PR #4, opened 2016-10-20
Update consensus_FAQ.md: Changed Text mentioning "Three" to "Two" algorithms

msoumeit:
## Description

The text says that there are three consensus implementations but actually there are only two implementations as of now.
## Motivation and Context

The change improves the documentation.

Signed-off-by: Sumit Mehrotra sumit@sumitmehrotra.com


hurf:

PR #30, opened 2017-03-05
Typos in docs: fabric_model

cca88:
The :ref:`MSP`s  doesnt parse.
There is no SHA0, only SHA-256.

Small corrections.

## Motivation and Context

Fixes #

Typos.

## Checklist:

Signed-off-by:

### DIFF
--- a/docs/source/fabric_model.rst
+++ b/docs/source/fabric_model.rst
@@ -53,7 +53,7 @@ channel. Each peer maintains a copy of the ledger for each channel of which they
 - Prior to appending a block, a versioning check is performed to ensure that states for assets that were read have not changed since chaincode execution time
 - There is immutability once a transaction is validated and committed
 - A channel's ledger contains a configuration block defining policies, access control lists, and other pertinent information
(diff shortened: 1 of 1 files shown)


cca88:
Actually this sentence here is totally unclear:

``To further obfuscate the data, values within chaincode can be encrypted (in part or in total) using common cryptographic algorithms such as SHA0-256, etc. before appending to the ledger.``

because a hash function like "SHA256" does not "encrypt".  What is meant by this?

hacera-jonathan:
Hi @cca88 (Christian),

Yes, I agree. This needs much more work!

BTW: Just for context: *while good cryptographic hash functions are deterministic, quick to compute and have the additional properties of being collision resistant and pre-image and second-pre-image resistant... they are designed to provide/guarantee integrity and not necessarily secrecy or confidentiality*.

So, I'm with you - these hashes are not used just for further data obfuscation [or "should not be used for data obfuscation", I hope? ;-)]

Also, what's "SHA0-256" ?

Thanks, Jonathan

ghost:

PR #73, opened 2017-10-06
MacOS users need to edit the query.js file

bennythejudge:
## Description
query.js uses `localhost` in `network_url` but this won't work on MacOS.
One needs to use the IP address of the VM used by docker-machine.

## Motivation and Context
Running query.js as it is fails on MacOS

## How Has This Been Tested?
Tested on my Mac laptop

## Checklist:

Signed-off-by: bennythejudge


bennythejudge:
I need to do more reading about the process to submit PRs.

PR #77, opened 2017-10-26
Modify regular expressions to be compatible with backslash for compatible windows

kutenglaoshu:
if header.Name is 'src\\chaincode\\main.go', it will check failed, in order to compatible windows path, should modify regular expressions from `(/)?src/.*` to `(/)?src/|\\.*`


ghost:

hpurmann:
Hey @rjones-lf!

Your link to zephyrproject doesn't work anymore (yields a 404). I've already created a LF account using my GitHub login. Why is this bad / not recommended? What can I do to fix whatever problem there might be?

PR #79, opened 2017-10-31
change chaincode path

Fatekiller23:
use original chaincode path, can not find corresponding go source file.
and i find it in "github.com/hyperledger/fabric/examples/chaincode/go/chaincode_example02/"
may be this is the correct path?
after i use this path, i can install chaincode.

## Description

image

image


ghost:

PR #82, opened 2017-11-29
Correct the URL for the bootstrap bash script

FelixFan1122:
The URL will retrieve the HTML page, not the bash file. Updating it with the one pointing at the raw file.

## Motivation and Context

Fixes #

## Checklist:

Signed-off-by:


milgner:
`https://github.com/hyperledger/fabric/raw/master/scripts/bootstrap.sh` works as well. It's the link behind the "Raw" button.

ryjones:

PR #91, opened 2017-12-29
CLI duration incorrectly documented

ronaldpetty:
byfn.sh help shows 10000 milliseconds and the docs show 10000 seconds, for humans, changed 10000 to 10.

## Motivation and Context

Fixes #

## Checklist:

Signed-off-by:

### DIFF
--- a/docs/source/channel_update.rst
+++ b/docs/source/channel_update.rst
@@ -126,7 +126,7 @@ variables.
 
   docker exec -it cli bash
 
-By default the CLI container exits after 10000 seconds.  If the container has
+By default the CLI container exits after 10 seconds.  If the container has
 exited, make sure to restart it before continuing.  First, check the status of
 your containers:
 


ronaldpetty:
Actually the timeout is incorrectly documented in many places including the byfn.sh.  Need to triple check all the places first.

ronaldpetty:
note to self, might consider changing the timeout as follows (since no time works for everyone):

* cmd; sleep $TIMEOUT becomes
* cmd; tail -f /dev/null (but compose does equiv of -itd + docker wait so the term waits just long enough)

christo4ferris:
@ronaldpetty Thanks for submitting this fix. Apologies for any delay in responding, apparently the bot that monitors PRs has fallen and cannot get up;-) We manage change to Hyperledger Fabric and the various related repos via Gerrit. The contribution process is defined in the [CONTRIBUTIONS](http://hyperledger-fabric.readthedocs.io/en/master/CONTRIBUTING.html) document. Please close this and resubmit the change via Gerrit. Looks like a clean fix. Thanks.

PR #105, opened 2018-02-04
updated the doc with correct CLI default exit time

vikas1188:
## Description
The default CLI exit time is 10 sec as per the configuration while following this document. & I found this as a big issue and very hard to notice and can be a big impediment for anyone who is following this doc. 
If you want to update this information by yourself, you can do it in your language. I've added the important details already as an edit to this document.

## Motivation and Context
It helps developer to know to go to correct path. CLI container was exiting the moment I was starting it, because the default time was 10 sec. If I were to knew it, I would not have wasted 4 hrs to this. I don't want anyone else to face this issue that they can just add the delay while using the ``./byfn.sh`` script.

Fixes #

## How Has This Been Tested?
It is just a doc update.

## Checklist:

- [x] I have added a [Signed-off-by](https://github.com/hyperledger/fabric/blob/master/CONTRIBUTING.md#legal-stuff).

Signed-off-by: Vikas <vikas.singh1188@gmail.com>

### DIFF
--- a/docs/source/channel_update.rst
+++ b/docs/source/channel_update.rst
@@ -126,9 +126,13 @@ variables.
 
   docker exec -it cli bash
 
-By default the CLI container exits after 10000 seconds.  If the container has
-exited, make sure to restart it before continuing.  First, check the status of
-your containers:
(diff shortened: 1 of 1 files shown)


ryjones:

PR #121, opened 2018-03-21
identity cleanup 

joealewine:


ryjones:

joealewine:
Thanks, Ry. Totally did this by mistake -- was trying to update my own fork.

PR #122, opened 2018-03-21
Update chaincode4ade.rst

gandalf15:
There is a wrong port. In the previous version, there was correct 7051. Moreover, I wondered why the query works if we do not have query function. I did not immediately notice it.


ryjones:

PR #134, opened 2018-03-30
Update membership.md

ronaldpetty:
Typo in name of Local MSP.

Signed-off-by: Ronald Petty <ronald.petty@rx-m.com>


ronaldpetty:
@bramwelt any idea why the DCO is not working here?  I am new to it, so probably missed something.  Thanks for the help.

PR #135, opened 2018-04-05
Update Vagrantfile

haseeb1431:
I was trying to download and it is downloading from atlash.hashicorp which has been retired on March 30. I just found a way to update the URL as follow. If there is a better solution please advise.


ryjones:
Thanks for the contribution! this is a read-only mirror. please read the [contribution guide](http://hyperledger-fabric.readthedocs.io/en/release-1.1/CONTRIBUTING.html).

PR #136, opened 2018-04-05
Fixed: Typo error

0xbuidler:


0xbuidler:
My bad, I didn't see that it was read-only :)

PR #148, opened 2018-06-03
Updated membership.md with some tiny tweaks.

aracataca:
Some tiny tweaks: 'a apple' to 'an apple' sort of tweaks.


aracataca:
MSPs seem quite complex. Might I suggest more pictures perhaps?

PR #164, opened 2018-08-09
Decrease log level of a normal behaviour

edevil:
When we are iterating over a MockStateRangeQueryIterator, the Next() method will set Current to nil after the last element is processed.


ryjones:
Thanks for the contribution! However, this is a read-only mirror. please read the [contribution guide](http://hyperledger-fabric.readthedocs.io/en/release-1.1/CONTRIBUTING.html). Also, feel free to join us on [chat!](https://chat.hyperledger.org/channel/fabric)

edevil:
@ryjones I've submitted a CR in Gerrit: https://gerrit.hyperledger.org/r/#/c/25790/

PR #182, opened 2018-09-14
fix a bug when create a channel the orderer may occur error

lq186:
when create a channel, the first block may not write to the ledger yet, and the ledgerResources.Height() may got 0, as a result the ledgerResources.Height() - 1 got 18446744073709551615, last block got nil and 
create channel error

### DIFF
--- a/orderer/common/multichannel/chainsupport.go
+++ b/orderer/common/multichannel/chainsupport.go
@@ -16,6 +16,7 @@ import (
 	"github.com/hyperledger/fabric/protos/utils"
 
 	"github.com/pkg/errors"
+	"time"
 )
 
 // ChainSupport holds the resources for a particular channel.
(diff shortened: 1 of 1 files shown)


kletkeman:
Are you perhaps accidentally swallowing errors? The easy way of introducing that type of bug is to use async await in your app without surrounding the call with a try catch block. There is no other way to trap errors thrown by fabric in the node SDK if you like the async await programming model.

 

The other possibility is that you are catching a race condition by talking to fabric from another thread after the channel is known but before the config block is committed. That too would be a bug since you clearly must wait for confirmation that the operation is complete at the client app that created the channel.

lq186:
Thank you very much. The second possibility is right. I accessed the ledger resource before it written to file.

PR #187, opened 2018-09-20
Update identity.md

richbl:
Grammar error: missing determiner (the object "it", in reference to "the message").


richbl:
Will try again with proper DCO in place

PR #192, opened 2018-10-04
FAB-12264

ghost:
PR with the solution of the issue FAB-12264. 
Basically added the command ```export``` in front of some env variables declaration at the 

> build your first network

 tutorial


ghost:
https://jira.hyperledger.org/browse/FAB-12264

PR #211, opened 2019-01-15
when i use Pulling from hyperledger/fabric-tools,it always shut down in the middle process.I dont konw why...

asn417:
how to download the v1.0


GoodTimeNET:
use git download it,and in the fabric floder  invoke  command  git chechout release-1.0

PR #217, opened 2019-01-30
[update] bootstrap.sh dsb option

iwin2471:
I saw dsb switch case block and  I thought cases should be true like this


iwin2471:
nope I was wrong I thought not a bypass option lol

PR #245, opened 2019-05-14
Fix a wrong word.

yyri:
They key point of understanding is that policy change is managed by a policy within the policy itself. 
"They" should be "The"


ryjones:
Thanks for the contribution! However, this is a read-only mirror. please read the [contribution guide](http://hyperledger-fabric.readthedocs.io/en/release-1.4/CONTRIBUTING.html). Also, feel free to join us on [chat!](https://chat.hyperledger.org/channel/fabric)

PR #266, opened 2019-07-16
Update build_network.rst

gnorbsl:
docker-compose would throw errors that IMAGE_TAG has not been set and would default to "". Added export for IMAGE_TAG before calling docker-compose to start the network

### DIFF
--- a/docs/source/build_network.rst
+++ b/docs/source/build_network.rst
@@ -523,7 +523,7 @@ First let's start our network:
 
 .. code:: bash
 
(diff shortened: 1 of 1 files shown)


ryjones:
Thanks for the contribution! However, this is a read-only mirror. please read the [contribution guide](http://hyperledger-fabric.readthedocs.io/en/release-1.4/CONTRIBUTING.html). Also, feel free to join us on [chat!](https://chat.hyperledger.org/channel/fabric)

PR #293, opened 2019-11-17
Update prereqs.rst

dehekkendekrekker:
I added this change because I'm using a new minimal image to build hyperledger upon, that does not come with git preinstalled. When running the curl command, the script borked. I therefore thought it was a good idea to add it to the prerequisits page.


lindluni:
@parall4x We are migrating Fabric from Gerrit to GitHub this week, if you'd like keep this open instead of using Gerrit that is perfectly fine. Once we finish the migration the maintainers will review this pull request

dehekkendekrekker:
@btl5037 I got your message too late, so closed it. I've used the gerrit route to submit the change instead.

PR #304, opened 2019-11-22
[FAB-11241] Move Profiling init Service external

yeasy:
Now the profile enable flag is inside the function, moving it into the main process
will make the entire logic more explicit.

#### Type of change

#### Description

Now the profile enable flag is inside the function, moving it into the main process
will make the entire logic more explicit.

#### Related issues

https://jira.hyperledger.org/browse/FAB-11241

### DIFF
--- a/orderer/common/server/main.go
+++ b/orderer/common/server/main.go
@@ -236,7 +236,9 @@ func Main() {
 		go clusterGRPCServer.Start()
 	}
 
-	initializeProfilingService(conf)
+	if conf.General.Profile.Enabled {
+		go initializeProfilingService(conf)
+	}
 	ab.RegisterAtomicBroadcastServer(grpcServer.Server(), server)
 	logger.Info("Beginning to serve requests")
 	grpcServer.Start()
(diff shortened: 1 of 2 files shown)


ale-linux (dismissed):

orderer/common/server/main.go
```diff
@@ -372,13 +374,9 @@ func initializeLogging() {
 
 // Start the profiling service if enabled.
 func initializeProfilingService(conf *localconfig.TopLevel) {
-	if conf.General.Profile.Enabled {
-		go func() {
```
guoger, line 376: the UT `TestInitializeProfilingService` needs to be altered to accommodate this change - this becomes a blocking method now

guoger (changes requested):

orderer/common/server/main.go
mastersingh24: I believe you will need to modify https://github.com/hyperledger/fabric/blob/master/orderer/common/server/main_test.go#L76 to use a goroutine

mastersingh24 (changes requested):

orderer/common/server/main.go
yeasy: Nice catch!

mastersingh24 (commented):
LGTM - will wait for CI to pass

orderer/common/server/main_test.go
```diff
@@ -73,7 +73,7 @@ func TestInitializeProfilingService(t *testing.T) {
 		l.Close()
 		return l.Addr().String()
 	}()
-	initializeProfilingService(
+	go initializeProfilingService(
```
guoger, line 76: not yours, but i think this go routine has been leaking. probably worth opening a JIRA to clean up this test

guoger (approved):
LGTM, just some future improvement worth a new JIRA in backlog (help-wanted)

Also, it uses time.Sleep to wait for profiling service to start, where `Eventually` polling mechanism looks like a much better choice. Would be nice to note that in the JIRA opened

mastersingh24 (approved):

PR #305, opened 2019-11-22
FAB-16120 Adding IT for MSP inconsistence

DereckLuo:

#### Type of change

- Test update

#### Description
- Adding integration test to replace behave test in **fabric-test**. 
- Testing correct endorsement behavior when a peer with _member_ identity trying to endorse a chaincode with _OR (Org1.peer, Org2.peer)_ endorsement policy.
- Minor typo change in integration test.

#### Related issues
https://jira.hyperledger.org/browse/FAB-16120

### DIFF
--- a/integration/msp/msp_test.go
+++ b/integration/msp/msp_test.go
@@ -58,6 +58,12 @@ var _ = Describe("MSP identity test on a network with mutual TLS required", func
 		By("setting TLS ClientAuthRequired to be true for all peers and orderers")
 		network.ClientAuthRequired = true
 
+		By("disabling NodeOU for org2")
+		// Org2 Peer0 is used to test chaincode endorsement policy not satisfied due to peer's MSP
+		// does not define Node OU.
+		Org2 := network.Organization("Org2")
+		Org2.EnableNodeOUs = false
+
 		network.GenerateConfigTree()
 		network.Bootstrap()
 
@@ -66,7 +72,8 @@ var _ = Describe("MSP identity test on a network with mutual TLS required", func
 		process = ifrit.Invoke(networkRunner)
 		Eventually(process.Ready(), network.EventuallyTimeout).Should(BeClosed())
 
-		peer := network.Peers[0]
+		org1Peer0 := network.Peer("Org1", "peer0")
+		org2Peer0 := network.Peer("Org2", "peer0")
 		orderer := network.Orderer("orderer")
 
 		By("creating and joining channels")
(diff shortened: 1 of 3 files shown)


integration/msp/msp_test.go
```diff
@@ -58,6 +58,10 @@ var _ = Describe("MSP identity test on a network with mutual TLS required", func
 		By("setting TLS ClientAuthRequired to be true for all peers and orderers")
 		network.ClientAuthRequired = true
 
+		By("disabling NodeOY for org2, and setting all peers in org2 with member identity")
```
mastersingh24:
```suggestion
By("disabling NodeOU for org2, and setting all peers in org2 with member identity")
```

mastersingh24 (dismissed):

integration/msp/msp_test.go
```diff
@@ -58,6 +58,10 @@ var _ = Describe("MSP identity test on a network with mutual TLS required", func
 		By("setting TLS ClientAuthRequired to be true for all peers and orderers")
 		network.ClientAuthRequired = true
 
+		By("disabling NodeOU for org2, and setting all peers in org2 with member identity")
```
jyellick: What does "and setting all peers in org2 with member identity" mean?

jyellick (commented):
As best as I can tell, this test is verifying that if you set an endorsement policy to require a peer role, for an org which does not have node OUs enabled, that that org won't be able to validly endorse.  But it's not at all clear from the description of the test that this is the scenario being tested.

It actually seems like it's testing two scenarios, one, where the policy is not satisfied because the peer's signer cert does not have the peer node OU, and one where the policy is not satisfied because the peer's MSP does not define the peer node OU.  It would be nice to explicitly call this out in the test, and to name which orgs/peers are testing which scenarios.

integration/nwo/configtx_template.go
```diff
@@ -26,6 +27,20 @@ Organizations:{{ range .PeerOrgs }}
...
+      Rule: OR('{{.MSPID}}.member')
+    Writers:
+      Type: Signature
+      Rule: OR('{{.MSPID}}.member')
+    Endorsement:
+      Type: Signature
+      Rule: OR('{{.MSPID}}.member')
+    Admins:
+      Type: Signature
+      Rule: OR('{{.MSPID}}.member')
```
jyellick: I don't think we want this.  Even without node-OUs, your admins should still require the admin role.

integration/nwo/configtx_template.go
```diff
@@ -50,13 +66,38 @@ Organizations:{{ range .PeerOrgs }}
...
```
jyellick: And same here

integration/nwo/configtx_template.go
```diff
@@ -50,13 +66,38 @@ Organizations:{{ range .PeerOrgs }}
...
+    {{- end }}
 {{ end }}
 {{- range .OrdererOrgs }}
 - &{{ .MSPID }}
   Name: {{ .Name }}
   ID: {{ .MSPID }}
   MSPDir: {{ $w.OrdererOrgMSPDir . }}
   Policies:
+  {{- if .EnableNodeOUs }}
+    Readers:
```
jyellick, line 91: I'd think we'd want to remove the '.peer' role from there, and add the .orderer role to both.

jyellick (changes requested):
The bulk of it looks good, but a few issues with the template

jyellick (approved):

jyellick (approved):

PR #307, opened 2019-11-22
[FAB-16978] MSP Key Concept Updates

pamandrejko:

#### Type of change
Documentation update for Key Concepts
- Documentation update

#### Description
Updating content for Fabric 2.0 Node OU support
#### Additional details
FAB-16978
We've received alot of feedback about MSP confusion and are trying to address it here as well as emphasize the new nodeOU support that was added. 

#### Related issues

---
This change is [Reviewable](https://reviewable.io/reviews/hyperledger/fabric/307)


docs/source/membership/membership.md
```diff
@@ -1,43 +1,86 @@
...
+identities. An MSP, on the other hand, is like the list of card providers
+accepted by the store, determining which identities are the trusted
+members (actors) of the store  payment network.
+
+![MSP1a](./membership.msp.diagram.png)
+
+*Identities are similar to your credit cards that are used to prove
+you can pay. The MSP is similar to the list of accepted credit cards.*
+
+Or consider a consortium of banks that each operate peer
```
nikhil550: I would rephrase the paragraph below to make it a bit easier to follow:

Or consider a consortium of banks that operate a blockchain network. Each bank would operate peer and ordering nodes, and endorse transactions submitted to the network. However, each bank would also have departments and account holders. The account holders would belong to each organization, but would not run nodes on the network. They would only interact with the system from their mobile or web application.

docs/source/membership/membership.md
```diff
@@ -1,43 +1,86 @@
...
+nodes that both endorse and validate transactions. They could also optionally
+operate ordering nodes. The Bank A account holders have identities affiliated
+with Bank A, and the Bank B account holders have identities affiliated
+with Bank B. But the account holders don't operate nodes, they simply
+interact with the system from their mobile or web application.
+
+So how does the network recognize and differentiate the identities?
```
nikhil550: I would switch the two senences to match the order in the introduction and order of importance:

So how does the network recognize and differentiate the identities? MSP's are used to define the organizations that belong to the network, whose members are trusted by other network members. MSPs are also the mechanism that provide members with a set of roles and permissions within the network.

docs/source/membership/membership.md
pamandrejko: Thanks. Done

docs/source/membership/membership.md
```diff
@@ -191,35 +311,77 @@ ORG2.MSP does exist in the network definition.*
...
+*The figure above shows the subfolders in a local MSP on the file system*
 
-As you can see, there are nine elements to an MSP. It's easiest to think of these elements
-in a directory structure, where the MSP name is the root folder name with each
-subfolder representing different elements of an MSP configuration.
+* **config.yaml:**  Used to configure the identity classification feature
+  in Fabric by enabling "Node OUs" and defining the accepted roles.
 
-Let's describe these folders in a little more detail and see why they are important.
+* **admincerts (Deprecated from Fabric v1.4.3 and higher):** This folder
```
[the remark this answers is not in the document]
pamandrejko: Yeah - We discussed this on the call and agreed on the call that deprecated was the term. We'll see what the others say.

docs/source/membership/membership.md
pamandrejko: Done.

docs/source/membership/membership.md
```diff
@@ -50,62 +94,200 @@ multiple MSPs and name them accordingly, e.g., `ORG2-MSP-NATIONAL` and
...
+who has administrative or participatory rights at that level (peer
+admins will not necessarily be channel admins, and vice versa).  This
+allows for authenticating member messages outside the context of a
+channel and to define the permissions over a particular node
+(who has the ability to install chaincode on a peer, for example).
+
+An orderer local MSP is also defined on the file system of the node
+and only applies to that node. Like peer nodes, orderers are also
+owned by a single organization and therefore have a single MSP to
+list the actors or nodes it trusts.
```
joealewine: Might also want to add a note here about ordering services that are owned by multiple organizations, as it will be a useful way to delineate some of these concepts. 

* Each node has a local MSP, defining the admins of that node. One or more nodes will be owned by an organization. This organization has an MSP defining its admins. The organization, the admin of the organization, the admin of the node, and the node itself should all have the same root of trust. 

* These ordering service organizations collectively run the system channel, which has a channel MSP, in which the organizations collectively make decisions relevant to the system channel (managing the consortium, mostly importantly, and the default policies which by default are inherited during the creation of a channel).

docs/source/membership/membership.md
```diff
@@ -50,62 +94,200 @@ multiple MSPs and name them accordingly, e.g., `ORG2-MSP-NATIONAL` and
...
-defined for it. Peers and orderers on a channel will all share the same view of channel
+channel level**. Peers and orderers on a channel share the same view of channel
 MSPs, and will therefore be able to correctly authenticate the channel participants.
 This means that if an organization wishes to join the channel, an MSP incorporating
 the chain of trust for the organization's members would need to be included in the
 channel configuration. Otherwise transactions originating from this organization's
-identities will be rejected.
+identities will be rejected. Whereas local MSPs are represented as a folder
+structure on the file system, channel MSPs are described in a channel
+config.json file.
```
joealewine: Better off just saying the channel configuration, which in its native state is a protobuf file contained in a block, not a json file with the metadata stripped out of it.

docs/source/membership/membership.md
```diff
@@ -50,62 +94,200 @@ multiple MSPs and name them accordingly, e.g., `ORG2-MSP-NATIONAL` and
...
+
+![MSP1d](./ChannelMSP.png)
+
+*Snippet from a channel config.json file that includes two organization
+MSPs.*
+
+**Channel MSPs identify who has permissions at a channel level**.
+
+The channel MSP defines the _relationship_ between the identities of
```
joealewine: See, I think "authority" could work here in the place of "permissions". It's not as good as "role", which implies something you're compelled to do, rather than just allowed to do, but it's a small step in the right direction.

docs/source/membership/membership.md
```diff
@@ -50,62 +94,200 @@ multiple MSPs and name them accordingly, e.g., `ORG2-MSP-NATIONAL` and
...
+channel members.
+
+**Every organization participating in a channel must have an MSP
+defined for it**. In fact, it is recommended that there is a one-to-one mapping
+between organizations and MSPs. If all network members are part of
+a single organization or MSP, data privacy is sacrificed. Multiple
+organizations facilitate privacy by segregating ledger data to only channel
+members. If more granularity is required within an
+organization, the organization can be divided into organizational units (OUs)
+as described above.
```
joealewine: I think the main point about organization MSPs is that they define who is empowered to act on behalf of the organization. This includes not just the interior configuration of the org as defined in the channel config (who the org admins are, for example), but also approving of administrative tasks in which the organization has a role (adding new members to the channel, for example). 

I know you say this below but I think it's the first/key point to highlight.

docs/source/membership/membership.md
```diff
@@ -50,62 +94,200 @@ multiple MSPs and name them accordingly, e.g., `ORG2-MSP-NATIONAL` and
...
 
-The key difference here between local and channel MSPs is not how they function
---- both turn identities into roles --- but their **scope**.
+**The organization MSP identifies the organization admins**. The org
+admin (identified in the MSP) has the right to act on behalf of the org.
+But the organization itself can have a limited set of permissions outlined in
+the channel MSP.
```
joealewine: This is a good point, and highlights an important distinction. The orgs in the channel get to decide what the policies are --- which orgs can do what. But it falls to the orgs themselves to decide who speaks for the org.

docs/source/membership/membership.md
```diff
@@ -50,62 +94,200 @@ multiple MSPs and name them accordingly, e.g., `ORG2-MSP-NATIONAL` and
...
+
+The key difference here between local and channel MSPs
+is not how they function -- both turn identities into roles -- but their **scope**.
+Each MSP lists roles and permissions at a particular level of administration.
```
joealewine: Exactly. This is the point that needs to be made early and often in this doc.

docs/source/membership/membership.md
```diff
@@ -132,94 +314,52 @@ operation, all organizations on the channel must agree to it. Therefore, the
...
-As you can see, there are nine elements to an MSP. It's easiest to think of these elements
-in a directory structure, where the MSP name is the root folder name with each
-subfolder representing different elements of an MSP configuration.
+*The figure above shows the subfolders in a local MSP on the file system*
 
-Let's describe these folders in a little more detail and see why they are important.
+* **config.yaml:**  Used to configure the identity classification feature
+  in Fabric by enabling "Node OUs" and defining the accepted roles.
+
+* **admincerts (Deprecated from Fabric v1.4.3 and higher):** This folder
```
joealewine: This information belongs with the rest of the implementation details, imo. And personally I might put it first in a section about implementation.

docs/source/membership/membership.md
```diff
@@ -132,94 +314,52 @@ operation, all organizations on the channel must agree to it. Therefore, the
...
+
+  **Note:** Prior to Fabric v1.4.3, admins were defined by explicitly putting certs in the
+  `admincerts` folder in the local MSP directory of your peer. **With Fabric v1.4.3 or higher,
+  certificates in this folder are no longer required.** Instead, it is recommended that when
+  the user is registered with the CA, that the `admin` Role is used to designate the node
+  administrator. Then, the identity is recognized as an `admin` by the Node OU value
+  in their signcert. As a reminder, in order to leverage the admin role, the "identity
+  classification" feature must be enabled in the config.yaml above by setting "Node OUs"
+  to `Enable: true`.
+
```
joealewine: Explicitly calling out and explaining this folder structure is AWESOME stuff.

docs/source/membership/membership.md
```diff
@@ -132,94 +314,52 @@ operation, all organizations on the channel must agree to it. Therefore, the
...
+  It's worth noting that just because an actor has the role of an administrator it doesn't
+  mean that they can administer particular resources! The actual power a given identity has
+  with respect to administering the system is determined by the policies that manage system
+  resources. For example, a channel policy might specify that `ORG1-MANUFACTURING`
+  administrators have the rights to add new organizations to the channel, whereas the
+  `ORG1-DISTRIBUTION` administrators have no such rights.
 
-* **Root CAs:** This folder contains a list of self-signed X.509 certificates of
+* **cacerts:** This folder contains a list of self-signed X.509 certificates of
   the Root CAs trusted by the organization represented by this MSP.
```
joealewine: Rather than calling out that certificates use the x.509 format, maybe just call it out once somewhere and just say "certificates" thereafter. Also, are root CA's necessarily self signed? Genuine question. And if so, is calling that out really necessary?

docs/source/membership/membership.md
```diff
@@ -132,94 +314,52 @@ operation, all organizations on the channel must agree to it. Therefore, the
...
   This is the most important folder because it identifies the CAs from
   which all other certificates must be derived to be considered members of the
-  corresponding organization.
-
+  corresponding organization to form the chain of trust.
 
-* **Intermediate CAs:** This folder contains a list of X.509 certificates of the
+* **intermediatecerts:** This folder contains a list of X.509 certificates of the
   Intermediate CAs trusted by this organization. Each certificate must be signed by
   one of the Root CAs in the MSP or by an Intermediate CA whose issuing CA chain ultimately
```
joealewine: Of any intermediate CAs.

docs/source/membership/membership.md
```diff
@@ -236,106 +376,75 @@ Let's describe these folders in a little more detail and see why they are import
   Like the Root CA folder, this folder defines the CAs from which certificates must be
   issued to be considered members of the organization.
 
+* **tlscacerts:** This folder contains a list of self-signed X.509 certificates of the
+  Root CAs trusted by this organization **for TLS communications**. An example of a TLS
```
joealewine: There isn't TLS communication, right? It's not a separate layer of communication (I don't think). It's just the certificates that go along with communication to secure it.

joealewine (commented):
Two additional points that might be worth making clear:

1. How signatures are attached to actions. 
2. How other members/peers/ordering service validate these signatures against their local copy of channel MSP.

docs/source/membership/membership.md
```diff
@@ -1,43 +1,86 @@
...
+a permissioned blockchain network. To transact on a Fabric network you
+need to:
+
+- Have an identity issued by a CA that is trusted by the network.
+- Become a member of an _organization_ that is recognized and approved
+by the network members. The MSP is how the identity is linked to the
+membership of an organization. Membership is achieved by adding your
+public key (also known as certificate, signing cert, or signcert) to
+your organization’s MSP.
+
```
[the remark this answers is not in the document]
pamandrejko: I changed it to `To transact on a Fabric network a member needs to:`

docs/source/membership/membership.md
```diff
@@ -50,62 +94,200 @@ multiple MSPs and name them accordingly, e.g., `ORG2-MSP-NATIONAL` and
...
-**Local MSPs are defined for clients (users) and for nodes (peers and orderers)**.
-Local MSPs define the permissions for that node (who the peer admins are, for example).
-The local MSPs of the users allow the user side to authenticate itself in its transactions
-as a member of a channel (e.g. in chaincode transactions), or as the owner of a
-specific role into the system (an org admin, for example, in configuration transactions).
+**Local MSPs are defined for clients and for nodes (peers and
+orderers)**.
+Local MSPs define the permissions for that node (who the peer admins
+are, for example). The local MSPs of clients (the account holders in
+the banking scenario above), allow the user to authenticate
```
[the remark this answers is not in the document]
pamandrejko: I moved it up. Let's see if that helps.

docs/source/membership/membership.md
pamandrejko: I moved it up to just after What is an MSP?

docs/source/membership/membership.md
pamandrejko: From the Fabric CA users guide:
The fabric-ca server init command generates a self-signed X.509 certificate. 
And is part of the CA server initialization process. 
After that it issues the certs.

I removed several of the X.509 references where they were redundant.

docs/source/membership/membership.md
pamandrejko: I changed it to **for secure communications between nodes using TLS**

docs/source/membership/membership.md
pamandrejko: I decided to remove this entire bullet and go with your earlier suggestion instead. `The MSP defines which members are empowered to act on behalf of the organization....`

docs/source/membership/membership.md
```diff
@@ -1,43 +1,86 @@
...
+*Identities are similar to your credit cards that are used to prove
+you can pay. The MSP is similar to the list of accepted credit cards.*
+
+Or consider a consortium of banks that each operate peer
+nodes that both endorse and validate transactions. They could also optionally
+operate ordering nodes. The Bank A account holders have identities affiliated
+with Bank A, and the Bank B account holders have identities affiliated
+with Bank B. But the account holders don't operate nodes, they simply
+interact with the system from their mobile or web application.
+
```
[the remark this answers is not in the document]
pamandrejko: I did add that transitional paragraph. Great suggestion. 

But regarding your second comment, I was only trying to convey that the bank has users with different types of identities and how do you differentiate them on the network. I feel like this level of detail is too much for the start of this topic?

docs/source/membership/membership.md
```diff
@@ -50,100 +93,175 @@ multiple MSPs and name them accordingly, e.g., `ORG2-MSP-NATIONAL` and
...
-node or user. However, as channel MSPs are available to all nodes in the
+node. However, as channel MSPs are available to all nodes in the
 channel, they are logically defined once in the channel configuration. However,
 **a channel MSP is also instantiated on the file system of every node in the
 channel and kept synchronized via consensus**. So while there is a copy of each
 channel MSP on the local file system of every node, logically a channel MSP
 resides on and is maintained by the channel or the network.
 
-### MSP Levels
+## MSP Levels
```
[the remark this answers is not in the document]
pamandrejko: I agree there is alot of overlap here. I merged the content of the 2 sections into 1 section now. ## MSP domains.

docs/source/membership/membership.md
```diff
@@ -50,100 +93,175 @@ multiple MSPs and name them accordingly, e.g., `ORG2-MSP-NATIONAL` and
...
-to reflect these separate lines of business. When a CA issues X.509 certificates,
-the `OU` field in the certificate specifies the line of business to which the
-identity belongs.
-
-We'll see later how OUs can be helpful to control the parts of an organization who
-are considered to be the members of a blockchain network. For example, only
-identities from the `ORG1-MANUFACTURING` OU might be able to access a channel,
-whereas `ORG1-DISTRIBUTION` cannot.
+### Organizational Units (OUs) and MSPs
+
```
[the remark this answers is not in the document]
pamandrejko: OK I added more color here to try to explain the difference. See if it helps.

docs/source/membership/membership.md
```diff
@@ -1,43 +1,86 @@
...
+membership of an organization. Membership is achieved by adding your
+public key (also known as certificate, signing cert, or signcert) to
+your organization’s MSP.
+
+## What is an MSP?
+
+Despite its name, the Membership Service Provider does not actually provide
+anything. Rather, the implementation of the MSP requirement is a folder
+that contains a list of related identities and is used to store users's
+certificates or in the case of an organization or channel, related identities.
```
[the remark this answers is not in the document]
pamandrejko: I changed it to `Rather, the implementation of the MSP requirement is a folder that contains a list of related identities and is used to store users's certificates or in the case of channel configuration, a list of organizations.`

docs/source/membership/membership.md
```diff
@@ -1,43 +1,86 @@
...
-permissions over a particular component (who has the ability to install chaincode on
-a peer, for example).
+member of a channel. It is the MSP that turns an identity into a **role** by
+identifying specific roles an actor might play within the scope of the
+organization the MSP represents (e.g., admins, or as members of a
+sub-organization group), and sets the basis for defining **access
+privileges** in the context of a network and channel e.g., channel admins,
+readers, writers). Also, when a user is registered with a CA, a role of
+admin, peer, client, orderer, or member must be associated with the
+user.
```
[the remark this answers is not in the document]
pamandrejko: The problem I've discovered is that before you can understand the folder structure you need to first understand Local and Channel MSPs and Organizations and OUs. So that's why the folder structure is last -- unfortunately.

docs/source/membership/membership.md
```diff
@@ -50,62 +94,200 @@ multiple MSPs and name them accordingly, e.g., `ORG2-MSP-NATIONAL` and
...
+file and contain a list of organizational units whose members are considered
+to be part of the organization represented by this MSP. This is particularly
+useful when you want to restrict the members of an organization to the ones
+holding an identity (signed by one of MSP designated CAs) with a specific
+Node OU role in it. For example, an endorsement policy can specify that
+only members of ORG1 can endorse a transaction. But when Node OU roles are
+enabled, a more granular restriction can be implemented,
+requiring that only peers in ORG1 can endorse a transaction instead of
+any member of ORG1.
+
```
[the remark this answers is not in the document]
pamandrejko: I'm a little reluctuant to introduce another term here, but I get what you are saying. Role is used by the Fabric CA so I was trying to stick with that since it does map -- especially along the lines of the OU discussion. At least that is where I am coming from

docs/source/membership/membership.md
```diff
@@ -50,62 +94,200 @@ multiple MSPs and name them accordingly, e.g., `ORG2-MSP-NATIONAL` and
...
-### Local and Channel MSPs
+## MSP domains
 
-MSPs appear in two types of places in a blockchain network:
+MSPs occur in two domains in a blockchain network:
 
-* In channel configuration (**channel MSP**)
 * Locally on an actor's node (**local MSP**)
+* In channel configuration (**channel MSP**)
+
```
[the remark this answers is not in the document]
pamandrejko: I like your suggestion and I tried moving all of this around, but ultimately I found that before you can understand the folder structure you have to to understand the difference between local MSPs, channel msps, organizations and OUs. It's the folder structure that ties all of that together.

docs/source/membership/membership.md
pamandrejko: incorporated this text into the section.

docs/source/membership/membership.md
pamandrejko: From the Fabric CA users guide:
The fabric-ca server init command generates a self-signed X.509 certificate.
And is part of the CA server initialization process.
After that it issues the certs.

I removed several of the X.509 references where they were redundant.

docs/source/membership/membership.md
pamandrejko: But ended up moving it back down because it depends on understanding local msps, channel msps, organizations and OU. :-(

docs/source/membership/membership.md
```diff
@@ -1,341 +1,197 @@
...
+### Local MSPs
+
+**Local MSPs are defined for clients and for nodes (peers and orderers)**.
+Local MSPs define the permissions for a node (who the peer admins are, for example). The local MSPs of clients (the account holders in the banking scenario above), allow the user to authenticate itself in its transactions as a member of a channel (e.g. in chaincode transactions), or as the owner of a specific role into the system (an org admin, for example, in configuration transactions).
+
+**Every node must have a local MSP defined**, as it defines who has administrative or participatory rights at that level (peer admins will not necessarily be channel admins, and vice versa).  This allows for authenticating member messages outside the context of a channel and to define the permissions over a particular node (who has the ability to install chaincode on a peer, for example). Note that one or more nodes can be owned by an organization. An MSP defines the organization admins. And the organization, the admin of the organization, the admin of the node, and the node itself should all have the same root of trust.
```
nikhil550, line 51: I would be a bit more clear here. The node MSP not only defines the identity of that node, it also identifies have the ability to operate that node.

docs/source/membership/membership.md
```diff
@@ -1,341 +1,197 @@
...
+
+Specifying OUs is optional. If OUs are not used, all of the identities that are part of an MSP --- as identified by the Root CA and Intermediate CA folders --- will be considered members of the organization.
+
+### Node OU Roles and MSPs
+
+Additionally, there is a special kind of OU, sometimes referred to as a `Node OU`, that can be used to confer a role onto an identity. These Node OU roles are defined in the `$FABRIC_CFG_PATH/msp/config.yaml` file and contain a list of organizational units whose members are considered to be part of the organization represented by this MSP. This is particularly useful when you want to restrict the members of an organization to the ones holding an identity (signed by one of MSP designated CAs) with a specific Node OU role in it. For example, an endorsement policy can specify that only members of ORG1 can endorse a transaction. But when Node OU roles are enabled, a more granular restriction can be implemented, requiring that only peers in ORG1 can endorse a transaction instead of any member of ORG1.
```
nikhil550: I think that this sentance can be shortened for clarity: For example, an endorsement policy can specify that only members of ORG1 can endorse a transaction. But when Node OU roles are enabled, a more granular restriction can be implemented, requiring that only peers in ORG1 can endorse a transaction instead of any member of ORG1.

> For example, with node OU's you can implement a more granular endorsement policy to requires Org1 peers to endorse a transaction, rather than any member of Org1

docs/source/membership/membership.md
```diff
@@ -1,341 +1,197 @@
...
+
+Let's explore the MSP elements that render the functionality we've described so far.
+
+A local MSP folder contains the following sub-folders:
+
+![MSP6](./membership.diagram.6.png)
+
+*The figure above shows the subfolders in a local MSP on the file system*
+
+* **config.yaml:**  Used to configure the identity classification feature in Fabric by enabling "Node OUs" and defining the accepted roles.
```
nikhil550, line 145: I understand that you are trying to match the image, but I think this whole section would read better if you were trying to rank by order of importance

- cacerts
- intermediate certs
- signcerts
- keystore
- admincerts
- config.yaml
- tlscacerts
- intermediatetls

docs/source/membership/membership.md
```diff
@@ -1,341 +1,197 @@
...
+
+**The system channel MSP includes the MSPs of all the organizations that participate in an ordering service.** An ordering service will likely include ordering nodes from multiple organizaitons and collectively these organizations run the ordering service, most importantly managing the consortium of organizations and the default policies that are inherited by the application channels.
+
+**Local MSPs are only defined on the file system of the node or user** to which they apply. Therefore, physically and logically there is only one local MSP per
+node. However, as channel MSPs are available to all nodes in the channel, they are logically defined once in the channel configuration. However, **a channel MSP is also instantiated on the file system of every node in the channel and kept synchronized via consensus**. So while there is a copy of each channel MSP on the local file system of every node, logically a channel MSP resides on and is maintained by the channel or the network.
```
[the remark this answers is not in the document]
pamandrejko: I did like this picture and the explanation that followed. Welcome your suggestions if you have something else in mind.

docs/source/membership/membership.md
```diff
@@ -1,341 +1,197 @@
...
+
+### Organizational Units (OUs) and MSPs
+
+An organization can also be divided into multiple **organizational units**, each of which has a certain set of responsibilities. Think of an OU as a department
+inside an organization. For example, the `ORG1` organization might have both `ORG1.MANUFACTURING` and `ORG1.DISTRIBUTION` OUs to reflect these separate lines of business. When a CA issues X.509 certificates, the `OU` field in the certificate specifies the line of business to which the identity belongs. These values can then be used in policy definitions in order to restrict access or in smart contracts for attribute-based access control.
+
```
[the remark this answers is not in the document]
pamandrejko: I added `A benefit of using OUs like this is that these values can then be used in policy definitions in order to restrict access or in smart contracts for attribute-based access control. Otherwise, separate MSPs would need to be created for each organization.`

docs/source/membership/membership.md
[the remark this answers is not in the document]
pamandrejko: One of the things I somehow missed when I was trying to understand MSPs is that  this config.yaml resides inside the /msp folder, so I feel it is important to make that clear here.

docs/source/membership/membership.md
pamandrejko: `Local MSPs define the permissions for a node (who are the peer admins who can operate the node, for example).`

docs/source/membership/membership.md
pamandrejko: I was trying to match the order they appear when you view the folder, but I can see your point. We can try that. I updated the image as well so the list and the image match.

docs/source/membership/membership.md
```diff
@@ -1,341 +1,197 @@
...
+![MSP1a](./membership.msp.diagram.png)
+
+*Identities are similar to your credit cards that are used to prove you can pay. The MSP is similar to the list of accepted credit cards.*
+
+Or consider a consortium of banks that operate a blockchain network. Each bank operates peer and ordering nodes, and the peers endorse transactions submitted to the network. However, each bank would also have departments and account holders. The account holders would belong to each organization, but would not run nodes on the network. They would only interact with the system from their mobile or web application. So how does the network recognize and differentiate these identities? MSP's are used to define the organizations that belong to the network, whose members are trusted by other network members. MSPs are also the mechanism that provide members with a set of roles and permissions within the network.
+
+Finally, consider if you want to join an _existing_ network, you need a way to turn your identity into something that is recognized
```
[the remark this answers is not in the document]
pamandrejko: `So how does the network recognize and differentiate these identities? A CA was used to create the identities, but like the card example, those identities cant just be issued, they need to be recognized by the network.`

docs/source/membership/membership.md
```diff
@@ -1,341 +1,194 @@
...
+
+![MSP1a](./membership.msp.diagram.png)
+
+*Identities are similar to your credit cards that are used to prove you can pay. The MSP is similar to the list of accepted credit cards.*
+
+Consider a consortium of banks that operate a blockchain network. Each bank operates peer and ordering nodes, and the peers endorse transactions submitted to the network. However, each bank would also have departments and account holders. The account holders would belong to each organization, but would not run nodes on the network. They would only interact with the system from their mobile or web application. So how does the network recognize and differentiate these identities? A CA was used to create the identities, but like the card example, those identities cant just be issued, they need to be recognized by the network. MSP's are used to define the organizations that are trusted by the network members. MSPs are also the mechanism that provide members with a set of roles and permissions within the network.
```
joealewine: "cant" should be "can't." Also should be MSPs, not "MSP's". 

I think this last sentence could be improved. You've gotten to the point where you've established MSPs conceptually and that they define an organization. The next logical step is to say that MSPs are also used to validate. Something like:

"Because the MSPs defining these organizations are known to the members of a network, they can then be used to validate that network entities that attempt to perform actions are allowed to."

docs/source/membership/membership.md
```diff
@@ -1,341 +1,197 @@
...
+
+Or consider a consortium of banks that operate a blockchain network. Each bank operates peer and ordering nodes, and the peers endorse transactions submitted to the network. However, each bank would also have departments and account holders. The account holders would belong to each organization, but would not run nodes on the network. They would only interact with the system from their mobile or web application. So how does the network recognize and differentiate these identities? MSP's are used to define the organizations that belong to the network, whose members are trusted by other network members. MSPs are also the mechanism that provide members with a set of roles and permissions within the network.
+
+Finally, consider if you want to join an _existing_ network, you need a way to turn your identity into something that is recognized
+by the network. The MSP is the mechanism that enables you to participate on a permissioned blockchain network. To transact on a Fabric network a member needs to:
+
+- Have an identity issued by a CA that is trusted by the network.
```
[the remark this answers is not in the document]
joealewine: I agree with Nik. There's not enough conceptual separation. Might be better as an ordered list rather than a conceptual list. Something like:

To transact on a Fabric network a member needs to:

1. Have an identity issued by a CA trusted by the network. 
2. Either create an MSP or have this identity added as an admin of the organization. 
3. Have the MSP added as a trusted member of the network. 
4. Make sure the configuration of the network includes this MSP. 

Then a link in #4 to the relevant section in the policies doc or updating a channel doc about how this is achieved.

docs/source/membership/membership.md
```diff
@@ -1,341 +1,194 @@
...
+
+Finally, consider if you want to join an _existing_ network, you need a way to turn your identity into something that is recognized by the network. The MSP is the mechanism that enables you to participate on a permissioned blockchain network. To transact on a Fabric network a member needs to:
+
+- Have an identity issued by a CA that is trusted by the network.
+- Become a member of an _organization_ that is recognized and approved by the network members. The MSP is how the identity is linked to the membership of an organization. Membership is achieved by adding the member's public key (also known as certificate, signing cert, or signcert) to the organization’s MSP.
+
+## What is an MSP?
+
+Despite its name, the Membership Service Provider does not actually provide anything. Rather, the implementation of the MSP requirement is a folder that contains a list of related identities and is used to store users's certificates or in the case of channel configuration, a list of organizations. Whereas Certificate Authorities generate the certificates that represent identities, the MSP contains a list of permissioned identities.
+
```
joealewine, line 30: Rather, MSPs are a set of folders that are added to the configuration of network and used to define an organization both inwardly (organizations decide who its admins are) and outwardly (allowing other organizations to validate that entities have the authority to do what they are attempting to do.

docs/source/membership/membership.md
```diff
@@ -1,341 +1,194 @@
...
+The MSP identifies which Root CAs and Intermediate CAs are accepted to define the members of a trust domain by listing the identities of their members, or by identifying which CAs are authorized to issue valid identities for their members.
+
+But the power of an MSP goes beyond simply listing who is a network participant or member of a channel. It is the MSP that turns an identity into a **role** by identifying specific privileges an actor has on a node or channel. Note that when a user is registered with a Fabric CA, a role of admin, peer, client, orderer, or member must be associated with the user.
```
joealewine: Last sentence --- might make sense to say that there are both roles and Fabric Roles. Something like:

This is related to the "role" explicitly defined by the Fabric CA when an identity is registered. Identities registered with the "peer" role will, naturally, be given to a peer. Similarly, identities registered with the "admin" role will be given to organization admins. All identities must be registered with a role, choosing between "admin", "peer", "client", "orderer", or "member". The structure of the channel MSP will then determine the role these identities have on the channel (for example, defining than an organization, and therefore its admin, must approve of changes to the channel configuration).

docs/source/membership/membership.md
```diff
@@ -1,341 +1,194 @@
...
+In addition, an MSP can allow for the identification of a list of identities that have been revoked --- as discussed in the [Identity](../identity/identity.html) documentation --- but we will talk about how that process also extends to an MSP.
+
+## MSP domains
+
+MSPs occur in two domains in a blockchain network:
 
-* In channel configuration (**channel MSP**)
 * Locally on an actor's node (**local MSP**)
+* In channel configuration (**channel MSP**)
+
```
joealewine, line 43: I think we need to add "organization" here. While it is contained in a channel configuration, it is, both conceptually and functionally, a separate layer of administration. Organizations can change the parameters within its MSP at will, without needing signatures from anyone else. This is separate from channel MSPs which determine permissions at the channel level (like the consensus type, etc). 

I would bring some of the material from the "what role does an organization play in an MSP" here and define it as a separate domain of administration (using some of the language above, if you want). Then I would link down to the section where you're talking down to OUs. It's a bit disjointed, but I think it will make sense as in this section we're talking about domains (levels of administration) while the OUs is more technically information about implementation. 

Ideally this information would be before the channel MSP stuff.

docs/source/membership/membership.md
```diff
@@ -1,341 +1,194 @@
...
+In the example above, there are 4 possible Node OU `ROLES` for the MSP:
+
+   * client
+   * peer
+   * admin
+   * orderer
+
+This convention allows you to distinguish MSP roles by the OU present in the CommonName attribute of the X509 certificate. The example above says that any certificate issued by cacerts/ca.sampleorg-cert.pem which has OU=client will identified as a client, OU=peer as a peer, etc. Starting with Fabric v1.4.3, there is also an OU for the orderer and for admins. The new admins role means that you no longer have to explicitly place certs in the admincerts folder of the MSP directory. Rather, the `admin` role present in the user's signcert qualifies the identity as an admin user.
+
```
joealewine, line 122: in which OU=client will be identified

docs/source/membership/membership.md
```diff
@@ -1,341 +1,194 @@
...
+These Role and OU attributes are assigned to an identity when the Fabric-CA or SDK is used to `register` a user with the CA. It is the subsequent `enroll` user command that generates the certificates in the users's /msp folder.
```
joealewine: I think we've basically standardized on Fabric CA and not Fabric-CA. 

Also should be users' and `/msp`

docs/source/membership/membership.md
```diff
@@ -1,341 +1,194 @@
...
+
+These Role and OU attributes are assigned to an identity when the Fabric-CA or SDK is used to `register` a user with the CA. It is the subsequent `enroll` user command that generates the certificates in the users's /msp folder.   
+
+![MSP1c](./ca-msp-visualization.png)
+
+The resulting ROLE and OU attributes are visible inside the X.509 signing certificate located in the /signcerts folder. The `ROLE` attribute is identified as `hf.Type` and  refers to an actor's role within its organization, (specifying, for example, that an actor is a `peer`). See the following snippet from a signing certificate shows how the Roles and OUs are represented in the certificate.
```
joealewine: /signcerts looks weird if it's not inside backtics, don't you think?

docs/source/membership/membership.md
```diff
@@ -1,341 +1,194 @@
...
+
+![MSP1d](./signcert.png)
+
+Finally, though this is a slight misuse of OUs, they can sometimes be used by different organizations in a consortium to distinguish each other. In such cases, the different organizations use the same Root CAs and Intermediate CAs for their chain of trust, but assign the `OU` field to identify members of each organization.
```
joealewine: I'd like to see some clarification on how this is misue (or whether it actually is). I know this language comes from the previous version of the doc but still. If it's misuse why say it?

docs/source/membership/membership.md
```diff
@@ -1,341 +1,194 @@
...
+
+  **Note:** Prior to Fabric v1.4.3, admins were defined by explicitly putting certs in the `admincerts` folder in the local MSP directory of your peer. **With Fabric v1.4.3 or higher, certificates in this folder are no longer required.** Instead, it is recommended that when the user is registered with the CA, that the `admin` role is used to designate the node administrator. Then, the identity is recognized as an `admin` by the Node OU role value in their signcert. As a reminder, in order to leverage the admin role, the "identity classification" feature must be enabled in the config.yaml above by setting "Node OUs" to `Enable: true`. We'll explore this more later.
+
+  It's worth noting for Channel MSPS that just because an actor has the role of an administrator it doesn't mean that they can administer particular resources. The actual power a given identity has with respect to administering the system is determined by the _policies_ that manage system resources. For example, a channel policy might specify that `ORG1-MANUFACTURING` administrators have the rights to add new organizations to the channel, whereas the `ORG1-DISTRIBUTION` administrators have no such rights.
+
```
joealewine, line 166: Channel MSPs. The point made by this paragraph is something that should be mentioned earlier in the section relating to the Channel MSP. I just made it in one of my earlier comments. Perhaps that language could be combined with this language.

docs/source/membership/membership.md
pamandrejko: I liked the clarification you suggested here and added it but left out the sentence about the channel MSP since we've not introduced that concept yet in the doc.

docs/source/membership/membership.md
pamandrejko: I actually had organization there and removed it due to a comment from another reviewer. But I do agree you have surfaced a gap and I like your idea. So to help tie all this together, under What role does an organization play in an MSP > Organizational Units (OUs) and MSPs> Node OU Roles and MSPs 
I added 
```**Note:** For Channel MSPs, just because an actor has the role of an administrator it doesn't mean that they can administer particular resources. The actual power a given identity has with respect to administering the system is determined by the _policies_ that manage system resources. For example, a channel policy might specify that `ORG1-MANUFACTURING` administrators, meaning identities with a role of `admin` and a Node OU of  `ORG1-MANUFACTURING`, have the rights to add new organizations to the channel, whereas the `ORG1-DISTRIBUTION` administrators have no such rights.```

Similar text is still repeated in the /admincerts folder discussion under the guise "As a reminder,.." since we can't assume people read the earlier text.

File: docs/source/membership/membership.md (hunk 22 above)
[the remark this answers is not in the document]
Comment (pamandrejko) [reply docs/source/membership/membership.md]: For your previous comment I added this:
How is that chain of trust used by the blockchain network?

Certificate Authorities issue identities by generating a public and private key which forms a key-pair that can be used to prove identity. Because a private key can never be shared publicly, a mechanism is required to enable that proof which is where the MSP comes in. For example, a peer uses its private key to digitally sign, or endorse, a transaction.  The MSP on the ordering service contains the peer's public key which is then used to verify that the signature attached to the transaction is valid. The private key is used to produce a signature on a transaction that only the corresponding public key, that is part of an MSP, can match. Thus, the MSP is the mechanism that allows that identity to be trusted and recognized by the rest of the network without ever revealing the member’s private key.

File: docs/source/membership/membership.md (hunk 39 above)
Comment (pamandrejko) [reply docs/source/membership/membership.md:166]: Combined with your earlier comment I added this information to the Organization and Node OUs section. I think that is the best place as it ties it all together there.

File: docs/source/membership/membership.md [hunk 40]
```diff
@@ -1,341 +1,197 @@
...
-network administration concerns** while **MSPs at a lower level handle
-identity for the administration of private resources**. MSPs are mandatory
-at every level of administration --- they must be defined for the network,
-channel, peer, orderer, and users.
+![MSP1d](./ChannelMSP.png)
+
+*Snippet from a channel config.json file that includes two organization
+MSPs.*
+
+**Channel MSPs identify who has authorities at a channel level**.
```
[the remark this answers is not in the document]
pamandrejko: I opened jira https://jira.hyperledger.org/browse/FAB-17335 to address this work.

docs/source/membership/membership.md
pamandrejko: I talked with Anthony to understand the origin of this comment and settled on this clarification:

Finally, OUs could be used by different organizations in a _consortium_ to distinguish each other. But in such cases, the different organizations have to use the same Root CAs and Intermediate CAs for their chain of trust, and assign the `OU` field to identify members of each organization. When every organization has the same CA or chain of trust, this makes the system more centralized than what might be desirable and therefore deserves careful consideration on a blockchain network.

denyeart:
@pamandrejko  Please squash the 4 commits together and ensure final commit is signed correctly so that DCO check can pass.

denyeart:
Looks like there is a conflict in docs/source/membership/membership.md.

sykesm (dismissed):
Waiting on @ale-linux for merge but looks good to me.

sykesm (approved):

PR #309, opened 2019-11-23
Fix the style-guides directory naming scheme

lindluni:
This change removes the duplicate style guide directory that is the same name as the other, just with a different case. This was introduced via a documentation cherry pick and causes folder reconciliation issues on case insensitive systems like OSX

#### Type of change

- Bug fix


manish-sethi (approved):

ghost (approved):

ghost (approved):

ale-linux (approved):

guoger:
https://gerrit.hyperledger.org/r/c/fabric/+/32812 magically prevents IT flake from happening for a long period on master branch. we probably need to cherry-pick that to 1.4

mastersingh24 (approved):

mastersingh24:
not sure why integration tests are timing out, but going to override and merge this given it's only a doc change

PR #312, opened 2019-11-23
[FAB-17132] Add Parallel Strategy to Integration Tests

lindluni:
**To see the results of this change look at the successful build of this PR**

https://dev.azure.com/Hyperledger/Fabric/_build/results?buildId=2527&view=results

#### Type of change

- Improvement to tests

#### Description

This change splits the integration suite into multiple parallel jobs.

Under this framework the build finishes in approximately 40 minutes as opposed to nearly 2 hours today which is in line with the time it takes to run unit tests

I've elected to remove haveged as it is not needed for our test suites, the BYFN E2E tests in Node and Java SDK repos require it. There is a bug in AZP where the request to APT to install deps returns a 503 frequently enough for it to be annoying and the builds fail. Since it adds nothing for us, just removing it to protect against that. The other deps happen to already be installed on the AZP-based nodes, so leaving them around as they are truly required, even if not needed right now. Future Pipelines that make use of Azure Scale Sets might need this.

Also switching to Ubuntu-18.04-based VM's, which in general in my tests seems to have better overall performance vs the 16.04 nodes

### DIFF
--- a/ci/azure-pipelines.yml
+++ b/ci/azure-pipelines.yml
@@ -16,7 +16,7 @@ variables:
 jobs:
   - job: VerifyBuild
     pool:
-      vmImage: ubuntu-16.04
+      vmImage: ubuntu-18.04
     steps:
       - template: install_deps.yml
       - checkout: self
@@ -27,7 +27,7 @@ jobs:
 
   - job: DocBuild
     pool:
-      vmImage: ubuntu-16.04
+      vmImage: ubuntu-18.04
     container:
       image: n42org/tox:3.4.0
     steps:
@@ -41,7 +41,7 @@ jobs:
 
   - job: UnitTests
     pool:
-      vmImage: ubuntu-16.04
+      vmImage: ubuntu-18.04
     steps:
       - template: install_deps.yml
       - checkout: self
@@ -54,8 +54,10 @@ jobs:
 
   - job: IntegrationTests
     pool:
-      vmImage: ubuntu-16.04
-    timeoutInMinutes: 120
+      vmImage: ubuntu-18.04
+    strategy:
+      parallel: 5
+    timeoutInMinutes: 90
     steps:
       - template: install_deps.yml
       - checkout: self
@@ -63,3 +65,4 @@ jobs:
         displayName: Checkout Fabric Code
       - script: make integration-test
         displayName: Run Integration Tests
+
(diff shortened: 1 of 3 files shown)


sykesm (dismissed):
There are less fragile ways to do this. What's being proposed makes it too easy to move a test suite around such that tests aren't run any longer. This also moves away from the makefile being the source of truth.

IMO, this is the wrong direction.

Further, if we're up to two hours, we have other problems to address.

lindluni:
I've updated to use the parallel strategy of AZP

Jobs have been taking about 01:40:00 for nearly two months now. And some of our current flakes cause deadlocks for 20 minutes which push us towards 2 hours before we fail

We can do something like this, which I didn't implement because we don't get to choose how we load balance easily. But I guess we can put split logic in the script that says if the number of agents is > 1 (meaning we arent running locally) we can assign the long running suites to particular agents, i.e (lifecycle to agent 1) (pvtdata and raft to agent 2) and then split the remaining between the last 2 nodes

lindluni:
This is obviously not the final solution, but hopefully this is closer to something you can get behind

lindluni:
While this solution isn't perfect, it does bring runtime under an hour, which is 20 minutes longer than the previous methodology, but ~40 minutes faster than we have today:

https://dev.azure.com/Hyperledger/Fabric/_build/results?buildId=2537&view=logs&j=17737827-9315-5d9e-c14a-16a17e90ef50

ci/azure-pipelines.yml
```diff
@@ -54,12 +54,15 @@ jobs:
 
   - job: IntegrationTests
     pool:
-      vmImage: ubuntu-16.04
-    timeoutInMinutes: 120
+      vmImage: ubuntu-18.04
+    strategy:
+      parallel: 5
```
guoger, line 59: question wrt azp - will this consume overall parallel job quota?

guoger (approved):

guoger:
@btl5037 i believe Matt's comment has been addressed (even though history is not visible during forced push) and can be marked resolved? i think we should get this merged so that other PRs can run faster...

ci/azure-pipelines.yml
lindluni: It does, but the majority of the parallel jobs finish in a rather short period, if we see that this is creating a backlog we can always adjust this number lower, but I don't expect it to happen.

lindluni:
Correct, Matt's comments were in regard to me hardcoding the directory splitting, this dynamically solves it and implements microsofts suggested framework for test splitting

mastersingh24 (approved):
This seems better than what we currently have to deal with

PR #313, opened 2019-11-24
[FAB-17133] Add Action to Trigger CI Builds

lindluni:
This change adds a GitHub Action that is used
to trigger CI builds when the PR is commented
on using "/ci-run"

You can test it out here: https://github.com/btl5037/fabric/pull/3
You can see the actual runs here: https://github.com/btl5037/fabric/actions

#### Type of change

- New feature

#### Description

This change adds a GitHub action to trigger CI builds on comment

Users can comment "/ci-run" on their PR conversation tab to retrigger CI. It first checks to see if an existing build is running before triggering the new build.

### DIFF
--- a/.github/workflows/trigger.yml
+++ b/.github/workflows/trigger.yml
@@ -1,0 +1,48 @@
+# Copyright the Hyperledger Fabric contributors. All rights reserved.
+#
+# SPDX-License-Identifier: Apache-2.0
+
+on:
+  issue_comment:
+    types: [created]
+name: Automatically Trigger Azure Pipeline
+jobs:
+  trigger:
+    name: TriggerAZP
+    if: github.event.issue.pull_request != '' && contains(github.event.comment.body, '/ci-run')
+    runs-on: ubuntu-latest
+    steps:
+    - name: Trigger Build
+      run: |
+        author=$(jq -r ".issue.user.login" "${GITHUB_EVENT_PATH}")
+        commenter=$(jq -r ".comment.user.login" "${GITHUB_EVENT_PATH}")
+        org=$(jq -r ".repository.owner.login" "${GITHUB_EVENT_PATH}")
+        pr_number=$(jq -r ".issue.number" "${GITHUB_EVENT_PATH}")
+        project=$(jq -r ".repository.name" "${GITHUB_EVENT_PATH}")
+        repo=$(jq -r ".repository.full_name" "${GITHUB_EVENT_PATH}")
+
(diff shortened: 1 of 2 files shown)


mastersingh24 (commented):
This is fine, but is it possible to actually just re-run failed jobs?

mastersingh24 (dismissed):

lindluni:
So it's not using the AZ CLI, but I had my moment of clarity when I stepped away last night and got some sleep, it looks like the rest API has an endpoint for retriggering just the failed builds, I am working on that now

denyeart:
> So it's not using the AZ CLI, but I had my moment of clarity when I stepped away last night and got some sleep, it looks like the rest API has an endpoint for retriggering just the failed builds, I am working on that now

Just curious - why is a github action comment for retriggering failed jobs better/different than clicking Re-Run on the failed job?
In general we'd like to write down and socialize all the common comment actions people should be using, so for consistency good to have comment actions. Just wondering if there is any difference relative to clicking Re-Run.

lindluni:
The only people authorized to use the re-run function or the /azp-run comment are repo maintainers/admins and people who have been added to the org in Azure DevOps (which is unfortunate). So the maintainers and admins can retrigger individual failed jobs, but there is no mechanism for an author who isn't a maintainer to retrigger ci on their own build.

@mastersingh24 I really dug into the the REST API today, there is no endpoint to re-trigger failed jobs only. Per above, as a maintainer you can do so via the checks tab, but the author cannot if they aren't a maintainer. I did open a ticket with Azure DevOp's to ask if and when this might be implemented and a feature request to add this. Unfortunately it's not perfect, but it's what we can do right now. Since author's can't re-trigger their CI on their own, this would give them some means to do so, even if not perfect. They only added rerun failed jobs last month, so its a fairly new feature, which is probably why there is nothing on the API for it.

lindluni:
I updated the job to only allow the author or those assigned as reviewers (which in our case is our maintainers) to allow the retriggering of the pipeline. It also doesn't retrigger if a build is already executing against it

mastersingh24 (approved):

kchristidis (approved):

kchristidis:
> Unfortunately it's not perfect, but it's what we can do right now. Since author's can't re-trigger their CI on their own, this would give them some means to do so, even if not perfect. They only added rerun failed jobs last month, so it's a fairly new feature, which is probably why there is nothing on the API for it.

Noted. Thanks for digging into it. This LGTM.

PR #315, opened 2019-11-25
Copy if rename fails during external build

guoger:
External builder attempts to persist build and release content
by renaming them to another dir from temp. However this would not
always work if temp dir is on a different mounted fs. This CR
changes it to try copy if rename fails.

FAB-17119

Change-Id: Ib6970807f637c7dd304efe66c479ce7a8b0bfa38

---
This change is [Reviewable](https://reviewable.io/reviews/hyperledger/fabric/315)

### DIFF
--- a/core/container/externalbuilder/copy.go
+++ b/core/container/externalbuilder/copy.go
@@ -1,0 +1,108 @@
+/*
+Copyright IBM Corp. All Rights Reserved.
+
+SPDX-License-Identifier: Apache-2.0
+*/
+
+package externalbuilder
+
+import (
+	"io"
+	"os"
+	"path/filepath"
+
+	"github.com/hyperledger/fabric/common/flogging"
+	"github.com/pkg/errors"
+)
+
+// MoveOrCopyDir attempts to copy src to dest by firstly trying to move, then copy upon failure.
+func MoveOrCopyDir(logger *flogging.FabricLogger, srcroot, destroot string) error {
+	mvErr := os.Rename(srcroot, destroot)
+	if mvErr == nil {
+		return nil
+	}
+
+	logger.Debugf("Failed to move %s to %s: %s, try copy instead", srcroot, destroot, mvErr)
+
+	info, err := os.Stat(srcroot)
+	if err != nil {
+		return errors.WithMessagef(err, "failed to stat dir: %s", srcroot)
+	}
+
+	if err = os.MkdirAll(destroot, info.Mode()); err != nil {
+		return errors.WithMessagef(err, "failed to make dir: %s", destroot)
+	}
+
+	cpErr := CopyDir(srcroot, destroot)
+	if cpErr == nil {
+		return nil
+	}
+
+	logger.Errorf("Failed to copy %s to %s: %s", srcroot, destroot, cpErr)
+
+	rmErr := os.RemoveAll(destroot)
+	if rmErr != nil {
+		logger.Errorf("Failed to clean targeting dir %s: %s", destroot, rmErr)
+	}
+
+	return errors.WithMessagef(cpErr, "failed to copy %s to %s", srcroot, destroot)
+}
+
+// CopyDir creates a copy of a dir
+func CopyDir(srcroot, destroot string) error {
+	return filepath.Walk(srcroot, func(path string, info os.FileInfo, err error) error {
+		if err != nil {
+			return err
+		}
+
+		srcsubpath, err := filepath.Rel(srcroot, path)
+		if err != nil {
+			return err
+		}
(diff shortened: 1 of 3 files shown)


lindluni:
@guoger Can you rebase this on master to pick up the new parallel strategy from Integration Tests.

lindluni:
Good news is, the force push looks like it cleared out that stray extra Integration Test check that was causing problems. Azure is investigating where that came from.

core/container/externalbuilder/copy.go
```diff
@@ -0,0 +1,106 @@
...
+
+		if info.IsDir() { // its a dir, make corresponding dir in the dest
+			if err = os.MkdirAll(destpath, info.Mode()); err != nil {
+				return err
+			}
+			return nil
+		}
+
+		// its a file, copy to corresponding path in the dest
+		if err = copyFile(path, destpath); err != nil {
```
jyellick, line 74: Just to confirm, this is safe (we are ensured the intermediate directories exist) because filepath.Walk gives us entries in lexical order, and a parent directory is always lexically before any of its child elements.

core/container/externalbuilder/copy.go
```diff
@@ -0,0 +1,106 @@
...
+			return err
+		}
+		return nil
+	})
+}
+
+func copyFile(srcpath, destpath string) error {
```
jyellick, line 81: This doesn't seem to appropriately handle sym-links.  Maybe that's okay, but seems like a notable limitation.  But I'd welcome feedback.

https://stackoverflow.com/questions/51779243/copy-a-folder-in-go pointed to https://github.com/otiai10/copy/blob/master/copy.go which seemed instructive.

core/container/externalbuilder/copy.go
```diff
@@ -0,0 +1,106 @@
...
+)
+
+// MoveOrCopyDir attempts to copy src to dest by firstly trying to move, then copy upon failure.
+func MoveOrCopyDir(logger *flogging.FabricLogger, srcroot, destroot string) error {
+	mvErr := os.Rename(srcroot, destroot)
+	if mvErr == nil {
+		return nil
+	}
+
+	logger.Warnf("Failed to move %s to %s: %s, try copy instead", srcroot, destroot, mvErr)
```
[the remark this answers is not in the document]
guoger: good point! will do

core/container/externalbuilder/copy.go
guoger: right, this relies on the fact that parent node is always visited before children, and we call `mkdir` to prepare the path. thx for bringing this up, i'll add comments

core/container/externalbuilder/copy.go
guoger: yes, i was thinking of symlinks but not sure if that's necessary in our case. wdyt?

core/container/externalbuilder/copy.go
```diff
@@ -0,0 +1,106 @@
...
+
+	info, err := os.Stat(srcroot)
+	if err != nil {
+		return errors.WithMessagef(err, "failed to stat")
```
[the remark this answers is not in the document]
guoger: my fault, i was intending to add path to error messages

core/container/externalbuilder/copy.go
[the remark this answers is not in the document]
guoger: good point!

core/container/externalbuilder/copy.go
guoger: right, i think this relies on the fact that parents are always visited before children. will add comments there. thx for bringing this up!

core/container/externalbuilder/copy.go
guoger: i was thinking of symlinks but not sure if symlink is needed in our case, wdyt?

core/container/externalbuilder/copy.go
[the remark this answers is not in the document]
guoger: good catch, i forgot to add dir path to error message

core/container/externalbuilder/copy.go
jyellick: @sykesm Do you think this is something we can punt on, or should try to support out of the gate?

jyellick: Discussed this with @sykesm and not handling symlinks should be fine for now.  If the link is relative and to a file, then we should get normal copy behavior.  There's no obvious reason to expect symlinks in the build output.

jyellick (dismissed):

jyellick (approved):

denyeart (approved):

jyellick (approved):

PR #317, opened 2019-11-25
Add Chris Gabriel as a Documentation maintainer

pamandrejko:

#### Type of change

- Documentation update

#### Description

Chris has been an instrumental member of our Fabric Documentation community for several years now, contributing content, reviewing content, and as a user of Fabric in his own Hyperhchain Labs business which he founded. 

With his breadth of Fabric experience, we'd like to have him become a Documentation maintainer.

### DIFF
--- a/MAINTAINERS.md
+++ b/MAINTAINERS.md
@@ -18,7 +18,8 @@ Maintainers
 | Manish Sethi   |  manish-sethi | manish-sethi | manish-sethi | <manish.sethi@gmail.com>
 | Matthew Sykes  |  sykesm    |   sykesm  |  sykesm   | <sykesmat@us.ibm.com>
 | Pam Andrejko  |   pama-ibm  |   pamandrejko | pandrejko | <pama@ibm.com>
-  Srinivasan Muralidharan | muralisr | muralisrini | muralisr | <srinivasan.muralidharan99@gmail.com>
+| Chris Gabriel |  cmgabriel | denali49 | cmgabriel | <chris@hyperchainlabs.com>
+| Srinivasan Muralidharan | muralisr | muralisrini | muralisr | <srinivasan.muralidharan99@gmail.com>
 | Yacov Manevich |  yacovm    |   yacovm   |   yacovm   | <yacovm@il.ibm.com>
 
 **Release Managers**
@@ -40,5 +41,3 @@ Maintainers
 | Yaoguo Jiang   |  jiangyaoguo | jiangyaoguo | jiangyaoguo | <jiangyaoguo@gmail.com>
 | Greg Haskins  |  greg.haskins | ghaskins  |  ghaskins | <gregory.haskins@gmail.com>
 | Keith Smith  |   smithbk   |  smithbk  |  smithbk  | <bksmith@us.ibm.com>
-
-


jmason90:
In addition to his work building his company, Hyperchain Labs, based on the Fabric platform and promoting Fabric around the world with customers and at conferences, Chris been helping to build the documentation for a long time. His Fabric implementations have been shown as demos many times to the documentation team to help provide guidance on how Fabric features can be used effectively.

yacovm:
Here is an idea - why don't we just make a docs repository where all the docs can be managed, and extract all documentation away from the fabric core repository?

lindluni:
We did talk about this today, we have https://github.com/hyperledger-archives/fabric-docs and could resurrect it from the archives. Chris started it years ago and it was never followed up on

C0rWin:
Reviving fabric-docs and extracting documentation to separate repo and making people who involved on documentation front be in charge of this repo makes more sense.

denyeart:
Our intent has been to split Fabric docs into its own repository after the transition to GitHub and Azure Pipelines. We only have one more repository to switch over (fabric-test), then will proceed with the documentation split.

I'd suggest that we identify the top documentation contributors and reviewers to seed the fabric docs repository maintainer list in the coming weeks, rather than pulling the trigger in the Fabric repository this week.

guoger:
I recall that splitting docs from main repo was proposed and debated a
while back, and decision was made *not* to split. Not sure what have
changed since then but I definitely like the idea of splitting and it would
make translation easier.

- J

sykesm:
Instead of extracting doc because it’s too hard to get changes merged, why not fix the review and merge process?

Once doc is extracted:

- tools to manage reference documentation (gendoc, cli help, etc) are no longer useful
- docs will skew away from the code
- maintainers will only gain responsibility to poll doc changes in a separate repo

denyeart:
@sykesm @guoger Deeper discussion ongoing in the mailing list, let's take the discussion there.

yacovm:
> Instead of extracting doc because it’s too hard to get changes merged, why not fix the review and merge process?
> 
> Once doc is extracted:
> 
>     * tools to manage reference documentation (gendoc, cli help, etc) are no longer useful
> 
>     * docs will skew away from the code
> 
>     * maintainers will only gain responsibility to poll doc changes in a separate repo

The reason to extract documentation to another repository is not because it's too hard to get changes merged, it's because _under the assumption_ that documentation maintainers should be a thing, they **should not** be maintainers in the fabric core repository. 

You cannot automate **meaningful** documentation, and **any** documentation you can automate in one repository you can **also** automate across several repositories. 

There is no mechanized way in which you can enforce that the documentation really documents the code, from obvious reasons (HP and such). We can only rely on the code contributors to update the documentation, or go over the documentation and update it if they don't.

christo4ferris (approved):
+1

ryjones:
it looks like a) this has stalled b) it has a merge conflict due to my shenanigans. I suggest this is rebased and merged. It doesn't seem like there is much pushback. Also - I suggest either a new section that calls these out functionally. Please alpha order additions.

pamandrejko:
Thanks Ry - we are getting ready to propose a new process for merging documentation-only content into Fabric - so we will either re-use this issue or create a new one. Hope to resolve shortly.

denyeart:
Given the updated direction for doc approvers/maintainers, let's close this one for now, it can always be re-opened if we want to re-use it for some reason.

PR #321, opened 2019-11-27
FAB-17161 improve error message

guoger:
Combination of OUs is not allowed, where error message returned
contains address of OU object, which is not human friendly. This
commit changes it to print actual OU identifiers in []string.

Test done: use disallowed combination of OUs for a Fabric network
and observe error message returned to contain clear text.

Change-Id: Ie7fd19e6542b7b78a736e5b4b6092ad637e37152

### DIFF
--- a/msp/identities.go
+++ b/msp/identities.go
@@ -12,6 +12,7 @@ import (
 	"crypto/x509"
 	"encoding/hex"
 	"encoding/pem"
+	"fmt"
 	"sync"
 	"time"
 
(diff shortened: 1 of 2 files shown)


msp/identities.go
```diff
@@ -107,6 +108,17 @@ func (id *identity) Validate() error {
 	return id.msp.Validate(id)
 }
 
+type OUIDs []*OUIdentifier
+
+func (o OUIDs) String() string {
+	var res []string
+	for _, id := range o {
+		res = append(res, id.OrganizationalUnitIdentifier)
```
ale-linux: You might also want to print a hex encoding of CertifiersIdentifier

ale-linux (dismissed):

msp/identities.go
guoger: thx for review! i wonder how that would help debugging log message?

ale-linux: The presence of CertifiersIdentifier determines which certification paths can issue valid OUs, so it will be helpful to have that information imho

guoger: updated it to include a short hex code

ale-linux: Thanks, LGTM

ale-linux (approved):

kchristidis (approved):

caod123:

PR #325, opened 2019-11-28
FAB-17158 ignore util test using build tag

guoger:
ledger_generate_test is used to generate test data, and is currently
skipped by default. However, it still runs BeforeEach and AfterEach,
which brings up a Fabric network, which is heavy, and leaves garbage
data dir behind. This patch adds a build tag 'generate' to file.

Test done: run `ginkgo -tags generate` to see that test being executed,
and `ginkgo` to see it being excluded.

### DIFF
--- a/integration/ledger/ledger_generate_test.go
+++ b/integration/ledger/ledger_generate_test.go
@@ -1,3 +1,5 @@
+// +build generate
+
 /*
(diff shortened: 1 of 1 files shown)


guoger:
cc @wenjianqiao

denyeart (approved):

manish-sethi (approved):

PR #328, opened 2019-11-29
fix FAB-17148 defer to delete 'bccsp-sw' directory

baidang201:
Signed-off-by: baidang201 <jinrishouji@sina.com>
Type of change
Bug fix

Description
fix when unit test in bccsp\sw，we find in the /tmp folder, has a lot of file like 'bccsp-sw' and can not delete auto by test program.

Additional details
the reason is when os.exit, it do not run defer to delete 'bccsp-sw' folder.
we can put os.exit in the head of function and import another defer for os.exit.

Related issues
https://jira.hyperledger.org/browse/FAB-17148

### DIFF
--- a/bccsp/sw/impl_test.go
+++ b/bccsp/sw/impl_test.go
@@ -56,6 +56,10 @@ func (tc testConfig) Provider(t *testing.T) (bccsp.BCCSP, bccsp.KeyStore, func()
 }
 
 func TestMain(m *testing.M) {
+	code := -1
+	defer func() {
+		os.Exit(code)
+	}()
 	tests := []testConfig{
 		{256, "SHA2"},
 		{256, "SHA3"},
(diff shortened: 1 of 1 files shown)


bccsp/sw/impl_test.go
```diff
@@ -76,10 +80,10 @@ func TestMain(m *testing.M) {
 		ret := m.Run()
```
guoger: this could be `code = m.Run()`

bccsp/sw/impl_test.go
```diff
@@ -76,10 +80,10 @@ func TestMain(m *testing.M) {
...
 		if ret != 0 {
 			fmt.Printf("Failed testing at [%d, %s]", config.securityLevel, config.hashFamily)
-			os.Exit(-1)
+			return
 		}
 	}
-	os.Exit(0)
+	code = 0
```
guoger: this can be removed if `code` var is used

guoger (changes requested):

bccsp/sw/impl_test.go
baidang201: i will delete last line 'code = 0'

bccsp/sw/impl_test.go
baidang201: sure

guoger (approved):

yacovm (approved):
LGTM, thanks for the fix

PR #329, opened 2019-11-29
Validation tests do not clean up their temporary directories

baidang201:
Signed-off-by: baidang201 <jinrishouji@sina.com>

Type of change
Bug fix

Description
fix when unit test in core\handlers\validation\builtin，we find in the /tmp folder, has a lot of file like 'v1.3-validationxxx' and can not delete auto by test program.

Additional details
the reason is when os.exit, it do not run defer to delete 'v1.3-validationxxx' folder.
we can put os.exit in the head of function and import another defer for os.exit.

Related issues
https://jira.hyperledger.org/browse/FAB-17146

### DIFF
--- a/core/handlers/validation/builtin/v12/validation_logic_test.go
+++ b/core/handlers/validation/builtin/v12/validation_logic_test.go
@@ -2272,10 +2272,14 @@ var mockMSPIDGetter = func(cid string) []string {
 }
 
 func TestMain(m *testing.M) {
+	code := -1
+	defer func() {
+		os.Exit(code)
+	}()
 	testDir, err := ioutil.TempDir("", "v1.2-validation")
 	if err != nil {
 		fmt.Printf("Could not create temp dir: %s", err)
-		os.Exit(-1)
+		return
 	}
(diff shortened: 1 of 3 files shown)


core/handlers/validation/builtin/v12/validation_logic_test.go
```diff
@@ -2272,10 +2272,14 @@ var mockMSPIDGetter = func(cid string) []string {
 }
 
 func TestMain(m *testing.M) {
+	code := -1
+	defer func() {
+		os.Exit(code)
+	}()
```
guoger, line 2278: i would probably wrap this into `testMain(m *testing.M) int { defer ... }` which returns exit code, and calls `os.Exit(testMain(m))` from `TestMain`. But it's effectively the same

guoger (approved):
LGTM, tmp dirs are correctly cleaned up after test. waiting for CI to pass

mastersingh24 (approved):

mastersingh24 (approved):

PR #337, opened 2019-12-01
[FAB-17166] Gate UT, IT and DocBuild

lindluni:
Adds a gate behind VerifyBuild that will tell the other jobs which of them should run depending on which directories in the source tree have been updated

#### Type of change

#### Description

This change introduces logic to check if the current commit set has made changes to just the /docs tree and tells CI to only run DocBuild if this is the case. This makes sure DocBuild doesn't have to wait on UT and IT to run. Otherwise it runs all tests

### DIFF
--- a/ci/scripts/evaluate_commits.sh
+++ b/ci/scripts/evaluate_commits.sh
@@ -1,0 +1,12 @@
+#!/bin/bash
+# Copyright IBM Corp. All Rights Reserved.
+#
+# SPDX-License-Identifier: Apache-2.0
+set -euo pipefail
+
+if [[ $(git diff-tree --no-commit-id --name-only 'HEAD^..HEAD') == "docs" ]]; then
+  echo "##vso[task.setvariable variable=buildDoc;isOutput=true]true"
+else
+  echo "##vso[task.setvariable variable=buildDoc;isOutput=true]true"
+  echo "##vso[task.setvariable variable=runTests;isOutput=true]true"
+fi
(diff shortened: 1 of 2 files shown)


ci/scripts/evaluate_commits.sh
```diff
@@ -0,0 +1,12 @@
+#!/bin/bash
+# Copyright IBM Corp. All Rights Reserved.
+#
+# SPDX-License-Identifier: Apache-2.0
+set -euo pipefail
+
+if [[ $(git diff-tree --no-commit-id --name-only 'HEAD^..HEAD') == "docs" ]]; then
```
guoger, line 7: would `HEAD^..HEAD` work as expected if a PR contains multiple commits?

lindluni: I tested this with the following commit chains:
- first commit touches doc, second touches code, ran against CI, and it triggered all jobs as expected
- ran again with first touches doc, second touches code, third touches doc, all jobs ran as expected
- Third change I ran with a single commit to code, all jobs ran as expected.
- Fourth change I ran with a single commit with just a doc change, only the docbuild ran.

lindluni: Thanks for bringing that up, hadn't run it against the second case, but I've confirmed it works as expected now

guoger (approved):
LGTM. CI creates a merge commit prior to running the script, which makes `HEAD^..HEAD` work as expected (against changes in **all** commits)

kchristidis (approved):

PR #338, opened 2019-12-02
Testing CI

lindluni:


guoger:
~i think we need to test the scenario where the **last** commit of PR modifies docs only, right?~

just saw the comment in other PR

PR #344, opened 2019-12-02
[FAB-15724] Use in-proc client for migration tests 

stephyee:
#### Type of change
- Test update

#### Description

In migration int tests,
- add additional orderer after solo->etcdraft migration has completed
- use in-proc client instead of chaincode/peer

#### Related issues

https://jira.hyperledger.org/browse/FAB-15724

### DIFF
--- a/integration/raft/migration_test.go
+++ b/integration/raft/migration_test.go
@@ -20,6 +20,7 @@ import (
 	"github.com/golang/protobuf/proto"
 	"github.com/hyperledger/fabric-protos-go/common"
 	protosorderer "github.com/hyperledger/fabric-protos-go/orderer"
+	"github.com/hyperledger/fabric-protos-go/orderer/etcdraft"
 	protosraft "github.com/hyperledger/fabric-protos-go/orderer/etcdraft"
 	"github.com/hyperledger/fabric/integration/nwo"
 	"github.com/hyperledger/fabric/integration/nwo/commands"
@@ -34,13 +35,12 @@ import (
 
 var _ = Describe("Kafka2RaftMigration", func() {
 	var (
-		testDir   string
-		client    *docker.Client
-		network   *nwo.Network
-		chaincode nwo.Chaincode
+		testDir string
+		client  *docker.Client
+		network *nwo.Network
 
 		process                ifrit.Process
-		peerProc, brokerProc   ifrit.Process
+		brokerProc             ifrit.Process
 		o1Proc, o2Proc, o3Proc ifrit.Process
 
 		o1Runner, o2Runner, o3Runner *ginkgomon.Runner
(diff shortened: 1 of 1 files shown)


integration/raft/migration_test.go
```diff
@@ -201,11 +189,11 @@ var _ = Describe("Kafka2RaftMigration", func() {
 
 			By("3) Verify: create new channel, executing transaction")
 			network.CreateAndJoinChannel(orderer, channel2)
-			nwo.EnableCapabilities(network, channel2, "Application", "V2_0", orderer, network.Peer("Org1", "peer0"), network.Peer("Org2", "peer0"))
-			nwo.DeployChaincode(network, channel2, orderer, chaincode, network.PeersWithChannel(channel2)...)
-			RunExpectQueryRetry(network, peer, channel2, 100)
-			RunExpectQueryInvokeQuery(network, orderer, peer, channel2, 100)
-			RunExpectQueryInvokeQuery(network, orderer, peer, channel2, 90)
+
+			env = CreateBroadcastEnvelope(network, peer, channel2, []byte("hola"))
```
yacovm: Doing an [nwo.Broadcast ](https://github.com/hyperledger/fabric/blob/master/integration/nwo/orderer_client.go#L22-L51)doesn't mean much, as it doesn't check that the transaction was included into some block.

Perhaps:

1. Add usage of a [Deliver](https://github.com/hyperledger/fabric/blob/master/integration/nwo/orderer_client.go#L54-L88) call to ensure a new block with the expected seq num has been created
2. Extract this repetitive pattern (send transaction and ensure block is created) into a function

yacovm (dismissed):
Looks good, just a comment that can be addressed in a later PR or in this PR

integration/raft/migration_test.go
```diff
@@ -131,9 +118,6 @@ var _ = Describe("Kafka2RaftMigration", func() {
 			By("Create & join first channel, deploy and invoke chaincode")
 			network.CreateAndJoinChannel(orderer, channel1)
 			nwo.EnableCapabilities(network, channel1, "Application", "V2_0", orderer, network.Peer("Org1", "peer0"), network.Peer("Org2", "peer0"))
```
jyellick: Application capabilities don't matter if peers/chaincode are not involved.

integration/raft/migration_test.go
jyellick: +1 Performing the Deliver check would improve the quality of this test dramatically.

integration/raft/migration_test.go
```diff
@@ -757,9 +784,6 @@ var _ = Describe("Kafka2RaftMigration", func() {
...
```
jyellick: Again here

integration/raft/migration_test.go
```diff
@@ -131,9 +118,6 @@ var _ = Describe("Kafka2RaftMigration", func() {
 			By("Create & join first channel, deploy and invoke chaincode")
 			network.CreateAndJoinChannel(orderer, channel1)
```
jyellick: No need to Join channel, creating is sufficient if no peers.

integration/raft/migration_test.go
```diff
@@ -34,10 +35,9 @@ import (
...
-		testDir   string
-		client    *docker.Client
-		network   *nwo.Network
-		chaincode nwo.Chaincode
+		testDir string
+		client  *docker.Client
+		network *nwo.Network
 
 		process                ifrit.Process
 		peerProc, brokerProc   ifrit.Process
```
jyellick: One of the primary goals in using the broadcast client directly is to eliminate the need for peers from this test, so I would expect peerProc to go away.

yacovm (approved):

mastersingh24 (approved):

PR #345, opened 2019-12-02
[FAB-15824] Scripting friendly CLI query methods

raftAtGit:
This commit introduces 5 new peer CLI commands
to make CLI more suitable for scripting.
Please see relevant Jira ticket for details and motivation.

These commands exit with status 0 if condition is satisfied, 99 o/w.
This makes them easy to use in scripting without error prone text parsing.

peer channel checkexists -c <channelID>
peer channel checkjoined -c <channelID>
peer channel checkanchors -c <channelID> -g <orgID>

peer chaincode checkinstalled -n <name> -v <version>
peer chaincode checkinstantiated -C <channelID> -n <name> [-v <version>]

Also installed yq to tools Docker image for processing yaml files.
This is required basically to process configtx.yaml file.

Signed-off-by: Hakan Eryargi <hakan.eryargi@gmail.com>

#### Type of change

- Bug fix
- New feature
- Test update
- Documentation update

### DIFF
--- a/images/tools/Dockerfile
+++ b/images/tools/Dockerfile
@@ -30,3 +30,11 @@ ENV FABRIC_CFG_PATH /etc/hyperledger/fabric
 VOLUME /etc/hyperledger/fabric
 COPY --from=tools /go/src/github.com/hyperledger/fabric/build/bin /usr/local/bin
 COPY --from=tools /go/src/github.com/hyperledger/fabric/sampleconfig ${FABRIC_CFG_PATH}
+
+# install yq via pip
+RUN apk add --no-cache curl python
+RUN curl "https://bootstrap.pypa.io/get-pip.py" -o "get-pip.py" && \
+    python get-pip.py && \
+    rm get-pip.py && \
+    pip install yq
+
--- a/internal/peer/chaincode/chaincode.go
+++ b/internal/peer/chaincode/chaincode.go
@@ -8,6 +8,7 @@ package chaincode
 
 import (
 	"fmt"
+	"os"
 	"time"
 
 	"github.com/hyperledger/fabric/bccsp"
@@ -49,6 +50,8 @@ func Cmd(cf *ChaincodeCmdFactory, cryptoProvider bccsp.BCCSP) *cobra.Command {
 	chaincodeCmd.AddCommand(signpackageCmd(cf, cryptoProvider))
 	chaincodeCmd.AddCommand(upgradeCmd(cf, cryptoProvider))
 	chaincodeCmd.AddCommand(listCmd(cf, cryptoProvider))
+	chaincodeCmd.AddCommand(checkinstalledCmd(cf))
+	chaincodeCmd.AddCommand(checkinstantiatedCmd(cf))
 
 	return chaincodeCmd
 }
@@ -77,6 +80,10 @@ var (
 	connectionProfile     string
 	waitForEvent          bool
 	waitForEventTimeout   time.Duration
+
+	// used for testing explicit os.Exit(<exit code>) scenarios
+	osExit   = os.Exit
+	exitCode = -1
 )
 
 var chaincodeCmd = &cobra.Command{
--- a/internal/peer/chaincode/checkinstalled.go
+++ b/internal/peer/chaincode/checkinstalled.go
@@ -1,0 +1,116 @@
+/*
+Copyright IBM Corp. All Rights Reserved.
+
+SPDX-License-Identifier: Apache-2.0
+*/
+
+package chaincode
+
+import (
+	"context"
+	"fmt"
+
+	"github.com/golang/protobuf/proto"
+	cb "github.com/hyperledger/fabric-protos-go/common"
+	pb "github.com/hyperledger/fabric-protos-go/peer"
+	"github.com/hyperledger/fabric/internal/peer/common"
+	"github.com/hyperledger/fabric/protoutil"
+	"github.com/pkg/errors"
+	"github.com/spf13/cobra"
+)
+
+func checkinstalledCmd(cf *ChaincodeCmdFactory) *cobra.Command {
+	checkCmd := &cobra.Command{
+		Use:   "checkinstalled",
+		Short: "Check if a certain version of chaincode is installed to peer.",
+		Long:  "Check if a certain version of chaincode is installed to peer.",
+		RunE: func(cmd *cobra.Command, args []string) error {
+			if len(args) != 0 {
+				return fmt.Errorf("trailing args detected: %s", args)
+			}
+			return checkinstalled(cmd, args, cf)
+		},
+	}
+
+	flagList := []string{
+		"name",
+		"version",
+		"peerAddresses",
+		"connectionProfile",
+	}
+	attachFlags(checkCmd, flagList)
+
+	return checkCmd
+}
+
+func checkinstalled(cmd *cobra.Command, args []string, cf *ChaincodeCmdFactory) error {
+	if chaincodeName == common.UndefinedParamValue {
+		return fmt.Errorf("Must supply chaincode name")
+	}
+
+	if chaincodeVersion == common.UndefinedParamValue {
+		return fmt.Errorf("Must supply chaincode version")
+	}
+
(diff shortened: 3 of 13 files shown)


jyellick:
Hello @raftAtGit thanks for submitting this PR!

A few notes.  First, I think this should probably be broken up until multiple PRs, or at least multiple commits.  From the [contribution guide](https://hyperledger-fabric.readthedocs.io/en/release-1.4/CONTRIBUTING.html)

> Minimize the lines of code per PR. Why? Maintainers have day jobs, too. If you send a 1,000 or 2,000 LOC change, how long do you think it takes to review all of that code? Keep your changes to < 200-300 LOC, if possible. If you have a larger change, decompose it into multiple independent changes. If you are adding a bunch of new functions to fulfill the requirements of a new capability, add them separately with their tests, and then write the code that uses them to deliver the capability.

Secondly, for the chaincode related commands, these are only compatible with the legacy lifecycle (which will hopefully be removed soon).  I'm not strictly opposed to adding new commands for the legacy lifecycle, but if we do, then it's probably worth maintaining parity in the new lifecycle.

raftAtGit:
Hi @jyellick thanks for the comments!

I can split this into multiple commits or PR's, but as this is the implementation of a single [Jira ticket](https://jira.hyperledger.org/browse/FAB-15824) I found it natural to wrap up in a single commit.

For chaincode lifecycle, possibly we will soon switch to Fabric 2.0, that time I can implement these for new lifecycle methods too. We will also need them for [our repo](https://github.com/APGGroeiFabriek/PIVT)

Initially I targeted 1.4 branch for these changes, but as I was advised to target master branch I adapted the changes for master branch. 

After this PL is hopefully accepted and merged into master, I'm planning to back port them to 1.4 branch, as simply a cherry-pick is not possible because of folder structure and others changes.

That's also one reason I would prefer keeping this as a single commit since it will be much convenient to back-port and create a PR for 1.4 branch.

raftAtGit:
@jyellick @yacovm any intentions to merge this PR before 2.0 release? it has been a long time..

mastersingh24:
I have a few issues with this:

1) I don't feel we should be adding more things to a CLI we eventually want to move away from
2) I believe that you can implement the same functions by writing a shell script which wraps existing functions (at least for most of the new stuff you added)
3) While the status/return code style you've implemented makes sense, it's inconsistent with the other functions

raftAtGit:
Hi @mastersingh24 thanks for the comments!

1. I Strongly believe any smart management software running on top of Fabric will eventually need this kind of functionality. Including **Cello**
2. Well, the whole idea of this commit is to avoid that kind of shell scripts: they are ugly, error prone and fragile. This is such a strong motivation that, I never intended to wrote those scripts but instead learned basic Go skills and implemented these CLI methods.
3. The CLI methods I have implemented both return status code and also prints human readable information. Status codes are intended for machine to machine, while printouts are intended for human machine interaction.

BTW, I'm not sure if you are aware: we had built our own CLI [Docker image](https://hub.docker.com/r/raft/hl-fabric-tools-extended/tags) based on version 1.4.1  six months ago and it's used in our [Helm charts](https://github.com/APGGroeiFabriek/PIVT) by many people globally. I'm also aware of some users which are using old non-declarative functionality and waiting this PR to be merged.

rpocase:
I've leveraged variants of @raftAtGit 's Helm charts for a few months now and the `checkX` functions have been invaluable. it is definitely not ideal to extend a feature that is intended to be removed, but with a hazy timeline for removal and the effort level for migration to the new workflow makes me believe this is a worthwhile endeavor for many users. 

> While the status/return code style you've implemented makes sense, it's inconsistent with the other functions

@mastersingh24 are there specific changes necessary for consistency? Is there any documentation on CLI conformance for HLF (e.g., stdout/stderr expectations and return codes) to help @raftAtGit  know exactly what needs to change?

mastersingh24:
> I've leveraged variants of @raftAtGit 's Helm charts for a few months now and the `checkX` functions have been invaluable. it is definitely not ideal to extend a feature that is intended to be removed, but with a hazy timeline for removal and the effort level for migration to the new workflow makes me believe this is a worthwhile endeavor for many users.
> 
> > While the status/return code style you've implemented makes sense, it's inconsistent with the other functions
> 
> @mastersingh24 are there specific changes necessary for consistency? Is there any documentation on CLI conformance for HLF (e.g., stdout/stderr expectations and return codes) to help @raftAtGit know exactly what needs to change?

I actually like the model that's implemented in these proposed commands ... it's really the older commands that can be improved.

In the end, this looks like someone adding a few commands ... but we end up being the ones stuck with maintaining all of this plus the existing things.  And we cannot keep taking on debt.

I also do not believe that using the CLI is the proper way to do things either ... but that's a different story.

Where I / we really want to go is to provide break up the current functionality into the CLI into discrete Go packages which others can then use to build any tool they desire.  Of course we'd have  base CLI as well but it would be a composition of these packages as well.

sykesm:
When this was a CR pending in Gerrit, one of the questions I had was why the [fabric-cli] project isn't the right place to do this work? The [fabric-cli] project was proposed as a top level project focused on redesigning and improving the CLI experience for fabric users and that seems to be the aim of these changes as well.

If you could apply your skill and passion there, I expect it would have a larger impact with fewer constraints from the existing implementation. Have you considered that path? Have you reached out to those that initiated that effort to see how you could help?

@mastersingh24, myself, and others would very much like to get out of the business of maintaining a CLI for fabric. I don't believe what we have offers a good experience and adding to it, in my opinion, sends a mixed message.

[fabric-cli]: https://github.com/hyperledger/fabric-cli

raftAtGit:
@sykesm I agree **fabric-cli** project looks like a good match for this functionality. But as mentioned in that repo:
>  it is NOT the "official" Fabric CLI and there is not yet any commitment that it is going to be.

And honestly I don't have time to refactor these changes for another project with an unclear future. We will go production in a few months and you can imagine I'm quite busy. 

Speaking of going to production, to my knowledge, globally we will be the first Fabric in Kubernetes in production and we are making use of our Helm charts and this extended CLI methods based on our home built Docker images.

Anyway, after some experimenting with Fabric 2.0-beta release, I've noticed legacy chaincode life cycle is not possible with v2_0 capability and v1_4_3 capability is not stable.

Not sure how these will evolve in time. 

From that perspective, even myself don't see much value in this commit is merged into master branch. I believe best option is to merge it to 1.4 branch. Our users will be happy as they can use official Fabric tools Docker image. 

Regarding concerns about stuck with maintenance above, you must be also happy as you can completely drop these methods with 2.0 release. I dont see much to maintain here but I understand the concern.

But hey, initially at Gerrit this patch targeted 1.4. branch 6 months ago. You asked me to move it to master branch and we ended up in this situation. Everyone's life would be much easy if the patch merged into 1.4. branch that time. ;)

sykesm:
> @sykesm I agree **fabric-cli** project looks like a good match for this functionality. But as mentioned in that repo:
> 
> > it is NOT the "official" Fabric CLI and there is not yet any commitment that it is going to be.
> 
> And honestly I don't have time to refactor these changes for another project with an unclear future. We will go production in a few months and you can imagine I'm quite busy.

Fair enough.

> Speaking of going to production, to my knowledge, globally we will be the first Fabric in Kubernetes in production

Cool. IBM Blockchain Platform v2 has been hosting Fabric on Kubernetes since last June so there may be others.

> I believe best option is to merge it to 1.4 branch. Our users will be happy as they can use official Fabric tools Docker image.
> 
> Regarding concerns about stuck with maintenance above, you must be also happy as you can completely drop these methods with 2.0 release. I dont see much to maintain here but I understand the concern.

We actually can't drop the commands or mechanisms. Existing channels that have not enabled the 2.0 capabilities are still using the legacy lifecycle. Our hope is to remove support for the legacy lifecycle in a future release but it'll be a little while.
 
> But hey, initially at Gerrit this patch targeted 1.4. branch 6 months ago. You asked me to move it to master branch and we ended up in this situation.

Yes, because we do development in master and pull things back as necessary. This is a fairly conventional model and seems appropriate for any "new features" that are added to the project.

While we won't be merging these changes, we do appreciate your contribution and the discussion. Thank you.

PR #356, opened 2019-12-03
[FAB-17177] Config block shouldn't verify itself in block replication

yacovm:
The cluster replication verifies blocks by pulling them in batches,
and performs hash chain verification + verifies the signatures of the last block in the chain.

It uses the bundle stored in the global config structures,
but in case it comes across a config block in the middle of the batch,
it switches to use a bundle constructed from that config block.

However, if the config block is the last block in the batch -
it accidentally verifies it twice:
 - once with the config block before it (as expected)
 - once with the config block itself (which is unexpected).

This change set simply adds a check to see if the last block in the batch
is a config block, and if so - doesn't verify it if it was already verified.

Change-Id: Id123c36e445a21b3081273ef0395aae759162818

### DIFF
--- a/orderer/common/cluster/util.go
+++ b/orderer/common/cluster/util.go
@@ -197,12 +197,14 @@ func VerifyBlocks(blockBuff []*common.Block, signatureVerifier BlockVerifier) er
 	}
 
 	var config *common.ConfigEnvelope
+	var isLastBlockConfigBlock bool
 	// Verify all configuration blocks that are found inside the block batch,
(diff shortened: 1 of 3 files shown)


C0rWin (dismissed):

jyellick (approved):

C0rWin (approved):

yacovm:
@btl5037 is it a known thing that 1.4.x integration tests are flaky? It passes on my machine, and this change set has been merged into master but here it refuses to pass for 2 days...

PR #363, opened 2019-12-04
FAB-15817 Prefer last config in signatures metadata field

wlahti:
This commit updates the orderer and peer to prefer retrieving the
last config from the SIGNATURES field and fallback to retrieving
it from the LAST_CONFIG field.

Signed-off-by: Will Lahti <wtlahti@us.ibm.com>

### DIFF
--- a/common/genesis/genesis.go
+++ b/common/genesis/genesis.go
@@ -48,5 +48,10 @@ func (f *factory) Block(channelID string) *cb.Block {
 	block.Metadata.Metadata[cb.BlockMetadataIndex_LAST_CONFIG] = protoutil.MarshalOrPanic(&cb.Metadata{
 		Value: protoutil.MarshalOrPanic(&cb.LastConfig{Index: 0}),
 	})
+	block.Metadata.Metadata[cb.BlockMetadataIndex_SIGNATURES] = protoutil.MarshalOrPanic(&cb.Metadata{
+		Value: protoutil.MarshalOrPanic(&cb.OrdererBlockMetadata{
+			LastConfig: &cb.LastConfig{Index: 0},
+		}),
+	})
 	return block
 }
--- a/common/ledger/blockledger/util.go
+++ b/common/ledger/blockledger/util.go
@@ -44,6 +44,7 @@ func (nfei *NotFoundErrorIterator) Close() {}
 func CreateNextBlock(rl Reader, messages []*cb.Envelope) *cb.Block {
 	var nextBlockNumber uint64
 	var previousBlockHash []byte
+	var err error
 
 	if rl.Height() > 0 {
 		it, _ := rl.Iterator(&ab.SeekPosition{
@@ -63,7 +64,6 @@ func CreateNextBlock(rl Reader, messages []*cb.Envelope) *cb.Block {
 		Data: make([][]byte, len(messages)),
 	}
 
-	var err error
 	for i, msg := range messages {
 		data.Data[i], err = proto.Marshal(msg)
 		if err != nil {
--- a/orderer/common/cluster/replication.go
+++ b/orderer/common/cluster/replication.go
@@ -479,7 +479,7 @@ func PullLastConfigBlock(puller ChainPuller) (*common.Block, error) {
 	if lastBlock == nil {
 		return nil, ErrRetryCountExhausted
 	}
-	lastConfNumber, err := lastConfigFromBlock(lastBlock)
+	lastConfNumber, err := protoutil.GetLastConfigIndexFromBlock(lastBlock)
 	if err != nil {
 		return nil, err
 	}
@@ -508,13 +508,6 @@ func latestHeightAndEndpoint(puller ChainPuller) (string, uint64, error) {
 		}
 	}
 	return mostUpToDateEndpoint, maxHeight, nil
-}
-
-func lastConfigFromBlock(block *common.Block) (uint64, error) {
-	if block.Metadata == nil || len(block.Metadata.Metadata) <= int(common.BlockMetadataIndex_LAST_CONFIG) {
-		return 0, errors.New("no metadata in block")
-	}
-	return protoutil.GetLastConfigIndexFromBlock(block)
 }
 
 // Close closes the ChainInspector
@@ -663,11 +656,17 @@ func ExtractGenesisBlock(logger *flogging.FabricLogger, block *common.Block) (st
 	metadata := &common.BlockMetadata{
 		Metadata: make([][]byte, 4),
 	}
+	metadata.Metadata[common.BlockMetadataIndex_SIGNATURES] = protoutil.MarshalOrPanic(&common.OrdererBlockMetadata{
+		LastConfig: &common.LastConfig{Index: 0},
+		// This is a genesis block, peer never verify this signature because we can't bootstrap
+		// trust from an earlier block, hence there are no signatures here.
+	})
 	metadata.Metadata[common.BlockMetadataIndex_LAST_CONFIG] = protoutil.MarshalOrPanic(&common.Metadata{
 		Value: protoutil.MarshalOrPanic(&common.LastConfig{Index: 0}),
 		// This is a genesis block, peer never verify this signature because we can't bootstrap
 		// trust from an earlier block, hence there are no signatures here.
 	})
+
 	blockdata := &common.BlockData{Data: [][]byte{payload.Data}}
 	b := &common.Block{
 		Header:   &common.BlockHeader{DataHash: protoutil.BlockDataHash(blockdata)},
--- a/orderer/common/cluster/util.go
+++ b/orderer/common/cluster/util.go
@@ -652,9 +652,6 @@ func LastConfigBlock(block *common.Block, blockRetriever BlockRetriever) (*commo
 	if blockRetriever == nil {
 		return nil, errors.New("nil blockRetriever")
 	}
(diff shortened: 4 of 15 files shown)


jyellick (changes requested):
We need to make sure that:

a) Old clients can still use new blockchains (ie, continue to populate the old metadata position)
b) New clients can still parse old blockchains (ie, fall-back to the old metadata position if the new one is empty)
c) New clients prefer the new metadata location (the old location is not signed)

common/genesis/genesis.go
```diff
@@ -45,8 +45,10 @@ func (f *factory) Block(channelID string) *cb.Block {
 	block := protoutil.NewBlock(0, nil)
 	block.Data = &cb.BlockData{Data: [][]byte{protoutil.MarshalOrPanic(envelope)}}
 	block.Header.DataHash = protoutil.BlockDataHash(block.Data)
-	block.Metadata.Metadata[cb.BlockMetadataIndex_LAST_CONFIG] = protoutil.MarshalOrPanic(&cb.Metadata{
```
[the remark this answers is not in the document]
jeffgarratt: I would say for now and forever.  This is exactly the type of change that will break every existing client.  Great care must be taken in proto modification such as this.  The ramifications are extreme.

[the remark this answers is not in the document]
jyellick: At some point, we are allowed to break client compatibility -- in v1.2 we removed the old eventhub service in favor of the Deliver API, for instance.  So, I don't think it's out of the realm of possibility that eventually, we stop populating this field.  However, it's a small amount of data, and the benefit to breaking these clients is similarly small.  We really want clients to move to using the other metadata field purely because it's signed and peers validate this signature.

common/ledger/blockledger/util.go
```diff
@@ -74,6 +78,11 @@ func CreateNextBlock(rl Reader, messages []*cb.Envelope) *cb.Block {
 	block := protoutil.NewBlock(nextBlockNumber, previousBlockHash)
 	block.Header.DataHash = protoutil.BlockDataHash(data)
 	block.Data = data
+	block.Metadata.Metadata[cb.BlockMetadataIndex_SIGNATURES] = protoutil.MarshalOrPanic(&cb.Metadata{
```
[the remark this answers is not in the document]
wlahti: It's due to the LastConfig now being nested inside the OrdererBlockMetadata instead of as its own separate value. Without setting this here, certain tests panic with a nil pointer when attempting to retrieve the index of a nil LastConfig (in regstrar_test.go's testLastConfigBlockNumber function). It's possibly the case that these tests weren't really testing that the last config was set correctly and merely relied on the default value.

common/genesis/genesis.go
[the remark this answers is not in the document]
wlahti: Done

orderer/common/cluster/replication.go
```diff
@@ -663,11 +663,12 @@ func ExtractGenesisBlock(logger *flogging.FabricLogger, block *common.Block) (st
 	metadata := &common.BlockMetadata{
 		Metadata: make([][]byte, 4),
 	}
-	metadata.Metadata[common.BlockMetadataIndex_LAST_CONFIG] = protoutil.MarshalOrPanic(&common.Metadata{
```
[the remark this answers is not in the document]
wlahti: Made sure to continue to set the old value here. Done.

common/ledger/blockledger/util.go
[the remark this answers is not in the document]
jyellick: Then I think we need to fix those tests.  The `CreateNextBlock` function has no business setting this metadata.  In fact, it can't set it correctly, because it doesn't know whether this block is a config block (in which case, the last config should be this block number.)  As implemented above, all blocks would point to block 0 as the last config, as there is no case where the value is updated.

orderer/common/cluster/replication.go
```diff
@@ -663,11 +663,17 @@ func ExtractGenesisBlock(logger *flogging.FabricLogger, block *common.Block) (st
 	metadata := &common.BlockMetadata{
 		Metadata: make([][]byte, 4),
 	}
+	metadata.Metadata[common.BlockMetadataIndex_SIGNATURES] = protoutil.MarshalOrPanic(&common.OrdererBlockMetadata{
+		LastConfig: &common.LastConfig{Index: 0},
+		// This is a genesis block, peer never verify this signature because we can't bootstrap
+		// trust from an earlier block, hence there are no signatures here.
+	})
 	metadata.Metadata[common.BlockMetadataIndex_LAST_CONFIG] = protoutil.MarshalOrPanic(&common.Metadata{
 		Value: protoutil.MarshalOrPanic(&common.LastConfig{Index: 0}),
```
jyellick, line 665: Because of the removed check on line 514, this could now result in an out of bounds dereference.  We either need to restore the original check, or, add an additional check before this dereference.

orderer/common/cluster/util.go
```diff
@@ -652,7 +652,7 @@ func LastConfigBlock(block *common.Block, blockRetriever BlockRetriever) (*commo
 	if blockRetriever == nil {
 		return nil, errors.New("nil blockRetriever")
 	}
-	if block.Metadata == nil || len(block.Metadata.Metadata) <= int(common.BlockMetadataIndex_LAST_CONFIG) {
+	if block.Metadata == nil || len(block.Metadata.Metadata) <= int(common.BlockMetadataIndex_SIGNATURES) {
```
jyellick: Either this check is not helping us, or we could have a similar out of bounds reference here.

orderer/common/multichannel/registrar_test.go
```diff
@@ -358,13 +362,13 @@ func TestCreateChain(t *testing.T) {
 }
 
 func testLastConfigBlockNumber(t *testing.T, block *cb.Block, expectedBlockNumber uint64) {
-	metadataItem := &cb.Metadata{}
```
jyellick, line 361: We should be checking that both of these are set, right?

orderer/common/server/main_test.go
```diff
@@ -295,8 +295,10 @@ func TestExtractSysChanLastConfig(t *testing.T) {
 	require.NoError(t, err)
 
 	nextBlock := blockledger.CreateNextBlock(rl, []*common.Envelope{configTx})
-	nextBlock.Metadata.Metadata[common.BlockMetadataIndex_LAST_CONFIG] = protoutil.MarshalOrPanic(&common.Metadata{
```
jyellick, line 298: Same here?

protoutil/blockutils.go
```diff
@@ -156,16 +156,30 @@ func GetConsenterMetadataFromBlock(block *cb.Block) (*cb.Metadata, error) {
...
+	m, err := GetMetadataFromBlock(block, cb.BlockMetadataIndex_SIGNATURES)
 	if err != nil {
-		return 0, err
+		return 0, errors.Wrap(err, "failed to get signatures")
 	}
-	lc := &cb.LastConfig{}
-	err = proto.Unmarshal(md.Value, lc)
+	// TODO FAB-15864 Remove this fallback when we can stop supporting upgrade from pre-1.4.1 orderer
+	if len(m.Value) == 0 {
+		m, err := GetMetadataFromBlock(block, cb.BlockMetadataIndex_LAST_CONFIG)
```
jyellick, line 173: Potential for out of bounds dereference.  I'd think it would be more appropriate to move the len checks in here rather than always sprinkle them in front of this call.

jyellick (changes requested):
I'd make sure to do a quick audit that in the tests you have not removed checks that the old field value be set, and that the fallback scenarios are appropriately tested.

protoutil/blockutils.go
```diff
@@ -130,7 +138,7 @@ func GetMetadataFromBlockOrPanic(block *cb.Block, index cb.BlockMetadataIndex) *
 func GetConsenterMetadataFromBlock(block *cb.Block) (*cb.Metadata, error) {
 	m, err := GetMetadataFromBlock(block, cb.BlockMetadataIndex_SIGNATURES)
 	if err != nil {
-		return nil, errors.Wrapf(err, "failed to retrieve metadata at index: %d", cb.BlockMetadataIndex_SIGNATURES)
+		return nil, errors.Wrap(err, "failed to retrieve metadata")
```
jyellick: If you're changing this, let's do a `WithMessage` instead of a `Wrap`, no need to have two stacks attached.

protoutil/blockutils.go
```diff
@@ -147,7 +155,7 @@ func GetConsenterMetadataFromBlock(block *cb.Block) (*cb.Metadata, error) {
 	res := &cb.Metadata{}
 	err = proto.Unmarshal(obm.ConsenterMetadata, res)
 	if err != nil {
-		return nil, errors.Wrapf(err, "failed to unmarshal consenter metadata")
+		return nil, errors.Wrap(err, "failed to unmarshal consenter metadata")
```
jyellick, line 158: Same here

jyellick (commented):
Looks pretty good to me, just a couple very minor things.

sykesm (approved):

protoutil/blockutils.go
wlahti: I think this one is fine as-is since it's wrapping an error from the proto package instead of a fabric package that already generated the stack. I did an audit (just after this review comment came in) of the other errors and updated at least one other place where it was wrapping instead of withmessaging.

jyellick (approved):

PR #364, opened 2019-12-04
[FAB-16951] Alternative mechanisms to find pkcs11 key

ldesrosi:
#### Type of change
- Improvement - Enable Hyperledger Fabric PKCS11 module to work with AWS CloudHSM.

#### Description

This modification adds a parameter called AltID to the PKCS11 BCCSP module.
When this parameter is set, the generateECKey and findKeyPairFromSKI functions
will use this alternative ID as the CKA_ID and CKA_LABEL to store and use the
associated private key.
This code change is required in situations where the HSM does not allow
modification of the CKA_ID after creation.
This code change includes the corresponding unit-test and has been tested against
both AWS CloudHSM and the SoftHSM.

#### Related issues
https://jira.hyperledger.org/browse/FAB-16951

Signed-off-by: Luc Desrosiers <ldesrosi@uk.ibm.com>

### DIFF
--- a/bccsp/pkcs11/pkcs11.go
+++ b/bccsp/pkcs11/pkcs11.go
@@ -124,13 +124,13 @@ func (csp *impl) getECKey(ski []byte) (pubKey *ecdsa.PublicKey, isPriv bool, err
 	session := csp.getSession()
 	defer csp.returnSession(session)
 	isPriv = true
-	_, err = findKeyPairFromSKI(p11lib, session, ski, privateKeyFlag)
+	_, err = findKeyPairFromSKI(p11lib, session, ski, csp.altId, privateKeyFlag)
 	if err != nil {
 		isPriv = false
 		logger.Debugf("Private key not found [%s] for SKI [%s], looking for Public key", err, hex.EncodeToString(ski))
 	}
 
-	publicKey, err := findKeyPairFromSKI(p11lib, session, ski, publicKeyFlag)
+	publicKey, err := findKeyPairFromSKI(p11lib, session, ski, csp.altId, publicKeyFlag)
 	if err != nil {
 		return nil, false, fmt.Errorf("Public key not found [%s] for SKI [%s]", err, hex.EncodeToString(ski))
 	}
@@ -215,9 +215,22 @@ func (csp *impl) generateECKey(curve asn1.ObjectIdentifier, ephemeral bool) (ski
 	session := csp.getSession()
 	defer csp.returnSession(session)
 
-	id := nextIDCtr()
-	publabel := fmt.Sprintf("BCPUB%s", id.Text(16))
-	prvlabel := fmt.Sprintf("BCPRV%s", id.Text(16))
+	keylabel := ""
+	updateSKI := false
+	if csp.altId == "" {
+		// Generate using the SKI process and then make keypair immutable according to csp.immutable
+		keylabel = fmt.Sprintf("BCP%s", nextIDCtr().Text(16))
+		updateSKI = true
+	} else if csp.altId != "" && csp.immutable {
+		// Generate the key pair using AltID process.
+		// No need to worry about immutable since AltID is used with Write-Once HSMs
+		keylabel = csp.altId
+		updateSKI = false
+	} else if csp.altId != "" && !csp.immutable {
+		// Raise an error since AltID is used with Write-Once HSMs
+		// So cannot make support AltID with immutable = false.
+		return nil, nil, fmt.Errorf("Cannot generate a mutable key using AltID.")
+	}
 
 	marshaledOID, err := asn1.Marshal(curve)
 	if err != nil {
@@ -230,10 +243,9 @@ func (csp *impl) generateECKey(curve asn1.ObjectIdentifier, ephemeral bool) (ski
 		pkcs11.NewAttribute(pkcs11.CKA_TOKEN, !ephemeral),
 		pkcs11.NewAttribute(pkcs11.CKA_VERIFY, true),
 		pkcs11.NewAttribute(pkcs11.CKA_EC_PARAMS, marshaledOID),
-		pkcs11.NewAttribute(pkcs11.CKA_PRIVATE, false),
 
-		pkcs11.NewAttribute(pkcs11.CKA_ID, publabel),
-		pkcs11.NewAttribute(pkcs11.CKA_LABEL, publabel),
+		pkcs11.NewAttribute(pkcs11.CKA_ID, []byte(keylabel)),
+		pkcs11.NewAttribute(pkcs11.CKA_LABEL, keylabel),
 	}
 
 	prvkeyT := []*pkcs11.Attribute{
@@ -243,8 +255,8 @@ func (csp *impl) generateECKey(curve asn1.ObjectIdentifier, ephemeral bool) (ski
 		pkcs11.NewAttribute(pkcs11.CKA_PRIVATE, true),
 		pkcs11.NewAttribute(pkcs11.CKA_SIGN, true),
 
-		pkcs11.NewAttribute(pkcs11.CKA_ID, prvlabel),
-		pkcs11.NewAttribute(pkcs11.CKA_LABEL, prvlabel),
+		pkcs11.NewAttribute(pkcs11.CKA_ID, []byte(keylabel)),
+		pkcs11.NewAttribute(pkcs11.CKA_LABEL, keylabel),
 
 		pkcs11.NewAttribute(pkcs11.CKA_EXTRACTABLE, false),
 		pkcs11.NewAttribute(pkcs11.CKA_SENSITIVE, true),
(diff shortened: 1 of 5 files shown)


ldesrosi:
Adding back comments from Gerritt:

> bccsp/pkcs11/conf.go
> Line 91:
> 
> This attribute has to be persisted IMO. How do you achieve this ?
> bccsp/pkcs11/impl.go
> Line 57:
> 
> same comment as before. This attribute has to be persisted.
> bccsp/pkcs11/pkcs11.go
> Line 275:
> 
> This block does not go hand-in-hand with your logic. It is not always true that absence of Alt-d will make modifiable to true.

Answers:
1. The alternative ID is associated via the config.  When the code runs in the context of an identity enrollment, the altID will be used to set the CKA_ID and CKA_Label
2. The modifiable variable is used to ensure that it maintains the same system behaviour as before the code was added.  Currently, the PKCS11 module creates the identity as modifiable on only after setting the CKA_ID and CKA_Label to the SKI does it change it to make it not-modifiable.

ashku:
> Adding back comments from Gerritt:
> 
> > bccsp/pkcs11/conf.go
> > Line 91:
> > This attribute has to be persisted IMO. How do you achieve this ?
> > bccsp/pkcs11/impl.go
> > Line 57:
> > same comment as before. This attribute has to be persisted.
> > bccsp/pkcs11/pkcs11.go
> > Line 275:
> > This block does not go hand-in-hand with your logic. It is not always true that absence of Alt-d will make modifiable to true.
> 
> Answers:
> 
>     1. The alternative ID is associated via the config.  When the code runs in the context of an identity enrollment, the altID will be used to set the CKA_ID and CKA_Label
> 
>     2. The modifiable variable is used to ensure that it maintains the same system behaviour as before the code was added.  Currently, the PKCS11 module creates the identity as modifiable on only after setting the CKA_ID and CKA_Label to the SKI does it change it to make it not-modifiable.

Answer 1 : That config is used in generation time , not invocation time. Once the attributes are populated , they have to be persisted. 
Answer 2 : Your code block that you changed is for copying , not modifying. You need to make that logic separate.

ashku (changes requested):
Answer 1 : That config is used in generation time , not invocation time. Once the attributes are populated , they have to be persisted.
Answer 2 : Your code block that you changed is for copying , not modifying. You need to make that logic separate.

ldesrosi:
@ashku I did an analysis a few weeks back to see what were the potential options.  I realized that it might be helpful to understand why I chose this approach.  I am attaching a small PDF that summarizes it.
[AWS CloudHSM impact analysis v2.1.pdf](https://github.com/hyperledger/fabric/files/3926383/AWS.CloudHSM.impact.analysis.v2.1.pdf)

ldesrosi:
To answer your previous point, the change will apply to both generation time and invocation time.   
  
If you look at section 3 (Lookup private key using a different correlation id) of the document attached you will find two options that were considered:
1. Add a new BCCSP PKCS11 option to specify the correlation ID (included in this PR)
2. Store the correlation ID in a reference table

For that second option, I looked at either adding value to the certificate or adding a reference table.  In both cases, it requires a significant amount of changes and it exposes the alternate ID to the higher layers of the architecture.

ashku:
Setting CKA_MODIFIABLE to false is user configurable option , it is not default option.
 Based on the doc attached , Copy Object is not supported on aws HSM. Can you confirm the above statement ?

ldesrosi:
You are correct, AWS CloudHSM does not support the CopyObject.  The list of supported operations can be found here: https://docs.aws.amazon.com/cloudhsm/latest/userguide/pkcs11-apis.html

ldesrosi:
> Setting CKA_MODIFIABLE to false is user configurable option  
  
Let me have another look at the code.  I understand your point.  

The CKA_MODIFIABLE only makes sense when the HSM supports the Set_Attribute or the CopyObject or else it is implicitely not modifiable.  However maybe I can use a different flag or rework the logic to make this clearer.

ldesrosi:
@ashku Looking at the code, the way I would see we could represent this is as follows:

1. Add an attribute that indicates that the HSM does not support modification of the key pair.  Something like: `hsmStyle: WriteOnce|WriteMany`
2. If the `hsmStyle` is set to `WriteOnce` then the `AltID` field becomes mandatory and the `immutable` field can be ignored since the key pair cannot be modified anyway.
3. If the `hsmStyle` is set to `WriteMany` then the `AltID` and `immutable` fields can be set if the admin decides to.

If that works, then it would change the block of logic accordingly.

jgkim:
@ldesrosi Is there any update on this issue?

ashku:
The implementation should not be so complex. I do not like the idea of maintaining another id. SKI is merely a reference to find Public and Private Key. The whole approach needs a revisit.

ashku:
@ldesrosi , based on our discussion , I realized that ski calculation at HSM cannot be decoupled from elliptic curve operation. That is smaller problem. The bigger problem is AWS HSM does not support CKA_EC_POINT operation. So, your none of the options of current design hold good. We need to revisit the design. Let me know if you need any help.

ldesrosi:
@ashku following our discussion, I have updated the pkcs11.go generateECKey method to validate the config parameters AltID and Immutable upfront.  It should improve the readability of the code.
  
Adding a few points from our previous discussions:  
  
1. SKI is a computed value from the EC Point from the public key.  
1. AWS HSM supports CKA_EC_POINT operation when using the public key, which is what the BCCSP PKCS11 module is doing.
1. Since the SKI is computed, the key pair has to be updated after creation to set the CKA_ID and CKA_LABEL to that value so it can be looked up after.
1. There is no other way around the generation of an alternative ID due to the fact that the AWS CloudHSM does not support ANY modification after the key pair generation.

ashku:
I was looking at AWS HSM doc https://docs.aws.amazon.com/cloudhsm/latest/userguide/pkcs11-attributes.html and found and found out that default value of CKA_MODIFIABLE and CKA_DESTROYABLE is true. Given that fact, what was reason for not taking that approach , meaning make a copy and destroy old copy.

ashku:
I see , CKA_ID cannot be modified. LGTM.

ahmedsajid:
Is this slated for 1.4.7?

mastersingh24:
I think this looks ok .... I think we should look at possibly making a different change in 2.0 / master, but for 1.4.x this seems ok.

We will likely need a doc update, although I don't think the PKCS11 docs were backported to 1.4.x in any case

denyeart:
@adecaro is going to find a crypto expert to review as well...

bccsp/pkcs11/pkcs11.go
```diff
@@ -230,10 +243,9 @@ func (csp *impl) generateECKey(curve asn1.ObjectIdentifier, ephemeral bool) (ski
 		pkcs11.NewAttribute(pkcs11.CKA_TOKEN, !ephemeral),
 		pkcs11.NewAttribute(pkcs11.CKA_VERIFY, true),
 		pkcs11.NewAttribute(pkcs11.CKA_EC_PARAMS, marshaledOID),
-		pkcs11.NewAttribute(pkcs11.CKA_PRIVATE, false),
 
-		pkcs11.NewAttribute(pkcs11.CKA_ID, publabel),
-		pkcs11.NewAttribute(pkcs11.CKA_LABEL, publabel),
+		pkcs11.NewAttribute(pkcs11.CKA_ID, []byte(keylabel)),
```
denyeart, line 247: Why the change from string to byte array as part of this PR?

bccsp/pkcs11/pkcs11.go
```diff
@@ -215,9 +215,22 @@ func (csp *impl) generateECKey(curve asn1.ObjectIdentifier, ephemeral bool) (ski
...
 	defer csp.returnSession(session)
 
-	id := nextIDCtr()
-	publabel := fmt.Sprintf("BCPUB%s", id.Text(16))
-	prvlabel := fmt.Sprintf("BCPRV%s", id.Text(16))
+	keylabel := ""
+	updateSKI := false
+	if csp.altId == "" {
+		// Generate using the SKI process and then make keypair immutable according to csp.immutable
+		keylabel = fmt.Sprintf("BCP%s", nextIDCtr().Text(16))
```
denyeart, line 222: Why the change from BCPUB/BCPRV to BCP as part of this PR?
What does BCP stand for?

denyeart (commented):
Just asking some picky questions since this is LTS release branch...

bccsp/pkcs11/pkcs11.go
ldesrosi: CKA_ID is of type byte array. 

When running the PKCS11 module against the AWS CloudHSM, if we pass a string, it fails with attribute type error.

bccsp/pkcs11/pkcs11.go
ldesrosi: Both BCPUB and BCPRV are temporary placeholders for the CKA_ID and CKA_LABEL.  Current code sets the value to the private key and public key to these values.

Keeping these two fields for that purpose made the modification more complex.  Given that the public key and private key should always have the same ID and Label, I felt it was a small change to make it more readable.  BCP stands for Blockchain Crypto Provider.

In any case, it does not impact the functionality, I can revert that change if there is a preference for that.

denyeart:
Ok, thanks Luc, I'll go ahead and merge.

denyeart (approved):

ahmedsajid:
Is there going to be a ticket for updating the doc and providing some instructions on how to use this @denyeart @ldesrosi?
Maybe updating sampleconfigs BCCSP section as well?

sykesm:
@denyeart @mastersingh24 - As this has broken the convention of "master first" and was merged into 1.4, who is driving the work to get this into master/release-2.1?

brjd:
@sykesm @denyeart @mastersingh24 I think this still wasn't merged into master or any releases other than 1.4, and so AWS CloudHSM doesn't work with Fabric v2.x.x. When will this get into master and a v2 release, please?

sykesm:
@braduf Pull requests are welcome.

denyeart:
Originally this was thought to be a tactical solution for v1.4.x. Given that no other strategic solution has been proposed for future versions, and that people seem satisfied with the v1.4.x solution, I agree the same thing should be PRed to master and then backported to release-2.3 and release-2.2 by somebody with knowledge of AWS CloudHSM.

ashku:
I agree Dev.

Thanks and Regards
Ashutosh Kumar
________________________________

PR #365, opened 2019-12-04
[FAB-17184] Skip pulling from remote peers if all remaining missing rwsets  are invalid when peer is configured to skip pulling invalid transactions

caod123:
#### Type of change

Improvement

#### Description

Prior to this change pvtdataprovider would enter the pull retry loop
and still attempt to fetch from peers even if all missing rwsets were
invalid

#### Related issues

[FAB-17184](https://jira.hyperledger.org/browse/FAB-17184)

### DIFF
--- a/gossip/privdata/pvtdataprovider.go
+++ b/gossip/privdata/pvtdataprovider.go
@@ -137,7 +137,6 @@ func (ec *eligibilityComputer) computeEligibility(pvtdataToRetrieve []*ledger.Tx
 			eligibleMissingKeys[key] = rwsetInfo{
 				invalid: invalid,
 			}
-
 			sources[key] = endorsersFromEligibleOrgs(ns, col, endorsers, policy.MemberOrgs())
 		}
 	}
@@ -225,13 +224,9 @@ func (pdp *PvtdataProvider) RetrievePvtdata(pvtdataToRetrieve []*ledger.TxPvtdat
 	pdp.logger.Debugf("Fetching %d collection private write sets from remote peers for a maximum duration of %s", len(pvtdataRetrievalInfo.eligibleMissingKeys), retryThresh)
 	startPull := time.Now()
 	for len(pvtdataRetrievalInfo.eligibleMissingKeys) > 0 && time.Since(startPull) < retryThresh {
-		pdp.populateFromRemotePeers(pvtdata, pvtdataRetrievalInfo)
-
-		// If succeeded to fetch everything, break to skip sleep
-		if len(pvtdataRetrievalInfo.eligibleMissingKeys) == 0 {
+		if needToRetry := pdp.populateFromRemotePeers(pvtdata, pvtdataRetrievalInfo); !needToRetry {
 			break
 		}
-
 		// If there are still missing keys, sleep before retry
 		pdp.sleeper.Sleep(pullRetrySleepInterval)
 	}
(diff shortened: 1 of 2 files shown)


gossip/privdata/pvtdataprovider.go
```diff
@@ -137,7 +138,9 @@ func (ec *eligibilityComputer) computeEligibility(pvtdataToRetrieve []*ledger.Tx
 			eligibleMissingKeys[key] = rwsetInfo{
 				invalid: invalid,
 			}
-
+			if invalid {
+				numInvalid++
+			}
```
denyeart: Before this for loop, we've already broken out if we're not supposed to store private data of invalid transactions, therefore this would only get incremented if we're supposed to store private data of invalid transactions. Is that what you intended?

caod123: storePvtdataOfInvalidTx is a separate flag that's tied to the 2.0 capability. skipPulllingInvalidTx is configured on a peer level. I don't believe there would be a situation where you would have skipPullingInvalidTx set to true and storePvtdataOfInvalidTx turned off (via not turning on 2.0 capability) because in that case the later check for skipping pulling invalid transactions would be unnecessary since we don't even consider invalid transactions

lindluni:
Can you rebase this on master to pick up the retryThreshold increase and see if that helps

caod123:
> Can you rebase this on master to pick up the retryThreshold increase and see if that helps

Sure

gossip/privdata/pvtdataprovider_test.go
```diff
@@ -984,6 +984,104 @@ func TestRetryFetchFromPeer(t *testing.T) {
...
+
+	retrievedPvtdata, err := pdp.RetrievePvtdata(pvtdataToRetrieve)
+	assert.NoError(t, err)
+
+	blockPvtdata := sortBlockPvtdata(retrievedPvtdata.GetBlockPvtdata())
+	assert.Equal(t, expectedBlockPvtdata, blockPvtdata)
+
+	// Check sleep and fetch were never called
+	assert.Equal(t, fakeSleeper.SleepCallCount(), 0)
+	assert.Equal(t, newFetcher.FetchCallCount(), 0)
```
guoger:
```suggestion
	assert.Len(t, newFetcher.Calls, 0)
```
so you don't need `callcount` var in struct

guoger (changes requested):
I wonder if we could return a `needToRetry bool` from `populateFromRemotePeers` to indicate whether it **fails** to pull any **valid** transaction? so we don't necessarily need to keep track of a counter at several places

gossip/privdata/pvtdataprovider_test.go
caod123: I thought about that originally but decided against it since it seemed unintuitive that `populateFromRemotePeers` should return anything. I don't have any strong opinions though.

caod123:
Force-push diff is a little messy because I also rebased off master.

gossip/privdata/pvtdataprovider.go
```diff
@@ -334,6 +346,10 @@ func (pdp *PvtdataProvider) populateFromTransientStore(pvtdata rwsetByKeys, pvtd
 					}
 					// populate the pvtdata with the RW set from the transient store
 					pvtdata[key] = col.Rwset
+					// decrement number of invalid eligible missing keys if writeset is invalid
+					if pvtdataRetrievalInfo.eligibleMissingKeys[key].invalid {
+						pvtdataRetrievalInfo.numInvalid--
+					}
```
guoger: i paused here to think, why are we decrementing `numInvalid` when we find it to be invalid... 

if we do go down this approach, can we rename `numInvalid` to be `numSkip`, and comments could be more explicit
> decrement number of skipped key since invalid missing key is populated from cache/transient store

guoger (dismissed):
> I thought about that originally but decided against it since it seemed unintuitive that populateFromRemotePeers should return anything. I don't have any strong opinions though.

IMHO, it's naturally to expect some hints to be returned when it comes to "do something with remote peers". More importantly, i think the code would be more manageable if we don't carry struct member `numInvalid` all along, and alter it in both `populateFromCache` and `populateFromTransient`, **especially when we have all the info needed to compute such result in `populateFromRemotePeers`**.

here's what i have in mind https://github.com/guoger/fabric/commit/aa2cbf9725bcda0de165f54fd587e62ff3cb18d5

let me know what you think

gossip/privdata/pvtdataprovider.go
caod123: > > I thought about that originally but decided against it since it seemed unintuitive that populateFromRemotePeers should return anything. I don't have any strong opinions though.
> 
> IMHO, it's naturally to expect some hints to be returned when it comes to "do something with remote peers". More importantly, i think the code would be more manageable if we don't carry struct member `numInvalid` all along, and alter it in both `populateFromCache` and `populateFromTransient`, **especially when we have all the info needed to compute such result in `populateFromRemotePeers`**.
> 
> here's what i have in mind [guoger@aa2cbf9](https://github.com/guoger/fabric/commit/aa2cbf9725bcda0de165f54fd587e62ff3cb18d5)
> 
> let me know what you think

I like it, this was the original approach I had intended so I'm fine with it if we're in agreement about returning the bool from `populateFromRemotePeers`

caod123: Let me know if you want me to force push your commit on my branch or if you'd like to just make a new PR I'm fine with that too and we can close this one.

guoger: @caod123 sorry for late response, please continue your PR here.

caod123:
@guoger since the previous requested change is no longer valid, can you dismiss the requested change to unblock the PR?

guoger (approved):
LGTM, but would like experts in this area to take a look as well.

C0rWin (approved):
Looks good

PR #367, opened 2019-12-04
Bump IT pull retry threshold to 7s

caod123:
#### Type of change

Improvement

#### Description

3s might have been a little too short for tests that actually needed
to fetch private data. Incrementing by just a little to reduce number of
flakes

### DIFF
--- a/integration/nwo/core_template.go
+++ b/integration/nwo/core_template.go
@@ -59,7 +59,7 @@ peer:
       leaderAliveThreshold: 10s
       leaderElectionDuration: 5s
     pvtData:
-      pullRetryThreshold: 3s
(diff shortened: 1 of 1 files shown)


jyellick (approved):
I'd love to see some more investigation into why a retry interval is needed at all. Assuming gossip is working correctly, I would expect for the first attempt to retrieve the private data to be successful, and no retry (and therefore retry timeout would have no effect).

mastersingh24 (approved):
I agree with @jyellick 
This seems like yet another band-aid

caod123:
Adding this followup investigation to https://jira.hyperledger.org/browse/FAB-16862

PR #372, opened 2019-12-05
[FAB-15035] Remove couchdb HTTP Request headers from debug log

caod123:
#### Type of change

Improvement

#### Description

This was printing out basic auth credentials and is probably unnecessary to log
in the first place

/cc @denyeart


denyeart (approved):

mastersingh24 (approved):
NICE

PR #376, opened 2019-12-06
[FAB-17220] Dynamically build TLS config in Raft client handshake

yacovm:
When we expand the root TLS CA in the channel config, **after**
Raft membership has expanded with an OSN that is issed a certificate
by a new TLS CA, the TLS client handshake uses the old root CA pool
and as a result the added orderer cannot be reached by the existing ones,
because their dialers reject its certificate.

This change set builds a dynamic transport credentials that
re-computes the TLS config in every TLS client handshake.

Expanded an integration test to ensure this works.

Change-Id: I6578ba49f16e14b97eb4eef4feccdecbfe1b7015

### DIFF
--- a/core/comm/client.go
+++ b/core/comm/client.go
@@ -14,7 +14,6 @@ import (
 
 	"github.com/pkg/errors"
 	"google.golang.org/grpc"
-	"google.golang.org/grpc/credentials"
 	"google.golang.org/grpc/keepalive"
 )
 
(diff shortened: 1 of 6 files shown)


core/comm/client.go
```diff
@@ -213,3 +213,37 @@ func (client *GRPCClient) NewConnection(address string, tlsOptions ...TLSOption)
 	}
 	return conn, nil
 }
+
```
mastersingh24: any reason this is not in creds.go where we have the analog to this for the server?

core/comm/client.go
```diff
@@ -213,3 +213,37 @@ func (client *GRPCClient) NewConnection(address string, tlsOptions ...TLSOption)
...
+type DynamicClientCredentials struct {
+	TLSConfig  *tls.Config
+	TLSOptions []TLSOption
+}
+
+func (dtc *DynamicClientCredentials) latestConfig() *tls.Config {
+	tlsConfigCopy := dtc.TLSConfig.Clone()
```
mastersingh24: Why Clone() ?

mastersingh24 (changes requested):

core/comm/client.go
yacovm: Will move it

core/comm/client.go
yacovm: Just for safety, to be extra cautious. 
This is a **tls.Config** so I make no assumptions on whether it can hold internal state or not by an existing TLS connection, and this has negligible computation overhead since it is just copying pointers anyway.

yacovm: Also note that we apply the mutator functions later on in this function, so I wanted the functions to be applied on a fresh shallow copy (which is deep enough of a copy, given the TLS config struct)

core/comm/client.go
yacovm: moved it

mastersingh24 (approved):

C0rWin (approved):

PR #379, opened 2019-12-08
[FAB-17220] Dynamically build TLS config in Raft client handshake

yacovm:
When we expand the root TLS CA in the channel config, *after*
Raft membership has expanded with an OSN that is issed a certificate
by a new TLS CA, the TLS client handshake uses the old root CA pool
and as a result the added orderer cannot be reached by the existing ones,
because their dialers reject its certificate.

This change set builds a dynamic transport credentials that
re-computes the TLS config in every TLS client handshake.

Expanded an integration test to ensure this works.

Change-Id: I6578ba49f16e14b97eb4eef4feccdecbfe1b7015

### DIFF
--- a/core/comm/client.go
+++ b/core/comm/client.go
@@ -14,7 +14,6 @@ import (
 
 	"github.com/pkg/errors"
 	"google.golang.org/grpc"
-	"google.golang.org/grpc/credentials"
 	"google.golang.org/grpc/keepalive"
 )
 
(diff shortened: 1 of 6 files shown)


jyellick (approved):

lindluni:
We did cherry-pick the CI changes from master into 1.4 if you want to rebase to pick up the parallel test splitting, it will still fail consistently, but it will be way faster. We haven't gotten to trying to figure out what is wrong with 1.4, we are focused on master this week with the 2.0 release

C0rWin (approved):

PR #380, opened 2019-12-09
[FAB-17135] Document private data enhancements in v2.0

denyeart:
#### Type of change

- Documentation update

#### Description

Document private data enhancements in v2.0 including:
- collection level endorsement policy
- implicit org-specific collections
- inclusion of private data in events
- memberOnlyWrite property

#### Related issues

https://jira.hyperledger.org/browse/FAB-17135

### DIFF
--- a/docs/source/private-data-arch.rst
+++ b/docs/source/private-data-arch.rst
@@ -72,10 +72,28 @@ Collection definitions are composed of the following properties:
 * ``memberOnlyRead``: a value of ``true`` indicates that peers automatically
   enforce that only clients belonging to one of the collection member organizations
   are allowed read access to private data. If a client from a non-member org
-  attempts to execute a chaincode function that performs a read of a private data,
+  attempts to execute a chaincode function that performs a read of a private data key,
   the chaincode invocation is terminated with an error. Utilize a value of
   ``false`` if you would like to encode more granular access control within
   individual chaincode functions.
+
+* ``memberOnlyWrite``: a value of ``true`` indicates that peers automatically
+  enforce that only clients belonging to one of the collection member organizations
+  are allowed to write private data from chaincode. If a client from a non-member org
+  attempts to execute a chaincode function that performs a write on a private data key,
+  the chaincode invocation is terminated with an error. Utilize a value of
+  ``false`` if you would like to encode more granular access control within
+  individual chaincode functions, for example you may want certain clients
+  from non-member organization to be able to create private data in a certain
+  collection.
+
+* ``endorsementPolicy``: An optional endorsement policy to utilize for the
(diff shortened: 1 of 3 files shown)


pamandrejko (commented):
Peer_event_services.rst
Contains: `By default, both services use the Channel Readers policy to determine whether
to authorize requesting clients for events.`
Should we replace `both` will `all` the services in this case?

pamandrejko (commented):
endorsement-policies.rst
Line 95 
Notice that, if the identity classification is enabled (see :doc:`msp`), one can
use the ``PEER`` role to restrict endorsement to only peers.
A better link would be  (see :doc:`msp#identity-classification`)

docs/source/private-data-arch.rst
```diff
@@ -183,6 +210,28 @@ along with the data in the chaincode APIs, for example
 
 A single chaincode can reference multiple collections.
 
+Referencing implicit collections from chaincode
+-----------------------------------------------
+
+Starting in v2.0, an implicit private data collection exists for each organization
+in a channel, so that you don't have to define collections if you'd like to
+utilize per-organization collections. Each org-specific implicit collection
```
pamandrejko: Would this be `an implicit private data collection is automatically created for each organization
in a channel when the organization joins the channel.`?

What happens if an organization leaves the channel, does the implicit collection go away/is removed?
I see this text below: `Note that collections cannot be deleted, as there may be prior private data hashes on the channel’s blockchain that cannot be removed.`

denyeart:
> Peer_event_services.rst
> Contains: `By default, both services use the Channel Readers policy to determine whether to authorize requesting clients for events.`
> Should we replace `both` will `all` the services in this case?

done

denyeart:
> endorsement-policies.rst
> Line 95
> Notice that, if the identity classification is enabled (see :doc:`msp`), one can
> use the `PEER` role to restrict endorsement to only peers.
> A better link would be (see :doc:`msp#identity-classification`)

I couldn't figure out the linking, we can add afterwards...

docs/source/private-data-arch.rst
denyeart: I updated the text a bit... implicit private data collections don't really exist until they are used.
And yes, just like regular private data collections, they will remain even after an org is removed from a channel since there are corresponding hashes on the blockchain.

manish-sethi: No, the implicit collection is not deleted when an organization leave the channel. The other organizations would still have hashes of the private data in this collection.

pamandrejko (commented):
LGTM

manish-sethi (approved):

PR #383, opened 2019-12-09
FAB-17170 externalbuilder pass metadata unmangled

jyellick:
#### Type of change

- Bug fix

#### Description

Today, the external builder receives a structure representing the three
fields of the metadata necessary to invoke the old docker platform
builds.  It then takes these fields and reserializes them to the JSON
metadata document.

In fact, the externalbuilder has no need for these deserialized fields,
and should pass the JSON document through without modifying it or losing
any additional fields.  This CR causes the package retrieving code to
return both the JSON document and the underlying bytes.  The underlying
bytes are passed to the externalbuilder while the three field structure
is passed to the dockerbuilder.

These paths are really quite bifurcated and could probably be further
split, with different package retrieving calls, it didn't seem worth it
purely to address this bug.


guoger (dismissed):
I like the rename...

mastersingh24 (dismissed):
This could be cleaner, but not worth the rework at this point just to fix the bug

guoger (dismissed):
I guess the latest force-push is a simple rebase... still puzzled by this UX of github :(

denyeart (approved):

PR #384, opened 2019-12-09
FAB-17160 Ensure peer can be started without any docker requirements

wlahti:
With the introduction of external builders for chaincode, docker is no longer required. This PR
ensures that a peer can be started with an external builder/launcher for chaincode and zero docker requirements. 

Signed-off-by: Will Lahti <wtlahti@us.ibm.com>

### DIFF
--- a/internal/peer/node/start.go
+++ b/internal/peer/node/start.go
@@ -495,42 +495,46 @@ func serve(args []string) error {
 		CCID:            scc.ChaincodeID(lifecycle.LifecycleNamespace),
 		HandlerRegistry: chaincodeHandlerRegistry,
 	}
-	var client *docker.Client
-	if coreConfig.VMDockerTLSEnabled {
-		client, err = docker.NewTLSClient(coreConfig.VMEndpoint, coreConfig.DockerCert, coreConfig.DockerKey, coreConfig.DockerCA)
-	} else {
-		client, err = docker.NewClient(coreConfig.VMEndpoint)
-	}
-	if err != nil {
-		logger.Panicf("cannot create docker client: %s", err)
+
+	if coreConfig.VMEndpoint == "" && len(coreConfig.ExternalBuilders) == 0 {
+		logger.Panic("VMEndpoint not set and no ExternalBuilders defined")
 	}
 
 	chaincodeConfig := chaincode.GlobalConfig()
 
-	dockerVM := &dockercontroller.DockerVM{
-		PeerID:        coreConfig.PeerID,
-		NetworkID:     coreConfig.NetworkID,
-		BuildMetrics:  dockercontroller.NewBuildMetrics(opsSystem.Provider),
-		Client:        client,
-		AttachStdOut:  coreConfig.VMDockerAttachStdout,
-		HostConfig:    getDockerHostConfig(),
-		ChaincodePull: coreConfig.ChaincodePull,
-		NetworkMode:   coreConfig.VMNetworkMode,
-		PlatformBuilder: &platforms.Builder{
-			Registry: platformRegistry,
-			Client:   client,
-		},
-		// This field is superfluous for chaincodes built with v2.0+ binaries
-		// however, to prevent users from being forced to rebuild leaving for now
-		// but it should be removed in the future.
-		LoggingEnv: []string{
-			"CORE_CHAINCODE_LOGGING_LEVEL=" + chaincodeConfig.LogLevel,
-			"CORE_CHAINCODE_LOGGING_SHIM=" + chaincodeConfig.ShimLogLevel,
-			"CORE_CHAINCODE_LOGGING_FORMAT=" + chaincodeConfig.LogFormat,
-		},
-	}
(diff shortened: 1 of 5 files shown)


internal/peer/node/start.go
```diff
@@ -496,13 +496,9 @@ func serve(args []string) error {
 		HandlerRegistry: chaincodeHandlerRegistry,
 	}
 	var client *docker.Client
```
guoger: as i commented in jira, i think we have fallback logic in https://github.com/hyperledger/fabric/blob/b430f818b3ab955776255ca8ac2f977aa8f409b1/core/container/container.go#L107-L118 and if we leave `var client docker.Client` nil here, that would trigger a panic. I think we'll need to change the logic in `DockerVM.Build` to error out

guoger (changes requested):

internal/peer/node/start.go
```diff
@@ -1076,6 +1072,18 @@ func computeChaincodeEndpoint(chaincodeAddress string, chaincodeListenAddress st
...
 
+func createDockerClient(coreConfig *peer.Config) (client *docker.Client, err error) {
+	if coreConfig.VMDockerTLSEnabled {
+		client, err = docker.NewTLSClient(coreConfig.VMEndpoint, coreConfig.DockerCert, coreConfig.DockerKey, coreConfig.DockerCA)
+	} else {
+		client, err = docker.NewClient(coreConfig.VMEndpoint)
+	}
+	if err != nil {
+		logger.Panicf("cannot create docker client: %s", err)
+	}
```
mastersingh24: I think this should be moved back up in the block at line 500

internal/peer/node/start.go
```diff
@@ -496,13 +496,9 @@ func serve(args []string) error {
...
 	var client *docker.Client
-	if coreConfig.VMDockerTLSEnabled {
-		client, err = docker.NewTLSClient(coreConfig.VMEndpoint, coreConfig.DockerCert, coreConfig.DockerKey, coreConfig.DockerCA)
-	} else {
-		client, err = docker.NewClient(coreConfig.VMEndpoint)
-	}
-	if err != nil {
-		logger.Panicf("cannot create docker client: %s", err)
+
+	if coreConfig.VMEndpoint != "" {
```
mastersingh24, line 507: shouldn't we also add a check for coreConfig.ExternalBuilders[] being empty ... we need either a VMEndpoint or a list of ExternalBuilders

mastersingh24 (dismissed):

internal/peer/node/start.go
jyellick: +1 at the very least we should print a warning indicating that the peer will not be able to execute chaincode.

I suppose it's possible that someone simply wants to stand up a peer to pull and validate blocks, though without a chaincode to access state, I don't see any real usefulness for this configuration.

internal/peer/node/start.go
wlahti: Oops, agreed.

denyeart:
Remember to add commit message and use PR template.

internal/peer/node/start.go
wlahti: Done

internal/peer/node/start.go
wlahti: Done

internal/peer/node/start.go
wlahti: Done

internal/peer/node/start.go
guoger: @jyellick one use case could be to construct state db for direct query against it?

internal/peer/node/start.go
```diff
@@ -495,14 +495,17 @@ func serve(args []string) error {
...
-		client, err = docker.NewClient(coreConfig.VMEndpoint)
+
+	if coreConfig.VMEndpoint == "" && len(coreConfig.ExternalBuilders) == 0 {
+		logger.Panic("VMEndpoint not set and no ExternalBuilders defined")
 	}
-	if err != nil {
-		logger.Panicf("cannot create docker client: %s", err)
+
+	var client *docker.Client
+	if coreConfig.VMEndpoint != "" {
```
guoger, line 507: shouldn't we also move the construct of `dockerVM` into this branch? and at fallback point, if `router.DockerVM == nil` we would return early and skip package retrieval.

guoger (changes requested):

internal/peer/node/start.go
jyellick: It's easy enough to create a no-op external builder -- for instance:

```
mkdir -p mybuilder/bin
ln -s /bin/false mybuilder/bin/detect
```

Now you have a functioning external builder that will never build anything, and you can successfully make your peer start.  If we find lots of users needing to do a workaround like this, we can look into relaxing the check, but we can always relax checks, but adding constraints later is hard.

jyellick (dismissed):

guoger (approved):
LGTM

jyellick (approved):

PR #385, opened 2019-12-10
FAB-17170 Peer CLI should encode mdata lowercase

jyellick:
#### Type of change

- Bug fix

#### Description

Once upon a time, the JSON fields in the chaincode package metadata.json
were upper case.  The peer CLI was written to reflect this fact.  When
they were changed to lower-case, the peer CLI was not updated, but
because of the fuzzy JSON matching rules in the golang JSON decoded it
was not noticed.  However, for external builders, the inconsistency is
more likely to be noticed as tools like jq etc. may not default to this
same fuzzy case-insensitive matching.

This change simply modifies the peer CLI to encode the metadata fields
as lower case.


jyellick:
Would self +2 if I could, this one is trivial.

denyeart (approved):

muralisrini (approved):

PR #392, opened 2019-12-11
FAB-15835 check error in deliverclient

guoger:
Add error check when flushing success after reading a block.
This does not fix the integration test flake described in JIRA but
hopefully makes it more investigable.

this is a cherry-pick from master

Change-Id: Ia00f6a41bd45786c476add04aa4c15e86baf75d4

### DIFF
--- a/peer/common/deliverclient.go
+++ b/peer/common/deliverclient.go
@@ -71,11 +71,26 @@ func (d *DeliverClient) readBlock() (*cb.Block, error) {
 	}
 	switch t := msg.Type.(type) {
 	case *ab.DeliverResponse_Status:
-		logger.Infof("Got status: %v", t)
+		logger.Infof("Expect block, but got status: %v", t)
 		return nil, errors.Errorf("can't read the block: %v", t)
 	case *ab.DeliverResponse_Block:
 		logger.Infof("Received block: %v", t.Block.Header.Number)
-		d.Service.Recv() // Flush the success message
+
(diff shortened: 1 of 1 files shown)


peer/common/deliverclient.go
```diff
@@ -71,11 +71,16 @@ func (d *DeliverClient) readBlock() (*cb.Block, error) {
...
 	switch t := msg.Type.(type) {
 	case *ab.DeliverResponse_Status:
-		logger.Infof("Got status: %v", t)
+		logger.Infof("Expect block, but got status: %v", t)
 		return nil, errors.Errorf("can't read the block: %v", t)
 	case *ab.DeliverResponse_Block:
 		logger.Infof("Received block: %v", t.Block.Header.Number)
-		d.Service.Recv() // Flush the success message
+		if resp, err := d.Service.Recv(); err != nil { // Flush the success message
+			logger.Errorf("Failed to flush success message: %s", err)
```
kchristidis, line 83: This is not the log message I was expecting to see here, but it's a nit admittedly, and I see your point about this being a means to an end.

kchristidis (dismissed):

peer/common/deliverclient.go
guoger: updated to be more precise, hopefully we could spot the ghost and get rid of these "debugging" code

guoger:
https://github.com/hyperledger/fabric/pull/413 should be able to resolve the flakiness, no need to add extra debugging info for now

PR #393, opened 2019-12-11
Allow seperate TLS config for cluster and client

guoger:
When intra cluster communication is configured to use seperate
listener, allow TLS being disabled for the client-facing listener,
and assume TLS is always enabled for cluster listener (no extra
config option).

FAB-15648

This is cherry-pick from master


jyellick (approved):
Seems like a near-cherry-pick of what's in master.

PR #397, opened 2019-12-11
Use org1peer0 in lifecycle integration test

wlahti:
#### Type of change
- Bug fix

#### Description
Given the recent sorting in the nwo PeersWithChannel function, use org1.peer0 for the various approval steps in the lifecycle test. This resolves the flakiness of this specific test but we'll still need to investigate why pulling the private data via gossip doesn't work consistently before the timeout. 

#### Related issues

FAB-16731


jyellick (approved):
Thanks for catching and fixing this.  It was my intent with #389 as I thought these peer names were 1 indexed, but seems like in this case we are zero indexed.

denyeart (approved):

PR #386, opened 2019-12-10
[FAB-16906] Chaincode launcher doc

joealewine:
Change-Id: I78dd2dac59e2343b2082fcef3f4c2b31a764f175
Signed-off-by: joe-alewine <Joe.Alewine@ibm.com>

#### Type of change

- Documentation update

#### Description

Initial doc for chaincode launcher functionality. 

#### Related issues

FAB-14086

PR #399, opened 2019-12-11
[FAB-14083] Chaincode as an external service documentation

pamandrejko:

Adding a new topic to the documentation for configuring chaincode as an external service.

- Documentation update

#### Description

Content includes an entry in the Operations Guide toc and the new topic which describes the process for configuring Chaincode as an external service.
Source material came from Murali here: https://docs.google.com/document/d/1WhJmbHt7JXNCStXWyM7eaTxUaA9PAlfHwVPI5RUnOAQ/edit?usp=sharing

#### Related issues
FAB-16906 External Builder and Launcher https://github.com/hyperledger/fabric/pull/386

#### Release Note
@denyeart We need to update the release notes and what's new if you have not already.


docs/source/cc_service.md
```diff
@@ -0,0 +1,297 @@
...
+
+## Configuring a peer to process external chaincode
+
+If you have reviewed the external build and chaincode launcher documentation then these steps will be familiar to you. We leverage those scripts to define the external chaincode service information. These scripts are located on the peer file system that is accessible to the peer process from the `core.yaml` file in the `externalBuilders` element under the `chaincode` stanza. An example is included in the steps which follow.
+
+### Create the set of external builder and launcher scripts on the peer
+
+In order to configure chaincode as an external service, you must use the scripts as follows:
+* **detect** - examines the chaincode package and accept if metadata.json `type` is set to `external`.
+* **build** - builds the chaincode and places the build artifacts in the `BUILD_OUTPUT_DIR`. For chaincode as an external service, the peer is not does not build the chaincode. Instead, the script extracts the chaincode endpoint information in the `connection.json` file and places it in the specified location.
```
denyeart:
```suggestion
* **build** - builds the chaincode and places the build artifacts in the `BUILD_OUTPUT_DIR`. For chaincode as an external service, the peer does not build the chaincode. Instead, the script extracts the chaincode endpoint information in the `connection.json` file and places it in the specified location.
```

docs/source/cc_service.md
```diff
@@ -0,0 +1,297 @@
...
+
+  # if the runtime is managed externally, we don’t have to start the chaincode from this script
+  # just return success
+
+  exit 0
+```
+
+## Writing chaincode to run as an external service
+
+Currently, the chaincode as an external service model model is only supported by GO chaincode shim. In Fabric v2.0, the GO shim API adds a `ChaincodeServer` type that developers should use to create a chaincode server.  The `Invoke` and `Query` APIs are unaffected. Developers should write to the `shim.ChaincodeServer` API, then build the chaincode and run it in the external environment of choice. Here is a simple sample chaincode program to illustrate the pattern:
```
denyeart:
```suggestion
Currently, the chaincode as an external service model is only supported by GO chaincode shim. In Fabric v2.0, the GO shim API adds a `ChaincodeServer` type that developers should use to create a chaincode server.  The `Invoke` and `Query` APIs are unaffected. Developers should write to the `shim.ChaincodeServer` API, then build the chaincode and run it in the external environment of choice. Here is a simple sample chaincode program to illustrate the pattern:
```

docs/source/cc_service.md
```diff
@@ -0,0 +1,297 @@
...
+                }
+        err := server.Start()
+        if err != nil {
+                fmt.Printf("Error starting Simple chaincode: %s", err)
+        }
+}
+```
+The key to running the chaincode as an external service is the use of `shim.ChaincodeServer`. This uses the new shim API `shim.ChaincodeServer` with the chaincode service properties described below:
+
+* **CCID** (string)- CCID should match chaincode's package name on peer. This the `CCID` associated with the installed chaincode as returned by the `peer lifecycle chaincode install <package>` CLI command. This can be obtained post-install using the "peer lifecycle chaincode queryinstalled" command.
```
denyeart:
```suggestion
* **CCID** (string)- CCID should match chaincode's package name on peer. This is the `CCID` associated with the installed chaincode as returned by the `peer lifecycle chaincode install <package>` CLI command. This can be obtained post-install using the "peer lifecycle chaincode queryinstalled" command.
```

denyeart (commented):
Did a quick read through.... Jason said he'll do technical review...

docs/source/cc_service.md
```diff
@@ -0,0 +1,297 @@
+# Chaincode as an external service
+
+Fabric v2.0 supports chaincode deployment and execution outside of Fabric that enables users to manage a chaincode runtime independently of the peer. This facilitates deployment of chaincode on Fabric cloud deployments such as Kubernetes. Instead of installing the chaincode on every peer, chaincode can now run as a service as opposed to a Fabric launched client. This capability leverages the Fabric v2.0 external builder and launcher functionality which enables operators to extend a peer with programs to build, launch, and discover chaincode. Before reading this topic you should become familiar with the [External Builds and Launchers](./cc_launcher.html) content.
```
jyellick: > Instead of installing the chaincode on every peer, chaincode can now run as a service as opposed to a Fabric launched client.

Grammar problem here.  

Instead of building and launching the chaincode on every peer, chaincode can now run as a service who's lifecycle is managed outside of Fabric.

docs/source/cc_service.md
```diff
@@ -0,0 +1,297 @@
...
+
+Prior to the availability of the external build and launcher functionality, the chaincode package would contain the source code for the chaincode. The new external build and launcher functionality now allows users to optionally customize the build process. And with respect to running the chaincode as an external service, the build process allows you to specify the endpoint information of the server where the chaincode is running. Hence the package simply consists of the externally running chaincode server endpoint information and TLS artifacts for secure connection. TLS is optional but highly recommended for all environments except a simple test environment.
```
jyellick: > Prior to the availability of the external build and launcher functionality, the chaincode package would contain the source code for the chaincode.

This is a bit disingenuous.  I would say "Prior to the availability of the external builders, the chaincode package content was required to be a set of source code files for a particular language which could be built and launched as a chaincode binary."

With the availability of external launchers, I expect many packages will continue to follow exactly this pattern, it is simply no longer a requirement.

docs/source/cc_service.md
```diff
@@ -0,0 +1,297 @@
...
+* [Packaging chaincode](#packaging-chaincode)
+* [Configuring a peer to process external chaincode](#configuring-a-peer-to-process-external-chaincode)
+* [External builder and launcher sample scripts](#external-builder-and-launcher-sample-scripts)
+* [Writing chaincode to run as an external service](#writing-chaincode-to-run-as-an-external-service)
+* [Deploying the chaincode](#deploying-the-chaincode)
+* [Running the chaincode as an external service](#running-the-chaincode-as-an-external-service)
+
+## Packaging chaincode
+
+With the Fabric v2.0 chaincode lifecycle, chaincode is packaged and installed in a `.targz` format. The following `myccpackage.tgz` archive  demonstrates the required structure:
```
jyellick: > `.targz`
`.tar.gz`

docs/source/cc_service.md
```diff
@@ -0,0 +1,297 @@
...
+
+```sh
```
jyellick, line 20: I think Matt's external builder doc does a clearer job of explaining the new packaging format, do we want to duplicate this info here?

docs/source/cc_service.md
```diff
@@ -0,0 +1,297 @@
...
+$ tar xvfz myccpackage.tgz
+code.tar.gz
+metadata.json
+```
+
+### code.tar.gz archive requirements
+
+The `code.tar.gz` archive must include a `connection.json` file which contains the chaincode endpoint and certificate information:
```
jyellick: This comes off as a little overly prescriptive to me.  Including a `connection.json` is the most obvious way to do it, but you could encode this information however you want, the information could come from a network resource, it really just matters that your external builder can produce a `connection.json` during its release step.

Overall, I think this section could be better restructured as 'an example external builder for a

docs/source/cc_service.md
```diff
@@ -0,0 +1,297 @@
...
+  "dial_timeout": "10s",
+  "tls_required": true,
+  "client_auth_required": "true",
+  "key_path": "path/rooted/in/release/directory/key.pem",
+  "cert_path": "path/rooted/in/release/directory/key.pem",
+  "root_cert_path": "path/rooted/in/release/directory/key.pem"
+}
+```
+
+**Note:** The TLS files can be placed anywhere in the `code.tar.gz` archive because the directory of the contents of the `.targz` will be provided to the external chaincode builder scripts. The `/bin/release` script, described later in this document, should move the files to the appropriate paths relative to the release directory.
```
jyellick: Maybe add another note that there is a JIRA open https://jira.hyperledger.org/browse/FAB-17228 to remove these file components and have the PEM come in as part of the JSON. (So this will very likely change before release)

docs/source/cc_service.md
```diff
@@ -0,0 +1,297 @@
...
+
+### metadata.json requirements
+
+The metadata.json file contains three keys:
```
jyellick: It need only the label key.  All other keys are free-form.  `type` is a suggested key for associating your package to a particular builder, but it is not required.

muralisrini (commented):
just a few nits @pamandrejko

docs/source/cc_service.md
```diff
@@ -0,0 +1,297 @@
...
+* **run** - Not required when using chaincode as an external service if the runtime of chaincode running on the external server is manged externally and the chaincode need not be started from this script.
+
+The scripts are required to exist in the peer `/bin` directory:
+```
+    <fully qualified path on the peer's env>
+    └── bin
+        ├── build
+        ├── detect
+        ├── release
+        └── run
```
jyellick: We don't need a run binary, should be fine to omit it.

docs/source/cc_service.md
```diff
@@ -0,0 +1,297 @@
...
+    echo "Expected 3 directories got $#"
+    exit 1
+fi
+
+SOURCE=$1
+OUTPUT=$3
+
+#external chaincodes expect connection.json file in the chaincode package
+if [ ! -f "$SOURCE/connection.json" ]  ; then
+    echo "$SOURCE/connection.json not found" > /tmp/err.txt
```
jyellick: Let's instead follow the pattern:

```
>&2 echo "$SOURCE/connection.json not found"
```

This ensures that the output goes into the peer logs and there's no contention for the tmp file.

docs/source/cc_service.md
```diff
@@ -0,0 +1,297 @@
...
+A typical `detect` script could contain:
+
+```sh
+
+#!/bin/bash
+
+set -euo pipefail
+
+if [ "$#" -ne 2 ]; then
+    echo "Expected 2 directories got $#"
```
jyellick: Ideally, all of these echos should be prefixed by `>&2` so as to not lose the output and instead redirect it to the peer logs.

docs/source/cc_service.md
```diff
@@ -0,0 +1,297 @@
...
+A typical `build` script could contain:
+
+```sh
+
+#!/bin/bash
+
+set -euo pipefail
+
+if [ "$#" -ne 3 ]; then
+    echo "Expected 3 directories got $#"
```
jyellick: Again.

docs/source/cc_service.md
```diff
@@ -0,0 +1,297 @@
...
+```sh
+
+#!/bin/bash
+
+set -euo pipefail
+
+set -x
+
+if [ "$#" -ne 2 ]; then
+    >&2 echo "Expected 2 directories got $#"
```
jyellick, line 191: This one looks nice already :) consistency.

docs/source/cc_service.md
```diff
@@ -0,0 +1,297 @@
...
+    exit 2
+fi
+
+BLD="$1"
+RELEASE="$2"
+
+#external chaincodes expect artifacts to be placed under "$RELEASE"/chaincode/server
+if [ -f $BLD/connection.json ]; then
+   mkdir -p "$RELEASE"/chaincode/server
+       cp $BLD/connection.json "$RELEASE"/chaincode/server
```
jyellick: Maybe a tab instead of spacing? But misallignment.

docs/source/cc_service.md
```diff
@@ -0,0 +1,297 @@
...
+       exit 0
+fi
+
+exit 1
+```    
+
+### /bin/run
```
jyellick: Let's not include this for completeness, it's not needed.  And if we were to include it, I would suggest it always return failure, not success.

jyellick (dismissed):

docs/source/cc_service.md
muralisrini: This merely talks about the files in `myccpackage.tgz` just so we can go into what the contents of those files would be for chaincode as a service use. The link for the complete explanation of the external builder is provided as a requisite in the opening paragraph,

docs/source/cc_service.md
muralisrini: Agreed. How about something like this 
```
The `code.tar.gz` archive must include information obtaining connection information for the chaincode endpoint. This information is packaged into `connection.json` in the `release` step (see below). In this example we package `connection.json` directly in `code.tar.gz` so the `release` step can just copy it over.
```

docs/source/cc_service.md
muralisrini: Well,  the user _could_ use it to do other things (including launching the chaincode locally or elsewhere). It would be nice to indicate that possibility .... we could indicate that in the doc and get rid of `run`.

docs/source/cc_service.md
muralisrini: Right, didn't mean to imply that original packaging alternative is gone.

docs/source/cc_service.md
```diff
@@ -0,0 +1,297 @@
...
+code.tar.gz
+metadata.json
+```
+
+### code.tar.gz archive requirements
+
+The `code.tar.gz` archive must include a `connection.json` file which contains the chaincode endpoint and certificate information:
+
+* **address** - chaincode server endpoint accessible from peer. Must be specified in “<host>:<port>” format.
+* **Dial_timeout** - interval to wait for connection to complete. Specified as a string qualified with units (examples : "10s", "500ms", "1m"). Default is “3s” if not specified.
```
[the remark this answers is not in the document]
pamandrejko: done

docs/source/cc_service.md
```diff
@@ -0,0 +1,297 @@
...
+* **tls_required** - true or false. If false "client_auth_required", "key_path", "cert_path", "root_cert_path" are not required. Default is “true”.
+* **client_auth_required** - if true then you need to specify "key_path" and "cert_path" for client authentication. Default is false. Ignored if tls_required is false.
+* **key_path** - path to the key file. This path is relative to the “release directory” (see “release” below). Required if client_auth_required is true. Ignored if tls_required is false.
+* **cert_path**  - path to the certificate file. This path is relative to the “release directory” (see “release” below). Must be specified if client_auth_required is true. Ignored if tls_required is false.
```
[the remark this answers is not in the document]
pamandrejko: Done

docs/source/cc_service.md
```diff
@@ -0,0 +1,297 @@
...
+
+For example:
+```json
+{
+  "address": "your.chaincode.host.com:9999",
+  "dial_timeout": "10s",
+  "tls_required": true,
+  "client_auth_required": "true",
+  "key_path": "path/rooted/in/release/directory/key.pem",
+  "cert_path": "path/rooted/in/release/directory/key.pem",
```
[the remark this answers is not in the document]
pamandrejko: done

docs/source/cc_service.md
```diff
@@ -0,0 +1,297 @@
...
+  "root_cert_path": "path/rooted/in/release/directory/key.pem"
```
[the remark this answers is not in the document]
pamandrejko: done

docs/source/cc_service.md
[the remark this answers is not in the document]
pamandrejko: done

docs/source/cc_service.md
pamandrejko: done

docs/source/cc_service.md
pamandrejko: done

docs/source/cc_service.md
sykesm: Agree with Jason here. There's a lot of flexibility, adding `connection.json` as a code artifact is _one_ way; you could also include an empty `code.tar.gz` and put the connection information into `metadata.json` to extract as necessary in build/release.

pamandrejko: I changed it to
`The `code.tar.gz` archive must include connection information for the chaincode endpoint. This information is packaged into `connection.json` in the `release` step (see below). In this example we package `connection.json` directly in `code.tar.gz` so the `release` step can just copy it over.`

docs/source/cc_service.md
sykesm: Why Upper_camel ? Doesn't adhere to either of the conventions we're currently using (camelCase and snake_case).

docs/source/cc_service.md
```diff
@@ -0,0 +1,297 @@
...
+* **client_auth_required** - if true then you need to specify "key_path" and "cert_path" for client authentication. Default is false. Ignored if tls_required is false.
+* **key_path** - path to the key file. This path is relative to the “release directory” (see “release” below). Required if client_auth_required is true. Ignored if tls_required is false.
+* **cert_path**  - path to the certificate file. This path is relative to the “release directory” (see “release” below). Must be specified if client_auth_required is true. Ignored if tls_required is false.
+* **root_cert_path**  - path to the root cert for authenticating the server. Required when tls_required is set to true.
+
+For example:
+```json
+{
+  "address": "your.chaincode.host.com:9999",
+  "dial_timeout": "10s",
```
sykesm, line 42: I guess it is `lower_snake_case` so the name in the doc needs to change.

docs/source/cc_service.md
sykesm: This seems to unnecessarily overlap with the general external builder doc. Would very much like to avoid that.

IMO, this doc should focus on `connection.json`, details around when and how a connection is initiated to running chaincode, the implication of things like indexes, and how to build chaincode to run as a server.

From Pam:
I changed this to:
```
### metadata.json requirements

When using chaincode as an external service, the "type" field in `metadata,json` file must be set to `external`. For example:


{"path":"","type":"external","label":"mycc"}

```
From Murali:
Its not a requirement from fabric  @pamandrejko (the `must be set to external` suggests it is). Users can use any means to inspect the package and check for external. How about something like the following ?

```
### use metadata.json to check for external chaincode service
We will use a "type" field in `metadata,json` file to indicate external chaincode service
    {"path":"","type":"external","label":"mycc"}
```

sykesm (dismissed):

docs/source/cc_service.md
pamandrejko: In general we try not to link to Jiras in our doc or even mention future potential functionality because often it does not get implemented.

As a compromise I added - 
`In the future, the PEM may be included in the JSON itself. `

The thing is we need to remember to come back and update the doc around this.

docs/source/cc_service.md
pamandrejko: done

docs/source/cc_service.md
pamandrejko: @muralisrini I thought type is required here (to be `external`) for your detect script to work? Or is that just an example of how you could do it?

docs/source/cc_service.md
pamandrejko: done

docs/source/cc_service.md
pamandrejko: done

docs/source/cc_service.md
pamandrejko: Done

docs/source/cc_service.md
pamandrejko: @muralisrini - I find this a little confusing here. Unless we could come up with a bona fide use case for why you would want to do this (it seems outside the scope of what we are trying to convey here?) 

Towards the top of the doc where we describe the 4 folders, I've removed `bin/run` and and replaced it with 
`You may notice that the external builder and launcher `bin/run` script is not required for the chaincode as an external service.`

And I removed the 
### /bin/run
section

docs/source/cc_service.md
pamandrejko: done.

docs/source/cc_service.md
pamandrejko: done

docs/source/cc_service.md
pamandrejko: done

docs/source/cc_service.md
pamandrejko: done

docs/source/cc_service.md
pamandrejko: I kind of agree with Murali here. Readers need to know what they need to do differently for this to work with the chaincode as an external service which is described in the subsections.
But to address your concern, I did go ahead and add a link here directly to the section in Matt's doc that talks about packaging.

docs/source/cc_service.md
muralisrini: Sounds good. Better to keep it simple and target the 90%. I'm sure users will utilize those possibilities on  need basis.

docs/source/cc_service.md
pamandrejko: I changed it to:
`When using chaincode as an external service, you can set the "type" field in `metadata.json` to `external` in order to indicate that the external service is being used. For example:

```json
{"path":"","type":"external","label":"mycc"}
````

**Review by denyeart — APPROVED** [review]

**Comment by denyeart** [comment]
comments addressed. lgtm.

File: docs/source/cc_service.md [hunk 23]
```diff
@@ -0,0 +1,276 @@
...
+
+```sh
+$ tar xvfz myccpackage.tgz
+code.tar.gz
+metadata.json
+```
+
+### code.tar.gz archive requirements
+
+The `code.tar.gz` archive must include connection information for the chaincode endpoint. This information is packaged into `connection.json` in the `/bin/release` step (see below). In this example we package `connection.json` directly in `code.tar.gz` so the `release` step can just copy it over.
```
sykesm, line 28: This still isn't correct. We are prescribing a sample implementation instead of documenting the function in fabric.

docs/source/cc_service.md
```diff
@@ -0,0 +1,276 @@
...
+For example:
+```json
+{
+  "address": "your.chaincode.host.com:9999",
+  "dial_timeout": "10s",
+  "tls_required": true,
+  "client_auth_required": "true",
+  "key_path": "path/rooted/in/release/directory/key.pem",
+  "cert_path": "path/rooted/in/release/directory/cert.pem",
+  "root_cert_path": "path/rooted/in/release/directory/rootcert.pem"
```
sykesm, line 47: There's a JIRA opened to change these to be values within the document instead of spreading them around the tree. That changes the key and the semantics and keep the definitions consistent between chaincode clients (what we have today) and chaincode servers (what this is talking about).

I really don't think it will make beta so this is mostly informational...

docs/source/cc_service.md
```diff
@@ -0,0 +1,276 @@
...
+
+### metadata.json requirements
+
+When using chaincode as an external service, you can set the `type` field in `metadata.json` to `external` in order to indicate that the external service is being used. For example:
+
+```json
+{"path":"","type":"external","label":"mycc"}
+```
+
+## Configuring a peer to process external chaincode
```
sykesm, line 61: I really think most of this unnecessarily overlaps with what we have in the other doc.

In my earlier comment I suggested the areas to focus on here; I'll go further and say that I think the connection.json should be documented or linked within the `/bin/release` portion of external builders as that's really the only intersection between the two.

muralisrini (approved):
Thanks Pam!

docs/source/cc_service.md
```diff
@@ -0,0 +1,276 @@
...
+```sh
+
+#!/bin/bash
+
+set -euo pipefail
+
+if [ "$#" -ne 2 ]; then
+    >&2 echo "Expected 2 directories got $#"
+    exit 2
+fi
```
sykesm, line 118: We're needlessly complicating examples by having these assertions. It looks like we're just copying the contents of tests.

sykesm (changes requested):

denyeart:
@muralisrini Will address the remaining comments in subsequent PR.

PR #407, opened 2019-12-12
External chaincode clients and servers should use consistent key/cert model

wlahti:
#### Type of change
- Bug fix

#### Description
External chaincode clients and servers should consistently provide PEM encoded certificates/keys instead of paths to files.

#### Related issues
FAB-17228


sykesm (approved):
lgtm

muralisrini (approved):

PR #412, opened 2019-12-13
Release 1.2

sunyiyao2019:
#### Type of change

- Bug fix
- New feature
- Test update
- Documentation update

#### Description
what's the different between the hyperledger fabric docker version 1.2.0 and latest. why i the run time of putstates gap is  so large.


guoger:
@sunyiyao2019 this looks like a question, and shouldn't be in form of PR. Please use mailing list or rocket.chat, or file a bug report in JIRA. Also, i don't quite understand your question, you might want to elaborate a bit. thx

PR #421, opened 2019-12-16
For avoiding SERVICE_UNAVAILABLE -- no Raft leader 

senthilkumar-ofs:
#### Type of change

- Bug fix
- New feature
- Test update
- Documentation update

### DIFF
--- a/docs/source/build_network.rst
+++ b/docs/source/build_network.rst
@@ -1160,7 +1160,7 @@ the network pass ``docker-compose-couch.yaml`` as well:
 
 .. code:: bash
 
-    docker-compose -f docker-compose-cli.yaml -f docker-compose-couch.yaml up -d
+    docker-compose -f docker-compose-cli.yaml -f docker-compose-couch.yaml -f docker-compose-etcdraft2.yaml up -d
 
 **abstore** should now work using CouchDB underneath.
 


senthilkumar-ofs (commented):
For avoiding SERVICE_UNAVAILABLE -- no Raft leader

lindluni:
We require signed commits, please follow the instructions here to sign your changes: https://github.com/hyperledger/fabric/pull/421/checks

guoger (changes requested):
Thanks for contribution! Please update PR message to reflect the intend of this change.

Also, i wonder why this change is needed, as single node orderer should also work. Are you seeing any errors running the sample?

senthilkumar-ofs:
> Thanks for contribution! Please update PR message to reflect the intend of this change.
> 
> Also, i wonder why this change is needed, as single node orderer should also work. Are you seeing any errors running the sample?

Yes. While running sample(first-network), got the above error. So updated the document

guoger:
> > Thanks for contribution! Please update PR message to reflect the intend of this change.
> > Also, i wonder why this change is needed, as single node orderer should also work. Are you seeing any errors running the sample?
> 
> Yes. While running sample(first-network), got the above error. So updated the document

normally it takes a few seconds for a leader to be elected, depending on election timeout config. So you may see this error if a channel has just been started, and you'll need to retry. Plus, adding more orderers would not solve the "problem".

Are you seeing this error constantly after several retry?

guoger:
close due to inactivity. pls reopen if needed.

PR #424, opened 2019-12-16
[FAB-17176] Make decode hooks consistent in viperutil

stephyee:

#### Type of change

Improvement to code

#### Related issues

https://jira.hyperledger.org/browse/FAB-17176

### DIFF
--- a/common/viperutil/config_util.go
+++ b/common/viperutil/config_util.go
@@ -109,163 +109,155 @@ func unmarshalJSON(val interface{}) (map[string]string, bool) {
 // customDecodeHook adds the additional functions of parsing durations from strings
 // as well as parsing strings of the format "[thing1, thing2, thing3]" into string slices
 // Note that whitespace around slice elements is removed
-func customDecodeHook() mapstructure.DecodeHookFunc {
+func customDecodeHook(f reflect.Type, t reflect.Type, data interface{}) (interface{}, error) {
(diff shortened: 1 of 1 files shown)


common/viperutil/config_util.go
```diff
@@ -109,163 +109,155 @@ func unmarshalJSON(val interface{}) (map[string]string, bool) {
...
-					var block *pem.Block
-					block, bytes = pem.Decode(bytes)
-					if block == nil {
-						break
-					}
-					if block.Type != "CERTIFICATE" || len(block.Headers) != 0 {
-						continue
-					}
-					result = append(result, string(pem.EncodeToMemory(block)))
+				if block.Type != "CERTIFICATE" || len(block.Headers) != 0 {
```
yacovm, line 249: I know this is like the original form, but - why do we really enforce a CERTIFICATE header here? 
This is a generic viper util, so what if we have here a public key or a private key ? Then the header is different. 

Also, i don't think that PEM parsing cares about header type at all - it only cares about -- BEGIN and -- END prefixes and ---- suffixes. 

Served as "food for thought" only.

jyellick: Removing all of 'viperutil' is on our TODO list.  I think we'd be better off without the magic (but not quite correct or consistent magic) around things like PEM blocks.

jyellick (approved):
LGTM

PR #430, opened 2019-12-17
Increase open file limit to address Gossip flakes

lindluni:
Attempting to address gossip flakes by increasing open file limit

### DIFF
--- a/ci/install_deps.yml
+++ b/ci/install_deps.yml
@@ -7,6 +7,8 @@ steps:
       sudo apt-get clean
       sudo apt-get update
       sudo apt-get install -y libtool gcc make
+      echo "vsts  hard  nofile  65535" | sudo tee -a /etc/security/limits.conf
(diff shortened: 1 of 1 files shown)


ci/azure-pipelines.yml
```diff
@@ -44,20 +44,28 @@ jobs:
...
+#    dependsOn: VerifyBuild
     pool:
       vmImage: ubuntu-18.04
     steps:
       - template: install_deps.yml
+      - script: cat /proc/sys/fs/file-max
+      - script: whoami
+      - script: sudo cat /etc/security/limits.conf
+      - script: ulimit -Sn
+      - script: ulimit -Hn
```
mastersingh24: you may need to run `su - vsts` to get a new session so the new limits take effect

jyellick: Looks like you approved this change on release-1.4, so marking this resolved.

jyellick (approved):

PR #438, opened 2019-12-19
[FAB-17275] Add Raft option of InsecureSkipVerify

yeasy:
This patchset exposes the option to allow skip the hostname checking in
the configuration, which can allow OSN to work behind network middleware
and also ease the code testing.

Change-Id: Icbcf5c2511f02c44576785514826afceb99719bb

#### Type of change

#### Description

Currently, OSN enabled TLS authentication for the raft communications.

In cases that OSN stays behind firewall/proxy/NATed devices, the published endpoint might not match that in the TLS cert. And the user may not care about it because there's following validation logics to compare the tls cert with those inside the channel config.

Hence we should expose the option to allow skip the hostname checking in the configuration.

With this option enabled, the OSN still requires to leverage the cert matching logic that is hardcoded to compare with the certs inside the channel config.

#### Related issues

https://jira.hyperledger.org/browse/FAB-17275

### DIFF
--- a/core/comm/client.go
+++ b/core/comm/client.go
@@ -70,6 +70,7 @@ func (client *GRPCClient) parseSecureOptions(opts SecureOptions) error {
 	}
 
 	client.tlsConfig = &tls.Config{
+		InsecureSkipVerify:    opts.InsecureSkipVerify,
 		VerifyPeerCertificate: opts.VerifyCertificate,
 		MinVersion:            tls.VersionTLS12} // TLS 1.2 only
 	if len(opts.ServerRootCAs) > 0 {
--- a/sampleconfig/orderer.yaml
+++ b/sampleconfig/orderer.yaml
@@ -47,6 +47,10 @@ General:
         # Consensus messages are dropped if the buffer is full, and transaction
         # messages are waiting for space to be freed.
         SendBufferSize: 10
+        # InsecureSkipVerify will disable the checking on tls certificate chain and the hostname.
+        # Be careful to change to true, and only use this when doing test or the OSNs are behind
+        # network middle devices.
+        InsecureSkipVerify: false
         # ClientCertificate governs the file location of the client TLS certificate
         # used to establish mutual TLS connections with other ordering service nodes.
         ClientCertificate:
--- a/core/comm/config.go
+++ b/core/comm/config.go
@@ -110,6 +110,8 @@ type SecureOptions struct {
 	CipherSuites []uint16
 	// TimeShift makes TLS handshakes time sampling shift to the past by a given duration
 	TimeShift time.Duration
+	// InsecureSkipVerify disable the checking of the certificate chain and the hostname
+	InsecureSkipVerify bool
 }
 
 // KeepaliveOptions is used to set the gRPC keepalive settings for both
--- a/orderer/common/localconfig/config.go
+++ b/orderer/common/localconfig/config.go
@@ -74,6 +74,7 @@ type Cluster struct {
 	SendBufferSize                       int
 	CertExpirationWarningThreshold       time.Duration
 	TLSHandshakeTimeShift                time.Duration
+	InsecureSkipVerify                   bool
 }
 
 // Keepalive contains configuration for gRPC servers.
@@ -224,6 +225,7 @@ var Defaults = TopLevel{
 			ReplicationRetryTimeout:              time.Second * 5,
 			ReplicationPullTimeout:               time.Second * 5,
 			CertExpirationWarningThreshold:       time.Hour * 24 * 7,
+			InsecureSkipVerify:                   false,
 		},
 		LocalMSPDir: "msp",
 		LocalMSPID:  "SampleOrg",
--- a/orderer/common/server/main.go
+++ b/orderer/common/server/main.go
@@ -482,13 +482,14 @@ func initializeClusterClientConfig(conf *localconfig.TopLevel) comm.ClientConfig
 	}
 
 	cc.SecOpts = comm.SecureOptions{
-		TimeShift:         conf.General.Cluster.TLSHandshakeTimeShift,
-		RequireClientCert: true,
-		CipherSuites:      comm.DefaultTLSCipherSuites,
-		ServerRootCAs:     serverRootCAs,
-		Certificate:       certBytes,
-		Key:               keyBytes,
-		UseTLS:            true,
+		TimeShift:          conf.General.Cluster.TLSHandshakeTimeShift,
+		RequireClientCert:  true,
+		CipherSuites:       comm.DefaultTLSCipherSuites,
+		ServerRootCAs:      serverRootCAs,
+		Certificate:        certBytes,
+		Key:                keyBytes,
+		UseTLS:             true,
+		InsecureSkipVerify: conf.General.Cluster.InsecureSkipVerify,
 	}
 
 	return cc


mastersingh24:
I'm not a proponent of adding insecure options.  Way back when, I intentionally never exposed this option in the GRPCClient.

core/comm/client.go
```diff
@@ -70,6 +70,7 @@ func (client *GRPCClient) parseSecureOptions(opts SecureOptions) error {
 	}
 
 	client.tlsConfig = &tls.Config{
+		InsecureSkipVerify:    opts.InsecureSkipVerify,
```
mastersingh24, line 73: I really do not think we should allow this to be set.  Maybe we can allow the ServerName to be set (aka hostname override) but we should not allow this to be set.

yeasy: @mastersingh24 Thanks gari, I've thought of the same idea before. And the serverName can only be set with connecting to one server, but there will be multiple potential servers to talk with.

Adding this new option will not break existing compatibility, as by default it is turned off.

The OSN will check to match the tlscert with those stored inside the channel config. Hence even with this option enabled, the communication is still safe.

Besides, this is useful when the OSN stays behind network middlebox or doing tests.

I've added explanation in the configuration file to warn not setting this unless clear to know what it means.

yacovm (changes requested):
>  In cases that OSN stays behind firewall/proxy/NATed devices, the published endpoint might not match that in the TLS cert. And the user may not care about it because there's following validation logics to compare the tls cert with those inside the channel config.

There are 2 different use cases here you're trying to solve:

- The remote OSN publishes an endpoint **foo** but actually the certificate is valid for **bar**: in such a case, I think you should just make the remote OSN to publish an endpoint that matches the certificate, or just have the certificate match the endpoint. 
- The remote OSN is behind a TLS terminating proxy.


I don't understand how this patch solves the latter case. 

- If the OSN is behind a TLS **non terminating** proxy or a network layer device then there is nothing to be done as it can work just fine (as long as you have correct network routing)
- If the OSN is behind a TLS **terminating** proxy then the TLS certificate that the client will see, is different than the TLS certificate that the remote OSN actually has, and then it will be aborted in the Fabric side **after** the TLS handshake ends. 

Did you check that this patch really works? Can you provide an integration test that uses a different TLS server certificate for an OSN than appears in the config and show that it really works? 

Lastly - In general I'm against the ability to run Fabric in an insecure fashion, and I think we need to first be convinced that this is indeed a needed and sought after change.

sampleconfig/orderer.yaml
```diff
@@ -47,6 +47,10 @@ General:
         # Consensus messages are dropped if the buffer is full, and transaction
         # messages are waiting for space to be freed.
         SendBufferSize: 10
+        # InsecureSkipVerify will disable the checking on tls certificate chain and the hostname.
+        # Be careful to change to true, and only use this when doing test or the OSNs are behind
+        # network middle devices.
+        InsecureSkipVerify: false
```
yacovm, line 53: Even if we decide to support this, we shouldn't expose this anyway. We can just support it in the code but not expose this in the config sample.

core/comm/client.go
yacovm: We can use a similar approach to the address overrides we have in the peer:

```
        addressOverrides:
        #  - from:
        #    to:
        #    caCertsFile:
        #  - from:
        #    to:
        #    caCertsFile:
```

But I'm still struggling to understand why this should be solved programmatically and not by issuing correct certificates or making the DNS environment fit what is in the certificates.

yeasy: @yacovm As i mentioned, the issue to resolve here is that will mismatched certificate. If the certificate is matching, then no need to enable this option.

yeasy:
> > In cases that OSN stays behind firewall/proxy/NATed devices, the published endpoint might not match that in the TLS cert. And the user may not care about it because there's following validation logics to compare the tls cert with those inside the channel config.
> 
> There are 2 different use cases here you're trying to solve:
> 
> * The remote OSN publishes an endpoint **foo** but actually the certificate is valid for **bar**: in such a case, I think you should just make the remote OSN to publish an endpoint that matches the certificate, or just have the certificate match the endpoint.
> * The remote OSN is behind a TLS terminating proxy.
> 
> I don't understand how this patch solves the latter case.
> 
> * If the OSN is behind a TLS **non terminating** proxy or a network layer device then there is nothing to be done as it can work just fine (as long as you have correct network routing)
> * If the OSN is behind a TLS **terminating** proxy then the TLS certificate that the client will see, is different than the TLS certificate that the remote OSN actually has, and then it will be aborted in the Fabric side **after** the TLS handshake ends.
> 
> Did you check that this patch really works? Can you provide an integration test that uses a different TLS server certificate for an OSN than appears in the config and show that it really works?
> 
> Lastly - In general I'm against the ability to run Fabric in an insecure fashion, and I think we need to first be convinced that this is indeed a needed and sought after change.

Yes, I have validated in my own testbed. If a test case is needed, there will be one later.

The typical scenario is that the OSN is behind some TLS **non terminating** proxy/firewall, and the other OSN can only see its public addr that is on the proxy/firewall. We cannot guarantee that the OSN always has the correct certificate from the proxy/firewall. In this case, we either use the configuration or hack the source code.

Besides, fabric's code will check the TLS certificate data additionally, hence the OSN is still safe here.

sampleconfig/orderer.yaml
yeasy: Sure, we can consider to hide this config item, as others.
But IMHO, we should allow users to have more flexibility to control the code they are using as OSS, instead of playing hiding and finding?

core/comm/client.go
yacovm: But I am not sure this is the way to solve it. The entity that sets the endpoint and the entity that sets the certificate, is the same entity. When you create a certificate, you get to choose the SAN on it and it should match the endpoint. 

Besides, I'm telling you that we can map the hosts of the endpoint to different hosts locally like the peer does in the example above, instead of just using insecureSkipVerify

sampleconfig/orderer.yaml
yacovm: We already have a bunch of [easter eggs](https://hyperledger-fabric.readthedocs.io/en/release-1.4/raft_configuration.html#local-configuration) in the Raft config, give or take :)

yacovm (changes requested):
> Yes, I have validated in my own testbed. If a test case is needed, there will be one later.

You validated only the non TLS termination case, because I think the TLS termination use case doesn't work with the current code. 

> The typical scenario is that the OSN is behind some TLS non terminating proxy/firewall, and the other OSN can only see its public addr that is on the proxy/firewall. We cannot guarantee that the OSN always has the correct certificate from the proxy/firewall. In this case, we either use the configuration or hack the source code.

If you cannot control the certificate of the OSN then it's probably because it's a commercial CA issued certificate, and in such a case it should be routable from the internet, right? 
If it is - then you just need to put the endpoint in the Raft channel config to match the certificate. 

What am I overlooking here? 

Now, another problem I see with this patch is that it will make code security scanners [issue a security warning](https://github.com/securego/gosec/blob/d8f249a0792bc3b1098b6958f99552e440f520c7/rules/tls.go#L67-L74).  
This would be used as an excuse to say that Fabric's code base is insecure and people would open JIRAs [like they did](https://jira.hyperledger.org/browse/FAB-4249) a few years ago (except the code in the link was used only for unit tests).

yeasy:
> > Yes, I have validated in my own testbed. If a test case is needed, there will be one later.
> 
> You validated only the non TLS termination case, because I think the TLS termination use case doesn't work with the current code.
> 
> > The typical scenario is that the OSN is behind some TLS non terminating proxy/firewall, and the other OSN can only see its public addr that is on the proxy/firewall. We cannot guarantee that the OSN always has the correct certificate from the proxy/firewall. In this case, we either use the configuration or hack the source code.
> 
> If you cannot control the certificate of the OSN then it's probably because it's a commercial CA issued certificate, and in such a case it should be routable from the internet, right?
> If it is - then you just need to put the endpoint in the Raft channel config to match the certificate.
> 
> What am I overlooking here?
> 
> Now, another problem I see with this patch is that it will make code security scanners [issue a security warning](https://github.com/securego/gosec/blob/d8f249a0792bc3b1098b6958f99552e440f520c7/rules/tls.go#L67-L74).
> This would be used as an excuse to say that Fabric's code base is insecure and people would open JIRAs [like they did](https://jira.hyperledger.org/browse/FAB-4249) a few years ago (except the code in the link was used only for unit tests).

We've discussed a lot with details in the jira issue. And I agree that this security scanner issue should be resolved finally.

PR #440, opened 2019-12-20
[FAB-17288] Fix peer lifecycle chaincode package on Windows

lehors:
#### Type of change
Bug fix

#### Description
With this change the command succeeds and creates a tar file
containing forwardslashes like the one created on Linux and Mac.

```
$ tar tf code.tar.gz
src/
src/vendor/
src/vendor/github.com/
src/vendor/github.com/PuerkitoBio/
...
src/vendor/gopkg.in/
src/vendor/gopkg.in/yaml.v2/
src/fabcar.go
src/go.mod
src/go.sum
...
src/vendor/gopkg.in/yaml.v2/yamlprivateh.go
src/vendor/modules.txt

And the chaincode deploys and runs successfully:
$ ./network.sh deployCC
deploying chaincode on channel 'mychannel'

Vendoring Go dependencies ...
~/Projects/Go/src/github.com/hyperledger/fabric-samples/chaincode/fabcar/go ~/Projects
/Go/src/github.com/hyperledger/fabric-samples/test-network
~/Projects/Go/src/github.com/hyperledger/fabric-samples/test-network
Finished vendoring Go dependencies
++ peer lifecycle chaincode package fabcar.tar.gz --path ../chaincode/fabcar/go/ --lang
golang --label fabcar_1
++ res=0
++ set +x

===================== Chaincode is packaged on peer0.org1 =====================

Installing chaincode on peer0.org1...
++ peer lifecycle chaincode install fabcar.tar.gz
++ res=0
++ set +x
2019-12-19 11:11:05.115 CET [cli.lifecycle.chaincode] submitInstallProposal -> INFO 001
Installed remotely: response:<status:200 payload:"\nIfabcar_1:b9a210d6c0c37bd69049e51b1
9de4e78a2e58f01cad4ea446000a53a5440b115\022\010fabcar_1" >
2019-12-19 11:11:05.115 CET [cli.lifecycle.chaincode] submitInstallProposal -> INFO 002
Chaincode code package identifier: fabcar_1:b9a210d6c0c37bd69049e51b19de4e78a2e58f01cad
4ea446000a53a5440b115
...
===================== Invoke transaction successful on peer0.org1 peer0.org2 on channel
'mychannel' =====================

Querying chaincode on peer0.org1...
===================== Querying on peer0.org1 on channel 'mychannel'... =====================

Attempting to Query peer0.org1 ...1576750333 secs
++ peer chaincode query -C mychannel -n fabcar -c '{"Args":["queryAllCars"]}'
++ res=0
++ set +x

[{"Key":"CAR0","Record":{"make":"Toyota","model":"Prius","colour":"blue","owner":"Tomok
o"}},{"Key":"CAR1","Record":{"make":"Ford","model":"Mustang","colour":"red","owner":"Br
ad"}},{"Key":"CAR2","Record":{"make":"Hyundai","model":"Tucson","colour":"green","owner
":"Jin Soo"}},{"Key":"CAR3","Record":{"make":"Volkswagen","model":"Passat","colour":"ye
llow","owner":"Max"}},{"Key":"CAR4","Record":{"make":"Tesla","model":"S","colour":"blac
k","owner":"Adriana"}},{"Key":"CAR5","Record":{"make":"Peugeot","model":"205","colour":
"purple","owner":"Michel"}},{"Key":"CAR6","Record":{"make":"Chery","model":"S22L","colo
ur":"white","owner":"Aarav"}},{"Key":"CAR7","Record":{"make":"Fiat","model":"Punto","co
lour":"violet","owner":"Pari"}},{"Key":"CAR8","Record":{"make":"Tata","model":"Nano","c
olour":"indigo","owner":"Valeria"}},{"Key":"CAR9","Record":{"make":"Holden","model":"Ba
rina","colour":"brown","owner":"Shotaro"}}]
 ===================== Query successful on peer0.org1 on channel 'mychannel' =====================
```

In addition go test now also succeeds.

#### Additional details

Using the filepath package in this case leads to the wrong result
because we do not want the result to be platform dependent. Care
must therefore be taken to ensure the output uses forward slashes.

Signed-off-by: Arnaud J Le Hors <lehors@us.ibm.com>

### DIFF
--- a/core/chaincode/platforms/golang/platform.go
+++ b/core/chaincode/platforms/golang/platform.go
@@ -38,7 +38,7 @@ func (p *Platform) Name() string {
 }
 
 // ValidatePath is used to ensure that path provided points to something that
-// looks like go chainccode.
+// looks like go chaincode.
 //
 // NOTE: this is only used at the _client_ side by the peer CLI.
 func (p *Platform) ValidatePath(rawPath string) error {
@@ -137,8 +137,8 @@ func (p *Platform) GetDeploymentPayload(codepath string) ([]byte, error) {
 	for _, pkg := range dependencyPackageInfo {
 		for _, filename := range pkg.Files() {
 			sd := SourceDescriptor{
-				Name: path.Join("src", pkg.ImportPath, filename),
-				Path: filepath.Join(pkg.Dir, filename),
+				Name: filepath.ToSlash(filepath.Join("src", pkg.ImportPath, filename)),
+				Path: path.Join(pkg.Dir, filename),
 			}
 			fileMap[sd.Name] = sd
 		}
@@ -272,7 +272,7 @@ func DescribeCode(path string) (*CodeDescriptor, error) {
 
 		return &CodeDescriptor{
 			Module:       true,
-			MetadataRoot: filepath.Join(modInfo.Dir, relImport, "META-INF"),
+			MetadataRoot: filepath.ToSlash(filepath.Join(modInfo.Dir, relImport, "META-INF")),
 			Path:         modInfo.ImportPath,
 			Source:       modInfo.Dir,
 		}, nil
@@ -290,7 +290,7 @@ func describeGopath(importPath string) (*CodeDescriptor, error) {
 
 	return &CodeDescriptor{
 		Path:         importPath,
-		MetadataRoot: filepath.Join(sourcePath, "META-INF"),
+		MetadataRoot: path.Join(sourcePath, "META-INF"),
 		Source:       sourcePath,
 	}, nil
 }
@@ -370,10 +370,10 @@ func (s SourceMap) Sources() Sources {
 func (s SourceMap) Directories() []string {
 	dirMap := map[string]bool{}
 	for filename := range s {
-		dir := filepath.Dir(filename)
+		dir := filepath.ToSlash(filepath.Dir(filename))
 		for dir != "." && !dirMap[dir] {
 			dirMap[dir] = true
-			dir = filepath.Dir(dir)
+			dir = filepath.ToSlash(filepath.Dir(dir))
 		}
 	}
 
@@ -431,7 +431,7 @@ func findSource(cd *CodeDescriptor) (SourceMap, error) {
 			if strings.HasPrefix(info.Name(), ".") {
 				return nil
 			}
-			name = filepath.Join("META-INF", name)
+			name = filepath.ToSlash(filepath.Join("META-INF", name))
 			err := validateMetadata(name, path)
 			if err != nil {
 				return err
@@ -441,7 +441,8 @@ func findSource(cd *CodeDescriptor) (SourceMap, error) {
 		default:
 			name = filepath.Join("src", cd.Path, name)
 		}
-
+		name = filepath.ToSlash(name)
+		path = filepath.ToSlash(path)
 		sources[name] = SourceDescriptor{Name: name, Path: path}
 		return nil
 	}
--- a/core/chaincode/platforms/golang/list.go
+++ b/core/chaincode/platforms/golang/list.go
@@ -7,12 +7,14 @@ SPDX-License-Identifier: Apache-2.0
 package golang
 
 import (
+	"bytes"
 	"context"
 	"encoding/json"
 	"fmt"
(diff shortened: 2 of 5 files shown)


core/chaincode/platforms/golang/platform.go
```diff
@@ -272,7 +272,7 @@ func DescribeCode(path string) (*CodeDescriptor, error) {
```
sykesm, line 272: Above, on line 166, is where I'd expect name munging to occur. The 'name' element is what is used in the name and the file path is used to stat the data.

sykesm (changes requested):
Thanks for finding this problem and making a go at a fix.

Looking at the changes, I don't fully grok why they're spread as far as they are, why we need to have checks for any path separators in the code, or why we'd want to use filename management functions over opaque JSON payloads. (The last two, in particular, would prevent me from merging.)

Can you help me understand why a more targeted change where the tar payload is built is not appropriate? For example, using `filepath.ToSlash` where the directory headers are added in `GetDeploymentPayload` (~ line 155) and on the call to `util.WriteFileToPackage` (~ line 166) isn't sufficient?

Basically, most (maybe not all, however) of the places where `filepath` has been used it's because we're intending to deal with real platform specific file names.

Thanks.

sykesm:
I reviewed the JIRA item and I can see why list has to change (slightly) but still worry a bit about path munging. I think we can get rid of the custom list format and use the `-json` option to get properly quoted and escaped paths; with a correct data model, we can fix the path munging.

sykesm:
@lehors can you please take a look at the changes in my branch and see if they address your issues?

https://github.com/hyperledger/fabric/compare/master...sykesm:windoze

lehors:
> I reviewed the JIRA item and I can see why list has to change (slightly) but still worry a bit about path munging. I think we can get rid of the custom list format and use the `-json` option to get properly quoted and escaped paths; with a correct data model, we can fix the path munging.

Indeed, that might work. Fixing GetDeploymentPayload isn't enough because the code breaks before even getting into it when it tries to process the outpout of go list -deps. The backslahes don't jazz well with JSON marshaling.

lehors:
> @lehors can you please take a look at the changes in my branch and see if they address your issues?
> 
> [master...sykesm:windoze](https://github.com/hyperledger/fabric/compare/master...sykesm:windoze)

Almost. It's better in that it generates a tarfile that works/can be successfully deployed but looking into it I see the tarfile still contains some spurious entries:

```
$ tar tf code.tar.gz 
src/
src\\vendor/
src\\vendor\\github.com/
src\\vendor\\github.com\\PuerkitoBio/
src\\vendor\\github.com\\PuerkitoBio\\purell/
src\\vendor\\github.com\\PuerkitoBio\\urlesc/
...
src\\vendor\\google.golang.org\\grpc\\stats/
src\\vendor\\google.golang.org\\grpc\\status/
src\\vendor\\google.golang.org\\grpc\\tap/
src\\vendor\\gopkg.in/
src\\vendor\\gopkg.in\\yaml.v2/
src/fabcar.go
src/go.mod
src/go.sum
src/vendor/github.com/PuerkitoBio/purell/.gitignore
src/vendor/github.com/PuerkitoBio/purell/.travis.yml
src/vendor/github.com/PuerkitoBio/purell/LICENSE
...
src/vendor/gopkg.in/yaml.v2/yamlh.go
src/vendor/gopkg.in/yaml.v2/yamlprivateh.go
src/vendor/modules.txt
```

If my memory serves me well, there is some filtering code in utils/writer.go that assumes paths contain forward slashes and fails when they contain backslashes.

sykesm:
Looks like it's the directories that are a bit messed. Pushed an update and I think I took care of it. (Only go does "the right thing" ™ and adds directory entries to the tar file so I just needed to stop treating the entry names as proper file names.)

lehors:
Yes, that did it. LGTM now.

sykesm:
@lehors, do you agree this is superseded by https://github.com/hyperledger/fabric/pull/458? If so, are you okay with closing this?

Thanks.

lehors:
Happy to close this in favor of #458 which is better. It is worth noting that it is better but not smaller though. :-P

PR #442, opened 2019-12-23
update docs to include methods to fix missing wal

guoger:

#### Type of change

- Documentation update

#### Related issues

https://jira.hyperledger.org/browse/FAB-17283

### DIFF
--- a/docs/source/raft_configuration.md
+++ b/docs/source/raft_configuration.md
@@ -320,7 +320,7 @@ monitor:
 
 ## Troubleshooting
 
-* The more stress you put on your nodes, the more you might have to change certain
+1. The more stress you put on your nodes, the more you might have to change certain
 parameters. As with any system, computer or mechanical, stress can lead to a drag
 in performance. As we noted in the conceptual documentation, leader elections in
 Raft are triggered when follower nodes do not receive either a "heartbeat"
@@ -329,6 +329,35 @@ amount of time. Because Raft nodes share the same communication layer across
 channels (this does not mean they share data --- they do not!), if a Raft node is
 part of the consenter set in many channels, you might want to lengthen the amount
 of time it takes to trigger an election to avoid inadvertent leader elections.
+
+2. As aforementioned, write-ahead logs (WAL) and snapshots should be persisted in
+order to attain safety property of Raft protocol. However, it may happen if a node
+loses its data by accident, and the node would panic with following message:
+```
+tocommit(X) is out of range [lastIndex(Y)]. Was the raft log corrupted, truncated, or lost?
+```
+Manual operation would be required to resurrect problematic node in this case.
+* Safe recovery:  
+The only way to handle the corruption or loss of a WAL is to remove the corrupted
+consenter from the quorum, and add a new one in its place, which can be done following
+the instructions in the reconfiguration section in this doc. This is the most appropriate
+option, as it does not violate safety property of Raft protocol. However, this is
+an expensive operation and affects all channels this node participates in.
+* Unsafe recovery:  
+It is possible to recover from a WAL corruption without removing the consenter from the
+consenter set and re-adding it. However, these procedures violate the Raft protocol safety
+guarantees and, in certain unlikely cases, could result in a blockchain fork. To minimize
+these possibilities, ensure that your cluster has quorum on all channels for a few minutes
+before proceeding.
+  * Option 1: delete WAL and snapshot directories of problematic channel from the corrupted
+node, restart current leader node, then start corrupted orderer. This essentially treats
+corrupted orderer as a fresh node without any data. This is the simplest option that requires
+restarting leader node, which often leads to system downtime.
+  * Option 2: to avoid downtime caused by restarting leader node, you could also delete the
+WAL and snapshot directories of problematic channel from the corrupted node, and then copy
+a snapshot from an alive node with an index greater than `X` in above panic message. The index
+can be inferred from snapshot file names, for example `0000000000000002-0000000000000006.snap`
+stores a snapshot with index 6. Then, start corrupted node and wait for it to catch up.
 
 <!--- Licensed under Creative Commons Attribution 4.0 International License
 https://creativecommons.org/licenses/by/4.0/) -->


docs/source/raft_configuration.md
```diff
@@ -330,5 +330,26 @@ channels (this does not mean they share data --- they do not!), if a Raft node i
...
+- Option 2: wipe out wal and snapshot directories of problematic channel, restart
+current leader node, then restart corrupted orderer. This essentially treats corrupted
+orderer as a fresh node without any data.
+- Option 3: to avoid downtime caused by restarting leader node, you could also wipe out
+wal and snapshot directories of problematic channel, and copy a snapshot from an alive node
+with greater index than X in above panic message. The index can be inferred from snapshot
+file names, for example `0000000000000002-0000000000000006.snap` stores a snapshot with
+index 6. Then start node and wait for it to catch up.
+
+Note: option 2&3 do violate safety property of Raft protocol and should be used with caution.
```
C0rWin: and this makes me wonder why would be enlist 2 and 3 if both violates safety? is there a practical use case where users will choose 2 or 3 over the first one?

C0rWin (changes requested):

docs/source/raft_configuration.md
```diff
@@ -330,5 +330,26 @@ channels (this does not mean they share data --- they do not!), if a Raft node i
...
+order to attain safety proerty of Raft protocol. However, it may happen if a node
+loses its data by accident, and the node would panic with following message:
+```
+tocommit(X) is out of range [lastIndex(Y)]. Was the raft log corrupted, truncated, or lost?
+```
+Human intervention is required to resurrect problematic node in this case:
+- Option 1: remove node and readd it to consenter set via config update, following
+instructions in reconfiguration section in this doc. This is the most appropriate
+option and does not violate safety property of Raft protocol.
+- Option 2: wipe out wal and snapshot directories of problematic channel, restart
```
yacovm: of problematic channel **in the problematic node**.

docs/source/raft_configuration.md
```diff
@@ -330,5 +330,26 @@ channels (this does not mean they share data --- they do not!), if a Raft node i
...
 of time it takes to trigger an election to avoid inadvertent leader elections.
 
+2. As aforementioned, write-ahead logs (wal) and snapshots should be persisted in
+order to attain safety proerty of Raft protocol. However, it may happen if a node
+loses its data by accident, and the node would panic with following message:
+```
+tocommit(X) is out of range [lastIndex(Y)]. Was the raft log corrupted, truncated, or lost?
+```
+Human intervention is required to resurrect problematic node in this case:
+- Option 1: remove node and readd it to consenter set via config update, following
```
yacovm: re-add

docs/source/raft_configuration.md
```diff
@@ -330,5 +330,26 @@ channels (this does not mean they share data --- they do not!), if a Raft node i
...
+instructions in reconfiguration section in this doc. This is the most appropriate
+option and does not violate safety property of Raft protocol.
+- Option 2: wipe out wal and snapshot directories of problematic channel, restart
+current leader node, then restart corrupted orderer. This essentially treats corrupted
```
yacovm: i think we first need to restart the corrupted orderer, no? What if the leader is restarted and then the node says its index is high because it still hasn't been restarted?

docs/source/raft_configuration.md
```diff
@@ -330,5 +330,26 @@ channels (this does not mean they share data --- they do not!), if a Raft node i
...
+orderer as a fresh node without any data.
+- Option 3: to avoid downtime caused by restarting leader node, you could also wipe out
+wal and snapshot directories of problematic channel, and copy a snapshot from an alive node
```
yacovm: of problematic channel **from the corrupted node**

yacovm (commented):
Proposed some additions in the spirit of: "This is a floor polish, not edible"

docs/source/raft_configuration.md
guoger: the first one is fairly heavy as it affects all the channels that this node participates in. Added some text to document this reason.

docs/source/raft_configuration.md
guoger: When this problem occurs, the corrupted node should already be down due to panic, therefore not very important to emphasize on the order.

guoger:
> Proposed some additions in the spirit of: "This is a floor polish, not edible"

This is a phrase i'm unfamiliar with :P

docs/source/raft_configuration.md
C0rWin: but @guoger don't you agree that option 3 could potentially cause more damage?

guoger: @C0rWin I think the potential damage they have is the same - chain being forked in extreme cases. However the advantage that option 3 has over option 2 is reduced downtime, especially for 3 node clusters. This is the main reason i think it's worth to document option 3. wdyt?

docs/source/raft_configuration.md
```diff
@@ -330,5 +330,29 @@ channels (this does not mean they share data --- they do not!), if a Raft node i
 part of the consenter set in many channels, you might want to lengthen the amount
 of time it takes to trigger an election to avoid inadvertent leader elections.
 
+2. As aforementioned, write-ahead logs (wal) and snapshots should be persisted in
+order to attain safety property of Raft protocol. However, it may happen if a node
+loses its data by accident, and the node would panic with following message:
+```
+tocommit(X) is out of range [lastIndex(Y)]. Was the raft log corrupted, truncated, or lost?
+```
+Manual operation would be required to resurrect problematic node in this case.
```
jyellick, line 339: I would break this section into two perhaps?

Safe recovery:
The only way to handle the corruption or loss of a WAL is to remove the corrupted consenter from the quorum, and add a new one in its place.  ...

Unsafe recovery: 

It is possible to recover from a WAL corruption without removing the consenter from the consenter set and re-adding it.  However, these procedures violate the Raft protocol safety guarantees and, in certain unlikely cases, could result in a blockchain fork.  To minimize these possibilities, ensure that your cluster has quorum on all channels for a few minutes before proceeding. (I think this is good enough? We just need a quorum to advance the index via heartbeats past any possible conflicting index in the broken node's WAL, right?)

Option 1: ...

Option 2: ...

guoger:
@C0rWin i've replied to your change request. pls take a look, thx

docs/source/raft_configuration.md
```diff
@@ -330,5 +330,34 @@ channels (this does not mean they share data --- they do not!), if a Raft node i
 part of the consenter set in many channels, you might want to lengthen the amount
 of time it takes to trigger an election to avoid inadvertent leader elections.
 
+2. As aforementioned, write-ahead logs (wal) and snapshots should be persisted in
+order to attain safety property of Raft protocol. However, it may happen if a node
```
pamandrejko: Is the abbreviation wal or WAL? It's used both ways in this section.

docs/source/raft_configuration.md
```diff
@@ -330,5 +330,34 @@ channels (this does not mean they share data --- they do not!), if a Raft node i
...
+guarantees and, in certain unlikely cases, could result in a blockchain fork. To minimize
+these possibilities, ensure that your cluster has quorum on all channels for a few minutes
+before proceeding.
+  * Option 1: wipe out wal and snapshot directories of problematic channel from the corrupted
+node, restart current leader node, then start corrupted orderer. This essentially treats
+corrupted orderer as a fresh node without any data. This is the simplest option that requires
+restarting leader node, which often leads to system downtime.
+  * Option 2: to avoid downtime caused by restarting leader node, you could also wipe out
+wal and snapshot directories of problematic channel from the corrupted node, and copy
+a snapshot from an alive node with greater index than `X` in above panic message. The index
```
pamandrejko: Option 2: To avoid downtime caused by restarting the leader node, you could also delete the
wal and snapshot directories of the problematic channel from the corrupted node, and then copy
a snapshot from an alive node with an index greater than `X` in above panic message.

docs/source/raft_configuration.md
```diff
@@ -330,5 +330,34 @@ channels (this does not mean they share data --- they do not!), if a Raft node i
...
+can be inferred from snapshot file names, for example `0000000000000002-0000000000000006.snap`
+stores a snapshot with index 6. Then, start corrupt node and wait for it to catch up.
+
```
pamandrejko: Should this be Then, restart the corrupted node and wait for it to catch up.

pamandrejko (commented):
A few edits..

docs/source/raft_configuration.md
joealewine: I'm not sure what "safety property" means in this context. Is "safety" an actual thing or a concept (safe recovery)?

docs/source/raft_configuration.md
```diff
@@ -330,5 +330,34 @@ channels (this does not mean they share data --- they do not!), if a Raft node i
...
+order to attain safety property of Raft protocol. However, it may happen if a node
+loses its data by accident, and the node would panic with following message:
+```
+tocommit(X) is out of range [lastIndex(Y)]. Was the raft log corrupted, truncated, or lost?
+```
+Manual operation would be required to resurrect problematic node in this case.
+* Safe recovery:  
+The only way to handle the corruption or loss of a WAL is to remove the corrupted
+consenter from the quorum, and add a new one in its place, which can be done following
+instructions in reconfiguration section in this doc. This is the most appropriate
```
joealewine: following the instructions in the reconfiguration section of this topic

docs/source/raft_configuration.md
```diff
@@ -330,5 +330,34 @@ channels (this does not mean they share data --- they do not!), if a Raft node i
...
+option, as it does not violate safety property of Raft protocol. However, this is heavy
+operation and affects all channels this node participates in.
```
joealewine: Again, not sure what "safety property" means here. Do you mean something like "does not violate the safe operation of a Raft ordering service" or "does not violate the integrity of the ordering service"?

docs/source/raft_configuration.md
joealewine: Why are we documenting an unsafe practice we explicitly don't recommend?

yeasy: Should the snapshot be copied only from the leader node here?
Besides, out of this topic, it might be a good idea to have orderer provided some command to rollback its local ledger to some height and then re-track, like the peer does today.

docs/source/raft_configuration.md
guoger: changed all to be WAL

docs/source/raft_configuration.md
guoger: The unsafe options are significantly simpler and the risk is fairly low. We do not explicitly recommend one over another (and we do have some sentences saying how to lower the probability of chain fork). But we definitely should document the risk, to be complete.

docs/source/raft_configuration.md
guoger: corrupt -> corrupted: done

it should be start instead of restart, cuz the corrupted node actually panics and could not be running

docs/source/raft_configuration.md
guoger: https://en.wikipedia.org/wiki/Safety_property

docs/source/raft_configuration.md
guoger: it can be copied from either leader or other followers. as for orderer ledger rollback, i believe it's addressing a different problem other than corrupted WAL. But since we are here, what use cases do you have in mind?

joealewine: Why wouldn't we explicitly recommend the safe path? 

I still think telling people how to do something we're calling "unsafe" is a bad idea.

guoger: That's definitely a valid concern. IMHO though, explicit "use with caution" is better than "hidden feature", especially when the safe option does introduce tremendous amount of overhead, and the unsafe option is actually not very unsafe. I think this is still a handy solution to most of users.

@jyellick what do you think?

yacovm: > it might be a good idea to have orderer provided some command to rollback its local ledger to some height and then re-track, like the peer does today.


Fabric guarantees block finality. We can't revert blocks, otherwise peers would get forked.

docs/source/raft_configuration.md
```diff
@@ -320,7 +320,7 @@ monitor:
 
 ## Troubleshooting
 
-* The more stress you put on your nodes, the more you might have to change certain
+1. The more stress you put on your nodes, the more you might have to change certain
 parameters. As with any system, computer or mechanical, stress can lead to a drag
 in performance. As we noted in the conceptual documentation, leader elections in
 Raft are triggered when follower nodes do not receive either a "heartbeat"
```
sykesm, line 326: `either a "heartbeat" messages or an "append" message`

I think that should be `a "heartbeat" message or an "append" message`

docs/source/raft_configuration.md
```diff
@@ -330,5 +330,34 @@ channels (this does not mean they share data --- they do not!), if a Raft node i
 part of the consenter set in many channels, you might want to lengthen the amount
 of time it takes to trigger an election to avoid inadvertent leader elections.
 
+2. As aforementioned, write-ahead logs (WAL) and snapshots should be persisted in
+order to attain safety property of Raft protocol. However, it may happen if a node
```
sykesm, line 334:
```suggestion
order to attain safety property of Raft protocol. However, if a node
```

docs/source/raft_configuration.md
```diff
@@ -330,5 +330,34 @@ channels (this does not mean they share data --- they do not!), if a Raft node i
...
+loses its data by accident, and the node would panic with following message:
```
sykesm, line 335:
```suggestion
loses its data by accident, the node would panic with following message:
```

docs/source/raft_configuration.md
```diff
@@ -330,5 +330,34 @@ channels (this does not mean they share data --- they do not!), if a Raft node i
...
+```
+tocommit(X) is out of range [lastIndex(Y)]. Was the raft log corrupted, truncated, or lost?
+```
+Manual operation would be required to resurrect problematic node in this case.
```
sykesm, line 339:
```suggestion
Manual intervention is required to recover the problematic node.
```

docs/source/raft_configuration.md
```diff
@@ -330,5 +330,34 @@ channels (this does not mean they share data --- they do not!), if a Raft node i
...
+* Safe recovery:  
+The only way to handle the corruption or loss of a WAL is to remove the corrupted
+consenter from the quorum, and add a new one in its place, which can be done following
```
sykesm, line 342:
```suggestion
consenter from the quorum and add a new one in its place. This can be accomplished by following
```

docs/source/raft_configuration.md
```diff
@@ -330,5 +330,34 @@ channels (this does not mean they share data --- they do not!), if a Raft node i
...
+the instructions in the reconfiguration section in this doc. This is the most appropriate
```
sykesm, line 343: doclink/ref?

sykesm, line 343:
```suggestion
the instructions in the reconfiguration section in this doc. This is generally the best
```

docs/source/raft_configuration.md
```diff
@@ -330,5 +330,34 @@ channels (this does not mean they share data --- they do not!), if a Raft node i
...
+option, as it does not violate safety property of Raft protocol. However, this is
```
sykesm, line 344:
```suggestion
option as it does not violate safety property of Raft protocol. However, this is
```

docs/source/raft_configuration.md
```diff
@@ -330,5 +330,34 @@ channels (this does not mean they share data --- they do not!), if a Raft node i
...
+an expensive operation and affects all channels this node participates in.
+* Unsafe recovery:  
+It is possible to recover from a WAL corruption without removing the consenter from the
+consenter set and re-adding it. However, these procedures violate the Raft protocol safety
+guarantees and, in certain unlikely cases, could result in a blockchain fork. To minimize
```
sykesm, line 349:
```suggestion
guarantees and could result in a blockchain fork. To minimize
```

I feel it's more appropriate to use language that implies a higher risk.

docs/source/raft_configuration.md
```diff
@@ -330,5 +330,34 @@ channels (this does not mean they share data --- they do not!), if a Raft node i
...
+these possibilities, ensure that your cluster has quorum on all channels for a few minutes
+before proceeding.
+  * Option 1: delete WAL and snapshot directories of problematic channel from the corrupted
```
sykesm, line 352:
```suggestion
  * Option 1: delete WAL and snapshot directories of the problematic channel from the corrupted
```

docs/source/raft_configuration.md
```diff
@@ -330,5 +330,34 @@ channels (this does not mean they share data --- they do not!), if a Raft node i
...
+node, restart current leader node, then start corrupted orderer. This essentially treats
+corrupted orderer as a fresh node without any data. This is the simplest option that requires
+restarting leader node, which often leads to system downtime.
```
sykesm, line 355: How about 

While this is the simplest option, restarting leader node often leads to system downtime.

docs/source/raft_configuration.md
```diff
@@ -330,5 +330,34 @@ channels (this does not mean they share data --- they do not!), if a Raft node i
...
+  * Option 2: to avoid downtime caused by restarting leader node, you could also delete the
```
sykesm, line 356:
```suggestion
  * Option 2: to avoid downtime caused by restarting the leader node, you could also delete the
```

docs/source/raft_configuration.md
```diff
@@ -330,5 +330,34 @@ channels (this does not mean they share data --- they do not!), if a Raft node i
...
+WAL and snapshot directories of problematic channel from the corrupted node, and then copy
```
sykesm, line 357:
```suggestion
WAL and snapshot directories of the problematic channel from the corrupted node, and then copy
```

docs/source/raft_configuration.md
```diff
@@ -330,5 +330,34 @@ channels (this does not mean they share data --- they do not!), if a Raft node i
...
+a snapshot from an alive node with an index greater than `X` in above panic message. The index
```
sykesm, line 358:
```suggestion
a snapshot from a live node with an index greater than `X` in above panic message. The index
```

sykesm (commented):
There are a number of minor edits that would make this a bit more readable.

sykesm:
@guoger, do you plan to come back to this? It’s been a couple of months.

While the updates are needed, if these are stalled and won’t be merged, we may as well close it out.

guoger:
according to review comments, it seemed that this PR causes more confusion rather than clarity, closing for now.

PR #445, opened 2020-01-02
[FAB-17279] Discovery supports collection level endorsement policies

yacovm:
This change set adapts discovery to take into account collection level
endorsement policies.

Since discovery supports multiple endorsement policies in a single invocation
chain already (to support chaincode to chaincode writes), I changed the
method that retrieves a single endorsement policy by chaincode namespace
to retrieve a slice of several endorsement policies as well as extended the
parameter to take into account the collections of the said chaincode.

However, this change set has no effect on production since the discovery backend
needs to change in order to actually fetch the collection specific endorsement policies
if collections are specified.

Change-Id: Id93911c9560a9db8f896a06f71c2ac07e801784a

### DIFF
--- a/discovery/client/client_test.go
+++ b/discovery/client/client_test.go
@@ -326,7 +326,7 @@ func createDiscoveryService(sup *mockSupport) discovery.DiscoveryServer {
 		Id:      []byte{1, 2, 3},
 	})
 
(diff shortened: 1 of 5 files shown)


yacovm:
Wait, don't merge just yet :)

C0rWin (approved):

PR #447, opened 2020-01-02
[FAB-17297] Add curl fallback logic

lindluni:
Add curl as a fallback method in the even wget is not available

### DIFF
--- a/scripts/bootstrap.sh
+++ b/scripts/bootstrap.sh
@@ -67,10 +67,18 @@ download() {
     local BINARY_FILE=$1
     local URL=$2
     echo "===> Downloading: " "${URL}"
-    wget "${URL}" || rc=$?
-    tar xvzf "${BINARY_FILE}" || rc=$?
-    rm "${BINARY_FILE}"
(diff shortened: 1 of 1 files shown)


scripts/bootstrap.sh
```diff
@@ -67,10 +67,16 @@ download() {
...
+    if [[ $(command -v wget) ]]; then
+        wget "${URL}" || rc=$?
+        tar xvzf "${BINARY_FILE}" || rc=$?
+        rm "${BINARY_FILE}"
+    else
+        echo "wget not available, falling back to curl"
+        curl -s -L "${URL}" | tar xz || rc=$?
+    fi
+
+    if [[ -n "$rc" ]]; then
```
denyeart, line 81: Since this is getting complicated again... we would benefit from a script comment explaining why wget is preferred over curl.

mastersingh24:
I really don't see why we need to do this.   Frankly, we need to move away from this whole download script stuff at some point.  But why can't we just prereq wget?

PR #449, opened 2020-01-03
Fabcar uses first-network instead basic-network

chaitanya5:
#### Type of change - 

- Documentation update

#### Description

In the fabric docs,in section Writing your first application- We learn about fabcar which uses first-network but it's mentioned in ccp path that it uses basic-network configuration.json. Intsead it uses first-network.json.

#### Related issues

<!--
Checklist (DELETE AFTER READING):

### DIFF
--- a/docs/source/write_first_app.rst
+++ b/docs/source/write_first_app.rst
@@ -271,11 +271,11 @@ The application connects to the network using a gateway:
 This code creates a new gateway and then uses it to connect the application to
 the network. ``ccp`` describes the network that the gateway will access with the
 identity ``user1`` from ``wallet``. See how the ``ccp`` has been loaded from
-``../../basic-network/connection.json`` and parsed as a JSON file:
+``../../first-network/connection.json`` and parsed as a JSON file:
 
 .. code:: bash
 
-  const ccpPath = path.resolve(__dirname, '..', '..', 'basic-network', 'connection.json');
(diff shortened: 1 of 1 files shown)


chaitanya5 (commented):
Files changed

nikhil550 (approved):

guoger (changes requested):
i don't think this example is using first network (it is mentioned to instruct users to terminate first network, if that's been launched before)

chaitanya5:
> i don't think this example is using first network (it is mentioned to instruct users to terminate first network, if that's been launched before)

Don't wanna sound harsh but fabcar actually uses first-network and not basic-network, if it had used basic-network, then the ccp path that is used in the client fabric-sdk files would point the ccp path to basic-network's connection.js file instead it points to first-network's connection.json file.
I'm providing the links below of invoke.js and query.js files of fabcar which use fabric-sdk module to talk to the chain code, where the ccp path is clearly mentioned pointing to first-network's connection.json file.

https://github.com/hyperledger/fabric-samples/blob/master/fabcar/javascript/invoke.js
https://github.com/hyperledger/fabric-samples/blob/master/fabcar/javascript/query.js

guoger (approved):

guoger:
@chaitanya5 my mistake, fabcar does rely on byfn now

PR #455, opened 2020-01-06
Remove gerrit specific .gitreview

caod123:
#### Type of change
Improvement (improvement to code, performance, etc)

#### Description

This file is no longer needed now that we're off gerrit


ryjones (commented):
LGTM

sykesm (approved):

PR #458, opened 2020-01-07
FAB-17288 - Fix windows path name handling in chaincode packaging

sykesm:
Update chaincode packaging code to appropriately handle path separator differences between Windows and Unix platforms. This is done in a couple of areas:

1. When calculating package dependencies for go, we used a custom template to generate JSON when calling list. This template did not properly encode backslashes used in Windows paths which resulted in decoding errors. This was addressed by removing the custom format and using the `-json` option.
2. There were (maybe still are) several places where path name munging assumes a slash is the path separator. We made changes to differentiate OS paths from tar file entry paths and attempted to use the appropriate separators in each context.
3. A few minor test changes were made to enable our unit tests to run on Windows.

Thanks to @lehors for finding this, making a go at a fix, and, ultimately, testing these changes out.

### DIFF
--- a/core/chaincode/platforms/golang/list.go
+++ b/core/chaincode/platforms/golang/list.go
@@ -21,31 +21,19 @@ import (
 
 const listTimeout = 3 * time.Minute
 
-const packageListFormat = `
-{{- if eq .Goroot false -}}
-{
-    "import_path": "{{ .ImportPath }}",
-    "incomplete": {{ .Incomplete }},
-    "dir": "{{ .Dir }}",
-    "go_files" : [{{ range $i, $file := .GoFiles  }}{{ if $i }}, {{ end }}"{{ $file }}"{{end}}],
-    "c_files":   [{{ range $i, $file := .CFiles   }}{{ if $i }}, {{ end }}"{{ $file }}"{{end}}],
-    "cgo_files": [{{ range $i, $file := .CgoFiles }}{{ if $i }}, {{ end }}"{{ $file }}"{{end}}],
-    "h_files":   [{{ range $i, $file := .HFiles   }}{{ if $i }}, {{ end }}"{{ $file }}"{{end}}],
-    "s_files":   [{{ range $i, $file := .SFiles   }}{{ if $i }}, {{ end }}"{{ $file }}"{{end}}],
-    "ignored_go_files": [{{ range $i, $file := .IgnoredGoFiles }}{{ if $i }}, {{ end }}"{{ $file }}"{{end}}]
-}
-{{- end -}}`
-
+// PackageInfo is the subset of data from `go list -deps -json` that's
+// necessary to calculate chaincode package dependencies.
 type PackageInfo struct {
-	ImportPath     string   `json:"import_path,omitempty"`
-	Dir            string   `json:"dir,omitempty"`
-	GoFiles        []string `json:"go_files,omitempty"`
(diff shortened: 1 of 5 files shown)


lehors:
I tested this change and can confirm it solves the problem.
It is better than my proposed change #440 because instead of trying to force the use of forward slashes throughout (to make the existing code, which assumes forward slashes, work) it fixes the code to work independently of whether forward or black slashes are used. The resulting code is cleaner.

core/chaincode/platforms/util/writer.go
```diff
@@ -13,102 +13,67 @@ import (
...
-	vmLogger.Debugf("rootDirectory = %s", rootDirectory)
+	rootDirectory := filepath.Clean(srcPath)
 
-	//append "/" if necessary
-	updatedExcludeDirs := make([]string, 0)
-	for _, excludeDir := range excludeDirs {
-		if excludeDir != "" && strings.LastIndex(excludeDir, "/") < len(excludeDir)-1 {
-			excludeDir = excludeDir + "/"
-			updatedExcludeDirs = append(updatedExcludeDirs, excludeDir)
-		}
```
guoger, line 47: just making sure i understand this correctly. if an excluded dir does end with "/", it would not get appended to `updatedExcludeDirs`. This is wrong and fixed in this commit, right?

core/chaincode/platforms/util/writer.go
```diff
@@ -13,102 +13,67 @@ import (
...
-	}
+	logger.Debugw("writing folder to package", "rootDirectory", rootDirectory)
+	excludeDirs = append([]string{".git"}, excludeDirs...)
```
guoger: any reason we want ".git" to be the 1st element? (naturally i would do `append(excludeDirs, ".git")` instead)

core/chaincode/platforms/golang/list.go
```diff
@@ -92,6 +80,9 @@ func gopathDependencyPackageInfo(goos, goarch, pkg string) ([]PackageInfo, error
 		if packageInfo.Incomplete {
 			return nil, fmt.Errorf("failed to calculate dependencies: incomplete package: %s", packageInfo.ImportPath)
 		}
+		if packageInfo.Goroot {
```
guoger, line 83: why not `Standard`? (and i wonder how it worked before w/o this check...)

core/chaincode/platforms/util/writer.go
sykesm: fixed.

core/chaincode/platforms/golang/list.go
sykesm: Look at the old template, please. This is doing exactly what we had before - excluding packages that are from the go root - which includes more than “standard”.

core/chaincode/platforms/util/writer.go
sykesm: No. This block added a slash to all dirs (that didn’t already have one) so later checks for prefixes would be on a directory boundary.

Most of the old code wasn’t very clear because it assumed cleaned paths and common prefixes. These changes explicitly clean paths and use names from file info instead of paths when filtering.

guoger (approved):
Thanks for explanation, LGTM

lehors:
Not a big deal but the first commit message should read "Use `go list -deps -json` to build go cc dependencies" rather than "Use `go list -json` to build go cc dependencies".

PR #462, opened 2020-01-08
[FAB-17307] InitRequired param check for ApproveForMyorg

satota2:
This patch adds a check mechanism for the `InitRequired` parameter for the chaincode definition of the current sequence number to `ApproveForMyOrg`.

#### Type of change

- Bug fix

#### Description

ApproveForMyOrg puts a chaincode definition entry for either the currently defined sequence number or the next sequence number.
According to comments on the current implementation, for the definition of the current sequence number (which means it has been committed), this command should reject the requested definition unless the requested one must match exactly the current one.

However, the current ApproveForMyOrg command does not check the InitRequired parameter, so the operators can update the InitRequired parameter.
When an organization does this update, it returns the chaincode to unapproved status for the organization.

To fix the bug, it is necessary to add a check mechanism for the InitRequired parameter to ApproveForMyOrg.

#### Related issues

https://jira.hyperledger.org/browse/FAB-17307


sykesm (approved):
LGTM; would like @jyellick to confirm.

mastersingh24:
> LGTM; would like @jyellick to confirm.
LGTM as well

jyellick (approved):

jyellick:
Thanks for finding and fixing!

PR #463, opened 2020-01-08
we should add the time to wait for the container ready, because diffe…

bjzhang03:
…rent machines has different capacity. The 45s is not enough for majority.

#### Type of change

- Test update

### DIFF
--- a/integration/runner/defaults.go
+++ b/integration/runner/defaults.go
@@ -12,7 +12,7 @@ import (
 	"github.com/hyperledger/fabric/integration/helpers"
 )
 
-const DefaultStartTimeout = 45 * time.Second
+const DefaultStartTimeout = 500 * time.Second
 
 // DefaultNamer is the default naming function.
 var DefaultNamer NameFunc = helpers.UniqueName


bjzhang03:
When I run the unit-test,it always tell me that cannot start the docker couchdb containers,and this confused me.Finally, I found that the reason is that my machine does not have enough capacity to setup the container with in 45s.
My machine has 8 core CPUs,and 16G memory, but it does not work. So I think it need to add the start timeout.

guoger:
do you have a clue why it fails to start within 45s? 8core/16G should be more than enough...

bjzhang03:
> do you have a clue why it fails to start within 45s? 8core/16G should be more than enough...

I have closed all others process,but it also is not work, until I change the timeout.
I have pulled the images ahead,so we can exclude the reason of downing images,and other process factor.
I think that we should larger the timeout, because we can not assume that each one have the computer as well as us.
So, larger the timeout and add the FAQ to instruct this problem can be a better solution.

guoger:
> So, larger the timeout and add the FAQ to instruct this problem can be a better solution.

IMHO, a better solution is to investigate why exactly it takes surprisingly longer to start couchdb container.

sykesm:
Within the integration tests, the *vast* majority of timeouts are no more than one minute. The CouchDB container should be starting in well under 10 seconds. Given our timeouts for running commands, starting, peers, orderers, kafka, zookeeper, etc are all under a minute, I don’t see why couch should be higher.

In unit tests, you can simply point to a running couch DB by setting the `COUCHDB_ADDR` environment variable.

PR #465, opened 2020-01-08
FAB-17315 english colon[:] illegal filename problem on windows when lifecycle chaincode install

KevinXiong2018:
#### Type of change

- Bug fix

#### Description

Fix the problem caused by illegal English colon characters in the file name generated under the directory production\lifecycle\chaincodes on the Windows platform when lifecycle chaincode install
#### Related issues

https://jira.hyperledger.org/browse/FAB-17315

### DIFF
--- a/core/chaincode/persistence/persistence.go
+++ b/core/chaincode/persistence/persistence.go
@@ -230,14 +230,14 @@ func (s *Store) GetChaincodeInstallPath() string {
 }
 
 func packageID(label string, hash []byte) string {
-	return fmt.Sprintf("%s:%x", label, hash)
+	return fmt.Sprintf("%s~%x", label, hash)
 }
 
 func CCFileName(packageID string) string {
 	return packageID + ".tar.gz"
 }
 
-var packageFileMatcher = regexp.MustCompile("^(.+):([0-9abcdef]+)[.]tar[.]gz$")
+var packageFileMatcher = regexp.MustCompile("^(.+)~([0-9abcdef]+)[.]tar[.]gz$")
 
 func installedChaincodeFromFilename(fileName string) (chaincode.InstalledChaincode, bool) {
 	matches := packageFileMatcher.FindStringSubmatch(fileName)
--- a/core/chaincode/persistence/persistence_test.go
+++ b/core/chaincode/persistence/persistence_test.go
@@ -227,11 +227,11 @@ var _ = Describe("Persistence", func() {
 		It("saves a new code package successfully", func() {
 			packageID, err := store.Save("testcc", pkgBytes)
 			Expect(err).NotTo(HaveOccurred())
-			Expect(packageID).To(Equal("testcc:3fec0187440286d404241e871b44725310b11aaf43d100b053eae712fcabc66d"))
+			Expect(packageID).To(Equal("testcc~3fec0187440286d404241e871b44725310b11aaf43d100b053eae712fcabc66d"))
 			Expect(mockReadWriter.WriteFileCallCount()).To(Equal(1))
 			pkgDataFilePath, pkgDataFileName, pkgData := mockReadWriter.WriteFileArgsForCall(0)
 			Expect(pkgDataFilePath).To(Equal(""))
-			Expect(pkgDataFileName).To(Equal("testcc:3fec0187440286d404241e871b44725310b11aaf43d100b053eae712fcabc66d.tar.gz"))
+			Expect(pkgDataFileName).To(Equal("testcc~3fec0187440286d404241e871b44725310b11aaf43d100b053eae712fcabc66d.tar.gz"))
 			Expect(pkgData).To(Equal([]byte("testpkg")))
 		})
 
@@ -243,7 +243,7 @@ var _ = Describe("Persistence", func() {
 			It("does nothing and returns the packageID", func() {
 				packageID, err := store.Save("testcc", pkgBytes)
 				Expect(err).NotTo(HaveOccurred())
-				Expect(packageID).To(Equal("testcc:3fec0187440286d404241e871b44725310b11aaf43d100b053eae712fcabc66d"))
+				Expect(packageID).To(Equal("testcc~3fec0187440286d404241e871b44725310b11aaf43d100b053eae712fcabc66d"))
 				Expect(mockReadWriter.WriteFileCallCount()).To(Equal(0))
 			})
 		})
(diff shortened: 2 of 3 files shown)


guoger:
never tried running fabric on windows but i believe colon is indeed illegal for filename. we probably want to give a different name when storing package on fs. However, do we need to change pkgid schema as well? @jyellick

lindluni:
We did have this Jira around performing this work, which was eventually decided (maybe we would come back and visit this in the future). If this is going to be undertaken, this is the Jira and the exhaustive list should exhausted

sykesm:
> never tried running fabric on windows

We don't test or support running the peer or orderer on windows, but we do support the CLI in that space. We need to get under if/why the persistence package would be used in the CLI space and, if it is used and used appropriately, address the issue.

If it turns out that the persistence is only the server side (which is my hope), Windows is not a supported host environment for the peer at this time.

We'll dig in a bit....

guoger:
> > never tried running fabric on windows
> 
> We don't test or support running the peer or orderer on windows

Is this written somewhere? (glad to see we don't need to support server side actually...)

> 
> If it turns out that the persistence is only the server side (which is my hope), Windows is not a supported host environment for the peer at this time.

IIRC, that persistency **is** only on server side.

KevinXiong2018:
Yes, we do not plan to support running fabric services under windows or as a chain code development environment now. There are a few things that can be determined:
1. On Linux or Mac systems, use an underscore _ instead of a colon: as the delimiter of package.tar.gz, all are valid characters, and have no effect on the fabric server currently running on Linux. 
2. The English colon is illegal in the file name under windows. We will lost the compatibility of running fabric on windows later.
3. Through the simple modification above, I have successfully run fabric 2.0 servers on my windows 10.

sykesm:
> 3. Through the simple modification above, I have successfully run fabric 2.0 servers on my windows 10.

I understand that the colon is the drive separator character on windows and that it's a problem for that environment.  This does not change the fact that we do not test or support running the peer on Windows.

Additionally, the underscore is a valid character in chaincode ID labels under the new lifecycle. The changes you have proposed break the existing code when labels contain underscores.

So there are two concerns:
1. The code changes that are proposed are incomplete and break the logic associated with `packageFileMatcher`.
2. Even if this change is incorporated, there is no guarantee that additional changes won't be necessary and, as with this change, we lack the resources or infrastructure to test them completely at this time.

denyeart:
Does anybody know where the valid characters for label are defined? It seems that colons within the label actually works (except in Windows). At least, nothing validates that colon shouldn't be included, my trial on Mac worked. Is the label:hash actually parsed somewhere? @jyellick @ale-linux 

In general I agree with Matt... peers are not tested on Windows, but we should least attempt to avoid any new breaks in v2.0 lifecycle.

sykesm:
regexp for the label was fixed and tested here: https://github.com/hyperledger/fabric/pull/480

guoger:
I understand that we are not attempting to support/test windows platform but I do empathize the need to setup dev env on it, and i think this is not really a drastic change asking for a compromise among platforms.

Also, wouldn't it be easier for an admin to pull down the pkg file and inspect, if it does not contain colon? Furthermore, if one day (i wish that day would not come though) we need to support the platform, this would not be another breaking change.

Why was colon chosen as separator initially and why cannot it be changed prior to release?

jyellick:
I initially typed up something a bit longer, but the short history of "why is there a colon in the name", would be because it was the shortest path to get things working, that we could always come back and mangle the package-id before writing it to the filesystem if need be.  This bug / PR seems to give us a reason to reassess.

I think we can live with a delimiter which is also in the list of allowable label characters.  Since we know the format is always `<label><delimiter><fixed-length-hex>.tar.gz` it's certainly possible to create safe parsing of labels no matter what delimiter we pick.  

Even on UNIX, `:` is a bit of an annoying character to have in filenames, and since we seem to have users running up against real problems, I'm certainly amenable to seeing it changed.

In the old lifecycle, it appears that the the `:` was substituted for a `.`.  Since no one seemed to complain about that, it would be somewhat natural choice to me, though I'm open to other suggestions.

KevinXiong2018:
> > 1. Through the simple modification above, I have successfully run fabric 2.0 servers on my windows 10.
> 
> I understand that the colon is the drive separator character on windows and that it's a problem for that environment. This does not change the fact that we do not test or support running the peer on Windows.
> 
> Additionally, the underscore is a valid character in chaincode ID labels under the new lifecycle. The changes you have proposed break the existing code when labels contain underscores.
> 
> So there are two concerns:
> 
> 1. The code changes that are proposed are incomplete and break the logic associated with `packageFileMatcher`.
> 2. Even if this change is incorporated, there is no guarantee that additional changes won't be necessary and, as with this change, we lack the resources or infrastructure to test them completely at this time.

Yes, without breaking the current LabelRegexp expression (`^[[:alnum:]][[:alnum:]_.+-]*$`) , Maybe we just need to find a common character that is not in [0~9][a~z][A~Z][ _.+-], acts as file name delimiter for packageID, and it is a valid file name character for linux, Mac, windows, etc.

I chose the English wave symbol ~ as this separator, and the code has been updated.

We know that the English colon is an illegal file name character for Windows. If we can fix this problem before the Fabric 2.0 release, when we plan to support Windows in the future, we do not need to modify the name of the existing packageID file containing the colon on Linux and Mac servers.

mastersingh24:
>I think we can live with a delimiter which is also in the list of allowable label characters. Since we know the format is always <label><delimiter><fixed-length-hex>.tar.gz it's certainly possible to create safe parsing of labels no matter what delimiter we pick.


Why do we actually need a delimiter if we have <fixed-length-hex>?  Seems we can always just split <label><fixed-length-hex> based on the fact we know the length of <fixed-length-hex>.

sykesm:
> Why do we actually need a delimiter if we have ? Seems we can always just split based on the fact we know the length of .

We can split based on the tail length and any character can be used as the delimiter that is not "special."  We all agree this can be done. We can also encode the names as base32 if we choose; there are many ways to do this.

I want to reiterate my two points:
  1. These changes are incomplete.
  2. There's no guarantee we won't hit other issues.

If we want these changes, we really need to address the completeness aspect here.

sykesm:
> 1. These changes are incomplete.

The use of `~` looks to address the "completeness" concern.

sykesm (changes requested):
@jyellick if you're okay with `~`, this works.

@KevinXiong2018, please squash your commits and rebase on current master.

mastersingh24:
I'm ok with `~`

jyellick:
I would prefer that we move to a more standard delimiter, like `.`, and use suffix matching, rather than regexp splitting to separate the label.  That being said, the move to `~` is still an improvement/fix for the current issue, so I'm fine merging as is if others would prefer.

KevinXiong2018:
> your commits

My commits have been squashed successfully.

sykesm:
@jyellick I'll handle suffix matching on top.

sykesm (approved):

sykesm (approved):

PR #471, opened 2020-01-08
Small edits to upgrade docs

joealewine:
Make sure people have the right commands in the
right places. Fix docker run command that was
meant to enable an environment variables file

Change-Id: Ica0317e86ef4854f47bd3b7c8a5de4347e1b948d

#### Type of change

- Documentation update

#### Description

Add to previous change by eliminating an incorrect docker run command (this wasn't meant to start the peer but enable a file with environment variables. 

Also add context to upgrade doc about upgrade-dbs command and the relevant environment variables.


joealewine:
@suryalnvs 

@sykesm 

My edits to yesterday's change. That command before the environment variables wasn't supposed to start the peer --- it was just for environment variables. So I went and fixed that.

sykesm (approved):

PR #469, opened 2020-01-08
[FAB-17299] Support collection level endorsement policies for discovery

caod123:
#### Type of change

- New feature

#### Description
* PoliciesByChaincode now returns one InquireablePolicy per collection,
either the collection's endorsement policy if one exists or the
chaincode policy by default
* chaincode.Metadata now stores CollectionPolicies for _lifecycle
namespace chaincodes that have collection endorsement policies

#### Related issues

[FAB-17299](https://jira.hyperledger.org/browse/FAB-17299)
This PR cannot be merged until [FAB-17308](https://jira.hyperledger.org/browse/FAB-17308) is complete; however, submitting PR now for review/feedback.

/cc @yacovm @denyeart

PR #472, opened 2020-01-08
[FAB-17299] Support collection level endorsement policies for discovery

caod123:
#### Type of change

- New feature

#### Description
* PoliciesByChaincode now returns one InquireablePolicy per collection,
either the collection's endorsement policy if one exists or the
chaincode policy by default
* chaincode.Metadata now stores CollectionPolicies for _lifecycle
namespace chaincodes that have collection endorsement policies

#### Related issues

[FAB-17299](https://jira.hyperledger.org/browse/FAB-17299)
This PR cannot be merged until [FAB-17308](https://jira.hyperledger.org/browse/FAB-17308) is complete; however, submitting PR now for review/feedback.

/cc @yacovm @denyeart

### DIFF
--- a/core/chaincode/lifecycle/metadata_provider.go
+++ b/core/chaincode/lifecycle/metadata_provider.go
@@ -13,7 +13,6 @@ import (
 	"github.com/hyperledger/fabric-protos-go/peer"
 	"github.com/hyperledger/fabric/common/chaincode"
 	"github.com/hyperledger/fabric/common/policies"
-	"github.com/hyperledger/fabric/common/policydsl"
 	"github.com/hyperledger/fabric/protoutil"
 	"github.com/pkg/errors"
 )
@@ -21,7 +20,7 @@ import (
 // LegacyMetadataProvider provides metadata for a lscc-defined chaincode
 // on a specific channel.
 type LegacyMetadataProvider interface {
-	Metadata(channel string, cc string, includeCollections bool) *chaincode.Metadata
+	Metadata(channel string, cc string, collections ...string) *chaincode.Metadata
 }
 
 // ChaincodeInfoProvider provides metadata for a _lifecycle-defined
@@ -105,28 +104,54 @@ func (mp *MetadataProvider) toSignaturePolicyEnvelope(channelID string, policyBy
 }
 
 // Metadata implements the metadata retriever support interface for service discovery
-func (mp *MetadataProvider) Metadata(channel string, ccName string, includeCollections bool) *chaincode.Metadata {
+func (mp *MetadataProvider) Metadata(channel string, ccName string, collections ...string) *chaincode.Metadata {
 	ccInfo, err := mp.ChaincodeInfoProvider.ChaincodeInfo(channel, ccName)
 	if err != nil {
 		logger.Debugf("chaincode '%s' on channel '%s' not defined in _lifecycle. requesting metadata from lscc", ccName, channel)
 		// fallback to legacy metadata via cclifecycle
-		return mp.LegacyMetadataProvider.Metadata(channel, ccName, includeCollections)
+		return mp.LegacyMetadataProvider.Metadata(channel, ccName, collections...)
 	}
 
 	spe, err := mp.toSignaturePolicyEnvelope(channel, ccInfo.Definition.ValidationInfo.ValidationParameter)
 	if err != nil {
 		logger.Errorf("could not convert policy for chaincode '%s' on channel '%s', err '%s'", ccName, channel, err)
-		spe = policydsl.MarshaledRejectAllPolicy
+		return nil
 	}
 
 	// report the sequence as the version to service discovery since
 	// the version is no longer required to change when updating any
 	// part of the chaincode definition
 	ccMetadata := &chaincode.Metadata{
-		Name:              ccName,
-		Version:           strconv.FormatInt(ccInfo.Definition.Sequence, 10),
-		Policy:            spe,
-		CollectionsConfig: ccInfo.Definition.Collections,
+		Name:               ccName,
+		Version:            strconv.FormatInt(ccInfo.Definition.Sequence, 10),
+		Policy:             spe,
+		CollectionPolicies: map[string][]byte{},
+		CollectionsConfig:  ccInfo.Definition.Collections,
+	}
+
+	if ccInfo.Definition.Collections == nil {
+		return ccMetadata
+	}
+
+	// process any existing collection endorsement policies
+	for _, collectionName := range collections {
+		for _, conf := range ccInfo.Definition.Collections.Config {
+			staticCollConfig := conf.GetStaticCollectionConfig()
+			if staticCollConfig == nil {
+				continue
+			}
+			if staticCollConfig.Name == collectionName {
+				if staticCollConfig.EndorsementPolicy != nil {
+					ep := protoutil.MarshalOrPanic(staticCollConfig.EndorsementPolicy)
+					cspe, err := mp.toSignaturePolicyEnvelope(channel, ep)
+					if err != nil {
+						logger.Errorf("could not convert collection policy for chaincode '%s' collection '%s' on channel '%s', err '%s'", ccName, collectionName, channel, err)
+						return nil
+					}
+					ccMetadata.CollectionPolicies[collectionName] = cspe
+				}
+			}
+		}
 	}
 
 	return ccMetadata
--- a/discovery/support/chaincode/support.go
+++ b/discovery/support/chaincode/support.go
@@ -18,7 +18,7 @@ import (
 var logger = flogging.MustGetLogger("discovery.DiscoverySupport")
 
 type MetadataRetriever interface {
-	Metadata(channel string, cc string, loadCollections bool) *chaincode.Metadata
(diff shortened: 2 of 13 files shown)


caod123:
Reopened #469 as a draft PR

core/chaincode/lifecycle/metadata_provider.go
```diff
@@ -119,14 +119,36 @@ func (mp *MetadataProvider) Metadata(channel string, ccName string, includeColle
 		spe = cauthdsl.MarshaledRejectAllPolicy
 	}
 
+	// process any existing collection endorsement policies
+	collectionPolicies := map[string][]byte{}
+	if ccInfo.Definition.Collections != nil {
```
yacovm: [Return early, return often](http://www.itamarweiss.com/personal/2018/02/28/return-early-pattern.html).

Can we please simply return the object and skip populating the collection policies if the **Definition.Collections** is nil?

Can we please switch the: if **staticCollConfig.EndorsementPolicy != nil**  to be if if **staticCollConfig.EndorsementPolicy == nil**  and then continue the loop?

discovery/support/chaincode/support.go
```diff
@@ -35,20 +35,58 @@ func NewDiscoverySupport(ci MetadataRetriever) *DiscoverySupport {
...
+
+	// chaincodeData.CollectionPolicies will be nil when using legacy lifecycle (lscc)
+	// because collection endorsement policies are not supported in lscc
+	if chaincodeData.CollectionPolicies == nil || len(collections) == 0 {
+		return []policies.InquireablePolicy{chaincodePolicy}
+	}
+	// process any collection policies
+	for _, collectionName := range collections {
+		// add the collection policy if it exists otherwise default to the chaincode policy
+		if policy, collectionPolicyExists := chaincodeData.CollectionPolicies[collectionName]; collectionPolicyExists {
```
yacovm: Let's do it without the **if**, just read from the map, and if it doesn't exist then continue. Makes less indentation.

discovery/support/chaincode/support.go
```diff
@@ -35,20 +35,58 @@ func NewDiscoverySupport(ci MetadataRetriever) *DiscoverySupport {
...
+			pol := &common2.SignaturePolicyEnvelope{}
+			if err := proto.Unmarshal(policy, pol); err != nil {
+				logger.Warningf("Failed unmarshaling collection policy for chaincode '%s' collection '%s': %s", cc, collectionName, err)
```
yacovm: This logging should be an error IMO

discovery/support/chaincode/support.go
```diff
@@ -35,20 +35,58 @@ func NewDiscoverySupport(ci MetadataRetriever) *DiscoverySupport {
...
+				return nil
+			}
+			if len(pol.Identities) == 0 || pol.Rule == nil {
+				logger.Warningf("Invalid collection policy, either Identities(%v) or Rule(%v) are empty", pol.Identities, pol.Rule)
```
yacovm: This logging should be an error

discovery/support/chaincode/support.go
```diff
@@ -35,20 +35,58 @@ func NewDiscoverySupport(ci MetadataRetriever) *DiscoverySupport {
...
 	}
 	if len(pol.Identities) == 0 || pol.Rule == nil {
-		logger.Warningf("Invalid policy, either Identities(%v) or Rule(%v) are empty:", pol.Identities, pol.Rule)
+		logger.Warningf("Invalid policy, either Identities(%v) or Rule(%v) are empty", pol.Identities, pol.Rule)
 		return nil
 	}
-	return []policies.InquireablePolicy{inquire.NewInquireableSignaturePolicy(pol)}
+	chaincodePolicy := inquire.NewInquireableSignaturePolicy(pol)
+
+	inquireablePolicies := map[policies.InquireablePolicy]bool{}
```
yacovm: 1) You can do a map from **Foo** to **struct{}**, it's the common pattern for a set in Go.
2) How does this even work? **policies.InquireablePolicy** is an interface. How does Golang know to find it again in the map if it already exists? Look at an example I made [here](https://play.golang.org/p/n2PF37yKT6D).

I think that what we really need is to make the keys of the map, the byte representation of the policies and not the derived interfaces.

yacovm (changes requested):
See comments

core/chaincode/lifecycle/metadata_provider.go
caod123: Ack

discovery/support/chaincode/support.go
caod123: Ack, will also change the 2 warnings for the chaincode policy further above

discovery/support/chaincode/support.go
caod123: Ack

discovery/support/chaincode/support.go
caod123: Ack

discovery/support/chaincode/support.go
caod123: I changed it to `make(map[string]struct{})` and set the keys to the string representation of the policy bytes (since byte slices themselves can't be keys)

caod123 (commented):
Pushing a followup commit so it's easier to see suggested changes. Will squash after changes approved

core/chaincode/lifecycle/metadata_provider.go
```diff
@@ -119,37 +119,41 @@ func (mp *MetadataProvider) Metadata(channel string, ccName string, collections
...
+			if staticCollConfig == nil {
+				continue
+			}
+			if staticCollConfig.Name == collectionName {
+				if staticCollConfig.EndorsementPolicy != nil {
+					ep := protoutil.MarshalOrPanic(staticCollConfig.EndorsementPolicy)
+					cspe, err := mp.toSignaturePolicyEnvelope(channel, ep)
+					if err != nil {
+						logger.Errorf("could not convert collection policy for chaincode '%s' collection '%s' on channel '%s', err '%s'", ccName, collectionName, channel, err)
+						cspe = cauthdsl.MarshaledRejectAllPolicy
```
yacovm: `cspe = cauthdsl.MarshaledRejectAllPolicy`

I missed this in the previous review. This wasn't introduced by you, but I don't think this should be here at all. 
If we failed converting this policy, then we should simply return nil and not use this RejectAllPolicy.

It makes sense to use RejectAllPolicy in validation, but this is a user facing API and we should return an error and not something that would make discovery analysis go haywire :-)

yacovm (changes requested):
Looks pretty good, just something that wasn't introduced by you but i think we should fix here and now

core/chaincode/lifecycle/metadata_provider.go
caod123: Sounds good, will do for both the collection policy as well as chaincode policy.

caod123 (commented):
Went ahead and squashed the PR

discovery/support/chaincode/support_test.go
```diff
@@ -9,66 +9,85 @@ package chaincode
...
 			sup := NewDiscoverySupport(&mockMetadataRetriever{res: test.input})
-			res := sup.PoliciesByChaincode("", "")
-			if test.shouldBeNil {
+			var res []policies.InquireablePolicy
+			if test.input == ccmd3 {
+				res = sup.PoliciesByChaincode("", "", "col1")
+			} else {
+				res = sup.PoliciesByChaincode("", "")
+			}
+			if test.expectedReturn == nil {
```
yacovm: can't we just use **assert.Equal** here?

discovery/support/chaincode/support_test.go
```diff
@@ -9,66 +9,85 @@ package chaincode
...
 
 	for _, test := range tests {
 		test := test
 		t.Run(test.name, func(t *testing.T) {
 			sup := NewDiscoverySupport(&mockMetadataRetriever{res: test.input})
-			res := sup.PoliciesByChaincode("", "")
-			if test.shouldBeNil {
+			var res []policies.InquireablePolicy
+			if test.input == ccmd3 {
+				res = sup.PoliciesByChaincode("", "", "col1")
```
yacovm: let's please extract this variable ("col1") to the test parameters and not use  **if test.input == ccmd3** :)

integration/discovery/testdata/collections_config_org1_org2.json
```diff
@@ -5,5 +5,12 @@
         "requiredPeerCount": 1,
         "maxPeerCount": 2,
         "blockToLive":1000000
+    },
+    {
+        "name": "collectionDetails",
```
yacovm: where is this collection used?

discovery/support/chaincode/support_test.go
wenjianqiao: Done.

discovery/support/chaincode/support_test.go
wenjianqiao: Done.

integration/discovery/testdata/collections_config_org1_org2.json
wenjianqiao: Removed since it is not used.

yacovm (approved):

PR #478, opened 2020-01-09
Reduce the number of peers in test networks

sykesm:
Try to slightly reduce the resource requirements of our tests by reducing the number of server processes we launch by default by removing the second peer in each org in our "basic" networks.

### DIFF
--- a/integration/nwo/standard_networks.go
+++ b/integration/nwo/standard_networks.go
@@ -6,6 +6,7 @@ SPDX-License-Identifier: Apache-2.0
 
 package nwo
 
+// BasicSolo is a configuration wtih two organizations and one peer per org.
 func BasicSolo() *Config {
 	return &Config{
 		Organizations: []*Organization{{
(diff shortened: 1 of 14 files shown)


integration/nwo/standard_networks.go
```diff
@@ -86,6 +75,31 @@ func BasicSolo() *Config {
 	}
 }
 
+// FullSolo is a configuration wtih two organizations and two peers per org.
+func FullSolo() *Config {
```
denyeart, line 79: Not a big deal, but I'd prefer more descriptive names rather than BasicSolo() and FullSolo(), e.g. SoloWithTwoOrgsOnePeerEach() and SoloWithTwoOrgsTwoPeersEach(). That way when looking at a test you know exactly what you are dealing with.

sykesm:
Opened [FAB-17343](https://jira.hyperledger.org/browse/FAB-17343) for the orderer integration test flake.

sykesm:
Opened [FAB-17344](https://jira.hyperledger.org/browse/FAB-17344) for the unit test flake.

mastersingh24:
> Try to slightly reduce the resource requirements of our tests by reducing the number of server processes we launch by default by removing the second peer in each org in our "basic" networks.

Excellent!

mastersingh24 (approved):
LGTM - I think we can modify the names in a later change if @denyeart  wants to

PR #476, opened 2020-01-09
Update dead link to Go Chaincode SDK

fasmat:
#### Type of change

- Documentation update

#### Description

The link to the go chaincode SDK was missing and the documentation was pointing to the old location of the shim package.

#### Additional details

N/A

#### Related issues

N/A

PR #479, opened 2020-01-09
Go SDK link should point to 1.4 release

fasmat:
#### Type of change

- Documentation update

#### Description

The links should point to the 1.4 release of the shim package and its documentation as discussed in #476.

#### Additional details

N/A

#### Related issues

#476


sykesm (approved):
LGTM; @pamandrejko

sykesm:
Let me elaborate a bit on the comments I made in #476.

For those that continue to use 1.4, we obviously want the links in the doc to work. We would also prefer for the links to point to versions that correspond to the release we're documenting. This change accomplishes that.

Beyond that, however, we probably need to improve the doc around the actual imports and code for 1.4 chaincode. Those building chaincode for 1.4 need to decide if they want to migrate to fabric-chaincode-go (which is backwards compatible) or vendor their own version of the shim from release-1.4.

PR #482, opened 2020-01-09
[FAB-17339] Refactor couchDB unit test

DereckLuo:

#### Type of change

- Test update

#### Description

- Fixed couchDB stopExternalResource in ./core/ledger/kvledger/txmgmt/privacyenabledstate. Only stop external couchDB container in this function. Move all DB cleanups in cleanup function.
- For package statecouchdb, un-export **testVDBEnv** and attach __couchDBAddress__ and __cleanupCouchDB__ function to the struct. Which during test, a single global instance of the testVDBEnv is being used, and couchDB external containers will only start once for this package.
- Replaced original array based test environments with map based test environments.

#### Related issues

https://jira.hyperledger.org/browse/FAB-17339

/cc @manish-sethi

### DIFF
--- a/core/ledger/kvledger/txmgmt/pvtstatepurgemgmt/purge_mgr_test.go
+++ b/core/ledger/kvledger/txmgmt/pvtstatepurgemgmt/purge_mgr_test.go
@@ -21,11 +21,16 @@ import (
 	"github.com/stretchr/testify/assert"
 )
 
+const (
+	levelDBtestEnvName = "levelDB_TestEnv"
+	couchDBtestEnvName = "couchDB_TestEnv"
+)
+
 // Tests will be run against each environment in this array
 // For example, to skip CouchDB tests, remove &couchDBLockBasedEnv{}
-var testEnvs = []privacyenabledstate.TestEnv{
-	&privacyenabledstate.LevelDBCommonStorageTestEnv{},
-	&privacyenabledstate.CouchDBCommonStorageTestEnv{},
+var testEnvs = map[string]privacyenabledstate.TestEnv{
+	levelDBtestEnvName: &privacyenabledstate.LevelDBCommonStorageTestEnv{},
+	couchDBtestEnvName: &privacyenabledstate.CouchDBCommonStorageTestEnv{},
 }
 
 func TestMain(m *testing.M) {
@@ -106,11 +111,7 @@ func testPurgeMgr(t *testing.T, dbEnv privacyenabledstate.TestEnv) {
 }
 
 func TestPurgeMgrForCommittingPvtDataOfOldBlocks(t *testing.T) {
-	dbEnvs := []privacyenabledstate.TestEnv{
-		&privacyenabledstate.LevelDBCommonStorageTestEnv{},
-		&privacyenabledstate.CouchDBCommonStorageTestEnv{},
-	}
-	for _, dbEnv := range dbEnvs {
+	for _, dbEnv := range testEnvs {
 		t.Run(dbEnv.GetName(), func(t *testing.T) { testPurgeMgrForCommittingPvtDataOfOldBlocks(t, dbEnv) })
 	}
 }
@@ -163,7 +164,7 @@ func testPurgeMgrForCommittingPvtDataOfOldBlocks(t *testing.T, dbEnv privacyenab
 }
 
 func TestKeyUpdateBeforeExpiryBlock(t *testing.T) {
-	dbEnv := &privacyenabledstate.LevelDBCommonStorageTestEnv{}
+	dbEnv := testEnvs[levelDBtestEnvName]
 	ledgerid := "testledger-perge-mgr"
 	btlPolicy := btltestutil.SampleBTLPolicy(
 		map[[2]string]uint64{
@@ -203,7 +204,7 @@ func TestKeyUpdateBeforeExpiryBlock(t *testing.T) {
 }
 
 func TestOnlyHashUpdateInExpiryBlock(t *testing.T) {
-	dbEnv := &privacyenabledstate.LevelDBCommonStorageTestEnv{}
+	dbEnv := testEnvs[levelDBtestEnvName]
 	ledgerid := "testledger-perge-mgr"
 	btlPolicy := btltestutil.SampleBTLPolicy(
 		map[[2]string]uint64{
@@ -250,7 +251,7 @@ func TestOnlyHashUpdateInExpiryBlock(t *testing.T) {
 }
 
 func TestOnlyHashDeleteBeforeExpiryBlock(t *testing.T) {
-	dbEnv := &privacyenabledstate.LevelDBCommonStorageTestEnv{}
+	dbEnv := testEnvs[levelDBtestEnvName]
 	ledgerid := "testledger-perge-mgr"
 	btlPolicy := btltestutil.SampleBTLPolicy(
 		map[[2]string]uint64{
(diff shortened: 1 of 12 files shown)


core/ledger/util/couchdb/couchdb_test.go
```diff
@@ -74,9 +75,10 @@ func TestMain(m *testing.M) {
 
 func testMain(m *testing.M) int {
 	// Switch to CouchDB
-	address, cleanup := couchdbtest.CouchDBSetup("", "")
-	testAddress = address
-	defer cleanup()
+	if testAddress == "" {
+		testAddress, cleanupCouchDB = couchdbtest.CouchDBSetup((nil))
```
manish-sethi, line 84: I thought that you were planning to move couchdb setup outside testmain... like other packages

core/ledger/kvledger/txmgmt/statedb/statecouchdb/statecouchdb_test.go
```diff
@@ -23,19 +23,17 @@ import (
...
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 )
 
 func TestMain(m *testing.M) {
 	flogging.ActivateSpec("statecouchdb=debug")
 
-	address, cleanup := couchdbtest.CouchDBSetup("", "")
-	couchAddress = address
+	startCouchDB()
```
manish-sethi: Same as other comment about TestMain

manish-sethi (commented):
LGTM - a couple minor comments. If you want to fix them separately, I am fine with that as well.

manish-sethi (dismissed):

core/ledger/kvledger/txmgmt/pvtstatepurgemgmt/purge_mgr_test.go
```diff
@@ -163,7 +159,7 @@ func testPurgeMgrForCommittingPvtDataOfOldBlocks(t *testing.T, dbEnv privacyenab
 }
 
 func TestKeyUpdateBeforeExpiryBlock(t *testing.T) {
-	dbEnv := &privacyenabledstate.LevelDBCommonStorageTestEnv{}
+	dbEnv := testEnvs[0]
```
lindluni: This seems a little fragile, what do we gain by pulling it from the array? If we were to support another storage engine in the future, if it was added to the start of testEnvs instead of the end, it would break the expected outcome here.

core/ledger/kvledger/txmgmt/pvtstatepurgemgmt/purge_mgr_test.go
```diff
@@ -203,7 +199,7 @@ func TestKeyUpdateBeforeExpiryBlock(t *testing.T) {
 }
 
 func TestOnlyHashUpdateInExpiryBlock(t *testing.T) {
-	dbEnv := &privacyenabledstate.LevelDBCommonStorageTestEnv{}
+	dbEnv := testEnvs[0]
```
lindluni: Same as above

core/ledger/kvledger/txmgmt/pvtstatepurgemgmt/purge_mgr_test.go
```diff
@@ -250,7 +246,7 @@ func TestOnlyHashUpdateInExpiryBlock(t *testing.T) {
 }
 
 func TestOnlyHashDeleteBeforeExpiryBlock(t *testing.T) {
-	dbEnv := &privacyenabledstate.LevelDBCommonStorageTestEnv{}
+	dbEnv := testEnvs[0]
```
lindluni: Same as above

core/ledger/kvledger/txmgmt/pvtstatepurgemgmt/purge_mgr_test.go
DereckLuo: thanks for the review. very good point, it's definitely very fragile. 
The intent here is that couchDB external container is only being started once for the entire package (doing so by calling _testEnv.Init_ at the beginning of every test), and **TestMain** will loop through _testEnvs_ after all tests finished in order to cleanup couchDB. If we declare new test environments in specific test, **TestMain** will not cleanup couchDB properly. 
Open for any suggestions.

lindluni: Ahh, I see. You could use a map (which you'd still be able to iterate over), but here you could reference by its key, `testEnvs["level"]`

sykesm: Agree; map is preferred.

manish-sethi: Yup, this map based approach is already being used in some other package as well - https://github.com/hyperledger/fabric/blob/d80c5f5a60d59a78891e1c958018a4bfeb892182/core/ledger/kvledger/txmgmt/txmgr/lockbasedtxmgr/pkg_test.go#L65.

DereckLuo: changes applied accordingly. thanks

core/ledger/kvledger/txmgmt/pvtstatepurgemgmt/purge_mgr_test.go
```diff
@@ -28,6 +28,16 @@ var testEnvs = []privacyenabledstate.TestEnv{
 	&privacyenabledstate.CouchDBCommonStorageTestEnv{},
 }
 
+const (
+	levelDBtestEnvName = "levelDB_TestEnv"
+	couchDBtestEnvName = "couchDB_TestEnv"
+)
+
+var testEnvsMap = map[string]privacyenabledstate.TestEnv{
```
lindluni: Why not just replace testEnvs with the map:

```
var testEnvs = map[string]privacyenabledstate.TestEnv {
	"levelDB": &privacyenabledstate.LevelDBCommonStorageTestEnv{},
	"couchDB": &privacyenabledstate.CouchDBCommonStorageTestEnv{},
}
```

and then you can iterate over it with:

```
for _, dbEnv := range testEnvs {
	t.Run(dbEnv.GetName(), func(t *testing.T) { testPurgeMgrForCommittingPvtDataOfOldBlocks(t, dbEnv) })
}
```
and reference directly with `coudbDB` and `levelDB` adding the constants and the secondary map is just an abstraction of testEnvs

core/ledger/kvledger/txmgmt/statedb/statecouchdb/statecouchdb_test_export.go
```diff
@@ -16,9 +16,11 @@ import (
 	"github.com/hyperledger/fabric/common/metrics/disabled"
 	"github.com/hyperledger/fabric/core/ledger/kvledger/txmgmt/statedb"
 	"github.com/hyperledger/fabric/core/ledger/util/couchdb"
+	"github.com/hyperledger/fabric/core/ledger/util/couchdbtest"
```
manish-sethi, line 17: The new addition of the code in this file is not correct. The consumers of "TestVDBEnv" defined in this file, start couch and stop couchdb explicitly. I would suggest to review the code of the consumers and if you want to move the code for starting/stopping here, better to enclose this in the TestVDBEnv but still controlling the start and stop explicitly in the consumer so that it can be done once at the package level...

manish-sethi (approved):
Thanks Derek

PR #483, opened 2020-01-09
Add code coverage to fabric verify builds.

Vijaypunugubati:
Adding code coverage to fabric verify builds.
- Improvement


ci/azure-pipelines.yml
```diff
@@ -57,6 +57,8 @@ jobs:
         displayName: Install SoftHSM
       - script: make unit-test
         displayName: Run Unit Tests
+      - script: make profile
+        displayName: Code Coverage
```
sykesm, line 61: Isn't this running unit tests again with profile mode on? Why would we do it twice?

sykesm: Also, this is running as part of the PR. Do we really need code coverage in PR checks?  Maybe we do but I would value fast execution over a report.

sykesm (changes requested):
IIUC, this change would make units run twice with the current makefile.

```
.PHONY: profile
profile: export JOB_TYPE=PROFILE
profile: unit-test
```

Also, the CR has a merge commit at the bottom that shouldn't be there. In a simple PR like this, I'd expect to see a single commit that we can rebase and merge on the target branch. Can you please remove the merge commit?

sykesm:
Let's have an offline discussion about if/where/when we want to have code coverage reports generated. Unless there's a known consumer that will take some action on these, I don't know that having them really creates value above and beyond the current ability to generate them from the make process.

Thanks.

PR #485, opened 2020-01-10
Reject re-defining identical uncommitted cc def

guoger:
When user attempts to update an uncommitted chaincode definition
with identical content as before, endorser should fail the invocation.
Otherwise, it results in an empty wset, which would be validated
against LifecycleEndorsement policy by default, instead of the policy
of org's own implicit collection.

### DIFF
--- a/core/chaincode/lifecycle/lifecycle.go
+++ b/core/chaincode/lifecycle/lifecycle.go
@@ -407,6 +407,30 @@ func (ef *ExternalFunctions) ApproveChaincodeDefinitionForOrg(chname, ccname str
 	}
 
 	privateName := fmt.Sprintf("%s#%d", ccname, requestedSequence)
+
+	// if requested sequence is not committed, and attempt is made to update its content,
+	// we need to check whether new definition actually contains updated content, to avoid
+	// empty write set.
+	if requestedSequence == currentSequence+1 {
+		uncommittedMetadata, ok, err := ef.Resources.Serializer.DeserializeMetadata(NamespacesName, privateName, orgState)
+		if err != nil {
+			return errors.WithMessage(err, "could not fetch uncommitted definition")
+		}
+
+		if ok {
+			logger.Debugf("Attempting to redefine uncommitted definition at sequence %d", requestedSequence)
+
+			uncommittedParameters := &ChaincodeParameters{}
+			if err := ef.Resources.Serializer.Deserialize(NamespacesName, privateName, uncommittedMetadata, uncommittedParameters, orgState); err != nil {
+				return errors.WithMessagef(err, "could not deserialize namespace %s as chaincode", privateName)
+			}
+
+			if err := uncommittedParameters.Equal(cd.Parameters()); err == nil {
+				return errors.Errorf("attempted to redefine uncommitted sequence (%d) for namespace %s with unchanged content", requestedSequence, ccname)
+			}
+		}
+	}
+
 	if err := ef.Resources.Serializer.Serialize(NamespacesName, privateName, cd.Parameters(), orgState); err != nil {
 		return errors.WithMessage(err, "could not serialize chaincode parameters to state")
 	}
@@ -423,7 +447,7 @@ func (ef *ExternalFunctions) ApproveChaincodeDefinitionForOrg(chname, ccname str
 		return errors.WithMessage(err, "could not serialize chaincode package info to state")
 	}
 
-	logger.Infof("successfully approved definition %s, name '%s' on channel '%s'", cd, ccname, chname)
+	logger.Infof("Successfully approved definition %s, name '%s' on channel '%s'", cd, ccname, chname)
 
 	return nil
 }
--- a/core/chaincode/lifecycle/lifecycle_test.go
+++ b/core/chaincode/lifecycle/lifecycle_test.go
@@ -662,7 +662,7 @@ var _ = Describe("ExternalFunctions", func() {
 			})
 		})
 
-		Context("when the sequence number already has a definition", func() {
+		Context("when the sequence number already has a committed definition", func() {
 			BeforeEach(func() {
 				err := resources.Serializer.Serialize("namespaces", "cc-name", &lifecycle.ChaincodeDefinition{
 					Sequence: 5,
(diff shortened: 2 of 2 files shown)


core/chaincode/lifecycle/lifecycle.go
```diff
@@ -407,6 +407,24 @@ func (ef *ExternalFunctions) ApproveChaincodeDefinitionForOrg(chname, ccname str
...
+	if ok {
+		logger.Debugf("Attempting to redefine uncommitted definition at sequence %d", requestedSequence)
+
+		uncommittedParameters := &ChaincodeParameters{}
+		if err := ef.Resources.Serializer.Deserialize(NamespacesName, privateName, uncommittedMetadata, uncommittedParameters, orgState); err != nil {
+			return errors.WithMessagef(err, "could not deserialize namespace %s as chaincode", privateName)
+		}
+
+		if err := uncommittedParameters.Equal(cd.Parameters()); err == nil {
+			return errors.Errorf("attempted to redefine uncommitted sequence (%d) for namespace %s with unchanged content", requestedSequence, ccname)
```
sykesm: This breaks a (bad?) test:
https://dev.azure.com/Hyperledger/Fabric/_build/results?buildId=5076&view=logs&j=17737827-9315-5d9e-c14a-16a17e90ef50&t=4c34d36c-6bff-5a33-cfcd-b209c5bac4c3&l=4342

core/chaincode/lifecycle/lifecycle.go
```diff
@@ -423,7 +441,7 @@ func (ef *ExternalFunctions) ApproveChaincodeDefinitionForOrg(chname, ccname str
 		return errors.WithMessage(err, "could not serialize chaincode package info to state")
 	}
 
-	logger.Infof("successfully approved definition %s, name '%s' on channel '%s'", cd, ccname, chname)
+	logger.Infof("Successfully approved definition %s, name '%s' on channel '%s'", cd, ccname, chname)
```
sykesm, line 450: why did we change this message? Historically, we haven't tried to convert our messages to lower case, only errors.

core/chaincode/lifecycle/lifecycle.go
guoger: test is correct. i forgot the case that pkg id can still be updated after def being committed. fixed the code (should've wait for CI go green before submitting for review...)

core/chaincode/lifecycle/lifecycle.go
guoger: it's changing from lower case to upper case, to align with other log messages.

sykesm: This file is odd. We have many errors that are upper case but our messages are lower case. Doesn't make sense. In most of fabric, our info messages are upper case.

Doesn't hold this up but we should probably revisit errors/messages in here as a separate item.

sykesm (approved):
Assuming tests work out..

ale-linux:
Sorry for jumping in late: what's the problem with the current code? An empty write-set without events needs no endorsement by definition. Could you spell out what the problem actually is please?

sykesm:
Opened JIRA for the unit test failure:
https://jira.hyperledger.org/browse/FAB-17375

jyellick:
> An empty write-set without events needs no endorsement by definition.

@ale-linux I don't think that's right.

https://github.com/hyperledger/fabric/blob/d80c5f5a60d59a78891e1c958018a4bfeb892182/core/committer/txvalidator/v20/plugindispatcher/dispatcher.go#L160

Basically, we always validate some endorsement policy for the origin namespace.  

This all comes back to a v1.2 change https://github.com/hyperledger/fabric/commit/0d25786b89d0c5c89be4b8e44845217adf7448ef

This was done to address the fact that a client could construct a tx with an event, and have the tx and event marked valid automatically because there were no writes done.  With V1.2+ application capabilities on, we always include the initial namespace in our EP checks.

ale-linux:
Yeah, I know that change, but I still think that it was a bit misguided. I accept that if the tx emits an event, we should consider it as a "write" action and subject it to some endorsement policy (and currently, it's not very clear which policy should be, if there are no writes but only reads and an event, as you and I have discussed). In the case of cc lifecycle, the empty write-set is basically a no-op, so I'm not sure I would fail it.. what is the risk of not doing so?

jyellick:
@ale-linux So for the immediate issue, we have the behavior that a no-op tx is validated against the cc endorsement policy, and because it was only executed by a single peer, is marked invalid.  As I see it, we have basically three options to address this.

1. As this PR attempts, simply inform the client that this invocation is a no-op, and therefore they should not bother submitting it to ordering.  Pros: it pushes the failure earlier (to endorsement time rather than commit time), and it produces no future backwards compatibility concerns  Cons: A no-op is not actually harmful, and it makes automation a bit harder.
2. When the invocation of this function would be a no-op, include a an unnecessary re-write to a key.  Say, setting the sequence to its current value.  Pros: it eliminates the failure, and has limited future backwards compatibility concerns, Cons: It's pretty hacky and unintuitive to inject unneeded writes, just to signal which EP to evaluate.
3. Modify the 'misguided' logic in the validation path to allow no-op txes through.  Pros: it eliinates the failure, Cons: This has significant backwards compatibility concerns, and much broader implications and certainly deserves more discussion.

Although the discussion around (3) might deserve it's own JIRA, since we are here:  If we have a tx which claims to invoke a chaincode, and has a read-set, but has no affect on state, and no events, then what is the value of submitting it to ordering at all?  Should transactions be 'valid by default'? That seems dangerous from a security perspective to me.

The only 'useful' thing I can think of for a no-op tx is demonstrating a 'strong read', that at block N, the chaincode invocation of function Foo was consistent with the given result.  But, in order for this to be valuable, we need to know it was executed correctly, which goes back to needing an endorsement check.

core/chaincode/lifecycle/lifecycle.go
guoger: will do a follow-up

mastersingh24 (approved):
LGTM and I think this is a good option.   Will let @jyellick merge

ale-linux:
Makes sense @jyellick , let's treat the "empty write set with no events" tx in a separate JIRA since it's a bigger change and requires some more thinking to ensure we don't create forks. LGTM

PR #487, opened 2020-01-10
[FAB-17303] Support PrivateDataImplicitCollection properties for private data dissemination

caod123:
#### Type of change
- New feature

#### Description
Currently supports `maxPeerCount` and `requiredPeerCount`

#### Related issues
[FAB-17303](https://jira.hyperledger.org/browse/FAB-17303)


sampleconfig/configtx.yaml
```diff
@@ -72,6 +72,31 @@ Organizations:
...
+        PrivateDataImplicitCollection:
+            # MaxPeerCount is the maximum number of other peers (across authorized organizations)
+            # that each endorsing peer will attempt to distribute the private data to. If an
+            # endorsing peer becomes unavailable between endorsement time and commit time, other
+            # peers that are collection members but who did not yet receive the private data at
+            # endorsement time, will be able to pull the private data from peers the private data
+            # was disseminated to. If this value is set to 0, the private data is not disseminated
+            # at endorsement time, forcing private data pulls against endorsing peers on all authorized
+            # peers at commit time.
+            MaxPeerCount: 1
```
yacovm: >              MaxPeerCount: 1 

why is this 1 by default? Shouldn't it be high like 100 or something?

caod123: Wasn't sure how relevant the sample config was supposed to be. `1` was never supposed to be considered as a default. As a matter of fact currently the property doesn't even need to be specified and will just default to 0 if not specified. Of course this can all be changed to support actual defaults and suggested sample values if necessary. I'm open to suggestions.

yacovm: We don't need to change the code, we should just put here a sane default which is in my opinion, to send to everyone that is eligible.

caod123: Sounds good. I'll talk to @denyeart to see if he has a preference for what it should be.

caod123: Talked to Dave and we decided on no code level default for when the properties aren't specified and a sampleconfig default of `requiredPeerCount: 0` and `maxPeerCount: 1`.

caod123:
@jyellick would you be able to review this?

jyellick (changes requested):
This overall looks pretty good, but you've missed a significant component.  You need to tie processing of this configuration element to the version of the channel capabilities.  Notably, if this config element is specified for pre-v2.0 fabric, the config must be rejected (just as it would be by any pre-v2.0 fabric binary).  Only for channels with the v2.0 channel level capability should allow configurations with this element to be processed.

caod123:
> This overall looks pretty good, but you've missed a significant component. You need to tie processing of this configuration element to the version of the channel capabilities. Notably, if this config element is specified for pre-v2.0 fabric, the config must be rejected (just as it would be by any pre-v2.0 fabric binary). Only for channels with the v2.0 channel level capability should allow configurations with this element to be processed.

Ah I knew I was forgetting something, but I couldn't remember. Thanks for the reminder.

caod123:
@jyellick let me know if that looks about right

common/capabilities/application.go
```diff
@@ -139,6 +139,12 @@ func (ap *ApplicationProvider) StorePvtDataOfInvalidTx() bool {
 	return ap.v142 || ap.v20
 }
 
+// PrivateDataImplicitCollectionConfig returns whether org-specific implicit collection configuration
+// may be specific in the channel application org config.
```
denyeart: specified

sampleconfig/configtx.yaml
```diff
@@ -72,6 +72,31 @@ Organizations:
             - Host: 127.0.0.1
               Port: 7051
 
+        # PrivateDataImplicitCollection stores the configuration for this org's implicit
+        # collection.
+        PrivateDataImplicitCollection:
+            # MaxPeerCount is the maximum number of other peers (across authorized organizations)
```
denyeart: (within the organization)

sampleconfig/configtx.yaml
```diff
@@ -72,6 +72,31 @@ Organizations:
...
+            # that each endorsing peer will attempt to distribute the private data to. If an
+            # endorsing peer becomes unavailable between endorsement time and commit time, other
+            # peers that are collection members but who did not yet receive the private data at
+            # endorsement time, will be able to pull the private data from peers the private data
+            # was disseminated to. If this value is set to 0, the private data is not disseminated
+            # at endorsement time, forcing private data pulls against endorsing peers on all authorized
+            # peers at commit time.
+            MaxPeerCount: 1
+            # RequiredPeerCount is the minimum number of peers (across authorized organizations)
```
denyeart: (within the organization)

sampleconfig/configtx.yaml
denyeart: I would put RequiredPeerCount before MaxPeerCount. That's how it is in the associated collection config doc, and it is most natural to think of dissemination for "0 to 1" peers rather than dissemination for "1 to 0" peers.

denyeart (commented):
I think we're missing the part that wires the new config properties into the runtime. At least, in my tests, I'm still getting dissemination to 0 peers, even though there are two peers in my org that have established gossip connection.

common/channelconfig/applicationorg.go
```diff
@@ -54,6 +60,12 @@ func NewApplicationOrgConfig(id string, orgGroup *cb.ConfigGroup, mspConfig *MSP
 		},
 	}
 
+	if !aoc.Capabilities().PrivateDataImplicitCollectionConfig() {
+		if _, ok := orgGroup.Values[PrivateDataImplicitCollectionKey]; ok {
```
denyeart: Hmmm, I tried not configuring V2_0 capability while PrivateDataImplicitCollection was configured... it got to this 'if' statement, but didn't enter it (didn't return the error as I expected).

denyeart (commented):
We'll also need integration test. If an actual functional test is difficult, at least add to an integration test configtx.yaml to ensure the new config doesn't break anything.

common/capabilities/application.go
caod123: Ack

sampleconfig/configtx.yaml
caod123: Ack

sampleconfig/configtx.yaml
caod123: Ack

sampleconfig/configtx.yaml
caod123: Sounds good, I'm not sure why I had MaxPeerCount first, but I distinctly remember doing so because I saw somewhere it was before Required. I can't find where now so maybe I was just seeing things. I'll swap the orders where necessary.

common/channelconfig/applicationorg.go
caod123: Weird, maybe the key was mispelled. I'll take a look.

caod123:
I pushed the comment updates but left the integration test and capability check since I'm still not sure what's missing to get the config to read properly. I'll keep taking a stab as it but will probably need to sync with @jyellick to dig deeper into the config deserialization to find out what I'm missing/doing wrong.

common/channelconfig/applicationorg.go
caod123: I think I'm doing something wrong with how the orgGroup.Values is being built since it can't even seem to find the `PrivateDataImplicitCollectionKey` in the list of values. That would explain both why the property doesn't seem to work when the capability is configured and specified and also why this line isn't erroring when the capability isn't there.

integration/pvtdata/pvtdata_test.go
```diff
@@ -103,6 +103,64 @@ var _ bool = Describe("PrivateData", func() {
...
+				}
+			}
+
+			nwo.UpdateConfig(network, orderer, channelID, config, updatedConfig, false, network.Peers[0], network.Peers...)
+
+			By("deploying new lifecycle chaincode and adding marble1")
+			testChaincode := chaincode{
+				Chaincode: newLifecycleChaincode,
+				isLegacy:  false,
+			}
```
jyellick, line 156: What are we actually testing here? Would any of this fail if line 150 were commented out?

jyellick (commented):
This seems like it's getting close, though as I've given it a more thorough review there are still a few things that are not the fault of your changeset here, but that make me question the approach.

We have:

1) Added new stuff to `configtx.yaml` and all of the associated gorp in order to successfully parse it, and encoded it into the channel config.
2) Exercise no-where in an integration test this `configtx.yaml` value.
3) Don't anticipate any normal users providing this information via `configtx.yaml`

This comes back to two things:

1) The defaults for this value are kind of terrible.  For a highly available setup, you'll almost definitely want MaxPeerCount to be greater than 0.
2) It's not clear that you'd want the same defaults for all your channels.  In fact, I'm also beginning to question whether MemberOnlyRead and MemberOnlyWrite make sense as global channel constants.  There are definitely a number of usecases where MemberOnlyWrite breaks things, but it's also a good idea to turn on as a more secure default.

So, if by (1), we probably want this value set in every channel, and by (2) we don't think a sane default makes sense for all channels, then it probably doesn't make sense to allow the org to provide a default via the consortium definition (this change doesn't do this).

So, we don't want a default, but we do want it to be specified, which leads us to an interesting conundrum.  When channels are created, typically only one org does the creation request.  Because of this, it doesn't really make sense to set the private data implicit collection stuff as part of the channel creation, because it would require getting a signature from every org who's config is modified.

Which brings us back to the problem originally enumerated.  Since the info isn't set in the genesis block for the orderer system channel, and it's not set in the channel creation tx, then really, there's no way for `configtxgen` to use this info, so why are we implementing it?  We allow you to specify anchor peers in `configtx.yaml`, but without the terrible and hacky `-outputAnchorPeersUpdate` flag which we are removing, they will become similarly superfluous.

I know we're trying to squeeze this in before we release, but I do think we should probably talk through exactly what we're expecting users to do here.

gossip/service/eventer.go
```diff
@@ -112,6 +114,10 @@ func (ag *appGrp) AnchorPeers() []*peer.AnchorPeer {
 	return ag.anchorPeers
 }
 
+func (ag *appGrp) PrivateDataImplicitCollection() *peer.PrivateDataImplicitCollection {
```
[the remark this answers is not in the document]
caod123: Good catch, some of this code was added initially when I wasn't sure yet who would need to consume the configuration. After the PR grew in size some of these older functions slipped through after I had finalized where the values were actually used.

integration/pvtdata/pvtdata_test.go
caod123: I guess this test was more validating as a sanity check if the capability was properly set. Now that it's a channel capability which is defaulted to have V2_0 support in the integration test, none of this should fail anyways. In all honesty it would probably make more sense to add some implicit collection read/write support in the test chaincode itself to actually verify if dissemination works. That might be worth exploring on a separate issue though since this one is already growing fairly large.

PR #488, opened 2020-01-10
lifecycle cache should remove stale chaincode definition references

wlahti:
#### Type of change
- Bug fix

#### Description

When a new chaincode package is installed and subsequently referenced by a committed chaincode definition that takes the place of a previous chaincode definition (aka they have the same name), the previous (stale) chaincode definition should no longer be listed as referenced by the previous
chaincode package in the lifecycle cache.

#### Related issues

FAB-17372

### DIFF
--- a/core/chaincode/lifecycle/cache.go
+++ b/core/chaincode/lifecycle/cache.go
@@ -434,10 +434,50 @@ func (c *Cache) update(initializing bool, channelID string, dirtyChaincodes map[
 			continue
 		}
 
+		privateName := fmt.Sprintf("%s#%d", name, chaincodeDefinition.Sequence)
+		hashKey := FieldKey(ChaincodeSourcesName, privateName, "PackageID")
+		hashOfCCHash, err := orgState.GetStateHash(hashKey)
+		if err != nil {
+			return errors.WithMessagef(err, "could not check opaque org state for chaincode source hash for '%s' on channel '%s'", name, channelID)
+		}
+
+		localChaincode, ok := c.localChaincodes[string(hashOfCCHash)]
+		if !ok {
+			localChaincode = &LocalChaincode{
+				References: map[string]map[string]*CachedChaincodeDefinition{},
+			}
+			c.localChaincodes[string(hashOfCCHash)] = localChaincode
+		}
+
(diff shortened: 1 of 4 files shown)


core/chaincode/lifecycle/cache.go
```diff
@@ -434,11 +434,53 @@ func (c *Cache) update(initializing bool, channelID string, dirtyChaincodes map[
...
+
+		localChaincode, ok := c.localChaincodes[string(hashOfCCHash)]
+		if !ok {
+			localChaincode = &LocalChaincode{
+				References: map[string]map[string]*CachedChaincodeDefinition{},
+			}
+			c.localChaincodes[string(hashOfCCHash)] = localChaincode
+		}
+
+		if !initializing {
```
sykesm, line 452: I think this block would be easer to grok if it were collapsed a bit:

```go
		if !initializing {
			// check for existing local chaincodes that reference this chaincode name on this channel
			for _, lc := range c.localChaincodes {
				// track which channel, if any, we removed a local chaincode reference from
				if ref, ok := lc.References[channelID][name]; ok {
					if ref.InstallInfo == nil || localChaincode.Info == nil {
						continue
					}
					if ref.InstallInfo.PackageID == localChaincode.Info.PackageID {
						continue
					}

					// remove existing local chaincode reference, which referred to an old chaincode definition
					delete(lc.References[channelID], name)
					if len(lc.References[channelID]) == 0 {
						delete(lc.References, channelID)
					}
				}

				// TODO FAB-17046: finally, check to see if this chaincode is
				// referenced in any channel. if not, stop the chaincode here
				// if len(localChaincode.References) == 0 {
				// 	logger.Debug("chaincode package with label %s is no longer referenced and will be stopped", localChaincode.Info.Label)
				// }
			}
		}
```

I think things would be even nicer if we made had a type defined for References and created a `Remove` operation that encapsulates this bit entirely.

wlahti: Agreed. Thanks for the input! 

I've updated the existing block and will follow-up by defining a type for References and its related operations.

sykesm (approved):

PR #492, opened 2020-01-13
Chaincode persistence uses period separator

sykesm:
- Use a period separator between the chaincode label and hash
- Update chaincode store file matcher to explicitly look for files with
  a suffix of exactly 64 hex characters + ".tar.gz" to reduce depenency
  on special characters in the label.

#### Related Issue
[FAB-17315](https://jira.hyperledger.org/browse/FAB-17315)


mastersingh24 (approved):
LGTM but @jyellick should take a look as well

jyellick (approved):

PR #497, opened 2020-01-14
[FAB-17379] Fix print format for querycommitted

satota2:
This patch fixes the print format for `querycommitted` command to improve the readability.

#### Type of change

- Bug fix

#### Description

When running `peer lifecycle chaincode querycommitted` command with only a channel name (specified), the command displays multiple chaincode definitions without line breaks.

To improve readability, this patch adds line breaks to the print format.

#### Related issues

https://jira.hyperledger.org/browse/FAB-17379


wlahti:
Looks good to me. Thanks for catching this. In the integration tests we only use this query with json output so I had never noticed the lack of new lines between definitions. For completeness though, you could update the test case to include checking for a new line at the end of each chaincode definition.

sykesm (dismissed):
Please modify an existing test to make an assertion on the code you're changing (or add a test if needed).

satota2:
Thank you for having reviews.
New patch updates an existing test to check for a new line at the end of each chaincode definition.

denyeart (approved):

PR #498, opened 2020-01-15
[FAB-16895] Move ledger benchmark tests from fabric-test repo

manish-sethi:

#### Type of change

- Test update

#### Description

Move ledger benchmark tests from _fabric-test_ repo to _fabric_ repo under _core/ledger_ package
so that it can eventually be deleted from _fabric-test_ repo.

### DIFF
--- a/core/ledger/kvledger/benchmark/chainmgmt/chains.go
+++ b/core/ledger/kvledger/benchmark/chainmgmt/chains.go
@@ -1,0 +1,157 @@
+/*
+Copyright IBM Corp. All Rights Reserved.
+
+SPDX-License-Identifier: Apache-2.0
+*/
+
+package chainmgmt
+
+import (
+	"fmt"
+	"os"
+	"path/filepath"
+	"sync"
+
+	"github.com/hyperledger/fabric-protos-go/common"
+	"github.com/hyperledger/fabric/common/configtx/test"
+	"github.com/hyperledger/fabric/core/ledger"
+	"github.com/hyperledger/fabric/core/ledger/ledgermgmt"
+	"github.com/hyperledger/fabric/core/ledger/ledgermgmt/ledgermgmttest"
+	"github.com/hyperledger/fabric/core/ledger/util/couchdb"
+)
+
+// ChainID is a type used for the ids for the chains for experiments
+type ChainID int
+
+func (chainID ChainID) String() string {
+	return fmt.Sprintf("chain-%04d", chainID)
+}
+
+// SimulationResult is a type used for simulation results
+type SimulationResult []byte
+
+// chainsMgr manages chains for experiments
+type chainsMgr struct {
+	ledgerMgr *ledgermgmt.LedgerMgr
+	mgrConf   *ChainMgrConf
+	batchConf *BatchConf
+	initOp    chainInitOp
+	chainsMap map[ChainID]*Chain
+	wg        *sync.WaitGroup
+}
+
+func newChainsMgr(mgrConf *ChainMgrConf, batchConf *BatchConf, initOp chainInitOp) *chainsMgr {
+	dataDir := filepath.Join(mgrConf.DataDir, "ledgersData")
+	ledgermgmtInitializer := ledgermgmttest.NewInitializer(dataDir)
+	ledgermgmtInitializer.Config.HistoryDBConfig.Enabled = true
+	if os.Getenv("useCouchDB") == "true" {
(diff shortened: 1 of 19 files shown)


core/ledger/kvledger/benchmark/chainmgmt/sanity_test.go
```diff
@@ -0,0 +1,33 @@
...
+package chainmgmt
+
+import (
+	"os"
+	"testing"
+
+	"github.com/stretchr/testify/require"
+)
+
+// TestChainMgmt is a basic sanity check test to catch any errors that could be caused by changes in the ledgermgmt ot kvledger packages
```
sykesm:
```suggestion
// TestChainMgmt is a basic sanity check test to catch any errors that could be caused by changes in the ledgermgmt or kvledger packages
```

core/ledger/kvledger/benchmark/chainmgmt/chains.go
```diff
@@ -0,0 +1,153 @@
...
+	ledgermgmtInitializer.Config.HistoryDBConfig.Enabled = true
+	if os.Getenv("useCouchDB") == "true" {
+		ledgermgmtInitializer.Config.StateDBConfig.StateDatabase = "CouchDB"
+		ledgermgmtInitializer.Config.StateDBConfig.CouchDB = &couchdb.Config{
+			Address:            "localhost:5984",
+			RedoLogPath:        filepath.Join(dataDir, "couchdbRedologs"),
+			UserCacheSizeMBs:   500,
+			MaxBatchUpdateSize: 500,
+		}
+	}
```
sykesm: Seems that we should be using the same patterns that we use in the rest of the tests and avoid hard-coded hosts and ports of cloud and rely on what's in COUCHDB_ADDR.

core/ledger/kvledger/benchmark/chainmgmt/chains.go
```diff
@@ -0,0 +1,153 @@
...
+	return chains
+}
+
+func (m *chainsMgr) waitForChainsToExhaustAllBlocks() {
+	m.wg.Wait()
+	m.ledgerMgr.Close()
+}
+
+// Chain extends ledger.PeerLedger and the experiments invoke ledger functions via a chain type
```
sykesm:
```suggestion
// Chain embeds ledger.PeerLedger and the experiments invoke ledger functions via a chain type
```
There is no inheritance in `go`.

core/ledger/kvledger/benchmark/chainmgmt/conf.go
```diff
@@ -0,0 +1,21 @@
+/*
+Copyright IBM Corp. All Rights Reserved.
+
+SPDX-License-Identifier: Apache-2.0
+*/
+
+package chainmgmt
+
+// ChainMgrConf captures the configurations meant at the level of chainMgr
```
sykesm:
```
// ChainMgrConf captures the configurations meant at the level of chainMgr
```
Not sure I grok that.

core/ledger/kvledger/benchmark/chainmgmt/conf.go
```diff
@@ -0,0 +1,21 @@
...
+// DataDir field specifies the filesystem location where the chains data is maintained
```
sykesm:
```
// DataDir field specifies the filesystem location where the chains data is maintained
// NumChains field specifies the number of chains to instantiate
```
If this are field doc, shouldn't they be with the fields? (Here and below.)

core/ledger/kvledger/benchmark/experiments/util.go
```diff
@@ -0,0 +1,143 @@
...
+	"john",
+}
+
+//TestValue is a struct for holding the test value
+type TestValue struct {
+	Value string
+}
+
+func constructKey(keyNumber int) string {
+	return fmt.Sprintf("%s%09d", "key_", keyNumber)
```
sykesm:
```suggestion
	return fmt.Sprintf("key_%09d", keyNumber)
```

sykesm (commented):
Mostly nits with two exceptions:

1. Please use an actual temporary directory as returned by ioutil.TempDir or similar.
2. Please use COUCHDB_ADDR instead of hard-coded address

Biggest concern overall is the amount of bash that's being used to glue this together; it's kind of insane for what it's doing. Even things like using the output of the test command to look for the `FAIL` string instead of honoring the exit code.

denyeart (approved):

PR #502, opened 2020-01-15
Fix go routine leak in core/scc tests

sykesm:
- Implement CloseSend on the in-process stream
- Consume registration message from system chaincode in test
- Use dot import for gomega in `scc_test.go`

FAB-16112


sykesm:
CI failure looks like FAB-17376

jyellick (approved):

PR #508, opened 2020-01-16
Stop _lifecycle chaincodes that are no longer referenced

wlahti:
#### Type of change
- New feature

#### Description
When a chaincode installed with _lifecycle is no longer referenced by any chaincode definition in any channel, it should be stopped.

#### Related issues
FAB-17046

### DIFF
--- a/core/chaincode/handler.go
+++ b/core/chaincode/handler.go
@@ -392,8 +392,8 @@ func (h *Handler) ProcessStream(stream ccintf.ChaincodeStream) error {
 				chaincodeLogger.Debugf("received EOF, ending chaincode support stream: %s", rmsg.err)
 				return rmsg.err
 			case rmsg.err != nil:
-				err := errors.Wrap(rmsg.err, "receive failed")
-				chaincodeLogger.Errorf("handling chaincode support stream: %+v", err)
+				err := errors.Wrap(rmsg.err, "receive from chaincode support stream failed")
+				chaincodeLogger.Debugf("%+v", err)
 				return err
 			case rmsg.msg == nil:
 				err := errors.New("received nil message, ending chaincode support stream")
--- a/core/chaincode/lifecycle/cache.go
+++ b/core/chaincode/lifecycle/cache.go
@@ -450,9 +450,9 @@ func (c *Cache) update(initializing bool, channelID string, dirtyChaincodes map[
 		}
 
 		if !initializing {
-			// check for existing local chaincodes that reference this chaincode name on this channel
+			// check for existing local chaincodes that reference this chaincode
+			// name on this channel
 			for _, lc := range c.localChaincodes {
-				// track which channel, if any, we removed a local chaincode reference from
 				if ref, ok := lc.References[channelID][name]; ok {
 					if ref.InstallInfo == nil || localChaincode.Info == nil {
 						continue
@@ -461,16 +461,26 @@ func (c *Cache) update(initializing bool, channelID string, dirtyChaincodes map[
 						continue
 					}
 
-					// remove existing local chaincode reference, which referred to an old chaincode definition
+					// remove existing local chaincode reference, which referred to a
+					// previous chaincode definition
 					delete(lc.References[channelID], name)
 					if len(lc.References[channelID]) == 0 {
 						delete(lc.References, channelID)
 
-						// TODO FAB-17046: finally, check to see if this chaincode is
-						// referenced in any channel. if not, stop the chaincode here
-						// if len(lc.References) == 0 {
-						// 	logger.Debug("chaincode package with label %s is no longer referenced and will be stopped", localChaincode.Info.Label)
-						// }
+						// check to see if this "local" chaincode is installed (an entry
+						// is added into local chaincodes for active chaincode definition
+						// references regardless of whether the peer has a chaincode
+						// package installed)
+						if lc.Info == nil {
+							continue
+						}
+
+						// finally, check to see if this chaincode is referenced in any
+						// channel. if not, stop the chaincode here
+						if len(lc.References) == 0 {
+							logger.Debugf("chaincode package with label %s is no longer referenced and will be stopped", lc.Info.Label)
+							c.chaincodeCustodian.NotifyStoppable(lc.Info.PackageID)
+						}
 					}
 				}
 			}
--- a/core/chaincode/lifecycle/custodian.go
+++ b/core/chaincode/lifecycle/custodian.go
@@ -15,10 +15,12 @@ import (
 //go:generate counterfeiter -o mock/chaincode_launcher.go --fake-name ChaincodeLauncher . ChaincodeLauncher
 type ChaincodeLauncher interface {
 	Launch(ccid string) error
+	Stop(ccid string) error
 }
 
 // ChaincodeCustodian is responsible for enqueuing builds and launches
-// of chaincodes as they become available.
+// of chaincodes as they become available and stops when chaincodes
+// are no longer referenced by an active chaincode definition.
 type ChaincodeCustodian struct {
 	cond       *sync.Cond
 	mutex      sync.Mutex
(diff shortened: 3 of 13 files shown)


integration/e2e/e2e_test.go
```diff
@@ -211,10 +212,61 @@ var _ = Describe("EndToEnd", func() {
...
+				Expect(err).NotTo(HaveOccurred())
+				hashStr := fmt.Sprintf("%x", util.ComputeSHA256(fileBytes))
+				return hashStr
+			}
+
+			modulePkgID := getPackageID(chaincode)
+			gopathPkgID := getPackageID(gopathChaincode)
+			filters := map[string][]string{
+				"name": {modulePkgID, gopathPkgID},
+			}
```
sykesm: Can we name this something like `initialContainerFilter` and use `upgradedChaincodeFilter` for the second set?

If not, nbd...

integration/lifecycle/lifecycle_test.go
```diff
@@ -66,8 +67,20 @@ var _ = Describe("Lifecycle", func() {
 		core.Ledger.State.StateDatabase = "CouchDB"
 		core.Ledger.State.CouchDBConfig.CouchDBAddress = couchAddr
 		processes[couchDB.Name] = couchProcess
+		org1TermFile := filepath.Join(testDir, "org1-term-file")
+		os.Setenv("ORG1_TERM_FILE", org1TermFile)
+		core.Chaincode.ExternalBuilders[0].EnvironmentWhitelist = append(core.Chaincode.ExternalBuilders[0].EnvironmentWhitelist, "ORG1_TERM_FILE")
```
sykesm: Is there always one external builder? (Honestly don't remember.) If not, we probably need more than referencing index 0.

core/chaincode/lifecycle/custodian.go
```diff
@@ -88,17 +102,25 @@ func (cc *ChaincodeCustodian) Work(buildRegistry *container.BuildRegistry, build
...
-				logger.Debugf("skipping build of chaincode '%s' as it is already in progress", chore.chaincodeID)
-				continue
-			}
-			err := builder.Build(chore.chaincodeID)
-			if err != nil {
-				logger.Warningf("could not build chaincode '%s': %s", chore.chaincodeID, err)
+			continue
+		}
+
+		if chore.stoppable {
```
sykesm, line 108: if // continue, if // continue // else looks like a switch.

Not something you need to address but I think this code could use a little TLC.

integration/e2e/e2e_test.go
```diff
@@ -211,10 +212,61 @@ var _ = Describe("EndToEnd", func() {
 			nwo.CommitChaincode(network, "testchannel", orderer, chaincode, testPeers[0], testPeers...)
 			nwo.InitChaincode(network, "testchannel", orderer, chaincode, testPeers...)
 
+			By("listing the containers after committing the chaincode definition")
```
sykesm, line 215: This is tricky. The test is placed in e2e so we can reduce the number of networks and, hopefully, improve the speed of the tests. I like that.

On the other hand, this is pretty unrelated to the kafka network test - especially since we're de-prioritizing kafka going forward.

Is there a better place to put this within lifecycle that would not extend the execution times greatly? The fact that we are deploying chaincode with docker means they will always take a long time...

sykesm (changes requested):
Looks pretty good. It would be nice to cleanup the test chaincode a bit and, if possible, move the legacy chaincode stop tests out of e2e.

Most of the other comments are questions.

integration/lifecycle/lifecycle_test.go
wlahti: nwo.New creates a network with just one external builder.

integration/e2e/e2e_test.go
wlahti: Done.

core/chaincode/lifecycle/cache.go
```diff
@@ -461,16 +461,22 @@ func (c *Cache) update(initializing bool, channelID string, dirtyChaincodes map[
...
 						delete(lc.References, channelID)
 
-						// TODO FAB-17046: finally, check to see if this chaincode is
-						// referenced in any channel. if not, stop the chaincode here
-						// if len(lc.References) == 0 {
-						// 	logger.Debug("chaincode package with label %s is no longer referenced and will be stopped", localChaincode.Info.Label)
-						// }
+						if lc.Info == nil {
+							continue
+						}
```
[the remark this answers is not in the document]
wlahti: This happens when a chaincode definition exists on the channel that doesn't have an associated chaincode package installed on the peer.

core/chaincode/handler.go
```diff
@@ -392,8 +392,8 @@ func (h *Handler) ProcessStream(stream ccintf.ChaincodeStream) error {
 				chaincodeLogger.Debugf("received EOF, ending chaincode support stream: %s", rmsg.err)
 				return rmsg.err
 			case rmsg.err != nil:
-				err := errors.Wrap(rmsg.err, "receive failed")
-				chaincodeLogger.Errorf("handling chaincode support stream: %+v", err)
+				err := errors.Wrap(rmsg.err, "receive from chaincode support stream failed")
+				chaincodeLogger.Debugf("%+v", err)
```
sykesm, line 396: Not exactly what I meant but good enough.

core/chaincode/lifecycle/cache.go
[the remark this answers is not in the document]
sykesm: Thank you for adding the comment.

integration/lifecycle/lifecycle_test.go
sykesm: Sure, and while true today, it may not be generally true. I commented on this because it's fragile. We don't have to hit it here but we should avoid this kind of thing in the future.

If you're using a topology or config that's generated from a shared source, it can change.

integration/lifecycle/lifecycle_test.go
```diff
@@ -206,6 +219,12 @@ var _ = Describe("Lifecycle", func() {
 		By("listing the installed chaincodes and verifying the channel/chaincode definitions that are using the chaincode package")
 		nwo.QueryInstalledReferences(network, "testchannel", chaincode.Label, chaincode.PackageID, network.Peer("Org2", "peer0"), []string{"My_1st-Chaincode", "Version-0.0"})
 
+		By("checking for term files that are written when the chaincode is stopped via SIGTERM")
+		for _, termFile := range termFiles {
+			_, err = os.Stat(termFile)
+			Expect(err).To(BeNil())
+		}
```
sykesm: There's a matcher for this: BeARegularFile

integration/e2e/e2e_test.go
sykesm: This one seemed to slip through the cracks. Do we have a place in lifecycle that makes sense?

If not, let's move; otherwise we should probably move.

integration/chaincode/module/main.go
```diff
@@ -200,10 +203,56 @@ func (t *SimpleChaincode) respond(stub shim.ChaincodeStubInterface, args []strin
...
+
 func main() {
-	err := shim.Start(&SimpleChaincode{})
-	if err != nil {
-		fmt.Fprintf(os.Stderr, "Exiting Simple chaincode: %s", err)
-		os.Exit(2)
+	termFile := os.Getenv("ORG1_TERM_FILE")
+	if termFile == "" {
+		termFile = os.Getenv("ORG2_TERM_FILE")
+	}
```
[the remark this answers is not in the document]
sykesm: Again, not exactly what I meant but improved.

sykesm (dismissed):

integration/e2e/e2e_test.go
wlahti: Created a task to move this: https://jira.hyperledger.org/browse/FAB-17425

integration/lifecycle/lifecycle_test.go
wlahti: Done

integration/lifecycle/lifecycle_test.go
wlahti: Done

sykesm (dismissed):
Works for me. While there are things we can do to clean this up, they can be done as follow-ups in master if we choose to.

wlahti:
ci-run still seems weird when there's a failed test (due to a flake) so I amended my commit as-is to update the commit hash and force pushed to retrigger everything. Hopefully everything passes again this time.

sykesm (approved):
LGTM

PR #512, opened 2020-01-17
Static leader should not give up retrieving blocks

wlahti:
#### Type of change

- Bug fix

#### Description
When a peer is configured as a static org leader for gossip, it should
not give up on retrieving blocks in the event that the orderering
service goes down for an extended period of time (i.e. longer than the
peer.deliveryclient.reconnectTotalTimeThreshold). Otherwise, the peer
must be restarted before it will act as the leader again.

#### Related issues

FAB-17327

### DIFF
--- a/core/deliverservice/deliveryclient.go
+++ b/core/deliverservice/deliveryclient.go
@@ -88,6 +88,7 @@ type deliverClient struct {
 // how it verifies messages received from it,
 // and how it disseminates the messages to other peers
 type Config struct {
+	IsStaticLeader bool
 	// ConnFactory returns a function that creates a connection to an endpoint
 	ConnFactory func(channelID string, endpointOverrides map[string]*comm.OrdererEndpoint) func(endpointCriteria comm.EndpointCriteria) (*grpc.ClientConn, error)
 	// ABCFactory creates an AtomicBroadcastClient out of a connection
@@ -292,7 +293,10 @@ func (d *deliverServiceImpl) newClient(chainID string, ledgerInfoProvider blocks
 	}
 	backoffPolicy := func(attemptNum int, elapsedTime time.Duration) (time.Duration, bool) {
 		if elapsedTime >= reconnectTotalTimeThreshold {
-			return 0, false
+			if !d.conf.IsStaticLeader {
+				return 0, false
+			}
+			logger.Warning("peer is a static leader, ignoring peer.deliveryclient.reconnectTotalTimeThreshold")
 		}
 		sleepIncrement := float64(time.Millisecond * 500)
 		attempt := float64(attemptNum)
--- a/gossip/service/gossip_service.go
+++ b/gossip/service/gossip_service.go
@@ -62,15 +62,17 @@ type DeliveryServiceFactory interface {
 }
 
 type deliveryFactoryImpl struct {
+	isStaticLeader bool
 }
 
 // Returns an instance of delivery client
-func (*deliveryFactoryImpl) Service(g GossipService, ec OrdererAddressConfig, mcs api.MessageCryptoService) (deliverclient.DeliverService, error) {
+func (d *deliveryFactoryImpl) Service(g GossipService, ec OrdererAddressConfig, mcs api.MessageCryptoService) (deliverclient.DeliverService, error) {
 	return deliverclient.NewDeliverService(&deliverclient.Config{
-		CryptoSvc:   mcs,
-		Gossip:      g,
-		ConnFactory: deliverclient.DefaultConnectionFactory,
-		ABCFactory:  deliverclient.DefaultABCFactory,
+		IsStaticLeader: d.isStaticLeader,
+		CryptoSvc:      mcs,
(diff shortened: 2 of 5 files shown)


core/deliverservice/client.go
```diff
@@ -30,18 +30,19 @@ type broadcastSetup func(blocksprovider.BlocksDeliverer) error
 // If further attempts should be made, it returns:
 // 	- a time duration after which the next attempt would be made, true
 // Else, a zero duration, false
-type retryPolicy func(attemptNum int, elapsedTime time.Duration) (time.Duration, bool)
+type retryPolicy func(attemptNum int, elapsedTime time.Duration, yieldLeadership bool) (time.Duration, bool)
```
sykesm: After making a pass through these changes, I don't think this update is needed. I think we can encapsulate the behavior of static or dynamic leaders within the retry functions without changing the contract throughout the code.

I believe that means this file needs no changes.

core/deliverservice/client_test.go
```diff
@@ -165,7 +165,7 @@ func testOrderingServiceConnFailure(t *testing.T, bdc blocksDelivererConsumer) {
 		setupInvoked++
 		return nil
 	}
-	backoffStrategy := func(attemptNum int, elapsedTime time.Duration) (time.Duration, bool) {
+	backoffStrategy := func(attemptNum int, elapsedTime time.Duration, yieldLeadership bool) (time.Duration, bool) {
```
sykesm: As all of these changes are being made to adhere to the updated backoff callback signature, I don't think we need them.

I believe that means this file needs no changes.

core/deliverservice/deliveryclient.go
```diff
@@ -290,16 +291,19 @@ func (d *deliverServiceImpl) newClient(chainID string, ledgerInfoProvider blocks
...
+			}
+			logger.Warning("peer is a static leader, ignoring peer.deliveryclient.reconnectTotalTimeThreshold")
 		}
 		sleepIncrement := float64(time.Millisecond * 500)
 		attempt := float64(attemptNum)
 		return time.Duration(math.Min(math.Pow(2, attempt)*sleepIncrement, reconnectBackoffThreshold)), true
 	}
 	connProd := comm.NewConnectionProducer(d.conf.ConnFactory(chainID, d.connConfig.OrdererEndpointOverrides), d.connConfig.toEndpointCriteria())
-	bClient := NewBroadcastClient(connProd, d.conf.ABCFactory, broadcastSetup, backoffPolicy)
+	bClient := NewBroadcastClient(connProd, d.conf.ABCFactory, broadcastSetup, backoffPolicy, d.conf.IsStaticLeader)
```
sykesm: If we don't wire this, we should be able to encapsulate the desired behavior within the `backoffPolicy`.

In my mind, that means we have two policies - the one that existed before these changes and one for a static configuration that always returns `0, false`.

gossip/service/gossip_service.go
```diff
@@ -58,19 +58,20 @@ type GossipService interface {
...
-	Service(g GossipService, oac OrdererAddressConfig, msc api.MessageCryptoService) (deliverclient.DeliverService, error)
+	Service(g GossipService, oac OrdererAddressConfig, msc api.MessageCryptoService, isStaticLeader bool) (deliverclient.DeliverService, error)
 }
 
 type deliveryFactoryImpl struct {
 }
 
 // Returns an instance of delivery client
-func (*deliveryFactoryImpl) Service(g GossipService, ec OrdererAddressConfig, mcs api.MessageCryptoService) (deliverclient.DeliverService, error) {
+func (*deliveryFactoryImpl) Service(g GossipService, ec OrdererAddressConfig, mcs api.MessageCryptoService, isStaticLeader bool) (deliverclient.DeliverService, error) {
```
sykesm: Can we move the isStaticLeader bool to the deliveryFactoryImpl struct and avoid the contract changes by passing through `InitGossipService` from peer/node/start.go?

That would seem to avoid a lot of these changes...

sykesm (changes requested):
I think we can reduce the scope of these changes that use the retryPolicy. That should make this a bit simpler.

core/deliverservice/deliveryclient.go
```diff
@@ -292,7 +293,10 @@ func (d *deliverServiceImpl) newClient(chainID string, ledgerInfoProvider blocks
 	}
 	backoffPolicy := func(attemptNum int, elapsedTime time.Duration) (time.Duration, bool) {
 		if elapsedTime >= reconnectTotalTimeThreshold {
-			return 0, false
+			if !d.conf.IsStaticLeader {
+				return 0, false
+			}
+			logger.Warning("peer is a static leader, ignoring peer.deliveryclient.reconnectTotalTimeThreshold")
```
sykesm, line 299: Just to reiterate the intent of my comments in earlier revisions and our discussion this morning, the idea of having two separate implementations of the policy would be to isolate the two different behaviors we exhibit.

sykesm (approved):

PR #520, opened 2020-01-20
Update _lifecycle resources in configtx.yaml

wlahti:
#### Type of change

- Documentation update

#### Description

Update _lifecycle resources in configtx.yaml, which included an outdated resource name (QueryNamespaaceDefinitions) and also lacked one resource (CheckCommitReadiness).

#### Related issues

https://jira.hyperledger.org/browse/FAB-17412

### DIFF
--- a/sampleconfig/configtx.yaml
+++ b/sampleconfig/configtx.yaml
@@ -141,10 +141,13 @@ Application: &ApplicationDefaults
         # (e.g.,who can receive Block events). This section does NOT specify the resource's
         # definition or API, but just the ACL policy for it.
         #
-        # User's can override these defaults with their own policy mapping by defining the
+        # Users can override these defaults with their own policy mapping by defining the
         # mapping under ACLs in their channel definition
 
         #---New Lifecycle System Chaincode (_lifecycle) function to policy mapping for access control--#
+
+        # ACL policy for _lifecycle's "CheckCommitReadiness" function
+        _lifecycle/CheckCommitReadiness: /Channel/Application/Writers
 
         # ACL policy for _lifecycle's "CommitChaincodeDefinition" function
         _lifecycle/CommitChaincodeDefinition: /Channel/Application/Writers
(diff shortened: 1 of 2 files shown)


sampleconfig/configtx.yaml
```diff
@@ -141,19 +141,22 @@ Application: &ApplicationDefaults
...
         #---New Lifecycle System Chaincode (_lifecycle) function to policy mapping for access control--#
 
+        # ACL policy for _lifecycle's "CheckCommitReadiness" function
+        _lifecycle/CheckCommitReadiness: /Channel/Application/Writers
+
         # ACL policy for _lifecycle's "CommitChaincodeDefinition" function
         _lifecycle/CommitChaincodeDefinition: /Channel/Application/Writers
 
         # ACL policy for _lifecycle's "QueryChaincodeDefinition" function
         _lifecycle/QueryChaincodeDefinition: /Channel/Application/Readers
```
nikhil550, line 157: Are these two different?

       # ACL policy for _lifecycle's "QueryChaincodeDefinition" function
        _lifecycle/QueryChaincodeDefinition: /Channel/Application/Readers

        # ACL policy for _lifecycle's "QueryChaincodeDefinitions" function
        _lifecycle/QueryChaincodeDefinitions: /Channel/Application/Readers

nikhil550 (dismissed):

sampleconfig/configtx.yaml
wlahti: Yes, they are two different queries.

sampleconfig/configtx.yaml
```diff
@@ -141,19 +141,22 @@ Application: &ApplicationDefaults
...
         # definition or API, but just the ACL policy for it.
         #
-        # User's can override these defaults with their own policy mapping by defining the
+        # Users can override these defaults with their own policy mapping by defining the
         # mapping under ACLs in their channel definition
 
         #---New Lifecycle System Chaincode (_lifecycle) function to policy mapping for access control--#
 
+        # ACL policy for _lifecycle's "CheckCommitReadiness" function
+        _lifecycle/CheckCommitReadiness: /Channel/Application/Writers
```
denyeart, line 150: Seems like this one should be /Channel/Application/Readers ?

wlahti: I was on the fence here. I think CheckCommitReadiness is mostly intended to be run by an admin right before they commit the chaincode definition. It could be confusing (or at least less than user-friendly) if a user is able to check the commit readiness successfully but then is blocked from performing the actual commit since they're a channel reader but not a writer.

sykesm (approved):
I'll leave the JIRA open and we can discuss reader (accurate) v. writer (expected users) with Jason.

EDIT: Closed the issue with a comment to revisit if the code needs to change; right now this update matches the code.

sampleconfig/configtx.yaml
sykesm: https://jira.hyperledger.org/browse/FAB-17412?focusedCommentId=66842&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-66842

PR #523, opened 2020-01-20
Improve chaincode lifecycle peer log messages

denyeart:
#### Type of change

- Improvement

#### Description

Suppressed commit endorsement success message when commit endorsement fails.
Clarified that approve and commit transaction are endorsements rather than actual commits.
Moved chaincode definition to end of messages and placed in brackets for readability.

### DIFF
--- a/core/chaincode/lifecycle/lifecycle.go
+++ b/core/chaincode/lifecycle/lifecycle.go
@@ -284,7 +284,7 @@ func (ef *ExternalFunctions) CheckCommitReadiness(chname, ccname string, cd *Cha
 		return nil, err
 	}
 
-	logger.Infof("Successfully checked commit readiness of chaincode definition %s, name '%s' on channel '%s'", cd, ccname, chname)
+	logger.Infof("Successfully checked commit readiness of chaincode name '%s' on channel '%s' with definition {%s}", ccname, chname, cd)
 
 	return approvals, nil
 }
@@ -304,8 +304,6 @@ func (ef *ExternalFunctions) CommitChaincodeDefinition(chname, ccname string, cd
 	if err = ef.Resources.Serializer.Serialize(NamespacesName, ccname, cd, publicState); err != nil {
 		return nil, errors.WithMessage(err, "could not serialize chaincode definition")
 	}
-
-	logger.Infof("Successfully committed definition %s, name '%s' on channel '%s'", cd, ccname, chname)
 
 	return approvals, nil
 }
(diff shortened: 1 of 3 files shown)


core/chaincode/lifecycle/lifecycle.go
```diff
@@ -305,8 +305,6 @@ func (ef *ExternalFunctions) CommitChaincodeDefinition(chname, ccname string, cd
 		return nil, errors.WithMessage(err, "could not serialize chaincode definition")
 	}
 
-	logger.Infof("Successfully committed definition %s, name '%s' on channel '%s'", cd, ccname, chname)
```
manish-sethi, line 308: Did you remove this because it is printed immediately after the one at line 287 above? If this is the case, isn't this a better message to retain instead (perhaps with a small modification to suit the tone of message at line 448 below)?

denyeart: I removed it from here because the commit may still fail in the caller, based on the returned approvals. In fact, that is what triggered this PR... we saw the commit request failing endorsement, but the peer claimed it was successful in this message. Moving the message until after the approvals have been verified in the caller fixes the issue.

sykesm (commented):
I know this is something we've discussed before, but, when we add or change a message to improve an experience, we should also be adding a test that ensures it gets rendered as we desire. It also makes it clear what it will look like to the end user.

Can we please cover at least one of these scenarios with a test that captures the formatted version of `cd` in the messages?

core/chaincode/lifecycle/lifecycle.go
manish-sethi: Ah, I see... this portion of the code split across scc and lifecycle often causes confusion to me.

denyeart:
I don't agree that every message requires a test, we have many occurrences of messages without tests. This PR doesn't change the cd content. If you want coverage for all messages, we could create a follow-on tech debt issue.

manish-sethi:
I am also not a big fan of micro-testing the code at the level of testing the log messages.

sykesm:
> I don't agree that every message requires a test, we have many occurrences of messages without tests. This PR doesn't change the cd content. If you want coverage for all messages, we could create a follow-on tech debt issue.

That's fine. I asked for *one* case to show that the serialized form of `cd` is consumable.

sykesm (dismissed):
> I don't agree that every message requires a test, we have many occurrences of messages without tests. This PR doesn't change the cd content. If you want coverage for all messages, we could create a follow-on tech debt issue.

That's fine. I asked for one case to show that the serialized form of cd is consumable.

sykesm:
> I don't agree that every message requires a test, we have many occurrences of messages without tests.

I want to push on this a bit. I'm not asking for all messages to be tested. I'm focusing on this one for a reason.  When the new lifecycle code was originally added, the author turned `ChaincodeDefinition` into a `fmt.Stringer`. That code has *no tests* and we merged it. Your changes are building on that untested base.

This code needs some sort of test. Explicit test is preferred on the Stringer than on the messages but there's nothing covering this stuff.

> This PR doesn't change the cd content. If you want coverage for all messages, we could create a follow-on tech debt issue.

Again, not all messages, these messages - as they build on an untested base. I'm all for you opening a tech debt issue and fixing it.

denyeart:
I agree that chaincode definition should be tested to ensure the string representation is suitable for log messages. I've added a test for that.
There was no precedent for testing the actual logging, and I don't want to add such a precedent.

sykesm:
Flake is recurrence of FAB-17344

sykesm (approved):

PR #527, opened 2020-01-21
[FAB-15920] Fix peer not respond when couchDB building indexes

DereckLuo:

#### Type of change

- Bug fix

#### Description

- Fixing peer not respond when couchDB building indexes. Moved initialization of
peer gRPC server and gossip service to post ledger manager initialization. Which
all ledger initialization will be compelte before peer gRPC starts.
- Un-exported Server from Peer struct, and moved the assigning of peer gPRC server 
into peer initialization function. 
#### Related issues

[FAB-15920](https://jira.hyperledger.org/browse/FAB-15920)

### DIFF
--- a/core/peer/peer.go
+++ b/core/peer/peer.go
@@ -110,7 +110,7 @@ func (p *Peer) updateTrustedRoots(cm channelconfig.Resources) {
 	trustedRoots = append(trustedRoots, p.ServerConfig.SecOpts.ServerRootCAs...)
 
 	// now update the client roots for the peerServer
-	err := p.Server.SetClientRootCAs(trustedRoots)
+	err := p.server.SetClientRootCAs(trustedRoots)
 	if err != nil {
 		msg := "Failed to update trusted roots from latest config block. " +
 			"This peer may not be able to communicate with members of channel %s (%s)"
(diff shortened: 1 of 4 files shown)


sykesm (changes requested):
I don't understand what this is fixing. The change defers the instantiation of the peer gRPC server but it doesn't do anything to change where registration with the server occurs or where the server starts.

This change also means that the peer instance can be referenced in a state where it's not completely initialized; there's no way to know the fallout of that change.

So, can you help me understand how this fixes the issue that was reported? Is there a test that helps us ensure it's been addressed?

sykesm:
@DereckLuo stopped by for a conversation. The issue is that the comm layer creates a net.Listener to provide to the gRPC server. As soon as we create a listener (done as part of the comm package),  clients can attempt to establish a connection to the server. The server, however, can only respond to requests after the server is started. This is the window this change was going after.

To add a bit of complication, services must be registered with with the gRPC server before Start is called. That means we can't just start the server and have it reject things until we register services later

Deferring the creation of the server defers the creation of the listener and reduces the size of the window between listen and serve. This is all good. The complication is that the peer is used by many components in the system and we don't want to inject a partially initialized peer to dependencies. If go down that path, we need to investigate the implications.

In concrete terms, the peer references the gRPC server to help keep TLS information up to date from channel configuration updates and other sources. If we start passing around a peer instance without this ability, bad things _may_ happen.

In summary, I understand the why and agree in principle with the change; we just need to handle the fallout.

DereckLuo:
Thanks for the review and guide. Changes has been applied accordingly.

sykesm (approved):
Thanks for making the server private and moving the population to Initialize. We can make additional improvements but this is good enough given what we have.

Thanks for the chat.

PR #532, opened 2020-01-22
Fix integration test for static leader

wlahti:
#### Type of change
- Test update

#### Description
Minor typo in the setup of the peer runners, which lead to the test not actually checking for the expected message.

#### Related issues
https://jira.hyperledger.org/browse/FAB-17327


wlahti:
Actually, it looks like this is already in the release-2.0 branch. Closing the PR.

PR #533, opened 2020-01-22
[FAB-17129] Configure peer to use PKCS#11 as BCCSP 

stephyee:
#### Type of change

Test update

#### Description

Configures and starts a peer using PKCS#11 as BCCSP

#### Related issues

[FAB-17129](https://jira.hyperledger.org/browse/FAB-17129)


sykesm:
We may come back to visit this but I think we should close it out for now. We should probably keep the branch around for a little bit...

PR #539, opened 2020-01-23
Fail fast if 'jq' is missing

yacovm:
If 'jq' is missing from the PATH then the chaincode installation fails and it
is not crystal clear why it failed unless you inspect the logs with care.

This change set makes the integration tests fail fast when attempting to install
chaincode and 'jq' doesn't exist in the path.

It prints:

'jq' is needed to build chaincode but it wasn't find in the PATH

and terminates the test.

Change-Id: Ic730ec6b90343c1da06818b16fb291963cbc20ac

### DIFF
--- a/integration/nwo/deploy.go
+++ b/integration/nwo/deploy.go
@@ -11,6 +11,7 @@ import (
 	"fmt"
 	"io/ioutil"
 	"os"
+	"os/exec"
 	"strconv"
 	"strings"
 
@@ -20,10 +21,10 @@ import (
 	"github.com/hyperledger/fabric/common/util"
 	"github.com/hyperledger/fabric/integration/nwo/commands"
 	"github.com/hyperledger/fabric/protoutil"
+	"github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
 	"github.com/onsi/gomega/gbytes"
 	"github.com/onsi/gomega/gexec"
-
-	. "github.com/onsi/gomega"
 	. "github.com/onsi/gomega/gstruct"
 )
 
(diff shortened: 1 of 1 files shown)


integration/helpers/images.go
```diff
@@ -38,3 +40,19 @@ func UniqueName() string {
...
+			continue
+		}
+		for _, f := range info {
+			if f.Name() == cmd {
+				return true
+			}
+		}
+	}
+	return false
+}
```
lindluni: Can't we just use https://golang.org/pkg/os/exec/#LookPath in place of where you call this function

integration/nwo/deploy.go
```diff
@@ -170,6 +171,11 @@ func PackageChaincodeLegacy(n *Network, chaincode Chaincode, peer *Peer) {
 }
 
 func InstallChaincode(n *Network, chaincode Chaincode, peers ...*Peer) {
+	// Ensure 'jq' exists in path, because we need it to build chaincode
+	if !helpers.ExistsInPath("jq") {
+		ginkgo.Fail("'jq' is needed to build chaincode but it wasn't find in the PATH")
```
lindluni: found, not "find"

integration/nwo/deploy.go
```diff
@@ -187,6 +193,11 @@ func InstallChaincode(n *Network, chaincode Chaincode, peers ...*Peer) {
 }
 
 func InstallChaincodeLegacy(n *Network, chaincode Chaincode, peers ...*Peer) {
+	// Ensure 'jq' exists in path, because we need it to build chaincode
+	if !helpers.ExistsInPath("jq") {
+		ginkgo.Fail("'jq' is needed to build chaincode but it wasn't find in the PATH")
```
lindluni: found, not "find"

integration/helpers/images.go
yacovm: I was looking for something like this but couldn't find it, thanks.

mastersingh24 (approved):
LGTM - but why exactly do we need/use jq?   :-)

yacovm:
> LGTM - but why exactly do we need/use jq? :-)

https://github.com/hyperledger/fabric/blob/master/integration/externalbuilders/binary/bin/build

PR #547, opened 2020-01-24
[FAB-17293] Remove rollback code from private data store

manish-sethi:

#### Type of change
- Bug fix

#### Description

FAB-17293 reports a corner case bug with a rare possibility. However, if this is to occur, it can cause a problem in restarting the peer. This bug could be triggered if during writing a block in blockstorage, block gets appended to file but during indexing, an I/O error is thrown. This would cause a panic in the peer and upon restart, it would be detected that the private data storage is behind the blockstorage (which is not an assumed scenario). However, the reverse situation is supported in the code - i.e., the private data store can be ahead of block store (in order to support
reset and rollback). This CR simply removes the rollback function on private data store for the last block. This has already been removed in 2.0 and in 1.4, this was logically of no use anymore.

#### Related issues

### DIFF
--- a/core/ledger/ledgerstorage/store.go
+++ b/core/ledger/ledgerstorage/store.go
@@ -143,7 +143,6 @@ func (s *Store) CommitWithPvtData(blockAndPvtdata *ledger.BlockAndPvtData) error
 	}
 
 	if err := s.AddBlock(blockAndPvtdata.Block); err != nil {
-		s.pvtdataStore.Rollback()
 		return err
 	}
 
--- a/core/ledger/pvtdatastorage/store.go
+++ b/core/ledger/pvtdatastorage/store.go
@@ -19,15 +19,10 @@ type Provider interface {
 }
 
 // Store manages the permanent storage of private write sets for a ledger
-// Beacsue the pvt data is supposed to be in sync with the blocks in the
-// ledger, both should logically happen in an atomic operation. In order
-// to accomplish this, an implementation of this store should provide
-// support for a two-phase like commit/rollback capability.
 // The expected use is such that - first the private data will be given to
 // this store (via `Prepare` funtion) and then the block is appended to the block storage.
-// Finally, one of the functions `Commit` or `Rollback` is invoked on this store based
-// on whether the block was written successfully or not. The store implementation
-// is expected to survive a server crash between the call to `Prepare` and `Commit`/`Rollback`
+// Finally, the functions `Commit` invoked on this store. The store implementation
+// is expected to survive a server crash between the call to `Prepare` and `Commit`.
 type Store interface {
 	// Init initializes the store. This function is expected to be invoked before using the store
 	Init(btlPolicy pvtdatapolicy.BTLPolicy)
@@ -51,15 +46,13 @@ type Store interface {
 	// for which this peer is a member; `ineligible` denotes that the missing private data belong to a
 	// collection for which this peer is not a member.
 	// This call does not commit the pvt data and store missing private data. Subsequently, the caller
-	// is expected to call either `Commit` or `Rollback` function. Return from this should ensure
+	// is expected to call `Commit` function. Return from this should ensure
 	// that enough preparation is done such that `Commit` function invoked afterwards can commit the
 	// data and the store is capable of surviving a crash between this function call and the next
 	// invoke to the `Commit`
 	Prepare(blockNum uint64, pvtData []*ledger.TxPvtData, missingPvtData ledger.TxMissingPvtDataMap) error
 	// Commit commits the pvt data passed in the previous invoke to the `Prepare` function
 	Commit() error
-	// Rollback rolls back the pvt data passed in the previous invoke to the `Prepare` function
-	Rollback() error
 	// ProcessCollsEligibilityEnabled notifies the store when the peer becomes eligible to recieve data for an
 	// existing collection. Parameter 'committingBlk' refers to the block number that contains the corresponding
 	// collection upgrade transaction and the parameter 'nsCollMap' contains the collections for which the peer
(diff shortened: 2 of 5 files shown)


core/ledger/pvtdatastorage/store.go
```diff
@@ -58,8 +58,6 @@ type Store interface {
 	Prepare(blockNum uint64, pvtData []*ledger.TxPvtData, missingPvtData ledger.TxMissingPvtDataMap) error
 	// Commit commits the pvt data passed in the previous invoke to the `Prepare` function
 	Commit() error
-	// Rollback rolls back the pvt data passed in the previous invoke to the `Prepare` function
-	Rollback() error
```
lindluni, line 62: Line 28 in this file references this function in it's comment, needs to be removed

manish-sethi: Thanks @btl5037. Good catch. I removed all the references to term "rollback" in the comments.

core/ledger/pvtdatastorage/store_impl_test.go
```diff
@@ -829,24 +825,6 @@ func TestRollBack(t *testing.T) {
 	assert.True(testDataKeyExists(t, store, datakeyTx0))
 	assert.True(testDataKeyExists(t, store, datakeyTx5))
 	assert.True(testMissingDataKeyExists(t, store, eligibleMissingdatakey))
```
wenjianqiao: Since PR removed Rollback() from store interface, is there any value to keep TestRollBack?

wenjianqiao: Since this PR removed Rollback() from pvtdata store, should remove TestRollBack(). Or rename the test if it provides more coverage.

core/ledger/pvtdatastorage/store_impl_test.go
```diff
@@ -829,24 +825,6 @@ func TestRollBack(t *testing.T) {
...
-		datakeyTx0.blkNum = uint64(i)
-		datakeyTx5.blkNum = uint64(i)
-		eligibleMissingdatakey.blkNum = uint64(i)
-		assert.True(testDataKeyExists(t, store, datakeyTx0))
-		assert.True(testDataKeyExists(t, store, datakeyTx5))
-		assert.True(testMissingDataKeyExists(t, store, eligibleMissingdatakey))
-	}
 }
 
 // TODO Add tests for simulating a crash between calls `Prepare` and `Commit`/`Rollback` - [FAB-13099]
```
wenjianqiao: Forgot to remove `Rollback` from the comment?

core/ledger/pvtdatastorage/store.go
```diff
@@ -51,15 +46,13 @@ type Store interface {
 	// for which this peer is a member; `ineligible` denotes that the missing private data belong to a
 	// collection for which this peer is not a member.
 	// This call does not commit the pvt data and store missing private data. Subsequently, the caller
-	// is expected to call either `Commit` or `Rollback` function. Return from this should ensure
+	// is expected to call either `Commit` function. Return from this should ensure
```
wenjianqiao: remove 'either'

core/ledger/pvtdatastorage/store_impl_test.go
manish-sethi: Yes, this can be removed - Done.

core/ledger/pvtdatastorage/store_impl_test.go
manish-sethi: Done. In fact, some of these discrepancies I believe are present in master as well.

core/ledger/pvtdatastorage/store.go
manish-sethi: done

wenjianqiao:
Looks good to me.

wenjianqiao:
One minor comment. Can you add Jira number to the PR title?

manish-sethi:
> One minor comment. Can you add Jira number to the PR title?

done

core/ledger/ledgerstorage/store.go
```diff
@@ -143,7 +143,6 @@ func (s *Store) CommitWithPvtData(blockAndPvtdata *ledger.BlockAndPvtData) error
 	}
 
 	if err := s.AddBlock(blockAndPvtdata.Block); err != nil {
-		s.pvtdataStore.Rollback()
```
cendhu, line 146: Given that we are removing the _rollback()_, having both _prepare()_ and _commit()_ may not make sense now. As already there is a huge difference in code between v1.4 and v2.0, having these additional differences might create more confusion in the future. Why not pick this PR https://github.com/hyperledger/fabric/commit/b33e5aaea8ebb618ff9881d7f6ac5edd4709625f#diff-596a7deb3a7e702099aac438aede9984 to v1.4 to reduce the code differences? If not, are we trying to minimize the changes made to v1.4?

mastersingh24: generally, we try to minimize changes where possible and sometimes this means we can't cherry pick a change but have to "re-implement" part of it.

manish-sethi: Yes, minimizing the changes was the main motive. In addition, on the code difference part, 1.4 and 2.0 looks significantly different now in many aspects and I have accepted the mode of maintaining two code bases and try to make changes in 1.4 branch such that it looks closer to earlier releases of 1.4.x as opposed to close to 2.0

cendhu: Thanks, @mastersingh24 and @manish-sethi for the clarification. The changes look good to me.

denyeart:
I've been slow to review because I wanted to test some error, rollback, and reset scenarios myself before merging into LTS stream. My tests were successful. Merging!

denyeart (approved):

PR #549, opened 2020-01-24
[FAB-17439] Expose peer's MSPID to chaincode as an env var CORE_PEER_MSPID

caod123:
#### Type of change
- New feature

#### Description

See [FAB-17439](https://jira.hyperledger.org/browse/FAB-17439)

### DIFF
--- a/core/container/dockercontroller/dockercontroller.go
+++ b/core/container/dockercontroller/dockercontroller.go
@@ -110,6 +110,7 @@ type DockerVM struct {
 	NetworkMode     string
 	PlatformBuilder PlatformBuilder
 	LoggingEnv      []string
+	MSPID           string
 }
 
 // HealthCheck checks if the DockerVM is able to communicate with the Docker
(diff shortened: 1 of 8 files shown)


core/container/dockercontroller/dockercontroller.go
```diff
@@ -247,7 +248,7 @@ const (
 	TLSClientRootCertFile string = "/etc/hyperledger/fabric/peer.crt"
 )
 
-func (vm *DockerVM) GetEnv(ccid string, tlsConfig *ccintf.TLSConfig) []string {
+func (vm *DockerVM) GetEnv(ccid string, tlsConfig *ccintf.TLSConfig, mspid string) []string {
```
lindluni: Haven't you already wired the MSPID to the DockereVM, no need to pass it into the function, can pull it right off of `vm`

caod123: good catch thanks!

integration/externalbuilders/binary/bin/run
```diff
@@ -20,6 +20,7 @@ export CORE_PEER_TLS_ENABLED="true"
 export CORE_TLS_CLIENT_CERT_FILE="$ARTIFACTS/client.crt"
 export CORE_TLS_CLIENT_KEY_FILE="$ARTIFACTS/client.key"
 export CORE_PEER_TLS_ROOTCERT_FILE="$ARTIFACTS/root.crt"
+export CORE_PEER_LOCALMSPID="$(jq -r .mspid "$ARTIFACTS/chaincode.json")"
```
jyellick, line 23: I'm not opposed to adding this line here -- though it's odd to add it only in the binary external builder and not in the golang one, especially since no chaincode (as yet) consumes this variable.

jyellick (approved):

PR #550, opened 2020-01-24
[FAB-17303] Set defaults for RequiredPeerCount and MaxPeerCount for implicit and explicit collections

caod123:
# Type of change

- Test update

#### Description

Per discussion in [FAB-17303](https://jira.hyperledger.org/browse/FAB-17303) we've decided to default `maxPeerCount: 1` and `requiredPeerCount: 0` for implicit collections and explicit collection configs that don't specify the properties.


denyeart (approved):
looks good, and also tested end-to-end with success.

PR #552, opened 2020-01-25
[FAB-17374] chaincode service integration test

muralisrini:
#### Type of change

-Test update

#### Description

Adds integration test to fabric using fabric-chaincode-go change from
https://github.com/hyperledger/fabric-chaincode-go/pull/19.

The test can be followed via integration/e2e/chaincode_service_test.go
which was adapted from the "basic solo.." integration test but adds
support for
* controlling starting the chaincode
* generating connection.json

The goal is to provide a basis for creating other chaincode service
tests in future.

Signed-off-by: muralisr <srinivasan.muralidharan99@gmail.com>


lindluni:
In general we've been trying to do better at the handling of input artifacts to tests. In the case of the private keys and certificates we've been trying to generate them during the test where possible. When not possible we should at least document in the code how those certificates and keys were generated, even if they were simply generated with `openssl`, we should document the commands used to create them

muralisrini:
I just used existing TLS artifacts from core/comm/testdata.  Didn't generate tls on the fly (with openssl or other dependencies) as this was just well understood standard TLS and not fabric specific crypto (such as MSPs).  But if we are already using openssl or something in integration test suite for TLS generation (I didn't want to use cryptogen for just generating TLS) I can try and automate that process (suggest in another PR though for expediency ?)

integration/nwo/config.go
```diff
@@ -9,15 +9,16 @@ package nwo
...
+	Organizations    []*Organization `yaml:"organizations,omitempty"`
+	Consortiums      []*Consortium   `yaml:"consortiums,omitempty"`
+	SystemChannel    *SystemChannel  `yaml:"system_channel,omitempty"`
+	Channels         []*Channel      `yaml:"channels,omitempty"`
+	Consensus        *Consensus      `yaml:"consensus,omitempty"`
+	Orderers         []*Orderer      `yaml:"orderers,omitempty"`
+	Peers            []*Peer         `yaml:"peers,omitempty"`
+	Profiles         []*Profile      `yaml:"profiles,omitempty"`
+	Templates        *Templates      `yaml:"templates,omitempty"`
+	ChaincodeServers []*Chaincode    `yaml:"chaincode_servers,omitempty"`
```
sykesm, line 21: This is not part of the network configuration but specific to a small subset of tests. There's no need to have this here.

integration/nwo/chaincode_client_template.go
```diff
@@ -0,0 +1,18 @@
+/*
```
sykesm, line 1: This really shouldn't be a template as it's not part of the configuration, it's part of the requirement for a single chaincode implementation.

Also, once `ChaincodeServerAddress` is removed from the config (since it shouldn't be there), there will be nothing to template against.

integration/chaincode/extcc/main.go
```diff
@@ -0,0 +1,263 @@
+/*
```
sykesm, line 1: Why do we need yet another copy of the simple chaincode? Can't we just create a main that reuses the logic?

Just adding a subdirectory where we separate the launch from the logic (eg. `integration/chaincode/simple/cmd/{server,client}`) should be sufficient.

integration/chaincode/extcc/main.go
```diff
@@ -0,0 +1,263 @@
...
+import (
+	"fmt"
+	"os"
+	"strconv"
+
+	"github.com/hyperledger/fabric-chaincode-go/shim"
+	pb "github.com/hyperledger/fabric-protos-go/peer"
+)
+
+var cert string = `-----BEGIN CERTIFICATE-----
```
sykesm, line 18: More statically generated fixtures. We need to stop doing this and I know I've mentioned it in the past as area we are focusing on.

If we have fixtures that are quick to generate, we prefer they are generated in the test; if complicated to generate, we fall back to `go:generate` commands and/or scripts. If neither of those work, we expecte explicit documentation about how they were generated.

integration/e2e/chaincode_service_test.go
```diff
@@ -0,0 +1,175 @@
...
+			RunRespondWith(network, orderer, peer, "testchannel")
+
+			By("waiting for DeliverFiltered stats to be emitted")
+			metricsWriteInterval := 5 * time.Second
+			Eventually(datagramReader, 2*metricsWriteInterval).Should(gbytes.Say("stream_request_duration.protos_Deliver.DeliverFiltered."))
+
+			CheckPeerStatsdStreamMetrics(datagramReader.String())
+			CheckPeerStatsdMetrics(datagramReader.String(), "org1_peer0")
+			CheckPeerStatsdMetrics(datagramReader.String(), "org2_peer0")
+			CheckOrdererStatsdMetrics(datagramReader.String(), "ordererorg_orderer")
```
sykesm, line 168: Why are we copying all of these assertions?

integration/e2e/chaincode_service_test.go
```diff
@@ -0,0 +1,175 @@
...
+
+			By("setting up a channel from a base profile")
+			additionalPeer := network.Peer("Org2", "peer0")
+			network.CreateChannel("baseprofilechannel", orderer, peer, additionalPeer)
```
sykesm, line 172: What does this have to do with the external chaincode server?

integration/e2e/chaincode_service_test.go
```diff
@@ -0,0 +1,175 @@
...
+			// start external chaincode service
+			ccrunner := network.ChaincodeServerRunner(extcc, []string{extcc.PackageID, address})
+			ccserver = ifrit.Invoke(ccrunner)
+			Eventually(ccserver.Ready(), network.EventuallyTimeout).Should(BeClosed())
+
+			// init the chaincode, if required
+			if extcc.InitRequired {
+				By("initing the chaincode")
+				nwo.InitChaincode(network, "testchannel", orderer, extcc, network.PeersWithChannel("testchannel")...)
+			}
```
sykesm, line 153: This is hard-coded above.

integration/nwo/deploy.go
```diff
@@ -135,6 +145,8 @@ func PackageAndInstallChaincode(n *Network, chaincode Chaincode, peers ...*Peer)
 	if _, err := os.Stat(chaincode.PackageFile); os.IsNotExist(err) {
 		switch chaincode.Lang {
 		case "binary":
+			fallthrough
+		case "extcc":
```
sykesm, line 149: Please avoid using `fallthrough`. In this case, just list the two cases:
```
case "binary", "extcc":
```

integration/nwo/network.go
```diff
@@ -144,9 +144,11 @@ type Network struct {
 	StatsdEndpoint     string
 	ClientAuthRequired bool
 
-	PortsByBrokerID  map[string]Ports
-	PortsByOrdererID map[string]Ports
-	PortsByPeerID    map[string]Ports
+	PortsByBrokerID          map[string]Ports
+	PortsByOrdererID         map[string]Ports
+	PortsByPeerID            map[string]Ports
+	PortsByChaincodeServerID map[string]Ports
```
sykesm, line 150: The external chaincode is not part of the network proper; if you need to allocate a port to listen on, just call `network.ReservePort` to reserve the port for the chaincode instance(s) you need.

There's no need to change the model.

integration/nwo/deploy.go
```diff
@@ -46,6 +46,11 @@ type Chaincode struct {
 	Label               string
 	SignaturePolicy     string
 	ChannelConfigPolicy string
+	Server              bool
+}
+
+func (c *Chaincode) ID() string {
+	return c.Name
```
sykesm, line 53: This accessor should not be needed. afaict, the only call is when you're reserving ports and that should be part of the test.

integration/nwo/commands/chaincodeserver.go
```diff
@@ -0,0 +1,23 @@
...
+	ServerArgs []string
+}
+
+func (c ChaincodeServerStart) SessionName() string {
+	return fmt.Sprintf("%s-server", c.PackageID)
+}
+
+func (c ChaincodeServerStart) Args() []string {
+	return c.ServerArgs
+}
```
sykesm, line 23: This is not a command that is part of fabric; you should use a session specific to your test to launch your chaincode and shut it down.

integration/nwo/network.go
```diff
@@ -156,6 +158,7 @@ type Network struct {
 	Profiles         []*Profile
 	Consortiums      []*Consortium
 	Templates        *Templates
+	ChaincodeServers []*Chaincode
```
sykesm, line 161: This should not be part of the network model.

integration/nwo/network.go
```diff
@@ -173,19 +176,22 @@ func New(c *Config, rootDir string, client *docker.Client, startPort int, compon
...
-		Peers:         c.Peers,
-		SystemChannel: c.SystemChannel,
-		Channels:      c.Channels,
-		Profiles:      c.Profiles,
-		Consortiums:   c.Consortiums,
-		Templates:     c.Templates,
+		PortsByBrokerID:          map[string]Ports{},
+		PortsByOrdererID:         map[string]Ports{},
+		PortsByPeerID:            map[string]Ports{},
+		PortsByChaincodeServerID: map[string]Ports{},
```
sykesm, line 183: This should not be part of the network model.

integration/nwo/network.go
```diff
@@ -173,19 +176,22 @@ func New(c *Config, rootDir string, client *docker.Client, startPort int, compon
...
+		Organizations:    c.Organizations,
+		Consensus:        c.Consensus,
+		Orderers:         c.Orderers,
+		Peers:            c.Peers,
+		SystemChannel:    c.SystemChannel,
+		Channels:         c.Channels,
+		Profiles:         c.Profiles,
+		Consortiums:      c.Consortiums,
+		Templates:        c.Templates,
+		ChaincodeServers: c.ChaincodeServers,
```
sykesm, line 194: This should not be part of the network model.

integration/nwo/network.go
```diff
@@ -633,6 +647,9 @@ func (n *Network) GenerateConfigTree() {
 	for _, p := range n.Peers {
 		n.GenerateCoreConfig(p)
 	}
+	for _, cc := range n.ChaincodeServers {
+		n.GenerateChaincodeConnectionConfig(cc)
+	}
```
sykesm, line 652: This should not be part of the network model; these are generated by the external builder implementations.

integration/nwo/network.go
```diff
@@ -1438,13 +1455,14 @@ type PortName string
...
-	OperationsPort PortName = "Operations"
-	ClusterPort    PortName = "Cluster"
+	ChaincodePort       PortName = "Chaincode"
+	EventsPort          PortName = "Events"
+	HostPort            PortName = "HostPort"
+	ListenPort          PortName = "Listen"
+	ProfilePort         PortName = "Profile"
+	OperationsPort      PortName = "Operations"
+	ClusterPort         PortName = "Cluster"
+	ChaincodeServerPort PortName = "Cluster"
```
sykesm, line 1465: This should not be part of the network model.

integration/nwo/network.go
```diff
@@ -1464,6 +1482,11 @@ func BrokerPortNames() []PortName {
 	return []PortName{HostPort}
 }
 
+// PeerPortNames returns the list of ports that need to be reserved for a Peer.
+func ChaincodeServerPortNames() []PortName {
+	return []PortName{ChaincodeServerPort}
+}
+
```
sykesm, line 1489: This should not be part of the network model.

integration/nwo/network.go
```diff
@@ -1473,6 +1496,18 @@ func (n *Network) BrokerAddresses(portName PortName) []string {
...
+	return fmt.Sprintf("127.0.0.1:%d", n.ChaincodeServerPort(ccsrv))
+}
+
+// ChaincodeServerPort returns the named port reserved for the chaincode service.
+func (n *Network) ChaincodeServerPort(ccSrv *Chaincode) uint16 {
+	ccSrvPorts := n.PortsByChaincodeServerID[ccSrv.ID()]
+	Expect(ccSrvPorts).NotTo(BeNil())
+	return ccSrvPorts[ChaincodeServerPort]
+}
+
```
sykesm, line 1510: This should not be part of the network model; the necessary port can be reserved in the test.

integration/nwo/network.go
```diff
@@ -1517,6 +1552,21 @@ func (n *Network) nextColor() string {
...
+	return ginkgomon.New(ginkgomon.Config{
+		AnsiColorCode:     n.nextColor(),
+		Name:              c.PackageID,
+		Command:           cmd,
+		StartCheck:        `Starting chaincode .* at .*`,
+		StartCheckTimeout: 15 * time.Second,
+	})
+}
+
```
sykesm, line 1569: This should not be part of the network; this is part of the test... The network does not launch server chaincode; server chaincode is launched and peer connects to it.

integration/nwo/network.go
```diff
@@ -1607,3 +1657,34 @@ func (n *Network) GenerateCoreConfig(p *Peer) {
...
+	t, err := template.New("peer").Funcs(template.FuncMap{
+		"Chaincode": func() *Chaincode { return chaincode },
+	}).Parse(n.Templates.ChaincodeConnectionTemplate())
+	Expect(err).NotTo(HaveOccurred())
+
+	time.Sleep(15 * time.Second)
+	pw := gexec.NewPrefixedWriter(fmt.Sprintf("[%s#core.yaml] ", chaincode.ID()), ginkgo.GinkgoWriter)
+	err = t.Execute(io.MultiWriter(core, pw), n)
+	Expect(err).NotTo(HaveOccurred())
+}
```
sykesm, line 1690: These do not belong in the network. Your test should explicitly create the `connection.json` that you need.

integration/nwo/standard_networks.go
```diff
@@ -100,6 +104,18 @@ func FullSolo() *Config {
...
+		Expect(cc.Server).To(Equal(true))
+	}
+
+	config := BasicSolo()
+
+	config.ChaincodeServers = ccServs
+
+	return config
+}
+
```
sykesm, line 118: There is no need for a template; this should be removed.

integration/nwo/templates.go
```diff
@@ -8,10 +8,11 @@ package nwo
...
 type Templates struct {
-	ConfigTx string `yaml:"configtx,omitempty"`
-	Core     string `yaml:"core,omitempty"`
-	Crypto   string `yaml:"crypto,omitempty"`
-	Orderer  string `yaml:"orderer,omitempty"`
+	ConfigTx            string `yaml:"configtx,omitempty"`
+	Core                string `yaml:"core,omitempty"`
+	Crypto              string `yaml:"crypto,omitempty"`
+	Orderer             string `yaml:"orderer,omitempty"`
+	ChaincodeConnection string `yaml:"chaincode_connection,omitempty"`
```
sykesm, line 15: This template is not needed.

integration/nwo/templates.go
```diff
@@ -41,3 +42,10 @@ func (t *Templates) OrdererTemplate() string {
 	}
 	return DefaultOrdererTemplate
 }
+
+func (t *Templates) ChaincodeConnectionTemplate() string {
+	if t.ChaincodeConnection != "" {
+		return t.ChaincodeConnection
+	}
+	return DefaultChaincodeConnectionTemplate
+}
```
sykesm, line 51: This is not needed and should not exist.

sykesm (changes requested):
The majority of the changes to the network should not have been made. The test implementation you are creating for this function should be fairly self-contained without modifying the network abstraction.

- Port reservation for your chaincode server instance should be accomplished by reserving a port in your test.
- The `connection.json` document should be created by your test or your external builder.
- The template and port reservation logic need to be removed from the nwo package.
- The certificate and key material needed for the tests should either be generated in the tests or have explicit information on how to generate them. This is especially true of TLS related assets as they do expire.
- The chaincode looks to be another copy of SimpleChaincode. Instead of making a full duplicate, it would be preferred to simply modify the main. This would highlight the differences between chaincode as client and chaincode as server and demonstrate that the chaincode does not have to change - simply how it is launched.
- Launching a server chaincode is the responsibility of the test and not the network. As such, there is no need to create a command for it or model it.

In summary, outside of a couple of simple changes to the nwo package (ex. short-circuit when Server is true), most code should be in a suite that is specific to chaincode-as-server. That suite should not be copying unrelated assertions from e2e but focus on the behavior and function related to external chaincode servers.

muralisrini:
@sykesm the comments are largely related to isolating chaincode service setup and management to a chaincode-service related package (and moving out of nwo). I'll take a look at how this is can be organized and in the process address the comments above today/tomorrow.

sykesm:
> @sykesm the comments are largely related to isolating chaincode service setup and management to a chaincode-service related package (and moving out of nwo). I'll take a look at how this is can be organized and in the process address the comments above today/tomorrow.

Thanks. Do you want to pull the bump of fabric-chaincode-go out of this into a separate PR? (Alternatively, I can do that on your behalf.) Given the the PR checks passed, they don't seem to break legacy chaincode.

muralisrini:
If you can,  do bump up the  fabric-chaincode-go please. Thanks!

sykesm:
> If you can, do bump up the fabric-chaincode-go please. Thanks!

#563 for master
#564 for release-2.0

Please review and merge. Thanks.

sykesm:
@muralisrini, it's been a couple of weeks. Any idea when you might come back to this?

Thanks.

lindluni:
It would be really nice to get this work done, as we march closer to a new LTS release, having the integration test would be great as people start to migrate from 1.4 LTS to 2.x LTS and start using new features, so we can make sure it functions as expected moving forward.

mastersingh24:
I'm closing this as it's seen no activity in months

integration/e2e/chaincode_service_test.go
psingh-io: @sykesm Did not understand what is hardcoded between lines 150 and 153. Can you please elaborate?

sykesm: Line 137 calls `DeployChaincode`. This logic part of that function.

```
// DeployChaincode is a helper that will install chaincode to all peers that
// are connected to the specified channel, approve the chaincode on one of the
// peers of each organization in the network, commit the chaincode definition
// on the channel using one of the peers, and wait for the chaincode commit to
// complete on all of the peers. It uses the _lifecycle implementation.
// NOTE V2_0 capabilities must be enabled for this functionality to work.
func DeployChaincode(n *Network, channel string, orderer *Orderer, chaincode Chaincode, peers ...*Peer) {
	if len(peers) == 0 {
		peers = n.PeersWithChannel(channel)
	}
	if len(peers) == 0 {
		return
	}

	PackageAndInstallChaincode(n, chaincode, peers...)

	// approve for each org
	ApproveChaincodeForMyOrg(n, channel, orderer, chaincode, peers...)

	// commit definition
	CheckCommitReadinessUntilReady(n, channel, chaincode, n.PeerOrgs(), peers...)
	CommitChaincode(n, channel, orderer, chaincode, peers[0], peers...)

	// init the chaincode, if required
	if chaincode.InitRequired {
		InitChaincode(n, channel, orderer, chaincode, peers...)
	}
}
```

psingh-io: @sykesm For chaincode as an external service, chaincode package ID is required to launch the chaincode. The package id is available after deployment of the chaincode. 
The DeployChaincode method has been updated with following just before chaincode initialization in this PR:
	if chaincode.Server {
		return
	}

After the chaincode is deployed, the package id is used to launch the chaincode process and it is then initialized - hence the init call outside DeployChaincode.
This appears to be acceptable solution.

Alternatively, since package id is hash of the chaincode package file - it can / be computed before deployment and chaincode process be launched prior to deploying the chaincode. In this way Init will succeed.
Please let me know this alternate path shall be explored.

PR #559, opened 2020-01-27
[FAB-17440] Improve docs on peer lifecycle chaincode with JSON output

satota2:
This patch adds example usages with JSON output option to the commands
reference for peer lifecycle chaincode queryinstalled and querycommitted.

This patch also updates the syntax of configtxgen in the commands reference
that were not synchronized with the implementation.

Signed-off-by: Tatsuya Sato <Tatsuya.Sato@hal.hitachi.com>

#### Type of change

- Documentation update

#### Description

Some of the `peer lifecycle chaincode` subcommands support output format as JSON with `–output` option. 
Fabric admins need to use this JSON output to see the detailed information on chaincode definition
because default output shows only partial information.

On the other hand, the current command reference for `peer lifecycle chaincode` describe example usages withthe JSON output only for the `checkcommitreadiness` subcommand.

To let more Fabric admins know about the useful option, the reference should explicitly describe example usages of the output options for the remaining `queryinstalled` and ` querycommitted ` as well.

#### Related issues

https://jira.hyperledger.org/browse/FAB-17440

### DIFF
--- a/docs/source/commands/peerlifecycle.md
+++ b/docs/source/commands/peerlifecycle.md
@@ -375,6 +375,36 @@ Get installed chaincodes on peer:
 Package ID: myccv1:a7ca45a7cc85f1d89c905b775920361ed089a364e12a9b6d55ba75c965ddd6a9, Label: myccv1
 ```
 
+  * You can also use the `--output` flag to have the CLI format the output as
+    JSON.
+
+    ```
+    peer lifecycle chaincode queryinstalled --peerAddresses peer0.org1.example.com:7051 --output json
+    ```
+
+    If successful, the command will return the chaincodes you have installed as JSON.
+
+    ```
+    {
+      "installed_chaincodes": [
+        {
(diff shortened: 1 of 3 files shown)


docs/source/commands/peerlifecycle.md
```diff
@@ -375,6 +375,36 @@ Get installed chaincodes on peer:
...
+                                                            "name": "mycc",
+                                                            "version": "1"
+                                                    }
+                                            ]
+                                    }
+                            }
+                    }
+            ]
+    }
+    ```
```
lindluni: Can we decrease the size of the tabs to two spaces, something that reduces that amount of whitespace.

In other places as well where JSON blocks were added

satota2: Thank you for having reviews.
New patch decreases the size of the tabs to two spaces for their JSON blocks.

denyeart (approved):

denyeart:
@satota2 I've merged in master, but please cherry pick to release-2.0 as well so that this can get into the upcoming v2.0 readthedocs.

PR #578, opened 2020-01-28
[FAB-17448] Fix couchDB startup inconsistence in package couchDB

DereckLuo:

#### Type of change

- Bug fix

#### Description

- Fixed couchDB unit tests are not starting couchDB container correctly.
Multiple functions are missing startCouchDB function call.
#### Related issues

[FAB-17448](https://jira.hyperledger.org/browse/FAB-17448)

### DIFF
--- a/core/ledger/util/couchdb/couchdb_test.go
+++ b/core/ledger/util/couchdb/couchdb_test.go
@@ -229,7 +229,7 @@ func TestBadCouchDBInstance(t *testing.T) {
 }
 
 func TestDBCreateSaveWithoutRevision(t *testing.T) {
-
+	startCouchDB()
 	database := "testdbcreatesavewithoutrevision"
 	err := cleanup(database)
 	assert.NoError(t, err, "Error when trying to cleanup  Error: %s", err)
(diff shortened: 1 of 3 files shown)


core/ledger/util/couchdb/couchdb_test.go
```diff
@@ -631,7 +633,7 @@ func TestDBRequestTimeout(t *testing.T) {
 }
 
 func TestDBTimeoutConflictRetry(t *testing.T) {
-
+	startCouchDB()
```
lindluni, line 636: For tests that have their own CreateCouchInstance call, why do we need to start the DB instance?

DereckLuo: If I understand this correctly, **CreateCouchInstance** uses __config__ to start a couchDB instance, which inside __config__ contains address for couchDB container, the call itself does not start the container, only creates an instance. 
**startCouchDB** will start the external couchDB container if it has not been started by previous tests, and populate a global variable to store couchDB address which will be consumed by __config__.

lindluni: Ah, got it, thank you!

manish-sethi (approved):

PR #588, opened 2020-01-29
What's New updates for v2.0.0 release

denyeart:
#### Type of change

- Documentation update

#### Description

Updates to What's New doc for v2.0.0 release.

### DIFF
--- a/docs/source/whatsnew.rst
+++ b/docs/source/whatsnew.rst
@@ -1,17 +1,21 @@
-What's new in the v2.0 Beta release
-===================================
+What's new in Hyperledger Fabric v2.0
+=====================================
 
-The Beta release of Hyperledger Fabric v2.0 allows users to try out upcoming
-features of Fabric v2.0. While the v2.0 release is not yet production ready,
-the Beta is feature complete, meaning it includes all the features that are
-expected to be delivered in the final v2.0 release.
+The first Hyperledger Fabric major release since v1.0, Fabric v2.0
+delivers important new features and changes for users and operators alike,
+including support for new application and privacy patterns, enhanced
+governance around smart contracts, and new options for operating nodes.
 
-Let’s take a peek at some of the highlights of Fabric v2.0....
+What has stayed the same is the ability to upgrade network components on
+your own terms, with support for rolling upgrades from v1.4.x, and the ability
+to enable the new capabilities only when member organizations are ready.
 
-Decentralized chaincode lifecycle
----------------------------------
+Let's take a look at some of the highlights of the Fabric v2.0 release...
 
-Fabric v2.0 introduces decentralized governance for chaincode, with a new
+Decentralized governance for smart contracts
+--------------------------------------------
+
+Fabric v2.0 introduces decentralized governance for smart contracts, with a new
 process for installing a chaincode on your peers and starting it on a channel.
 The new Fabric chaincode lifecycle allows multiple organizations to come to
 agreement on the parameters of a chaincode, such as the chaincode endorsement
@@ -27,7 +31,7 @@ offers several improvements over the previous lifecycle:
   trust models (such as that of the previous lifecycle model) as well as
   decentralized models requiring a sufficient number of organizations to
   agree on an endorsement policy and other details before the chaincode
-  ecomes active on a channel.
+  becomes active on a channel.
 
 * **More deliberate chaincode upgrade process** In the previous chaincode
   lifecycle, the upgrade transaction could be issued by a single organization,
@@ -55,7 +59,7 @@ offers several improvements over the previous lifecycle:
   like to track different types of assets in their own ‘copy’ of the chaincode.
 
 * **Chaincode packages do not need to be identical across channel members**
-  Organizations can customize a chaincode for their own use case, for example
+  Organizations can extend a chaincode for their own use case, for example
   to perform different validations in the interest of their organization.
   As long as the required number of organizations endorse chaincode transactions
   with matching results, the transaction will be validated and committed to the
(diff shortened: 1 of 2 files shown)


docs/source/whatsnew.rst
```diff
@@ -71,43 +75,36 @@ only when the channel application capability is updated to v2.0.
...
+applications to ensure organizations consent to data transactions before
+they are committed to the ledger.
 
-* **Chaincode as an external service** Traditionally, chaincodes are launched
-  by and then connect to the peer. It is now possible to run chaincode as
-  an external service, for example in a Kubernetes pod, which a peer can
-  connect to and utilize for chaincode execution. See :doc:`cc_service` for more
-  information.
+* **Automated checks** As mentioned above, organizations can add automated
+  checks to chaincode functions to validate additional information before
```
joealewine, line 87: Should have a period after this, other wise you'll have a bolded word followed by an unbolded word, which will look weird.

docs/source/whatsnew.rst
```diff
@@ -71,43 +75,36 @@ only when the channel application capability is updated to v2.0.
...
+  endorsing a transaction proposal.
 
-See :doc:`cc_launcher` to learn more about the external chaincode launcher feature.
+* **Decentralized agreement** Human decisions can be modeled into a chaincode process
+  that spans multiple transactions. The chaincode may require actors from
```
joealewine, line 91: See above comment.

joealewine (changes requested):

docs/source/whatsnew.rst
denali49: You are technically correct Joe, although I think the doc reads fine because the bolded text emphasis is consistent throughout the doc when viewing the html.  If using bold emphasis on the text is meant to serve as a header to the sub-section, then perhaps instead of a period, the follow-on text should be underneath?  I am however, inclined to vote for leaving as-is.

docs/source/whatsnew.rst
```diff
@@ -55,7 +59,7 @@ offers several improvements over the previous lifecycle:
   like to track different types of assets in their own ‘copy’ of the chaincode.
 
 * **Chaincode packages do not need to be identical across channel members**
-  Organizations can customize a chaincode for their own use case, for example
+  Organizations can extend a chaincode for their own use case, for example
   to perform different validations in the interest of their organization.
   As long as the required number of organizations endorse chaincode transactions
   with matching results, the transaction will be validated and committed to the
```
pamandrejko, line 65: Is it worth additionally pointing out here that each organization can run the chaincode in the language of their choice?

docs/source/whatsnew.rst
joealewine: I hadn't noticed the other instances of this in the doc, but I would absolutely change those as well. 

Screen Shot 2020-01-29 at 9 47 11 AM

Looks wrong. Is wrong. Imo.

docs/source/whatsnew.rst
```diff
@@ -138,6 +135,33 @@ To learn more about the new private data patterns, see :doc:`private-data/privat
 for information about private data sharing and :doc:`private-data-arch` (reference information)
 for documentation about private data collection configuration and implicit collections.
```
pamandrejko: Is this section all one sentence? If so there needs to be commas. Otherwise I think we need periods after each for clause and capitalize the `f`.

docs/source/whatsnew.rst
joealewine: I don't mind the option of having the text underneath, either.

denali49: I viewed the entire doc in html.  I suggest then we change all from:
**Bold text** Follow-on text...
To:
**Bold Text**
Follow-on-text...

docs/source/whatsnew.rst
denali49: I think you have to hit the expansion pane icon on the left-hand side to expose the text that occurs in the doc prior to the lower-case 'f' you are enquiring about.

docs/source/whatsnew.rst
```diff
@@ -138,6 +135,33 @@ To learn more about the new private data patterns, see :doc:`private-data/privat
...
+
+* **Alternatives to containers** Chaincode is no longer required to be run
+  in Docker containers, and may be executed in the operator’s choice of
+  environment (including containers).
+
+* **External builder executables** An operator can provide a set of external
+  builder executables to override how the peer builds and launches chaincode.
+
+* **Chaincode as an external service** Traditionally, chaincodes are launched
+  by and then connect to the peer. It is now possible to run chaincode as
```
caod123: nit: This first sentence sounds a bit weird. Maybe `Traditionally, chaincodes are launched by and then connected to the peer.`?

docs/source/whatsnew.rst
denali49: For consistency, we should probably add a 'style guide' topic to the doc working group meeting agenda.

docs/source/whatsnew.rst
mastersingh24: > Is it worth additionally pointing out here that each organization can run the chaincode in the language of their choice?

I think we agreed / decided not to emphasize this initially

docs/source/whatsnew.rst
```diff
@@ -1,17 +1,21 @@
-What's new in the v2.0 Beta release
-===================================
+What's new in Hyperledger Fabric v2.0
+=====================================
 
-The Beta release of Hyperledger Fabric v2.0 allows users to try out upcoming
-features of Fabric v2.0. While the v2.0 release is not yet production ready,
-the Beta is feature complete, meaning it includes all the features that are
-expected to be delivered in the final v2.0 release.
+As the first Hyperledger Fabric major release since v1.0, Fabric v2.0
```
mastersingh24:
```suggestion
The first Hyperledger Fabric major release since v1.0, Fabric v2.0
```

docs/source/whatsnew.rst
```diff
@@ -1,17 +1,21 @@
...
+delivers important new features and changes to users and operators alike,
```
mastersingh24:
```suggestion
delivers important new features and changes for users and operators alike,
```

wlahti:
```suggestion
delivers important new features and changes for users and operators alike,
```
Nit: I think this reads better as "for" instead of "to" here.

docs/source/whatsnew.rst
denyeart: no, not something we want to encourage! it's a slippery slope that gets more treacherous as you go down. e.g. json marshaling may be different on each language. so, i'm trying to describe the new 'feature' in a limited way, e.g. as a way to extend a chaincode for additional organization validations.

docs/source/whatsnew.rst
```diff
@@ -138,6 +135,33 @@ To learn more about the new private data patterns, see :doc:`private-data/privat
 for information about private data sharing and :doc:`private-data-arch` (reference information)
 for documentation about private data collection configuration and implicit collections.
 
+External chaincode launcher
+---------------------------
+
+While chaincode is still run in a docker container by default in Fabric v2.0,
```
mastersingh24:
```suggestion
```

docs/source/whatsnew.rst
```diff
@@ -1,17 +1,21 @@
...
-the Beta is feature complete, meaning it includes all the features that are
-expected to be delivered in the final v2.0 release.
+As the first Hyperledger Fabric major release since v1.0, Fabric v2.0
+delivers important new features and changes to users and operators alike,
+including support for new application and privacy patterns, enhanced
+governance around smart contracts, and new options for operating nodes.
 
-Let’s take a peek at some of the highlights of Fabric v2.0....
+What has stayed the same is the ability to upgrade network components on
+your own terms, with support for rolling upgrades from v1.4.x, and ability
```
wlahti:
```suggestion
your own terms, with support for rolling upgrades from v1.4.x, and the ability
```
Nit

docs/source/whatsnew.rst
```diff
@@ -138,6 +135,33 @@ To learn more about the new private data patterns, see :doc:`private-data/privat
 for information about private data sharing and :doc:`private-data-arch` (reference information)
 for documentation about private data collection configuration and implicit collections.
 
+External chaincode launcher
+---------------------------
+
+While chaincode is still run in a docker container by default in Fabric v2.0,
+the external chaincode launcher feature empowers operators to build and launch
```
mastersingh24:
```suggestion
The external chaincode launcher feature empowers operators to build and launch
```

docs/source/whatsnew.rst
wlahti: +1. Seems worthwhile to point out this additional flexibility the new chaincode enables.

docs/source/whatsnew.rst
```diff
@@ -138,6 +135,33 @@ To learn more about the new private data patterns, see :doc:`private-data/privat
...
+chaincode with the technology of their choice.
```
mastersingh24:
```suggestion
chaincode with the technology of their choice.  Use of external builders and launchers is not required as the default behavior builds and runs chaincode in the same manner as prior releases using the Docker API.
```

docs/source/whatsnew.rst
denyeart: refectoring this sentence.

docs/source/whatsnew.rst
wlahti: +1

docs/source/whatsnew.rst
denyeart: I had tried @denali49 suggestion in the rst source, but it ends up rendered the same. I'm going to leave it as-is for now, the doc masters can work the magic in a follow-on PR if it doesn't look right.

docs/source/whatsnew.rst
denyeart: Changing it to "Traditionally, chaincodes are launched by the peer, and then connect back to the peer."

docs/source/whatsnew.rst
denyeart: changing.

docs/source/whatsnew.rst
denyeart: changing.

denyeart: changing.

docs/source/whatsnew.rst
denyeart: changing.

docs/source/whatsnew.rst
denyeart: changing.

docs/source/whatsnew.rst
denyeart: changing.

docs/source/whatsnew.rst
denyeart: changing.

docs/source/whatsnew.rst
denyeart: If we don't mention at all, I suspect people will think it is a flaw. So I'm thinking we need to make it explicit, although position it as a way to extend chaincodes in a limited way like adding organization specific validations. I think that aspect is both valuable and feasible for organizations. But not go so far as to recommend completely different chaincodes or different languages. If you still think it should be entirely removed, I can remove it though...

joealewine (approved):

PR #602, opened 2020-01-30
Fix conditional_packages processing in UT script

sykesm:
An empty conditional_packages list would result in an error in join_by
that resulted in an empty test list.

FAB-17454 #done


lindluni:
LGTM: That's a simpler fix than what I was considering to fix it

mastersingh24 (approved):

PR #608, opened 2020-01-30
[FAB-17458] check LifecycleV20 capability before getting cc info

wenjianqiao:

#### Type of change
- Bug fix

#### Description
Check LifecycleV20 capability before getting new lifecycle chaincode info
for endorsement and get legacy cc info if LifecycleV20 is not enabled.

#### Related issues
https://jira.hyperledger.org/browse/FAB-17458

### DIFF
--- a/core/chaincode/lifecycle/endorsment_info_test.go
+++ b/core/chaincode/lifecycle/endorsment_info_test.go
@@ -20,21 +20,34 @@ import (
 
 var _ = Describe("ChaincodeEndorsementInfoSource", func() {
 	var (
-		cei               *lifecycle.ChaincodeEndorsementInfoSource
-		resources         *lifecycle.Resources
-		fakeLegacyImpl    *mock.LegacyLifecycle
-		fakePublicState   MapLedgerShim
-		fakeQueryExecutor *mock.SimpleQueryExecutor
-		fakeCache         *mock.ChaincodeInfoCache
-		testInfo          *lifecycle.LocalChaincodeInfo
-		builtinSCCs       scc.BuiltinSCCs
+		cei                     *lifecycle.ChaincodeEndorsementInfoSource
+		resources               *lifecycle.Resources
+		fakeLegacyImpl          *mock.LegacyLifecycle
+		fakePublicState         MapLedgerShim
+		fakeQueryExecutor       *mock.SimpleQueryExecutor
+		fakeCache               *mock.ChaincodeInfoCache
+		fakeChannelConfigSource *mock.ChannelConfigSource
+		fakeChannelConfig       *mock.ChannelConfig
+		fakeAppConfig           *mock.ApplicationConfig
+		fakeCapabilities        *mock.ApplicationCapabilities
+		testInfo                *lifecycle.LocalChaincodeInfo
+		builtinSCCs             scc.BuiltinSCCs
 	)
 
 	BeforeEach(func() {
 		fakeLegacyImpl = &mock.LegacyLifecycle{}
+		fakeChannelConfigSource = &mock.ChannelConfigSource{}
+		fakeChannelConfig = &mock.ChannelConfig{}
+		fakeAppConfig = &mock.ApplicationConfig{}
+		fakeCapabilities = &mock.ApplicationCapabilities{}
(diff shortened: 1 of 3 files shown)


wenjianqiao:
The change has a side effect, causing integration test failure at https://github.com/hyperledger/fabric/blob/master/integration/lifecycle/interop_test.go#L214.
This is because that the endorsed tx in v1_4_2 capability does not have _lifecycle in readset any more. As a result, no read conflict during commit. 

In the above test, since the chaincode has been upgraded using new lifecycle (v2_0 capability), the tx commit (endorsed in v1_4_2 capability before cc upgrade) should return a validation error. Will work offline to solve the problem and update the PR.

Update: talked to @jyellick. See below comment https://github.com/hyperledger/fabric/pull/608#discussion_r374171141

integration/lifecycle/interop_test.go
```diff
@@ -209,9 +209,11 @@ var _ = Describe("Release interoperability", func() {
 			}
 			nwo.DeployChaincode(network, "testchannel", orderer, chaincode)
 
-			By("committing the old transaction, expecting to hit an MVCC conflict")
+			By("committing the old transaction")
+			// FAB-17458: because the endorsed tx doesn't have _lifecycle in read set,
+			// it has no read conflict with the cc upgrade via new lifecycle.
 			err = CommitTx(network, env, endorsers[0], deliveryClient, ordererClient, userSigner, txid)
-			Expect(err).To(MatchError(ContainSubstring("transaction invalidated with status (MVCC_READ_CONFLICT)")))
+			Expect(err).NotTo(HaveOccurred())
```
wenjianqiao, line 216: In this test, a tx was endorsed in V1_4_2 capability which didn't have _lifecycle in read set. Then the app capability was upgraded to V2_0 and chaincode was upgraded using new lifecycle. Due to no read conflict (the old tx was using lscc and the cc upgrade was using _lifecycle), the old tx was validated and committed. 
Talked to @jyellick, this behavior is acceptable, but open for discussion if there is any concern.

wenjianqiao: One concern is that an inflight tx on old business logic may be committed after the cc is already upgraded with new business logic. 
To keep the existing behavior (MVCC_READ_CONFLICT), one idea is to add 'lscc' to the read set when upgrading cc using new lifecycle. However, it is not preferred because we want to get rid off 'lscc'.

denyeart: @ale-linux worked on the transition from old to new lifecycle, so would be good to get his opinion as well.

denyeart: We discussed further and concluded as Jason did, that the behavior is acceptable. If application really wants to ensure that the latest chaincode is used, they should maintain their own version key in the user chaincode state, read it for every transaction, and increment it upon deploying the new chaincode. This way, if a transaction endorsed at prior version attempts to be committed with a new chaincode already in place, it will fail MVCC validation check. This would work on old lifecycle, new lifecycle, as well as when upgrading from old chaincode to new chaincode.

core/chaincode/lifecycle/endorsment_info_test.go
```diff
@@ -243,4 +256,52 @@ var _ = Describe("ChaincodeEndorsementInfoSource", func() {
...
+		})
+
+		It("wraps and returns the error", func() {
+			res, err := cei.ChaincodeEndorsementInfo("channel-id", "cc-name", fakeQueryExecutor)
+			Expect(err).To(MatchError("fake-error"))
+			Expect(res).To(Equal(&lifecycle.ChaincodeEndorsementInfo{
+				Version:           "legacy-version",
+				EndorsementPlugin: "legacy-plugin",
+				ChaincodeID:       "legacy-id",
+			}))
```
denyeart: Does it really return the error and the result?
Also check the It - "wraps and returns the error"

denyeart (approved):
Looks good to me, but I'd like one more review please. I've nominated Ale, Manish, or Jason.

core/chaincode/lifecycle/endorsment_info_test.go
wenjianqiao: Done

jyellick (approved):

wenjianqiao:

lindluni:

PR #615, opened 2020-01-31
Update bootstrap.sh

olelbis:
Wrong CA_BINARY_FILE, ${ARCH} and ${CA_VERSION} must be separated by dot intead of dash if CA version is 1.4.4

#### Type of change

- Bug fix

#### Description

The binary CA file is not downloaded because there is probably a typo in a file name.
To overcome this problem it is necessary to change the separator  between the architecture and version from dash to dot

### DIFF
--- a/scripts/bootstrap.sh
+++ b/scripts/bootstrap.sh
@@ -165,7 +165,7 @@ else
 fi
 
 BINARY_FILE=hyperledger-fabric-${ARCH}-${VERSION}.tar.gz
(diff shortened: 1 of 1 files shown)


lindluni:
Thank you for the contribution. The issue is actually in our packaging process, I uploaded new binaries today and it has the wrong format in the tarball output. I will correct the binaries in GitHub to fix this situation.

Thanks for bringing it to our attention

lindluni:
I fixed the names of the binaries, you should be good now

PR #626, opened 2020-02-03
[FAB-17059]: Change collection membership eligibility checks to only compare mspID

caod123:
#### Type of change

- Bug fix

#### Description

This attempts to fix a bug where eligibility evaluation was failing when ca certs were rolled

#### Additional details

See [FAB-17059](https://jira.hyperledger.org/browse/FAB-17059)

/cc @ale-linux

### DIFF
--- a/core/common/privdata/membershipinfo.go
+++ b/core/common/privdata/membershipinfo.go
@@ -16,27 +16,47 @@ var logger = flogging.MustGetLogger("common.privdata")
 
 // MembershipProvider can be used to check whether a peer is eligible to a collection or not
 type MembershipProvider struct {
+	mspID                       string
 	selfSignedData              common.SignedData
 	IdentityDeserializerFactory func(chainID string) msp.IdentityDeserializer
 }
 
 // NewMembershipInfoProvider returns MembershipProvider
-func NewMembershipInfoProvider(selfSignedData common.SignedData, identityDeserializerFunc func(chainID string) msp.IdentityDeserializer) *MembershipProvider {
-	return &MembershipProvider{selfSignedData: selfSignedData, IdentityDeserializerFactory: identityDeserializerFunc}
+func NewMembershipInfoProvider(mspID string, selfSignedData common.SignedData,
+	identityDeserializerFunc func(chainID string) msp.IdentityDeserializer) *MembershipProvider {
+	return &MembershipProvider{mspID: mspID, selfSignedData: selfSignedData, IdentityDeserializerFactory: identityDeserializerFunc}
 }
 
 // AmMemberOf checks whether the current peer is a member of the given collection config.
 // If getPolicy returns an error, it will drop the error and return false - same as a RejectAll policy.
-func (m *MembershipProvider) AmMemberOf(channelName string, collectionPolicyConfig *common.CollectionPolicyConfig) (bool, error) {
+// It is used when a chaincode is upgraded to see if the peer's org has become eligible after	a collection
+// change.
+func (m *MembershipProvider) AmMemberOf(channelName string,
+	collectionPolicyConfig *common.CollectionPolicyConfig) (bool, error) {
 	deserializer := m.IdentityDeserializerFactory(channelName)
+
+	// Do a simple check to see if the mspid matches any principal identities in the SignaturePolicy - FAB-17059
+	if collectionPolicyConfig.GetSignaturePolicy() != nil {
+		memberOrgs := getMemberOrgs(collectionPolicyConfig.GetSignaturePolicy().GetIdentities(), deserializer)
+
+		for _, org := range memberOrgs {
+			if m.mspID == org {
+				return true, nil
+			}
+		}
+	}
+
+	// Fall back to default access policy evaluation otherwise
 	accessPolicy, err := getPolicy(collectionPolicyConfig, deserializer)
 	if err != nil {
 		// drop the error and return false - same as reject all policy
 		logger.Errorf("Reject all due to error getting policy: %s", err)
 		return false, nil
 	}
+
 	if err := accessPolicy.Evaluate([]*common.SignedData{&m.selfSignedData}); err != nil {
 		return false, nil
 	}
+
 	return true, nil
 }
--- a/gossip/privdata/coordinator.go
+++ b/gossip/privdata/coordinator.go
@@ -127,6 +127,7 @@ type Support struct {
 }
 
 type coordinator struct {
+	mspID          string
 	selfSignedData common.SignedData
 	Support
 	transientBlockRetention        uint64
@@ -142,9 +143,10 @@ type CoordinatorConfig struct {
 }
 
 // NewCoordinator creates a new instance of coordinator
-func NewCoordinator(support Support, selfSignedData common.SignedData, metrics *metrics.PrivdataMetrics,
+func NewCoordinator(mspID string, support Support, selfSignedData common.SignedData, metrics *metrics.PrivdataMetrics,
 	config CoordinatorConfig) Coordinator {
 	return &coordinator{Support: support,
+		mspID:                          mspID,
 		selfSignedData:                 selfSignedData,
 		transientBlockRetention:        config.TransientBlockRetention,
 		metrics:                        metrics,
@@ -823,7 +825,15 @@ func (c *coordinator) accessPolicyForCollection(chdr *common.ChannelHeader, name
 }
 
 // isEligible checks if this peer is eligible for a given CollectionAccessPolicy
+// It is used upon commit to determine if the peer is eligible to retrieve/persist
+// the private data collection data for a given transaction.
 func (c *coordinator) isEligible(ap privdata.CollectionAccessPolicy, namespace string, col string) bool {
+	// Simple check to see if mspID part of collection's MemberOrgs list - FAB-17059
+	if util.Contains(c.mspID, ap.MemberOrgs()) {
+		return true
+	}
+
+	// If not part of list fall back to default policy evaluation logic
 	filt := ap.AccessFilter()
 	eligible := filt(c.selfSignedData)
 	if !eligible {
--- a/gossip/service/gossip_service.go
+++ b/gossip/service/gossip_service.go
@@ -294,7 +294,9 @@ func (g *gossipServiceImpl) InitializeChannel(chainID string, oac OrdererAddress
 		SkipPullingInvalidTransactions: viper.GetBool("peer.gossip.pvtData.skipPullingInvalidTransactionsDuringCommit"),
 	}
 
-	coordinator := privdata2.NewCoordinator(privdata2.Support{
+	selfSignedData := g.createSelfSignedData()
+	mspID := string(g.secAdv.OrgByPeerIdentity(selfSignedData.Identity))
+	coordinator := privdata2.NewCoordinator(mspID, privdata2.Support{
 		ChainID:            chainID,
 		CollectionStore:    support.Cs,
 		Validator:          support.Validator,
@@ -302,7 +304,7 @@ func (g *gossipServiceImpl) InitializeChannel(chainID string, oac OrdererAddress
 		Committer:          support.Committer,
 		Fetcher:            fetcher,
 		CapabilityProvider: support.CapabilityProvider,
-	}, g.createSelfSignedData(), g.metrics.PrivdataMetrics, coordinatorConfig)
+	}, selfSignedData, g.metrics.PrivdataMetrics, coordinatorConfig)
 
 	reconcilerConfig := privdata2.GetReconcilerConfig()
 	var reconciler privdata2.PvtDataReconciler
--- a/integration/pvtdata/pvtdata_test.go
+++ b/integration/pvtdata/pvtdata_test.go
@@ -12,10 +12,12 @@ import (
 	"fmt"
 	"io/ioutil"
 	"os"
+	"os/exec"
 	"path/filepath"
 	"strings"
 	"syscall"
 
+	"github.com/gogo/protobuf/proto"
 	. "github.com/onsi/ginkgo"
 	. "github.com/onsi/gomega"
 
@@ -23,7 +25,10 @@ import (
 	"github.com/hyperledger/fabric/core/ledger/util"
 	"github.com/hyperledger/fabric/integration/nwo"
 	"github.com/hyperledger/fabric/integration/nwo/commands"
+	"github.com/hyperledger/fabric/msp"
 	"github.com/hyperledger/fabric/protos/common"
+	mspp "github.com/hyperledger/fabric/protos/msp"
+	"github.com/hyperledger/fabric/protos/utils"
 	"github.com/onsi/gomega/gbytes"
 	"github.com/onsi/gomega/gexec"
 	"github.com/tedsuo/ifrit"
@@ -99,6 +104,7 @@ var _ bool = Describe("PrivateData", func() {
 			process       ifrit.Process
 			orderer       *nwo.Orderer
 			expectedPeers []*nwo.Peer
+			peerProcesses map[string]ifrit.Process
 		)
 
 		BeforeEach(func() {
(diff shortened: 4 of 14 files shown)


gossip/service/gossip_service.go
```diff
@@ -111,6 +111,7 @@ type gossipServiceImpl struct {
 	lock            sync.RWMutex
 	mcs             api.MessageCryptoService
 	peerIdentity    []byte
+	mspID           string
```
yacovm: you can extract the MSPID by taking the `peerIdentity` and passing into the `secAdv` that is right below this `mspID` field and casting it to string.

caod123: Thanks, I was trying to find a way of extracting the mspID without having to pass it down the entire stack. Any other thoughts about this approach? I tried to ask @ale-linux but haven't heard back yet.

caod123: Also that covers extracting the mspID for the gossip path -> coordinator. Is there an easier way of extracting from the membership update path via the `membershipInfoProvider`? I'd like to entirely do away with the `selfSignedData` in both paths since I think we're only using them for the existing policy evaluation logic, but I guess for now we can keep them for the sake of falling back in case the new method doesn't work.

yacovm: You can apply the same function of the `secAdv` to the Identity in the self signed data, of course

caod123: That works for the gossip path since it has a `secAdv` but the `membershipInfoProvider` only takes an `IdentityDeserializerFactory` and `SignedData`

peer/node/start.go
```diff
@@ -871,7 +873,7 @@ func secureDialOpts() []grpc.DialOption {
 // 2. Init the message crypto service;
 // 3. Init the security advisor;
 // 4. Init gossip related struct.
-func initGossipService(policyMgr policies.ChannelPolicyManagerGetter, metricsProvider metrics.Provider,
+func initGossipService(mspID string, policyMgr policies.ChannelPolicyManagerGetter, metricsProvider metrics.Provider,
```
yacovm: why do you need to put here this mspID in this `initGossipService` ? 

You have the identity passed (`serializedIdentity`) and below you also have the `secAdv` as well.

caod123: ah oops I missed this one, I meant to remove all of the passed mspIDs in the gossip path after we discussed the `secAdv` but forgot about this one. Thanks for catching it!

core/common/privdata/membershipinfo.go
```diff
@@ -7,36 +7,98 @@ SPDX-License-Identifier: Apache-2.0
...
+func (m *MembershipProvider) AmMemberOf(channelName string,
+	collectionPolicyConfig *common.CollectionPolicyConfig) (bool, error) {
 	deserializer := m.IdentityDeserializerFactory(channelName)
+
+	// Do a simple check to see if the mspid matches any principal identities in the SignaturePolicy
+	if collectionPolicyConfig.GetSignaturePolicy() != nil {
+		memberOrgs, err := getMemberOrgs(collectionPolicyConfig.GetSignaturePolicy().GetIdentities(), deserializer)
+		if err != nil {
+			logger.Errorf("Failed getting member orgs from MSP Principals: %s", err)
+			return false, nil
```
ale-linux: Why don't we fall back here?

core/common/privdata/membershipinfo.go
```diff
@@ -7,36 +7,98 @@ SPDX-License-Identifier: Apache-2.0
...
+		}
+		for _, member := range memberOrgs {
+			if m.mspID == member {
+				return true, nil
+			}
+		}
+		return false, nil
```
ale-linux: and here? In these 2 cases I would just let `getPolicy` make the decision for us

ale-linux (changes requested):
Thanks Danny! One thing that I'm still missing is why the two separate paths (one through AmMemberOf and the other through isEligible. I would also add some additional comments in the code explaining why we're doing what we're doing. Finally, we'll need some kind of integration tests for this particular scenario.

core/common/privdata/membershipinfo.go
```diff
@@ -7,36 +7,98 @@ SPDX-License-Identifier: Apache-2.0
...
 	if err := accessPolicy.Evaluate([]*common.SignedData{&m.selfSignedData}); err != nil {
 		return false, nil
 	}
+
 	return true, nil
 }
+
+// Returns a list of member orgs from a list of MSPPrincipals
+func getMemberOrgs(identities []*mspp.MSPPrincipal, deserializer msp.IdentityDeserializer) ([]string, error) {
+	memberOrgs := []string{}
```
C0rWin: a nit, but doesn't it make sense to have here a map from string to bool/struct{}? this should make your later check for inclusion easier.

caod123: Good call thanks, I'll make the change

core/common/privdata/membershipinfo.go
```diff
@@ -7,36 +7,98 @@ SPDX-License-Identifier: Apache-2.0
...
-	return &MembershipProvider{selfSignedData: selfSignedData, IdentityDeserializerFactory: identityDeserializerFunc}
+func NewMembershipInfoProvider(mspID string, selfSignedData common.SignedData,
+	identityDeserializerFunc func(chainID string) msp.IdentityDeserializer) *MembershipProvider {
+	return &MembershipProvider{mspID: mspID, selfSignedData: selfSignedData, IdentityDeserializerFactory: identityDeserializerFunc}
 }
 
 // AmMemberOf checks whether the current peer is a member of the given collection config.
 // If getPolicy returns an error, it will drop the error and return false - same as a RejectAll policy.
-func (m *MembershipProvider) AmMemberOf(channelName string, collectionPolicyConfig *common.CollectionPolicyConfig) (bool, error) {
+func (m *MembershipProvider) AmMemberOf(channelName string,
```
denyeart, line 34: AmMemberOf is used when a chaincode is upgraded, to see if the peer's org has become eligible after a collection change. May want to indicate this in the comment.

core/common/privdata/membershipinfo.go
```diff
@@ -7,36 +7,98 @@ SPDX-License-Identifier: Apache-2.0
...
 	}
+
 	if err := accessPolicy.Evaluate([]*common.SignedData{&m.selfSignedData}); err != nil {
 		return false, nil
 	}
+
 	return true, nil
 }
+
+// Returns a list of member orgs from a list of MSPPrincipals
```
denyeart: Comment should start with "getMemberOrgs" right?

gossip/privdata/coordinator.go
```diff
@@ -824,6 +826,14 @@ func (c *coordinator) accessPolicyForCollection(chdr *common.ChannelHeader, name
 
 // isEligible checks if this peer is eligible for a given CollectionAccessPolicy
 func (c *coordinator) isEligible(ap privdata.CollectionAccessPolicy, namespace string, col string) bool {
+	// Simple check to see if mspID part of collection's MemberOrgs list
```
denyeart: isEligible is used upon commit to determine if the peer is eligible to retrieve/persist the private data collection data for a given transaction. May want to indicate this in the comment.

denyeart (commented):
Agree with the other feedback, let's add as many comments as we can so that the behavior becomes clear.

denyeart:
I've tested in an e2e scenario with some additional debug, and confirmed the new simple checks are working as expected on the private data requester side, but not called on the private data provider side (provider will continue to use the policy evaluation).

caod123:
The integration test ended up being fairly complicated. Please let me know if you have questions or concerns regarding the reproduction case. If desired I can squash into 1 commit. Once this PR is merged I'll port to 2.1

core/common/privdata/membershipinfo.go
```diff
@@ -7,36 +7,97 @@ SPDX-License-Identifier: Apache-2.0
...
-	return &MembershipProvider{selfSignedData: selfSignedData, IdentityDeserializerFactory: identityDeserializerFunc}
+func NewMembershipInfoProvider(mspID string, selfSignedData common.SignedData,
+	identityDeserializerFunc func(chainID string) msp.IdentityDeserializer) *MembershipProvider {
+	return &MembershipProvider{mspID: mspID, selfSignedData: selfSignedData, IdentityDeserializerFactory: identityDeserializerFunc}
 }
 
 // AmMemberOf checks whether the current peer is a member of the given collection config.
 // If getPolicy returns an error, it will drop the error and return false - same as a RejectAll policy.
-func (m *MembershipProvider) AmMemberOf(channelName string, collectionPolicyConfig *common.CollectionPolicyConfig) (bool, error) {
+// It is used when a chaincode is upgraded to see if the peer's org has become eligible after	a collection
```
cendhu, line 32: A nit. Not sure whether we need to comment about the usages of a function. We have tools/editors that help in finding all the callers/callee/references. I am saying this because we use this function twice during a chaincode upgrade.
1.  to check whether the peer is already a member (using the existing collection policy);
2.  to check whether the peer is a newly added member (using the new policy if the check during (1) returns false).

Using both (1) and (2), we check whether a peer has become a new member of a collection. I understand that this comment was added as a result of some questions raised during the review (when `AmMemberOf` and `isEligible` are used). Hence, I am fine if we plan to retain it.

core/common/privdata/membershipinfo.go
```diff
@@ -7,36 +7,97 @@ SPDX-License-Identifier: Apache-2.0
...
+
 	return true, nil
 }
+
+// getMemberOrgs returns a map containing member orgs from a list of MSPPrincipals
+func getMemberOrgs(identities []*mspp.MSPPrincipal, deserializer msp.IdentityDeserializer) (map[string]struct{}, error) {
+	memberOrgs := make(map[string]struct{})
+
+	// get member org MSP IDs from the envelope
+	for _, principal := range identities {
```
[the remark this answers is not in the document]
caod123: Yep, sure I can move it to a shared place. I left it separate because we ended up changing it to return a map instead of a string slice. I can change the consumers of SimpleCollection (coordinator) to also check the map instead.

core/common/privdata/membershipinfo.go
caod123: I added it in response to Dave's request for additional clarification above. I can move it down so it's not in the function signature comment and rather within the function if that helps to clarify the specific use case.

core/common/privdata/collection.go
```diff
@@ -29,7 +29,7 @@ type Collection interface {
 
 	// MemberOrgs returns the collection's members as MSP IDs. This serves as
 	// a human-readable way of quickly identifying who is part of a collection.
-	MemberOrgs() []string
```
cendhu: We can still keep it as `[]string` as the goal is to minimize changes made to release 1.4. We have two other options: 

1. we can convert the map to `[]string` at https://github.com/hyperledger/fabric/blob/1007a5f21fb6b2821535a4e23f2acfe9924122af/core/common/privdata/simplecollection.go#L89
2. we can return `[]string` from `getMemberOrgs` instead of a map. We can do a linear search on the `[]string` to find the match. Given that we use this function only during the chaincode upgrade, no optimization (usage of the map) is required IMO. If optimization is required (though it is not), we in fact need to check for the match after each deserialization of identity rather than collecting all identities at once and then do the matching. 

I think we can go with option (2).

cendhu (commented):
Thanks, @caod123 , for the changes.

integration/pvtdata/pvtdata_test.go
```diff
@@ -230,38 +358,90 @@ var _ bool = Describe("PrivateData", func() {
...
+					)
+					if peer.ID() == "org2.peer1" {
+						// use Admin2 user for peer1.org2
+						sess, err = network.PeerUserSession(peer, "Admin2", commands.ChannelInfo{
+							ChannelID: "testchannel",
+						})
+						Expect(err).NotTo(HaveOccurred())
+						Eventually(sess, network.EventuallyTimeout).Should(gexec.Exit(0))
+					} else {
+						sess, err = network.PeerUserSession(peer, "User1", commands.ChannelInfo{
```
yacovm: if you have "getLedgerHeight" why do you need this "ChannelInfo" command? can't you use the getLedgerheight ?

integration/pvtdata/pvtdata_test.go
```diff
@@ -199,29 +247,109 @@ var _ bool = Describe("PrivateData", func() {
...
+			err = proto.Unmarshal(oldConfig.Config, oldMspConfig)
+			Expect(err).NotTo(HaveOccurred())
+			err = proto.Unmarshal(newConfig.Config, newMspConfig)
+			Expect(err).NotTo(HaveOccurred())
+
+			By("merging the two MSP configs")
+			oldMspConfig.RootCerts = append(oldMspConfig.RootCerts, newMspConfig.RootCerts...)
+			oldMspConfig.TlsRootCerts = append(oldMspConfig.TlsRootCerts, newMspConfig.TlsRootCerts...)
+			oldMspConfig.FabricNodeOus.PeerOuIdentifier.Certificate = nil
+			oldMspConfig.FabricNodeOus.ClientOuIdentifier.Certificate = nil
```
yacovm: all this huge block where you edit the MSPConfig can and should be extracted to its own function.

core/common/privdata/collection.go
caod123: I believe that was what it was before, I originally changed to a map via an earlier suggestion. I do see that the extent of the change is a little too broad for 1.4 though so I'll probably just do the map in the 2.0/master branch fix.

integration/pvtdata/pvtdata_test.go
caod123: getLedgerHeight specifically uses `User1` for the peer user session. Unfortunately any operations on the new peer must use the `Admin2` user so a lot of the logic in this integration test revolves around passing in a custom user for the peer user session commands that need to be run on `org2.peer1`. Because this seemed like a pretty isolated test case I figured it wasn't worth extending the functions to take a user in and just leave them to default to `User1`

yacovm: 1. Why can't you use User1 ? 
2. Can't you add a parameter to the **getLedgerHeight** function to distinguish?

caod123: User1 certs can't be replaced because org2.peer0 still uses them, instead I copied the generated admin user over as Admin2 specifically to use for org2.peer1 operations. I don't mind extending `getLedgerHeight` but figured it wasn't worth it for just this use case.

cendhu:
Thanks, Danny, for the changes.

integration/pvtdata/pvtdata_test.go
```diff
@@ -199,29 +247,105 @@ var _ bool = Describe("PrivateData", func() {
...
+			oldAdminUserPath := filepath.Join(network.RootDir, "crypto", "peerOrganizations", "org2.example.com", "users", "Admin2@org2.example.com")
+			tempAdminUserPath := filepath.Join(tempCryptoDir, "peerOrganizations", "org2.example.com", "users", "Admin@org2.example.com")
+			os.RemoveAll(oldAdminUserPath)
+			err = exec.Command("cp", "-r", tempAdminUserPath, oldAdminUserPath).Run()
+			Expect(err).NotTo(HaveOccurred())
+			err = os.Rename(
+				filepath.Join(oldAdminUserPath, "msp", "signcerts", "Admin@org2.example.com-cert.pem"),
+				filepath.Join(oldAdminUserPath, "msp", "signcerts", "Admin2@org2.example.com-cert.pem"),
+			)
+			Expect(err).NotTo(HaveOccurred())
```
ale-linux: Can we bubble lines 256-285 out in a helper function?

integration/pvtdata/pvtdata_test.go
```diff
@@ -199,29 +247,105 @@ var _ bool = Describe("PrivateData", func() {
...
+
+			By("merging the two MSP configs")
+			mergedMspConfig := mergeMspConfigs(oldMspConfig, newMspConfig)
+
+			By("updating the channel config")
+			updatedConfig.ChannelGroup.Groups["Application"].Groups["org2"].Values["MSP"].Value = utils.MarshalOrPanic(&mspp.MSPConfig{
+				Type:   oldConfig.Type,
+				Config: utils.MarshalOrPanic(mergedMspConfig),
+			})
+			nwo.UpdateConfig(network, orderer, "testchannel", currentConfig, updatedConfig, false, network.Peer("org2", "peer0"))
```
ale-linux: Same as the comment above, I wouldn't inline the plumbing but just create helper functions and call them to perform the significant steps, so that someone reading the test would understand the test case

ale-linux (dismissed):
Thanks Danny! Some nits from my side

core/common/privdata/membershipinfo.go
```diff
@@ -16,27 +16,50 @@ var logger = flogging.MustGetLogger("common.privdata")
...
 	deserializer := m.IdentityDeserializerFactory(channelName)
+
+	// Do a simple check to see if the mspid matches any principal identities in the SignaturePolicy - FAB-17059
+	if collectionPolicyConfig.GetSignaturePolicy() != nil {
+		memberOrgs, err := getMemberOrgs(collectionPolicyConfig.GetSignaturePolicy().GetIdentities(), deserializer)
+		if err != nil {
+			logger.Errorf("Failed getting member orgs from MSP Principals: %s", err)
+		}
+
+		for _, org := range memberOrgs {
```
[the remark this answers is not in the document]
caod123: Yes, you got that right. I agree with your interpretation of what we should do with getMemberOrgs, I'll have it just continue the loop instead of returning on error. In that case I'll probably also just remove the error from being returned from `getMemberOrgs`.

integration/pvtdata/pvtdata_test.go
```diff
@@ -199,29 +247,105 @@ var _ bool = Describe("PrivateData", func() {
...
+				Output: tempCryptoDir,
+			})
+			Expect(err).NotTo(HaveOccurred())
+			Eventually(sess, network.EventuallyTimeout).Should(gexec.Exit(0))
+
+			By("copying the new msp certs for peer1.org2 to the original crypto dir")
 			org2peer1 := network.Peer("org2", "peer1")
-			network.JoinChannel("testchannel", orderer, org2peer1)
-			org2peer1.Channels = append(org2peer1.Channels, &nwo.PeerChannel{Name: "testchannel", Anchor: false})
+			oldPeerMSPPath := filepath.Join(network.RootDir, "crypto", "peerOrganizations", "org2.example.com", "peers", "peer1.org2.example.com", "msp")
```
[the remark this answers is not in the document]
caod123: This should have just been `PeerLocalMSPDir(org2peer1)` in the first place. I'll break up the one below it

core/common/privdata/membershipinfo.go
denyeart: I've confused between AmMemberOf and isEligible, so I'm fine to keep it.

denyeart (approved):

PR #632, opened 2020-02-04
Minor simplification of ApproveChaincodeDefinition logic

wlahti:
#### Type of change

#### Description

Minor simplification to streamline the code path.

#### Related issues

https://jira.hyperledger.org/browse/FAB-17441

### DIFF
--- a/core/chaincode/lifecycle/lifecycle.go
+++ b/core/chaincode/lifecycle/lifecycle.go
@@ -457,7 +457,6 @@ func (ef *ExternalFunctions) ApproveChaincodeDefinitionForOrg(chname, ccname str
 
 			if err := uncommittedParameters.Equal(cd.Parameters()); err == nil {
 				// also check package ID updates
-				uncommittedPackageID := ""
 				metadata, ok, err := ef.Resources.Serializer.DeserializeMetadata(ChaincodeSourcesName, privateName, orgState)
 				if err != nil {
 					return errors.WithMessagef(err, "could not deserialize chaincode-source metadata for %s", privateName)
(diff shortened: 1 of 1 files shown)


core/chaincode/lifecycle/lifecycle.go
```diff
@@ -467,11 +466,10 @@ func (ef *ExternalFunctions) ApproveChaincodeDefinitionForOrg(chname, ccname str
...
 						return errors.WithMessagef(err, "could not deserialize chaincode package for %s", privateName)
 					}
-					uncommittedPackageID = ccLocalPackage.PackageID
-				}
 
-				if uncommittedPackageID == packageID {
-					return errors.Errorf("attempted to redefine uncommitted sequence (%d) for namespace %s with unchanged content", requestedSequence, ccname)
+					if ccLocalPackage.PackageID == packageID {
+						return errors.Errorf("attempted to redefine uncommitted sequence (%d) for namespace %s with unchanged content", requestedSequence, ccname)
+					}
```
guoger, line 472: what if the `ok` at line 464 is false? we would still fail the check before this change, no?

wlahti: Not sure I'm following you. You're referring to the error directly above? 

Unless I'm missing something here, the only place we should care about checking the package id at this point in the code is if there's already been an approval for an uncommitted definition. If we receive a packageID here that's empty, we have problems elsewhere.

guoger: will there ever be a case where ChaincodeSource metadata is empty (so the `ok` at line 464 is `false`), and we are passing in a non-empty pkgID? In this case, we used to error out, but now proceed. I might likely be misinterpreting this code though...

ale-linux (approved):

core/chaincode/lifecycle/lifecycle.go
ale-linux: I think this PR doesn't change the logic - only avoids a local variable declaration (which probably the compiler simplifies anyway). I'll go ahead and merge, if you feel there's an issue you can open a JIRA or revert

PR #642, opened 2020-02-06
[FAB-17479] Migrated Kafka cluster can be safely expanded later

yacovm:
When new orderer nodes are added to a migrated Kafka cluster,
but are not added to all channels:

- The genesis block if written to the ledger to track future addition.
- However, the genesis block is a Kafka one, and therefore the orderer
  tries (wrongly) to launch a Kafka chain which fails.
- In addition, the orderer never tries to track future addition
  of the said channel, because this mechanism is only activated
  for Raft orderers.

This change set:
- Makes the Kafka chain activation be skipped in case
  that the orderer was migrated.
- Starts tracking future addition of the orderer to the channel.

An integration test is included.

Change-Id: Ib3598998073e884747b5966df72b70cac572b1c5


guoger (approved):

orderer/consensus/etcdraft/consenter.go
```diff
@@ -30,16 +30,13 @@ import (
...
-type CreateChainCallback func()
-
 //go:generate mockery -dir . -name InactiveChainRegistry -case underscore -output mocks
 
 // InactiveChainRegistry registers chains that are inactive
 type InactiveChainRegistry interface {
 	// TrackChain tracks a chain with the given name, and calls the given callback
 	// when this chain should be created.
-	TrackChain(chainName string, genesisBlock *common.Block, createChain CreateChainCallback)
+	TrackChain(chainName string, genesisBlock *common.Block, createChain func())
```
hacera-jonathan, line 39: Yup...

hacera-jonathan (approved):
I'd still wait for Jason Y,  but from my end: LGTM

Good catch Yacov

C0rWin (approved):

PR #644, opened 2020-02-07
[FAB-17479] Migrated Kafka cluster can be safely expanded later

yacovm:
When new orderer nodes are added to a migrated Kafka cluster,
but are not added to all channels:

- The genesis block if written to the ledger to track future addition.
- However, the genesis block is a Kafka one, and therefore the orderer
  tries (wrongly) to launch a Kafka chain which fails.
- In addition, the orderer never tries to track future addition
  of the said channel, because this mechanism is only activated
  for Raft orderers.

This change set:
- Makes the Kafka chain activation be skipped in case
  that the orderer was migrated.
- Starts tracking future addition of the orderer to the channel.

An integration test is included.

Change-Id: Ib3598998073e884747b5966df72b70cac572b1c5

### DIFF
--- a/orderer/consensus/kafka/consenter.go
+++ b/orderer/consensus/kafka/consenter.go
@@ -40,12 +40,8 @@ func New(config localconfig.Kafka, mp metrics.Provider, healthChecker healthChec
 
 	metrics := NewMetrics(mp, brokerConfig.MetricRegistry)
 
-	_, isNoopICR := icr.(*NoopInactiveChainRegistry)
-	isRaft := !isNoopICR && icr != nil // a non-noop and not nil InactiveChainRegistry is a Raft one
-
 	return &consenterImpl{
 		mkChain:               mkChain,
-		isRaft:                isRaft,
 		inactiveChainRegistry: icr,
 		brokerConfigVal:       brokerConfig,
 		tlsConfigVal:          config.TLS,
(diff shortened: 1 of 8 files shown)


orderer/consensus/kafka/consenter.go
```diff
@@ -52,17 +60,33 @@ func New(config localconfig.Kafka, metricsProvider metrics.Provider, healthCheck
...
-	healthChecker   healthChecker
-	metrics         *Metrics
+	mkChain               func(string)
+	isRaft                bool
+	brokerConfigVal       *sarama.Config
+	tlsConfigVal          localconfig.TLS
+	retryOptionsVal       localconfig.Retry
+	kafkaVersionVal       sarama.KafkaVersion
+	topicDetailVal        *sarama.TopicDetail
+	metricsProvider       metrics.Provider
```
guoger: why is `metricsProvider` added here?

orderer/consensus/kafka/consenter.go
```diff
@@ -52,17 +60,33 @@ func New(config localconfig.Kafka, metricsProvider metrics.Provider, healthCheck
...
+	// TrackChain tracks a chain with the given name, and calls the given callback
+	// when this chain should be created.
+	TrackChain(chainName string, genesisBlock *cb.Block, createChain func())
+}
+
+type NoopInactiveChainRegistry struct{}
+
+func (*NoopInactiveChainRegistry) TrackChain(chainName string, genesisBlock *cb.Block, createChain func()) {
+}
+
```
guoger: I think this is not really necessary. We could pass a *nil* chain registry to kafka consenter if it's still operating in kafka mode, or an actual `inactiveChainRegistry` otherwise. So we just need to call `TrackChain` if it's not nil (no type assertion needed). wdyt?

guoger (changes requested):

orderer/consensus/kafka/consenter.go
yacovm: I guess bad rebase?

orderer/consensus/kafka/consenter.go
yacovm: makes sense, will adjust.

guoger (approved):

C0rWin (approved):

PR #667, opened 2020-02-13
Fix URLs that don't exist in ledger doc

qpakzk:
#### Type of change

Documentation update

#### Description

Fix URLs that don't exist in ledger doc.

### DIFF
--- a/docs/source/ledger/ledger.md
+++ b/docs/source/ledger/ledger.md
@@ -10,9 +10,9 @@ values.
 
 In this topic, we're going to cover:
 
-* [What is a Ledger?](#what-is-a-ledger?)
+* [What is a Ledger?](#what-is-a-ledger)
 * [Storing facts about business objects](#ledgers-facts-and-states)
-* [A blockchain ledger](#a-blockchain-ledger)
+* [A blockchain ledger](#the-ledger)
 * [The world state](#world-state)
 * [The blockchain data structure](#blockchain)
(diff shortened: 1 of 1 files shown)


docs/source/ledger/ledger.md
```diff
@@ -375,7 +375,7 @@ the same chaincode can access a given namespace.
 
 A blockchain is not namespaced. It contains transactions from many different
 smart contract namespaces. You can read more about chaincode namespaces in this
-[topic](./developapps/chaincodenamespace.html).
+[topic](../developapps/chaincodenamespace.html).
```
pamandrejko, line 378: Can you fix this sentence while you are in here -> `A blockchain is not namespaced.`

to
A blockchain is not a namespace.

pamandrejko (dismissed):
Thanks for finding and fixing these. 
I did a local build and checked all the links you fixed and they work now.

I did notice one small typo unrelated to the fixes you made, and was wondering if you could get that one too as part of this and then we can approve and merge this?

docs/source/ledger/ledger.md
denyeart: I don't love the original text "A blockchain is not namespaced". But I don't think the suggestion "A blockchain is not a namespace" improves upon it.
So for now I suggest we merge the link fixes since that was the intent of this PR, and leave the other sentence as it was.

denyeart (approved):

PR #669, opened 2020-02-13
[FAB-17462] Create signature for config update

DereckLuo:
#### Type of change

- New feature

#### Description

This PR is part of an effort to create an easier way to create and update the channel configuration for fabric networks. This specific PR creates a signature for the provided configuration update.

#### Related issues

[FAB-17467](https://jira.hyperledger.org/browse/FAB-17467)

### DIFF
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -1,0 +1,56 @@
+/*
+Copyright IBM Corp. All Rights Reserved.
+
+SPDX-License-Identifier: Apache-2.0
+*/
+
+package config
+
+import (
+	"crypto/rand"
+
+	"github.com/golang/protobuf/proto"
+	"github.com/hyperledger/fabric-protos-go/common"
+)
+
+// SignConfigUpdate signs the given configuration update with a
+// specific signer.
+func SignConfigUpdate(configUpdate *common.ConfigUpdate, signer *Signer) (*common.ConfigSignature, error) {
+	signatureHeader, err := signer.CreateSignatureHeader()
+	if err != nil {
+		return nil, err
+	}
+
+	configSignature := &common.ConfigSignature{
+		SignatureHeader: MarshalOrPanic(signatureHeader),
+	}
+	configUpdateBytes := MarshalOrPanic(configUpdate)
+	configSignature.Signature, err = signer.Sign(rand.Reader, concatenateBytes(configSignature.SignatureHeader, configUpdateBytes))
+	if err != nil {
+		return nil, err
+	}
+
+	return configSignature, nil
+}
+
+// concatenateBytes combines multiple arrays of bytes, for signatures or digests
+// over multiple fields.
+func concatenateBytes(data ...[]byte) []byte {
+	bytes := []byte{}
+	for _, d := range data {
+		for _, b := range d {
+			bytes = append(bytes, b)
+		}
+	}
+	return bytes
+}
+
+// MarshalOrPanic serializes a protobuf message and panics if this
+// operation fails.
+func MarshalOrPanic(pb proto.Message) []byte {
+	data, err := proto.Marshal(pb)
+	if err != nil {
+		panic(err)
+	}
+	return data
+}
--- a/pkg/config/config_test.go
+++ b/pkg/config/config_test.go
@@ -1,0 +1,33 @@
+/*
+Copyright IBM Corp. All Rights Reserved.
+
+SPDX-License-Identifier: Apache-2.0
+*/
+
+package config
+
+import (
+	"testing"
+
+	"github.com/golang/protobuf/proto"
+	"github.com/hyperledger/fabric-protos-go/common"
+	. "github.com/onsi/gomega"
+)
+
+func TestSignConfigUpdate(t *testing.T) {
+	t.Run("success", func(t *testing.T) {
+		gt := NewGomegaWithT(t)
+
+		signer, err := NewSigner([]byte(publicKey), []byte(privateKey), "test-msp")
+		gt.Expect(err).NotTo(HaveOccurred())
+		configSignature, err := SignConfigUpdate(&common.ConfigUpdate{}, signer)
+		gt.Expect(err).NotTo(HaveOccurred())
+
+		expectedCreator, err := signer.Serialize()
+		gt.Expect(err).NotTo(HaveOccurred())
+		signatureHeader := &common.SignatureHeader{}
+		err = proto.Unmarshal(configSignature.SignatureHeader, signatureHeader)
+		gt.Expect(err).NotTo(HaveOccurred())
+		gt.Expect(signatureHeader.Creator).To(Equal(expectedCreator))
+	})
+}
--- a/pkg/config/signer.go
+++ b/pkg/config/signer.go
@@ -1,0 +1,225 @@
+/*
+Copyright IBM Corp. All Rights Reserved.
+
+SPDX-License-Identifier: Apache-2.0
+*/
+
+package config
+
+import (
+	"crypto"
+	"crypto/ecdsa"
+	"crypto/rand"
+	"crypto/x509"
+	"encoding/asn1"
+	"encoding/pem"
+	"errors"
+	"fmt"
+	"io"
+	"math/big"
+
+	"github.com/hyperledger/fabric-protos-go/common"
+	"github.com/hyperledger/fabric-protos-go/msp"
+)
+
+// Signer holds the information necessary to sign a configuration update.
+// It implements the crypto.Signer interface for ECDSA keys.
+type Signer struct {
+	cert       *x509.Certificate
+	mspID      string
+	privateKey *ecdsa.PrivateKey
+	publicKey  *ecdsa.PublicKey
+}
+
+type ecdsaSignature struct {
+	R, S *big.Int
+}
+
+// NewSigner creates a new signer for configuration updates. The publicCert
+// and privateKey should be pem encoded.
+func NewSigner(publicCert []byte, privateCert []byte, mspID string) (*Signer, error) {
+	if mspID == "" {
+		return nil, errors.New("failed to create new signer, mspID can not be empty")
+	}
+
+	cert, err := getCertFromPem(publicCert)
+	if err != nil {
+		return nil, fmt.Errorf("failed to get cert from pem: %v", err)
+	}
+
+	publicKey, err := ecdsaPublicKeyImport(cert)
+	if err != nil {
+		return nil, fmt.Errorf("failed to get ECDSA public key: %v", err)
+	}
+
+	pkBlock, _ := pem.Decode(privateCert)
+	if pkBlock == nil {
+		return nil, errors.New("failed to decode private key from pem")
+	}
+
+	privateKey, err := ecdsaPrivateKeyImport(pkBlock.Bytes)
+	if err != nil {
+		return nil, fmt.Errorf("failed to get ECDSA private key: %v", err)
+	}
+
+	signer := &Signer{
+		cert:       cert,
+		mspID:      mspID,
+		privateKey: privateKey,
+		publicKey:  publicKey,
+	}
+	return signer, nil
+}
+
+// Serialize returns a byte array representation of this identity.
+func (s *Signer) Serialize() ([]byte, error) {
+	pb := &pem.Block{Bytes: s.cert.Raw, Type: "CERTIFICATE"}
+	pemBytes := pem.EncodeToMemory(pb)
+	if pemBytes == nil {
+		return nil, errors.New("failed to encode pem block")
+	}
+
+	// serialize identities by prepending the MSPID and appending
+	// the ASN.1 DER content of the cert
+	sID := &msp.SerializedIdentity{Mspid: s.mspID, IdBytes: pemBytes}
+	idBytes := MarshalOrPanic(sID)
+
+	return idBytes, nil
+}
+
+// Public returns the public key of the signer.
+func (s *Signer) Public() crypto.PublicKey {
+	return s.publicKey
+}
+
+// Cert returns the certificate of the signer.
+func (s *Signer) Cert() *x509.Certificate {
+	return s.cert
+}
+
+// MSPId returns the MSP ID of the signer.
+func (s *Signer) MSPId() string {
+	return s.mspID
+}
+
+// Sign performs ECDSA sign with signer's private key on given digest. It
+// ensures signatures are created with Low S values since Fabric normalizes
+// all signatures to Low S.
+// See https://github.com/bitcoin/bips/blob/master/bip-0146.mediawiki#low_s
+// for more detail.
+func (s *Signer) Sign(reader io.Reader, digest []byte) (signature []byte, err error) {
+	if reader == nil {
+		return nil, errors.New("failed to sign, reader can not be nil")
+	}
+
+	rr, ss, err := ecdsa.Sign(reader, s.privateKey, digest)
+	if err != nil {
+		return nil, err
+	}
+
+	// ensure Low S signatures
+	sig := toLowS(
+		s.privateKey.PublicKey,
+		ecdsaSignature{
+			R: rr,
+			S: ss,
+		},
+	)
+
+	return asn1.Marshal(sig)
+}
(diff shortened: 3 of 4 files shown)


pkg/config/signer.go
```diff
@@ -0,0 +1,244 @@
...
+	if err != nil {
+		return nil, fmt.Errorf("failed to generate random nonce: %s", err)
+	}
+	return nonce, nil
+}
+
+func getRandomNonce() ([]byte, error) {
+	key := make([]byte, 24)
+
+	_, err := rand.Read(key)
```
ale-linux, line 162: we should check that this is not a short read

pkg/config/signer_test.go
```diff
@@ -0,0 +1,221 @@
...
+	"math/big"
+	"testing"
+
+	"github.com/gogo/protobuf/proto"
+	"github.com/hyperledger/fabric-protos-go/msp"
+	. "github.com/onsi/gomega"
+)
+
+var (
+	publicKey = `
```
ale-linux: how have these fixtures been generated? Ideally we should just run go generate in the folder to re-generate them. Failing that, can we at least have a comment describing how to re-generate them?

pkg/config/signer.go
```diff
@@ -0,0 +1,244 @@
...
+func toLowS(k *ecdsa.PublicKey, s *big.Int) (*big.Int, bool, error) {
+	lowS, err := isLowS(k, s)
+	if err != nil {
+		return nil, false, err
+	}
+
+	if !lowS {
+		// Set s to N - s that will be then in the lower part of signature space
+		// less or equal to half order
+		s.Sub(k.Params().N, s)
```
ale-linux: untested

ale-linux (changes requested):
Thanks for this PR - a few nits from my side

pkg/config/signer.go
```diff
@@ -0,0 +1,244 @@
...
+
+	// serialize identities by prepending the MSPID and appending
+	// the ASN.1 DER content of the cert
+	sID := &msp.SerializedIdentity{Mspid: s.mspID, IdBytes: pemBytes}
+	idBytes := MarshalOrPanic(sID)
+
+	return idBytes, nil
+}
+
+// Public returns the public key of the signer.
```
yacovm, line 90: what is the use case that calls for this method? (Public key extraction).  
The public key is an empty interface, what does it tell us? Perhaps it should be worth delaying this and introduce this later when we need it?

pkg/config/signer.go
```diff
@@ -0,0 +1,244 @@
...
+
+	_, err := rand.Read(key)
+	if err != nil {
+		return nil, fmt.Errorf("failed to get random bytes: %s", err)
+	}
+	return key, nil
+}
+
+// isLowS checks that s is a low-S.
+func isLowS(k *ecdsa.PublicKey, s *big.Int) (bool, error) {
```
yacovm: ***ecdsa.PublicKey** embeds **elliptic.Curve** which has **Params() \*CurveParams** and it has the order of the curve base point which you can divide by 2. why do we need the **curveHalfOrders** map?

pkg/config/signer.go
```diff
@@ -0,0 +1,244 @@
...
+	if reader == nil {
+		return nil, errors.New("failed to sign, reader can not be nil")
+	}
+
+	rr, ss, err := ecdsa.Sign(reader, s.privateKey, digest)
+	if err != nil {
+		return nil, err
+	}
+
+	ss, _, err = toLowS(&s.privateKey.PublicKey, ss)
```
yacovm: If this is to be in _pkg_ then let's at least _document_ why we're doing this.

pkg/config/config_test.go
```diff
@@ -0,0 +1,33 @@
...
+
+SPDX-License-Identifier: Apache-2.0
+*/
+
+package config
+
+import (
+	"testing"
+
+	"github.com/gogo/protobuf/proto"
```
yacovm: it's the wrong import

pkg/config/signer_test.go
```diff
@@ -0,0 +1,221 @@
...
+
+import (
+	"crypto/ecdsa"
+	"crypto/rand"
+	"crypto/x509"
+	"io"
+	"math/big"
+	"testing"
+
+	"github.com/gogo/protobuf/proto"
```
yacovm: wrong import

wlahti: Oops, we'll fix this.

pkg/config/config_test.go
wlahti: Oops, we'll fix this.

pkg/config/signer.go
wlahti: Dereck and I won't pretend to have a full understanding of ecdsa. Feedback like this is much needed and appreciated!

pkg/config/signer.go
wlahti: We were just implementing the [crypto.Signer](https://golang.org/pkg/crypto/#Signer) interface here. Not sure that we actually need to but we were trying to at least match the Sign function.

pkg/config/signer.go
wlahti: Document the "toLowS" part? We could use some input from some crypto experts on what some of these operations actually represent. As is, we pulled most of the crypto code here from the fabric bccsp/msp packages.

pkg/config/signer.go
wlahti: We pulled this code directly from [protutil/commonutils](https://github.com/hyperledger/fabric/blob/master/protoutil/commonutils.go#L257-L265). It seems consistent with the example code in [crypto.Rand](https://golang.org/pkg/crypto/rand/#Read) but if you think there's something we should to do improve it, we can certainly incorporate that.

pkg/config/signer_test.go
wlahti: Good catch. Didn't mean to include these fixtures instead of generating them. We'll fix this.

pkg/config/signer.go
wlahti: Ah, found another place in fabric where we have a "toLowS" function. And it's documented! https://github.com/hyperledger/fabric/blob/master/internal/cryptogen/csp/csp.go#L137-L143 We'll incorporate that info here. 

edit: And there are a lot of other useful comments in that file that we'll reuse here.

pkg/config/signer_test.go
wlahti: Done.

pkg/config/config_test.go
wlahti: Done.

pkg/config/signer.go
wlahti: Done.

pkg/config/signer.go
wlahti: Done.

pkg/config/signer.go
wlahti: I updated the low S functions based on the implementation in [cryptogen](https://github.com/hyperledger/fabric/blob/master/internal/cryptogen/csp/csp.go) and copied in the related tests.

pkg/config/signer_test.go
wlahti: Looking into this now. I'll update the PR when things are ready.

pkg/config/signer.go
yacovm: It's fine due to:

> On return, n == len(b) if and only if err == nil. 

From: https://golang.org/pkg/crypto/rand/#Read

pkg/config/signer.go
```diff
@@ -0,0 +1,225 @@
...
+		return nil, fmt.Errorf("failed to get random bytes: %s", err)
+	}
+	return key, nil
+}
+
+// When using ECDSA, both (r,s) and (r, -s mod n) are valid signatures.
+// In order to protect against signature malleability attacks, Fabric
+// normalizes all signatures to a canonical form where s is at most half
+// the order of the curve. In order to make signatures compliant with what
+// Fabric expects, toLowS creates signatures in this canonical form.
```
yacovm, line 173: This makes me wonder - why are we enforcing this on signatures produced by leaf certificates like peers or orderers?
I understand why we need this for CA signatures on identities, but - what is the attack we're protecting against here? Duplication of SignedData (triplet of message, signature and identity)? Seems unlikely since I can anyway sign as many times as I want, and we de-duplicate based on identity.

Bitcoin [did it](https://github.com/bitcoin/bips/blob/master/bip-0146.mediawiki) because of: 
>  For non-segregated witness transactions, signature malleability will change the txid and invalidate any unconfirmed child transactions.

But Fabric signatures do not affect the transaction ID, and they always sign an opaque marshaled proto message.

@ale-linux  ?

pkg/config/signer_test.go
wlahti: Done. Adapted from the code here: https://golang.org/src/crypto/tls/generate_cert.go

pkg/config/signer.go
ale-linux: Yeah, that's not tested either, because it was also copied... the original sources are here ./bccsp/utils/ecdsa.go, and you'll find tests there.

ale-linux (changes requested):
You can add the missing test coverage from ./bccsp/utils/ecdsa.go - thx!

pkg/config/signer.go
wlahti: The tests are there already. I copied those as well when I switched.

ale-linux: Ah sorry, I missed that - should have re-run.. thanks!

pkg/config/signer.go
ale-linux: I also don't see any issues with endorser signatures; at the same time, it's more uniform to have the same approach everywhere, and we don't know what new type of signature leaf nodes might be required to produce in the future, so it seems prudent to me to stick with this restriction.

ale-linux (approved):

pkg/config/config.go
```diff
@@ -0,0 +1,56 @@
...
+import (
+	"crypto/rand"
+
+	"github.com/golang/protobuf/proto"
+	"github.com/hyperledger/fabric-protos-go/common"
+)
+
+// SignConfigUpdate signs the given configuration update with a
+// specific signer.
+func SignConfigUpdate(configUpdate *common.ConfigUpdate, signer *Signer) (*common.ConfigSignature, error) {
```
sykesm, line 18: In this context, "Signer" is confusing. Yes, it's a concrete type that implements `crypto.Signer` but it's not a Signer. What you've created with `NewSigner` is essentially a SigningIdentity. We should consider names that more closely reflect what we're dealing with.

sykesm (changes requested):
I realize this has been merged but there are number of things that I think we need to improve before progressing to the next task.

pkg/config/config.go
```diff
@@ -0,0 +1,56 @@
...
+
+// MarshalOrPanic serializes a protobuf message and panics if this
+// operation fails.
+func MarshalOrPanic(pb proto.Message) []byte {
+	data, err := proto.Marshal(pb)
+	if err != nil {
+		panic(err)
+	}
+	return data
+}
```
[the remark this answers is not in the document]
yacovm: If you fail Marshaling a protobuf message it's probably due to a software error and not a transient thing.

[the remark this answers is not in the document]
sykesm: Yes, and in a library (not a server) it should be the caller that determines how to report that error and handle it.

https://github.com/golang/go/wiki/CodeReviewComments#dont-panic

pkg/config/config_test.go
```diff
@@ -0,0 +1,33 @@
...
+import (
+	"testing"
+
+	"github.com/golang/protobuf/proto"
+	"github.com/hyperledger/fabric-protos-go/common"
+	. "github.com/onsi/gomega"
+)
+
+func TestSignConfigUpdate(t *testing.T) {
+	t.Run("success", func(t *testing.T) {
```
caod123, line 18: Does this need to have a t.Run right here? It's the only test in this function

pkg/config/signer_test.go
```diff
@@ -0,0 +1,289 @@
...
+				gt.Expect(err).To(MatchError(tc.expectedErr))
+			} else {
+				gt.Expect(err.Error()).To(ContainSubstring(tc.expectedErr))
+			}
+		})
+	}
+}
+
+func TestECDSAPublicKeyImport(t *testing.T) {
+	t.Run("certificate does not contain valid ecdsa publicKey", func(t *testing.T) {
```
caod123, line 109: See comment above as well as for the ones below. If you want to annotate the test you can add the context to the matcher https://onsi.github.io/gomega/#annotating-assertions
t.Run is useful for sub-testing (table testing etc) and allows for finer grained control when trying to run a specific test from CLI. If there's only one test in the function you can still run that specific test via the CLI using the top level function name

pkg/config/config_test.go
DereckLuo: agreed. when we implemented the test, we were thinking about adding more test cases in this function. it's definitely not necessary now.

pkg/config/config.go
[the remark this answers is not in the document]
DereckLuo: done

pkg/config/signer.go
```diff
@@ -0,0 +1,225 @@
...
+	publicKey  *ecdsa.PublicKey
+}
+
+type ecdsaSignature struct {
+	R, S *big.Int
+}
+
+// NewSigner creates a new signer for configuration updates. The publicCert
+// and privateKey should be pem encoded.
+func NewSigner(publicCert []byte, privateCert []byte, mspID string) (*Signer, error) {
```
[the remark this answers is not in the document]
DereckLuo: done

pkg/config/signer_test.go
```diff
@@ -0,0 +1,289 @@
...
+)
+
+var publicKey, privateKey []byte
+
+func TestMain(m *testing.M) {
+	publicKey, privateKey = generatePublicAndPrivateKey()
+
+	os.Exit(m.Run())
+}
+
```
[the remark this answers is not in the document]
DereckLuo: done

pkg/config/signer.go
```diff
@@ -0,0 +1,225 @@
...
+	}
+
+	return &common.SignatureHeader{
+		Creator: creator,
+		Nonce:   nonce,
+	}, nil
+}
+
+// createNonce generates a nonce using the crypto/rand package.
+func createNonce() ([]byte, error) {
```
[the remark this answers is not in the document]
DereckLuo: done

pkg/config/signer.go
```diff
@@ -0,0 +1,225 @@
...
+
+	publicKey, err := ecdsaPublicKeyImport(cert)
+	if err != nil {
+		return nil, fmt.Errorf("failed to get ECDSA public key: %v", err)
+	}
+
+	pkBlock, _ := pem.Decode(privateCert)
+	if pkBlock == nil {
+		return nil, errors.New("failed to decode private key from pem")
+	}
```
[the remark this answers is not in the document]
DereckLuo: done

pkg/config/signer.go
```diff
@@ -0,0 +1,225 @@
...
+)
+
+// Signer holds the information necessary to sign a configuration update.
+// It implements the crypto.Signer interface for ECDSA keys.
+type Signer struct {
+	cert       *x509.Certificate
+	mspID      string
+	privateKey *ecdsa.PrivateKey
+	publicKey  *ecdsa.PublicKey
+}
```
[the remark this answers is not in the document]
DereckLuo: done

pkg/config/config.go
DereckLuo: done

pkg/config/config_test.go
DereckLuo: done

pkg/config/signer_test.go
DereckLuo: done

DereckLuo:
fixes for additional requests has been applied in this PR: https://github.com/hyperledger/fabric/pull/687

PR #670, opened 2020-02-13
Add release notes for Fabric v1.4.5

denyeart:
Add release notes for Fabric v1.4.5

### DIFF
--- a/release_notes/v1.4.5.md
+++ b/release_notes/v1.4.5.md
@@ -1,0 +1,156 @@
+v1.4.5 Release Notes - Date TBD, 2020
+=====================================
+
+What's New in Hyperledger Fabric v1.4.5
+---------------------------------------
+
+The following enhancements are included in this release:
+
+- **FAB-15648: Allow separate TLS config for Raft cluster and client**
+
+  When intra-cluster Raft communication is configured to use separate
+  listener, allow TLS to be disabled for the client-facing listener,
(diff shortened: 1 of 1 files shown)


yacovm (approved):

robmurgai (approved):
Overall looks good.  A couple of points for thought:

The deprecations make sense and it would be nice to see them replicate somehow in our ReadTheDocs as good practices.  For instance, use Raft-based ordering service with a single node instead of Solo ordering service when building your first network, etc in our Test-Network.

denyeart:
> Overall looks good. A couple of points for thought:
> 
> The deprecations make sense and it would be nice to see them replicate somehow in our ReadTheDocs as good practices. For instance, use Raft-based ordering service with a single node instead of Solo ordering service when building your first network, etc in our Test-Network.

@robmurgai 
Makes sense. Most of the deprecations already have similar comments in doc, e.g. for single node Raft:

test network:
"The sample network uses a single node Raft ordering service that is operated by the ordering organization."

And in ordering doc topic:
"Existing users of Solo should move to a single node Raft network for equivalent function."

But some of the other deprecations should be clarified in doc or samples, I'll open some issues for this.

denyeart:
Unit test gossip flakes. Since the PR does not touch any code, I'll go ahead and merge without the unit tests.

PR #671, opened 2020-02-14
[FAB-17453] 'peer lifecycle chaincode package' mandates does not need to init crypto

wuxuer:
Description

A developer might never have any credentials if He is just developing and packaging,
so the MSP initialization process should be removed while packaging with
lifecycle.

Related issues

https://jira.hyperledger.org/browse/FAB-17453

#### Type of change

- Bug fix

### DIFF
--- a/internal/peer/common/common.go
+++ b/internal/peer/common/common.go
@@ -305,6 +305,12 @@ func InitCmd(cmd *cobra.Command, args []string) {
 		LogSpec: loggingSpec,
 	})
 
+	// chaincode packaging does not require material from the local MSP
+	if cmd.CommandPath() == "peer lifecycle chaincode package" {
+		mainLogger.Debug("peer lifecycle chaincode package does not need to init crypto")
+		return
+	}
+
 	// Init the MSP
 	var mspMgrConfigDir = config.GetPath("peer.mspConfigPath")
 	var mspID = viper.GetString("peer.localMspId")
--- a/internal/peer/common/common_test.go
+++ b/internal/peer/common/common_test.go
@@ -10,6 +10,7 @@ import (
 	"fmt"
 	"os"
 	"path/filepath"
+	"strings"
 	"testing"
 
 	"github.com/hyperledger/fabric/common/flogging"
@@ -17,6 +18,7 @@ import (
 	"github.com/hyperledger/fabric/core/config/configtest"
 	"github.com/hyperledger/fabric/internal/peer/common"
 	"github.com/hyperledger/fabric/msp"
+	"github.com/spf13/cobra"
 	"github.com/spf13/viper"
 	"github.com/stretchr/testify/assert"
 )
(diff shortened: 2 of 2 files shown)


internal/peer/common/common.go
```diff
@@ -305,6 +305,13 @@ func InitCmd(cmd *cobra.Command, args []string) {
 		LogSpec: loggingSpec,
 	})
 
+	// 'peer lifecycle chaincode package' mandates does not need to init crypto
+	// For details, please refer to FAB-17453
+	if cmd != nil && strings.Contains(cmd.CommandPath(), "lifecycle") && cmd.Name() == "package" {
+		mainLogger.Debug("peer lifecycle chaincode package does not need to init crypto")
```
ale-linux, line 310: Looks good to me - could you please just add a test case?

ale-linux (dismissed):

internal/peer/common/common.go
wuxuer: OK. thx.
 I will try to add a test case.

internal/peer/common/common.go
```diff
@@ -305,6 +305,13 @@ func InitCmd(cmd *cobra.Command, args []string) {
 		LogSpec: loggingSpec,
 	})
 
+	// 'peer lifecycle chaincode package' mandates does not need to init crypto
+	// For details, please refer to FAB-17453
+	if cmd != nil && strings.Contains(cmd.CommandPath(), "lifecycle") && cmd.Name() == "package" {
```
sykesm: Checking to see if the command path contains the word "lifecycle" is too fragile, IMO. If you can write a debug message the explicitly states the path, we can certainly use that as the check.

Additionally, why are we checking for a nil cmd? If we get a nil command, something is semantically wrong. How are we to initialize a command when there is no command? Why are we initializing a command if there is no command?

It may be necessary given the existing (poor) structure of the code but if it is, it would be nice to fix the source of that issue as well.

sykesm (changes requested):
Thank you for the contribution. A few high level comments:

1. As a tactical fix, this seems to work. I would strongly prefer we do more than simply check for a (any) "lifecycle" subcommand in the command path.
2. Tests are important. If we have a behavior we are relying on (and I expect we do since we're fixing something) there should be a test explicitly asserts that behavior holds.

Before merging, I'd like to see those addressed.

Beyond that I am generally concerned with tactical fixes of this nature. The code used in the cli is already a mess and this isn't really doing anything to improve on that. In this case in particular, a "common" path is being changed to introduce exceptions based on the calling context; this means it is no longer common. That implies each of the commands require a mechanism to express what needs to be initialized or take over and explicitly initialize as needed.

internal/peer/common/common.go
wuxuer: Cmd cannot be empty in real environment, and the nil cmd only appear in test functions to simplify parameters, as TestInitCmd function in common_test.go. So I check if it is nil.

wuxuer: The root cause of this problem is that chaincodeCmd(in file internal/peer/lifecycle/chaincode/chaincode.go) registers the common.InitCmd function as a hook function, which has package, install, commit and other subcommads, but the cobra framework does not provide parameters to suppress this hook function for a subcommand. So we cannot treat these subcommands differently at this stage.
Define a new cobra.Command variable to register "package" command, which will not register function common.InitCmd as a hook function, instead of  chaincodeCmd. This approach may be better, but it will have some impact on the existing command line format,  which may require more community proposals. So I chose this tactical fix.

sykesm: > The root cause of this problem is that chaincodeCmd(in file internal/peer/lifecycle/chaincode/chaincode.go) registers the common.InitCmd function as a hook function

I understand the cause and I am familiar with this code. A larger non-tactical fix is certainly possible given the InitCmd calls are coming from the pre-execution hooks in the commands we register. Regardless, that wasn’t one of the things holding up this merge.

> Cmd cannot be empty in real environment, and the nil cmd only appear in test functions to simplify parameters, as TestInitCmd function in common_test.go. So I check if it is nil.

It’s in exactly _one_ test that we pass a nil. The test needs to be changed to pass an empty command instead of nil and the production code updated such that it does not check for a nil command.

> Checking to see if the command path contains the word "lifecycle" is too fragile

The point I was trying to make with that comment was that the production code change for a tactical fix like this should be as simple as:
```go
    if cmd.CommandPath() == "peer lifecycle chaincode package" {
        mainLogger.Debug("peer lifecycle chaincode package does not need to init crypto")
        return
    }
```

With that one block in the production code, your test would need to change to provide a “peer” command in your setup and the TestInitCmd test would have to pass `&cobra.Command{}` as the first argument to `InitCmd`. With that, the fix and tests would be complete.

sykesm (changes requested):
Thank you for adding tests. Below is an outline for the minor updates I’m suggesting.

```diff
diff --git a/internal/peer/common/common.go b/internal/peer/common/common.go
index 3a5d91d22..b2d46d092 100644
--- a/internal/peer/common/common.go
+++ b/internal/peer/common/common.go
@@ -305,9 +305,8 @@ func InitCmd(cmd *cobra.Command, args []string) {
                LogSpec: loggingSpec,
        })
 
-       // 'peer lifecycle chaincode package' mandates does not need to init crypto
-       // For details, please refer to FAB-17453
-       if cmd != nil && strings.Contains(cmd.CommandPath(), "lifecycle") && cmd.Name() == "package" {
+       // chaincode packaging does not require material from the local MSP
+       if cmd.CommandPath() == "peer lifecycle chaincode package" {
                mainLogger.Debug("peer lifecycle chaincode package does not need to init crypto")
                return
        }
diff --git a/internal/peer/common/common_test.go b/internal/peer/common/common_test.go
index 19db44b79..61e735b92 100644
--- a/internal/peer/common/common_test.go
+++ b/internal/peer/common/common_test.go
@@ -193,7 +193,7 @@ func TestInitCmd(t *testing.T) {
 
        origEnvValue := os.Getenv("FABRIC_LOGGING_SPEC")
        os.Setenv("FABRIC_LOGGING_SPEC", "chaincode=debug:test.test2=fatal:abc=error")
-       common.InitCmd(nil, nil)
+       common.InitCmd(&cobra.Command{}, nil)
        assert.Equal(t, "debug", flogging.LoggerLevel("chaincode"))
        assert.Equal(t, "info", flogging.LoggerLevel("test"))
        assert.Equal(t, "fatal", flogging.LoggerLevel("test.test2"))
@@ -206,6 +206,9 @@ func TestInitCmdWithoutInitCrypto(t *testing.T) {
        defer cleanup()
        defer viper.Reset()
 
+       peerCmd := &cobra.Command{
+               Use: "peer",
+       }
        lifecycleCmd := &cobra.Command{
                Use: "lifecycle",
        }
@@ -215,9 +218,10 @@ func TestInitCmdWithoutInitCrypto(t *testing.T) {
        packageCmd := &cobra.Command{
                Use: "package",
        }
-       // lifecycle chaincode package
+       // peer lifecycle chaincode package
        chaincodeCmd.AddCommand(packageCmd)
        lifecycleCmd.AddCommand(chaincodeCmd)
+       peerCmd.AddCommand(lifecycleCmd)
 
        // MSPCONFIGPATH is default value
        common.InitCmd(packageCmd, nil)
```

Thanks!

internal/peer/common/common.go
wuxuer: My understanding was a bit off before, I'm very sorry. It has been modified again according to your suggestion, please review.

sykesm (approved):
Thanks.

sykesm:

PR #672, opened 2020-02-14
[FAB-17472] Clarify doc and samples for NodeOU Certificate

denyeart:
#### Type of change

- Documentation and log message update

#### Description

Clarify in doc and samples that NodeOU classification
Certificate can be left unconfigured in scenarios
where any of the organization's CA's can issue NodeOU
classifications. This is the recommended configuration
so that it is possible to add CA or intermediate certs later,
without having to reissue existing credentials (only a single
Certificate can be configured for each NodeOU classification).

Also, change a Info message to Warning message, if the
configured Certificate cannot be found, for example when
running configtxgen.

### DIFF
--- a/docs/source/msp.rst
+++ b/docs/source/msp.rst
@@ -189,17 +189,24 @@ below:
 
    NodeOUs:
      Enable: true
+     # For each identity classification that you would like to utilize, specify
+     # an OU identifier.
+     # You can optionally configure that the OU identifier must be issued by a specific CA
+     # or intermediate certificate from your organization. However, it is typical to NOT
+     # configure a specific Certificate. By not configuring a specific Certificate, you will be
+     # able to add other CA or intermediate certs later, without having to reissue all credentials.
+     # For this reason, the sample below comments out the Certificate field.
      ClientOUIdentifier:
-       Certificate: "cacerts/cacert.pem"
+       # Certificate: "cacerts/cacert.pem"
        OrganizationalUnitIdentifier: "client"
      AdminOUIdentifier:
-       Certificate: "cacerts/cacert.pem"
+       # Certificate: "cacerts/cacert.pem"
        OrganizationalUnitIdentifier: "admin"
      PeerOUIdentifier:
-       Certificate: "cacerts/cacert.pem"
+       # Certificate: "cacerts/cacert.pem"
        OrganizationalUnitIdentifier: "peer"
      OrdererOUIdentifier:
-       Certificate: "cacerts/cacert.pem"
+       # Certificate: "cacerts/cacert.pem"
        OrganizationalUnitIdentifier: "orderer"
 
 Identity classification is enabled when ``NodeOUs.Enable`` is set to ``true``. Then the client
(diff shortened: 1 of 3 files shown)


docs/source/msp.rst
```diff
@@ -189,17 +189,22 @@ below:
 
    NodeOUs:
      Enable: true
+     # If a specific Certificate is not configured per identifier below,
+     # then the identity classification OU can be issued by any of the organization's configured CAs.
+     # This allows for adding or rotating CA or intermediate certs later,
+     # without having to reissue all credentials.
+     # Configure a Certificate only if you want a specific CA or intermediate cert to issue identity classification OUs.
```
pamandrejko: I'm unclear on the use case. Can you provide an example of what you mean? 
It seems like you are suggesting to not specify an OrganizationalUnitIdentifer for an identity? Sort of like leave it empty for now?

pamandrejko (changes requested):

pamandrejko (changes requested):
It would be good in the msp.rst if we could add a use case for doing this.

nikhil550:
```
 If a specific Certificate is not configured per identifier below,
  # then the identity classification OU can be issued by any of the organization's configured CAs.
  # This allows for adding or rotating CA or intermediate certs later,
  # without having to reissue all credentials.
  # Configure a Certificate only if you want a specific CA or intermediate cert to issue identity classification OUs.
```
Feel like this should be reversed in this section and the above doc. ie, the identity classification OU is used to configure which NodeOU are recognized as administrators or components of the network. You can configure the Certificate field if you want a Node OU to be issue by a specific CA or intermediate CA.

docs/source/msp.rst
denyeart: @pamandrejko The general use case is included in the comment and in the rst sentence below...

General Use Case: I want to be able to add or rotate my organization's CA or intermediate certs later, without having to replace all the existing client/admin/peer/orderer credentials at the same time.

This is not about "OrganizationalUnitIdentifier". It is about not configuring a Certificate per OU identifier. See the commented out Certificate lines below. The problem is that only a single Certificate can be configured, not multiple. So by configuring a single certificate here, it will render node OU classifications issued from other CA or intermediate certs from the same org invalid.

This came up in a side-by-side migration scenarios, where a second CA and second set of peers is expected to come online, while the original peers from the original CA are expected to keep working. To keep both sets working, an individual certificate must not be configured.  (It would have been nice to be able to configure multiple certificates, but that would require a capability change...).

With this knowledge, what do you recommend be added to make it more clear?

docs/source/msp.rst
```diff
@@ -213,7 +218,8 @@ a. ``OrganizationalUnitIdentifier``: Is the OU value that the x509 certificate n
 b. ``Certificate``: Set this to the path of the CA or intermediate CA certificate under which client
    (peer, admin or orderer) identities should be validated. The field is relative to the MSP root
    folder. This field is optional. You can leave this field blank and allow the certificate to be
```
pamandrejko: My suggestion:
b. ``Certificate``: (Optional) Set this to the path of the CA or intermediate CA certificate under which client
   (peer, admin or orderer) identities should be validated. The specified path is relative to the MSP root
   folder.  If you do not set this field, then the identities are validated under any CA defined in the MSP configuration, which could be desirable in the future if you need to add other CA or intermediate certificates.

The word `rotate` was really confusing me. Maybe it is important to you to keep that.

Do we have a best practice recommendation for them here? Like Normally, this field can be safely ignored?

pamandrejko (changes requested):

docs/source/msp.rst
denyeart: Thanks, I addressed the comments.

docs/source/msp.rst
denyeart: Thanks, I addressed the comments.
I changed the wording to say it is 'typical' to not configure the Certificate.

denyeart:
> ```
>  If a specific Certificate is not configured per identifier below,
>   # then the identity classification OU can be issued by any of the organization's configured CAs.
>   # This allows for adding or rotating CA or intermediate certs later,
>   # without having to reissue all credentials.
>   # Configure a Certificate only if you want a specific CA or intermediate cert to issue identity classification OUs.
> ```
> 
> Feel like this should be reversed in this section and the above doc. ie, the identity classification OU is used to configure which NodeOU are recognized as administrators or components of the network. You can configure the Certificate field if you want a Node OU to be issue by a specific CA or intermediate CA.

Thanks, I've addressed the comments.

ale-linux (approved):
Looks good Dave, thanks!

ale-linux:
Comments addressed

PR #673, opened 2020-02-14
Cherry-pick release 1.4 fix for FAB-17059 to master

caod123:
Cherry-pick of #626 with updates to work with master

* Also includes change to `getMemberOrgs` to return a `map[string]struct{}` instead of string slice for more efficient membership checks

/cc @ale-linux @denyeart

### DIFF
--- a/core/ledger/kvledger/tests/env.go
+++ b/core/ledger/kvledger/tests/env.go
@@ -217,7 +217,8 @@ func populateMissingsWithTestDefaults(t *testing.T, initializer *ledgermgmt.Init
 		}
 		cryptoProvider, err := sw.NewDefaultSecurityLevelWithKeystore(sw.NewDummyKeyStore())
 		assert.NoError(t, err)
-		membershipInfoProvider := privdata.NewMembershipInfoProvider(createSelfSignedData(cryptoProvider), identityDeserializerFactory)
+		mspID := "test-mspid"
+		membershipInfoProvider := privdata.NewMembershipInfoProvider(mspID, createSelfSignedData(cryptoProvider), identityDeserializerFactory)
 		initializer.MembershipInfoProvider = membershipInfoProvider
 	}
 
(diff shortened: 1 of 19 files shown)


core/ledger/kvledger/tests/env.go
```diff
@@ -217,7 +218,8 @@ func populateMissingsWithTestDefaults(t *testing.T, initializer *ledgermgmt.Init
 		}
 		cryptoProvider, err := sw.NewDefaultSecurityLevelWithKeystore(sw.NewDummyKeyStore())
 		assert.NoError(t, err)
-		membershipInfoProvider := privdata.NewMembershipInfoProvider(createSelfSignedData(cryptoProvider), identityDeserializerFactory)
+		mspID := viper.GetString("peer.localMspId")
```
mastersingh24: Do we really need to bring in viper here?  Can't we just set the mspID to a test value?

core/ledger/kvledger/tests/env.go
```diff
@@ -29,6 +29,7 @@ import (
 	"github.com/hyperledger/fabric/msp"
 	"github.com/hyperledger/fabric/msp/mgmt"
 	"github.com/hyperledger/fabric/protoutil"
+	"github.com/spf13/viper"
```
mastersingh24: see my other comment, but do we really need to use viper?   I actually worked pretty hard to remove viper last year

core/ledger/kvledger/tests/env.go
caod123: Probably not, this was ported from 1.4 when it was still using viper so I probably overlooked this when cherry-picking. I'll remove it, thanks for catching that!

core/ledger/kvledger/tests/env.go
caod123: sure we can remove it from this file, but anything that uses msp/mgmt will still use viper in its dependencies (creating self signed data requires `LoadLocalMSP`, etc)

mastersingh24: agreed .... someday we will get to that as well.  thx

mastersingh24 (approved):
LGTM

denyeart:
This passed my personal trials on master as well, so let's backport to release-2.0 for v2.0.1:

PR #675, opened 2020-02-15
Remove dead and irrelevant Jenkins references

lindluni:
Updates doc to remove references to Jenkins, many of these were dead before we ever moved off of Jenkins. Some will need to be replaced with future tooling, but nothing exists today to replace them with.

### DIFF
--- a/integration/README.rst
+++ b/integration/README.rst
@@ -111,9 +111,7 @@ description.
 
 Continuous Integration (CI) Execution
 -------------------------------------
-There is a target in the Hyperledger Fabric Makefile for executing `integration`_ tests.
-
-.. _integration: https://jenkins.hyperledger.org/view/fabric/job/fabric-verify-integration-tests-x86_64
+There is a target in the Hyperledger Fabric Makefile for executing integration tests.
 
 To execute the integration tests using the Makefile, execute the following:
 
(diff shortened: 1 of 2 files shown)


ale-linux (commented):
Isn't there an equivalent link for the new CI?

lindluni:
Not at the moment, but most of these links didn't make sense anyway, for example, the link to the coverage report was just a link to the Unit Test job page. You then had to choose a random unit test run, and the code coverage report was really just the output of `go test -cover ./...` that was in the log of everyones job run. 

The integration test link in the doc really had no context around it, so it made no sense.

For the rest of them there is no equivalent. We used to, for whatever reason, run BYFN as part of EVERY SINGLE REPO. Both SDK's, the chaincodes, Fabric and Fabric-CA all ran the same BYFN test scenario nightly and weekly and then we reported on that. Those links were just links to the results for all of these jobs, which were again, worthless.

The 4 links around SVT results, haven't been active for years.

lindluni:
In the future, we do plan to have a dashboard for fabric-test results with real information (not just saying whether SVT tests passed or not, as that provides no context into the actual quality of the code) but we aren't there yet

ale-linux (approved):
Understood - thx! LGTM

PR #679, opened 2020-02-16
Remove redundant tests from core/cc package

sykesm:
The integration style tests in the chaincode package take a very long time to run as they compile and run chaincode in docker containers. Given we already have project level integration tests that exercise this code, we can remove these without taking on any risk.


core/chaincode/chaincode_support_test.go
```diff
@@ -1299,3 +1311,184 @@ func TestMaxDuration(t *testing.T) {
...
+
+//getting a crash from ledger.Commit when doing concurrent invokes
+//It is likely intentional that ledger.Commit is serial (ie, the real
+//Committer will invoke this serially on each block). Mimic that here
+//by forcing serialization of the ledger.Commit call.
+//
+//NOTE-this should NOT have any effect on the older serial tests.
+//This affects only the tests in concurrent_test.go which call these
+//concurrently (100 concurrent invokes followed by 100 concurrent queries)
+var _commitLock_ sync.Mutex
```
yacovm, line 1356: I get that it was copied from **exectransaction_test.go**, but... ugh...

yacovm (approved):

PR #680, opened 2020-02-16
Reduce execution time of platforms/golang unit tests by disabling gzip compression

sykesm:
Packaging chaincode involves creating a compressed tar of the chaincode source and its transitive dependencies. In many cases, this is hundreds of files due to the dependency tree of grpc. It turns out that compressing hundreds of files takes a while and doing it multiple times
in tests results in very long test execution times.

By disabling gzip compression, the tests (running with race detection enabled) go from taking about 1m to about 10s.

This change creates an unexported package level var to hold the compression level to use that is initialized to gzip.DefaultCompression. An `init()` function was added to the test to explicitly set compression to gzip.NoCompression.

### DIFF
--- a/core/chaincode/platforms/golang/platform.go
+++ b/core/chaincode/platforms/golang/platform.go
@@ -108,6 +108,9 @@ func (p *Platform) ValidateCodePackage(code []byte) error {
 // Directory constant copied from tar package.
 const c_ISDIR = 040000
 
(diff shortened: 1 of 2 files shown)


lindluni:
Nice catch, simple fix and shaves some time off unit tests.

yacovm (approved):

PR #684, opened 2020-02-17
[FAB-17405]: Lifecycle CLI omits empty arrays from JSON output

wuxuer:
Description

When either "peer lifecycle chaincode queryinstalled -O json" or "peer
lifecycle chaincode querycommitted -O json" are called, they output a
JSON object that contains the results of the query. If the peer has no
installed or committed chaincodes, the array is also included in the
JSON.

Related issues

https://jira.hyperledger.org/browse/FAB-17405

#### Type of change

### DIFF
--- a/internal/peer/lifecycle/chaincode/common.go
+++ b/internal/peer/lifecycle/chaincode/common.go
@@ -7,11 +7,13 @@ SPDX-License-Identifier: Apache-2.0
 package chaincode
 
 import (
+	"bytes"
 	"context"
 	"encoding/json"
 	"fmt"
 	"io"
 
+	"github.com/golang/protobuf/jsonpb"
 	"github.com/golang/protobuf/proto"
 	pb "github.com/hyperledger/fabric-protos-go/peer"
 	"github.com/hyperledger/fabric/common/cauthdsl"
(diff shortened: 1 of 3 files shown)


sykesm (changes requested):
This needs further discussion. In fact, there isn't really agreement that the issue is really an issue at the moment.

Beyond that, the fix approach is very cumbersome. The tests also don't make use of the appropriate matchers.

wuxuer:
> This needs further discussion. In fact, there isn't really agreement that the issue is really an issue at the moment.
> 
> Beyond that, the fix approach is very cumbersome. The tests also don't make use of the appropriate matchers.

OK, thx. Then I close the PR now. 
If the issueis identified, consider submitting it at that time.

sykesm:
> OK, thx. Then I close the PR now.
> If the issueis identified, consider submitting it at that time.

While I didn't say it earlier, thank you for jumping in on this. Your effort is appreciated.

wuxuer:
> The tests also don't make use of the appropriate matchers.

And I have some doubts about your this proposal. 
Should I write a test function for function printResponseAsJSON instead of function Query, which calls function printResponseAsJSON?
Looking forward to your reply. Thank you!

PR #687, opened 2020-02-17
[FAB-17467] Create signature for config update

DereckLuo:

#### Type of change

#### Description
- additional fixes that addresses this PR: https://github.com/hyperledger/fabric/pull/669

#### Additional details
- Changed the structure name from **_Signer_** to **_SigningIdentity_**.
- Added parallel execution for all unit tests.
- Removed **_MarshalOrPanic_** function.
- Removed **_Public_** method from **_SigningIdentity_** , as it does not implement [crypto.Signer](https://golang.org/pkg/crypto/#Signer) anymore.
- Removed package level state **_publicKey_**, **_privateKey_** in unit tests. 

#### Related issues
[FAB-17467](https://jira.hyperledger.org/browse/FAB-17467)

### DIFF
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -8,16 +8,13 @@ package config
 
 import (
 	"crypto/rand"
-
-	"github.com/golang/protobuf/proto"
-
 	"errors"
 	"fmt"
 	"strings"
 	"time"
 
+	"github.com/golang/protobuf/proto"
 	"github.com/golang/protobuf/ptypes/timestamp"
-
 	cb "github.com/hyperledger/fabric-protos-go/common"
 	mb "github.com/hyperledger/fabric-protos-go/msp"
 )
@@ -121,9 +118,10 @@ func NewCreateChannelTx(profile *Profile, mspConfig *mb.FabricMSPConfig) (*cb.En
 	return env, nil
 }
 
-// SignConfigUpdate signs the given configuration update with a specific signer.
-func SignConfigUpdate(configUpdate *cb.ConfigUpdate, signer *Signer) (*cb.ConfigSignature, error) {
-	signatureHeader, err := signer.CreateSignatureHeader()
+// SignConfigUpdate signs the given configuration update with a
+// specific signing identity.
+func SignConfigUpdate(configUpdate *cb.ConfigUpdate, signingIdentity *SigningIdentity) (*cb.ConfigSignature, error) {
+	signatureHeader, err := signingIdentity.CreateSignatureHeader()
 	if err != nil {
 		return nil, fmt.Errorf("failed to create signature header: %v", err)
 	}
@@ -141,8 +139,7 @@ func SignConfigUpdate(configUpdate *cb.ConfigUpdate, signer *Signer) (*cb.Config
 	if err != nil {
 		return nil, fmt.Errorf("failed to marshal config update: %v", err)
 	}
-
-	configSignature.Signature, err = signer.Sign(rand.Reader, concatenateBytes(configSignature.SignatureHeader, configUpdateBytes))
+	configSignature.Signature, err = signingIdentity.Sign(rand.Reader, concatenateBytes(configSignature.SignatureHeader, configUpdateBytes))
 	if err != nil {
 		return nil, fmt.Errorf("failed to sign config update: %v", err)
 	}
@@ -551,11 +548,9 @@ func makePayloadHeader(ch *cb.ChannelHeader, sh *cb.SignatureHeader) (*cb.Header
 // concatenateBytes combines multiple arrays of bytes, for signatures or digests
 // over multiple fields.
 func concatenateBytes(data ...[]byte) []byte {
-	bytes := []byte{}
-	for _, d := range data {
-		for _, b := range d {
-			bytes = append(bytes, b)
-		}
+	var res []byte
+	for i := range data {
+		res = append(res, data[i]...)
 	}
-	return bytes
+	return res
 }
--- a/pkg/config/signer.go
+++ b/pkg/config/signer.go
@@ -7,7 +7,6 @@ SPDX-License-Identifier: Apache-2.0
 package config
 
 import (
-	"crypto"
 	"crypto/ecdsa"
 	"crypto/rand"
 	"crypto/x509"
@@ -23,10 +22,9 @@ import (
 	"github.com/hyperledger/fabric-protos-go/msp"
 )
 
-// Signer holds the information necessary to sign a configuration update.
-// It implements the crypto.Signer interface for ECDSA keys.
-type Signer struct {
-	cert       *x509.Certificate
+// SigningIdentity is an MSP Identity that can be used to sign configuration updates and transaction.
+type SigningIdentity struct {
+	pemCert    []byte // pem encoded publicCert
 	mspID      string
 	privateKey *ecdsa.PrivateKey
 	publicKey  *ecdsa.PublicKey
@@ -36,11 +34,11 @@ type ecdsaSignature struct {
 	R, S *big.Int
 }
 
-// NewSigner creates a new signer for configuration updates. The publicCert
-// and privateKey should be pem encoded.
-func NewSigner(publicCert []byte, privateCert []byte, mspID string) (*Signer, error) {
+// NewSigningIdentity creates a new signing identity for configuration updates.
+// The publicCert and privateKey should be pem encoded.
+func NewSigningIdentity(publicCert []byte, privateKey []byte, mspID string) (*SigningIdentity, error) {
 	if mspID == "" {
-		return nil, errors.New("failed to create new signer, mspID can not be empty")
+		return nil, errors.New("failed to create new signingIdentity, mspID can not be empty")
 	}
 
 	cert, err := getCertFromPem(publicCert)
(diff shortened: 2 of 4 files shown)


pkg/config/signer_test.go
```diff
@@ -25,21 +25,21 @@ import (
...
-func TestNewSigner(t *testing.T) {
+func TestNewSigningIdentity(t *testing.T) {
+	t.Parallel()
+
+	publicKey, privateKey := generatePublicAndPrivateKey()
+
 	t.Run("success", func(t *testing.T) {
 		gt := NewGomegaWithT(t)
-		signer, err := NewSigner([]byte(publicKey), []byte(privateKey), "test-msp")
+		signingIdentity, err := NewSigningIdentity([]byte(publicKey), []byte(privateKey), "test-msp")
```
caod123: `publicKey` and `privateKey` are already byte slices, you don't need to cast them again

pkg/config/signer_test.go
```diff
@@ -106,52 +108,43 @@ func TestNewSigner(t *testing.T) {
...
-}
+	publicKey, privateKey := generatePublicAndPrivateKey()
 
-func TestPublic(t *testing.T) {
-	t.Run("success", func(t *testing.T) {
-		gt := NewGomegaWithT(t)
-		signer, err := NewSigner([]byte(publicKey), []byte(privateKey), "test-msp")
-		gt.Expect(err).NotTo(HaveOccurred())
+	gt := NewGomegaWithT(t)
+	signingIdentity, err := NewSigningIdentity([]byte(publicKey), []byte(privateKey), "test-msp")
```
caod123: See above comment and all places this is casted

pkg/config/signer_test.go
```diff
@@ -106,52 +108,43 @@ func TestNewSigner(t *testing.T) {
...
-		gt.Expect(err).NotTo(HaveOccurred())
-		expectedPublicKey, err := ecdsaPublicKeyImport(expectedCert)
-		gt.Expect(err).NotTo(HaveOccurred())
-		publicKey := signer.Public()
-		gt.Expect(publicKey).To(Equal(expectedPublicKey))
-	})
+	sBytes, err := signingIdentity.Serialize()
+	gt.Expect(err).NotTo(HaveOccurred())
+	serializedIdentity := &msp.SerializedIdentity{}
+	err = proto.Unmarshal(sBytes, serializedIdentity)
```
caod123, line 142: Unchecked err
`Expect(err).NotTo(HaveOccurred())`

pkg/config/signer_test.go
```diff
@@ -86,16 +86,18 @@ func TestNewSigner(t *testing.T) {
 			spec:        "empty mspID",
 			publicKey:   []byte(publicKey),
 			privateKey:  []byte(privateKey),
```
caod123: See above comment and all other places in this test where it's casted

pkg/config/signer.go
```diff
@@ -82,32 +81,30 @@ func (s *Signer) Serialize() ([]byte, error) {
...
-// ensures signatures are created with Low S values since Fabric normalizes
+// Sign performs ECDSA sign with SigningIdentity's private key on given digest.
+// It ensures signatures are created with Low S values since Fabric normalizes
 // all signatures to Low S.
 // See https://github.com/bitcoin/bips/blob/master/bip-0146.mediawiki#low_s
 // for more detail.
-func (s *Signer) Sign(reader io.Reader, digest []byte) (signature []byte, err error) {
+func (s *SigningIdentity) Sign(reader io.Reader, digest []byte) (signature []byte, err error) {
 	if reader == nil {
 		return nil, errors.New("failed to sign, reader can not be nil")
```
caod123: nit: The consumer of this function knows that it called Sign and should handle any further error handling/reporting with that string. We can simplify this error to `errors.New("reader can not be nil")`, test will need to be updated as well

pkg/config/signer.go
```diff
@@ -149,21 +146,17 @@ func (s *Signer) CreateSignatureHeader() (*common.SignatureHeader, error) {
 
 // createNonce generates a nonce using the crypto/rand package.
 func createNonce() ([]byte, error) {
-	nonce, err := getRandomNonce()
+	nonce := make([]byte, 24)
+
+	_, err := rand.Read(nonce)
 	if err != nil {
-		return nil, fmt.Errorf("failed to generate random nonce: %s", err)
+		return nil, fmt.Errorf("failed to get random bytes: %s", err)
```
caod123: nit: we use `%v` just about everywhere else already in this package for errors, might as well standardize 🤷‍♂️There's really not a huge different between `%s` and `%v` so I'm fine with whichever

pkg/config/signer.go
```diff
@@ -149,21 +146,17 @@ func (s *Signer) CreateSignatureHeader() (*common.SignatureHeader, error) {
...
-	return nonce, nil
-}
-
-func getRandomNonce() ([]byte, error) {
-	key := make([]byte, 24)
 
-	_, err := rand.Read(key)
 	if err != nil {
-		return nil, fmt.Errorf("failed to get random bytes: %s", err)
+		return nil, fmt.Errorf("failed to generate random nonce: %s", err)
```
caod123: same nit as above regarding `%s`

pkg/config/signer.go
```diff
@@ -149,21 +146,17 @@ func (s *Signer) CreateSignatureHeader() (*common.SignatureHeader, error) {
...
 	}
-	return key, nil
+	return nonce, nil
 }
 
 // When using ECDSA, both (r,s) and (r, -s mod n) are valid signatures.
```
caod123: nit: This function isn't exported so it doesn't matter that much but exported function comments for GoDoc generally need to start with the function name. I would try to reorganize this function comment to start with `toLowS ...`

pkg/config/signer_test.go
```diff
@@ -106,52 +108,43 @@ func TestNewSigner(t *testing.T) {
...
-	t.Run("nil private key", func(t *testing.T) {
-		gt := NewGomegaWithT(t)
-		_, err := ecdsaPrivateKeyImport(nil)
-		gt.Expect(err.Error()).To(ContainSubstring("invalid key type. The DER must contain an ecdsa.PrivateKey"))
-	})
+	t.Parallel()
+
+	gt := NewGomegaWithT(t)
+	_, err := ecdsaPrivateKeyImport(nil)
+	gt.Expect(err.Error()).To(ContainSubstring("invalid key type. The DER must contain an ecdsa.PrivateKey"))
```
caod123, line 127: I much more prefer matching the actual error as opposed to just checking for a substring

pkg/config/signer_test.go
DereckLuo: done

pkg/config/signer_test.go
DereckLuo: done

pkg/config/signer_test.go
DereckLuo: done

pkg/config/signer_test.go
DereckLuo: done

pkg/config/signer.go
DereckLuo: done

pkg/config/signer.go
DereckLuo: done

pkg/config/signer.go
DereckLuo: done

pkg/config/signer.go
DereckLuo: done

pkg/config/signer_test.go
DereckLuo: in this case, and previous ones, the error message contains bytes of data that causes the error, such as _invalid private key_ bytes. Therefore, we decided not to match the exact error message.

pkg/config/signer.go
```diff
@@ -18,13 +17,13 @@ import (
 	"io"
 	"math/big"
 
+	"github.com/gogo/protobuf/proto"
```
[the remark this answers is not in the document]
DereckLuo: good catch, done

pkg/config/config.go
```diff
@@ -8,24 +8,33 @@ package config
...
+// specific signingIdentity.
+func SignConfigUpdate(configUpdate *common.ConfigUpdate, signingIdentity *SigningIdentity) (*common.ConfigSignature, error) {
+	signatureHeader, err := signingIdentity.CreateSignatureHeader()
 	if err != nil {
 		return nil, err
 	}
 
+	sh, err := proto.Marshal(signatureHeader)
+	if err != nil {
+		return nil, fmt.Errorf("failed to sign config update, failed to marshal signature header: %v", err)
```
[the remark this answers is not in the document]
DereckLuo: done

pkg/config/signer.go
```diff
@@ -35,44 +34,44 @@ type ecdsaSignature struct {
 	R, S *big.Int
 }
 
-// NewSigner creates a new signer for configuration updates. The publicCert
-// and privateKey should be pem encoded.
-func NewSigner(publicCert []byte, privateCert []byte, mspID string) (*Signer, error) {
+// NewSigningIdentity creates a new signingIdentity for configuration updates.
```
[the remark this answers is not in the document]
DereckLuo: done

pkg/config/config.go
```diff
@@ -8,24 +8,33 @@ package config
...
+// specific signingIdentity.
+func SignConfigUpdate(configUpdate *common.ConfigUpdate, signingIdentity *SigningIdentity) (*common.ConfigSignature, error) {
+	signatureHeader, err := signingIdentity.CreateSignatureHeader()
 	if err != nil {
 		return nil, err
 	}
 
+	sh, err := proto.Marshal(signatureHeader)
+	if err != nil {
+		return nil, fmt.Errorf("failed to marshal signature header: %v", err)
```
cendhu: Nit. Need to avoid `fmt.Errorf()` as per [error handling guideline](https://hyperledger-fabric.readthedocs.io/en/release-2.0/error-handling.html). Use `github.com/pkg/errors` package instead.

pkg/config/signer.go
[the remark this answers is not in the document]
cendhu: +1 @wlahti The term `signer` and `signer's public/private` key are widely used as far as I know based on the famous crypto book that I have been reading -- `Applied Cryptography: Protocols, Algorithms and Source Code in C`. Hence, `Signer` looks more natural to me than `SigningIdentity` here. This is similar to having `Validator` or `Committer` as the struct name (which we have used).

pkg/config/config.go
wlahti: We're treating this package as a library and are avoiding as many dependencies as possible as well as fabric-specific error/logging preferences.

pkg/config/signer.go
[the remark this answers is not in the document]
sykesm: A signer signs things; this is more than a signer as it also includes the MSPID and the certificate that goes into the the signature header. The overloaded use of “signer” also adds confusion with respect to `crypto.Signer` from the standard library.

Within Fabric, across components, we refer to this collection of identity, cert, and key as a `SigningIdentity`. Hence the original request to change.

https://github.com/hyperledger/fabric/blob/43ff898e6c3bfe05876a600a878603c02d5094c2/msp/msp.go#L78

I believe Will’s comment was that the godoc should use the term “signing identity” instead of the type; if the type is used, it should at least be done in consistent case. He can clarify if I’m wrong.

pkg/config/config.go
sykesm: As Will stated, we are avoiding dependencies and projecting our coding standards into consuming environments.

pkg/config/signer.go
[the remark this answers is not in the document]
wlahti: Correct, I was simply referring to the godoc in my previous comment. Things look good to me now.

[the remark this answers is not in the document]
cendhu: Got it. Thanks, @sykesm, for the detailed explanation.

pkg/config/config.go
cendhu: Got it. Thanks, @wlahti.

pkg/config/signer.go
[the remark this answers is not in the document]
yacovm: Even though **GetSigningIdentity** exists in the MSP interface, in my opinion it creates an un-needed coupling between signing and verification.

What if you have an idemix MSP for a specific organization **OrgA** but not for the rest of the organizations?  If a peer from **OrgB** now gets an API call to verify an idemix signature it needs to ask from the idemix MSP, which means an instance of the said MSP exists, however - the implementation of the idemix MSP on any org other than **OrgA** is undefined since only **OrgA** can issue idemix proofs.

If you ask me - the signing part should have never been placed inside the MSP interface, but it should have been in its own interface.

sykesm (dismissed):

DereckLuo:
rebased to current master. will address additional requests in a separate PR.

sykesm (approved):
Comments will be handled offline.

PR #688, opened 2020-02-17
[FAB-17491] Do not disseminate pvtdata for other org's implicit collection

wenjianqiao:

#### Type of change
- Bug fix

#### Description
For implicit collection, a peer should only disseminate pvtdata for it own org's implicit collection. Update GenerateImplicitCollectionForOrg to set maxPeerCount to 0 if an implicit collection is for another org so that the peer will not disseminate pvtdata for other org's implicit collection.

#### Related issues
https://jira.hyperledger.org/browse/FAB-17491

### DIFF
--- a/integration/pvtdata/implicit_coll_test.go
+++ b/integration/pvtdata/implicit_coll_test.go
@@ -1,0 +1,147 @@
+/*
+Copyright IBM Corp All Rights Reserved.
+
+SPDX-License-Identifier: Apache-2.0
+*/
+
+package pvtdata
+
+import (
+	"encoding/base64"
+	"encoding/json"
+	"fmt"
+	"path/filepath"
+
+	"github.com/hyperledger/fabric/integration/chaincode/kvexecutor"
+	"github.com/hyperledger/fabric/integration/nwo"
+	"github.com/hyperledger/fabric/integration/nwo/commands"
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+
+	"github.com/tedsuo/ifrit"
+)
+
+var _ bool = Describe("PrivateData implicit collection", func() {
+	var (
+		network       *nwo.Network
+		process       ifrit.Process
+		orderer       *nwo.Orderer
+		testChaincode chaincode
+	)
+
+	BeforeEach(func() {
+		By("setting up the network")
+		network = initThreeOrgsSetup(false)
+
+		By("setting the pull retry threshold to 0 on all peers")
+		// set pull retry threshold to 0 to disable pulling
+		for _, p := range network.Peers {
+			core := network.ReadPeerConfig(p)
+			core.Peer.Gossip.PvtData.PullRetryThreshold = 0
+			network.WritePeerConfig(p, core)
+		}
+
+		By("starting the network")
+		process, orderer = startNetwork(network)
+
+		By("deploying new lifecycle chaincode")
+		testChaincode = chaincode{
+			Chaincode: nwo.Chaincode{
+				Name:        "kvexecutor",
+				Version:     "1.0",
+				Path:        components.Build("github.com/hyperledger/fabric/integration/chaincode/kvexecutor/cmd"),
+				Lang:        "binary",
+				PackageFile: filepath.Join(network.RootDir, "kvexcutor.tar.gz"),
+				Label:       "kvexcutor",
+				Sequence:    "1",
+			},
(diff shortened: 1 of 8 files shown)


cendhu:
I understand that this PR makes a peer not to disseminate pvtData that is targeted for other org's implicit collection during the endorsement. This is because only the owning organization's peer should disseminate pvtData to other peers within the organization, depending on the fault-tolerant requirements of that implicit collection (as per [FAB-17463](https://jira.hyperledger.org/browse/FAB-17463)).  This CR achieves this by setting `maxPeerCount` to 0 for an implicit collection config that belongs to other organizations. As a result, when the [DistributePrivateData()](https://github.com/hyperledger/fabric/blob/master/core/endorser/endorser.go#L229) is called by the endorser, the [Distribute()](https://github.com/hyperledger/fabric/blob/master/gossip/service/gossip_service.go#L263) function in gossip service would not disseminate the pvtData that belongs to other org's implicit collection. However, we seem to still store all the pvtData in the transient store by calling [StorePvtData()](https://github.com/hyperledger/fabric/blob/master/gossip/service/gossip_service.go#L268) from [DistirubtePrivateData()](https://github.com/hyperledger/fabric/blob/master/core/endorser/endorser.go#L229). 

Though we purge entries from the transient store after committing `transientstoreMaxBlockRetention` number of blocks, is it okay to still store the data on a non-member peer temporarily? I understand that the pvtData is visible to the non-member during the endorsement and hence, there is no leakage of pvtData. However, storing the same for a short period is not needed I think.

wenjianqiao:
@cendhu Thanks for your comment. The behavior you pointed out is same for general (explicit) collections -  i.e., pvtdata may be stored in transient store during endorsement no matter if the peer is a collection member. Talked to Dave and Manish. It is not high priority and out of scope of this jira.

cendhu:
Thanks, @wenjianqiao. In general, we can avoid storing the pvtData in transient store on non-members for both implicit and explicit collections as the usage is nil. Even during the gossip pull, it seems to fetch the missing pvtData from [collection members only](https://github.com/hyperledger/fabric/blob/master/gossip/privdata/pvtdataprovider.go#L550). Yes, I agree that it is not a high priority item (as there is no data leakage), a very minor issue, and out of the scope of this JIRA.

denyeart:
@cendhu Thanks, you are right. I was thinking that pulling peers first attempted to pull from endorsing peers. But looking closer I see that the peers that pull at commit time won't even attempt to pull from non-member peers, so I agree there is no reason for a peer that is not a member of collection to write to its transient store. I've opened https://jira.hyperledger.org/browse/FAB-17516 as a future improvement.

integration/pvtdata/implicit_coll_test.go
```diff
@@ -0,0 +1,147 @@
...
+	It("disseminates implicit collection data owned by the peer's org ", func() {
+		peer1 := network.Peer("Org1", "peer0")
+		peer2 := network.Peer("Org2", "peer0")
+
+		By("writing private data to org1's and org2's implicit collections")
+		writeInput := []kvexecutor.KVData{
+			{Collection: "_implicit_org_Org1MSP", Key: "org1_key1", Value: "org1_value1"},
+			{Collection: "_implicit_org_Org2MSP", Key: "org2_key1", Value: "org2_value1"},
+		}
+		writeImplicitCollection(network, orderer, testChaincode.Name, writeInput, peer1, peer2)
```
denyeart, line 77: Since Org1 and Org2 implicit collection data is written to both Org1 and Org2 peers, we'd expect the private data to be disseminated to the authorized peers, both before and after this change.
Maybe I'm missing something, but does this test actually test that the dissemination is suppressed on peers that are not part of the implicit collection?

I was expecting to see maybe endorsement on only one of the two orgs, then we could verify that the other org didn't receive the private data.

wenjianqiao: Talked to Dave offline. https://jira.hyperledger.org/browse/FAB-17463 is needed to finish the verification (when we can configure different requiredPeerCount/maxPeerCount values for different peers). Therefore, the test will be enhanced/finalized in FAB-17463.

denyeart: ok, i've verified for myself that the new logic is good, will look for the final integration test in the next PR.

denyeart (approved):

wenjianqiao:

lindluni:

PR #693, opened 2020-02-19
[FAB-17487] Create config group for organization

DereckLuo:
#### Type of change

- New feature

#### Description

 Generates a ConfigGroup representing an Organization

#### Related issues

[FAB-17487](https://jira.hyperledger.org/browse/FAB-17487)

### DIFF
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -554,3 +554,35 @@ func concatenateBytes(data ...[]byte) []byte {
 	}
 	return res
 }
+
+// generateOrgConfigGroup returns an config group for a organization.
+// It defines the crypto material for the organization (its MSP).
+// It sets the mod_policy of all elements to "Admins".
+func generateOrgConfigGroup(org *Organization, mspConfig *mb.MSPConfig) (*cb.ConfigGroup, error) {
+	var err error
+
+	orgGroup := newConfigGroup()
+	orgGroup.ModPolicy = AdminsPolicyKey
+
+	if org.SkipAsForeign {
+		return orgGroup, nil
+	}
+
+	if err = addPolicies(orgGroup, org.Policies, AdminsPolicyKey); err != nil {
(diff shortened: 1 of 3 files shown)


pkg/config/config.go
```diff
@@ -559,3 +558,35 @@ func concatenateBytes(data ...[]byte) []byte {
...
+	if err != nil {
+		return nil, fmt.Errorf("failed to add msp value: %v", err)
+	}
+
+	if len(org.OrdererEndpoints) > 0 {
+		err = addValue(orgGroup, endpointsValue(org.OrdererEndpoints), AdminsPolicyKey)
+		if err != nil {
+			return nil, fmt.Errorf("failed to add orderer endpoints value: %v", err)
+		}
+	}
```
caod123, line 585: This is specific to orderer orgs. It's fine doing that here since we share the same Organization type struct between all orgs (consortium, orderer, and application) but we should also consider using this path for application and consortium orgs. The consumer will pass the organization config determining whether orderer endpoints or anchor peers should be set so we shouldn't have to distinguish between what type of org we're creating a config group for.

pkg/config/config_test.go
```diff
@@ -389,3 +386,169 @@ func createOrdererStandardPolicies() map[string]*Policy {
...
+		org := baseProfile().Application.Organizations[0]
+		org.OrdererEndpoints = []string{"123.45.677:8080"}
+		configGroup, err := generateOrgConfigGroup(org, mspConfig)
+		gt.Expect(err).NotTo(HaveOccurred())
+
+		buf := bytes.Buffer{}
+		err = protolator.DeepMarshalJSON(&buf, configGroup)
+		gt.Expect(err).NotTo(HaveOccurred())
+
+		gt.Expect(buf.String()).To(Equal(expectedPrintOrg))
```
sykesm, line 485: Please use MatchJSON

sykesm (approved):
We can follow-up offline.

PR #698, opened 2020-02-19
[FAB-17488] Compute a ConfigUpdate 

DereckLuo:
Signed-off-by: Dereck Luo <dereckluo@gmail.com>

#### Type of change

- New feature

#### Description

Compute a ConfigUpdate from a base and modified configuration transaction.

#### Related issues

[FAB-17488](https://jira.hyperledger.org/browse/FAB-17488)

### DIFF
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -299,7 +299,7 @@ func addPolicies(cg *cb.ConfigGroup, policyMap map[string]*Policy, modPolicy str
 		case SignaturePolicyType:
 			sp, err := FromString(policy.Rule)
 			if err != nil {
-				return fmt.Errorf("invalid signature policy rule '%s' error: %v", policy.Rule, err)
+				return fmt.Errorf("invalid signature policy rule: '%s' error: %v", policy.Rule, err)
 			}
 
 			signaturePolicy, err := proto.Marshal(sp)
@@ -534,6 +534,7 @@ func makePayloadHeader(ch *cb.ChannelHeader, sh *cb.SignatureHeader) (*cb.Header
 	if err != nil {
 		return nil, fmt.Errorf("failed to marshal channel header: %v", err)
 	}
+
 	signatureHeader, err := proto.Marshal(sh)
 	if err != nil {
 		return nil, fmt.Errorf("failed to marshal signature header: %v", err)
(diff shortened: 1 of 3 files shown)


pkg/config/config_test.go
```diff
@@ -389,3 +390,93 @@ func createOrdererStandardPolicies() map[string]*Policy {
...
+
+	base := common.Config{}
+	updated := common.Config{}
+
+	for _, test := range []struct {
+		name        string
+		channelID   string
+		expectedErr string
+	}{
+		{
```
wlahti, line 880: Let's make sure we always include the field name instead of relying on the order of the fields.

pkg/config/config_test.go
```diff
@@ -389,3 +390,93 @@ func createOrdererStandardPolicies() map[string]*Policy {
...
 	return policies
 }
+
+func TestComputeUpdate(t *testing.T) {
+	gt := NewGomegaWithT(t)
+
+	value1Name := "foo"
+	value2Name := "bar"
+	base := common.Config{
+		ChannelGroup: &common.ConfigGroup{
```
wlahti: Let's reuse the cb import instead of having two names for the same import.

pkg/config/config_test.go
stephyee: 🛠

pkg/config/config_test.go
stephyee: 🛠

caod123:
rebased

pkg/config/config_test.go
```diff
@@ -276,6 +326,90 @@ func TestNewCreateChannelTxFailure(t *testing.T) {
...
+			testName: "When creating the default config template with SignatureTypePolicy and bad rule fails",
+			profileMod: func() *Profile {
+				profile := baseProfile()
+				profile.Policies[ReadersPolicyKey].Type = SignaturePolicyType
+				profile.Policies[ReadersPolicyKey].Rule = "ANYY Readers"
+				return profile
+			},
+			err: errors.New("failed to create default config template: failed to create new channel group: " +
+				"failed to add policies: invalid signature policy rule: 'ANYY Readers' error: Cannot transition " +
+				"token types from VARIABLE [ANYY] to VARIABLE [Readers]"),
```
sykesm, line 401: After looking at the production code, I wanted to see what this error looked like.  Sadly, it looks like I thought it would. The error from `FromString` is not particularly consumable.

Later in this cycle we are going to have to spend some time on errors. For now, this is fine.

pkg/config/config.go
```diff
@@ -299,7 +299,7 @@ func addPolicies(cg *cb.ConfigGroup, policyMap map[string]*Policy, modPolicy str
 		case SignaturePolicyType:
 			sp, err := FromString(policy.Rule)
```
sykesm, line 300: Does this mean we're copying all of cauthdsl? That could be a problem.

sykesm (approved):
I'm concerned about making a copy of `cauthdsl` in this package. By doing that, we have two copies to maintain but no benefits as we're still pulling in all of the deps we'd otherwise have.

If we really need `cauthdsl`, let's understand how to truncate its dependency tree and use it directly instead of copying it.

PR #704, opened 2020-02-21
[FAB-16957] Fix nil pointer panic flake in deliver_test

caod123:
#### Type of change

- Bug fix

#### Description

Seems like info.GetStart().GetSpecified() can somehow return nil on
subsequent calls. Let's just assign it before the first nil check and
use it for the rest of the assertion.

#### Related issues

[FAB-16957](https://jira.hyperledger.org/browse/FAB-16957)

### DIFF
--- a/orderer/common/cluster/deliver_test.go
+++ b/orderer/common/cluster/deliver_test.go
@@ -32,6 +32,7 @@ import (
 	"github.com/onsi/gomega"
 	"github.com/pkg/errors"
 	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
 	"go.uber.org/zap"
 	"go.uber.org/zap/zapcore"
 	"google.golang.org/grpc"
@@ -271,8 +272,11 @@ func (ds *deliverServer) addExpectProbeAssert() {
 
 func (ds *deliverServer) addExpectPullAssert(seq uint64) {
 	ds.seekAssertions <- func(info *orderer.SeekInfo, _ string) {
-		assert.NotNil(ds.t, info.GetStart().GetSpecified())
-		assert.Equal(ds.t, seq, info.GetStart().GetSpecified().Number)
+		seekPosition := info.GetStart()
+		require.NotNil(ds.t, seekPosition)
+		seekSpecified := seekPosition.GetSpecified()
+		require.NotNil(ds.t, seekSpecified)
+		assert.Equal(ds.t, seq, seekSpecified.Number)
 		assert.Equal(ds.t, info.ErrorResponse, orderer.SeekInfo_BEST_EFFORT)
 	}
 }


orderer/common/cluster/deliver_test.go
```diff
@@ -271,8 +271,9 @@ func (ds *deliverServer) addExpectProbeAssert() {
 
 func (ds *deliverServer) addExpectPullAssert(seq uint64) {
 	ds.seekAssertions <- func(info *orderer.SeekInfo, _ string) {
-		assert.NotNil(ds.t, info.GetStart().GetSpecified())
-		assert.Equal(ds.t, seq, info.GetStart().GetSpecified().Number)
+		seekSpecified := info.GetStart().GetSpecified()
+		assert.NotNil(ds.t, seekSpecified)
+		assert.Equal(ds.t, seq, seekSpecified.Number)
```
yacovm, line 279: how does this fix the nil pointer panic? If **seekSpecified** is nil then at line 276 it will still cause a panic.
We need to use **require** instead of **assert** to fail-fast.

yacovm (changes requested):
Left a comment

orderer/common/cluster/deliver_test.go
caod123: good catch, in the attached logs on the bug, it was panicking on the 2nd call to `info.GetStart().GetSpecified()` so I just assumed we didn't expect it to be nil on the first call before the nil check. I figured at the time if it only ever panicked on the 2nd call then caching the first call would be enough to handle the nil check but I do agree changing to `require` is necessary here.

yacovm (approved):

orderer/common/cluster/deliver_test.go
yacovm: Why don't we just require that **info.GetStart()** won't be nil as well?

orderer/common/cluster/deliver_test.go
```diff
@@ -271,8 +272,9 @@ func (ds *deliverServer) addExpectProbeAssert() {
 
 func (ds *deliverServer) addExpectPullAssert(seq uint64) {
 	ds.seekAssertions <- func(info *orderer.SeekInfo, _ string) {
-		assert.NotNil(ds.t, info.GetStart().GetSpecified())
-		assert.Equal(ds.t, seq, info.GetStart().GetSpecified().Number)
+		seekSpecified := info.GetStart().GetSpecified()
```
yacovm: Let's require **info.GetStart()** not to be nil as well.

yacovm (changes requested):
Left a comment

orderer/common/cluster/deliver_test.go
caod123: Sure, I actually originally had a check for both but when I looked at the code for the GetStart() function it looked somewhat unlikely that it would ever return nil if info itself wasn't already nil.

yacovm (approved):

PR #705, opened 2020-02-21
[FAB-17521] remove protoutil from policyparser tests.

DereckLuo:
#### Type of change

- Bug fix

#### Description
- removed protoutil import from policyparser tests in ./pkg/config/, replaced with proto import instead.

#### Related issues
[FAB-17521](https://jira.hyperledger.org/browse/FAB-17521)

### DIFF
--- a/pkg/config/policyparser_test.go
+++ b/pkg/config/policyparser_test.go
@@ -10,9 +10,9 @@ import (
 	"reflect"
 	"testing"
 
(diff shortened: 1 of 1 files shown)


sykesm (approved):
This is fine. The idea of MarshalOrPanic is fine as well - just not the dependency on that pile of code.

So, if you want to create a copy of the helper to simplify this stuff, that's perfectly reasonable.

PR #708, opened 2020-02-21
Implement function to add an org to an existing consortium

wlahti:
#### Type of change

- New feature

#### Description

Implement a function to add an organization to an existing consortium

#### Related issues

[FAB-17510](https://jira.hyperledger.org/browse/FAB-17510)

### DIFF
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -683,3 +683,43 @@ func createSignedEnvelopeWithTLSBinding(
 
 	return env, nil
 }
+
+// AddOrgToConsortium adds an org definition to a named consortium in a given
+// channel configuration.
+func AddOrgToConsortium(org *Organization, consortium, channelID string, config *cb.Config, mspConfig *mb.MSPConfig) (*cb.ConfigUpdate, error) {
+	if org == nil {
+		return nil, errors.New("organization is empty")
+	}
+	if consortium == "" {
+		return nil, errors.New("consortium is empty")
+	}
+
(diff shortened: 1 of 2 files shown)


pkg/config/config.go
```diff
@@ -559,3 +571,43 @@ func concatenateBytes(data ...[]byte) []byte {
 	}
 	return bytes
 }
+
+// AddOrgToConsortium adds an org definition to a named consortium in a given
+// channel configuration.
+func AddOrgToConsortium(org *Organization, consortium, channelID string, config *cb.Config, mspConfig *mb.MSPConfig) (*cb.ConfigUpdate, error) {
+	if org == nil {
+		return nil, errors.New("organization is empty")
```
C0rWin, line 691: nit: "organization is nil" or "organization was not provided"

pkg/config/config.go
```diff
@@ -559,3 +571,43 @@ func concatenateBytes(data ...[]byte) []byte {
...
+	}
+	if consortium == "" {
+		return nil, errors.New("consortium is empty")
+	}
+
```
C0rWin, line 696: a nit: but IMO worth adding checks for rest of the parameters? Briefly inspecting the code flow nil config or mspConfig potentially could lead to a panic. (I know this cannot really happen, but would add this for sanity check).

caod123:
@wlahti feel free to reset this back to master for the dependent commit in the stack, it's been merged now

PR #716, opened 2020-02-22
[FAB-17523] Endorsing peer was not honoring RequiredPeerCount

denyeart:
#### Type of change

- Bug fix

#### Description

If there were not enough known eligible peers to meet a private data
collection RequiredPeerCount, endorsement was succeeding rather
than returning an error.

This was a regression caused by FAB-15389 in v1.4.4.
FAB-15389 shifted entirely to per-peer dissemination where required
peers were tagged up front before dissemination, however there was
no overall check to ensure that RequiredPeerCount was met. Prior
to FAB-15389, the check was handled deeper in the gossip dissemination.
Gossip still checks that each selected required peer disseminates
the private data.

This fix adds the overall RequiredPeerCount check for cases where
there are not enough known eligible peers.

Also, improvements are made to variable names, error messages, and
comments to make dissemination code easier to understand.

#### Related issues

https://jira.hyperledger.org/browse/FAB-17523

### DIFF
--- a/gossip/privdata/distributor.go
+++ b/gossip/privdata/distributor.go
@@ -189,7 +189,7 @@ func (d *distributorImpl) computeDisseminationPlan(txID string,
 			logger.Debugf("Computing dissemination plan for collection [%s]", collectionName)
 			dPlan, err := d.disseminationPlanForMsg(colAP, colFilter, pvtDataMsg)
 			if err != nil {
-				return nil, errors.WithStack(err)
+				return nil, errors.WithMessagef(err, "could not build private data dissemination plan for chaincode %s and collection %s", namespace, collectionName)
 			}
 			disseminationPlan = append(disseminationPlan, dPlan...)
 		}
@@ -227,8 +227,17 @@ func (d *distributorImpl) disseminationPlanForMsg(colAP privdata.CollectionAcces
 	m := pvtDataMsg.GetPrivateData().Payload
 
 	eligiblePeers := d.eligiblePeersOfChannel(routingFilter)
-	identitySets := d.identitiesOfEligiblePeers(eligiblePeers, colAP)
 
+	// With the shift to per peer dissemination in FAB-15389, we must first check
+	// that there are enough eligible peers to satisfy RequiredPeerCount.
+	if (len(eligiblePeers)) < colAP.RequiredPeerCount() {
+		return nil, errors.Errorf("required to disseminate to at least %d peers, but know of only %d eligible peers", colAP.RequiredPeerCount(), len(eligiblePeers))
+	}
+
+	// Group eligible peers by org so that we can disseminate across orgs first
+	identitySetsByOrg := d.identitiesOfEligiblePeersByOrg(eligiblePeers, colAP)
+
+	// peerEndpoints are used for dissemination debug only
 	peerEndpoints := map[string]string{}
 	for _, peer := range eligiblePeers {
 		epToAdd := peer.Endpoint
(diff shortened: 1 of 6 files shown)


denyeart:
Integration test will fail until https://github.com/hyperledger/fabric/pull/721 is merged.

gossip/privdata/distributor.go
```diff
@@ -227,8 +227,17 @@ func (d *distributorImpl) disseminationPlanForMsg(colAP privdata.CollectionAcces
 	m := pvtDataMsg.GetPrivateData().Payload
 
 	eligiblePeers := d.eligiblePeersOfChannel(routingFilter)
-	identitySets := d.identitiesOfEligiblePeers(eligiblePeers, colAP)
 
+	// With the shift to per peer dissemination in FAB-15389, we must first check
+	// that there are enough eligible peers to satisfy RequiredPeerCount.
+	if (len(eligiblePeers)) < colAP.RequiredPeerCount() {
+		return nil, fmt.Errorf("required to disseminate to at least %d peers, but know of only %d eligible peers", colAP.RequiredPeerCount(), len(eligiblePeers))
```
cendhu: We seem to have a good set of methods to ensure that pvtData is never lost and also the block commit performance is not impacted. 

1. PvtData should be disseminated to the `requiredPeerCount` number of peers. Otherwise, the endorsement is rejected (fixed by this PR).
2. As per the collection's endorsement policy, by default, the PvtData is replicated across the transient stores.
3. If a peer does not find PvtData in the transient store during the commit time, it would fetch from other peers. If the peer could not fetch the required pvtData from any other peers, it would record that particular pvtData is missing and proceed with the commit. Later, the reconciler fetches the missing pvtData from other peers. 
4. Most BaaS platforms also provide HA  — one from IBP https://cloud.ibm.com/docs/blockchain-sw?topic=blockchain-sw-ibp-console-ha 

As there could be duplicate communications (i.e., every endorser disseminating the pvt data even to the peer which already has the pvtData -- happens when there are multiple endorsers), such an approach in WAN could be bad for the performance. Anyway, this needs to be benchmarked to see whether it is really a problem. If we find this to be a problem, we can keep track of the previous senders and reduce the number of dissemination (for the future).

gossip/privdata/distributor.go
```diff
@@ -238,23 +247,30 @@ func (d *distributorImpl) disseminationPlanForMsg(colAP privdata.CollectionAcces
...
-	if maximumPeerCount > 0 {
-		for _, selectionPeers := range identitySets {
+
+	// PHASE 1 - Select one peer from each eligible org
+	if maximumPeerRemainingCount > 0 {
+		for _, selectionPeersForOrg := range identitySetsByOrg {
+
+			// Peers are tagged as a required peer for RequiredPeerCount up front before dissemination.
+			// TODO It would be better to attempt dissemination to MaxPeerCount first, and then verify that enough sends were acknowledged to meet RequiredPeerCount.
 			required := 1
```
cendhu: Nit. 
```suggestion
			numOfAcksRequired := 1
```
or
```suggestion
			acksRequired := 1
```
Improves readability IMO.

gossip/privdata/distributor.go
cendhu: `fmt.Errorf `to `errors.New()`? https://hyperledger-fabric.readthedocs.io/en/master/error-handling.html

cendhu: I think the onus is on the admin to configure `requiredPeerCount` and `maxPeerCount` depending on the endorsement policy to avoid duplicate transmissions.

gossip/service/gossip_service.go
```diff
@@ -261,7 +261,8 @@ func (g *GossipService) DistributePrivateData(channelID string, txID string, pri
 	}
 
 	if err := handler.distributor.Distribute(txID, privData, blkHt); err != nil {
-		logger.Error("Failed to distributed private collection, txID", txID, "channel", channelID, "due to", err)
+		err := errors.Wrapf(err, "failed to distribute private collection, txID %s, channel %s", txID, channelID)
```
manish-sethi: you would want to use errors.WithMessage instead.  errors.Wrapf will create a fresh stack trace instead, the original err already contains the full stack.

gossip/privdata/distributor.go
```diff
@@ -189,7 +189,7 @@ func (d *distributorImpl) computeDisseminationPlan(txID string,
 			logger.Debugf("Computing dissemination plan for collection [%s]", collectionName)
 			dPlan, err := d.disseminationPlanForMsg(colAP, colFilter, pvtDataMsg)
 			if err != nil {
-				return nil, errors.WithStack(err)
+				return nil, errors.Wrapf(err, "could not build private data dissemination plan for chaincode %s and collection %s", namespace, collectionName)
```
manish-sethi: same as other place "errors.WithMessage"

gossip/privdata/distributor.go
manish-sethi: errors.New would be desired instead of fmt.Errorf

gossip/privdata/distributor.go
manish-sethi: This is a little dicey. Earlier, it was errors.WithStack... Ideally, we should use errors.New at the origin of the error and everywhere else either propagate the error or errors.WithMessage. This all of a sudden "errors.WithStack" at a random place does not leave with certainty what form of err is being received...

gossip/privdata/distributor.go
cendhu: @manish-sethi mistakenly I removed my comment before you did +1 :) 

`fmt.Errorf` to `errors.New()`? https://hyperledger-fabric.readthedocs.io/en/master/error-handling.html

denyeart: Makes sense. I'm looking forward to your private data perf test results.

denyeart: I went with errors.Errorf()

denyeart (commented):
Addressed comments.

manish-sethi (approved):

PR #717, opened 2020-02-22
[FAB-17463] Allow peer to override implicit collection dissemination properties

wenjianqiao:

#### Type of change

#### Description
Added the following config properties to peer's core.yaml for implicit collection
dissemination policy. When a peer endorses a proposal that writes to its own
implicit collection, these values override the default values.
- peer.gossip.pvtData.ImplicitCollectionDisseminationPolicy.requiredPeerCount
- peer.gossip.pvtData.ImplicitCollectionDisseminationPolicy.maxPeerCount

These properties are applicable to all channels the peer has joined. The implication
is that requiredPeerCount has to be smaller than the number of peers in a channel
that has the lowest numbers of peers from the organization.

#### Related issues
https://jira.hyperledger.org/browse/FAB-17463

### DIFF
--- a/gossip/privdata/config.go
+++ b/gossip/privdata/config.go
@@ -7,14 +7,16 @@ SPDX-License-Identifier: Apache-2.0
 package privdata
 
 import (
+	"fmt"
 	"time"
 
 	"github.com/spf13/viper"
 )
 
 const (
-	reconcileSleepIntervalDefault = time.Minute
-	reconcileBatchSizeDefault     = 10
+	reconcileSleepIntervalDefault         = time.Minute
+	reconcileBatchSizeDefault             = 10
+	implicitCollectionMaxPeerCountDefault = 1
 )
 
 // PrivdataConfig is the struct that defines the Gossip Privdata configurations.
@@ -26,6 +28,19 @@ type PrivdataConfig struct {
 	ReconcileBatchSize int
 	// ReconciliationEnabled is a flag that indicates whether private data reconciliation is enabled or not.
 	ReconciliationEnabled bool
+	// ImplicitCollectionDisseminationPolicy specifies the dessemination policy for the peer's own implicit collection.
+	ImplicitCollDisseminationPolicy ImplicitCollectionDisseminationPolicy
+}
+
+// ImplicitCollectionDisseminationPolicy specifies the dessemination policy for the peer's own implicit collection.
+// It is not applicable to private data for other organizations' implicit collections.
+type ImplicitCollectionDisseminationPolicy struct {
+	// RequiredPeerCount defines the minimum number of eligible peers to which each endorsing peer must successfully
+	// disseminate private data for its own implicit collection. Default is 0.
+	RequiredPeerCount int
+	// MaxPeerCount defines the maximum number of eligible peers to which each endorsing peer will attempt to
+	// disseminate private data for its own implicit collection. Default is 1.
+	MaxPeerCount int
 }
 
 // GlobalConfig obtains a set of configuration from viper, build and returns the config struct.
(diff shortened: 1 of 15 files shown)


sampleconfig/core.yaml
```diff
@@ -204,6 +204,18 @@ peer:
             # transaction's private data from other peers need to be skipped during the commit time and pulled
             # only through reconciler.
             skipPullingInvalidTransactionsDuringCommit: false
+            # implicitCollectionDisseminationPolicy specifies the dessemination policy for the peer's own implicit collection.
+            # When a peer endorses a proposal that writes to its own implicit collection, below values override the default values
+            # for disseminating private data.
+            # Note that it is applicable to all channels the peer has joined. The implication is that requiredPeerCount has to
+            # be smaller than the number of peers in a channel that has the lowest numbers of peers from the organization.
+            implicitCollectionDisseminationPolicy:
```
cendhu, line 212: I understand the scope of this PR (i.e., allow the user to configure the dissemination plan). The following is out of scope for this PR but would like to convey my point. It might be good to expose `memberOnlyRead` and `memberOnlyWrite` as the config parameter for implicit collection. Though the implicit collection is used by the chaincode lifecycle, there are ACL policies in place to restrict who can approve the chaincode definition, etc... However, for the user chaincode, the implicit collection allows any member to read and write as the default value of `memberOnlyRead` and `memberOnlyWrite` is false. Isn't it? On the other hand, as the collection configuration specified in the core.yaml is applicable to all channels, it may not make sense to expose these configs in core.yaml. Hence, it is better to leave the ACL policy to the chaincode. We might need to document this behavior (better to open a JIRA for this so that we don't miss).

sampleconfig/core.yaml
```diff
@@ -204,6 +204,18 @@ peer:
...
+               # requiredPeerCount defines the minimum number of eligible peers to which the peer must disseminate private data
+               # for its own implicit collection. Default value is 0.
+               requiredPeerCount: 0
```
cendhu, line 215: Given that this configuration is applicable for all channels the peer has joined, it might lead to misconfiguration and difficulties in configuration IMO. For example, assume an organization with 4 peers. The organization wants to join 4 channels as well. The peers could join the channels in the following way (depending on a load of each channel and the need for load balancing endorsement requests):
- All 4 peers are part of channel c1 and c4.
- Only peer-1 is part of c2.
- Both peer-2 and peer-3 are part of c3.   
In the end, c1 and c4 have 4 peers. c2 has one peer. c3 has two peers.
peer-1: c1, c2, c4
peer-2: c1, c3, c4
peer-3: c1, c3, c4
peer-4: c1, c4

Given this scenario, how easy or hard for an admin to set these values correctly? In peer-1, the required peer count must be 0 as c2 is only available in peer-1, and so on. As a result, the transaction that gets endorsed on peer-1 for channel c1 would not do any push of pvtData to other peers for the fault-tolerant. I am sure that reading the above configuration itself is confusing but deciding some configuration value based on that is hard too. 

I think that the fault-tolerant property should be handled by the peer implicitly as it knows how many peers have joined a channel, etc... We seem to follow a simple approach and leave the load to the user. This would make the usability a bit harder IMO.

wenjianqiao:

sampleconfig/core.yaml
denyeart: @cendhu Thanks... discussed with Manish and Jason and we decided memberOnlyRead and memberOnlyWrite is not strategic, we'll document in FAB-17373 that users should utilize chaincode logic instead.

sampleconfig/core.yaml
denyeart: @cendhu Thanks. Discussed with Manish last week. There is no precedent in core.yaml for making config channel specific.  Additionally, it is typical for an enterprise to have the same number of peers in each channel, e.g. they frequently have two or three peers in each channel for redundancy, HA, and load balancing. Therefore we concluded that the approach in this PR is a reasonable compromise to keep things simple that should work for most use cases.
There has been a desire to refactor peer configuration in the future, and that would be an opportunity to make configurations channel specific across the board, if there turns out to be demand for it.

MaxPeerCount could potentially be handled by the peer dynamically based on number of available peers as you suggest, but that would not work for RequiredPeerCount, since the whole point of RequiredPeerCount is to fail endorsement in the scenario where other peers become not available.

So, I'm good with the approach in this PR, even though it is sub-optimal.

sampleconfig/core.yaml
cendhu: Yeah, I agree. If we make it a peer config, It would be applicable for all chaincodes which is not desirable.

sampleconfig/core.yaml
cendhu: Thanks, Dave. The reasoning for the current approach make sense and the rest can be handled when we refactor peer configuration.

gossip/privdata/config.go
```diff
@@ -50,4 +65,24 @@ func (c *PrivdataConfig) loadPrivDataConfig() {
 
 	c.ReconciliationEnabled = viper.GetBool("peer.gossip.pvtData.reconciliationEnabled")
 
+	requiredPeerCount := viper.GetInt("peer.gossip.pvtData.implicitCollectionDisseminationPolicy.requiredPeerCount")
+
+	maxPeerCount := implicitCollectionMaxPeerCountDefault
+	if viper.Get("peer.gossip.pvtData.implicitCollectionDisseminationPolicy.maxPeerCount") != nil {
+		// allow override maxPeerCount to 0 that will effectively disable dissemination on the peer
+		maxPeerCount = viper.GetInt("peer.gossip.pvtData.implicitCollectionDisseminationPolicy.maxPeerCount")
+	}
```
denyeart, line 74: Why check for nil on maxPeerCount, but not on requiredPeerCount?

denyeart (commented):
I confirmed the properties work in my environment as well. Just one small question...

sampleconfig/core.yaml
```diff
@@ -204,6 +204,18 @@ peer:
...
+               # maxPeerCount defines the maximum number of eligible peers to which the peer must disseminate private data
+               # for its own implicit collection. Default value is 1.
```
denyeart: 'must' is not correct here... maxPeerCount is the number of peers attempted, not the number of peers that 'must' be disseminated to.

gossip/privdata/config.go
wenjianqiao: Thank you, Dave. 
requiredPeerCount default value is 0 but maxPeerCount default value is 1. For requiredPeerCount, no matter if it is nil or 0, it will be 0. However, for maxPeerCount, we have to differentiate nil vs. 0, because maxPeerCount will be 1 (default value) if it is nil but will be 0 if it is 0. When maxPeerCount is set to 0 in core.yaml, it means to disable dissemination.

sampleconfig/core.yaml
wenjianqiao: Good catch, fixed.

denyeart (approved):

PR #718, opened 2020-02-23
Parallelize Unit Tests in CI

lindluni:
Parallelizes unit tests. Beyond making unit take less time to execute, this also means if you have a failure you don't have to wait 40 minutes for units to rerun as you only have to rerun the failed test.

### DIFF
--- a/ci/azure-pipelines.yml
+++ b/ci/azure-pipelines.yml
@@ -48,6 +48,8 @@ jobs:
     dependsOn: VerifyBuild
     pool:
       vmImage: ubuntu-18.04
+    strategy:
+      parallel: 3
     steps:
       - template: install_deps.yml
       - checkout: self
--- a/scripts/run-unit-tests.sh
+++ b/scripts/run-unit-tests.sh
@@ -1,222 +1,94 @@
 #!/bin/bash
-
 # Copyright IBM Corp. All Rights Reserved.
 #
 # SPDX-License-Identifier: Apache-2.0
-
-set -e
-set -o pipefail
-
-base_dir="$(cd "$(dirname "$0")/.." && pwd)"
+set -eo pipefail
 
 # regexes for packages to exclude from unit test
 excluded_packages=(
-    "/integration(/|$)"
-)
-
-# packages that must be run serially
-serial_packages=(
-    "github.com/hyperledger/fabric/gossip/..."
-)
-
-# packages which need to be tested with build tag pluginsenabled
-plugin_packages=()
-
-# packages which need to be tested with build tag pkcs11
-pkcs11_packages=(
-    "github.com/hyperledger/fabric/bccsp/..."
+  "/integration(/|$)"
 )
 
 # packages that are only tested when they (or their deps) change
 conditional_packages=(
-    "github.com/hyperledger/fabric/gossip/..."
+  "github.com/hyperledger/fabric/gossip"
 )
 
-# join array elements by the specified string
-join_by() {
-    local IFS="$1"; shift
-    [ "$#" -eq 0 ] && return 0
-    echo "$*"
-}
+# run everything when profiling
+if [[ "${JOB_TYPE}" = "PROFILE" ]]; then
+    conditional_packages=()
+fi
 
-contains_element() {
-    local key="$1"; shift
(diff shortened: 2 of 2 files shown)


sykesm (changes requested):
Can we talk about this again? We now know where 25 minutes is being spent...

denyeart:
The parallel integration tests have been a large help to productivity. I'm definitely in favor of doing the same for unit tests. Of course, we should continue to try to improve unit test time and flakes in general, but let's move forward with the parallel unit tests as well.

yacovm:
I wish we had a pull request like this a few years ago.

sykesm:
I'm going to stand my ground here. It doesn't make sense to me to spin up 8-10 VMs to run tests that *developers* are expected to run locally.

If we want to improve things, let's attack the source of the problems. For unit tests, we know what needs to change but we seem to be unwilling to do it.

denyeart:
@sykesm The orderer unit tests have been failing a lot, and it is quite painful to rerun all the unit tests each time.
It is quite possible to do both - invest in unit test improvements AND parallelize in CI for quicker turnaround and re-test. I don't understand the argument that we can only do one or the other.

sykesm:
> The orderer unit tests have been failing a lot, and it is quite painful to rerun all the unit tests each time.

Then we need to focus on fixing them. Same is true of gossip. Most of those tests share the same authors and test approach where many concurrent go routines are used in a way that results in non-deterministic execution.

> It is quite possible to do both - invest in unit test improvements AND parallelize in CI

Agreed. However, other than gossip (which can't be run in parallel because of resource conflicts), the tests are already running in parallel (see the -p flag on `go test`). That means the effort should be focused on test improvements instead of distributing the parallelism across virtual machines.

This is not a new discussion. I have been very vocal about this for two years. I fundamentally oppose a culture where we make it easy for maintainers to ignore persistent problems by simply clicking a button to re-run tests.

Facts:
- We currently have 27 items labled `ci-flakes` on the board.
- 19 `ci-flake` issues have been opened since December 2019. Most have been opened by @caod123, @btl5037, or @sykesm (me). One was opened by @denyeart. 
- 3 `ci-flake` issues have been resolved with a fix
- 1 `ci-flake` was closed as a duplicate
- Only 1-2 two flakes were updated with information about a recurrence.

Given "orderer unit tests have been failing a lot", those facts imply that maintainers aren't looking into the flakes.

In terms of test execution, [FAB-17508](https://jira.hyperledger.org/browse/FAB-17508) highlights one example of a watermelon in our tests. While the elapsed time in CI will differ from what I observed in my VM, simply caching a configuration object in our tests brings the execution time of the units down to [17m33s](https://dev.azure.com/Hyperledger/Fabric/_build/results?buildId=8376&view=logs&jobId=a11cdf60-0d11-5130-c801-1b4953009464). Compare that to the   [15m27s](https://dev.azure.com/Hyperledger/Fabric/_build/results?buildId=8322&view=logs&jobId=6e22d0ec-7766-52b0-d6b1-a8ff4f77545a&j=6e22d0ec-7766-52b0-d6b1-a8ff4f77545a) execution time of one of the parallel executions in this PR. Does the 2m6s difference really justify the consumption of 2 additional VMs? I don't believe it does.

In summary, before we try to "optimize" by doing multi-image parallelization

- let's fix the watermelons in the tests
- let's focus on improving the stability of the tests

denyeart:
222 ci-flakes have been opened since December 2018. I've opened 8 of them. That tells me that the remaining 27 are the difficult ones that require some serious time commitment. Regardless of the Jiras, we all know that gossip and orderer flakes are the frequent ones for recurrence.  The strategy of inflicting pain on all developers the last two years, in hopes that somebody will volunteer to fix these flakes, has not worked. It has only caused frustration and turned developers off. We need to change our approach. 

I propose:
1) Let's actually invest some dedicated resource in fixing the gossip and orderer flakes.
2) I am good to wait for FAB-17508.
3) If there is not a fundamental improvement a few weeks from now, let's proceed with the unit test parallelization.

lindluni:
The benefit of this is trivialized by #758. Thank you Matt for investigating and both you and Dereck for patching.

Closing

PR #737, opened 2020-02-25
Signing-related cleanup in pkg/config

wlahti:
#### Type of change

#### Description

Code cleanup/improvements to signing-related functions. Also, fix a data race in signer_test.go.

#### Related issues

### DIFF
--- a/pkg/config/signer.go
+++ b/pkg/config/signer.go
@@ -7,96 +7,25 @@ SPDX-License-Identifier: Apache-2.0
 package config
 
 import (
+	"crypto"
 	"crypto/ecdsa"
-	"crypto/rand"
 	"crypto/x509"
 	"encoding/asn1"
-	"encoding/pem"
-	"errors"
 	"fmt"
 	"io"
 	"math/big"
-
-	"github.com/golang/protobuf/proto"
-	"github.com/hyperledger/fabric-protos-go/common"
-	"github.com/hyperledger/fabric-protos-go/msp"
 )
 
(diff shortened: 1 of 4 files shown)


pkg/config/signer.go
```diff
@@ -7,155 +7,53 @@ SPDX-License-Identifier: Apache-2.0
...
-// SigningIdentity is an MSP Identity that can be used to sign configuration updates and transaction.
+// SigningIdentity is an MSP Identity that can be used to sign configuration
+// updates.
 type SigningIdentity struct {
-	pemCert    []byte // pem encoded publicCert
-	mspID      string
-	privateKey *ecdsa.PrivateKey
-	publicKey  *ecdsa.PublicKey
+	Certificate *x509.Certificate
+	PrivateKey  crypto.PrivateKey
```
DereckLuo, line 23: crypto.PrivateKey is an empty interface. Should we define privateKey type as crypto.Signer? as ecdsa.privateKey implement crypto.Signer.

DereckLuo (commented):
func (s *SigningIdentity) Public was removed from previous PR. We might want to add this in, so SigningIdentity will implement crypto.Signer.

sykesm:
Looks like a success notification was dropped from azure pipelines.

pkg/config/signer.go
sykesm: the interface is good here; while we really only handle ecdsa, it more accurately describes the dependency

sykesm (dismissed):

sykesm (approved):

PR #739, opened 2020-02-25
[FAB-17533] Move policy dsl related logic to separate package

caod123:
#### Type of change

#### Description

As part of the new config work, we want to minimize our dependencies on third party packages. Initially `Knetic/govaluate` was being imported by the config package via copied code from fabric's `cauthdsl` package. In an effort to reduce duplication, this PR moves the policyparser code that the config package is dependent on out to a separate shared package by `cauthdsl` and `config` that satisfies the minimal dependencies constraint of the config package.

#### Additional details

* Also consume this new policyparser package in pkg/config
* Move cauthdsl.SignedBy, cauthdsl.Or, cauthdsl.And, and cauthdsl.NOutOf
to policyparser package to avoid cyclic dependencies

#### Related issues

[FAB-17533](https://jira.hyperledger.org/browse/FAB-17533)


sykesm (commented):
I'm a bit confused about why the diff is showing us removing the original code and then moving the forked code on top. Seems weird. Perhaps this should have been two commits; one to move the common/cauthdsl code around and update the users and the other to remove the fork.

In context, `policyparser` does not look a good name. While `FromString` does parse a policy, the rest of the code is actually the DSL implementation. We see `And` and `NOutOf` as functions.

Stepping back even further, what is left in cauthdsl that we're not using?
In terms of depdencies, are we going to do something about logging as well to prevent pulling in the fabric logging component?

caod123:
addressed comments in latest commit stack

lindluni:
Rebase to pick up the bad merge fix

sykesm:
Integration test failure is [FAB-17535](https://jira.hyperledger.org/browse/FAB-17535).

common/policydsl/policydsl_builder.go
```diff
@@ -4,15 +4,14 @@ Copyright IBM Corp. 2016 All Rights Reserved.
 SPDX-License-Identifier: Apache-2.0
 */
 
-package cauthdsl
+package policydsl
```
sykesm, line 7: nit: the file name stutters.

it stuttered before and it still stutters. if you think of a better name, feel free to update it in a separate pr.

common/policydsl/policydsl_builder.go
```diff
@@ -28,26 +27,18 @@ var RejectAllPolicy *cb.SignaturePolicyEnvelope
...
 
 func init() {
-	var err error
-
 	AcceptAllPolicy = Envelope(NOutOf(0, []*cb.SignaturePolicy{}), [][]byte{})
-	MarshaledAcceptAllPolicy, err = proto.Marshal(AcceptAllPolicy)
-	if err != nil {
-		panic("Error marshaling trueEnvelope")
-	}
+	MarshaledAcceptAllPolicy = protoMarshalOrPanic(AcceptAllPolicy)
```
sykesm, line 31: nit: with the new `protoMarshalOrPanic` function, you shouldn’t need an explicit package init function. The var can simply be assigned to the result of calling the function with the result of Envelope(NOutOf(...))

sykesm (approved):

PR #740, opened 2020-02-25
cmd/peer tests fail due to port conflict

sykesm:
The PluginLoadingFailure test starts the peer without overriding the chaincode listen address. This results in a port conflict when running tests in parallel.

This change extends the pattern in the test that attempts to reserve an ephemeral port and then configures the peer to use it.


sykesm:
Opened [FAB-17535](https://jira.hyperledger.org/browse/FAB-17535) for the integration test flake.

denyeart (approved):

PR #742, opened 2020-02-25
[FAB-17529] Reduce and clarify errors in pkg/config

stephyee:

#### Type of change

#### Description

This change focuses on removing/rewording overly descriptive error messages

#### Related issues

### DIFF
--- a/pkg/config/config_test.go
+++ b/pkg/config/config_test.go
@@ -323,8 +323,8 @@ func TestNewCreateChannelTxFailure(t *testing.T) {
 				profile.Policies = nil
 				return profile
 			},
-			err: errors.New("failed to create default config template: failed to create new channel group: " +
-				"failed to add policies: no policies defined"),
+			err: errors.New("creating default config template: failed to add policies to " +
+				"channel group: no policies defined"),
 		},
 		{
 			testName: "When creating the default config template with no ApplicationGroupKey defined fails",
@@ -333,7 +333,7 @@ func TestNewCreateChannelTxFailure(t *testing.T) {
 				profile.Application = nil
 				return profile
 			},
-			err: errors.New("failed to create default config template: channel template config must contain " +
+			err: errors.New("creating default config template: channel template config must contain " +
 				"an application section"),
 		},
 		{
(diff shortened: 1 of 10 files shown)


pkg/config/config_test.go
```diff
@@ -323,8 +323,8 @@ func TestNewCreateChannelTxFailure(t *testing.T) {
 				profile.Policies = nil
 				return profile
 			},
-			err: errors.New("failed to create default config template: failed to create new channel group: " +
-				"failed to add policies: no policies defined"),
+			err: errors.New("creating default config template: failed to add policies to " +
+				"channel group: no policies defined"),
```
sykesm, line 327: super nit: it's much easier to read and search these when they aren't split across lines.

pkg/config/config_test.go
```diff
@@ -384,8 +384,8 @@ func TestNewCreateChannelTxFailure(t *testing.T) {
 				profile.Policies[ReadersPolicyKey].Rule = "ANYY Readers"
 				return profile
 			},
-			err: errors.New("failed to create default config template: failed to create new channel group: " +
-				"failed to add policies: invalid implicit meta policy rule: 'ANYY Readers' error: unknown rule type " +
+			err: errors.New("creating default config template: failed to add policies to channel group: " +
+				"invalid implicit meta policy rule: 'ANYY Readers': unknown rule type " +
 				"'ANYY', expected ALL, ANY, or MAJORITY"),
```
sykesm, line 389: seems very verbose but not worth changing right now.

sykesm (approved):
I'm still not sold on the use of "profile" in this code but that's something we can discuss as a follow-on.

pkg/config/config_test.go
stephyee: Fair enough, I've been splitting these on the suggestion by the linter.

pkg/config/config_test.go
stephyee: It's been trimmed down some. Is it reasonable for the consumer to know which rules are valid?

stephyee:
Agreed - there is a follow up work to rename profile to something more appropriate.

PR #749, opened 2020-02-26
[FAB-16233] Mitigate orderer/common/cluster/deliver_test.go

stephyee:
Signed-off-by: Danny Cao <dcao@us.ibm.com>

#### Type of change

- Test update

#### Description

Mitigate the timeout failure that occurs when the seekAssertions channel does not receive

#### Related issues

[FAB-17542](https://jira.hyperledger.org/browse/FAB-17542)

### DIFF
--- a/orderer/common/cluster/deliver_test.go
+++ b/orderer/common/cluster/deliver_test.go
@@ -180,9 +180,19 @@ func (ds *deliverServer) Deliver(stream orderer.AtomicBroadcast_DeliverServer) e
 	if err != nil {
 		panic(err)
 	}
+
+	// FAB-16233 This is meant to mitigate timeouts when
+	// seekAssertions does not receive a value
+	timer := time.NewTimer(1 * time.Minute)
+	defer timer.Stop()
+
+	select {
+	case <-timer.C:
+		panic("timed out waiting for seek assertions to receive a value")
 	// Get the next seek assertion and ensure the next seek is of the expected type
-	seekAssert := <-ds.seekAssertions
-	seekAssert(seekInfo, channel)
+	case seekAssert := <-ds.seekAssertions:
+		seekAssert(seekInfo, channel)
+	}
 
 	if seekInfo.GetStart().GetSpecified() != nil {
 		return ds.deliverBlocks(stream)


orderer/common/cluster/deliver_test.go
```diff
@@ -180,9 +180,23 @@ func (ds *deliverServer) Deliver(stream orderer.AtomicBroadcast_DeliverServer) e
 	if err != nil {
 		panic(err)
 	}
+
+	// FAB-16233 This is meant to mitigate timeouts when
+	// seekAssertions does not receive a value
+	timeoutC := make(chan bool, 1)
```
sykesm: We can do better. In addition to leaking a bunch of go routines, it’s a ton of boilerplate that is unnecessary; just use a Timer.

timer := time.NewTimer(3*time.Minute)
defer timer.Stop()

The select can then use timer.C.

sykesm (changes requested):

orderer/common/cluster/deliver_test.go
```diff
@@ -180,9 +180,23 @@ func (ds *deliverServer) Deliver(stream orderer.AtomicBroadcast_DeliverServer) e
...
+
+	go func() {
+		time.Sleep(3 * time.Minute)
+		timeoutC <- true
+	}()
+
+	select {
+	case <-timeoutC:
+		panic("timed out waiting for replicate chains to finish")
```
yacovm: why not assert.Fatalf or something that is related to the test framework?

caod123: Does assert have a Fatal that doesn't require consuming a testing.T?

sykesm: IIRC, this deliverServer instance is running in a go routine that is not managed by go test’s testing framework. If you use it, it simply delegates to T.FailNow which can’t be used from any go routine other than the one spawned by go test.

https://golang.org/pkg/testing/#T.FailNow

```
FailNow must be called from the goroutine running the test or benchmark function, not from other goroutines created during the test. Calling FailNow does not stop those other goroutines.
```

So, if there are any assertions like that in the deliver server, they’re also broken.

sykesm: For example, this code https://github.com/hyperledger/fabric/blob/5389a1d9696718cb7e93a20ca8b8a2cacff3360a/orderer/common/cluster/deliver_test.go#L634-L640

is calling resurrect from a go routine. If anything goes wrong in resurrect that results in this

https://github.com/hyperledger/fabric/blob/5389a1d9696718cb7e93a20ca8b8a2cacff3360a/orderer/common/cluster/deliver_test.go#L250

executing, it’s probably not going to work as you’d expect.

EDIT: It looks like it will fail the test but anything that results in FailNow is a different story.

From a glance at the tests, this pattern is all over.

orderer/common/cluster/deliver_test.go
```diff
@@ -180,9 +180,19 @@ func (ds *deliverServer) Deliver(stream orderer.AtomicBroadcast_DeliverServer) e
...
 	}
+
+	// FAB-16233 This is meant to mitigate timeouts when
+	// seekAssertions does not receive a value
+	timer := time.NewTimer(1 * time.Minute)
+	defer timer.Stop()
+
+	select {
+	case <-timer.C:
+		assert.Fail(ds.t, "timed out waiting for seek assertions to receive a value")
```
sykesm: Please read the other comments; we can try again.

sykesm (changes requested):

orderer/common/cluster/deliver_test.go
yacovm: > Does assert have a Fatal that doesn't require consuming a testing.T?

We have a testing.T in the instance as a field.

> IIRC, this deliverServer instance is running in a go routine that is not managed by go test’s testing framework. If you use it, it simply delegates to T.FailNow which can’t be used from any go routine other than the one spawned by go test.

What do you mean can't? It [still fails the test](https://play.golang.org/p/i5hPuxrPwSh), just not immediately. 
She has a select so it won't timeout, but it will fail (you just need to return an error or something). I just think that using the test framework to fail is more elegant than panicking, that's all.

sykesm: - The code already panics a few lines up from where this is being added.
- A fatal timeout is what we desire (ie. FailNow via require instead of assert). This is not asserting on a condition as much as mitigating test issues that are resulting in a hang.
- FailNow / require will not terminate tests out of a go routine.

I feel I need to stress that the only reason this is being added to terminate these tests in 3 minutes instead of 20 when things get locked up. While the preference would be to fix the cause(s) of the hangs, the scope is too large for people that have not worked in this area of the code.

sykesm (approved):

sykesm:
One small thing - this issue isn't fixing FAB-16233, it's mitigating it and associated with FAB-17542. So, this can close out FAB-17542 but it does not "fix" FAB-16233.

PR #760, opened 2020-02-27
[FAB-17543] Support adding anchor peers to existing config

caod123:
#### Type of change

- New feature

#### Description

Support adding anchor peers to an existing org

#### Related issues

[FAB-17543](https://jira.hyperledger.org/browse/FAB-17543)

### DIFF
--- a/pkg/config/application.go
+++ b/pkg/config/application.go
@@ -66,6 +66,50 @@ func newApplicationGroup(application *Application) (*cb.ConfigGroup, error) {
 	return applicationGroup, nil
 }
 
+// AddAnchorPeer adds an anchor peer to an existing channel config transaction.
+// It must add the anchor peer to an existing org and the anchor peer must not already
+// exist in the org.
+func AddAnchorPeer(config *cb.Config, orgName string, newAnchorPeer *AnchorPeer) error {
+	applicationOrgGroup, ok := config.ChannelGroup.Groups[ApplicationGroupKey].Groups[orgName]
+	if !ok {
+		return fmt.Errorf("application org %s does not exist in channel config", orgName)
+	}
+
+	anchorPeersProto := &pb.AnchorPeers{}
+
+	if anchorPeerConfigValue, ok := applicationOrgGroup.Values[AnchorPeersKey]; ok {
+		// Unmarshal existing anchor peers if the config value exists
+		err := proto.Unmarshal(anchorPeerConfigValue.Value, anchorPeersProto)
+		if err != nil {
+			return fmt.Errorf("failed unmarshalling %s's anchor peer endpoints: %v", orgName, err)
+		}
+	}
+
+	// Persist existing anchor peers if found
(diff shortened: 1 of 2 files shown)


pkg/config/application.go
```diff
@@ -66,6 +67,55 @@ func NewApplicationGroup(application *Application, mspConfig *mb.MSPConfig) (*cb
 	return applicationGroup, nil
 }
 
+// AddAnchorPeer adds an anchor peer to an existing channel config transaction.
+// It must add the anchor peer to an existing org and the anchor peer must not already
+// exist in the org.
+func AddAnchorPeer(config *cb.Config, orgName string, newAnchorPeer *AnchorPeer) error {
+	if _, orgExists := config.ChannelGroup.Groups["Application"].Groups[orgName]; !orgExists {
```
sykesm: 1. Here we use `"Application"` and below we use `ApplicationGroupKey`. It's inconsistent.
2. Why aren't we holding the result instead of just getting it again on line 78?

pkg/config/application.go
```diff
@@ -66,6 +67,55 @@ func NewApplicationGroup(application *Application, mspConfig *mb.MSPConfig) (*cb
...
+			return fmt.Errorf("failed unmarshalling existing anchor peers: %v", err)
+		}
+	}
+
+	anchorProtos := []*pb.AnchorPeer{}
+
+	// Persist existing anchor peers if found
+	if anchorPeersProto.AnchorPeers != nil {
+		anchorProtos = anchorPeersProto.AnchorPeers
+	}
```
sykesm: Why? If it's nil, the range won't do anything. The append will allocate if necessary.

pkg/config/application_test.go
```diff
@@ -132,6 +129,349 @@ func TestNewApplicationGroupSkipAsForeign(t *testing.T) {
...
+			orgName:       "Org1",
+			newAnchorPeer: &AnchorPeer{Host: "host1", Port: 123},
+			expectedErr:   "application org Org1 already contains anchor peer with host: host1, port: 123",
+		},
+	}
+
+	for _, tt := range tests {
+		tt := tt
+		t.Run(tt.testName, func(t *testing.T) {
+			t.Parallel()
```
sykesm, line 453: I don't know that the parallelization is buying anything here. Not necessarily opposed to it but we should set our expectations low.

sykesm (dismissed):
1. We should consistently use the `ApplicationGroupKey` constant if we're going to use it at all.
2. I don't know why we look in a map for a value to make sure it's there only to throw it away and get it again.
3. I don't understand why we need the intermediate values and explicit allocation of a slice in `AddAnchorPeers`. Those blocks are obscuring what should be an incredibly simple function.

pkg/config/application_test.go
caod123: I'll leave it in then, we can always revisit if we want to clean up the parallelization and it generally doesn't hurt.

sykesm (approved):

PR #761, opened 2020-02-27
[FAB-17529] Config package exported function and types comment hygiene

stephyee:

#### Type of change

#### Description
Clarifies comments for exported types and functions in the config package.

#### Related issues

### DIFF
--- a/pkg/config/application.go
+++ b/pkg/config/application.go
@@ -14,8 +14,8 @@ import (
 	pb "github.com/hyperledger/fabric-protos-go/peer"
 )
 
-// Application encodes the application-level configuration needed in config
-// transactions.
+// Application is a copy of the orderer configuration with the addition of an anchor peers
+// list in the organization definition.
 type Application struct {
 	Organizations []*Organization
 	Capabilities  map[string]bool
@@ -24,15 +24,15 @@ type Application struct {
 	ACLs          map[string]string
 }
 
-// AnchorPeer encodes the necessary fields to identify an anchor peer.
+// AnchorPeer defines the endpoint of peers for each application organization.
 type AnchorPeer struct {
 	Host string
 	Port int
 }
 
-// NewApplicationGroup returns the application component of the channel configuration.
-// By default, tt sets the mod_policy of all elements to "Admins".
-func NewApplicationGroup(application *Application, mspConfig *mb.MSPConfig) (*cb.ConfigGroup, error) {
+// newApplicationGroup returns the application component of the channel configuration.
+// By default, it sets the mod_policy of all elements to "Admins".
+func newApplicationGroup(application *Application, mspConfig *mb.MSPConfig) (*cb.ConfigGroup, error) {
 	var err error
 
 	applicationGroup := newConfigGroup()
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -24,7 +24,7 @@ import (
 	"github.com/hyperledger/fabric/common/policydsl"
 )
 
-// Channel encapsulates basic information for a channel config profile.
+// Channel is a channel configuration.
 type Channel struct {
 	Consortium   string
 	Application  *Application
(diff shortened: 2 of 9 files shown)


pkg/config/application.go
```diff
@@ -14,8 +14,8 @@ import (
 	pb "github.com/hyperledger/fabric-protos-go/peer"
 )
 
-// Application encodes the application-level configuration needed in config
-// transactions.
+// Application is a copy of the order configuration with the addition of an anchor peers
```
sykesm: "of the order configuration" ? Do you mean "orderer" configuration?

pkg/config/application.go
```diff
@@ -24,15 +24,15 @@ type Application struct {
 	ACLs          map[string]string
 }
 
-// AnchorPeer encodes the necessary fields to identify an anchor peer.
+// AnchorPeer defines the location of peers for each application organization.
```
sykesm: host:port is an address or an endpoint; not necessarily a "location"

pkg/config/application.go
```diff
@@ -24,15 +24,15 @@ type Application struct {
...
 type AnchorPeer struct {
 	Host string
 	Port int
 }
 
 // NewApplicationGroup returns the application component of the channel configuration.
```
sykesm: Doc command no longer matches the function name.

pkg/config/config.go
```diff
@@ -24,7 +24,7 @@ import (
 	"github.com/hyperledger/fabric/common/policydsl"
 )
 
-// Channel encapsulates basic information for a channel config profile.
+// Channel is the definition of the configuration for a channel.
```
sykesm: "the definition of the configuration" ? Seems a little hard to grok.

pkg/config/config.go
```diff
@@ -124,7 +125,7 @@ func NewCreateChannelTx(channelConfig *Channel, mspConfig *mb.FabricMSPConfig) (
 }
 
 // SignConfigUpdate signs the given configuration update with a
-// specific signing identity.
+// given specific signing identity and returns a config signature.
```
sykesm: "...with a given specific signing..."  ?

pkg/config/consortiums.go
```diff
@@ -25,7 +24,7 @@ type Consortium struct {
 // NewConsortiumsGroup returns the consortiums component of the channel configuration. This element is only defined for
```
sykesm: doc comment does not match the function name.

pkg/config/orderer.go
```diff
@@ -19,38 +19,49 @@ import (
 	eb "github.com/hyperledger/fabric-protos-go/orderer/etcdraft"
 )
 
-// Orderer encodes the orderer-level configuration needed in config transactions.
+// Orderer is the orderer implementation details.
```
sykesm: It doesn't hold implementation details. It may configure the ordering service behavior for the channel...

pkg/config/orderer.go
```diff
@@ -19,38 +19,49 @@ import (
...
+	// MaxMessageCount is the max message count.
+	MaxMessageCount uint32
+	// AbsoluteMaxBytes is the max block size (not including headers).
+	AbsoluteMaxBytes uint32
+	// PreferredMaxBytes is the preferred size of blocks.
 	PreferredMaxBytes uint32
 }
 
-// Kafka contains configuration for the Kafka-based orderer.
+// Kafka is an list of Kafka orderers.
```
sykesm: "is an list of" ?

sykesm (changes requested):
May want to enable the linter while doing some of the godoc work. It will point out when the function name is missing from doc and when the name in the doc does not match the function being documented.

Other than that, kudos for the improvements but some of the changes need a bit more word smithing.

sykesm (approved):

PR #762, opened 2020-02-27
Try to catch imports of gogo/proto during linting

sykesm:
List imports in fabric code and ensure gogo/proto is not directly imported.


lindluni:
+1

lindluni:
Hit known flake: added a comment on the Jira (which is already high priority)
https://jira.hyperledger.org/browse/FAB-17535

C0rWin (approved):

PR #765, opened 2020-02-28
[FAB-17544] Support removing anchor peers from an existing config

caod123:
#### Type of change

- New feature

#### Description

Support removing anchor peers from an existing org

#### Related issues

[FAB-17544](https://jira.hyperledger.org/browse/FAB-17544)

### DIFF
--- a/pkg/config/application.go
+++ b/pkg/config/application.go
@@ -9,6 +9,7 @@ package config
 import (
 	"fmt"
 
+	"github.com/golang/protobuf/proto"
 	cb "github.com/hyperledger/fabric-protos-go/common"
 	mb "github.com/hyperledger/fabric-protos-go/msp"
 	pb "github.com/hyperledger/fabric-protos-go/peer"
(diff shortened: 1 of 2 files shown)


pkg/config/application.go
```diff
@@ -66,6 +67,48 @@ func NewApplicationGroup(application *Application, mspConfig *mb.MSPConfig) (*cb
 	return applicationGroup, nil
 }
 
+// RemoveAnchorPeer removes an anchor peer from an existing channel config transaction.
+// The removed anchor peer and org it belongs to must both already exist.
+func RemoveAnchorPeer(config *cb.Config, orgName string, anchorPeerToRemove *AnchorPeer) error {
+	if _, orgExists := config.ChannelGroup.Groups["Application"].Groups[orgName]; !orgExists {
```
C0rWin: You might want to consider using `channelconfig.ApplicationGroupKey` instead of inline "Application" const.

pkg/config/application.go
```diff
@@ -66,6 +67,48 @@ func NewApplicationGroup(application *Application, mspConfig *mb.MSPConfig) (*cb
...
+		// Unmarshal existing anchor peers if the config value exists
+		err := proto.Unmarshal(anchorPeerConfigValue.Value, anchorPeersProto)
+		if err != nil {
+			return fmt.Errorf("failed unmarshalling existing anchor peers: %v", err)
+		}
+	}
+
+	existingAnchorPeers := anchorPeersProto.AnchorPeers
+
+	if len(existingAnchorPeers) == 0 {
```
C0rWin: I wonder whenever do you really needs this given that if anchor peers of length zero will skip for loop and return error at line 109. Unless you would like to indicate that no anchor peers was configured, when seems that added else clause will cover it.

pkg/config/application.go
caod123: good catch, forgot about that

pkg/config/application.go
caod123: good point, this error is probably not necessary as you mentioned the one on 109 would catch it and I think is meaningful enough to apply in this case as well

pkg/config/application.go
```diff
@@ -66,6 +67,44 @@ func NewApplicationGroup(application *Application, mspConfig *mb.MSPConfig) (*cb
...
 }
 
+// RemoveAnchorPeer removes an anchor peer from an existing channel config transaction.
+// The removed anchor peer and org it belongs to must both already exist.
+func RemoveAnchorPeer(config *cb.Config, orgName string, anchorPeerToRemove *AnchorPeer) error {
+	if _, orgExists := config.ChannelGroup.Groups[ApplicationGroupKey].Groups[orgName]; !orgExists {
+		return fmt.Errorf("application org %s does not exist in channel config", orgName)
+	}
+
+	applicationOrgGroup := config.ChannelGroup.Groups[ApplicationGroupKey].Groups[orgName]
```
sykesm: Reprise of the duplicate work. :)

C0rWin (approved):

PR #769, opened 2020-02-28
[FAB-17517] Only Initialize specified BCCSP provider

ahmedsajid:
#### Type of change

- Bug fix

#### Description

Code currently tries to initialize multiple providers based on
provided config Opts being nil or not.

This update ensures that only specified provider is initialized
based on ProviderName.

This fixes "Failed initializing PKCS11.BCCSP %!s(<nil>)" error
when the code complied with PKCS11 or PLUGINS enabled expects
configuration to not be nil even when Provider is set to SW.

Signed-off-by: Ahmed Sajid <ahmed.sajid@securekey.com>

#### Additional details

If this looks good I can do PR for release-2.0 and master.

#### Related issues

https://jira.hyperledger.org/browse/FAB-17517
#821 
#822

### DIFF
--- a/bccsp/factory/nopkcs11.go
+++ b/bccsp/factory/nopkcs11.go
@@ -35,48 +35,53 @@ type FactoryOpts struct {
 // Error is returned only if defaultBCCSP cannot be found
 func InitFactories(config *FactoryOpts) error {
 	factoriesInitOnce.Do(func() {
-		// Take some precautions on default opts
-		if config == nil {
-			config = GetDefaultOpts()
-		}
-
-		if config.ProviderName == "" {
-			config.ProviderName = "SW"
-		}
-
-		if config.SwOpts == nil {
-			config.SwOpts = GetDefaultOpts().SwOpts
-		}
-
-		// Initialize factories map
-		bccspMap = make(map[string]bccsp.BCCSP)
-
-		// Software-Based BCCSP
-		if config.SwOpts != nil {
-			f := &SWFactory{}
-			err := initBCCSP(f, config)
-			if err != nil {
-				factoriesInitError = errors.Wrapf(err, "Failed initializing BCCSP.")
-			}
-		}
-
-		// BCCSP Plugin
-		if config.PluginOpts != nil {
-			f := &PluginFactory{}
-			err := initBCCSP(f, config)
-			if err != nil {
-				factoriesInitError = errors.Wrapf(err, "Failed initializing PKCS11.BCCSP %s", factoriesInitError)
-			}
-		}
-
-		var ok bool
-		defaultBCCSP, ok = bccspMap[config.ProviderName]
-		if !ok {
-			factoriesInitError = errors.Errorf("%s\nCould not find default `%s` BCCSP", factoriesInitError, config.ProviderName)
-		}
+		factoriesInitError = initFactories(config)
 	})
 
 	return factoriesInitError
+}
+
+func initFactories(config *FactoryOpts) error {
+	// Take some precautions on default opts
+	if config == nil {
+		config = GetDefaultOpts()
+	}
+
+	if config.ProviderName == "" {
+		config.ProviderName = "SW"
+	}
+
+	if config.SwOpts == nil {
+		config.SwOpts = GetDefaultOpts().SwOpts
+	}
+
+	// Initialize factories map
+	bccspMap = make(map[string]bccsp.BCCSP)
+
+	// Software-Based BCCSP
+	if config.ProviderName == "SW" && config.SwOpts != nil {
+		f := &SWFactory{}
+		err := initBCCSP(f, config)
+		if err != nil {
+			return errors.Wrapf(err, "Failed initializing BCCSP.")
+		}
+	}
+
+	// BCCSP Plugin
+	if config.ProviderName == "PLUGIN" && config.PluginOpts != nil {
+		f := &PluginFactory{}
+		err := initBCCSP(f, config)
+		if err != nil {
+			return errors.Wrapf(err, "Failed initializing PLUGIN.BCCSP %s", factoriesInitError)
+		}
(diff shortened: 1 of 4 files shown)


denyeart:
The unit test failure is due to a gossip test flake, which is common in release-1.4.
Regardless, please add your own unit test to verify the fix and then we try to get all the unit tests to pass.

ahmedsajid:
@denyeart I have added unit tests.
Let me know anything else needs fixing.

gbolo:
Good job @ahmedsajid

fqutishat (approved):

lindluni:
The change looks good. But in general we open changes against the master branch and cherry pick them into earlier releases.

ahmedsajid:
@btl5037
The initial testing was done on v1.4.5.
Shall I open PR for Master and release-2.0?

bccsp/factory/nopkcs11.go
```diff
@@ -35,48 +35,53 @@ type FactoryOpts struct {
...
+			return errors.Wrapf(err, "Failed initializing PLUGIN.BCCSP %s", factoriesInitError)
 		}
-	})
+	}
 
-	return factoriesInitError
+	var ok bool
+	defaultBCCSP, ok = bccspMap[config.ProviderName]
+	if !ok {
+		return errors.Errorf("%s\nCould not find default `%s` BCCSP", factoriesInitError, config.ProviderName)
```
sykesm, line 82: Looks like you've simply copied a bad practice that exists in the code: errors are not messages and do not contain new lines.

bccsp/factory/pkcs11_test.go
```diff
@@ -41,6 +41,51 @@ func TestSetFactories(t *testing.T) {
...
+	err = setFactories(&FactoryOpts{
+		ProviderName: "PKCS11",
+		SwOpts:       &SwOpts{},
+		Pkcs11Opts:   &pkcs11.PKCS11Opts{},
+		PluginOpts:   &PluginOpts{},
+	})
+	assert.Error(t, err)
+	assert.Contains(t, err.Error(), "Failed initializing PKCS11.BCCSP")
+	assert.NotContains(t, err.Error(), "Failed initializing SW.BCCSP")
+	assert.NotContains(t, err.Error(), "Failed initializing PLUGIN.BCCSP")
```
sykesm, line 71: When we know the error, we use EqualError. That means we should have a single assertion here instead of 3.

bccsp/factory/nopkcs11_test.go
```diff
@@ -0,0 +1,51 @@
...
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+)
+
+func TestInitFactoriesWithMultipleProviders(t *testing.T) {
+	// testing SW Provider and ensuring other providers are not initialized
+	factoriesInitError = nil
```
sykesm, line 28: You're calling the unexported method - why is this required?

bccsp/factory/nopkcs11_test.go
```diff
@@ -0,0 +1,51 @@
...
+		ProviderName: "SW",
+		SwOpts:       &SwOpts{},
+		PluginOpts:   &PluginOpts{},
+	})
+	assert.Error(t, err)
+	assert.Contains(t, err.Error(), "Failed initializing BCCSP")
+	assert.NotContains(t, err.Error(), "Failed initializing PLUGIN.BCCSP")
+
+	// testing PLUGIN Provider and ensuring other providers are not initialized
+	factoriesInitError = nil
```
sykesm, line 40: Same thing here. Not sure why we need to reset the error when we're calling the unexported function directly.

bccsp/factory/nopkcs11.go
```diff
@@ -35,48 +35,53 @@ type FactoryOpts struct {
...
-		var ok bool
-		defaultBCCSP, ok = bccspMap[config.ProviderName]
-		if !ok {
-			factoriesInitError = errors.Errorf("%s\nCould not find default `%s` BCCSP", factoriesInitError, config.ProviderName)
+	// BCCSP Plugin
+	if config.ProviderName == "PLUGIN" && config.PluginOpts != nil {
+		f := &PluginFactory{}
+		err := initBCCSP(f, config)
+		if err != nil {
+			return errors.Wrapf(err, "Failed initializing PLUGIN.BCCSP %s", factoriesInitError)
```
sykesm, line 75: `factoriesInitError` should not be used here. It has not ben set yet.

bccsp/factory/nopkcs11.go
sykesm, line 82: `factoriesInitError` should not be used here. It has not been set yet. (It's set by the caller `Once`.)

bccsp/factory/nopkcs11_test.go
```diff
@@ -0,0 +1,51 @@
+// +build !pkcs11
+
+/*
+Copyright IBM Corp. 2016 All Rights Reserved.
```
sykesm, line 4: It's no longer 2016.

mastersingh24 (approved):

bccsp/factory/pkcs11.go
```diff
@@ -70,7 +70,7 @@ func setFactories(config *FactoryOpts) error {
 	}
 
 	// PKCS11-Based BCCSP
-	if config.Pkcs11Opts != nil {
+	if config.ProviderName == "PKCS11" && config.Pkcs11Opts != nil {
```
sykesm, line 73: The pkcs11 side should be made consistent with the nopkcs11 side. The `factoriesInitError` package state is not handled the same in the two methods. This is confusing and is causing odd constructs in the tests.

`factoriesInitError` should be set as a result of the once and not here.

sykesm (changes requested):
Comments are too late, I guess.

PR #770, opened 2020-02-28
Cleanup Makefile - remove unused env vars

denyeart:
#### Type of change

- Cleanup

#### Description

Remove unused env vars from Makefile, including:
- PREV_VERSION (also cleanup unused changelog Make target that referenced it)
- BASEIMAGE_RELEASE (as of v2.0, Fabric does not use baseimage)
- BASE_DOCKER_TAG (as of v2.0, Fabric does not use baseimage)
- BASE_DOCKER_NS (also remove from peer version info)

Also increment BASE_VERSION to prepare for v2.1.0.


denyeart:
Flake https://jira.hyperledger.org/browse/FAB-17535

Makefile
```diff
@@ -192,10 +187,6 @@ protos: gotools
 	@echo "Compiling non-API protos..."
 	./scripts/compile_protos.sh
 
-.PHONY: changelog
```
mastersingh24, line 195: How are we generating the change log now?

mastersingh24 (approved):

Makefile
denyeart: I've actually never used the make target for changelog or the PREV_VERSION variable... I've always just run the script directly, passing in the old and new tag, e.g.
`./scripts/changelog.sh v2.0.0 v2.0.1`

PR #774, opened 2020-03-01
Cleanup Makefile - remove unused env vars - release-2.0

denyeart:
#### Type of change

- Cleanup

#### Description

Remove unused env vars from Makefile, including:

- PREV_VERSION (also cleanup unused changelog Make target that referenced it)
- BASEIMAGE_RELEASE (as of v2.0, Fabric does not use baseimage)
- BASE_DOCKER_TAG (as of v2.0, Fabric does not use baseimage)
- BASE_DOCKER_NS (also remove from peer version info)

Also increment BASE_VERSION to prepare for v2.0.2 (if there is one...)


Makefile
```diff
@@ -192,10 +187,6 @@ protos: gotools
 	@echo "Compiling non-API protos..."
 	./scripts/compile_protos.sh
 
-.PHONY: changelog
-changelog:
-	./scripts/changelog.sh v$(PREV_VERSION) v$(BASE_VERSION)
```
sykesm, line 197: Does this mean we no longer need this script?

sykesm (approved):
Is the changelog script dead now? Looks like the release pipeline generates the changelog using azure tooling instead.

lindluni:
> Is the changelog script dead now? Looks like the release pipeline generates the changelog using azure tooling instead.

We discovered due to a limitation in the github API it's truncated at 2 or 3000 characters (can't remember which), so if you have a large changelog (i.e. 2.0) its not all there. So we still generate and attach it as a document

PR #777, opened 2020-03-02
Backport LabelHelp field to metrics option structs

saad-karim:
#### Type of change

Backport code from master to release 1.4

#### Description

Backport LabelHelp field to metrics option structs

### DIFF
--- a/common/metrics/provider.go
+++ b/common/metrics/provider.go
@@ -46,6 +46,10 @@ type CounterOpts struct {
 	// of these label names.
 	LabelNames []string
 
+	// LabelHelp provides help information for labels. When set, this information
+	// will be used to populate the documentation.
+	LabelHelp map[string]string
+
 	// StatsdFormat determines how the fully-qualified statsd bucket name is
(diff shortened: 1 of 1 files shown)


sykesm (commented):
Why are we doing this? The LabelHelp isn't consumed in Fabric 1.4.

sykesm (changes requested):
1. Why are we doing this? The LabelHelp isn't consumed in Fabric 1.4. When there is no consumer, it's hard to justify a backport.
2. If this *needs* to be pulled back, the backport is incomplete. The fields were added so they could be populated in the documentation.

Without 1, there's no need to backport. Without 2, there's no need to backport. Without 1&&2, there's really no need to backport.

saad-karim:
fabric-ca 1.4 uses `LabelHelp`, in IBP-Operator when we are vendoring 1.4.x version of fabric and fabric-ca. We get this compile time error: `lib/server/metrics/metrics.go:19:3: unknown field 'LabelHelp' in struct literal of type metrics.CounterOpts`

PR #781, opened 2020-03-02
Delete scripts/changelog.sh

sykesm:
This script is no longer used.


denyeart:
The script is still used... see the full discussion in https://github.com/hyperledger/fabric/pull/782.

mastersingh24:
Based on closing #782 , closing this as well for now

PR #782, opened 2020-03-02
Remove scripts/changelog.sh

sykesm:
This script is no longer referenced.


lindluni:
Dave still uses this to generate the changelog manually, we can't rely on the CI generation as the GitHub API truncates the changelog at 2 or 3,000 characters, so if its any larger than that the data isn't received and doesn't populate in the GitHub Release

lindluni:
I'm not sure removing it from the Makefile was the right choice (which I just noticed)

denyeart:
The GitHub generated changelog does not look correct in general. e.g. for the v2.0.1 release change log it includes many more commits than are in v2.0.1 (includes v2.0.0 changes until the truncation point). While our generated changelog from changelog.sh is correct.

That being said, I'm not going to argue for continued use of changelog.sh, if nobody else in the open source world feels a need to custom generate it.

@btl5037 why do you think changelog shouldn't be removed from makefile, if nobody was using it? It only forced us to maintain PREV_VERSION env variable that nobody else was using. I've always preferred to call changelog.sh directly and pass in the release tags myself, rather than depend on the env vars.

lindluni:
The changelog that we attach to our releases is generated using this script, when the time comes that fabric outlives our tenure, there is no obvious answer to how we generated our changelogs, it lives in a google doc we maintain out of band only.

That being said, if no one else has a problem removing it, I'm not going to put up a fight.

denyeart:
In summary -

If we want to keep the 'accurate' changelog per release, then we should keep the script, continue generating changelog in the release PR each release, and dismiss this PR.

If we are ok with the 'incorrect' changelog that github auto-generates, then we stop maintaining our own and merge this PR.

I would vote for the former, but don't feel too strong since the full commit and tag history is in github if somebody wants to investigate the detailed commit history  themselves.

PR #795, opened 2020-03-03
[FAB-17546, FAB-17545] Add rootCA to application org MSP, Add cert to revocation list

DereckLuo:
#### Type of change

- New feature

#### Description

- adding rootCA to application org MSP.
- adding certificate to application org revocation list.

#### Related issues

[FAB-17546](https://jira.hyperledger.org/browse/FAB-17546)
[FAB-17545](https://jira.hyperledger.org/browse/FAB-17545)


pkg/config/config.go
```diff
@@ -713,3 +713,7 @@ func createSignedEnvelopeWithTLSBinding(
 
 	return env, nil
 }
+
+func getOrgsFromApplication(config *cb.Config) map[string]*cb.ConfigGroup {
```
stephyee: nit: Given this helper is specific to application orgs we could move this application.go.

pkg/config/msp.go
```diff
@@ -0,0 +1,76 @@
...
+
+// addRootCAToOrgMSP adds a rootCA to org MSP config. It takes a pointer to
+// org MSP config, and modifies in place.
+func addRootCAToOrgMSP(rootCA *x509.Certificate, org *cb.ConfigGroup) error {
+	configValue := org.Values[MSPKey]
+
+	mspConfig := &mb.MSPConfig{}
+	err := proto.Unmarshal(configValue.Value, mspConfig)
+	if err != nil {
+		return fmt.Errorf("failed to unmarshal mspConfig: %v", err)
```
stephyee: We want to avoid chaining errors using **failed to**. For example, this error looks something like, `failed to add rootCA to application org: failed to unmarshal mspConfig: proto is nil or something"` .

pkg/config/msp.go
```diff
@@ -0,0 +1,110 @@
...
+		return nil, nil, fmt.Errorf("unmarshalling mspConfig: %v", err)
+	}
+
+	fabricMSPConfig := &mb.FabricMSPConfig{}
+	err = proto.Unmarshal(mspConfig.Config, fabricMSPConfig)
+	if err != nil {
+		return nil, nil, fmt.Errorf("unmarshalling mspConfig: %v", err)
+	}
+
+	return mspConfig, fabricMSPConfig, nil
```
caod123: Why are we returning both here? An `mb.MSPConfig` contains a single `Config` field with the marshaled `mb.FabricMSPConfig`, there's no need to return both. We can either return a MSPConfig with the marshaled FabricMSPConfig already inside it or just return the FabricMSPConfig and marshal it later down in `addMSPConfigToOrg`

pkg/config/msp.go
```diff
@@ -0,0 +1,110 @@
...
+}
+
+func addMSPConfigToOrg(org *cb.ConfigGroup, mspConfig *mb.MSPConfig, fabricMSPConfig *mb.FabricMSPConfig) error {
+	serializedMSPConfig, err := proto.Marshal(fabricMSPConfig)
+	if err != nil {
+		return fmt.Errorf("marshalling updated mspConfig: %v", err)
+	}
+
+	mspConfig.Config = serializedMSPConfig
```
caod123, line 609: See comment above

pkg/config/msp.go
```diff
@@ -0,0 +1,110 @@
...
+	if err != nil {
+		return fmt.Errorf("getting msp config: %v", err)
+	}
+
+	buffer := bytes.NewBuffer(nil)
+	err = pem.Encode(buffer, &pem.Block{Type: "CERTIFICATE", Bytes: rootCA.Raw})
+	if err != nil {
+		return err
+	}
+	fabricMSPConfig.RootCerts = append(fabricMSPConfig.RootCerts, buffer.Bytes())
```
caod123: Referring to my below comment, it looks like you still modify the FabricMSPConfig up here so you probably don't need to return the MSPConfig in the `getMSPConfig` function

pkg/config/msp.go
```diff
@@ -0,0 +1,110 @@
...
+// list of rootCerts for the specified application org MSP.
+func AddRootCAToMSP(rootCA *x509.Certificate, orgName string, config *cb.Config) error {
+	if (rootCA.KeyUsage & x509.KeyUsageCertSign) == 0 {
+		return errors.New("certificate KeyUsage must be x509.KeyUsageCertSign")
+	}
+	if !rootCA.IsCA {
+		return errors.New("certificate must be a CA certificate")
+	}
+
+	org := getOrgsFromApplication(config)[orgName]
```
caod123: I think this would read better as `getApplicationOrg(config, orgName)` and handle returning the specific org instead of a map

pkg/config/msp.go
```diff
@@ -0,0 +1,110 @@
...
+		return err
+	}
+
+	return nil
+}
+
+// RevokeCertificateFromMSP takes an x509 certificate and adds it to the
+// revocation list for the specified application org MSP.
+func RevokeCertificateFromMSP(cert *x509.Certificate, orgName string, config *cb.Config) error {
+	org := getOrgsFromApplication(config)[orgName]
```
caod123: See comment above about `getOrgsFromApplication`

pkg/config/application.go
```diff
@@ -169,3 +169,9 @@ func anchorPeersValue(anchorPeers []*pb.AnchorPeer) *standardConfigValue {
 		value: &pb.AnchorPeers{AnchorPeers: anchorPeers},
 	}
 }
+
+// getOrgsFromApplication which takes in a config proto and returns a map of
+// orgs in ChannelGroup Application.
+func getOrgsFromApplication(config *cb.Config) map[string]*cb.ConfigGroup {
+	return config.ChannelGroup.Groups[ApplicationGroupKey].Groups
```
caod123: See comment below about `getOrgsFromApplication`

pkg/config/msp_test.go
```diff
@@ -0,0 +1,628 @@
...
+		},
+		"sequence": "0"
+	}
+	`
+
+	expectedConfigProto := &cb.Config{}
+	err = protolator.DeepUnmarshalJSON(bytes.NewBufferString(expectedConfig), expectedConfigProto)
+	gt.Expect(err).NotTo(HaveOccurred())
+
+	gt.Expect(proto.Equal(config, expectedConfigProto)).To(BeTrue())
```
caod123: nit: this is fine, but if it ever does fail it will just show as `Expected false to be true`, the values themselves should be equivalent as well so `gt.Expect(config).To(Equal(expectedConfigProto))` should be fine here and will actually print out a diff if it's not equal

caod123 (changes requested):
1. We don't need to return both the MSPConfig and the contained FabricMSPConfig as well in the retriever. Just the FabricMSPConfig is probably fine
2. getOrgsFromApplication reads as if it had another intended usage besides just retrieving a single application org. If we don't need the map let's just change the function to retrieve a single org.

pkg/config/msp.go
```diff
@@ -74,3 +82,29 @@ func addRootCAToOrgMSP(rootCA *x509.Certificate, org *cb.ConfigGroup) error {
...
+	mspConfig, fabricMSPConfig, err := getMSPConfig(org)
+	if err != nil {
+		return fmt.Errorf("getting msp config: %v", err)
+	}
+
+	// TODO validate that this certificate was issued by this MSP
+
+	buffer := bytes.NewBuffer(nil)
+	err = pem.Encode(buffer, &pem.Block{Type: "CERTIFICATE", Bytes: cert.Raw})
```
yacovm: Just use [pem.EncodeToMemory](https://golang.org/pkg/encoding/pem/#EncodeToMemory)

yacovm: Type should be an **"X509 CRL"**, not a certificate.

pkg/config/msp.go
```diff
@@ -74,3 +82,29 @@ func addRootCAToOrgMSP(rootCA *x509.Certificate, org *cb.ConfigGroup) error {
...
+	if err != nil {
+		return err
+	}
+	fabricMSPConfig.RevocationList = append(fabricMSPConfig.RevocationList, buffer.Bytes())
```
yacovm: the bytes of a CRL are not an x509 certificate, but a CRL that looks like:

> type CertificateList struct {
> 	TBSCertList        TBSCertificateList
> 	SignatureAlgorithm AlgorithmIdentifier
> 	SignatureValue     asn1.BitString
> }

pkg/config/msp_test.go
wlahti: Done

pkg/config/msp.go
wlahti: Done

pkg/config/msp.go
stephyee: @yacovm What is the slice of byte slices for the fabricMSPConfig.RevocationList represented by in the CRL?

pkg/config/msp.go
wlahti: Done

pkg/config/msp.go
wlahti: Done

pkg/config/msp.go
wlahti: Done

pkg/config/msp.go
wlahti: Done

pkg/config/msp.go
wlahti: Done

pkg/config/msp.go
wlahti: Left a TODO to address this. Stomach is yelling at me to eat some lunch now. :)

yacovm: @stephyee , it's just an array of either PEM encoded or DER ASN1 encoded CRLs. Look at the [code that reads them](https://github.com/hyperledger/fabric/blob/master/msp/mspimplsetup.go#L197-L215), and you then can make the inverse code that writes them.

pkg/config/msp.go
```diff
@@ -0,0 +1,142 @@
...
+	// just parse the first CRL and add it there for now
+	crl, err := x509.ParseCRL(fabricMSPConfig.RevocationList[0])
+	if err != nil {
+		return fmt.Errorf("parsing CRL: %v", err)
+	}
+
+	crl.TBSCertList.RevokedCertificates = append(crl.TBSCertList.RevokedCertificates,
+		pkix.RevokedCertificate{
+			SerialNumber: cert.SerialNumber,
+			// RevocationTime: time.Now(),
```
stephyee: Why is this commented out?

wlahti: It was on purpose because I thought the ASN.1 marshaled bytes would change and break the test if we used the current time. Turns out it doesn't for some reason so I'll uncomment it. :)

pkg/config/msp.go
```diff
@@ -74,3 +100,63 @@ func addRootCAToOrgMSP(rootCA *x509.Certificate, org *cb.ConfigGroup) error {
...
+	configValue, ok := org.Values[MSPKey]
+	if !ok {
+		return nil, errors.New("org doesn't have MSP value")
+	}
+	return configValue, nil
+}
+
+// RevokeCertificateFromMSP takes an x509 certificate and adds it to the
+// revocation list for the specified application org MSP.
+func RevokeCertificateFromMSP(cert *x509.Certificate, orgName string, config *cb.Config) error {
```
yacovm: To revoke a certificate, you need the CA that issued the certificate to sign the CRL that revokes it. 
Even under the assumption that the function as it is now, correctly constructs the CRL and correctly adds it to the MSP config - won't make the certificate be actually revoked, because you're not signing the revocation with the issuer's private key.

wlahti: Does the CRL need to be signed each time a new certificate is added? The signing is easily handled by x509.CreateCRL (which we plan to incorporate when we implement the TODO to create a CRL if one doesn't already exist) so I was under the assumption that the previous signature on an existing CRL would still be valid.

yacovm: > Does the CRL need to be signed each time a new certificate is added? 

The stuff that is signed is the TBSCertificateList which contains the SNs of the revoked certificates. If you add another certificate then you obviously need to re-sign, otherwise the validation won't work.

My main point was though, that you need to somehow pass a private key to this method...

pkg/config/msp_test.go
```diff
@@ -8,16 +8,30 @@ package config
...
+	"github.com/golang/protobuf/proto"
+	cb "github.com/hyperledger/fabric-protos-go/common"
 	mb "github.com/hyperledger/fabric-protos-go/msp"
 	"github.com/hyperledger/fabric/common/tools/protolator"
+	"github.com/hyperledger/fabric/pkg/config/mock"
 	. "github.com/onsi/gomega"
 )
 
+//go:generate counterfeiter -o mock/signer.go --fake-name Signer . signer
+type signer interface {
```
sykesm: Why do we need a fake signer? Seems it’s quite easy to create a real signer and the only ref I see below uses it in the happy path such that we don’t validate args and we don’t return anything that’s actually referenced in the test.

(At least that’s how it looks from the diff.)

pkg/config/msp_test.go
```diff
@@ -94,6 +108,799 @@ func TestMSPToProtoFailure(t *testing.T) {
...
+			gt := NewGomegaWithT(t)
+			err = AddRootCAToMSP(tc.cert, "Org1", config)
+			gt.Expect(err).To(MatchError(tc.expectedErr))
+		})
+	}
+}
+
+// creates a CRL with an empty list of revoked certificates
+// and returns it as ASN.1 encoded bytes
+func createCRLBytesForTest() []byte {
```
sykesm: This all looks very wrong to me. Copying unexported bits of the standard library is usually a hint that something is amiss.

1. Why aren’t we generating a real certificate and key?
2. Once we have the actual key, we can issue certificates and then revoke them using the “official” [CreateCRL](https://golang.org/pkg/crypto/x509/#Certificate.CreateCRL) method.

pkg/config/mock/signer.go
```diff
@@ -0,0 +1,186 @@
+// Code generated by counterfeiter. DO NOT EDIT.
```
sykesm: I don’t believe this fake is necessary.

pkg/config/msp.go
```diff
@@ -262,3 +268,156 @@ func pemEncodeECDSAPrivateKey(priv *ecdsa.PrivateKey) ([]byte, error) {
...
+
+	// TODO validate that this certificate was issued by this MSP
+
+	if len(fabricMSPConfig.RevocationList) == 0 {
+		// TODO create a CRL here if one doesn't exist?
+		return errors.New("no CRL in RevocationList")
+	}
+
+	// just parse the first CRL and add it there for now
+	crl, err := x509.ParseCRL(fabricMSPConfig.RevocationList[0])
```
sykesm: Why would we be parsing the existing CRL? Why would we be changing it instead of adding to it?

pkg/config/msp.go
```diff
@@ -262,3 +268,156 @@ func pemEncodeECDSAPrivateKey(priv *ecdsa.PrivateKey) ([]byte, error) {
...
+
+	fabricMSPConfig, err := getFabricMSPConfig(org)
+	if err != nil {
+		return fmt.Errorf("getting msp config: %v", err)
+	}
+
+	// TODO validate that this certificate was issued by this MSP
+
+	if len(fabricMSPConfig.RevocationList) == 0 {
+		// TODO create a CRL here if one doesn't exist?
```
sykesm: Why aren’t we always creating a new CRL and appending?

pkg/config/msp.go
```diff
@@ -262,3 +268,156 @@ func pemEncodeECDSAPrivateKey(priv *ecdsa.PrivateKey) ([]byte, error) {
...
+	if !ok {
+		return nil, errors.New("org doesn't have MSP value")
+	}
+	return configValue, nil
+}
+
+// RevokeCertificateFromMSP takes an x509 certificate and adds it to the
+// revocation list for the specified application org MSP. It signs the CRL
+// using the specified crypto.Signer.
+func RevokeCertificateFromMSP(cert *x509.Certificate, orgName string, signer crypto.Signer, config *cb.Config) error {
```
sykesm: What does “FromMSP” mean in this context?

Why aren’t we supporting a (variadic) list of certificates (*x509.Certificate) or pkix.RevokedCertificates to revoke?

pkg/config/msp.go
```diff
@@ -262,3 +268,156 @@ func pemEncodeECDSAPrivateKey(priv *ecdsa.PrivateKey) ([]byte, error) {
...
+		return errors.New("no CRL in RevocationList")
+	}
+
+	// just parse the first CRL and add it there for now
+	crl, err := x509.ParseCRL(fabricMSPConfig.RevocationList[0])
+	if err != nil {
+		return fmt.Errorf("parsing CRL: %v", err)
+	}
+
+	crl.TBSCertList.RevokedCertificates = append(crl.TBSCertList.RevokedCertificates,
```
sykesm: This troubles me. I would fully expect us to use the CA certificate instance to create a new CRL and the associated private key to sign it.

Basically, everything from 388 onward is part of `CreateCRL` in the standard library.

pkg/config/msp_test.go
```diff
@@ -94,6 +108,799 @@ func TestMSPToProtoFailure(t *testing.T) {
...
+	org, err := getApplicationOrg(config, "Org1")
+	gt.Expect(err).NotTo(HaveOccurred())
+	fabricMSPConfig, err := getFabricMSPConfig(org)
+	gt.Expect(err).NotTo(HaveOccurred())
+	fabricMSPConfig.RevocationList = append(fabricMSPConfig.RevocationList, crlBytes)
+	err = addMSPConfigToOrg(org, fabricMSPConfig)
+	gt.Expect(err).NotTo(HaveOccurred())
+	config.ChannelGroup.Groups[ApplicationGroupKey].Groups["Org1"] = org
+
+	cert := &x509.Certificate{
```
sykesm: Please see earlier comment - I really think these tests need to use real certs and keys.

pkg/config/msp_test.go
```diff
@@ -94,6 +108,799 @@ func TestMSPToProtoFailure(t *testing.T) {
...
+		KeyUsage: x509.KeyUsageKeyAgreement,
+	}
+
+	fakeFailingSigner := &mock.Signer{}
```
sykesm: So this seems to be where a fake signer is used.  You can still cause a failure by changing the curve param ‘N’ value to the zero value.

sykesm (changes requested):

pkg/config/msp_test.go
wlahti: Done

pkg/config/msp_test.go
wlahti: Done

pkg/config/msp.go
wlahti: Done

pkg/config/msp.go
wlahti: Done

pkg/config/msp.go
wlahti: Done

pkg/config/msp_test.go
wlahti: Done

pkg/config/msp_test.go
wlahti: Removed this test for now since we're using the standard library call and not directly calling signer.Sign any more.

pkg/config/msp.go
```diff
@@ -589,3 +615,56 @@ func addRootCAToOrgMSP(org *cb.ConfigGroup, rootCA *x509.Certificate) error {
...
+	}
+
+	fabricMSPConfig, err := getFabricMSPConfig(org)
+	if err != nil {
+		return fmt.Errorf("getting msp config: %v", err)
+	}
+
+	// TODO validate that this certificate was issued by this MSP
+
+	revokeTime := time.Now()
```
sykesm, line 643: May need to make this UTC.

pkg/config/msp.go
```diff
@@ -589,3 +615,56 @@ func addRootCAToOrgMSP(org *cb.ConfigGroup, rootCA *x509.Certificate) error {
...
+	revokedCertificates := make([]pkix.RevokedCertificate, len(certs))
+	for i, cert := range certs {
+		revokedCertificates[i] = pkix.RevokedCertificate{
+			SerialNumber:   cert.SerialNumber,
+			RevocationTime: revokeTime,
+		}
+	}
+
+	crlBytes, err := caCert.CreateCRL(rand.Reader, caPrivKey, revokedCertificates, revokeTime, revokeTime.Add(365*24*time.Hour))
```
sykesm, line 652: I don't usually say this but you may want to pull the 365*24*time.Hour out to a constant just above the function. 

IIUC, most CRL's in the real world expire pretty much every day... Not something I expect for fabric but I'm not really sure what the recommendation would be either.

pkg/config/msp_test.go
```diff
@@ -524,498 +527,753 @@ func TestAddRootCAToMSP(t *testing.T) {
...
+	gt.Expect(err).NotTo(HaveOccurred())
+	gt.Expect(org1MSP.RevocationList).To(HaveLen(1))
+
+	caCert, caPrivKey := generateCACertAndPrivateKey("org1.example.com")
+	cert, _ := generateCertAndPrivateKeyFromCACert("Org1", caCert, caPrivKey)
+
+	err = RevokeCertificateFromMSP(config, "Org1", caCert, caPrivKey, cert)
+	gt.Expect(err).ToNot(HaveOccurred())
+
+	protolator.DeepMarshalJSON(os.Stdout, config)
```
sykesm, line 932: Why the marshal to Stdout? This is just going to create spam.

sykesm (approved):
I see a couple of minor nits that can be addressed afterwards.

pkg/config/msp_test.go
wlahti: Oops. I was just taking a look at the config and forgot to remove this line.

PR #796, opened 2020-03-03
Update release-1.4 to go 1.13

lindluni:
Adds the cherrypicks and changes necessary to support go1.13 in release-1.4 branch as go1.12 is EOL.

### DIFF
--- a/bccsp/idemix/bridge/bridge_test.go
+++ b/bccsp/idemix/bridge/bridge_test.go
@@ -98,7 +98,7 @@ var _ = Describe("Idemix Bridge", func() {
 			It("panic on rand failure", func() {
 				Issuer.NewRand = NewRandPanic
 				key, err := Issuer.NewKey(nil)
-				Expect(err.Error()).To(BeEquivalentTo("failure [new rand panic]"))
+				Expect(err).To(MatchError("failure [new rand panic]"))
 				Expect(key).To(BeNil())
 			})
 		})
@@ -107,13 +107,13 @@ var _ = Describe("Idemix Bridge", func() {
 
 			It("fails to unmarshal issuer public key", func() {
 				pk, err := Issuer.NewPublicKeyFromBytes([]byte{0, 1, 2, 3, 4}, nil)
-				Expect(err.Error()).To(BeEquivalentTo("failed to unmarshal issuer public key: proto: idemix.IssuerPublicKey: illegal tag 0 (wire type 0)"))
+				Expect(err).To(MatchError("failed to unmarshal issuer public key: proto: idemix.IssuerPublicKey: illegal tag 0 (wire type 0)"))
 				Expect(pk).To(BeNil())
 			})
 
 			It("fails to unmarshal issuer public key", func() {
 				pk, err := Issuer.NewPublicKeyFromBytes(nil, nil)
-				Expect(err.Error()).To(BeEquivalentTo("failure [runtime error: index out of range]"))
+				Expect(err).To(MatchError(ContainSubstring("failure [runtime error: index out of range")))
 				Expect(pk).To(BeNil())
 			})
 
@@ -141,7 +141,7 @@ var _ = Describe("Idemix Bridge", func() {
 
 					pk, err = Issuer.NewPublicKeyFromBytes(raw, nil)
 					Expect(err).To(HaveOccurred())
-					Expect(err.Error()).To(BeEquivalentTo("invalid issuer public key: zero knowledge proof in public key invalid"))
+					Expect(err).To(MatchError("invalid issuer public key: zero knowledge proof in public key invalid"))
 					Expect(pk).To(BeNil())
 				})
 
@@ -151,7 +151,7 @@ var _ = Describe("Idemix Bridge", func() {
 					Expect(raw).NotTo(BeEmpty())
 
 					pk, err := Issuer.NewPublicKeyFromBytes(raw, []string{"A"})
-					Expect(err.Error()).To(BeEquivalentTo("invalid number of attributes, expected [2], got [1]"))
+					Expect(err).To(MatchError("invalid number of attributes, expected [2], got [1]"))
 					Expect(pk).To(BeNil())
 				})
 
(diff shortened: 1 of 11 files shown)


btl5037:
The two tests I flagged to skip in here, I'll need some help with, I see no discernible reason for their failure. However, one of them the part that fails is testing the gRPC dependency directly and not our code and is replicating exactly what the other skipped test is also doing, so we can probably drop one of them.

The increase in the timeout for the temporal assertion I haven't been able to find anywhere in master a PR that might have fixed that to cherry-pick, so still not sure whats going on with the time drift there.

sykesm:
The gRPC dep update seems unrelated to a move to go 1.13. Why is that required as part of this?

lindluni:
> The gRPC dep update seems unrelated to a move to go 1.13. Why is that required as part of this?

After upgrading to go1.13 our tests for probing TLS on an gRPC endpoint, if a certificate is presented by the client that "shouldn't" satisfy the handshake, the `dialContext` on the endpoint still succeeds.

The exact change in go1.13 and gRPC that causes this, I'm not sure. But I can say we haven't touched any of the code that is failing, and the only resolution I could find to fix this was updating gRPC. You and I actually looked at this for a little at my desk before we got sent home, and it was your recommendation that it was probably in the gRPC dep.

I presume we didn't see this error when we originally migrated to go1.13, because we updated the gRPC dependency prior to migrating. I could go back and cherry pick releases of gRPC one at a time until I get to the one that fixes the problem and then dig into what actually fixed it if you want.

sykesm:
> After upgrading to go1.13 our tests for probing TLS on an gRPC endpoint, if a certificate is presented by the client that "shouldn't" satisfy the handshake, the `dialContext` on the endpoint still succeeds.

Not seeing it. Pulled down your changes, setup go 1.13.8, cleaned the tree, cleaned docker, removed the commit that updates gRPC and ran the tests. The only one I remember is the cherry-pick you made related to TLS 1.2.

I'm doing one more build to make sure I don't see it.

lindluni:
https://github.com/hyperledger/fabric/blob/master/common/crypto/tlsgen/ca_test.go

It was this test that was failing if you want to target it directly. I confirmed just now if I do the same (pull out the gRPC update) it fails. And I'm on 1.13.8.

lindluni:
Running it over here in CI without the gRPC commit: https://github.com/hyperledger/fabric/pull/1029

lindluni:
Here is the failure in CI without the gRPC chang: https://dev.azure.com/Hyperledger/Fabric/_build/results?buildId=11663&view=logs&j=a11cdf60-0d11-5130-c801-1b4953009464&t=849b73ad-cbaa-5065-7214-c0ac3d42920d

I forgot this also fixed a bug related to this test with temporal drift which you can also see in the CI log: https://github.com/hyperledger/fabric/blob/release-1.4/common/grpclogging/server_test.go#L191

lindluni:
There is one other failed test, with the same tlsProbe, but in my opinion that test can be removed as its directly testing the gRPC dependency and not the actual probing of the endpoint

sykesm:
The problem in the test is the same as the other comm tests: go 1.13 enables TLS 1.3 by default

```diff
diff --git a/common/crypto/tlsgen/ca_test.go b/common/crypto/tlsgen/ca_test.go
index a65478faa..d54895aa3 100644
--- a/common/crypto/tlsgen/ca_test.go
+++ b/common/crypto/tlsgen/ca_test.go
@@ -63,6 +63,7 @@ func TestTLSCA(t *testing.T) {
                tlsCfg := &tls.Config{
                        RootCAs:      x509.NewCertPool(),
                        Certificates: []tls.Certificate{cert},
+                       MaxVersion:   tls.VersionTLS12,
                }
                tlsCfg.RootCAs.AppendCertsFromPEM(ca.CertBytes())
                tlsOpts := grpc.WithTransportCredentials(credentials.NewTLS(tlsCfg))
```

```
$ go test -count 100 -run TestTLSCA ./common/crypto/tlsgen/...
ok      github.com/hyperledger/fabric/common/crypto/tlsgen      100.635s
```

Picking up over 18 months of gRPC changes does not seem like a good thing to do without more extensive tests. This limits the changes to a well understood change in go 1.13.

sykesm (changes requested):
Please pull out the update to gRPC and address the test issue by setting the max TLS version to 1.3.

https://github.com/golang/go/issues/33368

sykesm:
> The exact change in go1.13 and gRPC that causes this, I'm not sure.

There are two things that I'm aware of that influence this. First, go 1.13 enables TLS 1.3 by default. With 1.3, the client is the last one to speak in the handshake; if an error (ie. auth failure) occurs on the server side, the client needs to try to read from the transport to discover it.

The second is that gRPC made a bunch of changes between 1.17 and 1.20 related to whether or not a client will wait for the HTTP/2 handshake to complete before marking a connection as ready. At some point in this transition, the behavior would have changed such that gRPC would read after a TLS handshake and an error would have come back from the connection.

So, yes, it's possible that we moved to a level of gRPC waited for the HTTP/2 handshake before we moved master to 1.13. But, in 1.4, we don't have that level so we need to swizzle the test to use TLS 1.2.

lindluni:
I think I still need to deal with the temporalAssertion, but I'll wait for CI to tell me if thats the case.

Really appreciate your help on this, very stupid oversight on my part

sykesm:
> I think I still need to deal with the temporalAssertion

I'd recommend cherry-picking https://github.com/hyperledger/fabric/commit/aa9a468d10b03b4cb31656ffa7399c48d7df256d#diff-8b390b04f3d0db01337904c6ce79f418 and moving on.

sykesm:
> I'd recommend cherry-picking [aa9a468]

Well, that didn't work; I thought I had addressed most of it. Might want to bump the threshold to 1s if you want this in today, otherwise I'll look at it tomorrow. I think I know what it is.

lindluni:
Half a second worked before, I'll try it again

lindluni:
But you don't need to address it today, it's waited this long, it can wait another day

sykesm:
> But you don't need to address it today, it's waited this long, it can wait another day

#1034

sykesm (approved):

PR #798, opened 2020-03-04
fix typo in Anchor Peers

python279:
I think there was a typo in Anchor Peers definition.

#### Type of change

- Documentation update


pamandrejko (approved):
Thanks for fixing this!

pamandrejko:
It would be good to cherry pick this fix back to 1.4 as well since that is the LTS release.

PR #797, opened 2020-03-04
modify the errors notification

bjzhang03:
Signed-off-by: bjzhang <bjzhang1991@gmail.com>

#### Type of change
- Documentation update

#### Description

When I run the fabric,I found the mistake in errors,so I fix it.

PR #799, opened 2020-03-04
The operations subsystem is a singular entity

sykesm:
Update the error message to use the singular "subsystem" instead of "subsystems".

Relates to #797


yacovm (approved):
The singularity is nigh!

PR #800, opened 2020-03-04
[FAB-17583] Simplify NewCreateChannelTx to only consider new application channels

caod123:
#### Type of change

- Test update
- Documentation update

#### Description

We only need the consortium value, org names, and application group
values to create a new application channel

#### Related issues

[FAB-17583](https://jira.hyperledger.org/browse/FAB-17583)

### DIFF
--- a/pkg/config/application_test.go
+++ b/pkg/config/application_test.go
@@ -8,12 +8,12 @@ package config
 
 import (
 	"bytes"
-	"errors"
 	"testing"
 
 	"github.com/golang/protobuf/proto"
 	cb "github.com/hyperledger/fabric-protos-go/common"
 	mb "github.com/hyperledger/fabric-protos-go/msp"
+	pb "github.com/hyperledger/fabric-protos-go/peer"
 	"github.com/hyperledger/fabric/common/tools/protolator"
 	. "github.com/onsi/gomega"
 )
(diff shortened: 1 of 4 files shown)


pkg/config/application_test.go
```diff
@@ -69,21 +55,67 @@ func TestNewApplicationGroupFailure(t *testing.T) {
...
+				delete(application.Policies, WritersPolicyKey)
+			},
+			expectedErr: "no Writers policy defined",
+		},
+		{
+			testName: "When ImplicitMetaPolicy rules' subpolicy is missing",
+			applicationMod: func(application *Application) {
+				application.Policies[ReadersPolicyKey].Rule = "ALL"
+			},
+			expectedErr: "invalid implicit meta policy rule: 'ALL': expected two space separated " +
```
sykesm, line 93: nit: there's no need to separate these long lines to satisfy the "long line linter." While it doesn't matter too much in test, it does make it more difficult to find messages in production code and it's easier to drop spaces between words.

pkg/config/config_test.go
```diff
@@ -729,9 +617,9 @@ func TestNewOrgConfigGroup(t *testing.T) {
 		}
 	},
 	"values": {
-		"Endpoints": {
+		"AnchorPeers": {
 			"mod_policy": "Admins",
-			"value": "Cg8xMjMuNDUuNjc3OjgwODA=",
+			"value": "CgkKBWhvc3QxEHs=",
```
sykesm, line 622: Unrelated: Shouldn't this be getting handled by protolator? Do we need to open a bug for that?

sykesm (approved):

PR #801, opened 2020-03-04
Use ephemeral port for TLSCA test

lindluni:
Replaces random port generation with an ephemeral port in tlsca tests

### DIFF
--- a/common/crypto/tlsgen/ca_test.go
+++ b/common/crypto/tlsgen/ca_test.go
@@ -10,8 +10,6 @@ import (
 	"context"
 	"crypto/tls"
 	"crypto/x509"
-	"fmt"
-	"math/rand"
 	"net"
 	"testing"
 	"time"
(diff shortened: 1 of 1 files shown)


C0rWin (approved):

sykesm:
Units failed due to gossip tests.

lindluni:
> Units failed due to gossip tests.

I actually think I have a fix for this flake, got tired of hitting it

lindluni:
@sykesm  I need to go actually read the `run-unit-test` script, but how did this change trigger gossip, it's an update to a test only?

lindluni:
The only reason I made the comment I did in the other PR is that on my first push it actually failed in gossip tests as I forgot to change something, so they triggered, then on the second push gossip tests didnt trgger

sykesm:
> @sykesm I need to go actually read the `run-unit-test` script, but how did this change trigger gossip, it's an update to a test only?

The filter I was thinking of is not part of the unit-test script. That script just enumerates the import paths. I agree something is going on.

lindluni:
> > @sykesm I need to go actually read the `run-unit-test` script, but how did this change trigger gossip, it's an update to a test only?
> 
> The filter I was thinking of is not part of the unit-test script. That script just enumerates the import paths. I agree something is going on.

Opened as a reminder to investigate:

https://jira.hyperledger.org/browse/FAB-17586

PR #802, opened 2020-03-04
[FAB-15570] Make peer functions synchronous

lindluni:
Reduces the extreme flakiness of the `gossip/service` package

The startup of peers was done an asynchronous fashion which caused the system to slow down when multiple peers were starting at once. This caused the tests to often hit timeout limits.

This change modifies the startup to occur sequentially. Timing show a reduction in actual start of 4x by reducing deadlock in the system.

Reproducing the problem was trivial, simply running the tests was enough to hit it, timing tests were performed using the following command with a timer attached to the startup sequence in code:

`gotestsum -- -race -count=20 --run=TestWithStaticDeliverClientLeader ./...`

In asynchronous mode, for `count=20` between 6 and 8 failures would occur with each run, in synchronous mode, for `count=20` I had one timeout in 5 runs

Asynchronous tImings:
`8.005409
16.009951
6.009657
14.019486
14.011131
14.009675
10.010783
20.009928
26.035220
16.012245
22.014202
8.005213
8.005221
22.016637
14.004296
28.023713
10.005477
16.004200
10.009094
16.009174`

Synchronous timings:

`2.001079
10.009088
2.000459
2.000243
16.012208
2.000586
6.005354
2.002467
2.001784
2.002884
18.017998
10.003161
2.002959
2.000145
18.028490
10.005575
2.000483
10.004647
2.002277
2.001265`

### DIFF
--- a/gossip/service/gossip_service_test.go
+++ b/gossip/service/gossip_service_test.go
@@ -12,7 +12,6 @@ import (
 	"io/ioutil"
 	"net"
 	"os"
-	"sync"
 	"testing"
 	"time"
 
@@ -51,6 +50,8 @@ import (
 	"github.com/stretchr/testify/require"
 	"google.golang.org/grpc"
 )
+
+const TIMEOUT = 45 * time.Second
 
 func init() {
 	util.SetupTestLogging()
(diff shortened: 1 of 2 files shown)


lindluni:
This change affects 8 test paths, so it should dramatically reduce the number of flakes

C0rWin (approved):

lindluni:
Clearly something is broken with our unit-tests, this direct change to gossip didn't trigger gossip tests

C0rWin:
> Reproducing the problem was trivial, simply running the tests was enough to hit it, timing tests were performed using the following command with a timer attached to the startup sequence in code:

> `gotestsum -- -race -count=20 --run=TestWithStaticDeliverClientLeader ./...`


indeed checked locally, this PR definitely an improvement. thanks @btl5037!

sykesm:
> Clearly something is broken with our unit-tests, this direct change to gossip didn't trigger gossip tests

It’s simply that the test deps are not considered. The script only joins on production dependencies and ignores XDeps and TestDeps.

lindluni:
> > Clearly something is broken with our unit-tests, this direct change to gossip didn't trigger gossip tests
> 
> It’s simply that the test deps are not considered. The script only joins on production dependencies and ignores XDeps and TestDeps.

Okay that makes more sense

PR #803, opened 2020-03-05
Remove plugin_packages from run-unit-tests.sh

sykesm:
No longer used or needed.


lindluni:
Initial push didn't trigger CI for some reason

lindluni:
+1 for deleting dead code

denyeart (approved):
In the future would be good to expand the commit message to indicate why something is no longer used so that others understand. But I'll go ahead and merge...

PR #805, opened 2020-03-05
[FAB-17582] Use custom API type for FabricMSPConfig proto

caod123:
#### Type of change

- Test update

#### Description

- Custom API type is more consumer friendly. It accepts x509
certificates and ECDSA keys instead of byte slices and handles PEM
encoding them to byte slices for the FabricMSPConfig proto.
- General cleanup for application channel creation.
- Cleanup for AddOrgToConsortium

#### Related issues

[FAB-17582](https://jira.hyperledger.org/browse/FAB-17582)

### DIFF
--- a/pkg/config/application.go
+++ b/pkg/config/application.go
@@ -17,7 +17,7 @@ import (
 // Application is a copy of the orderer configuration with the addition of an anchor peers
 // list in the organization definition.
 type Application struct {
-	Organizations []*Organization
+	Organizations []Organization
 	Capabilities  map[string]bool
 	Policies      map[string]*Policy
 	ACLs          map[string]string
@@ -31,7 +31,7 @@ type AnchorPeer struct {
 
 // newApplicationGroup returns the application component of the channel configuration.
 // By default, it sets the mod_policy of all elements to "Admins".
-func newApplicationGroup(application *Application) (*cb.ConfigGroup, error) {
+func newApplicationGroup(application Application) (*cb.ConfigGroup, error) {
 	var err error
 
 	applicationGroup := newConfigGroup()
@@ -65,7 +65,7 @@ func newApplicationGroup(application *Application) (*cb.ConfigGroup, error) {
 // AddAnchorPeer adds an anchor peer to an existing channel config transaction.
 // It must add the anchor peer to an existing org and the anchor peer must not already
 // exist in the org.
-func AddAnchorPeer(config *cb.Config, orgName string, newAnchorPeer *AnchorPeer) error {
+func AddAnchorPeer(config *cb.Config, orgName string, newAnchorPeer AnchorPeer) error {
 	applicationOrgGroup, ok := config.ChannelGroup.Groups[ApplicationGroupKey].Groups[orgName]
 	if !ok {
 		return fmt.Errorf("application org %s does not exist in channel config", orgName)
@@ -108,7 +108,7 @@ func AddAnchorPeer(config *cb.Config, orgName string, newAnchorPeer *AnchorPeer)
 
 // RemoveAnchorPeer removes an anchor peer from an existing channel config transaction.
 // The removed anchor peer and org it belongs to must both already exist.
-func RemoveAnchorPeer(config *cb.Config, orgName string, anchorPeerToRemove *AnchorPeer) error {
+func RemoveAnchorPeer(config *cb.Config, orgName string, anchorPeerToRemove AnchorPeer) error {
 	applicationOrgGroup, ok := config.ChannelGroup.Groups[ApplicationGroupKey].Groups[orgName]
 	if !ok {
 		return fmt.Errorf("application org %s does not exist in channel config", orgName)
@@ -144,7 +144,7 @@ func RemoveAnchorPeer(config *cb.Config, orgName string, anchorPeerToRemove *Anc
 }
 
 // GetAnchorPeers retrieves existing anchor peers from a application organization.
-func GetAnchorPeers(config *cb.Config, orgName string) ([]*AnchorPeer, error) {
+func GetAnchorPeers(config cb.Config, orgName string) ([]AnchorPeer, error) {
 	applicationOrgGroup, ok := config.ChannelGroup.Groups[ApplicationGroupKey].Groups[orgName]
 	if !ok {
 		return nil, fmt.Errorf("application org %s does not exist in channel config", orgName)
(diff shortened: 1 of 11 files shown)


pkg/config/config_test.go
```diff
@@ -617,21 +556,21 @@ func TestNewOrgConfigGroup(t *testing.T) {
 		}
 	},
 	"values": {
-		"AnchorPeers": {
+		"Endpoints": {
 			"mod_policy": "Admins",
-			"value": "CgkKBWhvc3QxEHs=",
+			"value": "Cg1sb2NhbGhvc3Q6MTIz",
```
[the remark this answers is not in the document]
caod123: We figured out this was because we were using protolator on an arbitrary ConfigGroup. protolator is unable to determine how to decorate the ConfigGroup if we don't explicitly say what type of ConfigGroup it is. By wrapping it with a DynamicOrdererOrgGroup in the tests it now properly decodes the values.

pkg/config/msp.go
```diff
@@ -0,0 +1,264 @@
...
+			OrganizationalUnitIdentifier: identifier.OrganizationalUnitIdentifier,
+		}
+
+		fabricIdentifiers = append(fabricIdentifiers, fabricOUIdentifier)
+	}
+
+	return fabricIdentifiers
+}
+
+func buildPemEncodedCRL(crl []pkix.CertificateList) ([][]byte, error) {
```
mastersingh24:
```suggestion
func buildPemEncodedCRL(crls []pkix.CertificateList) ([][]byte, error) {
```

pkg/config/msp.go
```diff
@@ -0,0 +1,264 @@
...
+	pemEncodedCRL := [][]byte{}
+
+	for _, cert := range crl {
```
mastersingh24:
```suggestion
	for _, crl := range crls {
```

pkg/config/msp.go
```diff
@@ -0,0 +1,264 @@
...
+		asn1MarshalledBytes, err := asn1.Marshal(cert)
```
mastersingh24:
```suggestion
		asn1MarshalledBytes, err := asn1.Marshal(crl)
```

pkg/config/msp.go
mastersingh24: mostly a nit, but the input is really an array of crls

mastersingh24 (changes requested):

mastersingh24 (dismissed):

mastersingh24:
I'll let @sykesm have one more look as well

pkg/config/application.go
```diff
@@ -144,16 +144,16 @@ func RemoveAnchorPeer(config *cb.Config, orgName string, anchorPeerToRemove *Anc
 	return fmt.Errorf("could not find anchor peer %s:%d in %s's anchor peer endpoints", anchorPeerToRemove.Host, anchorPeerToRemove.Port, orgName)
 }
 
-// GetAnchorPeer retrieves existing anchor peers from a application organization.
-func GetAnchorPeer(config *cb.Config, orgName string) ([]*AnchorPeer, error) {
+// GetAnchorPeers retrieves existing anchor peers from a application organization.
+func GetAnchorPeers(config *cb.Config, orgName string) ([]*AnchorPeer, error) {
```
sykesm: It's not specific to this function (and I think I've mentioned it before), please consider value objects on the API.

pkg/config/config.go
```diff
@@ -50,10 +50,10 @@ type Resources struct {
...
 // Organization is an organization in the channel configuration.
 type Organization struct {
-	Name      string
-	ID        string
-	Policies  map[string]*Policy
-	MSPConfig *mb.FabricMSPConfig
+	Name     string
+	ID       string
+	Policies map[string]*Policy
+	MSP      *MSP
```
sykesm: Should these be references or values?

pkg/config/msp.go
```diff
@@ -0,0 +1,264 @@
...
+	CryptoConfig *CryptoConfig
+	// List of TLS root certificates trusted by this MSP.
+	// They are returned by GetTLSRootCerts.
+	TLSRootCerts []x509.Certificate
+	// List of TLS intermediate certificates trusted by this MSP;
+	// They are returned by GetTLSIntermediateCerts.
+	TLSIntermediateCerts []x509.Certificate
+	// fabric_node_ous contains the configuration to distinguish clients from peers from orderers
+	// based on the OUs.
+	NodeOus *NodeOUs
```
sykesm: values or references? In the case of the slices, they are references but should they contain pointers or values?

sykesm (dismissed):
This can be merged if you're comfortable there won't be a broken build due to a conflict.

pkg/config/application.go
caod123: Good point I'll push a quick followon commit to this PR that makes some changes to the types.

caod123:
@sykesm we pushed a followon commit to this PR that addresses some of the unnecessary pointers in the config package. This PR was also rebased on master so if CI passes it shouldn't break master.

pkg/config/application.go
```diff
@@ -17,7 +17,7 @@ import (
 // Application is a copy of the orderer configuration with the addition of an anchor peers
 // list in the organization definition.
 type Application struct {
-	Organizations []*Organization
+	Organizations []Organization
 	Capabilities  map[string]bool
 	Policies      map[string]*Policy
```
sykesm, line 22: Pointer or value?

pkg/config/application.go
```diff
@@ -143,16 +143,16 @@ func RemoveAnchorPeer(config *cb.Config, orgName string, anchorPeerToRemove *Anc
 	return fmt.Errorf("could not find anchor peer %s:%d in %s's anchor peer endpoints", anchorPeerToRemove.Host, anchorPeerToRemove.Port, orgName)
 }
 
-// GetAnchorPeer retrieves existing anchor peers from a application organization.
-func GetAnchorPeer(config *cb.Config, orgName string) ([]*AnchorPeer, error) {
+// GetAnchorPeers retrieves existing anchor peers from a application organization.
+func GetAnchorPeers(config cb.Config, orgName string) ([]AnchorPeer, error) {
```
sykesm, line 147: Given all the other functions take a pointer, this will be inconsistent and require a dereference.

sykesm (approved):

pkg/config/application.go
DereckLuo: it is easier to use it as pointers, as for now many tests are modifying policy values, which using pointer make it a bit easier. We can consider changing it to value down the line.

PR #814, opened 2020-03-09
Fixed GRPCS certificate verification problem when using cryptogen

StealthyCoder:
* Fixed ability to use cryptogen with GRPCS chaincode servers

* Fixed linter issues

#### Type of change

- Bug fix

#### Description

The issue was that I used cryptogen to generate the cryptographic material for two organisation aforehand. Then when running the setup inside a AWS EKS cluster the problem arose that the docker container executing the chaincode could not connect to the ChainCode address with TLS because the verification of the certificate failed. This was due to the fact the TLS credentials get generated on the fly inside the process and therefore you cannot get to it. 

My first quick fix was to write the SSL material to disk and use that but then the signing of the Chaincode failed but the connection was verified though. So the next step was to use the same certificates as the main Peer connection for ChainCode as well. 

#### Additional details

I generated the SSL certificates for two organisations aforehand and used those to run two separate pods inside a Kubernetes cluster. They connect through a single orderer and everything is exposed with services. The Kubernetes cluster also has a docker registry running inside and the two pods have a docker sidecar loaded in for executing the docker containers running the chaincode. 

I had a lot of trouble running the `make checks` as I kept running into Go project specific issues like usage of internal packages with `go vet` and I could not get a correct output. Is there a getting started document targeting how to exactly setup the local development environment?

### DIFF
--- a/core/chaincode/accesscontrol/access.go
+++ b/core/chaincode/accesscontrol/access.go
@@ -7,8 +7,12 @@ SPDX-License-Identifier: Apache-2.0
 package accesscontrol
 
 import (
+	"crypto/x509"
+	"encoding/pem"
 	"errors"
 	"fmt"
+
+	"github.com/hyperledger/fabric/core/peer"
 
 	"github.com/golang/protobuf/proto"
 	pb "github.com/hyperledger/fabric-protos-go/peer"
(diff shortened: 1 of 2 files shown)


StealthyCoder:
I wanted to hook it up to the JIRA issue but somehow I was unable. So it should be https://jira.hyperledger.org/browse/FAB-17590 this issue.

yacovm (changes requested):
The ephemeral TLS CA and dynamic generation of certificates is for a reason - it protects the peer from anyone connecting to the chaincode instead of the real chaincode, and it protects the peer from security incidents in which a specific chaincode has a vulnerability and this is used to connect to the peer as a different chaincode or as the same chaincode (by stealing the certificate).

sykesm:
Piggy-backing on @yacovm's comment, the existing code is working as designed. We generate a new certificate for the chaincode server each time the peer starts. When we build and start chaincode, we provide the appropriate TLS material to the instance we initiate.

Additionally, `cryptogen` has nothing to do with chaincode transports; it is simply a tool to bootstrap cryptographic material for non-production networks.

PR #819, opened 2020-03-09
Add integration test package that uses pkg/config

wlahti:
#### Type of change
- Test update

#### Description

- For now, just create an application channel and join all peers to it.

#### Related issues

- [FAB-17579](https://jira.hyperledger.org/browse/FAB-17579)

### DIFF
--- a/integration/config/config_test.go
+++ b/integration/config/config_test.go
@@ -1,0 +1,136 @@
+/*
+Copyright IBM Corp All Rights Reserved.
+
+SPDX-License-Identifier: Apache-2.0
+*/
+
+package config
+
+import (
+	"io/ioutil"
+	"os"
+	"syscall"
+
+	docker "github.com/fsouza/go-dockerclient"
+	"github.com/golang/protobuf/proto"
+	"github.com/hyperledger/fabric/integration/nwo"
+	"github.com/hyperledger/fabric/integration/nwo/commands"
+	"github.com/hyperledger/fabric/pkg/config"
+	"github.com/tedsuo/ifrit"
+
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+)
+
+var _ = Describe("Config", func() {
+	var (
+		client  *docker.Client
+		testDir string
+		network *nwo.Network
+		process ifrit.Process
+	)
+
+	BeforeEach(func() {
+		var err error
+		testDir, err = ioutil.TempDir("", "config")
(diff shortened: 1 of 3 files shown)


integration/config/config_test.go
```diff
@@ -0,0 +1,138 @@
...
+	BeforeEach(func() {
+		var err error
+		testDir, err = ioutil.TempDir("", "config")
+		Expect(err).NotTo(HaveOccurred())
+
+		client, err = docker.NewClientFromEnv()
+		Expect(err).NotTo(HaveOccurred())
+
+		network = nwo.New(nwo.BasicSolo(), testDir, client, StartPort(), components)
+		network.UseConfigPkg = true
```
sykesm: This is the kind of switch I wanted to avoid introducing to `nwo`. The primitives should be in place such that this test can simply generate the config transaction without requiring any other changes (outside of port allocation).

integration/nwo/network.go
```diff
@@ -683,15 +684,17 @@ func (n *Network) Bootstrap() {
...
-		sess, err := n.ConfigTxGen(commands.CreateChannelTx{
-			ChannelID:             c.Name,
-			Profile:               c.Profile,
-			BaseProfile:           c.BaseProfile,
-			ConfigPath:            n.RootDir,
-			OutputCreateChannelTx: n.CreateChannelTxPath(c.Name),
-		})
-		Expect(err).NotTo(HaveOccurred())
-		Eventually(sess, n.EventuallyTimeout).Should(gexec.Exit(0))
+		if !n.UseConfigPkg {
```
sykesm: This should be unnecessary. We can simply ignore the result of calling configtxgen.

integration/ports.go
```diff
@@ -35,6 +35,7 @@ const (
 	PrivateDataBasePort
 	RaftBasePort
 	SBEBasePort
+	ConfigBasePort
```
sykesm: Please keep these sorted in alphabetical order.

sykesm (changes requested):
Meat of the test looks fine but we shouldn't need to change nwo.

integration/config/config_test.go
wlahti: Done

integration/nwo/network.go
wlahti: Done

integration/ports.go
wlahti: Done

sykesm (approved):

PR #822, opened 2020-03-10
[FAB-17517] Only Initialize specified BCCSP provider

ahmedsajid:
#### Type of change

- Bug fix

#### Description

Code currently tries to initialize multiple providers based on
provided config Opts being nil or not.

This update ensures that only specified provider is initialized
based on ProviderName.

This fixes "Failed initializing PKCS11.BCCSP %!s(<nil>)" error
when the code complied with PKCS11 enabled expects
configuration to not be nil even when Provider is set to SW.

Signed-off-by: Ahmed Sajid <ahmed.sajid@securekey.com>

#### Related issues

https://jira.hyperledger.org/browse/FAB-17517
#769
#821

### DIFF
--- a/bccsp/factory/pkcs11.go
+++ b/bccsp/factory/pkcs11.go
@@ -69,7 +69,7 @@ func setFactories(config *FactoryOpts) error {
 	}
 
 	// PKCS11-Based BCCSP
-	if config.Pkcs11Opts != nil {
+	if config.ProviderName == "PKCS11" && config.Pkcs11Opts != nil {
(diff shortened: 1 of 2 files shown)


lindluni:
Thank you, the maintainers are back from the Global Forum tomorrow, I will push for a follow up on this.

mastersingh24:
I was a bit concerned here that PKCS11 might use the configured SW provider when `SoftVerify` is true, but it configures it's own instance

mastersingh24 (approved):

PR #824, opened 2020-03-10
[FAB-17600] retrieve policies for a specific org

DereckLuo:
#### Type of change

- New feature

#### Description
- retrieve policies for a specific org.
#### Related issues
[FAB-17600](https://jira.hyperledger.org/browse/FAB-17600)

### DIFF
--- a/pkg/config/policy.go
+++ b/pkg/config/policy.go
@@ -1,0 +1,249 @@
+/*
+Copyright IBM Corp. All Rights Reserved.
+
+SPDX-License-Identifier: Apache-2.0
+*/
+
+package config
+
+import (
+	"errors"
+	"fmt"
+	"strconv"
+	"strings"
+
+	"github.com/golang/protobuf/proto"
+	cb "github.com/hyperledger/fabric-protos-go/common"
+	mb "github.com/hyperledger/fabric-protos-go/msp"
+	"github.com/hyperledger/fabric/common/policydsl"
+)
+
+// GetPoliciesForConsortiums returns a map of policies for channel consortiums.
+func GetPoliciesForConsortiums(config cb.Config) (map[string]*Policy, error) {
+	consortiums, ok := config.ChannelGroup.Groups[ConsortiumsGroupKey]
+	if !ok {
+		return nil, errors.New("consortiums missing from config")
+	}
+
+	return getPolicies(consortiums.Policies)
+}
+
+// GetPoliciesForConsortium returns a map of policies for a specific consortium.
+func GetPoliciesForConsortium(config cb.Config, consortiumName string) (map[string]*Policy, error) {
+	consortium, ok := config.ChannelGroup.Groups[ConsortiumsGroupKey].Groups[consortiumName]
+	if !ok {
+		return nil, fmt.Errorf("consortium %s does not exist in channel config", consortiumName)
+	}
+
+	return getPolicies(consortium.Policies)
+}
+
+// GetPoliciesForConsortiumOrg returns a map of policies for a specific consortium org.
+func GetPoliciesForConsortiumOrg(config cb.Config, consortiumName, orgName string) (map[string]*Policy, error) {
+	org, ok := config.ChannelGroup.Groups[ConsortiumsGroupKey].Groups[consortiumName].Groups[orgName]
+	if !ok {
+		return nil, fmt.Errorf("consortium org %s does not exist in channel config", orgName)
+	}
+
+	return getPolicies(org.Policies)
+}
+
+// GetPoliciesForOrderer returns a map of policies for channel orderer.
+func GetPoliciesForOrderer(config cb.Config) (map[string]*Policy, error) {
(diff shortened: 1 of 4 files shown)


caod123 (changes requested):

pkg/config/policy.go
```diff
@@ -0,0 +1,249 @@
...
+	if !ok {
+		return nil, fmt.Errorf("application org %s does not exist in channel config", orgName)
+	}
+
+	return getPolicies(orgGroup.Policies)
+}
+
+// getPolicies returns a map of Policy from given map of ConfigPolicy in organization config group.
+func getPolicies(policies map[string]*cb.ConfigPolicy) (map[string]*Policy, error) {
+	if policies == nil {
```
sykesm, line 94: Why bother with this block? Ranging over nil results in 0 executions of the block. This just introduces an unnecessary path.

pkg/config/policy.go
```diff
@@ -0,0 +1,249 @@
...
+		return map[string]*Policy{}, nil
+	}
+
+	p := map[string]*Policy{}
+	for name, policy := range policies {
+		switch policy.Policy.Type {
```
sykesm, line 100: nit: cast on the switch instead of every case

pkg/config/policy.go
```diff
@@ -0,0 +1,249 @@
...
+		args = append(args, cb.ImplicitMetaPolicy_ALL.String())
+	case cb.ImplicitMetaPolicy_MAJORITY:
+		args = append(args, cb.ImplicitMetaPolicy_MAJORITY.String())
+	default:
+		return "", fmt.Errorf("unknown implicit meta policy rule type %v", imp.Rule)
+	}
+
+	args = append(args, imp.SubPolicy)
+
+	return strings.Join(args, " "), nil
```
sykesm, line 157: nit: we are joining exactly two strings - do we really need to use a slice and join?

pkg/config/policy.go
```diff
@@ -0,0 +1,249 @@
...
+}
+
+// signatureMetaToString converts a *cb.SignaturePolicyEnvelope to a string representation.
+func signatureMetaToString(sig *cb.SignaturePolicyEnvelope) (string, error) {
+	roles := []string{}
```
sykesm, line 162: nit: var roles []string

https://github.com/golang/go/wiki/CodeReviewComments#declaring-empty-slices

pkg/config/policy.go
```diff
@@ -0,0 +1,249 @@
...
+			return "", err
+		}
+		roles = append(roles, role)
+	}
+	return signaturePolicyToString(sig.Rule, roles)
+}
+
+// mspPrincipalToString converts a *mb.MSPPrincipal to a string representation.
+func mspPrincipalToString(principal *mb.MSPPrincipal) (string, error) {
+	var res strings.Builder
```
sykesm, line 175: nit: do we need a builder? If we do, it’s only used in one block. The declaration should be in that block.

pkg/config/policy.go
```diff
@@ -0,0 +1,249 @@
...
+	}
+}
+
+// signaturePolicyToString recursively converts a *cb.SignaturePolicy to a
+// string representation.
+func signaturePolicyToString(sig *cb.SignaturePolicy, IDs []string) (string, error) {
+	switch sig.Type.(type) {
+	case *cb.SignaturePolicy_NOutOf_:
+		nOutOf := sig.GetNOutOf()
+		policies := []string{}
```
sykesm, line 213: nit: Another empty slice;  If I’m reading this correctly, it basically gets initialized on line 224 with the value of nOutOf.N.

sykesm (approved):
Mostly nits. It’s probably worth reading the empty slices recommendations in the code review wiki.

pkg/config/policy.go
DereckLuo: the intent of this block is when policies are nil in given config group, we want to return a empty policy map instead of error out. 
but seems the original code already handles that. 
This block will be removed when fixes applied on top.

pkg/config/policy.go
DereckLuo: done

pkg/config/policy.go
DereckLuo: thought slice.join will be faster. changes applied.

pkg/config/policy.go
DereckLuo: done.

pkg/config/policy.go
DereckLuo: builder is used here as I considered this might be a faster way to build strings.
I'm open to any suggestions. 
also moved it into the block with incoming fixes.

pkg/config/policy.go
DereckLuo: done.

DereckLuo:
fixes addressed in this PR: https://github.com/hyperledger/fabric/pull/848

PR #826, opened 2020-03-10
Avoid port conflicts in configtxlator tests

sykesm:
When running tests in parallel:

```
  configtxlator: error: Could not bind to address '127.0.0.1:9998': listen
  tcp 127.0.0.1:9998: bind: address already in use
```

Since the test parses the server port out of the startup messages, simply change the args to specify port 0 to indicate we want an ephemeral port.


lindluni:
👏

C0rWin (approved):

PR #828, opened 2020-03-10
Move common/ledger/util/Buffer to consumer

sykesm:
- Relocate a small helper class out of the "common" utility package and
  into the single consuming package.
- Stop exporting the Buffer type as it's a detail of the consumer.


sykesm:
UT Failure: FAB-17376

denyeart (approved):

PR #841, opened 2020-03-11
[FAB-17576] Support retrieving MSP configuration for a specific org

caod123:
#### Type of change

- New feature
- Test update

#### Description

* Support API for retrieving MSP configuration for application/orderer/consortium orgs
* Add a dummy cert/private key/crl to baseMSP in tests and update tests to use them

#### Related issues

[FAB-17576](https://jira.hyperledger.org/browse/FAB-17576)


pkg/config/consortiums_test.go
```diff
@@ -113,44 +113,44 @@ func TestAddOrgToConsortium(t *testing.T) {
 									"Admins": {
 										"mod_policy": "Admins",
 										"policy": {
-											"type": 3,
```
sykesm, line 116: What's with all of the whitespace changes in here? Seems unrelated to the change.

pkg/config/msp.go
```diff
@@ -135,13 +137,264 @@ type NodeOUs struct {
 	// If true then an msp identity that does not contain any of the specified OU will be considered invalid.
 	Enable bool
 	// OU Identifier of the clients.
-	ClientOuIdentifier OUIdentifier
+	ClientOUIdentifier OUIdentifier
```
sykesm, line 140: Thank you.

pkg/config/msp.go
```diff
@@ -255,10 +508,24 @@ func pemEncodeX509Certificate(cert x509.Certificate) []byte {
...
-	privBytes, err := x509.MarshalPKCS8PrivateKey(priv)
+	privBytes, err := x509.MarshalECPrivateKey(priv)
 	if err != nil {
-		return nil, fmt.Errorf("marshalling PKCS8 private key: %v", err)
+		return nil, fmt.Errorf("marshalling ec private key: %v", err)
 	}
 
-	return pem.EncodeToMemory(&pem.Block{Type: "PRIVATE KEY", Bytes: privBytes}), nil
+	return pem.EncodeToMemory(&pem.Block{Type: "EC PRIVATE KEY", Bytes: privBytes}), nil
+}
```
sykesm, line 517: Can you explain this change? I'd like to understand the motivation.

PKCS#8 handles multiple algorithms - not just EC. Some clients (Java chaincode, for example) are able to deal with PKCS#8 keys but fail to decode SEC 1 keys.

pkg/config/msp_test.go
```diff
@@ -11,13 +11,350 @@ import (
...
 
+	"github.com/golang/protobuf/proto"
+	cb "github.com/hyperledger/fabric-protos-go/common"
 	mb "github.com/hyperledger/fabric-protos-go/msp"
 	"github.com/hyperledger/fabric/common/tools/protolator"
+
 	. "github.com/onsi/gomega"
 )
 
+const (
```
sykesm, line 25: And how was this "arbitrary" certificate generated?

pkg/config/msp_test.go
```diff
@@ -91,50 +440,61 @@ func TestMSPToProtoFailure(t *testing.T) {
...
+	crl, _ := x509.ParseCRL(crlBlock.Bytes)
+
 	return MSP{
-		Name:              "",
-		RootCerts:         []x509.Certificate{},
-		IntermediateCerts: []x509.Certificate{},
-		Admins:            []x509.Certificate{},
-		RevocationList:    []pkix.CertificateList{},
+		Name:              "MSPID",
+		RootCerts:         []x509.Certificate{*cert},
```
sykesm, line 459: With all of the back and forth on ref vs. value, I'd expect the cert to be a ref since that's how the stdlib deals with it.

sykesm (approved):
Please review the comments. We can deal with the outcome in a separate PR.

pkg/config/consortiums_test.go
caod123: Turns out when I copied the JSON output into the test some whitespaces changed? I'll make sure nothing unnecessary gets committed when updating the expected JSON in the future

pkg/config/msp_test.go
caod123: I just found a couple of example certs from the x509 golang tests, I think if you parse the certificate it's the classic acme co wil e coyote stuff. I mainly just needed a valid cert to ensure the marshalling/unmarshalling logic actually worked.

pkg/config/msp_test.go
caod123: I did notice that while working through this that there were a number of places where I had to reference or dereference certain variables. I'll probably throw a story in the backlog to revisit some of the types to change them to what makes sense in the context of how the variables are actually used (ie in this example they might as well be pointers)

pkg/config/msp.go
caod123: As mentioned in standup I was under the impression Fabric only supported EC private keys. After discussion in standup though I understand where you're coming from on this and will change it back to PKCS8

PR #842, opened 2020-03-12
[FAB-17605] handled InitFactories error

Eggsyz:
Signed-off-by: jilg <liguo_ji@outlook.com>

#### Type of change

- Bug fix

#### Description

A project using func InitFactories, it output errors when build it. But many places do not handle errors.

```

factory.InitFactories(nil)

func InitFactories(config *FactoryOpts) error {

}

```
#### Related issues

[FAB-17605](https://jira.hyperledger.org/browse/FAB-17605)

### DIFF
--- a/cmd/discover/main.go
+++ b/cmd/discover/main.go
@@ -7,6 +7,7 @@ SPDX-License-Identifier: Apache-2.0
 package main
 
 import (
+	"fmt"
 	"os"
 
 	"github.com/hyperledger/fabric/bccsp/factory"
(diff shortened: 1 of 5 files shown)


sykesm:
Fixes should be made and reviewed against master before pulling back to other releases.

Please consider opening a PR against master. For the areas that you've changed, it's likely that the processes should panic instead of just printing a message. When you do print a message, it should not go to stdout but to stderr (as an error).

PR #846, opened 2020-03-12
Fix Test Flake in TestLedgerHeightFromProperties

lomov02:
The current implementation runs test in parallel. This slow down the system and tends to introduce flakes.

To correct this we removed the parrallization and ran tests synchronously

Signed-off-by: Ihor Marchenko <ihormarchenko02@gmail.com>


wenjianqiao:
The change looks good to me. There are other parallel tests in gossip packages that may cause flakes, but we will fix them in separate PRs.

yacovm (approved):

PR #848, opened 2020-03-12
[FAB-17600] Fixes retrieve policies for a specific config group

DereckLuo:
#### Type of change

#### Description
- changed policy from map of pointers to map of values.
- removed empty slice.

### DIFF
--- a/pkg/config/policy.go
+++ b/pkg/config/policy.go
@@ -19,7 +19,7 @@ import (
 )
 
 // GetPoliciesForConsortiums returns a map of policies for channel consortiums.
-func GetPoliciesForConsortiums(config cb.Config) (map[string]*Policy, error) {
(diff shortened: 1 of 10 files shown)


pkg/config/policy.go
```diff
@@ -139,27 +135,27 @@ func getPolicies(policies map[string]*cb.ConfigPolicy) (map[string]*Policy, erro
 
 // implicitMetaToString converts a *cb.ImplicitMetaPolicy to a string representation.
 func implicitMetaToString(imp *cb.ImplicitMetaPolicy) (string, error) {
-	args := []string{}
+	var args string
```
sykesm, line 138: nit: is `args` the right name for this? Does it reflect what it holds?

sykesm (approved):

PR #853, opened 2020-03-13
[FAB-17576] Cleanup MSP config retrieval logic

caod123:
* Fix spelling of marshaling in config package
* Support any PKCS#8 private key for the key material instead of just
ecdsa private keys
* Fix broken tests on master in config package

Cleanup of #841 and master breakage


pkg/config/msp.go
```diff
@@ -95,9 +94,10 @@ type KeyInfo struct {
 	// the case of Software BCCSP as well as the HSM BCCSP would be
 	// the SKI of the key.
 	KeyIdentifier string
-	// KeyMaterial (optional) for the key to be imported; this is
-	// properly encoded key bytes, prefixed by the type of the key.
-	KeyMaterial *ecdsa.PrivateKey
+	// KeyMaterial (optional) for the key to be imported; this
+	// must be a supported PKCS#8 private key type of either
+	// *rsa.PrivateKey, *ecdsa.PrivateKey, or ed25519.PrivateKey.
+	KeyMaterial interface{}
```
sykesm: Andy why isn’t this a crypto.PrivateKey ?

pkg/config/msp.go
```diff
@@ -362,16 +362,16 @@ func parseCRL(crls [][]byte) ([]pkix.CertificateList, error) {
 	return certificateLists, nil
 }
 
-func parsePrivateKeyFromBytes(priv []byte) (*ecdsa.PrivateKey, error) {
+func parsePrivateKeyFromBytes(priv []byte) (interface{}, error) {
```
sykesm: Why isn’t this a crypto.PrivateKey

pkg/config/msp.go
```diff
@@ -507,13 +507,13 @@ func pemEncodeX509Certificate(cert x509.Certificate) []byte {
 	return pem.EncodeToMemory(&pem.Block{Type: "CERTIFICATE", Bytes: cert.Raw})
 }
 
-func pemEncodeECDSAPrivateKey(priv *ecdsa.PrivateKey) ([]byte, error) {
-	privBytes, err := x509.MarshalECPrivateKey(priv)
+func pemEncodePKCS8PrivateKey(priv interface{}) ([]byte, error) {
```
sykesm: crypto.PrivateKey

pkg/config/msp.go
caod123: wasn't aware there was a type for that, I just used what the `x509.MarshalPKCS8PrivateKey` accepted. I'll update it

pkg/config/msp_test.go
```diff
@@ -23,7 +23,9 @@ import (
 )
 
 const (
-	// Arbitrary valid pem encoded x509 certificate
+	// Arbitrary valid pem encoded x509 certificate from crypto/x509 tests.
+	// The contents of the certifcate don't matter, we just need a valid certificate
+	// to pass marshaling/unmamarshaling.
```
sykesm, line 28: Really? I can give you an RSA cert and this will still work?

sykesm (approved):
Pulling this in to fix the build.

- crypto.PrivateKey indicates we’re looking for a private key; it’s an empty interface
- still need to document how you’re creating the certs and keys or generate them. I realize dynamic will change your test assertion so generation commands / instructions are fine.

pkg/config/msp_test.go
caod123: I believe as long as it passes x509.ParseCertificate it will work. Not sure if an RSA cert would or not

PR #858, opened 2020-03-14
[FAB-17623] Replace OrdererAddressesKey with EndpointsKey in orderer org's Values

wuxuer:

#### Type of change

- Test update

#### Description

In common.Config, an orderer org's Values (map[string]*cb.ConfigValue) can only have MSP or Endpoints as its key. OrdererAddressesKey is a key for Values of ChannelGroup.
#### Related issues

https://jira.hyperledger.org/browse/FAB-17623


caod123:
+1

sykesm (approved):

PR #872, opened 2020-03-17
[FAB-17629] Correctly Compile Integration Test List

lindluni:
Uses `go list` to create the integration directory test list instead of `grep` which will properly ignore directories by default such as: vendor, scripts and hidden directories

### DIFF
--- a/scripts/run-integration-tests.sh
+++ b/scripts/run-integration-tests.sh
@@ -8,36 +8,27 @@
 # script, they are treated as the directories containing the tests to run.
 # When no arguments are provided, all integration tests are executed.
 
-set -e -u
+set -eu
 
 fabric_dir="$(cd "$(dirname "$0")/.." && pwd)"
+cd ${fabric_dir}
 
-cd "$fabric_dir"
+declare -a test_dirs
+test_dirs=($(
+ go list -f '{{if or (len .TestGoFiles | ne 0) (len .XTestGoFiles | ne 0)}}{{println .Dir}}{{end}}' ./... | \
+ grep integration | \
+ sed "s,${fabric_dir},.,g"
+))
 
-dirs=()
-if [ "${#}" -eq 0 ]; then
-  specs=()
-  specs=("$(grep -Ril --exclude-dir=vendor --exclude-dir=scripts "RunSpecs" . | grep integration)")
-  for spec in ${specs[*]}; do
-    dirs+=("$(dirname "${spec}")")
-  done
-else
-  dirs=("$@")
-fi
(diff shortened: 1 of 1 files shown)


scripts/run-integration-tests.sh
```diff
@@ -8,36 +8,40 @@
...
 
+declare -a testDirs
+for dir in ${tempTestDirs[@]}; do
+    listInside=$(go list -f '{{.TestGoFiles}}' ${dir})
+    listOutside=$(go list -f '{{.XTestGoFiles}}' ${dir})
+    if [[ "${listInside}" != "[]" || "${listOutside}" != "[]" ]]; then
+        testDirs+=(${dir})
+    fi
+done
+
```
sykesm: I think you're making more difficult than it needs to be. `go list` is your friend.

This gives you all of the packages with integration tests in them.
```
go list -f '{{if or (len .TestGoFiles | ne 0) (len .XTestGoFiles | ne 0)}}{{println .ImportPath}}{{end}}' ./... | grep integration
```

If you need the directories (for ginkgo), substitute `{{.Dir}}` for `{{.ImportPath}}`. if you need a relative path, you'll have to munge that by piping it through `sed "s,${fabric_dir},.,g"`.

lindluni: This is exactly what I was looking for actually, I couldn't figure out how to do conditionals in the format string for `go list`

THANK YOU!

lindluni:
Load balancing is a little better with the re-introduction of the two missed packages, down to 25 minutes on integration

scripts/run-integration-tests.sh
```diff
@@ -8,36 +8,31 @@
...
-  done
-else
-  dirs=("$@")
-fi
-
-totalAgents=${SYSTEM_TOTALJOBSINPHASE:-0}   # standard VSTS variables available using parallel execution; total number of parallel jobs running
-agentNumber=${SYSTEM_JOBPOSITIONINPHASE:-0} # current job position
-testCount=${#dirs[@]}
+total_agents=${SYSTEM_TOTALJOBSINPHASE:-0}   # standard VSTS variables available using parallel execution; total number of parallel jobs running
+agent_number=${SYSTEM_JOBPOSITIONINPHASE:-0} # current job position
```
sykesm: Why aren't you just using '1' as the default since you're effectively doing that in the next block.

btl5037: Back when we originally did this, there was some reason for it related to how the 0th agent handles it. This was an exact copy of the Microsoft example for doing it.

I can't find the doc anymore that tells you why because they deleted the golang example. So I'll remove it and see what happens.

sykesm (approved):

PR #873, opened 2020-03-17
[FAB-17603] Support consensus type state field in config value

stephyee:
#### Type of change

- New feature

#### Description
Support consensus type state field for use to allow a channel to be put in maintenance mode. 

#### Related issues
[FAB-17603](https://jira.hyperledger.org/browse/FAB-17603)

### DIFF
--- a/pkg/config/orderer.go
+++ b/pkg/config/orderer.go
@@ -36,6 +36,11 @@ type Orderer struct {
 	// Capabilities is a map of the capabilities the orderer supports.
 	Capabilities map[string]bool
 	Policies     map[string]Policy
+	// State defines the orderer mode of operation.
+	// Normal is during normal operation
+	// Maintenance is during consensus type migration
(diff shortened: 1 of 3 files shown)


pkg/config/orderer.go
```diff
@@ -36,6 +36,11 @@ type Orderer struct {
 	// Capabilities is a map of the capabilities the orderer supports.
 	Capabilities map[string]bool
 	Policies     map[string]Policy
+	// State defines the orderer mode of operation.
+	// Normal is during normal operation
+	// Maintenance is during consensus type migration
+	// Options: `STATE_NORMAL` and `STATE_MAINTENANCE`
```
sykesm, line 42: Do these strings warrant constants / typed values?

sykesm (approved):
Please consider creating a constant and/or typed value for State

pkg/config/orderer.go
stephyee: It does, I'll update in a follow on commit.

PR #874, opened 2020-03-17
[FAB-17603] Add constants for consensus state

stephyee:
#### Type of change

#### Description

Add constants per previous [pr](https://github.com/hyperledger/fabric/pull/873) comments

#### Related issues

[FAB-17603](https://jira.hyperledger.org/browse/FAB-17603)

### DIFF
--- a/pkg/config/orderer.go
+++ b/pkg/config/orderer.go
@@ -18,6 +18,9 @@ import (
 	eb "github.com/hyperledger/fabric-protos-go/orderer/etcdraft"
 )
 
+// ConsensusState defines the orderer mode of operation.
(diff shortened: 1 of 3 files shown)


pkg/config/orderer.go
```diff
@@ -18,6 +18,9 @@ import (
 	eb "github.com/hyperledger/fabric-protos-go/orderer/etcdraft"
 )
 
+// ConsensusState defines the orderer mode of operation.
+type ConsensusState string
```
sykesm, line 22: You may also want to consider making this a fmt.Stringer.

sykesm (approved):

PR #875, opened 2020-03-17
grpc{logging,metrics} test flake updates

sykesm:
1. Use the `grpc.WithBlock` option on `grpc.Dial` to ensure the network connection is established before attempting to use the client connection.
2. Perform a final `Recv` after calling `CloseSend` to ensure the server side stream processing has completed. This result of this call should be an `io.EOF` on successful stream termination or the error returned by the server.

FAB-17548


mastersingh24 (approved):
LGTM

PR #887, opened 2020-03-19
[FAB-17592] Support removing an orderer endpoint from existing config

wuxuer:
Description

Modify an existing channel configuration to remove an endpoint from an orderer org's config group.

Related issues

https://jira.hyperledger.org/browse/FAB-17592

#### Type of change

- New feature

### DIFF
--- a/pkg/config/orderer.go
+++ b/pkg/config/orderer.go
@@ -333,3 +333,38 @@ func AddOrdererEndpoint(config *cb.Config, orgName string, endpoint string) erro
 
 	return nil
 }
+
+// RemoveOrdererEndpoint removes an orderer's endpoint from an existing channel config transaction.
+// The removed endpoint and org it belongs to must both already exist.
+func RemoveOrdererEndpoint(config *cb.Config, orgName string, endpoint string) error {
+	ordererOrgGroup, ok := config.ChannelGroup.Groups[OrdererGroupKey].Groups[orgName]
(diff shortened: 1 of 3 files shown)


stephyee:
@hellosmallstone please resolve the conflicts.

pkg/config/orderer.go
```diff
@@ -333,3 +333,38 @@ func AddOrdererEndpoint(config *cb.Config, orgName string, endpoint string) erro
...
+
+	if ordererAddrConfigValue, ok := ordererOrgGroup.Values[EndpointsKey]; ok {
+		err := proto.Unmarshal(ordererAddrConfigValue.Value, ordererAddrProto)
+		if err != nil {
+			return fmt.Errorf("failed unmarshalling orderer org %s's endpoints: %v", orgName, err)
+		}
+	}
+
+	existingEndpoints := ordererAddrProto.Addresses
+	for i, e := range existingEndpoints {
```
sykesm, line 355: This is a bit of an unconventional loop for removing. Nothing wrong, per se, but it does mean we exit the function at an error and that we have a normal return in the middle.

sykesm (approved):

pkg/config/orderer.go
sykesm: Example: https://github.com/golang/go/wiki/SliceTricks#filtering-without-allocating

PR #893, opened 2020-03-20
[FAB-17568, FAB-17636] Support retrieving capabilities and orderer configuration

caod123:
#### Type of change

- New feature

#### Description

Also reworked ConfigMetadata type so API uses custom type and support unmarshaling/marshaling between proto and custom type.

#### Related issues

[FAB-17568](https://jira.hyperledger.org/browse/FAB-17568)
[FAB-17636](https://jira.hyperledger.org/browse/FAB-17636)

### DIFF
--- a/pkg/config/capabilities.go
+++ b/pkg/config/capabilities.go
@@ -1,0 +1,80 @@
+/*
+Copyright IBM Corp. All Rights Reserved.
+
+SPDX-License-Identifier: Apache-2.0
+*/
+
+package config
+
+import (
+	"errors"
+	"fmt"
+
+	"github.com/golang/protobuf/proto"
+	cb "github.com/hyperledger/fabric-protos-go/common"
+)
+
+// GetChannelCapabilities returns a map of enabled channel capabilities
+// from a config transaction.
+func GetChannelCapabilities(config *cb.Config) (map[string]bool, error) {
+	capabilites, err := getCapabilities(config.ChannelGroup)
+	if err != nil {
+		return nil, fmt.Errorf("retrieving channel capabilities: %v", err)
+	}
+
+	return capabilites, nil
+}
+
+// GetOrdererCapabilities returns a map of enabled orderer capabilities
+// from a config transaction.
+func GetOrdererCapabilities(config *cb.Config) (map[string]bool, error) {
+	orderer, ok := config.ChannelGroup.Groups[OrdererGroupKey]
+	if !ok {
+		return nil, errors.New("orderer missing from config")
+	}
+
+	capabilites, err := getCapabilities(orderer)
+	if err != nil {
+		return nil, fmt.Errorf("retrieving orderer capabilities: %v", err)
+	}
+
+	return capabilites, nil
+}
+
+// GetApplicationCapabilities returns a map of enabled application capabilities
+// from a config transaction.
+func GetApplicationCapabilities(config *cb.Config) (map[string]bool, error) {
+	application, ok := config.ChannelGroup.Groups[ApplicationGroupKey]
+	if !ok {
+		return nil, errors.New("application missing from config")
+	}
+
+	capabilites, err := getCapabilities(application)
+	if err != nil {
+		return nil, fmt.Errorf("retrieving application capabilities: %v", err)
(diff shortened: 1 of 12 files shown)


pkg/config/orderer_test.go
```diff
@@ -221,65 +232,453 @@ func TestUpdateOrdererConfiguration(t *testing.T) {
...
+	"channel_group": {
+		"groups": {
+			"Orderer": {
+				"groups": {
+					"OrdererOrg": {
+						"groups": {},
+						"mod_policy": "Admins",
+						"policies": {
+							"Admins": {
+							"mod_policy": "Admins",
```
sykesm, line 261: Seems we have whitespace stuff going on again. Not sure why the indentation keeps getting lost.

pkg/config/orderer_test.go
```diff
@@ -796,3 +1158,94 @@ func TestRemoveOrdererEndpointFailure(t *testing.T) {
...
+func baseSoloOrderer() Orderer {
+	return Orderer{
+		Policies:    ordererStandardPolicies(),
+		OrdererType: ConsensusTypeSolo,
+		Organizations: []Organization{
+			{
+				Name:     "OrdererOrg",
+				Policies: orgStandardPolicies(),
+				OrdererEndpoints: []string{
+					"localhost:123",
```
sykesm, line 1182: Here’s a bit of where we see how we have different ways of handling addresses. Earlier we had host string and port uint while here we have an actual address as expected by dial.

pkg/config/orderer_test.go
```diff
@@ -796,3 +1158,94 @@ func TestRemoveOrdererEndpointFailure(t *testing.T) {
...
+	orderer.Kafka = Kafka{
+		Brokers: []string{"broker1", "broker2"},
+	}
+
+	return orderer
+}
+
+func baseEtcdRaftOrderer() Orderer {
+	certBlock, _ := pem.Decode([]byte(dummyCert))
+	cert, _ := x509.ParseCertificate(certBlock.Bytes)
```
sykesm, line 1212: need an assertion on no error

sykesm (approved):
Minor comments that can be addressed in follow ups.

pkg/config/orderer_test.go
caod123: it might be because when I copy the output from my terminal the text wrapping is removing tab characters? feels unlikely but I'll keep an eye out again for when it happens. I'll leave as is for now until we have to change this expected when we update something that touches it again

pkg/config/orderer_test.go
caod123: I'm going to end up standardizing to an Address type because Anchor Peer logic already supports that and it's easier to convert in one direction than the other.

pkg/config/capabilities.go
```diff
@@ -0,0 +1,80 @@
...
+		return nil, nil
+	}
+
+	capabilitiesProto := &cb.Capabilities{}
+	err := proto.Unmarshal(capabilitiesValue.Value, capabilitiesProto)
+	if err != nil {
+		return nil, fmt.Errorf("unmarshalling capabilities: %v", err)
+	}
+
+	capabilities := map[string]bool{}
```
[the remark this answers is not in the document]
caod123: This is a map that needs to be initialized because the line below it sets a value in the map

pkg/config/orderer.go
```diff
@@ -60,6 +61,32 @@ type Kafka struct {
...
+// a channel configuration when the ConsensusType.Type is set to "etcdraft".
+type EtcdRaft struct {
+	Consenters []Consenter
+	Options    EtcdRaftOptions
+}
+
+// Consenter represents a consenting node (i.e. replica).
+type Consenter struct {
+	Host          string
+	Port          uint32
```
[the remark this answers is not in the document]
caod123: As mentioned below I ended up changing all address handling to use an Address type with a Host and integer Port because it's easier to string format that where the protos need it than the other way around (parsing the host and port for a url gets somewhat tricky when determining what's valid, if they can specify a protocol at the front etc among other things)

PR #896, opened 2020-03-22
Add warning message for endorsement policy failure

denyeart:
#### Type of change

- Improvement to logging

#### Description

Many users are unable to troubleshoot endorsement policy failures,
especially for implicit private data collections.

This change shifts a debug message to a warning message so that peer
will log the reason for the endorsement policy failure.

E.g. the following warning message will now help users troubleshoot the
reason for the failure:
"identity 0 does not satisfy principal: the identity is a member
of a different MSP (expected Org2MSP, got Org1MSP)"

### DIFF
--- a/common/cauthdsl/cauthdsl.go
+++ b/common/cauthdsl/cauthdsl.go
@@ -76,7 +76,7 @@ func compile(policy *cb.SignaturePolicy, identities []*mb.MSPPrincipal) (func([]
 				}
 				err := sd.SatisfiesPrincipal(signedByID)
 				if err != nil {
-					cauthdslLogger.Debugf("%p identity %d does not satisfy principal: %s", signedData, i, err)
+					cauthdslLogger.Warnf("%p identity %d does not satisfy principal: %s", signedData, i, err)
 					continue
 				}
 				cauthdslLogger.Debugf("%p principal evaluation succeeds for identity %d", signedData, i)


common/cauthdsl/cauthdsl.go
```diff
@@ -76,7 +76,7 @@ func compile(policy *cb.SignaturePolicy, identities []*mb.MSPPrincipal) (func([]
 				}
 				err := sd.SatisfiesPrincipal(signedByID)
```
yacovm, line 77: I'm not sure we should do this. The way policy is evaluated is by exhaustive search. This means that if you have an endorsement policy of AND(Org1, Org2) then if you have signatures from 2 and then from 1, you will see a warning despite the EP being satisfied.

yacovm (commented):
.

jyellick (changes requested):
-2

Just like with #897 please actually look at what this does to the logs.  It is perfectly normal and expected that an identity will not satisfy principals when evaluating a policy.

This change is even more problematic than the implicit meta policy one, as in this case, a channel with 20 orgs, and someone submitting a transaction will, in the worst case, emit 19 warnings before the outer 'ANY' implicit policy succeeds.

Once more, the 'proper' way to handle this, would be to build up an error message, and pass it back to the caller.  Only if the original top level policy invocation fails should the log be emitted at an appropriate log level.

denyeart:
Ok, it just looked so good in my single org channel logs :-)

PR #897, opened 2020-03-22
Add warning message for implicit meta endorsement policy failure

denyeart:
#### Type of change

- Improvement to logging

#### Description

Many users are unable to troubleshoot endorsement policy failures.

This change shifts a debug message to a warning message so that peer
will log the reason for the endorsement policy failure.

E.g. the following warning message will now help users troubleshoot the
reason for the failure:

"Evaluation Failed: Only 0 policies were satisfied,
but needed 2 of [ Org2MSP/Endorsement Org1MSP/Endorsement ]"

Also, the policy name is now added to associated error messages,
so that users know where to change the policy.

### DIFF
--- a/common/policies/implicitmeta.go
+++ b/common/policies/implicitmeta.go
@@ -73,7 +73,7 @@ func (imp *ImplicitMetaPolicy) EvaluateSignedData(signatureSet []*protoutil.Sign
 	defer func() {
 		if remaining != 0 {
 			// This log message may be large and expensive to construct, so worth checking the log level
-			if logger.IsEnabledFor(zapcore.DebugLevel) {
+			if logger.IsEnabledFor(zapcore.WarnLevel) {
 				var b bytes.Buffer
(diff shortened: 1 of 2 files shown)


jyellick (changes requested):
-2

Have you actually looked at logs with this tracing enabled?  In particular, look at an orderer log.  I expect you will see this warning spewed constantly, even when there are no policy failures.

The meta-policy concept is that result of the policy is a combination of the results of the sub-policies.  So, for something like the /Channel/Writers policy, which evaluates the /Channel/Orderer/Writers policy and the /Channel/Application/Writers policy, it's expected that usually, one of these policy evaluations will fail.  So, with every transaction in the log, you will see a warning message.

This is fixable, but requires a much more invasive change.  The correct way to fix this, is to build up the log message during the evaluations, but not evaluate it unless the top-level policy fails.  Essentially, pass back the "If I fail" log message each time, and eventually, once the caller (who is the only one who is aware whether the policy evaluation actually failed) can emit a message at an appropriate log level.

denyeart:
Ok, it just looked so good in my single org channel logs :-)

PR #899, opened 2020-03-23
[FAB-17574] Support adding an application org to a channel config

wuxuer:
Description

Modify an existing channel configuration to add an application org.

Related issues

https://jira.hyperledger.org/browse/FAB-17574

#### Type of change

- New feature


pkg/config/application_test.go
```diff
@@ -10,10 +10,12 @@ import (
 	"bytes"
 	"testing"
 
+	"github.com/hyperledger/fabric/common/tools/protolator"
+	"github.com/hyperledger/fabric/common/tools/protolator/protoext/peerext"
+
```
DereckLuo, line 15: nit: empty line in import.

DereckLuo (commented):
LGTM

pkg/config/application.go
```diff
@@ -209,3 +209,18 @@ func getApplicationOrg(config *cb.Config, orgName string) (*cb.ConfigGroup, erro
 	}
 	return org, nil
 }
+
+// AddApplicationOrg adds an organization to an existing config's Application configuration.
```
sykesm, line 213:
```suggestion
// AddApplicationOrg adds an organization to an existing Application configuration.
```

We really need to move away from possessives in the doc. I realize this is not the only place we’re doing it but I wanted to highlight it.

pkg/config/application_test.go
sykesm: agreed; we don’t need to separate these imports into a group.

sykesm (approved):

PR #901, opened 2020-03-23
Generate all certs and CRLs in pkg/config unit tests

wlahti:
#### Type of change

- Test update
- Bug fix

#### Description

- Generate certs and CRLs for unit tests instead of relying on fixtures.
- Fix bug where adding an org with a name that is already defined in the consortium was overwriting the existing org config (found while updating tests to use the generated certs/CRLs). 

#### Related issues

[FAB-17621](https://jira.hyperledger.org/browse/FAB-17621)
[FAB-17645](https://jira.hyperledger.org/browse/FAB-17645)


pkg/config/msp_test.go
```diff
@@ -23,57 +25,6 @@ import (
 	. "github.com/onsi/gomega"
 )
```
sykesm, line 27: Thank you.

pkg/config/msp_test.go
```diff
@@ -801,7 +774,7 @@ func TestAddRootCAToMSP(t *testing.T) {
 	},
 	"sequence": "0"
 }
-	`
+`, org1CertBase64, org1CertBase64, org1CertBase64, org1CertBase64, org1CertBase64, org1CertBase64, org1CertBase64, org1CRLBase64, org1CertBase64, certBase64, org1PKBase64, org1CertBase64, org1CertBase64, org1CertBase64, org2CertBase64, org2CertBase64, org2CertBase64, org2CertBase64, org2CertBase64, org2CertBase64, org2CertBase64, org2CRLBase64, org2CertBase64, org2PKBase64, org2CertBase64, org2CertBase64, org2CertBase64)
```
sykesm, line 777: This works but there’s a lot of repetition. Two alternatives to consider:

- Use indexes in the format string so you can reference the same item. `%[1]s` would be the first argument; you can use that format string multiple times.
- Consider using a template instead of a simple printf. Then you can use a struct with named fields and reference the fields in the template with `{{Org1Cert}}` style names.

pkg/config/msp_test.go
```diff
@@ -1260,14 +1252,15 @@ func TestRevokeCertificateFromMSPFailure(t *testing.T) {
...
-
-	privKeyBlock, _ := pem.Decode([]byte(dummyPrivateKey))
-	privKey, _ := x509.ParsePKCS8PrivateKey(privKeyBlock.Bytes)
-
-	crlBlock, _ := pem.Decode([]byte(dummyCRL))
-	crl, _ := x509.ParseCRL(crlBlock.Bytes)
+	cert, privKey := generateCACertAndPrivateKey("org1.example.com")
+	crlBytes, err := cert.CreateCRL(rand.Reader, privKey, nil, time.Now(), time.Now().Add(YEAR))
+	if err != nil {
+		log.Fatalf("Failed to create CRL: %s", err)
```
sykesm, line 1258: I think @DereckLuo may have a PR related to this but, functions like `baseMSP` are test helpers and can be provided with a `*testing.T`. We should be wiring that through and making assertions as necessary in preference to a panic when appropriate.

This applies to the functions that follow as well.

sykesm (approved):

pkg/config/msp_test.go
wlahti: Ah, I knew there had to be a way built into printf to handle using the same argument multiple times but couldn't figure out the right way to search for that functionality. :) Thanks. Yeah, I didn't like doing it this way but wanted to get something working for now.

pkg/config/msp_test.go
DereckLuo: I can make the changes in the same pr.

PR #908, opened 2020-03-24
Cleanup followup comments from FAB-17568

caod123:
* Standardize Address type and wherever an endpoint is used in the API
* Cleanup slice filtering
* Move some exported functions to the top of respective file

### DIFF
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -13,6 +13,7 @@ import (
 	"encoding/pem"
 	"errors"
 	"fmt"
+	"strconv"
 	"strings"
 	"time"
 
@@ -46,8 +47,15 @@ type Organization struct {
 	Policies map[string]Policy
 	MSP      MSP
 
-	AnchorPeers      []AnchorPeer
+	// AnchorPeers contains the endpoints of anchor peers for each application organization.
+	AnchorPeers      []Address
 	OrdererEndpoints []string
+}
+
+// Address contains the hostname and port for an endpoint.
+type Address struct {
+	Host string
+	Port int
 }
 
 type standardConfigValue struct {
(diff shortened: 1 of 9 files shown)


sykesm:
Unit Test failure is [FAB-17655](https://jira.hyperledger.org/browse/FAB-17655).

pkg/config/config.go
```diff
@@ -46,10 +47,17 @@ type Organization struct {
...
+	AnchorPeers      []Address
 	OrdererEndpoints []string
 }
 
+// Address contains the hostname and port for an endpoint.
+type Address struct {
+	Host string
+	Port int
+}
+
```
sykesm, line 60: What about a constructor to create one of these from a "host:port" string?

Alternatively, we could have made Address something like

```
type Address string

func (a Address) Host() string { ... }
func (a Address) Port() int { ... }
```

There are ways to bridge the gap but, at the end of the day, none of these are heavy hitters. The proto definition migrated away from the common host:port representation.

pkg/config/config.go
```diff
@@ -46,10 +47,17 @@ type Organization struct {
 	Policies map[string]Policy
 	MSP      MSP
 
-	AnchorPeers      []AnchorPeer
+	// AnchorPeers contains the endpoints of anchor peers for each application organization.
+	AnchorPeers      []Address
 	OrdererEndpoints []string
 }
 
+// Address contains the hostname and port for an endpoint.
```
sykesm, line 55: field is host, "host name" is English, and "hostname" is jargon. Which do we want?

sykesm (approved):
Works for me. Comments are mostly nits and alternative approaches for consideration.

PR #914, opened 2020-03-25
[FAB-17638] Support retrieving application configuration from an existing config

wuxuer:

#### Type of change

- New feature

#### Description
1.  Retrieve application configuration from an existing config;
2.  Fix some nits in https://github.com/hyperledger/fabric/pull/899;

#### Related issues
https://jira.hyperledger.org/browse/FAB-17638

### DIFF
--- a/pkg/config/application.go
+++ b/pkg/config/application.go
@@ -7,6 +7,7 @@ SPDX-License-Identifier: Apache-2.0
 package config
 
 import (
+	"errors"
 	"fmt"
 
 	"github.com/golang/protobuf/proto"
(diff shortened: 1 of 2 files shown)


caod123:
@hellosmallstone it looks like your unit tests for retrieving application config are failing now after rebasing. Apologies for neglecting this PR and making you rebase over and over, we will prioritize getting it reviewed and merged after the unit tests are addressed. Thanks!

pkg/config/application.go
```diff
@@ -217,7 +250,7 @@ func (c *ConfigTx) GetAnchorPeers(orgName string) ([]Address, error) {
 	return anchorPeers, nil
 }
 
-// AddApplicationOrg adds an organization to an existing config's Application configuration.
+// AddApplicationOrg adds an organization to an existing Application configuration.
 // Will not error if organization already exists.
```
caod123, line 254: I know this isn't part of your commit, but I'm wondering if maybe AddApplicationOrg SHOULD error if org already exists. I can't remember if this is consistent with other add/remove behavior in the rest of the package. We can always revisit this when we do some final passes over the package to clean it up

caod123 (commented):
Aside from the failing Unit Tests, LGTM

pkg/config/application.go
wuxuer: While adding orderer org, AddOrdererOrg doesn't judge if org already exists. However, while adding endpoint or anchor peer, the corresponding function returns error if it already exist. Seems to unify this behavior for consistency.

wuxuer:
@caod123 I have fixed the unit tests. 
The main reason for the previous failure is that the definition of the Organization in Application struct is an array, which is obtained by traversing the map structure. However, the map is traversed in random order.  When performing equal comparisons later, random errors may occur.
```
for orgName := **range applicationGroup.Groups** {
    orgConfig, err := c.GetApplicationOrg(orgName)
    ......
    applicationOrgs = append(applicationOrgs, orgConfig)
}
```
Defining Organizations in Application struct as a map type also seems to be an option.

pkg/config/application.go
```diff
@@ -23,19 +24,46 @@ type Application struct {
...
 }
 
-// AddApplicationOrg adds an organization to an existing config's Application configuration.
-// Will not error if organization already exists.
-func AddApplicationOrg(config *cb.Config, org Organization) error {
-	appGroup := config.ChannelGroup.Groups[ApplicationGroupKey]
+// GetApplicationConfiguration returns the existing application configuration values from a config
+// transaction as an Application type. This can be used to retrieve existing values for the application
+// prior to updating the application configuration.
+func (c *ConfigTx) GetApplicationConfiguration() (Application, error) {
```
cendhu, line 30: why are we not returning `*Application` (i mean the pointer to the Appplication)? Any specific reason?

mastersingh24: There's no need ... Application is simply a struct with no member functions.

sykesm (approved):

PR #918, opened 2020-03-25
Optionally silence CLI commands in integration tests

yacovm:
Sometimes during troubleshooting integration tests, we have commands
that either run several times due to Eventually-ies, or are just
chatty (like discover commands that return peers and their information).

This information is many times distracting and makes focusing on the logs
printed by the nodes harder.

This change set adds an option to silence CLIs via a flag.

Change-Id: I3480c4286dc8880b402831c1f9c8de7bd3538feb

### DIFF
--- a/integration/nwo/network.go
+++ b/integration/nwo/network.go
@@ -136,6 +136,7 @@ type Profile struct {
 
 // Network holds information about a fabric network.
 type Network struct {
+	SilenceCLI         bool
 	RootDir            string
 	StartPort          uint16
 	Components         *Components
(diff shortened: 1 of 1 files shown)


sykesm (changes requested):
Suppressing all output at the network level discards everything written by the command. As soon as someone enables that, there will be no debug data.

Without this change, ginkgo simply buffers the output and will only dump it when tests fail. This seems like appropriate and expected behavior.

yacovm:
> Without this change, ginkgo simply buffers the output and will only dump it when tests fail. This seems like appropriate and expected behavior.

With this change, this is the default behavior. 
Only if you opt out of logging CLI output, this has any effect.

> Suppressing all output at the network level discards everything written by the command. As soon as someone enables that, there will be no debug data.

There will be debug data of all nodes (peers, orderers) and sometimes that's all that is interesting to a developer that is running the integration test.

sykesm:
> Only if you opt out of logging CLI output, this has any effect.

The concern is that you're enabling this behavior at the level of all command sessions - programatically. If this is only intended for local debug, it's easy enough to change the code to use ioutil.Discard. If the proposal enabled a more fine-grained approach, I would be less concerned.
 
> There will be debug data of all nodes (peers, orderers) and sometimes that's all that is interesting to a developer that is running the integration test.

And it should come as no surprise that the session output code in nwo behaves the same as ginkgomon - and for the same reasons.

If you only want to see output from a peer or an orderer, each line of output is prefixed with the process type so it's pretty trivial to capture output and filter it with grep. If you want to remove noise from a command, that's also easy to suppress.

While I understand that the noise is annoying (I feel the same way), in automated environments, I'd rather have more data than less.

PR #919, opened 2020-03-26
[FAB-17605] Handling errors returned by InitFactories()

Eggsyz:
#### Type of change

- Bug fix

#### Description

A project using func InitFactories, it output errors when build it. But many places do not handle errors.

Signed-off-by: jilg <liguo_ji@outlook.com>

#### Related issues

[17605](https://jira.hyperledger.org/browse/FAB-17605?filter=-1)

### DIFF
--- a/discovery/client/signer_test.go
+++ b/discovery/client/signer_test.go
@@ -8,6 +8,7 @@ package discovery
 
 import (
 	"crypto/rand"
+	"fmt"
 	"io"
 	"sync"
 	"sync/atomic"
@@ -19,7 +20,10 @@ import (
 )
 
 func init() {
-	factory.InitFactories(nil)
+	err := factory.InitFactories(nil)
+	if err != nil {
+		panic(fmt.Errorf("Could not initialize BCCSP Factories [%s]", err))
+	}
 }
 
 func TestSameMessage(t *testing.T) {
--- a/gossip/comm/comm_test.go
+++ b/gossip/comm/comm_test.go
@@ -42,7 +42,10 @@ import (
 func init() {
 	util.SetupTestLogging()
 	rand.Seed(time.Now().UnixNano())
-	factory.InitFactories(nil)
+	err := factory.InitFactories(nil)
+	if err != nil {
+		panic(fmt.Errorf("Could not initialize BCCSP Factories [%s]", err))
+	}
 	naiveSec.On("OrgByPeerIdentity", mock.Anything).Return(api.OrgIdentityType{})
 }
 
--- a/gossip/gossip/channel/channel_test.go
+++ b/gossip/gossip/channel/channel_test.go
@@ -59,7 +59,10 @@ var disabledMetrics = metrics.NewGossipMetrics(&disabled.Provider{}).MembershipM
 
 func init() {
 	util.SetupTestLogging()
-	factory.InitFactories(nil)
+	err := factory.InitFactories(nil)
+	if err != nil {
+		panic(fmt.Errorf("Could not initialize BCCSP Factories [%s]", err))
+	}
 }
 
 var (
--- a/gossip/gossip/gossip_test.go
+++ b/gossip/gossip/gossip_test.go
@@ -42,7 +42,10 @@ func TestMain(m *testing.M) {
 	util.SetupTestLogging()
 	rand.Seed(int64(time.Now().Second()))
 	discovery.SetMaxConnAttempts(5)
-	factory.InitFactories(nil)
+	err := factory.InitFactories(nil)
+	if err != nil {
+		panic(fmt.Errorf("Could not initialize BCCSP Factories [%s]", err))
+	}
 	os.Exit(m.Run())
 }
 
--- a/gossip/gossip/orgs_test.go
+++ b/gossip/gossip/orgs_test.go
@@ -32,7 +32,10 @@ import (
 
 func init() {
 	util.SetupTestLogging()
-	factory.InitFactories(nil)
+	err := factory.InitFactories(nil)
+	if err != nil {
+		panic(fmt.Errorf("Could not initialize BCCSP Factories [%s]", err))
+	}
 }
 
 type configurableCryptoService struct {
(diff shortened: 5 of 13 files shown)


discovery/client/signer_test.go
```diff
@@ -19,7 +20,10 @@ import (
 )
 
 func init() {
-	factory.InitFactories(nil)
+	err := factory.InitFactories(nil)
+	if err != nil {
+		panic(fmt.Errorf("Could not initialize BCCSP Factories [%s]", err))
```
denyeart: assert.NoError() here as well.

gossip/comm/comm_test.go
```diff
@@ -42,7 +42,10 @@ import (
 func init() {
 	util.SetupTestLogging()
 	rand.Seed(time.Now().UnixNano())
-	factory.InitFactories(nil)
+	err := factory.InitFactories(nil)
+	if err != nil {
+		panic(fmt.Errorf("Could not initialize BCCSP Factories [%s]", err))
```
denyeart: assert.NoError() here as well.

gossip/gossip/channel/channel_test.go
```diff
@@ -59,7 +59,10 @@ var disabledMetrics = metrics.NewGossipMetrics(&disabled.Provider{}).MembershipM
 
 func init() {
 	util.SetupTestLogging()
-	factory.InitFactories(nil)
+	err := factory.InitFactories(nil)
+	if err != nil {
+		panic(fmt.Errorf("Could not initialize BCCSP Factories [%s]", err))
```
denyeart: assert.NoError() here as well.

gossip/gossip/gossip_test.go
```diff
@@ -42,7 +42,10 @@ func TestMain(m *testing.M) {
 	util.SetupTestLogging()
 	rand.Seed(int64(time.Now().Second()))
 	discovery.SetMaxConnAttempts(5)
-	factory.InitFactories(nil)
+	err := factory.InitFactories(nil)
+	if err != nil {
+		panic(fmt.Errorf("Could not initialize BCCSP Factories [%s]", err))
```
denyeart, line 47: assert.NoError() here as well.

gossip/gossip/orgs_test.go
```diff
@@ -32,7 +32,10 @@ import (
 
 func init() {
 	util.SetupTestLogging()
-	factory.InitFactories(nil)
+	err := factory.InitFactories(nil)
+	if err != nil {
+		panic(fmt.Errorf("Could not initialize BCCSP Factories [%s]", err))
```
denyeart: assert.NoError() here as well.

gossip/state/state_test.go
```diff
@@ -71,7 +71,10 @@ type joinChanMsg struct {
 
 func init() {
 	gutil.SetupTestLogging()
-	factory.InitFactories(nil)
+	err := factory.InitFactories(nil)
+	if err != nil {
+		panic(fmt.Errorf("Could not initialize BCCSP Factories [%s]", err))
```
denyeart: assert.NoError() here as well.

internal/configtxgen/encoder/encoder_suite_test.go
```diff
@@ -15,8 +16,10 @@ import (
 )
 
 func TestEncoder(t *testing.T) {
-	factory.InitFactories(nil)
-
+	err := factory.InitFactories(nil)
+	if err != nil {
+		panic(fmt.Errorf("Could not initialize BCCSP Factories [%s]", err))
```
denyeart: assert.NoError() here as well.

orderer/common/cluster/deliver_test.go
```diff
@@ -44,7 +44,10 @@ import (
 var gRPCBalancerLock = sync.Mutex{}
 
 func init() {
-	factory.InitFactories(nil)
+	err := factory.InitFactories(nil)
+	if err != nil {
+		panic(fmt.Errorf("Could not initialize BCCSP Factories [%s]", err))
```
denyeart: assert.NoError() here as well.

orderer/consensus/etcdraft/chain_test.go
```diff
@@ -55,7 +55,10 @@ const (
 )
 
 func init() {
-	factory.InitFactories(nil)
+	err := factory.InitFactories(nil)
+	if err != nil {
+		panic(fmt.Errorf("Could not initialize BCCSP Factories [%s]", err))
```
denyeart: assert.NoError() here as well.

denyeart (dismissed):

MHBauer (dismissed):
@denyeart how to get a `testing.T` in the `init()` functions?

denyeart (commented):
Ah, you're right, removed the comment from the init functions.

gossip/gossip/gossip_test.go
Eggsyz: this is testing.M, not testing.T

denyeart:
I'll defer to @sykesm on whether the panics are the most appropriate way to handle the errors.

sykesm:
The *vast* majority of calls to InitFactory simply pass in a nil configuration. If that function does not work, it's pretty much always fatal as there's little on the default config path that can go wrong; it's wired to work. The error return is only for cases where an explicit configuration is provided.

The places where we pass in an explicit configuration, the call can fail. It looks like there are only two places where that is the case and only one of those is production code:

- `msp/configbuilder.go`
- `core/testutil/config.go`

The orderer and peer implementations initialize bccsp factories by side-effect through direct or indirect calls to `msp.GetLocalMspConfig` so that's the only place where an error really can happen.

In the peer case, the path is through `InitCrypto` and, if that call fails, the process is terminated with an `os.Exit(1)`. In the orderer case, `GetLocalMspConfig` is called directly and, if that call fails, the process is terminated with a panic.

So... IMO, given how poorly structured bccsp is, if `InitFactories` fails, it should simply panic. The function shouldn't bother returning an error. While I generally prefer error returns to panic, crypto is so fundamentally core to Fabric that failures to set it up due to problems in the configuration or environment are always terminal. We should codify that.

PR #923, opened 2020-03-26
[FAB-17637] Support adding and removing capability

stephyee:
#### Type of change

- New feature

#### Description

- Support adding and removing a capability to Channel, Application, and Orderer
- Change capabilities map to slice

#### Related issues
[FAB-17637](https://jira.hyperledger.org/browse/FAB-17637)

### DIFF
--- a/pkg/config/capabilities.go
+++ b/pkg/config/capabilities.go
@@ -16,7 +16,7 @@ import (
 
 // GetChannelCapabilities returns a map of enabled channel capabilities
 // from a config transaction.
-func (c *ConfigTx) GetChannelCapabilities() (map[string]bool, error) {
+func (c *ConfigTx) GetChannelCapabilities() ([]string, error) {
 	capabilities, err := getCapabilities(c.base.ChannelGroup)
 	if err != nil {
 		return nil, fmt.Errorf("retrieving channel capabilities: %v", err)
@@ -27,7 +27,7 @@ func (c *ConfigTx) GetChannelCapabilities() (map[string]bool, error) {
 
 // GetOrdererCapabilities returns a map of enabled orderer capabilities
 // from a config transaction.
-func (c *ConfigTx) GetOrdererCapabilities() (map[string]bool, error) {
+func (c *ConfigTx) GetOrdererCapabilities() ([]string, error) {
 	orderer, ok := c.base.ChannelGroup.Groups[OrdererGroupKey]
 	if !ok {
 		return nil, errors.New("orderer missing from config")
@@ -43,7 +43,7 @@ func (c *ConfigTx) GetOrdererCapabilities() (map[string]bool, error) {
 
 // GetApplicationCapabilities returns a map of enabled application capabilities
 // from a config transaction.
-func (c *ConfigTx) GetApplicationCapabilities() (map[string]bool, error) {
+func (c *ConfigTx) GetApplicationCapabilities() ([]string, error) {
 	application, ok := c.base.ChannelGroup.Groups[ApplicationGroupKey]
 	if !ok {
 		return nil, errors.New("application missing from config")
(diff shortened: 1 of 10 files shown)


pkg/config/capabilities.go
```diff
@@ -57,7 +57,153 @@ func (c *ConfigTx) GetApplicationCapabilities() (map[string]bool, error) {
 	return capabilities, nil
 }
 
-func getCapabilities(configGroup *cb.ConfigGroup) (map[string]bool, error) {
+// AddChannelCapability adds capability to the provided channel config,
+// capability must be specified.
```
caod123: Regarding this comment, where does it enforce the fact that the capability being added must be specified.

pkg/config/capabilities.go
```diff
@@ -57,7 +57,153 @@ func (c *ConfigTx) GetApplicationCapabilities() (map[string]bool, error) {
...
+	err = addCapability(c.updated.ChannelGroup, capabilities, AdminsPolicyKey, capability)
+	if err != nil {
+		return err
+	}
+
+	return nil
+}
+
+// AddOrdererCapability adds capability to the provided channel config,
+// capability must be specified.
```
caod123: Same as above

pkg/config/capabilities.go
```diff
@@ -57,7 +57,153 @@ func (c *ConfigTx) GetApplicationCapabilities() (map[string]bool, error) {
...
+	err = addCapability(c.updated.ChannelGroup.Groups[OrdererGroupKey], capabilities, AdminsPolicyKey, capability)
+	if err != nil {
+		return err
+	}
+
+	return nil
+}
+
+// AddApplicationCapability adds capability to the provided channel config,
+// capability must be specified.
```
caod123: Same as above

pkg/config/capabilities.go
```diff
@@ -57,7 +57,153 @@ func (c *ConfigTx) GetApplicationCapabilities() (map[string]bool, error) {
...
+	err = removeCapability(c.updated.ChannelGroup, capabilities, AdminsPolicyKey, capability)
+	if err != nil {
+		return err
+	}
+
+	return nil
+}
+
+// RemoveOrdererCapability removes capability to the provided channel config,
+// capability must be specified.
```
caod123: Same as above

pkg/config/capabilities.go
```diff
@@ -57,7 +57,153 @@ func (c *ConfigTx) GetApplicationCapabilities() (map[string]bool, error) {
...
+	err = removeCapability(c.updated.ChannelGroup.Groups[OrdererGroupKey], capabilities, AdminsPolicyKey, capability)
+	if err != nil {
+		return err
+	}
+
+	return nil
+}
+
+// RemoveApplicationCapability removes capability to the provided channel config,
+// capability must be specified.
```
caod123: Same as above

pkg/config/capabilities.go
```diff
@@ -57,7 +57,153 @@ func (c *ConfigTx) GetApplicationCapabilities() (map[string]bool, error) {
...
+			if err != nil {
+				return fmt.Errorf("removing capability: %v", err)
+			}
+
+			return nil
+		}
+	}
+
+	return errors.New("capability not set")
+}
```
sykesm: Since duplicates are not expected, this works but it makes me uneasy to see a removal during iteration. I'd much rather see an approach that does not rely on unique values.

Also: https://github.com/golang/go/wiki/SliceTricks

pkg/config/capabilities.go
```diff
@@ -57,7 +57,153 @@ func (c *ConfigTx) GetApplicationCapabilities() (map[string]bool, error) {
...
+	}
+
+	return nil
+}
+
+func removeCapability(configGroup *cb.ConfigGroup, capabilities []string, modPolicy string, capability string) error {
+	for i, c := range capabilities {
+		if c == capability {
+			// remove the capability from slice
+			capabilities = append(capabilities[:i], capabilities[i:]...)
```
caod123: Matt left a comment a while back on some slice filtering tricks. Refer to https://github.com/hyperledger/fabric/blob/master/pkg/config/application.go#L103-L120 for an example in our package

pkg/config/capabilities.go
```diff
@@ -57,7 +57,153 @@ func (c *ConfigTx) GetApplicationCapabilities() (map[string]bool, error) {
...
+
+			err := addValue(configGroup, capabilitiesValue(capabilities), modPolicy)
```
sykesm: `addValue` getting called by `removeCapability` makes me wonder about the name. Makes me think `addValue` should be `setValue`.

sykesm (commented):
The remove loop makes me uneasy.

pkg/config/example_test.go
```diff
@@ -1417,3 +1417,399 @@ func ExampleNewCreateChannelTx() {
...
+}
+
+func ExampleConfigTx_AddOrdererCapability() {
+	baseConfig := fetchChannelConfig()
+	c := config.New(baseConfig)
+
+	err := c.AddOrdererCapability("V1_4")
+	if err != nil {
+		panic(err)
+	}
```
caod123, line 1774: Not sure if you wanted to add the protolator output here and the ones below

caod123 (commented):
Mostly nits

pkg/config/capabilities.go
stephyee: It doesn't enforce it, only expected. I can remove this from the doc.

pkg/config/example_test.go
stephyee: I did not add the output intentionally - given how little value it adds.

pkg/config/capabilities.go
stephyee: I'll throw a follow-up on the backlog for this given most of the removal functions use `addValue`.

pkg/config/capabilities.go
stephyee: 🛠

pkg/config/capabilities.go
stephyee: 🛠

pkg/config/capabilities.go
```diff
@@ -57,7 +57,150 @@ func (c *ConfigTx) GetApplicationCapabilities() (map[string]bool, error) {
 	return capabilities, nil
 }
 
-func getCapabilities(configGroup *cb.ConfigGroup) (map[string]bool, error) {
+// AddChannelCapability adds capability to the provided channel config.
```
sykesm, line 60: All of these godoc comments seem out of date at this point. We need a follow up to address these (if there isn't one already).

sykesm (approved):

PR #924, opened 2020-03-26
[FAB-17664] Add/Remove channel policies

DereckLuo:

#### Type of change

- New feature

#### Description
* Implemented add/remove channel policies.

#### Related issues
[FAB-17664](https://jira.hyperledger.org/browse/FAB-17664)

### DIFF
--- a/pkg/config/policies.go
+++ b/pkg/config/policies.go
@@ -114,6 +114,11 @@ func (c *ConfigTx) GetPoliciesForApplicationOrg(orgName string) (map[string]Poli
 	return getPolicies(orgGroup.Policies)
 }
 
+// GetPoliciesForChannel returns a map of policies for channel configuration.
+func (c *ConfigTx) GetPoliciesForChannel() (map[string]Policy, error) {
+	return getPolicies(c.base.ChannelGroup.Policies)
+}
+
 // AddApplicationPolicy modifies an existing application policy configuration.
 // When the policy exists it will overwrite the existing policy.
 func (c *ConfigTx) AddApplicationPolicy(modPolicy, policyName string, policy Policy) error {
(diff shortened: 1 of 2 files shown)


pkg/config/policies.go
```diff
@@ -219,6 +224,23 @@ func (c *ConfigTx) RemoveOrdererOrgPolicy(orgName, policyName string) error {
 	return removePolicy(c.updated.ChannelGroup.Groups[OrdererGroupKey].Groups[orgName], policyName, policies)
 }
 
+// AddChannelPolicy modifies an existing channel policy configuration.
+// When the policy exists it will overwrite the existing policy.
```
caod123, line 253: Is this the ideal behavior for Add? Not saying it's wrong but we should consider whether we should allow the user to overwrite existing policies via the Add method instead of a Replace method or via a Delete followed by an Add

pkg/config/policies.go
```diff
@@ -219,6 +224,23 @@ func (c *ConfigTx) RemoveOrdererOrgPolicy(orgName, policyName string) error {
...
+func (c *ConfigTx) AddChannelPolicy(modPolicy, policyName string, policy Policy) error {
+	return addPolicy(c.updated.ChannelGroup, modPolicy, policyName, policy)
+}
+
+// RemoveChannelPolicy removes an existing channel policy configuration.
```
caod123: Mentioning configuration at the end is a bit wordy. Probably good enough to say `RemoveChannelPolicy removes an existing channel level policy.`

caod123 (commented):
small comments, mostly some function comments and considerations

pkg/config/policies.go
```diff
@@ -219,6 +224,23 @@ func (c *ConfigTx) RemoveOrdererOrgPolicy(orgName, policyName string) error {
 	return removePolicy(c.updated.ChannelGroup.Groups[OrdererGroupKey].Groups[orgName], policyName, policies)
 }
 
+// AddChannelPolicy modifies an existing channel policy configuration.
```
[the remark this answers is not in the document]
DereckLuo: done

pkg/config/policies.go
DereckLuo: make sense, thought about the same thing during the implementation. 
However, currently all add methods in policies.go share the same implementation, it's probably worth it to consider the ideal behavior, and apply it over the entire package.

pkg/config/policies.go
DereckLuo: done.

DereckLuo:
rebased.

pkg/config/policies.go
sykesm: Add != Set; Set could imply Add but not the other way round, IMO.

sykesm (approved):

PR #895, opened 2020-03-20
[FAB-17611] IT: discover all peers before endorsement

wenjianqiao:
Signed-off-by: Wenjian Qiao <wenjianq@gmail.com>

#### Type of change
- Test update

#### Description
To mitigate a test flake with private data dissemination,
this PR adds code to discover all the peers to ensure that
the gossip service has been successfully initialized.

#### Related issues
https://jira.hyperledger.org/browse/FAB-17611

PR #925, opened 2020-03-26
[FAB-17675] Prevent gossip probe from registering as a connection

yacovm:
Part of FAB-17672:

Gossip bootstrap peer and anchor peer connection establishments are a two step process:

I)  Connect to the remote endpoint, and figure out its organization association
II) Connect to the remote endpoint once again, this time sending membership information according to what it is eligible to.

Unfortunately, if the responding peer connects to the initiator peer at the same time, the connection of (1) will be overwritten by the connection (gossip keeps a single connection among every pair of peers) from the responder peer and the initiator peer will consider the responder peer as dead.

In production, where peers run for a long time and constnatly retry, this is not a problem.

However, in integration tests that expect data to be disseminated in a timely manner, this might cause flakes.

this change set adds a flag that indicates this connection is not a long lasting one, and thus there is no point to add it to the connection store.

Change-Id: Id8f4ee12145748527233301356af830c3401ee02


gossip/comm/comm_impl.go
```diff
@@ -37,6 +37,10 @@ const (
 	DefSendBuffSize  = 20
 )
 
+var (
+	errProbe = errors.New("probe")
```
cendhu, line 41: Can't we have a more descriptive error message for better readability? 

Correct me if I have misunderstood: this error would be returned when the peer tries to authenticate a stream which is not a probe (by passing an appropriate flag) but it ends up being a probe. As a result, we don't store info in the connection store. If my understanding is correct, I am not sure whether it should be an error. Maybe a bool can be used instead as I mentioned in the other comment.

gossip/comm/comm_impl.go
```diff
@@ -163,7 +167,7 @@ func (c *commImpl) createConnection(endpoint string, expectedPKIID common.PKIidT
 
 	ctx, cancel = context.WithCancel(context.Background())
 	if stream, err = cl.GossipStream(ctx); err == nil {
-		connInfo, err = c.authenticateRemotePeer(stream, true)
+		connInfo, err = c.authenticateRemotePeer(stream, true, false)
 		if err == nil {
```
cendhu, line 171: Unrelated to this CR. Can be `err != nil` and return early to avoid the extra indentation.

gossip/comm/comm_impl.go
```diff
@@ -556,7 +564,13 @@ func (c *commImpl) GossipStream(stream proto.Gossip_GossipStreamServer) error {
 	if c.isStopping() {
 		return fmt.Errorf("Shutting down")
 	}
-	connInfo, err := c.authenticateRemotePeer(stream, false)
+	connInfo, err := c.authenticateRemotePeer(stream, false, false)
+
+	if err == errProbe {
```
cendhu, line 569: It can be confusing to the code reader. Some questions might arise like why we are ignoring an error. If we are trying to not store info in the conn store for probs, can't the return value be an `isProbe` bool?

yacovm: No, 3 return values is ugly, and this is an un-exported function.
It's perfectly fine to have a specific error that indicates something. 

Look at:

> if e, ok := err.(*QueryError); ok && e.Err == ErrPermission {
    // query failed because of a permission problem
}

From https://blog.golang.org/go1.13-errors.

gossip/comm/comm_impl.go
yacovm: This is by design.
I need to close the gRPC connection, and also to cancel the gRPC stream. 
So it's not just an early return. I want cleanup operations to be at the bottom of the function so that I manage them at the same place.

gossip/comm/comm_impl.go
yacovm: You misunderstood. This error just means that the remote peer probed us, so there is no point in doing anything with the returned first value.

cendhu: Okay. That's what I meant may be my wordings didn't convey that. That's why I suggested `isProbe bool` later in the comment.

gossip/comm/comm_impl.go
cendhu: I see. Got it.

gossip/comm/comm_impl.go
cendhu: Thanks for the clarification. Agree that 3 return value is ugly.

gossip/comm/comm_impl.go
yacovm: I see a function that returns an object and an error in the following way:
If the error is nil, use the first return value.
Else, use the other return value.

This perfectly fits my needs here.

wenjianqiao:
This PR helps to fix the test flake in FAB-17611. The changes have been included in https://github.com/hyperledger/fabric/pull/895 for verification purpose.

C0rWin (approved):

PR #927, opened 2020-03-27
move txvalidation flag util to internal/pkg from ledger

cendhu:
#### Type of change

#### Description

The [txvalidation-flag util package](https://github.com/hyperledger/fabric/blob/master/core/ledger/util/txvalidationflags.go) helps in creating transaction validation flags that would be assigned to the block metadata. This function is used by the [v14](https://github.com/hyperledger/fabric/tree/master/core/committer/txvalidator/v14) and [v20](https://github.com/hyperledger/fabric/tree/master/core/committer/txvalidator/v20) validation as it is needed by these packages to record invalid tx as per policy evaluation. This validation flag is also reused in the [mvcc validator ](https://github.com/hyperledger/fabric/tree/master/core/ledger/kvledger/txmgmt/validator) to mark valid/invalid transactions. Further, the txvalidation-flag util is referred in many other packages including test, hence, we move the util to `internal/pkg/txflags`

#### Additional details

Before starting the ledger checkpointing code, we would like to refactor the ledger package. If we continue to add new features with the current package structure, after a few years, it will be unmaintainable. We plan to have nearly 10 such PRs. Note that each PR may modify many files and we cannot do much about it. We might be okay to get some review burden than having an unmaintainable code in the longer term.

#### Related issues

[FAB-17677](https://jira.hyperledger.org/browse/FAB-17677)


lindluni:
Hit Flake: FAB-17611
Has potential fix #895

core/committer/txvalidator/txvalidationflags.go
```diff
@@ -4,7 +4,7 @@ Copyright IBM Corp. All Rights Reserved.
 SPDX-License-Identifier: Apache-2.0
 */
 
-package util
+package txvalidator
```
mastersingh24: I think we should actually move this to `internal/pkg/txflags`.
`internal/pkg` was created for code that is shared across multiple packages.

mastersingh24 (changes requested):

core/committer/txvalidator/txvalidationflags.go
cendhu: Nice. Thanks for the pointer. I wasn't aware of this `internal/pkg`. Will update the code.

mastersingh24: Yeah ... I'm glad you started on this work again.  I've been meaning to get back to (re)organizing a bunch of the code myself

cendhu: Yes, can't agree more. It is a much needed one.

common/ledger/blkstorage/fsblkstorage/blockfile_mgr_test.go
```diff
@@ -209,7 +209,7 @@ func TestBlockfileMgrGetTxByIdDuplicateTxid(t *testing.T) {
 		},
 		[]string{"txid-1", "txid-2", "txid-1"},
 	)
-	txValidationFlags := ledgerutil.NewTxValidationFlags(3)
+	txValidationFlags := txflags.NewTxValidationFlags(3)
```
manish-sethi: After moving this here, this functions (and the other one) can be renamed to remove redundant info in the package name and the function name. For instance, this function could simply be renamed as "New".

manish-sethi (commented):
LGTM. Just a nit-picking comment. I am fine if you do not want to take that suggestion. 
This PR got into a conflict now and I am fine with merging after you rebase

common/ledger/blkstorage/fsblkstorage/blockfile_mgr_test.go
cendhu: yes, I agree. Will do.

cendhu: The following changes are done. 
`TxValidationFlags` -> `ValidationFlags`
`NewTxValidationFlags()` -> `New()`
`NewTxValidationFlagsSetValue()` -> `NewWithValues()`

manish-sethi: > The following changes are done.
> `TxValidationFlags` -> `ValidationFlags`
> `NewTxValidationFlags()` -> `New()`
> `NewTxValidationFlagsSetValue()` -> `NewWithValues()`

Looks good. I guess that "TxValidationFlags" could be changed to "Values" to match others renames. But I am fine merging this as the larger objective is achieved.

manish-sethi (approved):

common/ledger/blkstorage/fsblkstorage/blockfile_mgr_test.go
cendhu: If we do that, wouldn't API become little ugly?  -- `Values` vs `NewWithValues()` 
`TxValidationFlags` -> `Values`
`NewTxValidationFlags()` -> `New()`
`NewTxValidationFlagsSetValue()` -> `NewWithValues()`

The reasoning behind the current naming is that t`xflags.ValidationFlags` denotes that we have `txflags` and one type of the flag is validation flags. I know that we will not have any more types but this makes code more readable IMO.

cendhu: I have contradicted with myself there. With that reasoning, even the `New()` can be made `NewValidationFlags()`. Let's see what Gari suggests.

manish-sethi: As I had mentioned earlier, I am fine with any consistent naming. One alternate could be to use `txvalidationflags` for the package name, so we don't miss `validation` in the name and at the same time avoid repetition of `flags` in package name and in the name of functions/types.

cendhu: Let's go with what we have now. Don't want to get blocked on this.

cendhu:
@mastersingh24 your approval is still pending. Can you please ack on the following?

@manish-sethi and I discussed offline on the package/variable/init function naming. Though it is very minor, we thought of doing it in the correct way as we are already changing 29 files. Though these discussions might look unnecessary, we feel that this knowledge will help in future PRs.

**Proposed Convention.** 
_package name_: `internal/pkg/txvalidation`
_variable name_: `Flags`
_Initialization_: `NewFlags()`, `NewInitializedFlags()`

**Current Convention** as used in this PR.
_package name_: `internal/pkg/txflags`
_variable name_: `ValidationFlags`
_Initialization_: `New()`, `NewWithValues()`

**If you agree with the proposed convention, I will make the modification.**

mastersingh24 (approved):

PR #928, opened 2020-03-28
Remove BCCSP Utilities

lindluni:
Removes a bunch of "utilities" in the BCCSP package and pushes the very simplistic code into the calling functions


bccsp/sw/fileks.go
```diff
@@ -113,35 +117,35 @@ func (ks *fileBasedKeyStore) GetKey(ski []byte) (bccsp.Key, error) {
...
+			return nil, fmt.Errorf("failed loading key [%x] [%s]", ski, err)
 		}
 
 		return &aesPrivateKey{key, false}, nil
 	case "sk":
 		// Load the private key
 		key, err := ks.loadPrivateKey(hex.EncodeToString(ski))
 		if err != nil {
-			return nil, fmt.Errorf("Failed loading secret key [%x] [%s]", ski, err)
+			return nil, fmt.Errorf("ailed loading secret key [%x] [%s]", ski, err)
```
cendhu:
```suggestion
			return nil, fmt.Errorf("failed loading secret key [%x] [%s]", ski, err)
```

bccsp/sw/fileks.go
```diff
@@ -355,27 +359,51 @@ func (ks *fileBasedKeyStore) loadKey(alias string) ([]byte, error) {
 func (ks *fileBasedKeyStore) createKeyStoreIfNotExists() error {
 	// Check keystore directory
 	ksPath := ks.path
-	missing, err := utils.DirMissingOrEmpty(ksPath)
 
-	if missing {
-		logger.Debugf("KeyStore path [%s] missing [%t]: [%s]", ksPath, missing, utils.ErrToString(err))
+	// Check if keystore directory exists
+	_, err := os.Stat(ksPath)
+	if os.IsNotExist(err) {
```
cendhu:
```suggestion
	if err != nil && !os.IsNotExist(err) {
	    return err
	}
```
It helps to reduce the extra indentation.

bccsp/sw/fileks.go
```diff
@@ -355,27 +359,51 @@ func (ks *fileBasedKeyStore) loadKey(alias string) ([]byte, error) {
...
 		if err != nil {
 			logger.Errorf("Failed creating KeyStore At [%s]: [%s]", ksPath, err.Error())
 			return nil
 		}
+	} else if err != nil {
+		return err
 	}
 
-	return nil
+	// Check if keystore directory is empty
```
cendhu: This is called **side effect** and **disinformation** as I specified in the coding guidelines here https://docs.google.com/document/d/1mU4S4U1dDpBk9AA180RMRB7g3OH887nqbcEPGD3px3Q/edit?usp=sharing 

- The name of the function is `CreateStoreIfNotExist()`
- What the function actually does? 
           1. creates the store if not exists
           2. creates the store if empty
- The function claims to do one thing but it does two things -- side effect and disinformation naming.

I understand that you didn't name it. Given that the function size is grown with additional details, I feel that the readability got impacted significantly.

I find the existing function more readable than this one. IMO, the existing code somewhat followed the formatting rule I specified in the above document but it used unjustifiable util. The existing function 
- checks whether the directory already exists or empty 
- if it so, it creates one 
It just took a few seconds to read the existing function.

 If possible, I would suggest fixing these issues.

bccsp/sw/keyimport.go
```diff
@@ -33,7 +33,9 @@ func (*aes256ImportKeyOptsKeyImporter) KeyImport(raw interface{}, opts bccsp.Key
 		return nil, fmt.Errorf("Invalid Key Length [%d]. Must be 32 bytes", len(aesRaw))
 	}
 
-	return &aesPrivateKey{utils.Clone(aesRaw), false}, nil
```
cendhu, line 36: Just curious. Why do we need to clone the byte slice? Do we expect the `raw` to get changed?

bccsp/sw/keyimport.go
```diff
@@ -48,7 +50,10 @@ func (*hmacImportKeyOptsKeyImporter) KeyImport(raw interface{}, opts bccsp.KeyIm
 		return nil, errors.New("Invalid raw material. It must not be nil.")
 	}
 
-	return &aesPrivateKey{utils.Clone(aesRaw), false}, nil
+	clone := make([]byte, len(aesRaw))
```
cendhu: The same question applies here too.

bccsp/sw/fileks.go
```diff
@@ -355,27 +359,51 @@ func (ks *fileBasedKeyStore) loadKey(alias string) ([]byte, error) {
...
+	f, err := os.Open(ksPath)
+	if err != nil && err != io.EOF {
+		return err
+	}
+	defer f.Close()
+
+	_, err = f.Readdir(1)
+	if err == io.EOF {
```
cendhu:
```suggestion
	if err != nil && err != io.EOF {
	    return err
	 }
```
This ensures consistency within the function (similar to the one at line 379).

cendhu (changes requested):

bccsp/sw/keyimport.go
mastersingh24: I don't think we need to "clone" here either

mastersingh24 (commented):
We need to do more of this type of thing

bccsp/sw/fileks.go
lindluni: I ended up pulling the `dirEmpty` and `dirExists` functions up instead, and pushing the `CreateStoreIfNotExist` code up to the caller. I could have renamed the function `CreateStoreIfEmptyOrNotExist` but I think it was excessive and the code is clear what it is doing in the `Init` function.

bccsp/sw/fileks.go
```diff
@@ -65,32 +66,51 @@ func (ks *fileBasedKeyStore) Init(pwd []byte, path string, readOnly bool) error
...
+	copy(ks.pwd, pwd)
+	ks.pwd = clone
+	ks.readOnly = readOnly
+
+	// Check if the keystore directory exists
+	exists, err := dirExists(path)
 	if err != nil {
 		return err
 	}
+	// If keystore directory does not exist, create the keystore
```
cendhu: Nit. You have refactored the code in a nice way and It is very clear from the code that it checks whether the dirExists and if not, create the store. IMO, this comment and the one at line 86 does not help. While reading the code, it creates an unnecessary pause and a breakage. Please refer to this PR for discussion on comments -- [895](https://github.com/hyperledger/fabric/pull/895). The same suggestion is applicable to dirEmpty related code below. 

Sorry If you think I am doing too much nitpicking. It is just that we can reduce 4 lines of code and avoid breakage during the code reading. If you think this will improve the code readability, please follow this rule in other future PRs too. IMO, these simple things also improve the code readability. 

Throughout the file, you can notice comments like (exists already, not added by you)
```
// Load the key
key, err := ks.loadKey(hex.EncodeToString(ski))
```
```
// Load the private key
key, err := ks.loadPrivateKey(hex.EncodeToString(ski))
```
```
// Load the public key
key, err := ks.loadPublicKey(hex.EncodeToString(ski))
```
If the golang speaks greek langauge instead of English, I think it is okay to have this English translation :-)

cendhu (commented):
Thank you, Brett, for the changes. A minor suggestion on comments.

bccsp/sw/fileks.go
lindluni: Agreed and removed.

And while it is nitpicking, we have to start somewhere. The more we educate everyone on best practices, and standardizing on those practices, hopefully the more good practices will spread throughout the codebase. 

Thanks Senthil!

bccsp/sw/fileks.go
```diff
@@ -219,7 +232,7 @@ func (ks *fileBasedKeyStore) searchKeystoreForSKI(ski []byte) (k bccsp.Key, err
 
 		return k, nil
 	}
-	return nil, fmt.Errorf("Key with SKI %s not found in %s", hex.EncodeToString(ski), ks.path)
+	return nil, fmt.Errorf("key with SKI %s not found in %s", hex.EncodeToString(ski), ks.path)
 }
```
sykesm, line 239: Just use %x to format as hex instead of using the hex package.

bccsp/sw/fileks_test.go
```diff
@@ -30,7 +30,7 @@ func TestInvalidStoreKey(t *testing.T) {
 
 	ks, err := NewFileBasedKeyStore(nil, filepath.Join(tempDir, "bccspks"), false)
 	if err != nil {
-		fmt.Printf("Failed initiliazing KeyStore [%s]", err)
+		fmt.Printf("Failed initiliazing KeyStore [%s]\n", err)
```
sykesm, line 33: Why the new line? Why the odd formatting of the error? Why printf and exit instead of failing the test with an assertion or `Fatalf`?

bccsp/sw/fileks_test.go
```diff
@@ -89,7 +89,7 @@ func TestBigKeyFile(t *testing.T) {
 
 	_, err = ks.GetKey(ski)
 	assert.Error(t, err)
-	expected := fmt.Sprintf("Key with SKI %s not found in %s", hex.EncodeToString(ski), ksPath)
+	expected := fmt.Sprintf("key with SKI %s not found in %s", hex.EncodeToString(ski), ksPath)
```
sykesm, line 92: %x can be used instead of the hex package

sykesm (approved):

bccsp/sw/fileks.go
lindluni: addressed in #936

bccsp/sw/fileks_test.go
lindluni: addressed in #936

bccsp/sw/fileks_test.go
lindluni: addressed in #936

PR #933, opened 2020-03-28
Move keys.go Utility to SW package

lindluni:
Moves the keys.go utility to the SW package where the consumers are and kills off production code that existed entirely for driving tests.

### DIFF
--- a/bccsp/pkcs11/impl_test.go
+++ b/bccsp/pkcs11/impl_test.go
@@ -743,7 +743,7 @@ func TestECDSAKeyImportFromECDSAPublicKey(t *testing.T) {
 		t.Fatalf("Failed getting ECDSA raw public key [%s]", err)
 	}
 
-	pub, err := utils.DERToPublicKey(pkRaw)
+	pub, err := x509.ParsePKIXPublicKey(pkRaw)
 	if err != nil {
 		t.Fatalf("Failed converting raw to ecdsa.PublicKey [%s]", err)
 	}
@@ -862,7 +862,7 @@ func TestKeyImportFromX509ECDSAPublicKey(t *testing.T) {
 		t.Fatalf("Failed getting ECDSA raw public key [%s]", err)
 	}
 
-	pub, err := utils.DERToPublicKey(pkRaw)
+	pub, err := x509.ParsePKIXPublicKey(pkRaw)
 	if err != nil {
 		t.Fatalf("Failed converting raw to ECDSA.PublicKey [%s]", err)
 	}
--- a/bccsp/signer/signer.go
+++ b/bccsp/signer/signer.go
@@ -8,10 +8,10 @@ package signer
 
 import (
 	"crypto"
+	"crypto/x509"
 	"io"
 
 	"github.com/hyperledger/fabric/bccsp"
-	"github.com/hyperledger/fabric/bccsp/utils"
 	"github.com/pkg/errors"
 )
 
@@ -47,7 +47,7 @@ func New(csp bccsp.BCCSP, key bccsp.Key) (crypto.Signer, error) {
 		return nil, errors.Wrap(err, "failed marshalling public key")
 	}
 
-	pk, err := utils.DERToPublicKey(raw)
+	pk, err := x509.ParsePKIXPublicKey(raw)
 	if err != nil {
 		return nil, errors.Wrap(err, "failed marshalling der to public key")
 	}
--- a/bccsp/sw/keys.go
+++ b/bccsp/sw/keys.go
@@ -4,7 +4,7 @@ Copyright IBM Corp. All Rights Reserved.
 SPDX-License-Identifier: Apache-2.0
 */
 
-package utils
+package sw
 
 import (
 	"crypto/ecdsa"
@@ -17,7 +17,6 @@ import (
 	"fmt"
 )
 
-// struct to hold info required for PKCS#8
 type pkcs8Info struct {
 	Version             int
 	PrivateKeyAlgorithm []asn1.ObjectIdentifier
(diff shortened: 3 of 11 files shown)


bccsp/pkcs11/impl_test.go
```diff
@@ -743,7 +743,10 @@ func TestECDSAKeyImportFromECDSAPublicKey(t *testing.T) {
 		t.Fatalf("Failed getting ECDSA raw public key [%s]", err)
 	}
 
-	pub, err := utils.DERToPublicKey(pkRaw)
+	if len(pkRaw) == 0 {
+		t.Fatalf("Invalid DER. It must be different from nil")
+	}
```
sykesm: I don't understand why this validation is needed. ParsePKIXPublicKey should return an error. A test will let us know if it's needed or not.

bccsp/pkcs11/impl_test.go
```diff
@@ -862,7 +865,10 @@ func TestKeyImportFromX509ECDSAPublicKey(t *testing.T) {
...
```
sykesm: Same here.

bccsp/signer/signer.go
```diff
@@ -47,7 +47,10 @@ func New(csp bccsp.BCCSP, key bccsp.Key) (crypto.Signer, error) {
 		return nil, errors.Wrap(err, "failed marshalling public key")
 	}
 
-	pk, err := utils.DERToPublicKey(raw)
+	if len(raw) == 0 {
+		return nil, errors.New("invalid DER. It must be different from nil")
+	}
+	pk, err := x509.ParsePKIXPublicKey(raw)
```
sykesm: And here.

bccsp/sw/keys.go
```diff
@@ -299,46 +288,22 @@ func PublicKeyToPEM(publicKey interface{}, pwd []byte) ([]byte, error) {
...
-
-// PublicKeyToEncryptedPEM converts a public key to encrypted pem
-func PublicKeyToEncryptedPEM(publicKey interface{}, pwd []byte) ([]byte, error) {
-	if publicKey == nil {
-		return nil, errors.New("Invalid public key. It must be different from nil.")
+		return nil, errors.New("invalid public key. It must be different from nil")
 	}
 	if len(pwd) == 0 {
-		return nil, errors.New("Invalid password. It must be different from nil.")
+		return nil, errors.New("invalid password. It must be different from nil")
```
sykesm: The function isn't exported and the length was checked by the caller to see if this should even be used. I don't know why it's here. I'd also expect an error if the password was "wrong".

bccsp/sw/keys.go
```diff
@@ -299,46 +288,22 @@ func PublicKeyToPEM(publicKey interface{}, pwd []byte) ([]byte, error) {
...
 	default:
-		return nil, errors.New("Invalid key type. It must be *ecdsa.PublicKey")
+		return nil, errors.New("invalid key type. It must be *ecdsa.PublicKey")
 	}
 }
 
-// PublicKeyToDER marshals a public key to the der format
-func PublicKeyToDER(publicKey interface{}) ([]byte, error) {
+func publicKeyToEncryptedPEM(publicKey interface{}, pwd []byte) ([]byte, error) {
 	if publicKey == nil {
```
sykesm: This is a nice check but falls into the default case below; if you want a specific case, put it there.

sykesm (commented):
Most of the checking that's done in here is necessary and should be removed with assertions in proper test coverage.

bccsp/sw/keys.go
```diff
@@ -54,30 +53,27 @@ func oidFromNamedCurve(curve elliptic.Curve) (asn1.ObjectIdentifier, bool) {
...
-func PrivateKeyToPEM(privateKey interface{}, pwd []byte) ([]byte, error) {
+func privateKeyToPEM(privateKey interface{}, pwd []byte) ([]byte, error) {
 	// Validate inputs
 	if len(pwd) != 0 {
-		return PrivateKeyToEncryptedPEM(privateKey, pwd)
+		return privateKeyToEncryptedPEM(privateKey, pwd)
 	}
 	if privateKey == nil {
-		return nil, errors.New("Invalid key. It must be different from nil.")
+		return nil, errors.New("invalid key. It must be different from nil")
```
sykesm: This block can be in in the switch. The type will be `nil`.

bccsp/sw/keys.go
```diff
@@ -110,7 +106,7 @@ func PrivateKeyToPEM(privateKey interface{}, pwd []byte) ([]byte, error) {
 
 		pkcs8Bytes, err := asn1.Marshal(pkcs8Key)
 		if err != nil {
-			return nil, fmt.Errorf("error marshaling EC key to asn1 [%s]", err)
```
sykesm: This covers the nil key.

bccsp/sw/keys.go
```diff
@@ -120,20 +116,19 @@ func PrivateKeyToPEM(privateKey interface{}, pwd []byte) ([]byte, error) {
...
+		return nil, errors.New("invalid key type. It must be *ecdsa.PrivateKey")
 	}
 }
 
-// PrivateKeyToEncryptedPEM converts a private key to an encrypted PEM
-func PrivateKeyToEncryptedPEM(privateKey interface{}, pwd []byte) ([]byte, error) {
+func privateKeyToEncryptedPEM(privateKey interface{}, pwd []byte) ([]byte, error) {
 	if privateKey == nil {
-		return nil, errors.New("Invalid private key. It must be different from nil.")
+		return nil, errors.New("invalid private key. It must be different from nil")
```
sykesm: this can be handled in the switch and should be covered by the default case below.

lindluni:
I will address the comments after I figure out why `goimports` is reporting bad imports. I can't reproduce it locally

bccsp/sw/keys.go
```diff
@@ -171,90 +165,86 @@ func DERToPrivateKey(der []byte) (key interface{}, err error) {
...
-func PEMtoPrivateKey(raw []byte, pwd []byte) (interface{}, error) {
+func pemToPrivateKey(raw []byte, pwd []byte) (interface{}, error) {
 	if len(raw) == 0 {
-		return nil, errors.New("Invalid PEM. It must be different from nil.")
+		return nil, errors.New("invalid PEM. It must be different from nil")
 	}
 	block, _ := pem.Decode(raw)
 	if block == nil {
-		return nil, fmt.Errorf("Failed decoding PEM. Block must be different from nil. [% x]", raw)
+		return nil, fmt.Errorf("failed decoding PEM. Block must be different from nil [% x]", raw)
```
sykesm, line 182: unrelated: message can be much better

bccsp/sw/keys.go
```diff
@@ -171,90 +165,86 @@ func DERToPrivateKey(der []byte) (key interface{}, err error) {
...
-	return nil, errors.New("Invalid key type. The DER must contain an ecdsa.PrivateKey")
+	return nil, errors.New("invalid key type. The DER must contain an ecdsa.PrivateKey")
 }
 
-// PEMtoPrivateKey unmarshals a pem to private key
-func PEMtoPrivateKey(raw []byte, pwd []byte) (interface{}, error) {
+func pemToPrivateKey(raw []byte, pwd []byte) (interface{}, error) {
 	if len(raw) == 0 {
-		return nil, errors.New("Invalid PEM. It must be different from nil.")
+		return nil, errors.New("invalid PEM. It must be different from nil")
```
sykesm: zero length and nil are not the same so the message is misleading. Also, a 0 length pem block will fail decoding making this unnecessary.

lindluni:
This is my file: https://github.com/btl5037/fabric/blob/keys/bccsp/sw/fileks.go

But the remote it is pulling has this:

image

lindluni:
Going to come back to this on the weekend

PR #940, opened 2020-03-29
Update doc for Go v1.14.1

denyeart:
#### Type of change

- Documentation update

#### Description

https://github.com/hyperledger/fabric/pull/934 updated Go to
v1.14.1. This change updates docs accordingly.

### DIFF
--- a/docs/source/commands/peerversion.md
+++ b/docs/source/commands/peerversion.md
@@ -8,7 +8,7 @@ information. For example:
  peer:
    Version: 2.1.0
    Commit SHA: b78d79b
-   Go version: go1.13.8
+   Go version: go1.14.1
    OS/Arch: linux/amd64
    Chaincode:
     Base Docker Label: org.hyperledger.fabric
--- a/docs/source/prereqs.rst
+++ b/docs/source/prereqs.rst
@@ -83,7 +83,7 @@ Go Programming Language
 Hyperledger Fabric uses the Go Programming Language for many of its
 components.
 
-  - `Go <https://golang.org/dl/>`__ version 1.13.x is required.
+  - `Go <https://golang.org/dl/>`__ version 1.14.x is required.
 
 Given that we will be writing chaincode programs in Go, there are two
 environment variables you will need to set properly; you can make these
--- a/docs/wrappers/peer_version_preamble.md
+++ b/docs/wrappers/peer_version_preamble.md
@@ -8,7 +8,7 @@ information. For example:
  peer:
    Version: 2.1.0
    Commit SHA: b78d79b
-   Go version: go1.13.8
+   Go version: go1.14.1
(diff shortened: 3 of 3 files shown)


docs/source/prereqs.rst
```diff
@@ -83,7 +83,7 @@ Go Programming Language
 Hyperledger Fabric uses the Go Programming Language for many of its
 components.
 
-  - `Go <https://golang.org/dl/>`__ version 1.13.x is required.
+  - `Go <https://golang.org/dl/>`__ version 1.14.x is required.
```
sykesm, line 86: This is untrue for app development and the reason why I did not make the update. This document seems to be all over the map with respect to prerequisites so it's unclear the context it's being used in.

For fabric development, yes, we want go 1.14 now. For apps, it doesn't matter.

This document, however, also talks about docker compose, node.js, npm, and python - things we don't use in Fabric. If the goal is for this to document prerequisites for development, it needs a lot of cleanup. If it's intended for chaincode or other client development, it needs to be improved.

In summary, if this is dev doc, let's yank all of the stuff that's irrelevant and update the go version. If it's generic, let's highlight the context under which we're making the prereq.

docs/wrappers/peer_version_preamble.md
```diff
@@ -8,7 +8,7 @@ information. For example:
  peer:
    Version: 2.1.0
    Commit SHA: b78d79b
-   Go version: go1.13.8
+   Go version: go1.14.1
```
sykesm, line 11: This is an example; it doesn't matter what it is.

docs/source/commands/peerversion.md
```diff
@@ -8,7 +8,7 @@ information. For example:
  peer:
    Version: 2.1.0
    Commit SHA: b78d79b
-   Go version: go1.13.8
+   Go version: go1.14.1
```
sykesm, line 11: This is an example; it doesn't matter what it is.

sykesm (approved):

docs/source/prereqs.rst
denyeart: Agree that the prereq doc overall is a mess and conflates many things. I've opened https://jira.hyperledger.org/browse/FAB-17701 to clean it up.
For now let's just go with the precedent of mentioning the Go version that Fabric uses in the associated release. Although this is not required for chaincode apps, I don't think we could go wrong with having chaincode developers use the same Go version as Fabric itself.
And since Matt already approved, I assume he's good with this for now.

docs/wrappers/peer_version_preamble.md
denyeart: Agreed it doesn't technically matter, but having it accurate relative to the Version field above can only reduce user confusion.

denyeart:

PR #945, opened 2020-03-30
mv cache package to statecouchdb

cendhu:
#### Type of change

#### Description

We have introduced the cache implementation in the [statedb package](https://github.com/hyperledger/fabric/tree/master/core/ledger/kvledger/txmgmt/statedb) as the initial plan was to integrate cache with both [statecouchdb](https://github.com/hyperledger/fabric/tree/master/core/ledger/kvledger/txmgmt/statedb/statecouchdb) and [stateleveldb](https://github.com/hyperledger/fabric/tree/master/core/ledger/kvledger/txmgmt/statedb/stateleveldb). However, we have decided to defer the cache integration with stateleveldb as the expected throughput improvement would be < 5%. This is because the goleveldb is an embedded DB (i.e., no cost associated with the access unlike CouchDB) and it utilizes OS page cache anyway to avoid the disk access. Further, the goleveldb is not a bottleneck as of now.

Hence, we move the cache implementation to the statecouchdb package itself. When there is a need to integrate cache with goleveldb, we can move it back.

#### Additional details

> Related concepts should be in the same package; Especially when it is used by only one package. 

as per one of the format rule specified in the [coding guidelines ](https://docs.google.com/document/d/1mU4S4U1dDpBk9AA180RMRB7g3OH887nqbcEPGD3px3Q/edit#)

#### Related issues

[FAB-17676](https://jira.hyperledger.org/browse/FAB-17676)


core/ledger/kvledger/txmgmt/statedb/statecouchdb/cache_value.proto
```diff
@@ -6,7 +6,7 @@ SPDX-License-Identifier: Apache-2.0
 
 syntax = "proto3";
 
-option go_package = "github.com/hyperledger/fabric/core/ledger/kvledger/txmgmt/statedb";
+option go_package = "github.com/hyperledger/fabric/core/ledger/kvledger/txmgmt/statecouchdb";
 
 package statedb;
```
mastersingh24: I don't think it truly matters, but should this be `statecouchdb` as well?

mastersingh24 (dismissed):
Makes sense to me.  One comment on the package for the CacheValue proto def

core/ledger/kvledger/txmgmt/statedb/statecouchdb/statecouchdb_test_export.go
```diff
@@ -74,7 +79,7 @@ func (env *testVDBEnv) stopExternalResource() {
 
 func (env *testVDBEnv) closeAndReopen() {
 	env.DBProvider.Close()
-	dbProvider, _ := NewVersionedDBProvider(env.config, &disabled.Provider{}, &statedb.Cache{})
+	dbProvider, _ := NewVersionedDBProvider(env.config, &disabled.Provider{}, env.sysNamespaces)
```
manish-sethi, line 82: In "env" you had obtained a pointer to cache. You would need that to point to new cache object here. Perhaps, no caching based test is calling this function yet and hence escaped.

manish-sethi (changes requested):

core/ledger/kvledger/txmgmt/statedb/statecouchdb/cache_value.proto
cendhu: Thanks for catching this. It was never intended. I missed out this change.

core/ledger/kvledger/txmgmt/statedb/statecouchdb/statecouchdb_test_export.go
cendhu: I don't understand. Earlier, the `privacyenabledstate` was creating the cache. As we have moved to `statecouchdb` package, now we create a `cache` in`NewVersionedDBProvider()`. When we `closeAndReopen()`, we need to create a new cache object which is anyway done by this line. If I am missing something, I am not sure what you meant here.

manish-sethi: Look at the `env.init()` function. You have this line `env.cache = dbProvider.cache`. 
After you invoke this function (`closeAndReopen`) in some test, `env.cache` would still be pointing to older cache object, not the new one that will be created.

core/ledger/kvledger/txmgmt/statedb/statecouchdb/cache_value.proto
cendhu: Done.

core/ledger/kvledger/txmgmt/statedb/statecouchdb/statecouchdb_test_export.go
cendhu: Ah :-) Probably, we should have named `closeAndReopen()` as `reInit()`. I am saying that because it would have made me look at `init()` once after changing this code. 

Done.

manish-sethi (approved):

PR #947, opened 2020-03-30
Move comm pkg to internal

mastersingh24:
#### Type of change

#### Description
comm is not currently intended to be an external library.


C0rWin (dismissed):

core/comm/alias.go
```diff
@@ -0,0 +1,109 @@
...
+type CredentialSupport = pkgcomm.CredentialSupport
+type DynamicClientCredentials = pkgcomm.DynamicClientCredentials
+type GRPCClient = pkgcomm.GRPCClient
+type GRPCServer = pkgcomm.GRPCServer
+type KeepaliveOptions = pkgcomm.KeepaliveOptions
+type Metrics = pkgcomm.Metrics
+type SecureOptions = pkgcomm.SecureOptions
+type ServerConfig = pkgcomm.ServerConfig
+type ServerStatsHandler = pkgcomm.ServerStatsHandler
+type TLSOption = pkgcomm.TLSOption
```
sykesm: I assume there will be follow-ups to remove these? If not, the move doesn't really do much.

cendhu:
@mastersingh24 @sykesm  A doubt but not related to this PR. If we have some util in the ledger and it is used only by the `core/ledger` (but different packages within the ledger), is it okay to move them to `internals/pkg/`. Is there a common guideline on what should go into `internals/pkg`? I understand that txflags used across different modules such as ledger, validation, etc... go here but I still wonder what to do with certain ledger utils.

core/comm/alias.go
mastersingh24: > I assume there will be follow-ups to remove these? If not, the move doesn't really do much.

Yep.  Just did not want to change all the consumers at once.

mastersingh24:
> @mastersingh24 @sykesm A doubt but not related to this PR. If we have some util in the ledger and it is used only by the `core/ledger` (but different packages within the ledger), is it okay to move them to `internals/pkg/`. Is there a common guideline on what should go into `internals/pkg`? I understand that txflags used across different modules such as ledger, validation, etc... go here but I still wonder what to do with certain ledger utils.

`internal/pkg` is intended to be used by internal packages which are intended to be used multiple different packages.

`internal` might be a better fit or perhaps an `internal` under `core/ledger` as well.

cendhu:
Thanks, Gari. I will keep that in mind.

mastersingh24:
ugh ... I hope these updates are not making the flakes worse

core/comm/alias.go
sykesm: > Yep. Just did not want to change all the consumers at once.

Then my suggestion is to hold this out until the next piece is ready. Additional commits should be added until the type aliases are removed.

sykesm (changes requested):
I'd prefer to see this deferred until the callers are updated. Type aliases are a good tool for compatibility but moving the code while preserving all of the aliases isn't really moving the ball forward.

If there's strong disagreement, I'll step aside but I do think we need to bring this to completion or not start it at all.

mastersingh24:
> I'd prefer to see this deferred until the callers are updated. Type aliases are a good tool for compatibility but moving the code while preserving all of the aliases isn't really moving the ball forward.
> 
> If there's strong disagreement, I'll step aside but I do think we need to bring this to completion or not start it at all.

done ... I originally thought there might be more conflicts so was going to update one caller at a time.  But here it is in one shot

sykesm (approved):
Thank you.

PR #951, opened 2020-03-30
[FAB-17109] Retrieve ReConnectBackoffThreshold as duration

stephyee:
#### Type of change

- Bug fix

#### Description

Parses ReconnectBackoffThreshold from config as duration then converts to float64 

#### Related issues

[FAB-17109](https://jira.hyperledger.org/browse/FAB-17109)

### DIFF
--- a/core/deliverservice/config.go
+++ b/core/deliverservice/config.go
@@ -20,7 +20,7 @@ import (
 )
 
 const (
-	DefaultReConnectBackoffThreshold   = float64(time.Hour)
+	DefaultReConnectBackoffThreshold   = time.Hour * 1
 	DefaultReConnectTotalTimeThreshold = time.Second * 60 * 60
 	DefaultConnectionTimeout           = time.Second * 3
 )
@@ -30,7 +30,7 @@ type DeliverServiceConfig struct {
 	// PeerTLSEnabled enables/disables Peer TLS.
 	PeerTLSEnabled bool
 	// ReConnectBackoffThreshold sets the delivery service maximal delay between consencutive retries.
-	ReConnectBackoffThreshold float64
+	ReConnectBackoffThreshold time.Duration
 	// ReconnectTotalTimeThreshold sets the total time the delivery service may spend in reconnection attempts
 	// until its retry logic gives up and returns an error.
 	ReconnectTotalTimeThreshold time.Duration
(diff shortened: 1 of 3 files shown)


yacovm (approved):

yacovm:
why is it a draft? what is left to do?

stephyee:
@yacovm nothing is left, I was only working on the fix for 1.4.3 alongside this.

core/deliverservice/config.go
```diff
@@ -97,7 +97,9 @@ func LoadOverridesMap() (map[string]*orderers.Endpoint, error) {
 func (c *DeliverServiceConfig) loadDeliverServiceConfig() {
 	c.PeerTLSEnabled = viper.GetBool("peer.tls.enabled")
 
-	c.ReConnectBackoffThreshold = viper.GetFloat64("peer.deliveryclient.reConnectBackoffThreshold")
+	reconnectBackoffThresholdDuration := viper.GetDuration("peer.deliveryclient.reConnectBackoffThreshold")
+
+	c.ReConnectBackoffThreshold = reconnectBackoffThresholdDuration.Seconds()
```
yacovm: I'm not sure we need the Seconds() here. This converts it to multiples of seconds, no? 

The code after this block checks if this is zero, and if it is - it assigns DefaultReConnectBackoffThreshold which isn't a multiple of seconds, but a multiple of nano seconds. 

Aren't the units mixed up?

Then, it assigns it to Max**RetryDelay:     time.Duration(d.conf.DeliverServiceConfig.ReConnectBackoffThreshold)** in **deliveryclient.go** so it is clearly a time.Duration, so we should just change the **ReConnectBackoffThreshold** field in **DeliverServiceConfig** also to time.Duration in my opinion, instead of leaving it as a float64

yacovm (changes requested):

core/deliverservice/config.go
stephyee: Yes, `Seconds()` was used to convert duration to a float. This does mix up the units for the provided value vs default, do you have any concerns if we change to always process the value as `time.Duration`?

yacovm: **time.Duration** is preferable. (No concerns)

stephyee: 🛠

yacovm (approved):

PR #955, opened 2020-03-31
mv encoding util to consumer (pvtdatastorage)

cendhu:
#### Type of change

#### Description

When [missing pvtData reconciler](https://github.com/hyperledger/fabric/blob/master/gossip/privdata/reconcile.go) asks for missing data info, it is necessary to pass the missing pvtdata info associated with the most recent block so that missing pvtdata in the state db can be fixed sooner to reduce the `pvtdata does not match with hashed data` error during endorserments. In order to give priority to missing pvtData in the most recent block, we use [reverse order-preserving encoding/decoding](https://github.com/hyperledger/fabric/blob/master/core/ledger/util/uint64_encoding.go) while constructing the missing data key. This simplifies the implementation of `GetMissingPvtDataInfoForMostRecentBlocks()`.

Currently, this reverse order encoding/decoding is present in the `core/ledger/util`. Given that it is used only in [pvtdatastorage package](https://github.com/hyperledger/fabric/tree/master/core/ledger/pvtdatastorage) and we need to keep such code close to the consumer, we move the endcoding/decoding util to pvtdatastorage package itself.

#### Related issues

[FAB-17697](https://jira.hyperledger.org/browse/FAB-17697)


mastersingh24 (approved):
LGTM - thanks

PR #956, opened 2020-03-31
[FAB-17681] Support removing a consortium

wuxuer:
Remove a consortium from Consortiums group in a channel config.

#### Type of change

- New feature

#### Related issues
https://jira.hyperledger.org/browse/FAB-17681

### DIFF
--- a/pkg/config/consortiums.go
+++ b/pkg/config/consortiums.go
@@ -21,6 +21,19 @@ type Consortium struct {
 	Organizations []Organization
 }
 
+// RemoveConsortium remove a consortium from a channel configuration
+func (c *ConfigTx) RemoveConsortium(consortiumName string) error {
+	consortiumGroup := c.updated.ChannelGroup.Groups[ConsortiumsGroupKey].Groups
+	_, ok := consortiumGroup[consortiumName]
+	if !ok {
+		return fmt.Errorf("consortium %s does not exist in channel config", consortiumName)
+	}
+
+	delete(consortiumGroup, consortiumName)
+
+	return nil
+}
+
 // AddOrgToConsortium adds an org definition to a named consortium in a given
 // channel configuration.
 func (c *ConfigTx) AddOrgToConsortium(org Organization, consortium string) error {
--- a/pkg/config/consortiums_test.go
+++ b/pkg/config/consortiums_test.go
@@ -601,6 +601,65 @@ func TestAddOrgToConsortiumFailures(t *testing.T) {
 	}
 }
 
+func TestRemoveConsortium(t *testing.T) {
+	t.Parallel()
+
+	gt := NewGomegaWithT(t)
+
+	consortiums := baseConsortiums(t)
+	consortiumsGroup, err := newConsortiumsGroup(consortiums)
+	gt.Expect(err).NotTo(HaveOccurred())
+
+	config := &cb.Config{
+		ChannelGroup: &cb.ConfigGroup{
+			Groups: map[string]*cb.ConfigGroup{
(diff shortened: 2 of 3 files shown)


pkg/config/consortiums.go
```diff
@@ -21,6 +21,19 @@ type Consortium struct {
 	Organizations []Organization
 }
 
+// RemoveConsortium remove a consortium from a channel configuration
+func (c *ConfigTx) RemoveConsortium(consortiumName string) error {
+	consortiumGroup := c.updated.ChannelGroup.Groups[ConsortiumsGroupKey].Groups
+	_, ok := consortiumGroup[consortiumName]
+	if !ok {
```
cendhu, line 28: Wondering whether it should return an error or be a no-op when no such `consortiumName` found. Given that it is a library, I think I can compare this to the data structure library in golang. In the map data structure, if we pass a non-existence key, it does not return an error. Please refer here https://github.com/golang/go/blob/master/src/builtin/builtin.go#L143 If an error is needed in `RemoveConsortium` case, I would like to understand the reasoning behind such a design.  

I also wondered why `delete()` does not return error. On a google search, I found this interesting discussion -- https://www.reddit.com/r/golang/comments/5tfx7i/why_delete_doesnt_return_a_bool/ 

> Perhaps because adding a boolean could be confusing and easy to misinterpret as "failed to delete" vs. "delete successful".

>  The caller usually either knows that the item is present, or doesn't care; either way, why bother reporting back?

wuxuer: If RemoveConsortium doesn't return error with a non-exist consortiumName, one calls this function and ComputeUpdate in turn to comupte the delta. This will end with error "no differences detected between original and updated config".
Therefore, it seems to be a better choice to return an error to remind the caller not to call other functions when the consortiumName does not exist.

cendhu: Then, my question is also applicable to `ComputeUpdate()`. Why does it return an `error` instead of returning an empty delta? Are you saying that if I implement an API to compare two files to create a diff, the API would return an error instead of returning empty diff when no difference is detected?

I think this library puts so many restrictions on the consumer.  I am not comfortable with such usage. Anyway, if maintainers have no issue with this library, I will step aside.

caod123: It seems like if we don't error and return an empty delta, the user may accidentally submit a config update transaction to the ordering service with no actual update in it. This may lead to confusion when attempting to debug why something they had intended to update was never changed when in fact it had errored earlier. For the RemoveConsortium example, imagine the user attempts to remove a consortium that exists but accidentally introduces a typo in the name that they do not catch. If they aren't quick to notice it, they may end up processing and computing the update and then wonder why the consortium they had originally intended on removing was never removed after submitting the config update. You could argue that the client of this library should check for empty deltas and handle error processing there, but what about in the case where you are processing multiple updates on the config before computing the update. Then your typo is lost and it is uncertain which parts of the config update had initially failed prior to generating the update envelope.

cendhu: The following looks like a natural flow to me. 

```
if isConsortiumExist(consortium) {
    removeConsortium(consortium)
    delta = computeUpdate()
    if delta != nil {
        //submit the transaction
    }
}
```

Even in golang, I can write a buggy program. I don't think libraries are created in such a way to avoid programming bugs/error. Currently, it looks more like a state-machine where the consumer cannot move from one state to another without passing certain criteria. 

Anyway, these are two different philosophies and quite hard to say which is good or bad. As you are leading this effort, I trust your judgment and leave it here.

caod123 (approved):

PR #959, opened 2020-03-31
[FAB-17109] Retrieve ReconnectBackoffThreshold as duration

stephyee:
#### Type of change
- Bug fix

#### Description

Retrieves and uses ReconnectBackoffThreshold from config as duration

#### Related issues

[FAB-17109](https://jira.hyperledger.org/browse/FAB-17109)

### DIFF
--- a/core/deliverservice/deliveryclient.go
+++ b/core/deliverservice/deliveryclient.go
@@ -29,7 +29,7 @@ var logger = flogging.MustGetLogger("deliveryClient")
 const (
 	defaultReConnectTotalTimeThreshold = time.Second * 60 * 60
 	defaultConnectionTimeout           = time.Second * 3
-	defaultReConnectBackoffThreshold   = float64(time.Hour)
+	defaultReConnectBackoffThreshold   = time.Hour
 )
 
 func getReConnectTotalTimeThreshold() time.Duration {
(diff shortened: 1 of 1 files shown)


core/deliverservice/deliveryclient.go
```diff
@@ -29,7 +29,7 @@ var logger = flogging.MustGetLogger("deliveryClient")
 const (
 	defaultReConnectTotalTimeThreshold = time.Second * 60 * 60
 	defaultConnectionTimeout           = time.Second * 3
-	defaultReConnectBackoffThreshold   = float64(time.Hour)
+	defaultReConnectBackoffThreshold   = time.Hour * 1
```
caod123: I believe you dont need the `*1`

core/deliverservice/deliveryclient.go
```diff
@@ -279,7 +279,7 @@ func (d *deliverServiceImpl) newClient(chainID string, ledgerInfoProvider blocks
 		}
 		sleepIncrement := float64(time.Millisecond * 500)
 		attempt := float64(attemptNum)
-		return time.Duration(math.Min(math.Pow(2, attempt)*sleepIncrement, reconnectBackoffThreshold)), true
+		return time.Duration(math.Min(math.Pow(2, attempt)*sleepIncrement, reconnectBackoffThreshold.Hours())), true
```
caod123: I don't think you want to return `Hours()` here since your sleep increment is just casting the duration to float64. The threshold should match the base unit used for the increment. Probably just casting to a float64 is enough.

caod123 (changes requested):

core/deliverservice/deliveryclient.go
stephyee: agreed

mastersingh24 (approved):

PR #962, opened 2020-03-31
[FAB-17660] Retrieve Channel configuration

DereckLuo:

#### Type of change
- New feature

#### Description
* Added retrieve channel configuration from config.
* Added retrieve consortiums from config.
* Moved channelID out of Channel struct.

#### Related issues
[FAB-177660](https://jira.hyperledger.org/browse/FAB-17660)

### DIFF
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -32,7 +32,6 @@ type Channel struct {
 	Consortiums  []Consortium
 	Capabilities []string
 	Policies     map[string]Policy
-	ChannelID    string
 }
 
 // Policy is an expression used to define rules for access to channels, chaincodes, etc.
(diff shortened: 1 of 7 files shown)


pkg/config/consortiums.go
```diff
@@ -49,6 +49,42 @@ func (c *ConfigTx) AddOrgToConsortium(org Organization, consortium string) error
 	return nil
 }
 
+// GetConsortiums returns a list of consortiums for the channel configuration.
+// Consortiums is only defined for the ordering system channel.
+func (c *ConfigTx) GetConsortiums() ([]Consortium, error) {
```
stephyee: Going forward we don't need the `GET` prefix for these exported ConfigTX retrieval functions

MHBauer: Sounds like there's an interesting story there. Were prefixes required previously?

stephyee: Interesting? Possibly but we have a [FAB-17662](https://jira.hyperledger.org/browse/FAB-17662) as a housekeeping item.

DereckLuo: done. will address rest of `GET` prefix in FAB-17662

pkg/config/config.go
```diff
@@ -95,13 +94,96 @@ func (c *ConfigTx) Updated() *cb.Config {
...
+	return updt, nil
+}
+
+// ChannelConfiguration returns a channel configuration value from a config transaction.
+func (c *ConfigTx) ChannelConfiguration() (Channel, error) {
+	channelGroup := c.base.ChannelGroup
+	var err error
+
+	_, ok := channelGroup.Values[ConsortiumKey]
+	var consortium string
```
caod123: I would consider moving all of these var declarations to the top into a `var(...)` block

pkg/config/config.go
```diff
@@ -95,13 +94,96 @@ func (c *ConfigTx) Updated() *cb.Config {
...
+	if ok {
+		consortiumProto := &cb.Consortium{}
+		err := unmarshalConfigValueAtKey(channelGroup, ConsortiumKey, consortiumProto)
+		if err != nil {
+			return Channel{}, err
+		}
+		consortium = consortiumProto.Name
+	}
+
+	_, ok = channelGroup.Groups[ApplicationGroupKey]
```
caod123: You can just inline the if check since you're not assigning the value for future use:
```
if _, ok := channelGroup.Groups[ApplicationGroupKey]; ok {...}
```

pkg/config/config.go
```diff
@@ -95,13 +94,96 @@ func (c *ConfigTx) Updated() *cb.Config {
...
+	var application Application
+	if ok {
+		application, err = c.GetApplicationConfiguration()
```
caod123, line 135: You may want to consider if we even need the if ok check above for these. Considering you're not erroring out here if the ApplicationGroup does not exist but just moving on with an empty Application, should we instead just call GetApplicationConfiguration and defer to its error if the group doesn't exist?

pkg/config/config.go
```diff
@@ -95,13 +94,96 @@ func (c *ConfigTx) Updated() *cb.Config {
...
+			return Channel{}, err
+		}
+	}
+
+	policies, err := c.GetPoliciesForChannel()
+	if err != nil {
+		return Channel{}, err
+	}
+
+	//TODO: channelID cannot be retrieved from proto.
```
caod123: can probably remove this TODO since you're removing `channelID` from the `Channel` type

caod123 (changes requested):
Some minor comments about the retrieval function

pkg/config/config.go
DereckLuo: The intend here is to support retrieve both `application channel` and `system channel`, therefore, it did not error out in the first place, but just returned empty Application and move on. We can definitely split this up into two functions, which we can target specific channel types. Any recommendations?

pkg/config/config.go
DereckLuo: done.

pkg/config/config.go
DereckLuo: done.

pkg/config/config.go
DereckLuo: done.

pkg/config/config.go
caod123: Makes sense, hadn't considered system channel yet.

caod123 (approved):

PR #954, opened 2020-03-31
go modules for fabric

MHBauer:
wip

#### Type of change

- Improvement - Infrastructure/Build

- [ ] ~`check_deps.sh` checking implemented with modules base~ don't care
- [ ] [Tools are installed into $(go env GOPATH)/bin. That's problematic in cases where $GOPATH is really a path or when $GOPATH is not on the $PATH.](https://github.com/hyperledger/fabric/pull/954#issuecomment-606870940)
- [ ] more?

#### Description

Would like to use some of fabric as a dependency for golang plugins.

#### Additional details

pkcs11 unit-tests always fail for me. On master and this branch.
idemix integration-tests always fail for me. On master and this
branch.

Completed work:
- internal dependency on gopath https://github.com/hyperledger/fabric/pull/970
- more gopath https://github.com/hyperledger/fabric/pull/1077
- tools using modules https://github.com/hyperledger/fabric/pull/1057
- gossip race detector fails https://github.com/hyperledger/fabric/pull/1069/files
- code using modules #1064
- document changes to developer experience #1064

#### Related issues

https://jira.hyperledger.org/browse/FAB-10562

A `tools.go` based dependency management.
https://github.com/golang/go/wiki/Modules#how-can-i-track-tool-dependencies-for-a-module

<!-- Checklist (DELETE AFTER READING):

- `Signed-off-by` added to commits (required for DCO check to pass)
- Tests have been added/updated (required for bug fixes and features)
- Unit and/or integration tests pass locally Run linters and checks
- locally using 'make checks' If change requires documentation
- updates, make updates in pull request, or open a separate issue and
- provide link Squash commits into a single commit, unless a stack of
- commits is intentional to assist reviewers or to preserve review
- comments.  For additional contribution guidelines see the project's
- CONTRIBUTING.md file ->

PR #964, opened 2020-03-31
don't wait to expire dead endpoints if exiting

MHBauer:
#### Type of change

- Bug fix

#### Description

deadEndpoints channel gets filled up. Exit occurs, and deadEndpoints is not drained. We can avoid putting more into the channel and creating hanging goroutines (many many many of them, this fills up my memory). There are still messages that get dropped on the floor sometimes that will cause a different hang.

#### Additional details

Seeing what looked like unbounded memory usage when running this in a loop.
There are anywhere between 80 and 600 messages for each side of the connection.
Each iteration through the loop results in a timer being created.
Avoid that and create one at the test start.

#### Related issues

FAB-17451 I reported this from https://github.com/hyperledger/fabric/pull/954#issuecomment-606809081

### DIFF
--- a/gossip/comm/comm_impl.go
+++ b/gossip/comm/comm_impl.go
@@ -602,10 +602,12 @@ func (c *commImpl) Ping(context.Context, *proto.Empty) (*proto.Empty, error) {
 }
 
 func (c *commImpl) disconnect(pkiID common.PKIidType) {
-	if c.isStopping() {
+	select {
+	case c.deadEndpoints <- pkiID:
+	case <-c.exitChan:
 		return
 	}
-	c.deadEndpoints <- pkiID
+
 	c.connStore.closeConnByPKIid(pkiID)
 }
 


MHBauer:
@denyeart maybe #3 from your bug. The build log had expired.

gossip/comm/comm_impl.go
```diff
@@ -602,10 +602,11 @@ func (c *commImpl) Ping(context.Context, *proto.Empty) (*proto.Empty, error) {
 }
 
 func (c *commImpl) disconnect(pkiID common.PKIidType) {
-	if c.isStopping() {
-		return
+	select {
+	case c.deadEndpoints <- pkiID:
+	case <-c.exitChan:
```
yacovm, line 607: maybe return here?

MHBauer: sure.

MHBauer:
That would be the other flake I mean.

/ci-run

MHBauer:
flake is https://jira.hyperledger.org/browse/FAB-16233

/ci-run

yacovm (changes requested):
Please keep this pull request focused and not have in it commits that are unrelated

MHBauer:
moved to separate PR #983. I would suggest merging that one first if this keeps flakin.

yacovm (approved):

PR #966, opened 2020-03-31
rm unnecessary import dependency from statecouchdb package 

cendhu:
#### Type of change

#### Description

The [statecouchdb package](https://github.com/hyperledger/fabric/tree/master/core/ledger/kvledger/txmgmt/statedb/statecouchdb) should not be dependent on too many other packages. It should be self-sufficient. Currently, we have imported [ccprovider](https://github.com/hyperledger/fabric/tree/master/core/common/ccprovider) just for a struct. Further, the test file imports a ccprovider to call a utility function. As these are not justifiable, we remove it and duplicate code as needed.

#### Additional details

In the ledger refactoring work, we plan to merge the following packages to reduce the folder depth
- merge kvledger with the ledger package
- merge lockbasedtxmgr to txmgr
- many more 

Due to some unnecessary import dependency, we get import cycles in the refactoring process. Hence, we remove one such unnecessary dependency from the statecouchdb package. In the upcoming PR, we will move structs that are used across packages from `core/ledger` to `internal/pkg/ledger`. This will also help us reduce the import cycles and do the refactoring without any blockers. 

#### Related issues

[FAB-17699](https://jira.hyperledger.org/browse/FAB-17699)

### DIFF
--- a/core/ledger/kvledger/txmgmt/statedb/statecouchdb/statecouchdb.go
+++ b/core/ledger/kvledger/txmgmt/statedb/statecouchdb/statecouchdb.go
@@ -16,7 +16,6 @@ import (
 	"github.com/hyperledger/fabric/common/flogging"
 	"github.com/hyperledger/fabric/common/ledger/dataformat"
 	"github.com/hyperledger/fabric/common/metrics"
-	"github.com/hyperledger/fabric/core/common/ccprovider"
 	"github.com/hyperledger/fabric/core/ledger/internal/version"
 	"github.com/hyperledger/fabric/core/ledger/kvledger/txmgmt/statedb"
 	"github.com/hyperledger/fabric/core/ledger/util/couchdb"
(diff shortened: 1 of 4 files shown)


core/ledger/kvledger/txmgmt/statedb/statecouchdb/statecouchdb.go
```diff
@@ -16,7 +16,6 @@ import (
 	"github.com/hyperledger/fabric/common/flogging"
 	"github.com/hyperledger/fabric/common/ledger/dataformat"
 	"github.com/hyperledger/fabric/common/metrics"
-	"github.com/hyperledger/fabric/core/common/ccprovider"
```
mastersingh24, line 19: Very nice ... we need to do more of this!

cendhu: Thanks, @mastersingh24 If you keep on encouraging me like this, many more will come. Anyway, we do have some time limit to complete these tasks before starting the checkpoint pruning. Will do as much as @manish-sethi and I can do to improve the code before that.

core/ledger/kvledger/txmgmt/statedb/statecouchdb/statecouchdb_test.go
```diff
@@ -817,10 +820,10 @@ func TestPaginatedQuery(t *testing.T) {
 		t.Fatalf("Couchdb state impl is expected to implement interface `statedb.IndexCapable`")
 	}
 
-	fileEntries, errExtract := ccprovider.ExtractFileEntries(dbArtifactsTarBytes, "couchdb")
+	dbIndexArtifacts, errExtract := extractFileEntries(dbArtifactsTarBytes)
```
manish-sethi: Now since this package expects simply a map, there is no need to create the tar file and then untar (and hence no need to copy the function `extractFileEntries` from ccprovider below). You can simply use the map at line 802.

manish-sethi (changes requested):

core/ledger/kvledger/txmgmt/statedb/statecouchdb/statecouchdb_test.go
cendhu: Valid point. Done.

manish-sethi (dismissed):

cendhu:
@manish-sethi fixed conflicts.

manish-sethi (approved):

PR #967, opened 2020-03-31
[FAB-17540] Fix race TestUpdateRootsFromConfigBlock

stephyee:
#### Type of change

- Bug fix

#### Description

Lock read/write access to tlsconfig to prevent data race

#### Related issues

[FAB-17540](https://jira.hyperledger.org/browse/FAB-17540)


yacovm:
I am confused.

The stack trace in the JIRA shows a race in the production code, where Fabric does:
```
WARNING: DATA RACE
Write at 0x00c001242690 by goroutine 216:
 github.com/hyperledger/fabric/core/comm.(*GRPCServer).SetClientRootCAs()
      /home/sykesm/workspace/fabric/src/github.com/hyperledger/fabric/core/comm/server.go:293 +0x315
  github.com/hyperledger/fabric/core/peer.(*Peer).updateTrustedRoots()
      /home/sykesm/workspace/fabric/src/github.com/hyperledger/fabric/core/peer/peer.go:113 +0x5c4
```
And the TLS handshake does:

```
Previous read at 0x00c001242690 by goroutine 128:
  crypto/tls.(*Conn).processCertsFromClient()
      /usr/local/go/src/crypto/tls/handshake_server.go:730 +0x405
  crypto/tls.(*serverHandshakeState).doFullHandshake()
      /usr/local/go/src/crypto/tls/handshake_server.go:514 +0xa23
  crypto/tls.(*serverHandshakeState).handshake()
      /usr/local/go/src/crypto/tls/handshake_server.go:105 +0x68c
  crypto/tls.(*Conn).serverHandshake()
      /usr/local/go/src/crypto/tls/handshake_server.go:60 +0x1cd
```


But you are changing only the test code.

Don't we potentially have a problem in the production code? 

The way I see it, the Fabric code changes: **gServer.tlsConfig.ClientCAs** and the TLS handshake reads these TLS root CAs. 

Perhaps we should consider implementing a [TransportCredentials](https://godoc.org/google.golang.org/grpc/credentials#TransportCredentials) which will atomically load an immutable TLSConfig ?

cendhu:
@yacovm -- the data race expert :-)

stephyee:
@yacovm Is it preferred to add a mutex to the serverConfig? I used the [clone()](https://godoc.org/crypto/tls#Config.Clone) which uses a mutex.

internal/pkg/comm/creds.go
```diff
@@ -56,7 +56,10 @@ func (sc *serverCreds) ClientHandshake(context.Context,
 
 // ServerHandshake does the authentication handshake for servers.
 func (sc *serverCreds) ServerHandshake(rawConn net.Conn) (net.Conn, credentials.AuthInfo, error) {
-	conn := tls.Server(rawConn, sc.serverConfig)
+	// safe clone that can be used concurrently for reads and writes
+	serverConfig := sc.serverConfig.Clone()
```
yacovm, line 60: This clone is not thread safe.

yacovm:
> @yacovm Is it preferred to add a mutex to the serverConfig? I used the [clone()](https://godoc.org/crypto/tls#Config.Clone) which uses a mutex.

```
func (c *Config) Clone() *Config {
	// Running serverInit ensures that it's safe to read
	// SessionTicketsDisabled.
	c.serverInitOnce.Do(func() { c.serverInit(nil) })
	var sessionTicketKeys []ticketKey
	c.mutex.RLock()
	sessionTicketKeys = c.sessionTicketKeys
	c.mutex.RUnlock()
```
It uses a mutex, but: 

1. It only synchronizes around the session ticket keys which is irrelevant to the client root CAs.
2. The **SetClientRootCAs** method which writes to the memory, doesn't lock the same mutex, so there is no happens-before relation enforced by the runtime.

You need to make all writes and reads to the client roots to be thread safe and memory coherent.

There are 2 paths for you to take:

1. Wrap every access that writes to the tlsConfig field of the GRPCServer struct with a write lock, and wrap every read access to that struct with a read lock. This is less code but is more error prone in the long run.
2. Remove the tlsConfig field, and replace it with something that enforces the user to access it in a memory coherent manner. For instance, you can wrap the **tls.Config** with a struct with a method that returns a copy of the inner **tls.Config** (the copying is taken under a read lock) and a method that mutates the **tls.Config** under a write lock.

yacovm (changes requested):
> @yacovm Is it preferred to add a mutex to the serverConfig? I used the [clone()](https://godoc.org/crypto/tls#Config.Clone) which uses a mutex.

```
func (c *Config) Clone() *Config {
	// Running serverInit ensures that it's safe to read
	// SessionTicketsDisabled.
	c.serverInitOnce.Do(func() { c.serverInit(nil) })
	var sessionTicketKeys []ticketKey
	c.mutex.RLock()
	sessionTicketKeys = c.sessionTicketKeys
	c.mutex.RUnlock()
```
It uses a mutex, but: 

1. It only synchronizes around the session ticket keys which is irrelevant to the client root CAs.
2. The **SetClientRootCAs** method which writes to the memory, doesn't lock the same mutex, so there is no [happens-before](https://go101.org/article/memory-model.html) relation enforced by the runtime.

It is true that the documentation says:
```
 It is safe to clone a Config that is being used concurrently by a TLS client or server. 
```
However this is misleading in our case, because the documentation didn't take into account that the Fabric code, changes the config as well. 

You need to make all writes and reads to the client roots to be thread safe and memory coherent.

There are 2 paths for you to take:

1. Wrap every access that writes to the tlsConfig field of the GRPCServer struct with a write lock, and wrap every read access to that struct with a read lock. This is less code but is more error prone in the long run.
2. Remove the tlsConfig field, and replace it with something that enforces the user to access it in a memory coherent manner. For instance, you can wrap the **tls.Config** with a struct with a method that returns a copy of the inner **tls.Config** (the copying is taken under a read lock) and a method that mutates the **tls.Config** under a write lock. An example can be seen in the [PredicateDialer](https://github.com/hyperledger/fabric/blob/master/orderer/common/cluster/util.go#L113-L123) of the cluster infrastructure of the orderer.

stephyee:
@yacovm thanks I'll consider both options you presented

internal/pkg/comm/creds.go
```diff
@@ -56,6 +58,9 @@ func (sc *serverCreds) ClientHandshake(context.Context,
 
 // ServerHandshake does the authentication handshake for servers.
 func (sc *serverCreds) ServerHandshake(rawConn net.Conn) (net.Conn, credentials.AuthInfo, error) {
+	sc.lock.Lock()
```
yacovm: TLS handshakes involve network I/O so we don't want to hold locks for a long time. Even if there is nothing competing this lock, it is usually considered bad practice to hold locks while doing network I/O.

I suggest that the TLS config will be cloned under a read lock and nothing more.

internal/pkg/comm/creds.go
```diff
@@ -28,26 +29,50 @@ var (
...
 // serverCreds is an implementation of grpc/credentials.TransportCredentials.
 type serverCreds struct {
-	serverConfig *tls.Config
+	serverConfig *TLSConfig
 	logger       *flogging.FabricLogger
 }
 
+type TLSConfig struct {
+	config *tls.Config
+	lock   *sync.Mutex
```
yacovm: no need to have a pointer mutex. a non pointer is just fine, and saves you initialization.

internal/pkg/comm/creds.go
```diff
@@ -28,26 +29,50 @@ var (
...
+}
+
+func NewTLSConfig(config *tls.Config, lock *sync.Mutex) *TLSConfig {
```
yacovm, line 58: this lock belongs to the TLSConfig, so why do you have a constructor that requires to create it outside and pass it?

internal/pkg/comm/server.go
```diff
@@ -73,11 +73,15 @@ func NewGRPCServerFromListener(listener net.Listener, serverConfig ServerConfig)
 	if secureConfig.UseTLS {
 		//both key and cert are required
 		if secureConfig.Key != nil && secureConfig.Certificate != nil {
+			grpcServer.tls = &TLSConfig{
```
yacovm: you have a constructor method, yet you create the instance without it.

Let's be consistent.

internal/pkg/comm/server.go
```diff
@@ -245,6 +252,8 @@ func (gServer *GRPCServer) AppendClientRootCAs(clientRoots [][]byte) error {
 
 // internal function to add a PEM-encoded clientRootCA
 func (gServer *GRPCServer) appendClientRootCA(clientRoot []byte) error {
+	gServer.tls.lock.Lock()
```
yacovm: this works, but is very brittle as no one promises you that the TLS config object is not going to be modified further in the code execution without grabbing the lock beforehand.

internal/pkg/comm/creds.go
```diff
@@ -28,26 +29,50 @@ var (
...
+	return &TLSConfig{
+		config: config,
+		lock:   lock,
+	}
+}
+
+func (sc *TLSConfig) Config() *tls.Config {
```
yacovm: I suggest that the TLSConfig struct would govern the lifecycle of its inner fields and not delegate it to the user of the struct:

1. Let's return a copy upon Config(). Note, that a shallow copy is trivially obtained via changing the return type to be a non pointer type.
2. Let's add a mutator method that just updates whatever we want, under the lock.

This removes the need to external locking completely and simplifies the usage of this struct.

yacovm (changes requested):
The comment above + I think it's nice to add a unit test.

internal/pkg/comm/creds.go
stephyee: I'm not sure about this, w/o the pointer this triggers `copylocks` check failure with govet.

yacovm: But with your latest version of the code, you don't reference the lock externally anymore.

internal/pkg/comm/creds.go
```diff
@@ -28,26 +30,64 @@ var (
...
 // serverCreds is an implementation of grpc/credentials.TransportCredentials.
 type serverCreds struct {
-	serverConfig *tls.Config
+	serverConfig *TLSConfig
 	logger       *flogging.FabricLogger
 }
 
+type TLSConfig struct {
+	config *tls.Config
+	lock   *sync.Mutex
```
yacovm: Let's use a RWMutex so it will be non blocking for TLS hanshakes.

internal/pkg/comm/creds.go
```diff
@@ -28,26 +30,64 @@ var (
...
+
+func NewTLSConfig(config *tls.Config, lock *sync.Mutex) *TLSConfig {
+	return &TLSConfig{
+		config: config,
+		lock:   lock,
+	}
+}
+
+func (t *TLSConfig) Config() tls.Config {
+	t.lock.Lock()
```
yacovm: Let's use a read lock here instead so you can do several TLS handshakes in parallel.

internal/pkg/comm/creds.go
```diff
@@ -28,26 +30,64 @@ var (
...
+	defer t.lock.Unlock()
+
+	if t.config != nil {
+		return *t.config.Clone()
+	}
+
+	return tls.Config{}
+}
+
+func (t *TLSConfig) AddCert(cert *x509.Certificate) {
```
yacovm: AddClientRootCA gives the method more meaning that AddCert

internal/pkg/comm/server.go
```diff
@@ -234,6 +237,7 @@ func (gServer *GRPCServer) Stop() {
 func (gServer *GRPCServer) AppendClientRootCAs(clientRoots [][]byte) error {
```
yacovm: This method (**AppendClientRootCAs**) is actually not used in production code. 
Let's get rid of it shall we?

internal/pkg/comm/creds.go
```diff
@@ -28,26 +30,64 @@ var (
...
 // grpc/credentials.TransportCredentials
 func NewServerTransportCredentials(
-	serverConfig *tls.Config,
+	serverConfig *TLSConfig,
 	logger *flogging.FabricLogger) credentials.TransportCredentials {
-
 	// NOTE: unlike the default grpc/credentials implementation, we do not
 	// clone the tls.Config which allows us to update it dynamically
-	serverConfig.NextProtos = alpnProtoStr
+	serverConfig.lock.Lock()
```
yacovm: i might be missing something, but - why do we need to lock here?

internal/pkg/comm/creds.go
```diff
@@ -28,26 +30,64 @@ var (
...
+	serverConfig *TLSConfig
 	logger       *flogging.FabricLogger
 }
 
+type TLSConfig struct {
+	config *tls.Config
+	lock   *sync.Mutex
+}
+
+func NewTLSConfig(config *tls.Config, lock *sync.Mutex) *TLSConfig {
```
yacovm: I think that now you don't need to inject the lock anymore, because I don't think you need to lock in **NewServerTransportCredentials**.

yacovm (changes requested):
Some more comments

internal/pkg/comm/creds.go
stephyee: I'm not sure what you mean by this statement, the `AddRootCA()` and `SetClientCAs()` still reference the TLConfig lock.

yacovm: Right, but these are all methods of the struct, so they reference it internally. 
By externally i mean that the lock is referenced by some other function which is not an instance method of the struct TLSConfig

internal/pkg/comm/creds.go
```diff
@@ -28,26 +30,61 @@ var (
...
+	defer t.lock.RUnlock()
+
+	if t.config != nil {
+		return *t.config.Clone()
+	}
+
+	return tls.Config{}
+}
+
+func (t *TLSConfig) AddRootCA(cert *x509.Certificate) {
```
yacovm: **AddClientRootCA** ? :-)

yacovm (approved):
Thanks for the patience and compliance to the code review comments.

yacovm:
I think this should be backported since it effects production behavior of TLS handshakes.

PR #969, opened 2020-03-31
Fix transaction parameterin in docs/test_network.md

obermeier:
The current command does not update the ledger because CAR9 does not exists. The real name is CAR009
If you flow the original doc this will lead to the error:
```Error: endorsement failure during invoke. response: status:500 message:"error in simulation: transaction returned with failure: Car CAR9 does not exist"```

#### Type of change

- Documentation update

#### Description

Go and Java implementations use a three-digit number for the key. [Example](https://github.com/hyperledger/fabric-samples/blob/4bb48a909cf7109f3915562fc08d1cbb4cb14dde/chaincode/fabcar/java/src/main/java/org/hyperledger/fabric/samples/fabcar/FabCar.java#L98)

Signed-off-by: Stefan Obermeier <scray@stefan-obermeier.de>

### DIFF
--- a/docs/source/test_network.md
+++ b/docs/source/test_network.md
@@ -319,7 +319,7 @@ Chaincodes are invoked when a network member wants to transfer or change an
 asset on the ledger. Use the following command to change the owner of a car on
 the ledger by invoking the fabcar chaincode:
 ```
-peer chaincode invoke -o localhost:7050 --ordererTLSHostnameOverride orderer.example.com --tls true --cafile ${PWD}/organizations/ordererOrganizations/example.com/orderers/orderer.example.com/msp/tlscacerts/tlsca.example.com-cert.pem -C mychannel -n fabcar --peerAddresses localhost:7051 --tlsRootCertFiles ${PWD}/organizations/peerOrganizations/org1.example.com/peers/peer0.org1.example.com/tls/ca.crt --peerAddresses localhost:9051 --tlsRootCertFiles ${PWD}/organizations/peerOrganizations/org2.example.com/peers/peer0.org2.example.com/tls/ca.crt -c '{"function":"changeCarOwner","Args":["CAR9","Dave"]}'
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: docs/source/test_network.md [hunk 1]
```diff
@@ -319,7 +319,7 @@ Chaincodes are invoked when a network member wants to transfer or change an
...
+peer chaincode invoke -o localhost:7050 --ordererTLSHostnameOverride orderer.example.com --tls true --cafile ${PWD}/organizations/ordererOrganizations/example.com/orderers/orderer.example.com/msp/tlscacerts/tlsca.example.com-cert.pem -C mychannel -n fabcar --peerAddresses localhost:7051 --tlsRootCertFiles ${PWD}/organizations/peerOrganizations/org1.example.com/peers/peer0.org1.example.com/tls/ca.crt --peerAddresses localhost:9051 --tlsRootCertFiles ${PWD}/organizations/peerOrganizations/org2.example.com/peers/peer0.org2.example.com/tls/ca.crt -c '{"function":"changeCarOwner","Args":["CAR009","Dave"]}'
```
mastersingh24, line 322: I think this may only be the case in Java ... maybe we should fix the Java fabcar sample instead?

sykesm: The ~~update~~ _original_ matches the output from line 315 so I believe (hope?) it's correct:

```
{"Key":"CAR9", "Record":{"make":"Holden","model":"Barina","colour":"brown","owner":"Shotaro"}}
```

Given the pointer to code, however, it looks like things don't match.  Something is wrong. :)

nikhil550: I ran the network and confirmed, the java chaincode uses a different index than golang and javascript.
```
 nikhilgupta$ peer chaincode query -C channel2 -n fabcar -c '{"Args":["queryCar","CAR009"]}'
{"owner":"Shotaro","model":"Barina","color":"brown","make":"Holden"}
```
For golang
```
peer chaincode query -C channel2 -n fabcar -c '{"Args":["queryCar","CAR9"]}'
{"owner":"Shotaro","make":"Holden","model":"Barina"}
```

obermeier: > I think this may only be the case in Java ... maybe we should fix the Java fabcar sample instead?

You are right! I tested it with the go code the wrong way. Sorry!

nikhil550: Ill put a small note in the doc for java users. I can remove it if we update the chaincode.

PR #970, opened 2020-04-01
Prepare for possible move to go modules

sykesm:
These changes isolate GOPATH dependencies and enable a possible move to go modules. These changes are focused on two areas:

**Discovery of the sampleconfig resources during test**

If the build is run in _module_ mode, the `sampleconfig` directory is calculated as relative to the module root of the test. If the build is run in _GOPATH_ mode, the directory is calculated as relative to the `GOPATH`

**Use "simple" chaincode instead of a module in most cases**

There were a number of places in the integration tests where we would attempt to build a chaincode module from an import path reference. This works in _GOPATH_ mode but fails in module mode. These changes replace the use of the `module` chaincode with `simple` chaincode where appropriate as `simple` chaincode could be built within the context of the module hosting the integration tests.


core/config/configtest/config.go
```diff
@@ -38,15 +43,55 @@ func dirExists(path string) bool {
...
 func GetDevConfigDir() string {
-	gopath := os.Getenv("GOPATH")
+	path, err := gomodDevConfigDir()
+	if err != nil {
+		path, err = gopathDevConfigDir()
+		if err != nil {
+			panic(err)
+		}
+	}
+	return path
```
MHBauer, line 53: The barest of nits. This reads better straight line with no indentation. I don't think the single exit point helps in a function this short. I think it's clearer in terms of, use gomod first, use gopath second.

```
	path, err := gomodDevConfigDir()
	if err == nil {
		return path
	}
	path, err = gopathDevConfigDir()
	if err == nil {
		return path
	}
	panic(err)
```
It can even be collapsed with the assignable-if
```
	var err error
	if path, err := gomodDevConfigDir(); err == nil {
		return path
	}
	if path, err := gopathDevConfigDir(); err == nil {
		return path
	}
	panic(err)

File: core/config/configtest/config.go [hunk 2]
```diff
@@ -38,15 +43,55 @@ func dirExists(path string) bool {
...
 		}
 	}
 
-	panic("unable to find sampleconfig directory on gopath")
+	return "", fmt.Errorf("unable to find sampleconfig directory on GOPATH")
+}
+
+func gomodDevConfigDir() (string, error) {
+	buf := bytes.NewBuffer(nil)
+	cmd := exec.Command("go", "env", "GOMOD")
```
MHBauer, line 77: This makes sense to me and seems more flexible.

integration/externalbuilders/golang/bin/build
```diff
@@ -27,7 +27,7 @@ if [ -f "$SOURCE/src/go.mod" ]; then
     cd "$SOURCE/src"
     go build -v -mod=readonly -o "$OUTPUT/chaincode" "$GO_PACKAGE_PATH"
 else
-    GO111MODULE=off go build -v  -o "$OUTPUT/chaincode" "$GO_PACKAGE_PATH"
+    GOPATH="$SOURCE" GO111MODULE=off go build -v  -o "$OUTPUT/chaincode" "$GO_PACKAGE_PATH"
```
MHBauer, line 30: So before this gopath was completely unset, and was the default set depending on go version, and the fact that it built was because everything was already checked out in the gopath. I think this makes sense.

MHBauer (approved):
I think the gomod/gopath determination for finding the file is elegant and correct for this situation.

I can't speak to the simplecc change. Glancing at the directory in the location, I don't know how to figure out what it is.The directory could use a Readme or a comment to explain it's existence in the location.

Agree as to being a separable change in preparation. :+1:

core/config/configtest/config.go
```diff
@@ -38,15 +43,55 @@ func dirExists(path string) bool {
 // maintained with the source tree. This should only be used in a
 // test/development context.
 func GetDevConfigDir() string {
```
cendhu, line 45: I like the structure -- highly readable. 

Just one minor suggestion: 

[In the formatting rules section](https://docs.google.com/document/d/1mU4S4U1dDpBk9AA180RMRB7g3OH887nqbcEPGD3px3Q/edit#), I have specified the following:

1. Closely related concepts should not be kept far within a file; Should never be kept in a different file.
2. If one function calls another, they should be vertically close. The caller should be placed above the callee.
3. Certain bits of code which has a conceptual affinity towards other code, they should be kept together. Affinity might be caused because a group of functions performs similar operations.
4. The most important concept first and the low-level details in the bottom.

While the current code structure satisfies these rules somewhat, it would be nice to have `gomodDevConfigDir()` immediately after the `GetDevConfigDir()` followed by `gopathDevConfigDir()`. The structure can look similar to how various function calls are made in the top-level function that is `GetDevConfigDir()`

When I read the top-level function, I searched for `gomodDevConfigDir` and had to browse the code a bit more to find out where it is located. 

As these points are highly debatable, I will leave the decision to you.

core/config/configtest/config.go
cendhu: I thought about what you suggested above. But, we need three `return path` statement and especially the last `return path` which comes after `panic(err)` is **a dead code**. IMO, the current structure looks good and no changes required.

cendhu: @MHBauer If the `panic(err)` can be avoided, what you suggested is applicable and makes the code more readable.

integration/chaincode/simple/cmd/main.go
```diff
@@ -8,16 +8,69 @@ package main
...
+
+	err := ioutil.WriteFile(termFile, []byte("term-file"), 0644)
 	if err != nil {
-		fmt.Fprintf(os.Stderr, "Exiting Simple chaincode: %s", err)
-		os.Exit(2)
+		fmt.Fprintf(os.Stderr, "failed to write term file to %s: %s", termFile, err)
+	}
+}
+
+func main() {
```
cendhu, line 50: This is the top-level function in this file. IMO, this should be at the beginning of this file. The reason is that, first, I read `handleSignals`, then I read `handleSigTerm` followed by `main()`. Finally, I had to again read `handleSignals` and `handleSigTerm` by keeping `main()` in the context. Probably, I need to improve my code reading skills. May be the rearrangement might help me.

core/config/configtest/config.go
mastersingh24: I'd say feel free to create a change on top of this if you like @cendhu

mastersingh24 (approved):
I like getting started down the path to Go modules

mastersingh24 (approved):

core/config/configtest/config.go
cendhu: Yeah, this is simple to do. Will push a PR.

PR #983, opened 2020-04-03
[FAB-17451] off-by-one error in TestCommSend

MHBauer:
Stop processing received messages as soon as possible.
Avoid leaking tickers momentarily.
Spelling correction: recieve -> receive.

Change-Id: Ieac2341c5e97c59971dee366a03b00a9b57830b9
Signed-off-by: Morgan Bauer <mbauer@us.ibm.com>

#### Type of change

- Bug fix (gossip flakes) TestCommSend

#### Description

The implementation changes with the reads happening separately vs only pulling one or the other at a time.

https://jira.hyperledger.org/browse/FAB-17451 maybe case 3, but also encountered in other PRs.

### DIFF
--- a/gossip/comm/comm_test.go
+++ b/gossip/comm/comm_test.go
@@ -624,35 +624,37 @@ func TestCommSend(t *testing.T) {
 		}
 	}()
 
-	c1recieved := 0
-	c2recieved := 0
+	c1received := 0
+	c2received := 0
 	// hopefully in some runs we'll fill both send and receive buffers and
 	// drop overflowing messages, but still finish, because the endless
 	// stream of messages inexorably gets through unless something is very
 	// broken.
-	totalMessagesRecieved := (DefSendBuffSize + DefRecvBuffSize) * 2
+	totalMessagesReceived := (DefSendBuffSize + DefRecvBuffSize) * 2
+	ticker := time.NewTicker(30 * time.Second)
+	defer ticker.Stop()
 RECV:
 	for {
 		select {
 		case <-ch1:
-			c1recieved++
-			if c1recieved == totalMessagesRecieved {
+			c1received++
+			if c1received == totalMessagesReceived {
 				close(stopch2)
 			}
 		case <-ch2:
-			c2recieved++
-			if c2recieved == totalMessagesRecieved {
+			c2received++
+			if c2received == totalMessagesReceived {
 				close(stopch1)
 			}
-		case <-time.After(30 * time.Second):
-			t.Fatalf("timed out waiting for messages to be recieved.\nc1 got %d messages\nc2 got %d messages", c1recieved, c2recieved)
+		case <-ticker.C:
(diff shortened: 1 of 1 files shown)


gossip/comm/comm_test.go
```diff
@@ -624,35 +624,56 @@ func TestCommSend(t *testing.T) {
...
+		defer close(c)
+		wg.Wait()
+		close(stopch1)
+		close(stopch2)
+	}()
+
+	ticker := time.NewTicker(30 * time.Second)
+	select {
+	case <-c:
+	case <-ticker.C:
```
yacovm: why not use **time.After** ? This is done one time, so who cares if the ticker is still scheduled to run?

gossip/comm/comm_test.go
```diff
@@ -624,35 +624,56 @@ func TestCommSend(t *testing.T) {
...
-			if c2recieved == totalMessagesRecieved {
-				close(stopch1)
+	totalMessagesReceived := (DefSendBuffSize + DefRecvBuffSize) * 2
+
+	wg := sync.WaitGroup{}
+	wg.Add(2)
+	// read from each until there are an acceptable number of
+	// messages and then stop the sender from sending more
+	// messages.
+	go func() {
```
yacovm: 1) Why not close stopCh1 in a defer when this go func exits?
2) Can't we create a function that spawns this go func and then just invoke it 2 times for ch1 and ch2 ? (to reduce code duplication)

gossip/comm/comm_test.go
```diff
@@ -624,35 +624,56 @@ func TestCommSend(t *testing.T) {
...
+				return
 			}
 		}
+	}()
+
+	c := make(chan struct{})
+	go func() {
+		defer close(c)
+		wg.Wait()
+		close(stopch1)
```
yacovm: stopCh1 and stopCh2 can be closed earlier.

yacovm (changes requested):
a few comments

MHBauer (commented):
whoops

gossip/comm/comm_test.go
```diff
@@ -624,35 +624,37 @@ func TestCommSend(t *testing.T) {
...
-	c2recieved := 0
+	c1received := 0
+	c2received := 0
 	// hopefully in some runs we'll fill both send and receive buffers and
 	// drop overflowing messages, but still finish, because the endless
 	// stream of messages inexorably gets through unless something is very
 	// broken.
-	totalMessagesRecieved := (DefSendBuffSize + DefRecvBuffSize) * 2
+	totalMessagesReceived := (DefSendBuffSize + DefRecvBuffSize) * 2
+	ticker := time.NewTicker(3 * time.Second)
```
yacovm: isn't 3 seconds a bit too fast for our CI?

MHBauer: ah, oops, leftover from local test

yacovm (approved):

PR #985, opened 2020-04-03
mv version package to core/ledger/internal/pkg

cendhu:
#### Type of change

#### Description

To ensure the **serializable isolation** level for transactions, we have implemented **optimistic concurrency control** (OCC) in v1.0. Refer to this paper for more details -- https://www.eecs.harvard.edu/~htk/publication/1981-tods-kung-robinson.pdf 

This OCC approach attaches a version to each state update and uses it to detect serializable violations. As this package is core to the `ledger` package and non-ledger packages must not use it, we move version package to `core/ledger/internal/pkg`

#### Additional details

A test in `core/common/validation/statebased` unnecessarily imported `version` package. This PR fixes that too as the `core/ledger/internal/pkg` can be imported only within the `core/ledger`

#### Related issues

[FAB-17707](https://jira.hyperledger.org/browse/FAB-17707)


mastersingh24 (approved):
Might not need to use `pkg` as `internal` would be ok, but at least we'll know this has multiple consumers.

cendhu:
@mastersingh24 Agree. After your comment, I did google search and found this article https://dave.cheney.net/2019/10/06/use-internal-packages-to-reduce-your-public-api-surface 

>  Packages within an internal/ directory are therefore said to be internal packages

Among the many goals of ledger refactoring, one of them is to reduce the folder depth. `core/ledger/internal/pkg/version` is definitely not helping as compared to what you suggested, i.e., `core/ledger/internal/version`. Hence, I will push a PR to remove the `pkg` within the internal.

PR #993, opened 2020-04-03
mv keys/update_batches to ledger/internal/state

cendhu:
#### Type of change

- Test Updates

#### Description

In the ledger package, we have 

1. a bunch of composite keys. 
2. an update batch to hold block's public, hashed, and private states (holds only the valid tx's write-set after the tx is validated by both vscc and serializability check). 
3. an update batch to hold these states combined to update the state database (uses actual key format used at the statedb side).

These `structs`/`methods` are used throughout the ledger package. Hence, we move these structs and related methods to `core/ledger/internal/state`

Add UT to improve the test coverage. Removed some dead-code too. 

#### Related issues

[FAB-17708](https://jira.hyperledger.org/browse/FAB-17688)

### DIFF
--- a/core/ledger/internal/state/mock/results_iterator.go
+++ b/core/ledger/internal/state/mock/results_iterator.go
@@ -4,7 +4,7 @@ package mock
 import (
 	"sync"
 
-	"github.com/hyperledger/fabric/core/ledger/kvledger/txmgmt/statedb"
+	"github.com/hyperledger/fabric/core/ledger/internal/state"
 )
 
 type ResultsIterator struct {
@@ -12,16 +12,16 @@ type ResultsIterator struct {
 	closeMutex       sync.RWMutex
 	closeArgsForCall []struct {
 	}
-	NextStub        func() (statedb.QueryResult, error)
+	NextStub        func() (state.QueryResult, error)
 	nextMutex       sync.RWMutex
 	nextArgsForCall []struct {
 	}
 	nextReturns struct {
-		result1 statedb.QueryResult
+		result1 state.QueryResult
 		result2 error
 	}
 	nextReturnsOnCall map[int]struct {
-		result1 statedb.QueryResult
+		result1 state.QueryResult
 		result2 error
 	}
 	invocations      map[string][][]interface{}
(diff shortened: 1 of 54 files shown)


manish-sethi:
In this CR, you simply moved all the types into ledger/internal package. Part of the reason that causes hinderance in readability in the code that we have multiple types of structs and batches across the code. One type of structures that we had in the beginning for statedb level and another types that we introduced for pvt and hashed data. I think that it would be better to evaluate if we can remove all together some structs. 

For instance, a high level idea is that everywhere we can simply use the struct that can hold everything and pass that straight to statedb and then perhaps we can get rid of statedb level update batches and related structures (and perhaps, the privacyenabledstatedb package altogether).

To be honest, otherwise, it was more clear having the statedb related structures in the statedb package and other packages are anyways dependent on it - wdyt?

manish-sethi (changes requested):
Holding this for now till we decide on the above comment

cendhu:
`UpdateBatch` is used in both `stateleveldb` and `statecouchdb` packages in addition to `validator` and `privacyenabledstate` -- 4 packages in total.

Pub, Hash, Pvt batch is mainly used by the `validator` and `privacyenabedstate` package -- two packages in total. 

These kinds of imports do not look correct to me and hence, I have moved all the structures to an internal package.

`privacyenabledstate` converts the Pub, Hash, Pvt batch to a single `UpdateBatch` by constructing keys for hash and pvt data before passing it to `stateleveldb` and `statecouchdb` package. This is applied because we have not made both statedbs aware of hash or Pvt data as they simply handle KV pairs. 

In the subsequent PRs, we will merge `privacyenabledstate` with the `statedb` package.

cendhu:
Based on the offline discussion, we decided to not move everything to the internal package. However, I am not convinced in using a single update batch across the package as I like the current abstraction that statedb provides. For now, I am closing this PR. To have a consensus on the final ledger package structure, I am planning to push one big PR which would represent the final structure that I have in mind. We can use that PR for discussion and decide on how we want to refactor going further.

PR #995, opened 2020-04-05
Documentation doesn't match sample files

baartch:
#### Type of change

- Bug fix

#### Description

The documentation does not match with the sample files. Which one is more accurate?
Usually the documentation lags behind...

Even thought, it seems that line 220 and 247 are probably more correct in the documentation, than in the issue.js file from the samples.
That's why I created a pull request on the samples repository as well.
https://github.com/hyperledger/fabric-samples/pull/153/commits/4b1e612e3d44a32349e6204a93c4c33fb5c6bad2


nikhil550:
Thanks for this PR. The code in 2.0 has been updated since the doc was written.

nikhil550 (dismissed):

denyeart:
@nikhil550 @TheBookThief  So which branches is the update required in?
Typically we make a fix to master branch first, and then backport to release branches if required.
Is there a reason this PR targets release-2.0 initially?

baartch:
I'm really sorry if I made it wrong. I'm an absolute newbie...
I closed this pull request and try to make a new one for the master branch.

PR #996, opened 2020-04-05
mv proto out of msg package (stateleveldb)

cendhu:
#### Type of change

#### Description

In `goleveldb`, we store a KV pair of format `{key, [value, version, metadata]}` for every state managed by the blockchain. As the `[value, version, metadata]` must be passed as a single `[]byte` (as per the goleveldb requirements), we use a proto message to serialize the combined values.

Currently, we have a separate package to have this proto. This is unjustifiable. Hence, we move this proto to `stateleveldb` package itself and change the proto message to `DBValue` from `VersionedValueProto` for better readability.

#### Additional details

I would have preferred the following 
```
message DBValue {
    bytes value = 1;
    bytes version = 2;
    bytes metadata = 3;
}
```
instead of
```
message DBValue {
    bytes version = 1;
    bytes value = 2;
    bytes metadata = 3;
}
```
However, we need to be very careful in changing this proto as it would break the backward compatibility. As per [protobuf guideline](https://developers.google.com/protocol-buffers/docs/proto), we cannot change the field number. Hence, we have changed only `version_bytes` to `version` and it is completely fine.

_**Note that we have enough backward compatibility tests in [core/ledger/kvledger/tests](https://github.com/hyperledger/fabric/tree/master/core/ledger/kvledger/tests) to ensure that we are not breaking anything in this PR.**_ 

#### Related issues

[FAB-17713](https://jira.hyperledger.org/browse/FAB-17713)


core/ledger/kvledger/txmgmt/statedb/stateleveldb/persisted_value.proto
```diff
@@ -6,12 +6,12 @@ SPDX-License-Identifier: Apache-2.0
...
 
-option go_package = "github.com/hyperledger/fabric/core/ledger/kvledger/txmgmt/statedb/stateleveldb/msgs";
+option go_package = "github.com/hyperledger/fabric/core/ledger/kvledger/txmgmt/statedb/stateleveldb";
 
-package msgs;
+package stateleveldb;
 
-message VersionedValueProto {
-    bytes version_bytes = 1;
+message PersistedValue {
```
manish-sethi: I don't find this change meaningful. "PersistedValue" is a very specific term that I may use for a variable when I assign it a value from a store. For a type, I don't find this term much meaningful.

manish-sethi (changes requested):

core/ledger/kvledger/txmgmt/statedb/stateleveldb/persisted_value.proto
cendhu: I see. Okay. Then, we need to pick between `VersionedValue` and `VersionedValueWithMetadata`. We cannot have the `Proto` in the variable name.  Or May be `DBValue` similar to `CacheValue` that we have.

manish-sethi: `DBValue` seems concise and meaningful to me

cendhu: Thanks, Manish. Will use that name then.

manish-sethi (approved):

PR #997, opened 2020-04-05
mv proto out of msg package (statecouchdb)

cendhu:
#### Type of change

#### Description

Currently, we have a proto message named `VersionFieldProto` inside `msg` package to enclose both `Version` and `Metadata`. This is disinformation and a separate `msg` package is unjustified. Hence, in this PR, we move the `VersionFieldProto` out of the `msg` package and rename it to `VersionAndMetadata` proto for better readability.

Further, this PR renames the variable named `VersionBytes` to `Version`. As this is just a renaming and no field number has changed, this does not break any backward compatibility.

#### Additional details

Unlike goleveldb, in CouchDB, we need to pass JSON bytes and attachments (if present) for each blockchain state. We have an internal JSON format defined in the `statecouchdb` package so that we can set a version and encapsulate the JSON passed by the chaincode. The version bytes is set in the `~version` field in the internal JSON value. When the state-based endorsement was introduced, we didn't introduce a separate metadata field instead created a proto that includes both version and metadata. Then the serialized proto message was set in the version field. 

It is a kind of hacky solution and not much can be done now as we need to provide backward compatibility (otherwise, we need to rebuild states).

#### Related issues

[FAB-17714](https://jira.hyperledger.org/browse/FAB-17714)


manish-sethi (dismissed):
LGTM. However, needs a rebase

cendhu:
rebased and fixed conflicts

manish-sethi (approved):

PR #999, opened 2020-04-05
Externally built and launched chaincodes cleanup on signal

wlahti:
#### Type of change

- Bug fix

#### Description

When a peer is terminated via SIGTERM or SIGINT, it should ensure any externally built and launched chaincodes have a chance to cleanup.

#### Related issues

[FAB-17155](https://jira.hyperledger.org/browse/FAB-17155)


core/container/externalbuilder/externalbuilder.go
```diff
@@ -380,23 +384,25 @@ func (b *Builder) Run(ccid, bldDir string, peerConnection *ccintf.PeerConnection
...
 		os.RemoveAll(launchDir)
 		return nil, errors.Wrapf(err, "builder '%s' run failed to start", b.Name)
 	}
 
 	go func() {
-		defer os.RemoveAll(launchDir)
+		sess.WaitGroup.Add(1)
 		sess.Wait()
+		os.RemoveAll(launchDir)
+		sess.WaitGroup.Done()
```
sykesm: I don't see how this can work. We're waiting before calling Done in the same synchronous block of code. Can you help me understand how this works?

core/container/container.go
```diff
@@ -78,6 +78,11 @@ type Router struct {
 	mutex           sync.Mutex
 }
 
+// Containers returns the map of containers that have been started.
+func (r *Router) Containers() map[string]Instance {
+	return r.containers
```
sykesm: 1. Do we actually need a map returned?
2. Do we want to return a reference to the instance we're managing or a copy?

I think both could be addressed by returning a JIT allocated slice for the consumer.

core/container/externalbuilder/externalbuilder.go
```diff
@@ -268,10 +268,12 @@ type Builder struct {
 	Logger       *flogging.FabricLogger
 	Name         string
 	MSPID        string
+	signalChan   <-chan struct{}
+	doneChan     chan<- struct{}
 }
 
 // CreateBuilders will construct builders from the peer configuration.
-func CreateBuilders(builderConfs []peer.ExternalBuilder, mspid string) []*Builder {
+func CreateBuilders(builderConfs []peer.ExternalBuilder, mspid string, signalChan <-chan struct{}, doneChan chan<- struct{}) []*Builder {
```
sykesm: Instead of wiring these, and instead of wiring them to the constructor, can we wire a context through to the appropriate methods?

A context would allow us to orchestrate cancelation using a relatively common pattern.

sykesm (changes requested):
Just a few comments on the implementation and approach for your consideration.

I'd also ask for an updated integration test that ensures that the run directories are actually removed when the peers are shutdown. This may require some explicit wiring to control where the build directories are created or to grab it out of one of the external builder implementations.

core/container/externalbuilder/externalbuilder.go
wlahti: The WaitGroup here is for everything waiting on the session, which in this case is two things: the instance itself calls session.Wait and then this builder.Run code calls session.Wait on the same session. The instance code in this PR waits on the WaitGroup to complete and then signals via the doneChan.

sykesm: So, I misread the code. Sorry for the confusion.

That said, I'm still not sure I'm on board with this. It's very odd to see a user of `Session` reaching into it to update an explicit WaitGroup and then call Wait. Most people would expect the wait to take care of all of that.

core/container/externalbuilder/session.go
```diff
@@ -24,13 +25,18 @@ type Session struct {
 	exited     chan struct{}
 	exitErr    error
 	waitStatus syscall.WaitStatus
+	signalChan <-chan struct{}
+	doneChan   chan<- struct{}
+	WaitGroup  sync.WaitGroup
```
sykesm: This should not be exported; notice nothing else in here is exported. I'm also not convinced it's needed.

core/container/externalbuilder/session_test.go
```diff
@@ -71,18 +71,37 @@ var _ = Describe("Session", func() {
...
+		sess, err := externalbuilder.Start(logger, cmd, signalCh, nil)
+		Expect(err).NotTo(HaveOccurred())
+
+		exitCh := make(chan error)
+		go func() { exitCh <- sess.Wait() }()
+
+		Consistently(exitCh).ShouldNot(Receive())
+		signalCh <- struct{}{}
+		Eventually(exitCh).Should(Receive(MatchError("closing session due to signal")))
+
```
sykesm, line 90: nit: unnecessary blank line

internal/peer/node/start.go
```diff
@@ -828,9 +834,32 @@ func serve(args []string) error {
 		}()
 	}
 
+	signalFunc := func() {
+		// notify all externally built chaincodes of signal
+		close(externalCCSignalChan)
```
sykesm: Why would you bury this here instead of a defer out of the caller? This makes the function a one-shot and if it's called multiple times, it's going to panic.

internal/peer/node/start.go
```diff
@@ -828,9 +834,32 @@ func serve(args []string) error {
...
+
+		// wait for a message from each externally built cc that
+		// has been launched that it has terminated (and cleaned
+		// up after itself) properly
+		for ccid, instance := range containerRouter.Containers() {
```
sykesm: Going back to my other context comment, if we wired with a parent context, we could cancel everything in one shot  instead of looping.

An even cleaner implementation would be a `Shutdown`, `Kill`, or `Stop` implementation on the router that handles the details of all of this. We would want that covered with a timer to ensure we don't wait too long for cleanup.

core/container/externalbuilder/instance.go
```diff
@@ -186,7 +186,13 @@ func (i *Instance) Wait() (int, error) {
 		return -1, errors.Errorf("instance was not successfully started")
 	}
 
+	i.Session.WaitGroup.Add(1)
 	err := i.Session.Wait()
+	i.Session.WaitGroup.Done()
```
sykesm: Given session is a synchronous wait, I'm not sure what this is doing.

core/container/externalbuilder/instance.go
```diff
@@ -186,7 +186,13 @@ func (i *Instance) Wait() (int, error) {
...
+	// have the Instance wait for all goroutines waiting on the Session
+	i.Session.WaitGroup.Wait()
+	// signal that this chaincode instance is done (i.e. successfully cleaned up)
+	i.Session.doneChan <- struct{}{}
```
sykesm: ironically, your comment is exactly what `Wait` does. It's the implementation that's making things more confusing.

core/container/externalbuilder/session.go
```diff
@@ -76,9 +84,13 @@ func (s *Session) waitForExit(logger *flogging.FabricLogger, stderr io.Reader) {
 // Wait waits for the running command to terminate and returns the exit error
 // from the command. If the command has already exited, the exit err will be
 // returned immediately.
-func (s *Session) Wait() error {
-	<-s.exited
-	return s.exitErr
+func (s *Session) Wait(optText ...string) error {
```
sykesm: What is `optText`? What is it used for?

core/container/externalbuilder/session.go
```diff
@@ -24,13 +25,18 @@ type Session struct {
...
 
 // Start will start the provided command and return a Session that can be used
-// to await completion or signal the process.
+// to await completion or signal the process. The session will also listen for
+// any messages on the signalChan, at which time it will perform its cleanup
+// and send a message on the doneChan.
 //
 // The provided logger is used log stderr from the running process.
-func Start(logger *flogging.FabricLogger, cmd *exec.Cmd) (*Session, error) {
+func Start(logger *flogging.FabricLogger, cmd *exec.Cmd, signalChan <-chan struct{}, doneChan chan<- struct{}) (*Session, error) {
```
sykesm: I don't understand these two channels. The `signalChan` seems to be carrying a cancelation notification and I really don't understand the `doneChan`'s role given the `Wait` implementation.

internal/peer/node/start.go
```diff
@@ -828,9 +834,32 @@ func serve(args []string) error {
...
+			if started {
+				// we don't care which cc we're receiving from,
+				// as long as we receive one message per cc that
+				// has been built and launched.
+				<-externalCCDoneChan
+			}
+		}
+		serve <- nil
+	}
+
```
sykesm: This does not belong way up here. The implementation that handles the shutdown needs to live down with the handlers.

sykesm (changes requested):
I really think we need to step back here. The goal is to continue to have directory cleanup performed after chaincode instances and build processes terminate. Both of those seem to be synchronized behind a `session.Wait()` call.

In order to trigger the termination of the chaincode when the peer is killed, we need to wire in a mechanism for asynchronous termination of the chaincode execution and/or build processes. The canonical implementation of that in go is the `context.Context`. The context could be provided to `exec.CommandContext` to cause the process represented by the session to be killed when an appropriate signal is received by the peer. The session will continue to wait for the process to terminate (it's calling `exec.Cmd#Wait`) so the existing cleanup should run at the correct time.

An even simpler implementation would be to create a `Shutdown` function on the container router that handles the termination and cleanup. In the case of external builders, it would use the `Stop` operation on the session to terminate the process backing each session using the TERM then KILL protocol. (You've already got all this code.) This function would simply be called on SIGTERM or SIGINT (with some sort of timeout covering it to prevent hangs at termination).

I really don't understand the new WaitGroup or some of the processing around it. To me, both of the options above seem like straightforward implementations. When the peer receives a signal, it should try to terminate chaincode before exiting.

The trick, if we want to call it that, is to prevent the peer from terminating until the shutdown and cleanup are complete.

core/container/externalbuilder/session.go
wlahti: Oops, this was just put in to debug my way through the flow and where exactly session.Wait was being called.

sykesm (changes requested):
We're getting closer. :)

Adding a new configuration option and wiring it all the way through seems unnecessary and is causing a fair bit of unrelated change. The directory you want to ensure is removed is the last argument provided to the run script. We control the run script implementation in the integration test and the environment it runs in (through the env provided to the peer and the environment whitelist).

Outside of the `Shutdown` implementation and wiring that into the signal handler, I'd expect everything else to be in test. The run script should do something to save the args in a place that you can find when the test completes. You can then ensure the directory exists while the peer is running and that it goes away after the peer is stopped.

core/container/externalbuilder/session.go
```diff
@@ -70,6 +72,7 @@ func (s *Session) waitForExit(logger *flogging.FabricLogger, stderr io.Reader) {
 	defer s.mutex.Unlock()
 	s.exitErr = err
 	s.waitStatus = s.command.ProcessState.Sys().(syscall.WaitStatus)
+	os.RemoveAll(s.launchDir)
```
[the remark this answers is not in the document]
wlahti: This was the simplest way to keep complexity down and ensure that the resources were cleaned up before returning. Otherwise, I haven't found a simple way to contain the os.RemoveAll calls and guarantee that they've executed before the peer exits.

core/container/container.go
```diff
@@ -152,3 +154,21 @@ func (r *Router) Stop(ccid string) error {
...
+		case <-time.After(r.TermTimeout):
+			vmLogger.Warningf("timeout while stopping ccid: %s", ccid)
+			continue
+		case <-done:
+			if err != nil {
+				vmLogger.Warningf("failed to stop ccid: %s err: %s", ccid, err)
+			}
+		}
+	}
+}
```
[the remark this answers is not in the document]
wlahti: Waiting on unit tests in case I completely rewrite all of the production code again. :)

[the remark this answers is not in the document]
sykesm: Fair enough.

sykesm:
Going through all of this again, I think the window you're trying to close is where the deferred cleanup of the run directory does not happen because it's async to the shutdown. So, yes, I see why you're trying to change things and why you are approaching this as you are.

I'd still like to keep the changes required to do this small and consistent with minimal external impacts so let me see if I can find a straightforward way to do that.

Outside of that, a top-level shutdown that is synchronous and wired into the signal handling is going to be required. I don't think there's any argument there. Let's use that as a starting point.

core/container/container.go
```diff
@@ -76,6 +77,7 @@ type Router struct {
 	containers      map[string]Instance
 	PackageProvider PackageProvider
 	mutex           sync.Mutex
+	TermTimeout     time.Duration
```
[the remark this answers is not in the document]
wlahti: Done

internal/peer/node/start.go
```diff
@@ -580,7 +580,9 @@ func serve(args []string) error {
 			},
 			LegacyCCPackageLocator: &ccprovider.CCInfoFSImpl{GetHasher: factory.GetDefault()},
 		},
+		TermTimeout: 5 * time.Second,
 	}
+	defer containerRouter.Shutdown()
```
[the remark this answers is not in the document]
wlahti: Done

core/container/container.go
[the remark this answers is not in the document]
wlahti: Updated this to stop the instances in parallel.

core/container/externalbuilder/externalbuilder.go
```diff
@@ -363,7 +365,7 @@ func newRunConfig(ccid string, peerConnection *ccintf.PeerConnection, mspid stri
 // Run starts the `run` script and returns a Session that can be used to
 // signal it and wait for termination.
 func (b *Builder) Run(ccid, bldDir string, peerConnection *ccintf.PeerConnection) (*Session, error) {
-	launchDir, err := ioutil.TempDir("", "fabric-run")
+	launchDir, err := ioutil.TempDir(b.LaunchDir, "fabric-run")
```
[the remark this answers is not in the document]
wlahti: Removed.

core/container/externalbuilder/externalbuilder.go
```diff
@@ -268,6 +268,7 @@ type Builder struct {
 	Logger       *flogging.FabricLogger
 	Name         string
 	MSPID        string
+	LaunchDir    string
```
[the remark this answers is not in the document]
wlahti: Removed.

integration/nwo/core_template.go
```diff
@@ -195,6 +195,7 @@ chaincode:
       environmentWhitelist: {{ range .EnvironmentWhitelist }}
          - {{ . }}
       {{- end }}
+      launchDir: {{ .LaunchDir }}
```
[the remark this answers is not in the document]
wlahti: Removed.

wlahti:
> We're getting closer. :)
> 
> Adding a new configuration option and wiring it all the way through seems unnecessary and is causing a fair bit of unrelated change. The directory you want to ensure is removed is the last argument provided to the run script. We control the run script implementation in the integration test and the environment it runs in (through the env provided to the peer and the environment whitelist).
> 
> Outside of the `Shutdown` implementation and wiring that into the signal handler, I'd expect everything else to be in test. The run script should do something to save the args in a place that you can find when the test completes. You can then ensure the directory exists while the peer is running and that it goes away after the peer is stopped.

Thanks for all the insightful reviews, Matt. I always learn a lot. :) 

Removed LaunchDir from the external builder config and used an environment variable in just the integration tests to track the artifacts that are being generated (and ultimately removed).

wlahti:
Rebased on master to keep things up-to-date. Any thoughts on next steps here, @sykesm? I guess you don't want to merge as-is for now and follow-up via another JIRA task asap?

integration/externalbuilders/binary/bin/run
```diff
@@ -14,6 +14,12 @@ fi
 OUTPUT=$1
 ARTIFACTS=$2
 
+# when set, keep track of the artifacts directories to verify
+# proper cleanup in integration test
+if [ -n "${RUN_ARTIFACTS_FILE+1}" ]; then
```
sykesm, line 19: curious: What is the `+1` for?

wlahti: This is to avoid an unbound variable for the integration tests that don't need to set this environment variable (aka every other one besides the one I've updated in this PR). Not sure it's the "best" way to handle it but it seemed decent after consulting around the internet. :)

wlahti:
Weekly rebase.

wlahti:
@sykesm Can we merge this as-is and create another task for getting the cleanup responsibility just right? At least this ensures we don't end up with the leftover "fabric-run" temp directories.

sykesm:
> @sykesm Can we merge this as-is and create another task for getting the cleanup responsibility just right? At least this ensures we don't end up with the leftover "fabric-run" temp directories.

I wanted to agree and merge but I just pulled it down and the integration tests are not actually testing that the directories are removed. The way it's structured, the test opens a file and passes it as the reader to a scanner to verify that there are files. Later, during the after each, it uses the same file and attempts to scan. Since the file was never closed and the reader has been advanced to the end, the scanner doesn't return anything. Since nothing is returned, the loop doesn't run.

Basically, the test is broken.

wlahti:
Good catch. I didn't realize exactly how scanners worked in conjunction with files. After each test run, I was also manually verifying that the files were gone but didn't realize the AfterEach() scanner loop wasn't running. Updated the test.

integration/e2e/e2e_test.go
```diff
@@ -179,15 +181,12 @@ var _ = Describe("EndToEnd", func() {
 			nwo.DeployChaincode(network, "testchannel", orderer, chaincode)
 
 			By("ensuring external cc run artifacts exist after deploying")
-			runArtifactsFile, err := os.Open(runArtifactsFilePath)
+			contents, err := ioutil.ReadFile(runArtifactsFilePath)
```
sykesm, line 184: One thing that should probably be added is a check to ensure that contents is not empty.

sykesm:
I pushed a commit on top that changes the cleanup to a callback instead of simple delegation; it keeps the session generic and removes the race.

I also tried to simplify the integration test slightly. I just noticed that the test should likely ensure that there are directories to validate so we can ensure the assertions run. If you're okay with what's here, we can merge.

sykesm (approved):
Pending agreement from @wlahti on the additional changes.

wlahti:
Looks good to me. Walked through the changes and thought through each part to learn from your improvements. Thanks, Matt!

PR #1002, opened 2020-04-06
3 small fixes that documentation matches samples

baartch:
#### Type of change
- Documentation update

#### Description
I stumbled over three parts in the documentation, that don't match with the sample files.

And the following two lines are also different in the sample file. You may check these too
https://github.com/hyperledger/fabric/blame/master/docs/source/developapps/application.md#L219
https://github.com/hyperledger/fabric/blame/master/docs/source/developapps/application.md#L246

### DIFF
--- a/docs/source/developapps/application.md
+++ b/docs/source/developapps/application.md
@@ -57,7 +57,7 @@ Towards the top of `issue.js`, you'll see two Fabric classes are brought
 into scope:
 
 ```JavaScript
-const { FileSystemWallet, Gateway } = require('fabric-network');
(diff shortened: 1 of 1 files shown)


denyeart:
@TheBookThief Thanks for the fix... it actually was an honest question when I asked which release branches this fix should be targeted for.
I assume it is valid for release-2.1 and release-2.0, therefore I'll add comments that should auto-backport PRs for those releases (there is a ReadTheDocs version per release branch).

denyeart (approved):

denyeart:

denyeart:

PR #1004, opened 2020-04-06
Enable race detector for gossip packages in unit test

sykesm:
This is a follow up on #1003 and enables the race detector for gossip packages. This PR should not be merged unless #1003 is.


yacovm (approved):

yacovm:
I don't think the CI servers handle the load very well.... perhaps consider to revert this?

PR #1013, opened 2020-04-07
[FAB-16879] Add stack trace to couchdb http errors

denyeart:
#### Type of change

- Error handling improvement

#### Description

If there is http error calling couchdb, no context was provided
to higher level code that handles the errors.

This change adds stack trace to the http error at the point that the error
is raised, so that admins can identify what was going on upon hitting the error.
In most places, the higher level error handlers are already printing stack
trace, if available, e.g. in the commit path and recommit lost block path.

Also, for each of the points in ledger code where go routines are used,
if stack trace was added in the go routine error, the stack would only
be available from the point the go routine was called and lower. This
is resolved by also adding stack trace to errors that are returned from
go routines.

### DIFF
--- a/core/ledger/kvledger/txmgmt/statedb/statecouchdb/batch_util.go
+++ b/core/ledger/kvledger/txmgmt/statedb/statecouchdb/batch_util.go
@@ -8,6 +8,8 @@ package statecouchdb
 
 import (
 	"sync"
+
+	"github.com/pkg/errors"
 )
 
 // batch is executed in a separate goroutine.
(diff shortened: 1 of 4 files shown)


lindluni:
You alright over there Dave? Little twitchy on the close/reopen button 😆

ryjones:
@btl5037 usually a sign of a `git push --force` to the wrong branch

denyeart:
I force pushed to the right branch, but it looks like after you close a PR you can't force push and re-open. Anyways, I made the change in https://github.com/hyperledger/fabric/pull/1015 instead.

PR #1014, opened 2020-04-07
Add test to ensure CRL can be signed by intermediate CA cert

wlahti:
#### Type of change

- Test update

#### Description

- Added a test to ensure a CRL can be created and signed by an intermediate CA cert. 
- Removed dead code from msp.go

#### Related issues

[FAB-17694](https://jira.hyperledger.org/browse/FAB-17694)

### DIFF
--- a/pkg/config/msp.go
+++ b/pkg/config/msp.go
@@ -697,38 +697,3 @@ func validateCACerts(caCerts []*x509.Certificate) error {
 
 	return nil
 }
-
-func addMSPConfigToOrg(org *cb.ConfigGroup, fabricMSPConfig *mb.FabricMSPConfig) error {
-	configValue, err := getOrgMSPValue(org)
-	if err != nil {
-		return err
-	}
-
-	mspConfig := &mb.MSPConfig{}
-	err = proto.Unmarshal(configValue.Value, mspConfig)
-	if err != nil {
-		return fmt.Errorf("unmarshalling mspConfig: %v", err)
-	}
-
-	serializedMSPConfig, err := proto.Marshal(fabricMSPConfig)
-	if err != nil {
-		return fmt.Errorf("marshalling updated mspConfig: %v", err)
-	}
-
-	mspConfig.Config = serializedMSPConfig
-
-	err = setValue(org, mspValue(mspConfig), AdminsPolicyKey)
-	if err != nil {
-		return err
-	}
-
-	return nil
-}
-
-func getOrgMSPValue(org *cb.ConfigGroup) (*cb.ConfigValue, error) {
-	configValue, ok := org.Values[MSPKey]
-	if !ok {
-		return nil, errors.New("org doesn't have MSP value")
-	}
-	return configValue, nil
-}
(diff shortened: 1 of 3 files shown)


pkg/config/msp_test.go
```diff
@@ -947,24 +947,53 @@ func TestCreateApplicationMSPCRL(t *testing.T) {
...
+		signingIdentity := &SigningIdentity{
+			Certificate: org2IntermediateCert,
+			PrivateKey:  org2IntermediatePrivKey,
+			MSPID:       "MSPID",
+		}
+		certToRevoke, _ := generateCertAndPrivateKeyFromCACert(t, "org2.example.com", org2IntermediateCert, org2IntermediatePrivKey)
+		org2CRL, err := c.CreateApplicationMSPCRL("Org2", signingIdentity, certToRevoke)
+		gt.Expect(err).NotTo(HaveOccurred())
+		gt.Expect(org2CRL.TBSCertList.RevokedCertificates).To(HaveLen(1))
+		gt.Expect(org2CRL.TBSCertList.RevokedCertificates[0].SerialNumber).To(Equal(certToRevoke.SerialNumber))
```
yacovm: I think that this is not enough and you should also call [CheckCRLSignature](https://golang.org/pkg/crypto/x509/#Certificate.CheckCRLSignature) to ensure the CRL was properly signed by the intermediate CA. 

The same thing should be done for root CAs without intermediate CAs, that issue leaf certificates.

wlahti: Good call. Definitely makes the test more thorough. Done!

pkg/config/msp.go
```diff
@@ -697,38 +697,3 @@ func validateCACerts(caCerts []*x509.Certificate) error {
...
-	return nil
-}
-
-func getOrgMSPValue(org *cb.ConfigGroup) (*cb.ConfigValue, error) {
-	configValue, ok := org.Values[MSPKey]
-	if !ok {
-		return nil, errors.New("org doesn't have MSP value")
-	}
-	return configValue, nil
-}
```
caod123, line 734: I'm assuming these functions weren't being used and hence removed.

pkg/config/msp_test.go
```diff
@@ -947,24 +947,57 @@ func TestCreateApplicationMSPCRL(t *testing.T) {
...
+	org2Cert, org2PrivKey, _ := certPrivKeyCRL(org2MSP)
+	org2IntermediateCert, org2IntermediatePrivKey := generateIntermediateCACertAndPrivateKey(t, "org2.example.com", org2Cert, org2PrivKey)
+	org2MSP.IntermediateCerts = append(org2MSP.IntermediateCerts, org2IntermediateCert)
+	err = originalConfigTx.UpdateApplicationMSP(org2MSP, "Org2")
 	gt.Expect(err).NotTo(HaveOccurred())
-	gt.Expect(newCRL.TBSCertList.RevokedCertificates).To(HaveLen(1))
-	gt.Expect(newCRL.TBSCertList.RevokedCertificates[0].SerialNumber).To(Equal(certToRevoke.SerialNumber))
 
-	// TODO create a CRL using an intermediate cert
+	t.Run("create CRL using a root cert", func(t *testing.T) {
```
caod123: nit: can we table test these? seems like ultimately they're both testing using a different certToRevoke which can be set up beforehand pretty easily. Alternatively you could just combine both of them into one test that uses the same setup and just does one check after the other on the same config.

caod123 (approved):

PR #1022, opened 2020-04-08
PKCS11: Validate session and get new if invalid

mrshah-at-work:
Signed-off-by: Mihir Shah <mrshah@us.ibm.com>

#### Type of change

- Bug fix

#### Description

As described in [JIRA FAB-17722](https://jira.hyperledger.org/browse/FAB-17722):
Currently the pkcs11 code in fabric's bccsp module is set to have a session cache and reuse sessions if available in cache. If a session goes bad (due to connection issues with HSM), the session is not evicted from cache and will be reused. If all sessions go bad, the client will never be able to recover and keep using bad sessions.

#### Additional details

This code change:
- Moves creation of a new session to a new function `CreateSession`
- Validates the existing session by calling `GetSessionInfo`
- If found invalid, creates a new session and expects the old one to be garbage collected.
- The current test in pkcs11_test.go `TestPKCS11GetSession`, will test the new function as well. Let me know if I should add additional test to test `CreateSession` function.

#### Related issues

https://jira.hyperledger.org/browse/FAB-17722

### DIFF
--- a/bccsp/pkcs11/pkcs11.go
+++ b/bccsp/pkcs11/pkcs11.go
@@ -54,18 +54,8 @@ func loadLib(lib, pin, label string) (*pkcs11.Ctx, uint, *pkcs11.SessionHandle, 
 		return nil, slot, nil, fmt.Errorf("Could not find token with label %s", label)
 	}
 
-	var session pkcs11.SessionHandle
-	for i := 0; i < 10; i++ {
-		session, err = ctx.OpenSession(slot, pkcs11.CKF_SERIAL_SESSION|pkcs11.CKF_RW_SESSION)
-		if err != nil {
-			logger.Warningf("OpenSession failed, retrying [%s]\n", err)
-		} else {
-			break
-		}
-	}
-	if err != nil {
-		logger.Fatalf("OpenSession [%s]\n", err)
-	}
+	session := createSession(ctx, slot, pin)
+
 	logger.Debugf("Created new pkcs11 session %+v on slot %d\n", session, slot)
 
 	if pin == "" {
@@ -84,25 +74,44 @@ func loadLib(lib, pin, label string) (*pkcs11.Ctx, uint, *pkcs11.SessionHandle, 
 func (csp *impl) getSession() (session pkcs11.SessionHandle) {
 	select {
 	case session = <-csp.sessions:
-		logger.Debugf("Reusing existing pkcs11 session %+v on slot %d\n", session, csp.slot)
+		_, err := csp.ctx.GetSessionInfo(session)
+		if err != nil {
+			logger.Warningf("Get session info failed [%s], closing existing session and getting a new session\n", err)
+			csp.ctx.CloseSession(session)
+			session = createSession(csp.ctx, csp.slot, csp.pin)
+		} else {
+			logger.Debugf("Reusing existing pkcs11 session %+v on slot %d\n", session, csp.slot)
+		}
 
 	default:
 		// cache is empty (or completely in use), create a new session
-		var s pkcs11.SessionHandle
-		var err error
-		for i := 0; i < 10; i++ {
-			s, err = csp.ctx.OpenSession(csp.slot, pkcs11.CKF_SERIAL_SESSION|pkcs11.CKF_RW_SESSION)
-			if err != nil {
-				logger.Warningf("OpenSession failed, retrying [%s]\n", err)
-			} else {
-				break
-			}
-		}
+		session = createSession(csp.ctx, csp.slot, csp.pin)
+	}
+	return session
+}
+
+func createSession(ctx *pkcs11.Ctx, slot uint, pin string) pkcs11.SessionHandle {
+	var s pkcs11.SessionHandle
+	var err error
+	for i := 0; i < 10; i++ {
+		s, err = ctx.OpenSession(slot, pkcs11.CKF_SERIAL_SESSION|pkcs11.CKF_RW_SESSION)
 		if err != nil {
-			panic(fmt.Errorf("OpenSession failed [%s]", err))
+			logger.Warningf("OpenSession failed, retrying [%s]\n", err)
+		} else {
+			break
 		}
-		logger.Debugf("Created new pkcs11 session %+v on slot %d\n", s, csp.slot)
-		session = s
+	}
+	if err != nil {
+		logger.Fatalf("OpenSession failed [%s]", err)
+	}
+	logger.Debugf("Created new pkcs11 session %+v on slot %d\n", s, slot)
+	session := s
+
+	err = ctx.Login(session, pkcs11.CKU_USER, pin)
+	if err != nil {
+		if err != pkcs11.Error(pkcs11.CKR_USER_ALREADY_LOGGED_IN) {
+			logger.Errorf("Login failed [%s]", err)
+		}
 	}
 	return session
 }
--- a/bccsp/pkcs11/impl.go
+++ b/bccsp/pkcs11/impl.go
@@ -54,7 +54,7 @@ func New(opts PKCS11Opts, keyStore bccsp.KeyStore) (bccsp.BCCSP, error) {
 	}
 
 	sessions := make(chan pkcs11.SessionHandle, sessionCacheSize)
-	csp := &impl{swCSP, conf, keyStore, ctx, sessions, slot, lib, opts.SoftVerify, opts.Immutable, opts.AltId}
+	csp := &impl{swCSP, conf, keyStore, ctx, sessions, slot, pin, lib, opts.SoftVerify, opts.Immutable, opts.AltId}
 	csp.returnSession(*session)
 	return csp, nil
 }
@@ -68,6 +68,7 @@ type impl struct {
 	ctx      *pkcs11.Ctx
 	sessions chan pkcs11.SessionHandle
 	slot     uint
+	pin      string
 
 	lib        string
 	softVerify bool


bccsp/pkcs11/pkcs11.go
```diff
@@ -84,25 +84,42 @@ func loadLib(lib, pin, label string) (*pkcs11.Ctx, uint, *pkcs11.SessionHandle,
 func (csp *impl) getSession() (session pkcs11.SessionHandle) {
 	select {
 	case session = <-csp.sessions:
-		logger.Debugf("Reusing existing pkcs11 session %+v on slot %d\n", session, csp.slot)
```
lindluni, line 87: You've removed the Debug statement on the happy path. Would be nice to keep so you can follow the flow of execution

bccsp/pkcs11/pkcs11.go
```diff
@@ -84,25 +84,42 @@ func loadLib(lib, pin, label string) (*pkcs11.Ctx, uint, *pkcs11.SessionHandle,
...
+		_, err := csp.ctx.GetSessionInfo(session)
```
lindluni, line 77: I had mentioned to Saad that  it is possible for the Session to be opened, but to be in an invalid state. So it's not enough to simply call this function. The SessionInfo object returns a DeviceError and State code. When State has a code of 0 (I believe, but I'll double check the spec), the session itself is in an invalid state, and then we also need to eject the session from the pool.

mrshah-at-work: Thanks Brett. When we checked, any error in the session would result in an error response. Please let us know if it is otherwise.

As per the session state, following are the states that it can be:
```
CKS_RO_PUBLIC_SESSION
CKS_RO_USER_FUNCTIONS
CKS_RW_PUBLIC_SESSION
CKS_RW_USER_FUNCTIONS
CKS_RW_SO_FUNCTIONS 
```

For the full info response, search for "CK_SESSION_INFO" in [this document](https://docs.oasis-open.org/pkcs11/pkcs11-base/v2.40/os/pkcs11-base-v2.40-os.html)

bccsp/pkcs11/pkcs11.go
mrshah-at-work: Added the debug statement back

bccsp/pkcs11/pkcs11.go
lindluni: You are right and agreed, the other consideration is section 5.1.1:  CKR_FUNCTION_FAILED which states that even if the ulDeviceError field contains an error code, its likely a subsequent call to the same function would succeed. So I'll retract my comment.

Thanks Mihir for linking to the spec.

bccsp/pkcs11/pkcs11.go
```diff
@@ -84,25 +84,44 @@ func loadLib(lib, pin, label string) (*pkcs11.Ctx, uint, *pkcs11.SessionHandle,
...
+		if err != nil {
+			logger.Warningf("Get session info failed [%s], getting new session", err)
```
lindluni: You missed a `\n` character at the end of this. But I also don't think the message conveys what the check is actually doing, maybe `Existing session invalid, creating a new session\n`, I'm not sure adding the error would provide value to the user, as we've already once established this session, so it's not something the user did that caused the session to fail.

bccsp/pkcs11/pkcs11.go
```diff
@@ -84,25 +84,44 @@ func loadLib(lib, pin, label string) (*pkcs11.Ctx, uint, *pkcs11.SessionHandle,
...
+	var err error
+	for i := 0; i < 10; i++ {
+		s, err = csp.ctx.OpenSession(csp.slot, pkcs11.CKF_SERIAL_SESSION|pkcs11.CKF_RW_SESSION)
 		if err != nil {
-			panic(fmt.Errorf("OpenSession failed [%s]", err))
+			logger.Warningf("OpenSession failed, retrying [%s]\n", err)
+		} else {
+			break
+		}
+	}
```
sykesm: Iterating in a tight loop is rarely a good idea. (It was bad where you copied it from and it's bad here.)

If it's not successful on the first attempt, it's unlikely it's going to be immediately accessible on the second. This should probably be an exponential backoff.

ashku:
Some point : Only one session is created although the code attempts to create session for 10 iteration and I do not think it is cached. So when you create a session do following , within  createSession() method

1. Check if context is valid/not nil if not load HSM library again by calling 
     ctx := pkcs11.New(lib) , then

1. Get a session  by calling
    session =  ctx.OpenSession(slot, pkcs11.CKF_SERIAL_SESSION|pkcs11.CKF_RW_SESSION)

1. You need to log in using 
     err = ctx.Login(session, pkcs11.CKU_USER, pin)
      
      if  user is already logged in , you'll go to  following  path
	if err != nil {
		if err != pkcs11.Error(pkcs11.CKR_USER_ALREADY_LOGGED_IN).

I see you are using all 3 steps ,  so that should be good. Just make sure that you execute step 1 in case csp.ctx is nil.

bccsp/pkcs11/pkcs11.go
sykesm: This looks like a copy and paste from loadLib. Makes me think that function should be reworked to call this one.

bccsp/pkcs11/pkcs11.go
```diff
@@ -84,25 +84,44 @@ func loadLib(lib, pin, label string) (*pkcs11.Ctx, uint, *pkcs11.SessionHandle,
 func (csp *impl) getSession() (session pkcs11.SessionHandle) {
 	select {
 	case session = <-csp.sessions:
-		logger.Debugf("Reusing existing pkcs11 session %+v on slot %d\n", session, csp.slot)
+		_, err := csp.ctx.GetSessionInfo(session)
```
sykesm, line 77: So, we failed to get the session info and assume that means the session is bad. I'll buy that.

However, if we think it's bad, why aren't we attempting to close it to ensure that any resources being held on the cgo side of this are getting released?

sykesm (changes requested):
In general, we start by making changes to master and then roll them back to older releases. I did not see this change on master and would expect to see it land there first.

Beyond that:
- I'd like to see the resources acquired when opening sessions returned by closing the sessions on the fail path.
- I'd like to understand why calling "OpenSession" ten times in a tight loop is going to give us the desired behavior.
- I'd like to see the code (if it continues to be duplicated) that is common between `loadLib` and `createSession` consolidated. I believe that's as simple as reordering the code in `loadLib` and calling `createSession` from there.
- While I didn't call it out inline, there's a lot of debug in here on happy paths and I'm not sure what it's buying. The "Logging in..." message in particular doesn't seem very useful.

sykesm:
> * I'd like to see the code (if it continues to be duplicated) that is common between `loadLib` and `createSession` consolidated. I believe that's as simple as reordering the code in `loadLib` and calling `createSession` from there.

In fairness, it looks like that code was present before your changes. While I don't like it, I can't hold it against you... :)

bccsp/pkcs11/pkcs11.go
mrshah-at-work: @sykesm thanks for the feedback. I did not intent to change the functionality of creation of session. So I retained the code as is, just moved it to a new function.

The intent of this PR is to check if a session is invalid and if we should be getting a new one.

mrshah-at-work: This is not a copy and paste from loadLib. It is the `default` case that was present already in `GetSession` function to create a new session when there were no cached sessions available.

bccsp/pkcs11/pkcs11.go
mrshah-at-work: I agree, I will add CloseSession before creating a new session.

mrshah-at-work:
> In general, we start by making changes to master and then roll them back to older releases. I did not see this change on master and would expect to see it land there first.

I apologize. I will make a PR on master. You can hold off this one until we get it in on master. As we are using 1.4, we made the change here.

mrshah-at-work:
> I'd like to understand why calling "OpenSession" ten times in a tight loop is going to give us the desired behavior.

This was the behavior before I made the changes. The intent of the PR is not to change the behavior of OpenSession, so I do not want to touch that.

mrshah-at-work:
> I'd like to see the code (if it continues to be duplicated) that is common between loadLib and createSession consolidated. I believe that's as simple as reordering the code in loadLib and calling createSession from there.

I agree. Sorry I missed it. I will make the change in loadLib

mrshah-at-work:
> While I didn't call it out inline, there's a lot of debug in here on happy paths and I'm not sure what it's buying. The "Logging in..." message in particular doesn't seem very useful.

I agree, that was our personal debug print that we missed. I will remove the "Logging in..." one. I was dinged on removing one of the existing happy path debug prints, so I will leave the others until you point out the ones that you want to remove.

sykesm (commented):
Change works for me but I still haven't seen it land in master.

As an aside, someone really needs to follow-up on this and do some cleanup on logging and on the busy-wait looping on session creation. Neither of those were introduced here.

mrshah-at-work:
> Change works for me but I still haven't seen it land in master.
> 
> As an aside, someone really needs to follow-up on this and do some cleanup on logging and on the busy-wait looping on session creation. Neither of those were introduced here.

Thanks. I was waiting on everything to look good in this one before I make another PR on master. I can make it now.

lindluni:
> Change works for me but I still haven't seen it land in master.
> 
> As an aside, someone really needs to follow-up on this and do some cleanup on logging and on the busy-wait looping on session creation. Neither of those were introduced here.

I opened: FAB-17727 and FAB-17729 as follow ons (jira is acting up so I can't  add components)

mrshah-at-work:
@sykesm I want to bring to your attention that as [this change of adding altId](https://github.com/hyperledger/fabric/commit/afbc42f5b768797947d02f8ec2f6e88b0d490da5) does not exist in master, I could not cherry pick commits from this PR into the [PR for master branch](https://github.com/hyperledger/fabric/pull/1047). Looks like that change was merged only in release-1.4.

The reason I am bringing this to your attention is that I had to make some code changes in the PR for master branch, so you might want to give it a deep review again. Thanks!

mastersingh24 (approved):
This looks ok to me.  @sykesm are you happy to merge this?

sykesm (approved):

PR #1024, opened 2020-04-08
[LEDGER] replace Metadata in query API with Limit and Bookmark

cendhu:
#### Type of change

#### Description

1. The `GetStateRangeScanIteratorWithMetadata()` at both `stateleveldb` and `statecouchdb` receive metadata such as pagination limit. 
2. The `ExecuteQueryWithMetadata()` at `statecouchdb` receives metadata such as the pagination bookmark and pagination limit.
3. The `ValidateRangeMetadata()` and `ValidateQueryMetadata()` validate the received metadata for the respective query.

However, the term `Metadata` is confusing with the `statemetadata` which stores the key-based endorsement. The API `SetXXXMetadata()` and `GetXXXMetadata()` are associated with the per state metadata. To avoid this confusion, we replace `Metadata` in range query and execute query API with `Pagination` to be explicit. As we have removed the metadata, we also removed `ValidateRangeMetadata()` and `ValidateQueryMetadata() `as they are not needed anymore.

`GetStateRangeScanIteratorWithMetadata()` -> `GetStateRangeScanIteratorWithPagination()`
`ExecuteQueryWithMetadata()` -> `ExecuteQueryWithPagination()`

#### Related issues

[FAB-17721](https://jira.hyperledger.org/browse/FAB-17721)


manish-sethi (changes requested):
1) I certainly find using "Options" instead of "Metadata" more meaningful but inconsistent naming causes more confusion. I would prefer to apply this change to this function `ExecuteQueryWithMetadata` as well.

2) Moving a tiny piece of code to internal package because it is used in two packages is not that useful - particularly, when you have a similar function for rich queries still sitting `statecouchdb` package. I find keeping the related code as close as possible more useful. 
wdyt?

cendhu:
1. While I was changing the code, I did change `Metadata` to `Options` for the `ExecuteQuery` as well. Then, I reverted the changes thinking that it can be done in a separate PR and concentrate on range queries only in this PR. Hence, let's do it in the next PR. 

2. If we need to keep it close, we need to duplicate the code. I don't think that either `stateleveldb` or `statecouchdb` should import `statedb` for a utility function if it cannot import an internal package. In fact, in my initial changes, I tried to remove the `map[string]interface{}` arguments from the range query as we are only passing the `limit` for now and extending this function later would not break any backward compatibility. IMO, it would be good to code for a feature that we currently support today. If we need to remove the introduced internal package, I would prefer to replace the map argument with the `limit` parameter and avoid this validation code altogether.  However, in this case, I am not sure whether we can name it `Options` instead `Limit` makes more sense. Otherwise, we can introduce a new struct named `queryOptions` and introduce `bookmark` and `limit`. Then, use it for both types of queries.

What do you think?

cendhu:
@Mergifyio rebase

denyeart:
I would prefer we simply change "WithMetadata" to "WithPagination" or "WithOptions". I think that would be more expressive, cleaner, and a smaller change, than saying things like "WithBookmarkAndLimit".  What if a database has 10 query options, you wouldn't want the API itself to mention all 10 of the query options.

core/ledger/kvledger/txmgmt/statedb/statecouchdb/statecouchdb.go
```diff
@@ -439,37 +438,21 @@ func (vdb *VersionedDB) GetStateMultipleKeys(namespace string, keys []string) ([
...
-		if limitOption, ok := metadata[optionLimit]; ok {
-			requestedLimit = limitOption.(int32)
-		}
-	}
 	db, err := vdb.getNamespaceDBHandle(namespace)
 	if err != nil {
 		return nil, err
 	}
-	return newQueryScanner(namespace, db, "", internalQueryLimit, requestedLimit, "", startKey, endKey)
+	return newQueryScanner(namespace, db, "", internalQueryLimit, limit, "", startKey, endKey)
```
denyeart: This was changed to "requestedLimit" on purpose.  There were too many 'limits' in the function, so it was more clear to give each limit a more meaningful name, e.g. internalQueryLimit and requestedLimit.

core/ledger/kvledger/txmgmt/statedb/statecouchdb/statecouchdb.go
```diff
@@ -570,7 +538,7 @@ func (vdb *VersionedDB) ExecuteQueryWithMetadata(namespace, query string, metada
 	if err != nil {
 		return nil, err
 	}
-	return newQueryScanner(namespace, db, queryString, internalQueryLimit, requestedLimit, bookmark, "", "")
+	return newQueryScanner(namespace, db, queryString, internalQueryLimit, limit, bookmark, "", "")
```
denyeart: same comment

core/ledger/kvledger/txmgmt/statedb/statecouchdb/statecouchdb.go
cendhu: Agree. Will change to `pageSize`.

cendhu:
> I would prefer we simply change "WithMetadata" to "WithPagination" or "WithOptions". I think that would be more expressive, cleaner, and a smaller change, than saying things like "WithBookmarkAndLimit". What if a database has 10 query options, you wouldn't want the API itself to mention all 10 of the query options.

I can use `WithPagination` then. We added this pagination support around May 2018 and we passed a map instead of the parameters directly. We then had separate map validation code -- one for the executeQuery and another for the range query. The range query validation code was placed in the `statedb` package and imported by both `statecouchdb` and `stateleveldb` packages to avoid code duplication. We did all this to support that 10 query options. In two years, we never supported any other options other than the pagination option. The map validation code and related test code are not justified. This PR has removed them and reduced 200 LOC. When we support that 10 query options, we can pass an option. For now, I can change `WithMetadata` to `WithPagination` instead of `WithLimit` and `WithBookmarkAndLimit`.

core/chaincode/handler.go
```diff
@@ -743,8 +743,9 @@ func (h *Handler) HandleGetStateByRange(msg *pb.ChaincodeMessage, txContext *Tra
 				startKey = metadata.Bookmark
 			}
 		}
-		rangeIter, err = txContext.TXSimulator.GetStateRangeScanIteratorWithMetadata(namespaceID,
-			startKey, getStateByRange.EndKey, paginationInfo)
+		pageSize := paginationInfo["limit"].(int32)
```
manish-sethi: No need to create a map and then lookup for a key...simply extract `metadata.PageSize` inside the above if loop?

core/chaincode/handler.go
```diff
@@ -858,8 +859,10 @@ func (h *Handler) HandleGetQueryResult(msg *pb.ChaincodeMessage, txContext *Tran
 			return nil, err
 		}
 		isPaginated = true
-		executeIter, err = txContext.TXSimulator.ExecuteQueryWithMetadata(namespaceID,
-			getQueryResult.Query, paginationInfo)
+		bookmark := paginationInfo["bookmark"].(string)
+		pageSize := paginationInfo["limit"].(int32)
```
manish-sethi: same comment as above

manish-sethi (changes requested):
A few comment, mainly on the chaincode side.

manish-sethi:
> > I would prefer we simply change "WithMetadata" to "WithPagination" or "WithOptions". I think that would be more expressive, cleaner, and a smaller change, than saying things like "WithBookmarkAndLimit". What if a database has 10 query options, you wouldn't want the API itself to mention all 10 of the query options.
> 
> I can use `WithPagination` then. We added this pagination support around May 2018 and we passed a map instead of the parameters directly. We then had separate map validation code -- one for the executeQuery and another for the range query. The range query validation code was placed in the `statedb` package and imported by both `statecouchdb` and `stateleveldb` packages to avoid code duplication. We did all this to support that 10 query options. In two years, we never supported any other options other than the pagination option. The map validation code and related test code are not justified. This PR has removed them and reduced 200 LOC. When we support that 10 query options, we can pass an option. For now, I can change `WithMetadata` to `WithPagination` instead of `WithLimit` and `WithBookmarkAndLimit`.

@denyeart, @cendhu  -  IMO, there is no need to have two variations for the queries. Simply the basic APIs (i.e.,  neither `WithPagination` nor `WithMetadata`/`WithOptions`) can take additional options. Even, in the current code the Basic API simply call the other one with default options. We cannot help it at the chaincode layer for compatibility reason but that should not be dragged to the last mile in the code.

core/ledger/kvledger/txmgmt/statedb/stateleveldb/stateleveldb.go
```diff
@@ -130,44 +130,28 @@ func (vdb *versionedDB) GetStateMultipleKeys(namespace string, keys []string) ([
 // startKey is inclusive
 // endKey is exclusive
 func (vdb *versionedDB) GetStateRangeScanIterator(namespace string, startKey string, endKey string) (statedb.ResultsIterator, error) {
-	return vdb.GetStateRangeScanIteratorWithMetadata(namespace, startKey, endKey, nil)
+	// limit = 0 denotes unlimited page size
```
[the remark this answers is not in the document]
cendhu: Done.

core/chaincode/handler.go
cendhu: Nice suggestion. Removed some 35 LOC.

cendhu:
@manish-sethi Valid point. I agree with you. I have created a separate JIRA [FAB-17754](https://jira.hyperledger.org/browse/FAB-17754) to address this once we decide on the following. To define and import `option` structs (one per query type--range and executeQuery), we need to agree on a location where we can keep all types exported by the ledger package to be used by both non-ledger packages and the ledger package. As it might take a bit longer to decide on that, we can merge this PR for now.

cendhu:
@btl5037 2 second wait time for CouchDB index creation does not seem to help 😞 We might need to increase the wait time a bit. I hit the flakes twice in this PR.

lindluni:
> @btl5037 2 second wait time for CouchDB index creation does not seem to help 😞 We might need to increase the wait time a bit. I hit the flakes twice in this PR. 

Yea. We hit it in 3 other PRs today as well :(

cendhu:
@btl5037 Do we use SSD or HDD? These tests run in parallel right? If it is SSD, 2 seconds should be adequate and surprising why it fails. If it is HDD, we need to increase the time. Btw, what is the allocated #vCPUs? The CouchDB also consumes 3x higher CPU than GoLevelDB. Even with SSD, if we have allocated lower #vCPUs and run tests in parallel, I am sure that 2 seconds is not adequate.

lindluni:
2vCPU and 4GB running on SSD

core/ledger/kvledger/txmgmt/statedb/statecouchdb/statecouchdb.go
```diff
@@ -439,37 +438,21 @@ func (vdb *VersionedDB) GetStateMultipleKeys(namespace string, keys []string) ([
...
+// GetStateRangeScanIteratorWithPagination implements method in VersionedDB interface
 // startKey is inclusive
 // endKey is exclusive
-// metadata contains a map of additional query options
-func (vdb *VersionedDB) GetStateRangeScanIteratorWithMetadata(namespace string, startKey string, endKey string, metadata map[string]interface{}) (statedb.QueryResultsIterator, error) {
-	logger.Debugf("Entering GetStateRangeScanIteratorWithMetadata  namespace: %s  startKey: %s  endKey: %s  metadata: %v", namespace, startKey, endKey, metadata)
-	// Get the internalQueryLimit from core.yaml
+// pageSize limits the number of results returned
+func (vdb *VersionedDB) GetStateRangeScanIteratorWithPagination(namespace string, startKey string, endKey string, pageSize int32) (statedb.QueryResultsIterator, error) {
+	logger.Debugf("Entering GetStateRangeScanIteratorWithOptions namespace: %s  startKey: %s  endKey: %s  options: %d", namespace, startKey, endKey, pageSize)
```
manish-sethi: out of sync with new function name and parameter names

core/ledger/kvledger/txmgmt/statedb/statecouchdb/statecouchdb.go
```diff
@@ -534,33 +517,18 @@ func isCouchInternalKey(key string) bool {
...
 	}
 	return queryResult, nil
 }
 
-// ExecuteQueryWithMetadata implements method in VersionedDB interface
-func (vdb *VersionedDB) ExecuteQueryWithMetadata(namespace, query string, metadata map[string]interface{}) (statedb.QueryResultsIterator, error) {
-	logger.Debugf("Entering ExecuteQueryWithMetadata  namespace: %s,  query: %s,  metadata: %v", namespace, query, metadata)
+// ExecuteQueryWithPagination implements method in VersionedDB interface
+func (vdb *VersionedDB) ExecuteQueryWithPagination(namespace, query, bookmark string, pageSize int32) (statedb.QueryResultsIterator, error) {
+	logger.Debugf("Entering ExecuteQueryWithMetadata  namespace: %s,  query: %s,  metadata: %d", namespace, query, pageSize)
```
manish-sethi: out of sync with new function name and parameter names

manish-sethi (dismissed):
LGTM. A couple nit-picking comments though, that you may fix when resolving conflicts.

core/ledger/kvledger/txmgmt/statedb/statecouchdb/statecouchdb.go
cendhu: Done.

core/ledger/kvledger/txmgmt/statedb/statecouchdb/statecouchdb.go
cendhu: Done.

manish-sethi (approved):

PR #1027, opened 2020-04-08
[FABB-148, FABB-149]

pamandrejko:
Added HSM updates for AWS HSM support

New `bccsp` parameters:
- 'Immutable`
- 'AltId` (required for AWS HSM)

And considerations around using them.

#### Type of change

- Documentation update

### DIFF
--- a/docs/source/hsm.md
+++ b/docs/source/hsm.md
@@ -1,0 +1,161 @@
+# Using a Hardware Security Module (HSM)
+
+You can use a Hardware Security Module (HSM) to generate and store the private
+keys used by your Fabric nodes. An HSM protects your private keys and handles
+cryptographic operations, which allows your peers and ordering nodes to sign and
+endorse transactions without exposing their private keys. Currently, Fabric only
+supports the PKCS11 standard to communicate with an HSM.
+
+## Configuring an HSM
+
+To use an HSM with your Fabric node, you need to update the BCCSP (Crypto Service
+Provider) section of the node configuration file such as core.yaml or
+orderer.yaml. In BCCSP section, you need to select PKCS11 as the provider and
+provide the path to the PKCS11 library that you would like to use. You also need
+to provide the label and pin of the token that you created for your cryptographic
+operations. You can use one token to generate and store multiple keys.
+
+The prebuilt Hyperledger Fabric Docker images are not enabled to use PKCS11. If
+you are deploying Fabric using docker, you need to build your own images and
+enable PKCS11 using the following command:
+```
+make docker GO_TAGS=pkcs11
+```
+You also need to ensure that the PKCS11 library is available to be used by the
+node by installing it or mounting it inside the container.
+
+### Example
+
+The following example demonstrates how to configure a Fabric node to use an HSM.
+
+First, you will need to install an implementation of the PKCS11 interface. This
+example uses the [softhsm](https://github.com/opendnssec/SoftHSMv2) open source
+implementation. After downloading and configuring softhsm, you will need to set
+the SOFTHSM2_CONF environment variable to point to the softhsm2 configuration
+file.
+
+You can then use softhsm to create the token that will handle the cryptographic
+operations of your Fabric node inside an HSM slot. In this example, we create a
+token labelled "fabric" and set the pin to "71811222". After you have created
+the token, update the configuration file to use PKCS11 and your token as the
+crypto service provider. You can find an example BCCSP section below:
+
+```
+#############################################################################
+# BCCSP (BlockChain Crypto Service Provider) section is used to select which
+# crypto library implementation to use
+#############################################################################
+bccsp:
+  default: PKCS11
+  pkcs11:
+    Library: /etc/hyperledger/fabric/libsofthsm2.so
+    Pin: 71811222
+    Label: fabric
+    hash: SHA2
+    security: 256
+    Immutable: false
+```
+
+By default, when private keys are generated using the HSM, the private key is mutable, meaning PKCS11 private key  attributes can be changed after the key is generated. Setting `Immutable` to `true` means that the private key attributes cannot be altered after key generation. Before you configure immutability by setting `Immutable: true`, ensure that PKCS11 object cloning is supported by the HSM.
+
+If you are using AWS HSM there is an additional step required:
+
+- Add the parameter, `AltId` to the `pkcs11` section of the `bccsp` block. When AWS HSM is being used, this parameter is used to assign a unique value for the Subject Key Identifier (SKI). Create a long secure string outside of Fabric and assign it to the `AltId` parameter. For example:
+
+  ```
+  #############################################################################
+  # BCCSP (BlockChain Crypto Service Provider) section is used to select which
+  # crypto library implementation to use
+  #############################################################################
+  bccsp:
+    default: PKCS11
+    pkcs11:
+      Library: /etc/hyperledger/fabric/libsofthsm2.so
+      Pin: 71811222
+      Label: fabric
+      hash: SHA2
+      security: 256
+      Immutable: false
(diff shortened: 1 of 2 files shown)


pamandrejko:
@ashku

docs/source/hsm.md
```diff
@@ -0,0 +1,161 @@
...
+  pkcs11:
+    Library: /etc/hyperledger/fabric/libsofthsm2.so
+    Pin: 71811222
+    Label: fabric
+    hash: SHA2
+    security: 256
+    Immutable: false
+```
+
+By default, when private keys are generated using the HSM, the private key is mutable, meaning PKCS11 private key  attributes can be changed. Setting `Immutable` to `true` means that the private key attributes cannot be altered. Before you configure immutability by setting `Immutable: true`, ensure that PKCS11 object cloning is supported by the HSM.
```
ashku:
```suggestion
By default, when private keys are generated using the HSM, the private key is mutable, meaning PKCS11 private key  attributes can be changed after it is generated. Setting `Immutable` to `true` means that the private key attributes cannot be altered after generation. Before you configure immutability by setting `Immutable: true`, ensure that PKCS11 object cloning is supported by the HSM.
```

ashku (changes requested):
small change @pamandrejko

docs/source/hsm.md
pamandrejko: Added.

ashku (dismissed):
LGTM

docs/source/hsm.md
```diff
@@ -0,0 +1,161 @@
...
+bccsp:
+  default: PKCS11
+  pkcs11:
+    Library: /etc/hyperledger/fabric/libsofthsm2.so
+    Pin: 71811222
+    Label: fabric
+    hash: SHA2
+    security: 256
+    Immutable: false
+```
```
denyeart, line 57: In v1.4.x, there is also properties for FileKeyStore and KeyStore. These were removed in v2.0 since they are not needed. But it is unclear if they have to be configured in v1.4.x... see:
https://github.com/hyperledger/fabric/blob/release-1.4/sampleconfig/core.yaml#L298-L299

denyeart:
For AWS HSM, I thought I heard a proxy was also required to be configured.

docs/source/hsm.md
```diff
@@ -0,0 +1,161 @@
...
+
+By default, when private keys are generated using the HSM, the private key is mutable, meaning PKCS11 private key  attributes can be changed after the key is generated. Setting `Immutable` to `true` means that the private key attributes cannot be altered after key generation. Before you configure immutability by setting `Immutable: true`, ensure that PKCS11 object cloning is supported by the HSM.
+
+If you are using AWS HSM there is an additional step required:
+
+- Add the parameter, `AltId` to the `pkcs11` section of the `bccsp` block. Create a long secure string outside of Fabric and assign it to the `AltId` parameter. For example:
+
```
denyeart, line 64: Can we mention why AltId is required and what it is used for?

pamandrejko:
@denyeart Thanks for all the comments!
Given the time constraint here - would it be reasonable to address all the issues not related to FABB-148 and FABB-149 via a separate PR? Also, so they can get appropriately addressed in both the v1.4 and 2.x branches.

docs/source/hsm.md
```diff
@@ -0,0 +1,161 @@
...
+3. Edit the Fabric CA client config file or environment variables to use your
+HSM as the crypto service provider. Then for each node, use the Fabric CA client
+to generate the component MSP folder by enrolling against the node identity. The
+enroll command will generate the private key inside your HSM.
+
+3. Update the BCCSP section of the peer or orderer configuration file to use
+PKCS11 and your token as the crypto service provider. Point to the MSP that was
+generated using the Fabric CA client. Once it is deployed, the peer or orderer
+node will be able sign and endorse transactions with the private key protected by
+the HSM.
```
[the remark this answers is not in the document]
ashku: Here is the flow : 

1. You as Enterprise has HSM Server , say , Gemalto. 
2. As optional step , you can have dedicated partition on HSM for Peer/Orderer
3. From your Fabric CA module , you load up HSM Client module , server for which is configured in step 1 and you generate key and sent request to FABRIC CA server to give signed certificate back to you.
4. Step 3 creates private key for you , which may be peer of orderer. This Private Key is stored on HSM. 

Above explains linkage between Private Key generation on HSM using Fabric CA client  and Orderer/Peer.

5. (Extra step for clarity) You configure PKCS11 of BCCSP for Message Signing and Verification. 

@denyeart , please let me know , if you have more question(s).

docs/source/hsm.md
```diff
@@ -0,0 +1,161 @@
...
+
+You can also use environment variables to override the relevant fields of the
+configuration file. If you are connecting to an HSM using the Fabric CA server,
+you need to set the following environment variables:
+
+```
+FABRIC_CA_SERVER_BCCSP_DEFAULT=PKCS11
+FABRIC_CA_SERVER_BCCSP_PKCS11_LIBRARY=/etc/hyperledger/fabric/libsofthsm2.so
+FABRIC_CA_SERVER_BCCSP_PKCS11_PIN=71811222
+FABRIC_CA_SERVER_BCCSP_PKCS11_LABEL=fabric
```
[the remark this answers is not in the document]
ashku: Same pattern for Peer and Orderer should work.

docs/source/hsm.md
ashku: Here is flow : 

1. Key pair is generated using PKCS11 interface.
2. SKI(Subject Key Identifier) is calculated based on KeyPair generated in step 1.
3. SKI value is set to PKCS11 CKA_ID field attribute of Public and Private key.

Step 3 is not supported in AWS. So , once crypto object is generated , you cannot modify it.

To circumvent problem with step 3 and for minimal code impact , a new variable was created named AltId. (The change was put by luc). AltId is independent of Private key , it is being generated before key pair is generated. AltId is being set for CKA_ID field of Private Key. @denyeart , please let me know , if you have more comments.

docs/source/hsm.md
ashku: FileKeyStore and KeyStore is used for Software based BCCSP and hence does not play in PKCS11 configuration. For that reason , it should be mentioned here IMO.

denyeart: I assume you mean 'should not' be mentioned here.
So is the suggestion to simply leave these fields blank in core.yaml for people using HSM on 1.4?

docs/source/hsm.md
denyeart: Makes sense, please help Pam put that into a couple doc sentences.

docs/source/hsm.md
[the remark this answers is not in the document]
denyeart: Makes sense, please help Pam clarify this in the existing bullets.

denyeart:
> @denyeart Thanks for all the comments!
> Given the time constraint here - would it be reasonable to address all the issues not related to FABB-148 and FABB-149 via a separate PR? Also, so they can get appropriately addressed in both the v1.4 and 2.x branches.

Yes, I'm ok with all the comments being addressed in subsequent PR. Therefore I'll approve with that understanding...

denyeart (dismissed):

docs/source/hsm.md
pamandrejko: Confirmed offline with Ash that he meant `FileKeyStore and KeyStore is used for Software based BCCSP and hence does not play in PKCS11 configuration. For that reason , it should not be mentioned here IMO.`

docs/source/hsm.md
pamandrejko: Add the parameter, `AltId` to the `pkcs11` section of the `bccsp` block. When AWS HSM is being used, this parameter is used to assign a unique value to the CKA_Id attribute of the private and public key.

pamandrejko:
I opened this jira to address the remaining comments https://jira.hyperledger.org/browse/FAB-17732

denyeart (approved):

PR #1031, opened 2020-04-08
Improve configtxgen test coverage for unhappy paths

betasve:
Signed-off-by: Svetoslav Blyahov <svetlio.blyahoff@gmail.com>

Adding a simple test for an unhappy path in configtxgen package to improve test coverage a bit

#### Type of change
- Test update

#### Description

Adding a test for covering the unhappy path of passing an empty string, instead of a config path in:
- configtxgen
  doOutputBlock
  doInspectBlock

Basically following some low-hanging fruits for improving test coverage in this part of the code base I've been researching recently. 

Aim in this PR is to see if such a minor improvements in test coverage have a good reception. If so I will continue with some follow ups in this test file and other test files related to that package. 

Thanks in advance and looking forward to hearing your comments and recommendations.

### DIFF
--- a/cmd/configtxgen/main_test.go
+++ b/cmd/configtxgen/main_test.go
@@ -46,6 +46,13 @@ func TestInspectBlock(t *testing.T) {
 	assert.NoError(t, doInspectBlock(blockDest), "Good block inspection request")
 }
 
+func TestInspectBlockErr(t *testing.T) {
+	config := genesisconfig.Load(genesisconfig.SampleInsecureSoloProfile, configtest.GetDevConfigDir())
+
+	assert.Error(t, doOutputBlock(config, "foo", ""), "Error writing genesis block:")
+	assert.Error(t, doInspectBlock(""), "Could not read block")
+}
+
 func TestMissingOrdererSection(t *testing.T) {
 	blockDest := filepath.Join(tmpDir, "block")
 


caod123 (approved):

wlahti:
Hello and thanks for contributing! My first thought was to make sure we're actually testing that the error matches the error we expect to receive using assert.EqualError. 

However, I noticed none of this file does that so, for now, this is probably fine as is. If you'd like to do some follow-up work, perhaps you can start there, for these errors and the others in the file.

betasve:
> Hello and thanks for contributing! My first thought was to make sure we're actually testing that the error matches the error we expect to receive using assert.EqualError.
> 
> However, I noticed none of this file does that so, for now, this is probably fine as is. If you'd like to do some follow-up work, perhaps you can start there, for these errors and the others in the file.

Hello, thanks for your reply. I will definitely do that. I was not sure if you consider testing for the exact error message a good practice or a bad one. 

As far as I was able to see [assert.EqualError](https://github.com/stretchr/testify/blob/master/assert/assertions.go#L1353) API, it has `errString` so using it I will again test the error message itself, am I right? Or you have something else in mind?

wlahti:
Yes, we always like to test that an error is what we expect it to be. While not critical in this particular codepath, it can help ensure that the functionality on error paths doesn't change unless we intend it to.

And yes, assert.EqualError tests the error message itself. You can replace the calls to assert.Error with it and it will test that both an error has occurred and that the error messages matches what you expect.

PR #1032, opened 2020-04-08
Add upgrade steps and troubleshooting to deployCC tutorial

nikhil550:
Signed-off-by: NIKHIL E GUPTA <negupta@us.ibm.com>

#### Type of change

- Documentation update

#### Description

Add chaincode upgrade steps to Deploying a chaincode tutorial to help users further understand the fabric lifecycle process. Also adding a troubleshooting section with common problems I have seen and experienced.

### DIFF
--- a/docs/source/deploy_chaincode.md
+++ b/docs/source/deploy_chaincode.md
@@ -144,11 +144,6 @@ To confirm that you are able to use the `peer` CLI, check the version of the bin
 peer version
 ```
 
-You also need to set the `CORE_PEER_MSPCONFIGPATH` to the location of the MSP folder of an admin or client user. Run the following command to establish the Org1 admin as the identity we're using.
-```
-export CORE_PEER_MSPCONFIGPATH=${PWD}/organizations/peerOrganizations/org1.example.com/users/Admin@org1.example.com/msp
-```
-
 You can now create the chaincode package using the [peer lifecycle chaincode package](commands/peerlifecycle.html#peer-lifecycle-chaincode-package) command:
 ```
 peer lifecycle chaincode package fabcar.tar.gz --path ../chaincode/fabcar/go/ --lang golang --label fabcar_1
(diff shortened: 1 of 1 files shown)


docs/source/deploy_chaincode.md
```diff
@@ -461,6 +461,130 @@ The response to the query should be the following list of cars:
...
+429dae4757ba        hyperledger/fabric-peer:latest                                                                                                                                          "peer node start"        About an hour ago   Up About an hour    7051/tcp, 0.0.0.0:9051->9051/tcp   peer0.org2.example.com
+7de5d19400e6        hyperledger/fabric-orderer:latest                                                                                                                                       "orderer"                About an hour ago   Up About an hour    0.0.0.0:7050->7050/tcp             orderer.example.com
+```
+
+Because we used the `--init-required` flag, we need to initialize the upgraded chaincode before it can be used. The following command will initialize the new chaincode by passing the ``--isInit`` flag. Because our chaincode uses the Fabric contract API, we can target a query function that will not update the ledger.
```
denyeart: It does not logically make sense to require init, and then have the init call do nothing but a query. Init is intended to update the ledger with some initialization data. I realize that you probably did this to demonstrate the process for initializing, but doing an init with a noop function call will be confusing to the reader.

denyeart (commented):
Nice updates... just found one nit about init.

nikhil550:
@denyeart Thanks Dave, I will update the tutorial to initialize the chaincode by creating a new car

denyeart:
> @denyeart Thanks Dave, I will update the tutorial to initialize the chaincode by creating a new car

@nikhil550 
Actually, init is not a pattern that we want to encourage. Maybe it would be better to keep the chaincode as is, remove the init required from the command, and just mention in text that you could pass init required if the chaincode required it.

nikhil550:
@denyeart I was worried about legacy behavior, since older chaincodes required Init. But if we want to move users away from it, I can update the tutorial to mention it but not use it. I will also update the chaincode for developers topic to mention that init is not required if you use the new smart contract class, which had been on my list of things to do.

nikhil550:
@denyeart Removed the ``Init`` flags from the commands, but made a brief mention of it for legacy users:
If you are using the low level APIs provided by the Fabric Chaincode Shim API, you could pass the `--init-required` flag to the command above to request the execution of the Init function to initialize the chaincode. The first invoke of the chaincode would need to target the Init function and include the `--isInit` flag before you could use the other functions in the chaincode to interact with the ledger.

Also removed this sentence since it is no longer true in 2.1.0:
You also need to set the `CORE_PEER_MSPCONFIGPATH` to the location of the MSP folder of an admin or client user. Run the following command to establish the Org1 admin as the identity we're using.

denyeart:
@nikhil550 The Init stuff looks good now.

But I don't understand why CORE_PEER_MSPCONFIGPATH isn't required anymore... what changed?

nikhil550:
@denyeart  Requiring an MSP folder just to package the chaincode was seen as a bug, and it was removed in 2.1.0. I ran the tutorial without it and everything worked.

denyeart:
@nikhil550 Ah yes, I missed that it was removed only for the package step. Merging...

denyeart (approved):

nikhil550:

PR #1040, opened 2020-04-09
Clear {U,G}name fields in package tar entries

sykesm:
For historic reasons, entries in chaincode packages are explicitly created with a numeric uid/gid 500/500 ownership to help remove variability in packages but the uname/gname fields were not cleared.

This change clears the uname and gname fields and ensures directories and files have the same owner information after extraction and prevents another level of variability in the package.


MHBauer (approved):
straightforward change and test update.

lindluni (approved):

PR #1043, opened 2020-04-09
Omit go.{mod,sum} from pkg when not in module mode

sykesm:
In certain environments, it's possible to package chaincode that structured as a module from an active GOPATH. This often happens when the path provided to the package command is an import path resolvable from the GOPATH instead of a file system path.

If the package could be successfully built in the packaging environment using the import path, the chaincode dependencies would be calculated and packaged from the GOPATH for compilation as a traditional go package.

In this scenario, the `go.mod` would be included in the chaincode package as packaging always includes all non-hidden files in the top level folder of the import path.

On the server, the presence of the `go.mod` implies that the build process should execute in module mode. When the dependencies have been vendored in the module, the build uses `-mod=vendor` flag to indicate the module requirements should be satisfied from the vendor folder.
Unfortunately, since the chaincode dependencies were packaged using GOPATH mode instead of module mode, there are some metadata files missing from the vendor folder that are expected by the module mode build process.

To help prevent this from occurring, we will explicitly omit go.mod and go.sum from top level folder of chaincode that is not packaged in module mode.

Bug fix - FAB-17725

### DIFF
--- a/core/chaincode/platforms/golang/platform.go
+++ b/core/chaincode/platforms/golang/platform.go
@@ -290,7 +290,7 @@ func DescribeCode(path string) (*CodeDescriptor, error) {
 func describeGopath(importPath string) (*CodeDescriptor, error) {
 	output, err := exec.Command("go", "list", "-f", "{{.Dir}}", importPath).Output()
 	if err != nil {
-		return nil, err
+		return nil, wrapExitErr(err, "'go list' failed")
 	}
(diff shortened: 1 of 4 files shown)


lehors:
Happy to report that this does address my issue and with this change BYFN keeps working unchanged.
Thanks.

lindluni (approved):

PR #1044, opened 2020-04-09
[FAB-17716] Fix test flake due to too many requests for /protos.Deliver

wenjianqiao:

#### Type of change
- Test update

#### Description
Currently peer.limits.concurrency.deliverService value is set to 1
in IT core_template.go. This value is too low and may cause test flakes
because some tests have multiple goroutines sending requests to DeliverService.

This PR sets a higher value for all tests in IT. Meanwhile, it overwrites
the value to 1 in pvtdata_test.go where maximum 1 goroutine sends requests
to DeliverService.

#### Related issues
https://jira.hyperledger.org/browse/FAB-17716

### DIFF
--- a/integration/pvtdata/pvtdata_test.go
+++ b/integration/pvtdata/pvtdata_test.go
@@ -624,150 +624,6 @@ var _ bool = Describe("PrivateData", func() {
 			})
 		})
 
-		Describe("marble APIs invocation and private data delivery", func() {
-			// call marble APIs: getMarblesByRange, transferMarble, delete, getMarbleHash, getMarblePrivateDetailsHash and verify ACL Behavior
-			assertMarbleAPIs := func() {
-				eligiblePeer := network.Peer("Org2", "peer0")
-				ccName := testChaincode.Name
-
-				// Verifies marble private chaincode APIs: getMarblesByRange, transferMarble, delete
-
-				By("adding five marbles")
-				for i := 0; i < 5; i++ {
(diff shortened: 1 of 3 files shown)


integration/pvtdata/pvtdata_test.go
```diff
@@ -905,6 +761,180 @@ var _ bool = Describe("PrivateData", func() {
...
+				Path:              components.Build("github.com/hyperledger/fabric/integration/chaincode/marbles_private/cmd"),
+				Lang:              "binary",
+				PackageFile:       filepath.Join(network.RootDir, "marbles-pvtdata.tar.gz"),
+				Label:             "marbles-private-20",
+				SignaturePolicy:   `OR ('Org1MSP.member','Org2MSP.member', 'Org3MSP.member')`,
+				CollectionsConfig: collectionConfig("collections_config1.json"),
+				Sequence:          "1",
+			}
+
+			By("setting DeliverService limit to 1 to verify grpc concurrency control")
```
denyeart: It is not obvious how this test verifies the grpc concurrency control. From talking to Wenjian I now understand it is an implicit verification via the rest of the test, not something that is tested explicitly in this test.
It could be added to any test. I think that should be better explained in the By here.
Better yet, add this implicit verification to a simpler test, not a the more complex private data test, as it gets lost in the weeds and has nothing to do with private data testing per se.

wenjianqiao: Thank you Dave. Talked offline. Since IT has only 2 packages (lifecycle and pvtdata) calling DeliverService, I will keep the concurrency limit test in pvtdata, but add a comment explaining why we are doing it.

denyeart (approved):

denyeart:
Looks good now.

denyeart:

PR #1055, opened 2020-04-10
[FAB-17709] Support creating system channel genesis block

wuxuer:
Add a function to create a system channel genesis block.

#### Type of change

- New feature

#### Related issues
https://jira.hyperledger.org/browse/FAB-17709

### DIFF
--- a/pkg/configtx/config.go
+++ b/pkg/configtx/config.go
@@ -8,6 +8,9 @@ SPDX-License-Identifier: Apache-2.0
 package configtx
 
 import (
+	"bytes"
+	"crypto/sha256"
+	"encoding/hex"
 	"errors"
 	"fmt"
 	"strconv"
@@ -206,23 +209,49 @@ func NewCreateChannelTx(channelConfig Channel, channelID string) (*cb.Envelope, 
 	return env, nil
 }
 
+// NewSystemChannelGenesisBlock creates a genesis block using the provided consortiums and orderer
+// configuration and returns a block.
+func NewSystemChannelGenesisBlock(channelConfig Channel, channelID string) (*cb.Block, error) {
+	var err error
+
+	if channelID == "" {
+		return nil, errors.New("system channel ID is required")
+	}
+
+	systemChannelGroup, err := newSystemChannelGroup(channelConfig)
+	if err != nil {
+		return nil, fmt.Errorf("creating system channel group: %v", err)
+	}
+
+	block, err := newSystemChannelBlock(systemChannelGroup, channelID)
+	if err != nil {
+		return nil, fmt.Errorf("creating system channel genesis block: %v", err)
+	}
+
+	return block, nil
+}
+
 // newChannelGroup defines the root of the channel configuration.
 func newChannelGroup(channelConfig Channel) (*cb.ConfigGroup, error) {
 	var err error
 
 	channelGroup := newConfigGroup()
 
-	if channelConfig.Consortium != "" {
-		err = setValue(channelGroup, consortiumValue(channelConfig.Consortium), "")
-		if err != nil {
-			return nil, err
-		}
+	if channelConfig.Consortium == "" {
+		return nil, errors.New("consortium is not defined in channel config")
+	}
+
+	err = setValue(channelGroup, consortiumValue(channelConfig.Consortium), "")
+	if err != nil {
+		return nil, err
 	}
 
 	channelGroup.Groups[ApplicationGroupKey], err = newApplicationGroup(channelConfig.Application)
 	if err != nil {
 		return nil, fmt.Errorf("failed to create application group: %v", err)
 	}
+
+	channelGroup.ModPolicy = AdminsPolicyKey
 
 	return channelGroup, nil
 }
@@ -247,15 +276,6 @@ func newSystemChannelGroup(channelConfig Channel) (*cb.ConfigGroup, error) {
 		return nil, err
 	}
 
-	if channelConfig.Consortium == "" {
-		return nil, errors.New("consortium is not defined in channel config")
-	}
-
-	err = setValue(channelGroup, consortiumValue(channelConfig.Consortium), AdminsPolicyKey)
-	if err != nil {
-		return nil, err
-	}
-
 	if len(channelConfig.Capabilities) == 0 {
 		return nil, errors.New("capabilities is not defined in channel config")
 	}
@@ -276,6 +296,8 @@ func newSystemChannelGroup(channelConfig Channel) (*cb.ConfigGroup, error) {
 		return nil, err
 	}
 	channelGroup.Groups[ConsortiumsGroupKey] = consortiumsGroup
+
+	channelGroup.ModPolicy = AdminsPolicyKey
 
 	return channelGroup, nil
 }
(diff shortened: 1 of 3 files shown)


DereckLuo (dismissed):
+1 LGTM

pkg/config/config.go
```diff
@@ -210,6 +212,28 @@ func NewCreateChannelTx(channelConfig Channel, channelID string) (*cb.Envelope,
...
+	}
+
+	ct, err := newSystemChannelGroup(channelConfig)
+	if err != nil {
+		return nil, fmt.Errorf("creating system channel group: %v", err)
+	}
+
+	block, err := newSystemChannelBlock(ct, channelID)
+	if err != nil {
+		return nil, fmt.Errorf("creating system channel block: %v", err)
```
caod123: likewise to my comment on the function name, I think this error will need some rework too. We probably want to focus more on the fact that this is the genesis block as otherwise `creating system channel block` can be misconstrued for what this function is intended to be used for.

pkg/config/config.go
```diff
@@ -210,6 +212,28 @@ func NewCreateChannelTx(channelConfig Channel, channelID string) (*cb.Envelope,
 	return env, nil
 }
 
+// NewSystemChannelBlock creates a genesis block using the provided consortiums and orderer
+// configuration and returns a block.
+func NewSystemChannelBlock(channelConfig Channel, channelID string) (*cb.Block, error) {
```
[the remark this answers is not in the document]
caod123: the most clear API function would be `NewSystemChannelGenesisBlock` but that might be a bit wordy

pkg/config/config.go
```diff
@@ -210,6 +212,28 @@ func NewCreateChannelTx(channelConfig Channel, channelID string) (*cb.Envelope,
...
+	var err error
+
+	if channelID == "" {
+		return nil, errors.New("system channel ID is required")
+	}
+
+	ct, err := newSystemChannelGroup(channelConfig)
```
caod123: you may want to update this function and the `newChannelGroup` function for application channels. Application channels must have the `channelConfig.Consortium` specified whereas system channels do not (looks like this behavior on master is currently swapped). `newSystemChannelGroup` and `newChannelGroup` are also both missing a modPolicy: `channelGroup.ModPolicy = channelConfig.AdminsPolicyKey`

caod123 (changes requested):

pkg/config/config.go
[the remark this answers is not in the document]
wuxuer: In order to clarify the meaning of the function, I changed the function to be NewSystemChannelGenesisBlock .

pkg/config/config.go
wuxuer: I also added "genesis" in the error.

pkg/config/config.go
wuxuer: I fixed the two problems in fucntion newSystemChannelGroup and newChannelGroup. Please review whether it is correct.

pkg/config/config.go
```diff
@@ -343,15 +369,6 @@ func newSystemChannelGroup(channelConfig Channel) (*cb.ConfigGroup, error) {
 		return nil, err
 	}
 
-	if channelConfig.Consortium == "" {
-		return nil, errors.New("consortium is not defined in channel config")
-	}
```
caod123: This check probably should move to `newChannelGroup`

pkg/config/config_test.go
```diff
@@ -409,6 +410,741 @@ func TestNewCreateChannelTxFailure(t *testing.T) {
 	}
 }
 
+func TestSystemChannelGenesisBlock(t *testing.T) {
```
caod123: nit: `TestNewSystemChannelGenesisBlock`

pkg/config/config_test.go
```diff
@@ -702,7 +1438,7 @@ func baseSystemChannelProfile(t *testing.T) Channel {
 		Orderer:      baseSoloOrderer(t),
 		Capabilities: []string{"V2_0"},
 		Policies:     standardPolicies(),
-		Consortium:   "testconsortium",
+		Consortium:   "",
```
caod123: You might as well remove this line entirely

pkg/config/config_test.go
```diff
@@ -659,7 +1395,7 @@ func TestChannelConfiguration(t *testing.T) {
 				Orderer:      baseOrderer,
 				Capabilities: []string{"V2_0"},
 				Policies:     policies,
-				Consortium:   "testconsortium",
+				Consortium:   "",
```
caod123: You can probably remove this alongside line 1384

pkg/config/config_test.go
```diff
@@ -409,6 +410,741 @@ func TestNewCreateChannelTxFailure(t *testing.T) {
...
+	actualChannelHeader := &cb.ChannelHeader{}
+	err = proto.Unmarshal(actualPayload.Header.ChannelHeader, actualChannelHeader)
+	gt.Expect(err).NotTo(HaveOccurred())
+	expectedChannelHeader.Timestamp = actualChannelHeader.Timestamp
+	expectedChannelHeader.TxId = actualChannelHeader.TxId
+
+	gt.Expect(actualChannelHeader).To(Equal(expectedChannelHeader))
+}
+
+func TestSystemChannelGenesisBlockFailure(t *testing.T) {
```
caod123: nit: `TestNewSystemChannelGenesisBlockFailure`

caod123 (changes requested):
Some small requested changes

pkg/config/config.go
wuxuer: Yes, I have fixed it and added an unit test case.

pkg/config/config_test.go
wuxuer: Sorry for the following four nits. I have fixed them all.

pkg/config/config.go
```diff
@@ -217,13 +242,17 @@ func newChannelGroup(channelConfig Channel) (*cb.ConfigGroup, error) {
 		if err != nil {
 			return nil, err
 		}
+	} else {
+		return nil, errors.New("consortium is not defined in channel config")
```
C0rWin: given you return within else, might make sense to revert logic of the `if-then-else` to save on indentation and return on error case immediately?

pkg/config/config.go
```diff
@@ -364,6 +386,64 @@ func defaultConfigTemplate(channelConfig Channel) (*cb.ConfigGroup, error) {
...
+// newSystemChannelBlock generates a genesis block by the config group and system channel ID
+// the block num is always zero
+func newSystemChannelBlock(cg *cb.ConfigGroup, channelID string) (*cb.Block, error) {
+	payloadChannelHeader := channelHeader(cb.HeaderType_CONFIG, msgVersion, channelID, epoch)
+	nonce, err := newNonce()
+	if err != nil {
+		return nil, fmt.Errorf("try to get nonce: %v", err)
+	}
+	payloadSignatureHeader := &cb.SignatureHeader{Creator: nil, Nonce: nonce}
+	payloadChannelHeader.TxId = computeTxID(payloadSignatureHeader.Nonce, payloadSignatureHeader.Creator)
```
C0rWin: given that you generate `TxId` from fields of `SignatureHeader` doesn't it make sense to pass pointer to the struct itself rather than fields? also it's seems that creator is always a nil.

pkg/config/config.go
```diff
@@ -364,6 +386,64 @@ func defaultConfigTemplate(channelConfig Channel) (*cb.ConfigGroup, error) {
...
+	}
+	envelope := &cb.Envelope{Payload: envelopePayload, Signature: nil}
+	blockData, err := proto.Marshal(envelope)
+	if err != nil {
+		return nil, fmt.Errorf("marshaling envelope: %v", err)
+	}
+
+	block := newBlock(0, nil)
+	block.Data = &cb.BlockData{Data: [][]byte{blockData}}
+	block.Header.DataHash = blockDataHash(block.Data)
```
C0rWin: IMO it might make sense to encapsulate this initialization within `newBlock`

C0rWin (commented):
thanks for your contribution I have a few small nits.

pkg/config/config.go
wuxuer: Thank you very much for your suggestion. I have reverted the logic.

pkg/config/config.go
wuxuer: Function computeTxID should be a general function. If the input parameter is *SignatureHeader, the versatility of this function will be greatly reduced. 
The design of function newBlock  is also based on the same logic, which constructs a block with no data and no metadata.
In addition, these two functions are referenced in the “protoutil/blockutils.go” file.
These are some of my thoughts, I don't know if they are suitable.

C0rWin (approved):

caod123 (approved):

PR #1063, opened 2020-04-11
replace Metadata in query with Limit and Bookmark

cendhu:
#### Type of change

#### Description

1. The `GetStateRangeScanIteratorWithMetadata()` at both `stateleveldb` and `statecouchdb` receive metadata such as pagination limit. 
2. The `ExecuteQueryWithMetadata()` at `statecouchdb` receives metadata such as the pagination bookmark and pagination limit.
3. The `ValidateRangeMetadata()` and `ValidateQueryMetadata()` validate the received metadata for the respective query.

However, the term `Metadata` is confusing with the `statemetadata` which stores the key-based endorsement. The API `SetXXXMetadata()` and `GetXXXMetadata()` are associated with the per state metadata. To avoid this confusion, we replace `Metadata` in range query API with `Limit` and in the execute query API with `BookmarkAndLimit` to be explicit. As we have removed the metadata, we also removed `ValidateRangeMetadata()` and `ValidateQueryMetadata() `as they are not needed anymore.

`GetStateRangeScanIteratorWithMetadata()` -> `GetStateRangeScanIteratorWithLimit()`
`ExecuteQueryWithMetadata()` -> `ExecuteQueryWithBookmarkAndLimit()`

#### Related issues

[FAB-17721](https://jira.hyperledger.org/browse/FAB-17721)


denyeart:
Sorry accidentally closed and re-opened while adding this comment...

I would prefer we simply change "WithMetadata" to "WithPagination" or "WithOptions". I think that would be more expressive, cleaner, and a smaller change, than saying things like "WithBookmarkAndLimit".  What if a database has 10 query options, you wouldn't want the API itself to mention all 10 of the query options.

cendhu:
@btl5037 Even from a new branch, i get `race: limit on 8128 simultaneously alive goroutines is exceeded, dying`. Any suggestion on what should I do?

cendhu:
As matt has reverted the gossip race detection, other PR should pass. Hence, I am closing this PR.

PR #1064, opened 2020-04-11
Move to Go modules

sykesm:
The commits in this PR should move us to go modules.

These changes require go 1.14+ as they assume the presence of a `vendor` folder at the top level of a module will be used by default, without requiring the `-mod=vendor` flag. (This means we don't have to augment any of the flags we use when building.)

Future improvements should be made around our go tools as they get installed into the default `$GOPATH/bin` directory (`$HOME/go/bin` by default). These tools should really be installed to a location within the build tree and our scripts should be updated to explicitly reference the install location instead of relying on the $PATH.

Note to reviewers: I recommend reviewing commit by commit. The largest number of changes are in the first commit where dep was replaced with modules as it changed a fair bit of the vendored code by pulling in files that were excluded by dep and other miscellaneous things.

[asciicast](https://asciinema.org/a/320081)


lindluni:
I opened FAB-17735 as a follow on to do the tools work

scripts/check_deps.sh
```diff
@@ -1,5 +1,28 @@
...
+# check if we have unused requirements
+go mod tidy -modfile="${dep_tempdir}/go.mod"
+
+for f in go.mod go.sum; do
+    if ! diff -q "${fabric_dir}/$f" "${dep_tempdir}/$f"; then
+        echo "It appears $f is stale. Please run 'go mod tidy' and 'go mod vendor'."
+        diff -u "${fabric_dir}/$f" "${dep_tempdir}/$f"
+        exit 1
+    fi
+done
```
mastersingh24, line 28: Cool

mastersingh24 (approved):

mastersingh24:
will wait for someone from docs to review as well over the next day.  if not, will merge

mastersingh24:
This makes me happy!

nikhil550:
Docs look good. We should update the prerequisites page here in a follow on PR: https://hyperledger-fabric.readthedocs.io/en/latest/prereqs.html#go-programming-language

sykesm:
> Docs look good. We should update the prerequisites page here in a follow on PR: https://hyperledger-fabric.readthedocs.io/en/latest/prereqs.html#go-programming-language

Agreed; however, I believe that article should be focused on the requirements for app and chaincode development instead of "Fabric" development. While there is some overlap. they are distinct roles and the requirements differ quite a bit.

For the "Go" section, in particular, Go 1.14 is not needed for chaincode development and GOPATH or not is really up to the chaincode developer. This was the gist of the offline discussion I initiated last week.

@nikhil550, are you (docs experts) going to make the updates to the "prereqs" page that is part of "getting started" or is that something you need me to draft?

Thanks.

PR #1065, opened 2020-04-11
build: unit-test only needs CouchDB docker image

sykesm:
Stop building and puling docker images that are not needed by the unit tests. The only docker asset that is required is the couchDB image.

This slightly improves the execution time of unit tests. in CI.


cendhu:
please check whether #1072 fix the CouchDB ut flakes.

lindluni (approved):

PR #1072, opened 2020-04-13
Fix CouchDB UT flakes -- index creation

cendhu:
#### Type of change

- Bug fix

#### Description

As per CI statistics, `TestHandleChaincodeDeployErroneousIndexFile` has failed 152 times across 173 pipeline runs in the last 14 days.

The potential reason for this error could be that the CouchDB creates an index in the background and responds to the REST API immediately. Currently, we wait only for `100 milliseconds` to allow CouchDB to build the index before issuing the query. This `100 milliseconds` sleep in inadequate even for some local test environment used for the development. Hence, this PR increases wait time to `2 seconds` and allow ample time for CouchDB to build the index before issuing the query.

#### Additional details

As `assert.Eventually()` only accepts a function that returns a bool, we are unable to print the exact error message :disappointed: 

#### Related issues

[FAB-17737](https://jira.hyperledger.org/browse/FAB-17737)

### DIFF
--- a/core/ledger/kvledger/txmgmt/statedb/statecouchdb/statecouchdb_test.go
+++ b/core/ledger/kvledger/txmgmt/statedb/statecouchdb/statecouchdb_test.go
@@ -600,14 +600,15 @@ func TestHandleChaincodeDeploy(t *testing.T) {
 	}
 
 	indexCapable.ProcessIndexesForChaincodeDeploy("ns1", indexData)
-	//Sleep to allow time for index creation
-	time.Sleep(100 * time.Millisecond)
-	//Create a query with a sort
 	queryString = `{"selector":{"owner":"fred"}, "sort": [{"size": "desc"}]}`
-
-	//Query should complete without error
-	_, err = db.ExecuteQuery("ns1", queryString)
-	assert.NoError(t, err)
+	queryUsingIndex := func() bool {
+		_, err = db.ExecuteQuery("ns1", queryString)
+		if err != nil {
+			return false
+		}
+		return true
+	}
+	assert.Eventually(t, queryUsingIndex, 2*time.Second, 100*time.Millisecond, "error executing query with sort")
 
 	//Query namespace "ns2", index is only created in "ns1".  This should return an error.
 	_, err = db.ExecuteQuery("ns2", queryString)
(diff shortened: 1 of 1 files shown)


lindluni:
Hit known integration flake, rerunning tests, but otherwise looks like a welcome change

core/ledger/kvledger/txmgmt/statedb/statecouchdb/statecouchdb_test.go
```diff
@@ -600,14 +600,15 @@ func TestHandleChaincodeDeploy(t *testing.T) {
...
-	time.Sleep(100 * time.Millisecond)
-	//Create a query with a sort
 	queryString = `{"selector":{"owner":"fred"}, "sort": [{"size": "desc"}]}`
-
-	//Query should complete without error
-	_, err = db.ExecuteQuery("ns1", queryString)
-	assert.NoError(t, err)
+	queryUsingIndex := func() bool {
+		_, err = db.ExecuteQuery("ns1", queryString)
+		if err != nil {
```
denyeart, line 606: Can't you print the error here, so at least it will be in the test logs?

denyeart (commented):
Good change, one comment.

core/ledger/kvledger/txmgmt/statedb/statecouchdb/statecouchdb_test.go
cendhu: It can be done but we need to consider the following two points before deciding on that. 

1. The [assert.Eventually](https://github.com/stretchr/testify/blob/master/assert/assertions.go#L1628) calls the `queryUsingIndex` on every tick. If the tick is not configured properly especially when we print logs to a terminal, before the first call prints the log, the second call might succeed, the main program ends, and unnecessary panic might happen when the old goroutine still tries to print. As the print to terminal takes a long time usually and the time taken is unpredictable, the tick configuration becomes little tricky. Without the log, even if we set the tick to 1 msec, it works perfectly. With log and even if we try to set an optimal tick time, I am afraid that we could hit random flakes. Further, with a log, it is quite hard to find the correct tick time to test whether the log is working or not. 
2. If we print multiple errors in the log like `TestHandleChaincodeDeployErroneousIndexFile: statecouchdb_test.go:662: error handling CouchDB request. Error:no_usable_index,  Status Code:400,  Reason:No index exists for this sort, try indexing by the sort fields.` and then saying that the test is passed might confuse the developers. 

As this is a test code, when an error occurs due to changes made to index processing, we definitely need to debug the error with more info. At that time, some changes might be required for local testing. I think this is justifiable and we do that often during the development. Hence, this PR isn't introducing some new requirements.

With this change, I expect no flakes in the CouchDB UT and hence, a generic error like `"error executing query with sort"` might be sufficient for now.

denyeart: ok, thanks for the assessment, I'm good with it as-is.

denyeart (approved):

PR #1080, opened 2020-04-13
[LEDGER] Improve readability of HandleChaincodeDeploy

cendhu:
#### Type of change

#### Description
The following are good examples of _avoid disinformation, i.e., context should be explicit in the code itself_ mentioned in the [coding guidelines](https://docs.google.com/document/d/1mU4S4U1dDpBk9AA180RMRB7g3OH887nqbcEPGD3px3Q/edit?usp=sharing).
 
1. `if directoryPathArray[3] == "indexes" {`
2. `if directoryPathArray[3] == "collections" && directoryPathArray[5] == "indexes" {`
3. `collectionName := directoryPathArray[4]`

What is **3**, **4**, and **5**? Hence, we replace the above lines with the following

```
 indexInfo := getIndexInfo(directoryPath)
 switch {
	case indexInfo.hasIndexForChaincode:
	case indexInfo.hasIndexForCollection:
 }
```

and followed _the most important concept first and the low-level details in the bottom._ as per the formatting rules specified in the above coding guidelines. The following may not be a perfect one. However, we try to avoid large code refactoring and restrict ourselves to the ledger package.  
```
        chaincodeIndexDirDepth    = 3
	collectionDirDepth        = 3
	collectionNameDepth       = 4
	collectionIndexDirDepth   = 5
```

#### Additional details

These changes were initially part of #966 but we decided to push a separate PR (this one) for the readability related changes.

#### Related issues

[FAB-17739](https://jira.hyperledger.org/browse/FAB-17739)

### DIFF
--- a/core/ledger/kvledger/txmgmt/privacyenabledstate/common_storage_db.go
+++ b/core/ledger/kvledger/txmgmt/privacyenabledstate/common_storage_db.go
@@ -32,6 +32,14 @@ const (
 	pvtDataPrefix  = "p"
 	hashDataPrefix = "h"
 	couchDB        = "CouchDB"
+	// Example for chaincode indexes:
+	// "META-INF/statedb/couchdb/indexes/indexColorSortName.json"
+	chaincodeIndexDirDepth = 3
+	// Example for collection scoped indexes:
+	// "META-INF/statedb/couchdb/collections/collectionMarbles/indexes/indexCollMarbles.json"
+	collectionDirDepth      = 3
+	collectionNameDepth     = 4
+	collectionIndexDirDepth = 5
 )
 
 // StateDBConfig encapsulates the configuration for stateDB on the ledger.
@@ -306,32 +314,28 @@ func (s *CommonStorageDB) HandleChaincodeDeploy(chaincodeDefinition *cceventmgmt
 	}
 
 	for directoryPath, indexFiles := range dbArtifacts {
-		// split the directory name
-		directoryPathArray := strings.Split(directoryPath, "/")
-		// process the indexes for the chain
 		indexFilesData := make(map[string][]byte)
 		for _, f := range indexFiles {
 			indexFilesData[f.FileHeader.Name] = f.FileContent
 		}
-		if directoryPathArray[3] == "indexes" {
+
+		indexInfo := getIndexInfo(directoryPath)
+		switch {
+		case indexInfo.hasIndexForChaincode:
 			err := indexCapable.ProcessIndexesForChaincodeDeploy(chaincodeDefinition.Name, indexFilesData)
 			if err != nil {
 				logger.Errorf("Error processing index for chaincode [%s]: %s", chaincodeDefinition.Name, err)
 			}
-			continue
-		}
-		// check for the indexes directory for the collection
-		if directoryPathArray[3] == "collections" && directoryPathArray[5] == "indexes" {
-			collectionName := directoryPathArray[4]
-			_, ok := collectionConfigMap[collectionName]
+		case indexInfo.hasIndexForCollection:
+			_, ok := collectionConfigMap[indexInfo.collectionName]
 			if !ok {
-				logger.Errorf("Error processing index for chaincode [%s]: cannot create an index for an undefined collection=[%s]", chaincodeDefinition.Name, collectionName)
-			} else {
-				err := indexCapable.ProcessIndexesForChaincodeDeploy(derivePvtDataNs(chaincodeDefinition.Name, collectionName),
-					indexFilesData)
-				if err != nil {
-					logger.Errorf("Error processing collection index for chaincode [%s]: %s", chaincodeDefinition.Name, err)
-				}
+				logger.Errorf("Error processing index for chaincode [%s]: cannot create an index for an undefined collection=[%s]",
+					chaincodeDefinition.Name, indexInfo.collectionName)
+				continue
+			}
+			err := indexCapable.ProcessIndexesForChaincodeDeploy(derivePvtDataNs(chaincodeDefinition.Name, indexInfo.collectionName), indexFilesData)
+			if err != nil {
+				logger.Errorf("Error processing collection index for chaincode [%s]: %s", chaincodeDefinition.Name, err)
 			}
 		}
 	}
@@ -387,4 +391,23 @@ func extractCollectionNames(chaincodeDefinition *cceventmgmt.ChaincodeDefinition
 		}
 	}
 	return collectionConfigsMap, nil
+}
+
+type indexInfo struct {
+	hasIndexForChaincode  bool
+	hasIndexForCollection bool
+	collectionName        string
+}
+
+func getIndexInfo(indexPath string) *indexInfo {
+	indexInfo := &indexInfo{}
+	dirsDepth := strings.Split(indexPath, "/")
+	switch {
+	case dirsDepth[chaincodeIndexDirDepth] == "indexes":
+		indexInfo.hasIndexForChaincode = true
+	case dirsDepth[collectionDirDepth] == "collections" && dirsDepth[collectionIndexDirDepth] == "indexes":
+		indexInfo.hasIndexForCollection = true
+		indexInfo.collectionName = dirsDepth[collectionNameDepth]
+	}
+	return indexInfo
 }


core/ledger/kvledger/txmgmt/privacyenabledstate/common_storage_db.go
```diff
@@ -305,34 +318,33 @@ func (s *CommonStorageDB) HandleChaincodeDeploy(chaincodeDefinition *cceventmgmt
...
-				logger.Errorf("Error processing index for chaincode [%s]: cannot create an index for an undefined collection=[%s]", chaincodeDefinition.Name, collectionName)
-			} else {
-				err := indexCapable.ProcessIndexesForChaincodeDeploy(derivePvtDataNs(chaincodeDefinition.Name, collectionName),
-					indexFilesData)
-				if err != nil {
-					logger.Errorf("Error processing collection index for chaincode [%s]: %s", chaincodeDefinition.Name, err)
-				}
-			}
+
+		if !hasIndexForCollection(directoryPath) {
```
manish-sethi: You can have this function return collection name as well

core/ledger/kvledger/txmgmt/privacyenabledstate/common_storage_db.go
```diff
@@ -32,6 +32,19 @@ const (
...
+	// Example for collection scoped indexes:
+	// "META-INF/statedb/couchdb/collections/collectionMarbles/indexes/indexCollMarbles.json"
+	collectionDirDepth      = 3
+	collectionNameDepth     = 4
+	collectionIndexDirDepth = 5
+	// Regex can be used instead of numbers but it cannot help to remove all
+	// above numbers. This index path format is hardcoded at many places. It
+	// requires a separate refactoring which involves core/common/ccprovider.
+	// We may need to move core/common/ccprovider to internal/pkgs and fix
+	// hardcoded path -- FAB-17698
```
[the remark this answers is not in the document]
cendhu: I Agree. Removed.

core/ledger/kvledger/txmgmt/privacyenabledstate/common_storage_db.go
```diff
@@ -283,7 +296,7 @@ func (s *CommonStorageDB) GetPrivateDataMetadataByHash(namespace, collection str
 // and the errors because of bad index files - the later being unfixable by the admin. Note that the error suppression
 // is acceptable since peer can continue in the committing role without the indexes. However, executing chaincode queries
 // may be affected, until a new chaincode with fixed indexes is installed and instantiated
-func (s *CommonStorageDB) HandleChaincodeDeploy(chaincodeDefinition *cceventmgmt.ChaincodeDefinition, dbArtifactsTar []byte) error {
+func (s *CommonStorageDB) HandleChaincodeDeploy(chaincodeDefinition *cceventmgmt.ChaincodeDefinition, dbIndexArtifactsTar []byte) error {
```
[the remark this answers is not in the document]
cendhu: Reverted the changes. I agree that the name has become a little longer. 

My thought process is that code for today while keeping the structure in a way that extensions are possible tomorrow. Though the `dbArtifactsTar` enables the easy extension, I am not sure whether we need this much ease. Looks like the `dbArtifactsTar` was added in May 2018. We haven't supported any new artifacts and I also doubt whether we support anything new going forward. As it only holds index, I renamed it. Further, I had a problem with `GetDBType()` too as we support index only for the CouchDB. I am sure that the probability of supporting a new DB is too low. I would prefer to be explicit in the code so that it would not confuse the reader like _what other artifacts are there?_ As it is unrelated to this PR, I am reverting the changes made.

core/ledger/kvledger/txmgmt/privacyenabledstate/common_storage_db.go
cendhu: To maintain consistency between `hasIndexForChaincode` and `hasIndexForCollection`, I think it is okay to have a separate `getCollectionName` function. Though we might split the directory path thrice, I left it as-is given that deployment is an infrequent task.

manish-sethi: Replacing hard-coded numbers by `consts` is certainly good but beyond that, to be honest, I am not sure, if this PR improves much on the existing code.
A better way to make it more readable would be to contain the path parsing logic in function that simply returns a struct something like this...
`dbArtifactInfo{
    isIndexSpec bool,
    isCollectionLevel bool,
   collectionName string,
}` 
wdyt?

manish-sethi (changes requested):
@cendhu - left you a comment (https://github.com/hyperledger/fabric/pull/1080/files#r409108625)

core/ledger/kvledger/txmgmt/privacyenabledstate/common_storage_db.go
cendhu: The goal of this PR is just to replace the hardcoded number as it is a bad pattern as described in the PR description. This PR retains most of the existing logic as-is.

As I mentioned, I am not comfortable calling it `dbArtifacts`. Hence, I am not sure about using `isIndexSpec bool` either. We have only index artifacts for now.

I can do the following if there is an agreement.

(1) Rename the `dbArtifacts` to `dbIndexArtifacts`
(2) The following combines all three functions into a single one.
```
type struct indexInfo {
     hasIndexForChaincode bool;
     hasIndexForCollection bool;
     collectionName string;
}
func getIndexInfo(indexPath string) {
     indexInfo := &indexInfo{}
     dirsDepth := strings.Split(path, "/")
     if dirsDepth[chaincodeIndexDirDepth] == "indexes" {
           indexInfo.hasIndexForChaincode = true
           return indexInfo
     }
     if dirsDepth[collectionDirDepth] == "collections" &&
           dirsDepth[collectionIndexDirDepth] == "indexes" {
           indexInfo.hasIndexForCollection = true
           dirsDepth[collectionNameDepth]
     }
     return indexInfo     
}
```
The readability might be subjective. To me, having three functions called `hasIndexForChaincode()`, `hasIndexForCollection()`, and `getCollectionName()` was more readable than this. There is nothing wrong in having small functions. When we have a single function called `getIndexInfo()` combining all three, it has an additional info like if we have chaincode index, skip collection check. When we have three function, this context was only present in the `HandleChaincodeDeploy()`.

(3) The HandleChaincodeDeploy would look like (does not differ much from the current format)
```
indexInfo := getIndexInfo(directoryPath)
if indexInfo.hasIndexForChaincode {
     ....
     continue
}
if !indexInfo.hasIndexForCollection {
     ....
     continue
}
_, ok := collectionConfigMap[indexInfo.collectionName]
```

manish-sethi: > The goal of this PR is just to replace the hardcoded number as it is a bad pattern as described in the PR description.

If you would have stopped there, I had no problem. You missed my comment perhaps when I say "Replacing hard-coded numbers by consts is certainly good but beyond that.."

To understand a simple path structure, I have to move around three super tiny one-liner functions, not sure how it is better than what is already there. Also, parsing same string thrice, is more unnatural style to me. I surely, do not want to spend our time on these nit-picking like discussions. I think as a middle ground, you can just use const instead of hard coding.

cendhu: I followed **the most important concept first and the low-level details in the bottom**. Most of the time, we would not need to read low-level details. 

**Example-1:**
Assume that tomorrow, we introduce an index for `implicit_collection`. Though it can be still under the collection folder, I am taking this example to explain the reasoning. The developer needs to read only a few information and add a function called `hasIndexForImplicitCollection(directoryPath)` and even the `directoryPath` can be of very different format as compared to the existing path.

**Example-2:**
Suppose say some developer wants to read how index is deployed on CouchDB. Reading the `HandleChaincodeDeploy()` function alone is adequate and the developer can skip low-level details implemented in `hasIndexForChaincode()` `hasIndexForCollection()` `getCollectionName()`.

**Example-3:**
Assume that we change the `directoryPath` of collection. The developer has to simply go to `hasIndexForCollection()` and make the changes.

```
    if hasIndexForChaincode(directoryPath) {
    if !hasIndexForCollection(directoryPath) {
    collectionName := getCollectionName(directoryPath)
```
has better readability than the following IMO
```
    if directoryPathArray[chaincodeIndexDirDepth] == "indexes" {
    if directoryPathArray[collectionDirDepth] == "collections" && directoryPathArray[collectionIndexDirDepth] == "indexes" {
    collectionName := directoryPathArray[collectionNameDepth]
```
> Also, parsing same string thrice, is more unnatural style to me.

As it is an infrequent task, some amount of code duplication is fine.

cendhu: >  I surely, do not want to spend our time on these nit-picking like discussions. 

I agree and I am sorry for that. Given that I have strong opinion on **the most important concept first and the low-level details in the bottom**, I am trying to convince you.

manish-sethi: > I agree and I am sorry for that. Given that I have strong opinion on **the most important concept first and the low-level details in the bottom**, I am trying to convince you.

Anyone would hardly disagree to any philosophy of this sort in abstract. The difference of opinions arise from the situation you applying this to and at what granularity.
In this particular case, you have three functions (mostly one liners), they have to be invoked in exact same order (else they will panic). Moreover, this sequence of call is exactly at one place.

cendhu: In the initial PR, I did have a check on the len of the `dirDepth` to avoid this panic. Then, I thought that I would get a comment saying why are we changing the existing logic. 

I don't think we will come to an agreement on this. I am fine with closing this PR.

cendhu: Agreed on a middle-ground and updated the PR.

manish-sethi (approved):
Merging this one. However, I would like to see two improvements. That can be done in a separate PR
1) Move the constants related to new function, just above the function.
2) Though, not introduced in the PR but just noticed that array element access without checking lengths. A wrong input could cause panic. I recall that you mentioned that you had added that check but not included finally. But, it would be good to check for safety against a bad input.

PR #1085, opened 2020-04-14
[FAB-17720] improve MSP certificates validation.

DereckLuo:
FAB-17720 #done

#### Type of change

#### Description
* Added intermediate cert validation for MSP, which it must be issued
by one of the root certs.

#### Related issues
[FAB-17720](https://jira.hyperledger.org/browse/FAB-17720)

### DIFF
--- a/pkg/configtx/msp.go
+++ b/pkg/configtx/msp.go
@@ -670,6 +670,20 @@ func (m *MSP) validateCACerts() error {
 	if err != nil {
 		return fmt.Errorf("invalid intermediate cert: %v", err)
 	}
+	//TODO: follow the workaround that msp code use to incorporate cert.Verify()
+	for _, ic := range m.IntermediateCerts {
+		validIntermediateCert := false
+		for _, rc := range m.RootCerts {
+			err := ic.CheckSignatureFrom(rc)
+			if err == nil {
+				validIntermediateCert = true
+				break
+			}
+		}
+		if !validIntermediateCert {
+			return fmt.Errorf("intermediate cert not signed by any root certs of this MSP. serial number: %d", ic.SerialNumber)
+		}
+	}
 
 	err = validateCACerts(m.TLSRootCerts)
 	if err != nil {
--- a/pkg/configtx/msp_test.go
+++ b/pkg/configtx/msp_test.go
@@ -442,14 +442,11 @@ func TestUpdateApplicationMSP(t *testing.T) {
 	org1CertBase64, org1PKBase64, org1CRLBase64 := certPrivKeyCRLBase64(t, org1MSP)
 	org2CertBase64, org2PKBase64, org2CRLBase64 := certPrivKeyCRLBase64(t, org2MSP)
 
-	newRootCert, _ := generateCACertAndPrivateKey(t, "anotherca-org1.example.com")
+	newRootCert, newRootPrivKey := generateCACertAndPrivateKey(t, "anotherca-org1.example.com")
 	newRootCertBase64 := base64.StdEncoding.EncodeToString(pemEncodeX509Certificate(newRootCert))
 	org1MSP.RootCerts = append(org1MSP.RootCerts, newRootCert)
 
-	newIntermediateCert := &x509.Certificate{
-		KeyUsage: x509.KeyUsageKeyEncipherment | x509.KeyUsageDigitalSignature | x509.KeyUsageCertSign,
-		IsCA:     true,
-	}
+	newIntermediateCert, _ := generateIntermediateCACertAndPrivateKey(t, "anotherca-org1.example.com", newRootCert, newRootPrivKey)
 	newIntermediateCertBase64 := base64.StdEncoding.EncodeToString(pemEncodeX509Certificate(newIntermediateCert))
 	org1MSP.IntermediateCerts = append(org1MSP.IntermediateCerts, newIntermediateCert)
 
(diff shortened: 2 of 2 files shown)


caod123 (dismissed):
Looks fine to me @wlahti can you take a quick glance as well since you have more context on this path?

pkg/config/msp.go
```diff
@@ -670,6 +670,21 @@ func (m *MSP) validateCACerts() error {
...
+		rootCertPool.AddCert(cert)
+	}
+	opts := x509.VerifyOptions{
+		Roots: rootCertPool,
+	}
+	// Even though currently Fabric does not care if a certificate is expired or not,
+	// this cert verification will check expiration of a certificate. This will error
+	// out if the provided intermediate certs are not valid or expired.
+	for _, cert := range m.IntermediateCerts {
+		if _, err := cert.Verify(opts); err != nil {
```
wlahti: While ideally we'd use this Verify function, we can't since, as you pointed out, it considers expiration. For consistency with the other code, we should simply use CheckSignatureFrom to ensure that the intermediate certs were signed by one of the root certs.

wlahti (changes requested):

pkg/config/msp.go
wlahti: I just took another look at the msp code to figure out what it's doing for cert validation. I'll update the JIRA item related to this PR to include the details of what I'm thinking. 

For now, I think we go ahead with updating this PR to use cert.CheckSignatureFrom() and merge it in.  After that, we should try to incorporate cert.Verify() following the workaround that the msp code uses.

DereckLuo: done

pkg/config/msp.go
```diff
@@ -670,6 +670,23 @@ func (m *MSP) validateCACerts() error {
...
+		for _, rc := range m.RootCerts {
+			err := ic.CheckSignatureFrom(rc)
+			if err == nil {
+				validIntermediateCert = true
+			}
+		}
+	}
+	if !validIntermediateCert {
+		return errors.New("invalid intermediate cert")
+	}
```
caod123: This logic only checks for a single valid intermediate cert. Not sure if you intended on validating all intermediate certs. If you want to you could return the error on 688 within the loop  if the err on 681 is not nil and if the loop passes all the way through you know all intermediate certs are valid.

caod123 (changes requested):

pkg/config/msp.go
DereckLuo: you are right. the original logic only checks if one intermediate cert is valid. fix applied.

pkg/config/msp.go
```diff
@@ -670,6 +670,23 @@ func (m *MSP) validateCACerts() error {
...
+	for _, cert := range m.RootCerts {
+		rootCertPool.AddCert(cert)
+	}
+	//TODO: follow the workaround that msp code use to incorporate cert.Verify()
+	for _, ic := range m.IntermediateCerts {
+		validIntermediateCert := false
+		for _, rc := range m.RootCerts {
+			err := ic.CheckSignatureFrom(rc)
+			if err == nil {
+				validIntermediateCert = true
```
caod123: can probably break here once we've validated against a root cert

caod123 (dismissed):

pkg/config/msp.go
DereckLuo: definitely

caod123 (dismissed):

pkg/config/msp.go
```diff
@@ -670,6 +670,24 @@ func (m *MSP) validateCACerts() error {
...
+				break
+			}
+		}
+		if !validIntermediateCert {
+			return errors.New("invalid intermediate cert")
```
wlahti: I know this error message may not be long-lived, but just saying that the intermediate cert is invalid doesn't really provide the user any actionable item to correct things. This should really state that the intermediate cert has not been signed by any of the root certs (and should probably include the cert's serial number). Something like: "intermediate cert not signed by any root certs of this MSP. serial number: 123456".

pkg/config/msp.go
```diff
@@ -670,6 +670,24 @@ func (m *MSP) validateCACerts() error {
 	if err != nil {
 		return fmt.Errorf("invalid intermediate cert: %v", err)
 	}
+	rootCertPool := x509.NewCertPool()
```
wlahti: I know we'll use this in the future commit related to the TODO below, but we shouldn't get in a habit of leaving around code we aren't using. :)

DereckLuo: goot catch, thanks.

yacovm:
When you setup an **MSP** via the [Setup](https://github.com/hyperledger/fabric/blob/d0192fda079234333e79333b48578e5481b4d221/msp/mspimplsetup.go#L474) method, it performs self sanity checks and errors if something is misconfigured. 

It also [takes into account channel capabilities](https://github.com/hyperledger/fabric/blob/d0192fda079234333e79333b48578e5481b4d221/msp/mspimpl.go#L116-L137). 

Can you explain why the code of the config library, essentially re-implements the **MSP** sanity checks and doesn't just attempt to create an **MSP** instance and **Setup()** it?

DereckLuo:
@yacovm thanks for the review. Currently in _configtx_ library, the intent is to build an interface for consumer to access/modify fabric configurations. 
Therefore, instead of directly using **bccspmsp** struct in fabric, we created a **MSP** struct in _configtx_ to hold all necessary information for membership configurations, and use it to modify fabric configuration protos. 

In this code, we want to have at least some sanity checks while consumers create the **configtx MSP** struct before it is being used to create configuration protos.

caod123 (approved):

PR #1089, opened 2020-04-14
Extend SigningIdentity with config update signing functions

wlahti:
#### Type of change

#### Description

Move and rename config update signing functions.

#### Related issues

[FAB-17741](https://jira.hyperledger.org/browse/FAB-17741)

### DIFF
--- a/pkg/config/signer.go
+++ b/pkg/config/signer.go
@@ -9,11 +9,17 @@ package config
 import (
 	"crypto"
 	"crypto/ecdsa"
+	"crypto/rand"
 	"crypto/x509"
 	"encoding/asn1"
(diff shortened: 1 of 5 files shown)


pkg/config/signer.go
```diff
@@ -76,3 +82,145 @@ func toLowS(key ecdsa.PublicKey, sig ecdsaSignature) ecdsaSignature {
 
 	return sig
 }
+
+// SignConfigUpdate signs the given configuration update with a
```
caod123, line 86: nit: can probably reword this now since there isnt a "specified" signing identity passed in anymore.

caod123 (approved):

PR #1091, opened 2020-04-15
Move keys.go Utility to SW package

lindluni:
Moves the keys.go utility to the SW package where the majority the consumers are and kills off production code that existed entirely for driving tests.


lindluni:
Had to close and reopen as a new branch as Azure was pulling the wrong refspec, but I've addressed the comments from the previous PR.

mastersingh24 (approved):
Given the desired scope of the change, this looks fine to me.
I do think we can remove a number of the "derTo" and "pemTo" functions in the future.

lindluni:
> Given the desired scope of the change, this looks fine to me.
> I do think we can remove a number of the "derTo" and "pemTo" functions in the future.

Agreed, most of them are just wrapping calls to standard library functions

PR #1095, opened 2020-04-15
[LEDGER] remove TxMgr interface

cendhu:
#### Type of change

#### Description

We have a [TxMgr](https://github.com/hyperledger/fabric/blob/master/core/ledger/kvledger/txmgmt/txmgr/txmgr.go#L27) interface as the initial plan was to implement two types of transaction manager:

1. lock-based txmgr
2. snapshot isolation based txmgr

In the last three years and in the many coming years, we may not implement another `TxMgr`. Hence, we remove the interface.

#### Additional details

As we decided to keep a single TxMgr implementation, we can avoid additional folder depth and reduce a package by moving `lockbasedtxmgr` to `txmgr`.

`TxStatInfo` defined in `txmgr` is imported in both `lockbasedtxmgr` and `validator`. If we move the `lockbasedtxmgr` to `txmgr`, we will get an import cycle. As `TxStatInfo` is used by `kvledger`, `metrics`, `lockbasedtxmgr`, and `validator`, we need to find a suitable place to keep these structs. Only if we can do that, we can move `lockbasedtxmgr` to `txmgr` -- [FAB-17757](https://jira.hyperledger.org/browse/FAB-17757)

Merge this only after merging  #1094 

#### Related issues

[FAB-17756](https://jira.hyperledger.org/browse/FAB-17756)


cendhu:
We will reopen this once we finish refactoring the mvcc validator.

PR #1098, opened 2020-04-15
DRY up pkg/config envelope creation

wlahti:
#### Type of change

#### Description

Reuse newEnvelope() when creating a signed envelope. 

#### Related issues

[FAB-17741](https://jira.hyperledger.org/browse/FAB-17741)

### DIFF
--- a/pkg/configtx/signer.go
+++ b/pkg/configtx/signer.go
@@ -130,7 +130,7 @@ func (s *SigningIdentity) SignConfigUpdateEnvelope(configUpdate *cb.ConfigUpdate
 		Signatures:   signatures,
 	}
 
-	signedEnvelope, err := s.createSignedEnvelope(cb.HeaderType_CONFIG_UPDATE, configUpdate.ChannelId, configUpdateEnvelope)
+	signedEnvelope, err := s.newSignedEnvelope(cb.HeaderType_CONFIG_UPDATE, configUpdate.ChannelId, configUpdateEnvelope)
 	if err != nil {
 		return nil, fmt.Errorf("failed to create signed config update envelope: %v", err)
 	}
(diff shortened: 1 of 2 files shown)


pkg/config/config.go
```diff
@@ -405,8 +405,9 @@ func newConfigGroup() *cb.ConfigGroup {
 	}
 }
 
-// newEnvelope creates an unsigned envelope of type txType using with the marshalled
-// cb.ConfigGroupEnvelope proto message.
+// newEnvelope creates an unsigned envelope of the desired type containing
+// a payload Header and the marshaled proto message as the payload Data.
+// message
```
stephyee: Did you forget to delete this line?

wlahti: Ah yep, didn't even notice it for a while after reading your comment. :) Removed.

pkg/configtx/signer.go
```diff
@@ -130,60 +130,50 @@ func (s *SigningIdentity) SignConfigUpdateEnvelope(configUpdate *cb.ConfigUpdate
...
 	return signedEnvelope, nil
 }
 
-// createSignedEnvelope creates a signed envelope of the desired type and signs it.
-func (s *SigningIdentity) createSignedEnvelope(txType cb.HeaderType, channelID string, envelope proto.Message) (*cb.Envelope, error) {
-	channelHeader := channelHeader(txType, msgVersion, channelID, epoch)
-
-	signatureHeader, err := s.signatureHeader()
+// newSignedEnvelope creates a signed envelope of the desired type and signs it.
+func (s *SigningIdentity) newSignedEnvelope(txType cb.HeaderType, channelID string, dataMsg proto.Message) (*cb.Envelope, error) {
```
caod123, line 142: could maybe even call this just `signedEnvelope` and just drop the `new` since that seems to be the pattern we're moving towards

caod123: I'm indifferent either way. I'll let you decide which you prefer, let me know if you want to merge as is or if you want to change it

caod123 (approved):

pkg/configtx/signer.go
wlahti: I say we keep it as-is for now.  We can do another pass to make sure we're consistent in our naming of functions after the current set of tasks are merged in.

PR #1099, opened 2020-04-15
[FAB-17732] HSM clarifications

pamandrejko:

Addressed Dave's leftover feedback from PR  https://github.com/hyperledger/fabric/pull/1027:
- Add peer/orderer environment variables
- Explain how do you get the HSM protected private key generated by Fabric CA client over to the peer/orderer HSM? What does "Once it is deployed mean"?

Added `Immutable : false` to the BCCSP parameters.

#### Type of change

- Documentation update

#### Description

Original updates were made to the release-1.4 branch to support AWS HSM via the BCCSP  `AltId` field. `AltID` is not supported yet in the master branch. When it is we will need to add it.
In the meantime, I addressed Dave's other comments that we will need to cherry-pick back to release-1.4 when after this is merged.

#### Related issues

FAB-17732

### DIFF
--- a/docs/source/hsm.md
+++ b/docs/source/hsm.md
@@ -8,11 +8,11 @@ supports the PKCS11 standard to communicate with an HSM.
 
 ## Configuring an HSM
 
-To use an HSM with your Fabric node, you need to update the BCCSP (Crypto Service
+To use an HSM with your Fabric node, you need to update the `bccsp` (Crypto Service
 Provider) section of the node configuration file such as core.yaml or
-orderer.yaml. In BCCSP section, you need to select PKCS11 as the provider and
-provide the path to the PKCS11 library that you would like to use. You also need
-to provide the label and pin of the token that you created for your cryptographic
+orderer.yaml. In the `bccsp` section, you need to select PKCS11 as the provider and
+enter the path to the PKCS11 library that you would like to use. You also need
+to provide the `Label` and `PIN` of the token that you created for your cryptographic
 operations. You can use one token to generate and store multiple keys.
 
 The prebuilt Hyperledger Fabric Docker images are not enabled to use PKCS11. If
@@ -38,7 +38,7 @@ You can then use softhsm to create the token that will handle the cryptographic
 operations of your Fabric node inside an HSM slot. In this example, we create a
 token labelled "fabric" and set the pin to "71811222". After you have created
 the token, update the configuration file to use PKCS11 and your token as the
-crypto service provider. You can find an example BCCSP section below:
+crypto service provider. You can find an example `bccsp` section below:
 
 ```
 #############################################################################
@@ -53,17 +53,36 @@ bccsp:
     Label: fabric
     hash: SHA2
     security: 256
+    Immutable: false
 ```
 
-You can also use environment variables to override the relevant fields of the
-configuration file. If you are connecting to an HSM using the Fabric CA server,
-you need to set the following environment variables:
+By default, when private keys are generated using the HSM, the private key is mutable, meaning PKCS11 private key  attributes can be changed after the key is generated. Setting `Immutable` to `true` means that the private key attributes cannot be altered after key generation. Before you configure immutability by setting `Immutable: true`, ensure that PKCS11 object copy is supported by the HSM.
+
+You can also use environment variables to override the relevant fields of the configuration file. If you are connecting to softhsm2 using the Fabric CA server, you could set the following environment variables or directly set the corresponding values in the CA server config file:
 
 ```
 FABRIC_CA_SERVER_BCCSP_DEFAULT=PKCS11
 FABRIC_CA_SERVER_BCCSP_PKCS11_LIBRARY=/etc/hyperledger/fabric/libsofthsm2.so
 FABRIC_CA_SERVER_BCCSP_PKCS11_PIN=71811222
 FABRIC_CA_SERVER_BCCSP_PKCS11_LABEL=fabric
+```
+
+If you are connecting to softhsm2 using the Fabric peer, you could set the following environment variables or directly set the corresponding values in the peer config file:
+
+```
+CORE_PEER_BCCSP_DEFAULT=PKCS11
+CORE_PEER_BCCSP_PKCS11_LIBRARY=/etc/hyperledger/fabric/libsofthsm2.so
+CORE_PEER_BCCSP_PKCS11_PIN=71811222
+CORE_PEER_BCCSP_PKCS11_LABEL=fabric
+```
+
+If you are connecting to softhsm2 using the Fabric orderer, you could set the following environment variables or directly set the corresponding values in the orderer config file:
+
+```
+ORDERER_GENERAL_BCCSP_DEFAULT=PKCS11
+ORDERER_GENERAL_PKCS11_LIBRARY=/etc/hyperledger/fabric/libsofthsm2.so
+ORDERER_GENERAL_PKCS11_PIN=71811222
+ORDERER_GENERAL_PKCS11_LABEL=fabric
 ```
 
 If you are deploying your nodes using docker compose, after building your own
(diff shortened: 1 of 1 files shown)


docs/source/hsm.md
```diff
@@ -53,11 +53,12 @@ bccsp:
...
     security: 256
+    Immutable: false
 ```
 
-You can also use environment variables to override the relevant fields of the
-configuration file. If you are connecting to an HSM using the Fabric CA server,
-you need to set the following environment variables:
+By default, when private keys are generated using the HSM, the private key is mutable, meaning PKCS11 private key  attributes can be changed after the key is generated. Setting `Immutable` to `true` means that the private key attributes cannot be altered after key generation. Before you configure immutability by setting `Immutable: true`, ensure that PKCS11 object cloning is supported by the HSM.
+
+You can also use environment variables to override the relevant fields of the configuration file. If you are connecting to softhsm2 using the Fabric CA server, you need to set the following environment variables:
```
caod123: Because the first sentence implies the environment variables are specifically for overriding the config file, it might sound better to say something along the lines of `you may set the following environment variables`. Just something to better imply that the properties can both be set via the config and overwritten with the env vars. Otherwise reading this sentence might imply that you must configure it via the environment variables.

docs/source/hsm.md
```diff
@@ -66,6 +67,24 @@ FABRIC_CA_SERVER_BCCSP_PKCS11_PIN=71811222
 FABRIC_CA_SERVER_BCCSP_PKCS11_LABEL=fabric
 ```
 
+If you are connecting to softhsm2 using the Fabric peer, you would set the following environment variables:
```
caod123: Likewise maybe `could` or `may` instead of `would` might work better here

docs/source/hsm.md
```diff
@@ -66,6 +67,24 @@ FABRIC_CA_SERVER_BCCSP_PKCS11_PIN=71811222
...
+
+```
+CORE_PEER_BCCSP_DEFAULT=PKCS11
+CORE_PEER_BCCSP_PKCS11_LIBRARY=/etc/hyperledger/fabric/libsofthsm2.so
+CORE_PEER_BCCSP_PKCS11_PIN=71811222
+CORE_PEER_BCCSP_PKCS11_LABEL=fabric
+```
+
+If you are connecting to softhsm2 using the Fabric orderer, you would set the following environment variables:
```
caod123: Same as above

docs/source/hsm.md
```diff
@@ -89,31 +108,31 @@ in the `signcerts` folder to retrieve the private key from inside the HSM.
...
-configuration file as you would make to a peer or ordering node. Because you can
-use Fabric CA to generate keys inside an HSM, the process of creating the local
-MSP folders is straightforward. Use the following steps:
+2. Use the Fabric CA client to register the peer or ordering node identities with your CA.
 
-1. Create an HSM token and point to it in the Fabric CA server configuration
-file. When the Fabric CA server starts, it will generate the CA signing
-certificate inside your HSM. If you are not concerned about exposing your CA
-signing certificate, you can skip this step.
+3. Before you deploy a peer or ordering node with HSM, you need to enroll the node admin, storing its private key in the HSM. Edit the BCCSP section of the Fabric CA client config file or associated environment variables to point to the HSM configuration for your peer or ordering node. In the Fabric CA Client configuration file, replace the default `SW` configuration with the `PKCS11` configuration and providing the values for your own HSM:
```
caod123: `Before you deploy a peer or ordering node with HSM support, you need to enroll the node admin by storing its private key in the HSM.`

`configuration and provide the values for your own HSM:`

docs/source/hsm.md
```diff
@@ -89,31 +108,31 @@ in the `signcerts` folder to retrieve the private key from inside the HSM.
...
+  ```
+  bccsp:
+    default: PKCS11
+    pkcs11:
+      Library: /etc/hyperledger/fabric/libsofthsm2.so
+      Pin: 71811222
+      Label: fabric
+      hash: SHA2
+      security: 256
+      Immutable: false
```
caod123, line 137: Is there any reason why some of these are capitalized while others aren't?

docs/source/hsm.md
```diff
@@ -89,31 +108,31 @@ in the `signcerts` folder to retrieve the private key from inside the HSM.
...
+  ```
 
-3. Edit the Fabric CA client config file or environment variables to use your
-HSM as the crypto service provider. Then for each node, use the Fabric CA client
-to generate the component MSP folder by enrolling against the node identity. The
-enroll command will generate the private key inside your HSM.
+  Then for each node, use the Fabric CA client to generate the peer or ordering node's MSP folder by enrolling against the node identity that you registered in step 2. The enroll command will use the HSM to generate the private key for the peer or ordering node and store it inside the HSM instead of in the /keystore folder of the associated MSP.
```
caod123: Probably better to put `/keystore` in a code block

caod123: I see in other parts of this doc we just reference `keystore` folder without a `/`. Not sure if you want to standardize which one we use.

docs/source/hsm.md
```diff
@@ -124,14 +143,9 @@ can use an HSM using the following steps:
 Then use your CA to generate the private key and signing certificate for each
 node, with the private key generated inside the HSM.
 
-2. Use your CA to build the node MSP folder. Place the signing certificate that
-you generated in step 1 inside the `signcerts` folder. You can leave the
-`keystore` folder empty.
+2. Use your CA to build the peer or ordering node MSP folder. Place the signing certificate that you generated in step 1 inside the `signcerts` folder. You can leave the `keystore` folder empty.
```
caod123, line 153: Here's where I mentioned above about `keystore`

docs/source/hsm.md
ashku: Private Key is not stored , it is generated and stored on HSM. These two aspects are different.

docs/source/hsm.md
```diff
@@ -89,31 +108,31 @@ in the `signcerts` folder to retrieve the private key from inside the HSM.
 The process for creating the MSP folders will be different depending on if you
 are using a Fabric Certificate Authority (CA) your own CA.
 
-### Using a Fabric CA
+### Using an HSM with a Fabric CA
+
+You can set up a Fabric CA to use an HSM by making the same edits to the CA server configuration file as you would make to a peer or ordering node. Because you can use the Fabric CA to generate keys inside an HSM, the process of creating the local MSP folders is straightforward. Use the following steps:
+
+1. Create an HSM token and point to it in the Fabric CA server configuration file. When the Fabric CA server starts, the HSM described in the configuration file will generate the CA signing certificate and store the private key inside the HSM. If you are not concerned about exposing your CA signing certificate, you can skip this step and only configure an HSM for your peer or ordering nodes.
```
ashku: Replace HSM token with Label. Also the precondition step should be : 

1. Create Partition on HSM Server and note down Label and PIN for partition.

1.  Follow HSM Vendor document to configure HSM Client and Server to communicate.

ashku (commented):
@pamandrejko , some comments. These comments are applicable to all HSM docs. Please make change everywhere.

docs/source/hsm.md
denyeart: "you need to enroll the node admin" - isn't it enrolling of the node's identity, not the node's admin identity.

docs/source/hsm.md
denyeart: "the HSM" ... similar to my prior comments, I think it would be better to say "the node's HSM", as I was initially confused between the CA's HSM and the node's HSM.

docs/source/hsm.md
pamandrejko: You would have to ask the developers that.

docs/source/hsm.md
pamandrejko: Added both options throughout the content.

docs/source/hsm.md
pamandrejko: updated references through the doc

docs/source/hsm.md
pamandrejko: Added a Before you begin section with these steps.

docs/source/hsm.md
ashku: There is no specific reason. What I can think of is , Capital letter characterizes HSM related attributes while lower case are generic security attributes.

denyeart (dismissed):
Looks good to me, let's see if Danny wants to do a final review...

docs/source/hsm.md
caod123: I talked to @jyellick about this and it appears the standard we're trying to move towards is lowerCamelCase. Unfortunately, I'm not too sure if simply changing all the fields to lower case will still work as intended with how fabric currently loads the yaml. For now it's probably fine as is until we can put together a proper test for verifying the pkcs11 fields are loaded and used properly.

caod123: In addition, it seems in orderer.yaml the fields are all capitalized, whereas in peer's core.yaml it isnt. Not sure about fabric ca I'll have to verify

docs/source/hsm.md
```diff
@@ -53,11 +53,12 @@ bccsp:
...
     hash: SHA2
     security: 256
+    Immutable: false
 ```
 
-You can also use environment variables to override the relevant fields of the
-configuration file. If you are connecting to an HSM using the Fabric CA server,
-you need to set the following environment variables:
+By default, when private keys are generated using the HSM, the private key is mutable, meaning PKCS11 private key  attributes can be changed after the key is generated. Setting `Immutable` to `true` means that the private key attributes cannot be altered after key generation. Before you configure immutability by setting `Immutable: true`, ensure that PKCS11 object cloning is supported by the HSM.
+
```
nikhil550, line 60: it may be worth mentioning what `PKCS11 object cloning` means, since I could not find any detail with a quick search.

nikhil550 (commented):
LGTM, just left one small comment.

docs/source/hsm.md
ashku: Cloning is same copy object as C_CopyObject Covered under Object Management Section 5.7.2 of pkcs11 interface document https://docs.oasis-open.org/pkcs11/pkcs11-base/v3.0/csprd01/pkcs11-base-v3.0-csprd01.pdf. Do you think `clone` should be replaced by` copy` @nikhil550 ?

nikhil550: that is definitely a lot easier to find.

pamandrejko: Thanks @ashku I changed `cloning` to `copy`

caod123 (approved):

caod123:
LGTM

PR #1106, opened 2020-04-16
Add the sample policy to remove the test warning info

bjzhang03:
Signed-off-by: bjzhang <bjzhang1991@gmail.com>

#### Type of change

- Test update

#### Description

When I run the test, it always tell the warning.
I found the reason and repair it by add the sample policy in it.

### DIFF
--- a/common/channelconfig/realconfig_test.go
+++ b/common/channelconfig/realconfig_test.go
@@ -34,6 +34,12 @@ func TestWithRealConfigtx(t *testing.T) {
 			Port: 7,
 		},
 	}
+	conf.Application.Policies = map[string]*genesisconfig.Policy{
+		"SamplePolicy": {
+			Type: "ImplicitMeta",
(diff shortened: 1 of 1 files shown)


bjzhang03:
warning

bjzhang03:
warning

mastersingh24:
@bjzhang03  - thanks for this change.  I'm ok with backporting this, but this pull request should be made against the master branch as well.

mastersingh24 (changes requested):
As noted in my comment, please make a pull request for this on the master branch as well

bjzhang03:
> @bjzhang03 - thanks for this change. I'm ok with backporting this, but this pull request should be made against the master branch as well.
The master use the different test case, and does not have this problem.

bjzhang03:
The master use the different test case, and does not have this problem.

mastersingh24 (approved):

PR #1113, opened 2020-04-17
[FAB-17742] Convert Add to Set in configtx package.

DereckLuo:

#### Type of change

#### Description
* Part 1 of converting adds to sets in configtx package,
which includes ACLs, Anchorpeers, and Capabilities.
#### Related issues
[FAB-17742](https://jira.hyperledger.org/browse/FAB-17742)

### DIFF
--- a/pkg/configtx/application.go
+++ b/pkg/configtx/application.go
@@ -67,13 +67,9 @@ func (c *ConfigTx) ApplicationConfiguration() (Application, error) {
 }
 
 // AddAnchorPeer adds an anchor peer to an existing channel config transaction.
-// It must add the anchor peer to an existing org and the anchor peer must not already
-// exist in the org.
+// If anchor peer endpoints already exist in configuration, this action will be a no-op.
 func (c *ConfigTx) AddAnchorPeer(orgName string, newAnchorPeer Address) error {
-	applicationOrgGroup, ok := c.updated.ChannelGroup.Groups[ApplicationGroupKey].Groups[orgName]
-	if !ok {
-		return fmt.Errorf("application org %s does not exist in channel config", orgName)
-	}
+	applicationOrgGroup := c.updated.ChannelGroup.Groups[ApplicationGroupKey].Groups[orgName]
 
 	anchorPeersProto := &pb.AnchorPeers{}
 
@@ -90,8 +86,7 @@ func (c *ConfigTx) AddAnchorPeer(orgName string, newAnchorPeer Address) error {
 
 	for _, anchorPeer := range anchorProtos {
 		if anchorPeer.Host == newAnchorPeer.Host && anchorPeer.Port == int32(newAnchorPeer.Port) {
-			return fmt.Errorf("application org %s already contains anchor peer endpoint %s:%d",
-				orgName, newAnchorPeer.Host, newAnchorPeer.Port)
+			return nil
 		}
 	}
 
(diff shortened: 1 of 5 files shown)


pkg/configtx/application.go
```diff
@@ -66,14 +66,10 @@ func (c *ConfigTx) ApplicationConfiguration() (Application, error) {
...
-// It must add the anchor peer to an existing org and the anchor peer must not already
-// exist in the org.
-func (c *ConfigTx) AddAnchorPeer(orgName string, newAnchorPeer Address) error {
-	applicationOrgGroup, ok := c.updated.ChannelGroup.Groups[ApplicationGroupKey].Groups[orgName]
-	if !ok {
-		return fmt.Errorf("application org %s does not exist in channel config", orgName)
-	}
+// SetAnchorPeer sets an anchor peer to an existing channel config transaction.
+// If anchor peer endpoints already exist in configuration, this action will be a no-op.
+func (c *ConfigTx) SetAnchorPeer(orgName string, newAnchorPeer Address) error {
```
mastersingh24: This is a tricky one ... I believe that technically we are **adding** and anchor peer rather than simply just **setting** one

pkg/configtx/capabilities.go
```diff
@@ -57,8 +51,10 @@ func (c *ConfigTx) ApplicationCapabilities() ([]string, error) {
 	return capabilities, nil
 }
 
-// AddChannelCapability adds capability to the provided channel config.
-func (c *ConfigTx) AddChannelCapability(capability string) error {
+// SetChannelCapability adds capability to the provided channel config.
+// If the provided capability already exist in current configuration, this action
+// will be a no-op.
+func (c *ConfigTx) SetChannelCapability(capability string) error {
```
mastersingh24: Similar comment to the one for `AddAnchorPeer` ... we are not really setting a single value here but rather adding this

mastersingh24 (commented):
See other comments, but overall I'm not sure that replacing "Add" with "Set" is appropriate in all of the cases here

@jyellick  @sykesm - opinion?

jyellick:
> See other comments, but overall I'm not sure that replacing "Add" with "Set" is appropriate in all of the cases here
> 
> @jyellick @sykesm - opinion?

My initial reaction was that 'Add' was a better choice here, but the more I've thought of it, the more I think 'Set' might be correct.  The behavior of these functions seems to be "If the key already exists, replace the value, and if the key does not exist, add the value".  I can't come up with a better verb than 'Set' for this sort of behavior.

I would agree that it's a bit confusing, particularly for the ACLs, that you are passing in a map of key-values with 'Set', to update another map.  I would intuitively expect the old map to be replaced with this new map, but instead we're doing a 'Set' on each key instead.  Maybe 'Add' is more clear here, or maybe there's a better verb, though nothing comes to mind.

DereckLuo:
Thanks for the review. Totally agree with the suggestions.
Couple changes that pushed out for this PR.
* reverted **AnchorPeers** and **Capabilities** back to use **Add** prefix.
* for **SetACLs**, replace the old map with the new map instead of merging the two maps.

pkg/configtx/application.go
caod123: Depends on how you interpret the data. You could argue you're setting the existence of that anchor peer. I do agree by and large this is really just an add since we treat the Anchor Peers data as a slice, but the idea was to attempt standardizing the terminology for add/update/set. Either way the comment should be updated if we decide to go with set since `SetAnchorPeer sets an anchor peer to an existing channel` doesn't exactly sound correct 😛I'm fine keeping this as Add if there's agreement on it.

mastersingh24 (commented):
Thanks, Dereck!  LGTM ... but will let @caod123  have a look as well

caod123 (approved):
LGTM

PR #1114, opened 2020-04-17
[FAB-17753] Support updating orderer and consortium MSP configuration

wuxuer:

#### Type of change

- New feature

#### Related issues
https://jira.hyperledger.org/browse/FAB-17753

### DIFF
--- a/pkg/configtx/consortiums.go
+++ b/pkg/configtx/consortiums.go
@@ -244,3 +244,12 @@ func implicitMetaAnyPolicy(policyName string) (*standardConfigPolicy, error) {
 		value: implicitMetaPolicy,
 	}, nil
 }
+
+// getConsortiumOrg returns the organization config group for a consortium org in the
+// provided config. It will panic if the consortium doesn't exist, and it
+// will return nil if the org doesn't exist in the config.
+func getConsortiumOrg(config *cb.Config, consortiumName string, orgName string) *cb.ConfigGroup {
+	consortiumsGroup := config.ChannelGroup.Groups[ConsortiumsGroupKey].Groups
+	consortiumGroup := consortiumsGroup[consortiumName]
+	return consortiumGroup.Groups[orgName]
+}
--- a/pkg/configtx/consortiums_test.go
+++ b/pkg/configtx/consortiums_test.go
@@ -908,6 +908,21 @@ func TestGetConsortiums(t *testing.T) {
 	gt.Expect(len(baseConsortiums)).To(Equal(len(consortiums)))
 }
 
+func TestGetConsortiumOrg(t *testing.T) {
+	t.Parallel()
+	gt := NewGomegaWithT(t)
+
+	consortiumGroup, err := baseConsortiumChannelGroup(t)
+	gt.Expect(err).NotTo(HaveOccurred())
+
+	config := &cb.Config{
+		ChannelGroup: consortiumGroup,
+	}
+
+	org1ConfigGroup := getConsortiumOrg(config, "Consortium1", "Org1")
+	gt.Expect(org1ConfigGroup).To(Equal(config.ChannelGroup.Groups[ConsortiumsGroupKey].Groups["Consortium1"].Groups["Org1"]))
+}
+
 func baseConsortiums(t *testing.T) []Consortium {
 	return []Consortium{
 		{
@@ -926,4 +941,18 @@ func baseConsortiums(t *testing.T) []Consortium {
 			},
 		},
 	}
+}
+
+func baseConsortiumChannelGroup(t *testing.T) (*cb.ConfigGroup, error) {
+	channelGroup := newConfigGroup()
+
+	consortiums := baseConsortiums(t)
+	consortiumsGroup, err := newConsortiumsGroup(consortiums)
+	if err != nil {
+		return nil, err
+	}
+
+	channelGroup.Groups[ConsortiumsGroupKey] = consortiumsGroup
+
+	return channelGroup, nil
 }
--- a/pkg/configtx/msp.go
+++ b/pkg/configtx/msp.go
@@ -117,19 +117,45 @@ func (c *ConfigTx) ConsortiumMSP(consortiumName, orgName string) (MSP, error) {
 // YEAR is a time duration for a standard 365 day year.
 const YEAR = 365 * 24 * time.Hour
 
+// CreateConsortiumOrgMSPCRL creates a CRL that revokes the provided certificates
+// for the specified consortium org signed by the provided SigningIdentity.
+func (c *ConfigTx) CreateConsortiumOrgMSPCRL(consortiumName string, orgName string, signingIdentity *SigningIdentity, certs ...*x509.Certificate) (*pkix.CertificateList, error) {
+	msp, err := c.ConsortiumMSP(consortiumName, orgName)
+	if err != nil {
+		return nil, fmt.Errorf("retrieving consortium org msp: %s", err)
+	}
+
+	return msp.newMSPCRL(signingIdentity, certs...)
+}
+
+// CreateOrdererMSPCRL creates a CRL that revokes the provided certificates
+// for the specified orderer org signed by the provided SigningIdentity.
+func (c *ConfigTx) CreateOrdererMSPCRL(orgName string, signingIdentity *SigningIdentity, certs ...*x509.Certificate) (*pkix.CertificateList, error) {
+	msp, err := c.OrdererMSP(orgName)
+	if err != nil {
+		return nil, fmt.Errorf("retrieving orderer msp: %s", err)
+	}
+
+	return msp.newMSPCRL(signingIdentity, certs...)
+}
+
 // CreateApplicationMSPCRL creates a CRL that revokes the provided certificates
-// for the specified application org signed by the provided SigningIdentity. It
-// returns the CRL as the PEM-encoded bytes. If any of the provided certs were
-// not signed by any of the root/intermediate CA cets in the MSP configuration,
-// it will return an error.
+// for the specified application org signed by the provided SigningIdentity.
 func (c *ConfigTx) CreateApplicationMSPCRL(orgName string, signingIdentity *SigningIdentity, certs ...*x509.Certificate) (*pkix.CertificateList, error) {
 	msp, err := c.ApplicationMSP(orgName)
 	if err != nil {
 		return nil, fmt.Errorf("retrieving application msp: %s", err)
 	}
 
-	err = msp.validateCertificates(signingIdentity.Certificate, certs...)
-	if err != nil {
+	return msp.newMSPCRL(signingIdentity, certs...)
+}
+
+// newMSPCRL creates a CRL that revokes the provided certificates for the specified org
+// signed by the provided SigningIdentity. If any of the provided certs were
+// not signed by any of the root/intermediate CA cets in the MSP configuration,
+// it will return an error.
+func (m *MSP) newMSPCRL(signingIdentity *SigningIdentity, certs ...*x509.Certificate) (*pkix.CertificateList, error) {
+	if err := m.validateCertificates(signingIdentity.Certificate, certs...); err != nil {
 		return nil, err
 	}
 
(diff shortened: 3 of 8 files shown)


pkg/configtx/consortiums_test.go
```diff
@@ -945,6 +945,65 @@ func TestGetConsortiums(t *testing.T) {
...
+	consortiumGroup, err := baseConsortiumChannelGroup(t)
+	gt.Expect(err).NotTo(HaveOccurred())
+
+	config := &cb.Config{
+		ChannelGroup: consortiumGroup,
+	}
+
+	org1ConfigGroup, err := getConsortiumOrg(config, "Consortium1", "Org1")
+	gt.Expect(err).NotTo(HaveOccurred())
+	gt.Expect(org1ConfigGroup).To(Equal(config.ChannelGroup.Groups[ConsortiumsGroupKey].Groups["Consortium1"].Groups["Org1"]))
```
cendhu:
```suggestion
	require.NoError(t, err)
	expectedConfigGroup := config.ChannelGroup.Groups[ConsortiumsGroupKey].Groups["Consortium1"].Groups["Org1"]
	require.Equal(t, expectedConfigGroup, org1ConfigGroup)
```
The above-suggested changes definitely do not hurt my eyes. 

What is the reasoning behind using `Gomega` over the `testify` package for unit tests? 

Quote the Gomega website:

> Gomega is a matcher/assertion library. It is best paired with the Ginkgo BDD test framework, but can be adapted for use in other contexts too.

Quote from ginkgo website

> Ginkgo is a Go testing framework built to help you efficiently write expressive and comprehensive tests using Behavior-Driven Development (“BDD”) style. It is best paired with the Gomega matcher library but is designed to be matcher-agnostic.

I understand the usage of Gomega with the ginkgo as we can easily search for a particular string in the stdout or stderr. This is good for integration tests of a distributed system. Why are we applying Gomega even for the UT? what is the advantage when it is hard to read the function chaining? 

Further, the testify seems to be more popular than Gomega https://go.libhunt.com/compare-testify-vs-gomega
Using testify would certainly be good for golang developers who try to contribute to Fabric. 

Comparison analysis of the golang testing framework: https://www.slideshare.net/DushyantBhalgami/comparative-analysis-of-golang-testing-frameworks 

Let's see what packages are used by existing golang projects:

1. https://github.com/etcd-io/etcd -- reflect and basic if-then-else with `t.Errorf `or `t.Fatal` from `testing` package.
2. https://github.com/docker/docker-ce -- `gotest.tools/v3/assert `and `gotest.tools/v3/assert/cmp`
3. https://github.com/hashicorp/consul -- reflect basic if-then-else with `t.Errorf` or `t.Fatal` from `testing` package.
4. https://github.com/kubernetes/kubernetes -- a combination of `github.com/stretchr/testify/assert` and basic if-then-else with `t.Errorf` or `t.Fatal `from `testing` package
5. https://github.com/golang/go --basic if-then-else with `t.Errorf` and `t.Fatal` from `testing` package
6. https://github.com/pingcap -- a combination of own assertion package `github.com/pingcap/check` and basic if-then-else with t.Errorf or t.Fatal from `testing` package
7. https://github.com/cockroachdb/cockroach -- `reflect` and basic if-then-else with `t.Errorf `or `t.Fatal`. from `testing` package

Looks like only Kubernetes uses testify and no one uses `Gomega`.  Docker and pingcap use some assertion library. 

We need to either continue with `testify` or follow basic  basic if-then-else with `t.Errorf `or `t.Fatal` from `testing` package similar to what many popular projects have done. But, PLEASE do not use Gomega for the unit test.

wlahti: While you make many valid points, I'm not sure this is the place to raise your concerns over the use of Gomega in unit tests. Perhaps you can raise these points in the fabric-maintainers channel on rocketchat if you feel strongly that we shouldn't use Gomega in any unit tests? For the time being, we've decided to use Gomega for our matchers in this package and will continue to do so to keep things consistent throughout the package. That is the same philosophy I've seen adopted in terms of mocks/fakes in fabric: stick with one type (counterfeiter v. mockery, for example) for the entire package and follow the precedent of any package you're modifying for a new feature / bug fix / etc.

cendhu: I don't want to drag it either to rocket chat or #fabric-dev. This is my code review asking reasoning for using the Gomega when there are simple ways to do the same assertion. In the ledger packages, we use simple assertions. You have provided the reasoning. Hence, please feel free to ignore my comments.  

Maybe I should ask Fabric Architect (@sykesm) that 

1. why are we not keeping UT consistent across Fabric?
2. why Gomega that is best suited with ginkgo is used for the UT when it is specifically designed for integration test with CLI?
3. why there is no guideline on this?  
4. why most of the famous existing golang projects use a simple UT framework while we have inconsistent and complex framework?
5. Do we really need `gt.Expect(org1ConfigGroup).To(BeNil())` when we can have a easily readable code such as `require.Nil(t, org1ConfigGroup)` or `if org1ConfigGroup != nil {t.Errorf(...)}`?

I am not asking for answers here. I am just asking you to think about it. If you think my points are valid, please instruct the team to avoid using Gomega in UT. Otherwise, let's use whatever we want.

caod123: Not to speak on behalf of @sykesm or anything but we've discussed these points before in our standups/IPMs so I'll try to provide some insight on our reasoning. Simply put, I personally prefer Gomega style matchers over testify's (and I believe a majority of our squad does as well), but that's entirely my opinion. It reads better left to right like a spoken sentence in English. I wouldn't say that Gomega should explicitly be used with integration tests. Although we don't tend to use them as much in the UTs, I much rather prefer Gomega's handling for asynchronous assertions from both an integration test standpoint and when applicable in a UT point of view. In terms of consistency, there are already a couple of other places in the Fabric UT that do a mixed approach using Gomega matchers without Ginkgo. Ultimately though, because this package won't live in Fabric, it's not as clear cut as to any standardization patterns with existing Fabric code. I understand your concerns and do think maybe we should take a step back at some point and discuss at a higher level some standardization patterns/general guidelines (both for Fabric and any additional libraries we create repos for), but for now we will continue to move with Gomega matchers as the config work is nearly wrapping up in its current iteration and putting effort on converting the assertions would hold up the next set of priorities we have coming up.

cendhu: Thanks @caod123 for the response. 

1. I work remotely and sorry for not knowing about earlier discussions on UT framework.
2. In the ledger squad, we have decided to stick to `assert/require` as Gomega is not used by any of the famous projects that I mentioned above. We aren't using it because of that but we decided to follow a simple `if-then-else` or `assert/require` which can be read by anyone without having prior knowledge on the framework used. We prefer not to put additional learning burden for the new developers. Further Gomega follows JAVA method chaining style such as `user.setName("skptricks").setAge(22).getUserDetails();` and it does not look like a golang way to doing things to me. 
3. Testify does have support for testing with asynchronous operations. I recently used `assert.Eventually` and `assert.Never` in one of the PR. In fact, the implementation logic seems to be similar:
**Gomega**: https://github.com/onsi/gomega/blob/master/internal/asyncassertion/async_assertion.go#L105 
**testify**: https://github.com/stretchr/testify/blob/master/assert/assertions.go#L1628

cendhu: > Ultimately though, because this package won't live in Fabric. but for now we will continue to move with Gomega matchers as the config work is nearly wrapping up in its current iteration and putting effort on converting the assertions would hold up the next set of priorities we have coming up.

I agree. My goal is not to create a burden for the ongoing work. For ledger refactoring work, I have been looking at many famous open golang projects, to see what they have used, and then analyze why they have followed a certain approach over another. One of them was related to the testing framework as well. There is no single golden golang project that we can refer too. While golang repo itself is a good reference, it is more of a library repo. Hence, I am looking at multiple repos and try to identify reasoning for choosing one over the other. 

> and putting effort on converting the assertions would hold up the next set of priorities we have coming up.

Thank you much @caod123  for at least giving a consideration (even if changes does not happen after some discussion).

caod123:
@hellosmallstone can you rebase and fix the unit test errors? Will try to merge this asap afterwards since it's been open for some time

wuxuer:
@caod123 Sorry for this. I have rebased and fixed the unit test errors and all checks have passed. Please review it. Thanks very muck.

pkg/configtx/consortiums.go
```diff
@@ -244,3 +244,21 @@ func implicitMetaAnyPolicy(policyName string) (*standardConfigPolicy, error) {
 		value: implicitMetaPolicy,
 	}, nil
 }
+
+// getConsortiumOrg returns the organization config group for a consortium org in the
+// provided config. It returns an error if a named consortium or a consortium org was not found
+// in the config with the specified name.
```
caod123: We've actually decided to no longer return an error if the consortium or consortium org is not found. This will lead to panicking if they don't exist but we figured that behavior was fine for the purposes of easily chaining API calls together as part of [FAB-17744](https://jira.hyperledger.org/browse/FAB-17744). Instead we've been adding a node to the function comments to indicate that it will panic if not provided.

pkg/configtx/consortiums.go
```diff
@@ -244,3 +244,21 @@ func implicitMetaAnyPolicy(policyName string) (*standardConfigPolicy, error) {
...
+func getConsortiumOrg(config *cb.Config, consortiumName string, orgName string) (*cb.ConfigGroup, error) {
+	consortiumsGroup := config.ChannelGroup.Groups[ConsortiumsGroupKey].Groups
+	consortiumGroup, ok := consortiumsGroup[consortiumName]
+	if !ok {
+		return nil, fmt.Errorf("consortium %s does not exist in channel config", consortiumName)
+	}
```
caod123: This check is no longer needed per the point above (as well as the returned error).

pkg/configtx/consortiums.go
```diff
@@ -244,3 +244,21 @@ func implicitMetaAnyPolicy(policyName string) (*standardConfigPolicy, error) {
...
+
+	org, ok := consortiumGroup.Groups[orgName]
+	if !ok {
+		return nil, fmt.Errorf("org %s does not exist in channel config", orgName)
+	}
```
caod123: Same with this

pkg/configtx/msp.go
```diff
@@ -607,6 +633,54 @@ func pemEncodePKCS8PrivateKey(priv crypto.PrivateKey) ([]byte, error) {
 	return pem.EncodeToMemory(&pem.Block{Type: "PRIVATE KEY", Bytes: privBytes}), nil
 }
 
+// UpdateConsortiumMSP updates the MSP config for the specified consortium org group.
+func (c *ConfigTx) UpdateConsortiumMSP(updatedMSP MSP, consortiumName string, orgName string) error {
```
caod123: We've decided to standardize on naming these `SetConsortiumMSP`, also this should probably be `SetConsortiumOrgMSP` but we use the same naming convention for `SetOrdererMSP` and `SetApplicationMSP` so I'm fine leaving it as `SetConsortiumMSP`.

pkg/configtx/msp.go
```diff
@@ -607,6 +633,54 @@ func pemEncodePKCS8PrivateKey(priv crypto.PrivateKey) ([]byte, error) {
...
+	err = setMSPConfigForConsortium(c.UpdatedConfig(), updatedMSP, consortiumName, orgName)
+	if err != nil {
+		return err
+	}
+
+	return nil
+}
+
+// UpdateOrdererMSP updates the MSP config for the specified orderer org group.
+func (c *ConfigTx) UpdateOrdererMSP(updatedMSP MSP, orgName string) error {
```
caod123: Should be `SetOrdererMSP` now

pkg/configtx/msp.go
```diff
@@ -607,6 +633,54 @@ func pemEncodePKCS8PrivateKey(priv crypto.PrivateKey) ([]byte, error) {
...
+	if err != nil {
+		return err
+	}
+
+	return nil
+}
+
 // UpdateApplicationMSP updates the MSP config for the specified application
 // org group.
 func (c *ConfigTx) UpdateApplicationMSP(updatedMSP MSP, orgName string) error {
```
caod123: Not part of your commit but might as well update this to `SetApplicationMSP` as well

caod123 (changes requested):

pkg/configtx/consortiums.go
wuxuer: I have removed the error info for function getApplicationOrg, getOrdererOrg, and getConsortiumOrg.

pkg/configtx/msp.go
wuxuer: Yeah, I have replaced Update with Set for the three functions.

caod123 (approved):

PR #1115, opened 2020-04-17
Warn if a docker network may cause issues on macOS

sykesm:
The default docker local network address pool includes addresses that are commonly used on home networks. This configuration can prevent containers from accessing the network host on the LAN addresses discovered by the peer. This often surfaces in integration tests as chaincode registration failures with a "no route to host" message.

This change attempts to discover and highlight the broken configuration so that it can be addressed. (pun, heh)

The network pools can be configured by adding some JSON to the Docker daemon configuration to prevent the overlaps. The comment near the warning message describes the syntax.

### DIFF
--- a/integration/nwo/network.go
+++ b/integration/nwo/network.go
@@ -11,9 +11,11 @@ import (
 	"fmt"
 	"io"
 	"io/ioutil"
+	"net"
 	"os"
 	"os/exec"
 	"path/filepath"
+	"runtime"
 	"sort"
 	"strconv"
 	"strings"
@@ -662,13 +664,7 @@ func (n *Network) GenerateConfigTree() {
 // written to ${rootDir}/${Channel.Name}_tx.pb.
 func (n *Network) Bootstrap() {
 	if n.DockerClient != nil {
-		_, err := n.DockerClient.CreateNetwork(
-			docker.CreateNetworkOptions{
-				Name:   n.NetworkID,
-				Driver: "bridge",
-			},
-		)
-		Expect(err).NotTo(HaveOccurred())
+		n.createDockerNetwork()
 	}
 
 	sess, err := n.Cryptogen(commands.Generate{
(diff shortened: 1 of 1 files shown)


sykesm:
Hit flake [FAB-17762](https://jira.hyperledger.org/browse/FAB-17762) multiple times.

cendhu:
#1092 should help for now.

sykesm:
> #1092 should help for now.

I'm not sure I see how. The change in that PR modifies timing but does not appear to address the root cause. I'll go comment over there.

integration/nwo/network.go
```diff
@@ -704,6 +700,91 @@ func (n *Network) Bootstrap() {
...
+//
+// https://github.com/moby/libnetwork/blob/1a17fb36132631a95fe6bb055b91e24a516ad81d/ipamutils/utils.go#L18-L20
+//
+// Docker can be configured to use different addresses by addding an
+// appropriate default-address-pools configuration element to "daemon.json".
+//
+// For example:
+//   "default-address-pools":[
+//       {"base":"172.80.0.0/16","size":24},
+//       {"base":"172.81.0.0/16","size":24}
```
mastersingh24, line 731: Just a nit, but wouldn't it be better to pick from the officially reserved ranges just in case someone looks at the comment?
```suggestion
//       {"base":"172.16.0.0/16","size":16},
//       {"base":"172.17.0.0/16","size":16}
```
but I guess those are in the default.  You can also override the bridge network with the "bip" config option but again I guess this is just a comment

integration/nwo/network.go
```diff
@@ -704,6 +700,91 @@ func (n *Network) Bootstrap() {
...
+
+// checkDockerNetworks attempts to discover if the docker network configuration
+// will prevent a container from accessing the host. This commonly happens when
+// using Docker for Mac on home networks because most home routers provide DHCP
+// addresses from 192.168.1.0/24 and the default Docker daemon config uses
+// 192.168.0.0/20 as one of the default local address pools.
+//
+// https://github.com/moby/libnetwork/blob/1a17fb36132631a95fe6bb055b91e24a516ad81d/ipamutils/utils.go#L18-L20
+//
+// Docker can be configured to use different addresses by addding an
```
mastersingh24, line 725:
```suggestion
// Docker can be configured to use different addresses by adding an
```

mastersingh24 (approved):

integration/nwo/network.go
sykesm: > Just a nit, but wouldn't it be better to pick from the officially reserved ranges just in case someone looks at the comment?

They're from the reserved range of 172.16.0.0/12. So, I'm not sure why it's wrong? The example is showing how to get 512 different /24 networks for local use that do no overlap with common prefixes.

integration/nwo/network.go
sykesm: I'll push a spelling fix on top.

caod123 (approved):

integration/nwo/network.go
sykesm: You are correct and my math is not. I'll address this with the spelling fix.

PR #1116, opened 2020-04-17
Remove concurrent session creation from raft suite

sykesm:
The session creation commands in nwo.Network were to be used with ginkgo and were not structured for concurrent access. However, one integration test suite that pushes the boundaries and starts sessions in parallel using go routines in the test suites.

Removing the go routine prevents a concurrent map access in the network, simplifies the test, and does not seem to impact the execution time of the suite. (Before and after execution were both over 9 minutes and the version without concurrency was 4 seconds faster locally.)

### DIFF
--- a/integration/raft/config_test.go
+++ b/integration/raft/config_test.go
@@ -17,7 +17,6 @@ import (
 	"os"
 	"path/filepath"
 	"strings"
-	"sync"
 	"syscall"
 	"time"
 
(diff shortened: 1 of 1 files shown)


yacovm (approved):
> However, one integration test suite that pushes the boundaries and starts sessions in parallel using go
> routines in the test suites.

Boundaries were meant to be broken you know...

PR #1119, opened 2020-04-17
 [FAB-17762] create index in deterministic order to avoid flake

sykesm:
ProcessIndexesForChaincodeDeploy returns an error on the first failure to create an index but, as the indexes are passed in a map, the order of the operations is non-deterministic.

TestHandleChaincodeDeployErroneousIndexFile relies on successful creation of a good index to support the query performed by the test and an error from ProcessIndexesForChaincodeDeploy related to an attempt to create an index from invalid JSON. As both documents were passed to a single call to ProcessIndexesForChaincodeDeploy, there was a 50/50 chance that the "good" index would not be processed.

This change uses two calls to ensure that the "good" index is created before the "bad" one is tested. While this addresses the test flake, there may be future changes to the production implementation of ProcessIndexesForChaincodeDeploy from the ledger team.


cendhu:
LGTM. Thanks. The production code would be fixed in [FAB-17762](https://jira.hyperledger.org/browse/FAB-17762)

lindluni (approved):

PR #1121, opened 2020-04-17
[LEDGER] make couchdb index processor deterministic

cendhu:
#### Type of change

- Bug fix

#### Description

We need to satisfy two requirements while processing the index files.

1. R1: all valid indexes should be processed.
2. R2: the order of index creation must be the same in all peers. For example, if the user passes two index files with the same index name but different index fields and we process these files in different orders in different peers, each peer would have different indexes (as one index definition could replace another if the index names are the same).

To satisfy R1, we log the error and continue to process the next index file. Note that we were logging the error earlier too but in [HandleChaincodeDeploy() ](https://github.com/hyperledger/fabric/blob/master/core/ledger/kvledger/txmgmt/privacyenabledstate/common_storage_db.go#L318) and didn't process rest of the index file as soon as an error occurred. Now, we log at the `statecouchdb` package itself and continue to process the rest of valid indexes. Further, we continue to log error in [HandleChaincodeDeploy() ](https://github.com/hyperledger/fabric/blob/master/core/ledger/kvledger/txmgmt/privacyenabledstate/common_storage_db.go#L318) as the error would be returned from `statecouchdb` package when the database could not be created for the chaincode. 

To satisfy R2, we sort the indexFilesData map based on the filenames and process each index as per the sorted order.

#### Additional details

Though the R2 was not satisfied earlier too (because a map was used in Extract Files util in the ccprovider.), we found this as #966 introduced a test flake. 

Given that we are fixing both R1 and R2, it would not break any backward compatibility when we do a rolling upgrade from one version to another. Only when a chaincode is deployed newly, the index files present in the deployment package would be processed. 

In this PR, we tackle both R1 and R2. Now the question is, do we really need a fake-logger to test the output of error logs? 

#### Related issues

[FAB-17762](https://jira.hyperledger.org/browse/FAB-17762)

### DIFF
--- a/core/ledger/kvledger/txmgmt/statedb/statecouchdb/statecouchdb.go
+++ b/core/ledger/kvledger/txmgmt/statedb/statecouchdb/statecouchdb.go
@@ -10,6 +10,7 @@ import (
 	"bytes"
 	"context"
 	"encoding/json"
+	"sort"
 	"sync"
 
 	"github.com/hyperledger/fabric/common/flogging"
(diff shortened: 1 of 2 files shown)


core/ledger/kvledger/txmgmt/statedb/statecouchdb/statecouchdb.go
```diff
@@ -256,10 +257,30 @@ func (vdb *VersionedDB) ProcessIndexesForChaincodeDeploy(namespace string, index
...
+	var indexFilesName []string
+	for fileName := range indexFilesData {
+		indexFilesName = append(indexFilesName, fileName)
+	}
+	sort.Strings(indexFilesName)
+	for _, fileName := range indexFilesName {
+		_, err = db.createIndex(string(indexFilesData[fileName]))
+		switch {
+		case err != nil:
+			logger.Errorf("error creating index from file [%s] for chaincode [%s] on channel [%s]: %s",
```
manish-sethi: better to use %+v for logging errors, so stack trace also gets logged.

core/ledger/kvledger/txmgmt/statedb/statecouchdb/statecouchdb_test.go
```diff
@@ -723,27 +723,92 @@ func TestHandleChaincodeDeployErroneousIndexFile(t *testing.T) {
...
+		if err != nil {
+			return false
+		}
+		return true
+	}
+	assert.Never(t, queryUsingIndex, 2*time.Second, 100*time.Millisecond, "error should have occurred as there is no index on color field")
+
+	// due to the change in order, the query using the color index would pass
+	queryString = `{"selector":{"owner":"fred"}, "sort": [{"color": "desc"}]}`
+	queryUsingIndex = func() bool {
```
manish-sethi: This pretty much seems the repeat of what is already tested in the previous block. Did I miss anything?

cendhu: True. Removed it.

core/ledger/kvledger/txmgmt/statedb/statecouchdb/statecouchdb.go
cendhu: Done.

manish-sethi (approved):

PR #1047, opened 2020-04-09
PKCS11: Validate session and get new if invalid - master

mrshah-at-work:
Signed-off-by: Mihir Shah <mrshah@us.ibm.com>

#### Type of change

- Bug fix

#### Description

Making the same change from https://github.com/hyperledger/fabric/pull/1022 into master

As described in [JIRA FAB-17722](https://jira.hyperledger.org/browse/FAB-17722):
Currently the pkcs11 code in fabric's bccsp module is set to have a session cache and reuse sessions if available in cache. If a session goes bad (due to connection issues with HSM), the session is not evicted from cache and will be reused. If all sessions go bad, the client will never be able to recover and keep using bad sessions.

#### Additional details

This code change:
- Moves creation of a new session to a new function `CreateSession`
- Validates the existing session by calling `GetSessionInfo`
- If found invalid, creates a new session and expects the old one to be garbage collected.
- The current test in pkcs11_test.go `TestPKCS11GetSession`, will test the new function as well. Let me know if I should add additional test to test `CreateSession` function.

#### Related issues

https://github.com/hyperledger/fabric/pull/1022
https://jira.hyperledger.org/browse/FAB-17722

PR #1124, opened 2020-04-18
Return Error From PKCS11 CreateSession

lindluni:
The current implementation of BCCSP performs a fatal logging function in the event it cannot open a session with the HSM. This is fine during startup when it's expected an exit might occur, but the createSession function is called during the standard code flow, i.e., when the cryptoServiceProvider's session pool is already completely in use (by default 10 sessions) a new session is created. So it is possible while running post-setup for the fatal log to occur when an expected error occurs, i.e., the HSM has reached its maximum session handles. Instead of issuing a fatal log call, we should return the error as it's probable the error is recoverable, or outside the control of Fabric itself (is a problem with the HSM).

Note: #1047 appears to have broke the tests for these function changes. If you run the PKCS11 tests using `go test -tags pkcs11 ./bccsp/pkcs11/...` the tests fail. When using `make unit-test` the failure doesn't occur (though it should, the test asserts that a panic occurs, but the test no longer panics due to the change Mihir introduced)

#### Type of change

#### Description

Prevents production environments from issuing a fatal logging call in a recoverable situation

### DIFF
--- a/bccsp/pkcs11/pkcs11.go
+++ b/bccsp/pkcs11/pkcs11.go
@@ -17,6 +17,7 @@ import (
 	"sync"
 
 	"github.com/miekg/pkcs11"
+	"github.com/pkg/errors"
 	"go.uber.org/zap/zapcore"
 )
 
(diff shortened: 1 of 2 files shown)


bccsp/pkcs11/pkcs11.go
```diff
@@ -51,46 +52,37 @@ func loadLib(lib, pin, label string) (*pkcs11.Ctx, uint, *pkcs11.SessionHandle,
...
 	}
 
-	session := createSession(ctx, slot, pin)
-
-	logger.Debugf("Created new pkcs11 session %+v on slot %d\n", session, slot)
-
-	if pin == "" {
-		return nil, slot, nil, fmt.Errorf("No PIN set")
-	}
-	err = ctx.Login(session, pkcs11.CKU_USER, pin)
```
denyeart, line 64: Why is this logic removed? It's not mentioned in the PR description... if it must be removed I'd suggest remove in a separate PR or add the motivation to the PR description.

lindluni: It's already performed in `createSession` it's accidentally duplicated code, removing it is inconsequential, the refactoring in #1047 caused this to become redundant and should have been removed there

bccsp/pkcs11/pkcs11.go
```diff
@@ -51,46 +52,37 @@ func loadLib(lib, pin, label string) (*pkcs11.Ctx, uint, *pkcs11.SessionHandle,
...
+func (csp *impl) getSession() (session pkcs11.SessionHandle, err error) {
 	select {
 	case session = <-csp.sessions:
-		_, err := csp.ctx.GetSessionInfo(session)
+		_, err = csp.ctx.GetSessionInfo(session)
 		if err != nil {
 			logger.Warningf("Get session info failed [%s], closing existing session and getting a new session\n", err)
 			csp.ctx.CloseSession(session)
-			session = createSession(csp.ctx, csp.slot, csp.pin)
+			session, err = createSession(csp.ctx, csp.slot, csp.pin)
```
caod123, line 73: Keep in mind this error will overwrite the one on line 69 for the purposes of the return on 82. Is there ever a case where we would not want to attempt creating a new Session if GetSessionInfo fails and we would want to return early? I'm guessing that's largely dependent on error handling specific errors from `GetSessionInfo`.

bccsp/pkcs11/pkcs11.go
```diff
@@ -129,7 +119,10 @@ func (csp *impl) returnSession(session pkcs11.SessionHandle) {
 // Look for an EC key by SKI, stored in CKA_ID
 func (csp *impl) getECKey(ski []byte) (pubKey *ecdsa.PublicKey, isPriv bool, err error) {
 	p11lib := csp.ctx
-	session := csp.getSession()
+	session, err := csp.getSession()
```
caod123, line 122: Do the callers of this function (and the ones below) need to care about getSession erroring and returning early? Or does nothing change in the caller's logic?

caod123 (commented):
General questions for better understanding. I'm more or less fine with the changes otherwise.

bccsp/pkcs11/pkcs11.go
lindluni: No the err on line 69 is essentially acting as an `ok` flag. Its used internally only to denote the session was invalid and we should attempt to close and open a new session to replace it (we don't attempt to derive the cause for the invalid session, so the error is simply logged and disgarded). The error propagated back to the caller should be the failure to create a new session, not the fact the the original session was invalid at some point in the flow.

bccsp/pkcs11/pkcs11.go
lindluni: My general assumption was, because there already existed error paths in these functions, there was proper error handling in the caller code. The fact is, if the session is bad, the subsequent calls would fail. You never had to worry about that happening before because if the session failed, the peer process exited.

Probably worth me checking into the callers and making sure things are handled properly up the stack

caod123: Yeah that's what I figured so my comment was more of a consideration. Otherwise LGTM

lindluni: It's a good consideration though, since we don't have an integration test on this code path, its difficult to say all of these pieces function perfectly well together, and something like this doesn't cause unintended consequences

caod123 (approved):

PR #1128, opened 2020-04-20
[FAB-17745] Move orderer types to orderer subpackage

stephyee:
#### Type of change

#### Description

Move Orderer related types to an orderer subpackage.

#### Related issues

[FAB-17745](https://jira.hyperledger.org/browse/FAB-17745)

### DIFF
--- a/pkg/configtx/constants.go
+++ b/pkg/configtx/constants.go
@@ -33,19 +33,11 @@ const (
 	// application, and orderer levels and this constant is used for all three.
 	CapabilitiesKey = "Capabilities"
 
-	ordererAdminsPolicyName = "/Channel/Orderer/Admins"
-
 	// EndpointsKey is the common.ConfigValue key name for the Endpoints message in the OrdererOrgGroup.
 	EndpointsKey = "Endpoints"
 
 	// MSPKey is the key for the MSP definition in orderer groups.
 	MSPKey = "MSP"
-
-	// BatchSizeKey is the common.ConfigValue type key name for the BatchSize message.
-	BatchSizeKey = "BatchSize"
-
-	// BatchTimeoutKey is the common.ConfigValue type key name for the BatchTimeout message.
-	BatchTimeoutKey = "BatchTimeout"
 
 	// AdminsPolicyKey is the key used for the admin policy.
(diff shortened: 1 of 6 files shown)


pkg/configtx/constants.go
```diff
@@ -111,9 +85,5 @@ const (
 	// SignaturePolicyType is the 'Type' string for signature policies.
 	SignaturePolicyType = "Signature"
 
-	// ConsensusStateNormal indicates normal orderer operation.
-	ConsensusStateNormal ConsensusState = "STATE_NORMAL"
-
-	// ConsensusStateMaintenance indicates the orderer is in consensus type migration.
-	ConsensusStateMaintenance ConsensusState = "STATE_MAINTENANCE"
+	ordererAdminsPolicyName = "/Channel/Orderer/Admins"
```
caod123, line 88: consider moving this too (`orderer.AdminsPolicyName`)? Although I can see it staying as well

stephyee: I did consider this, however this unexported constant is used in multiple files (other than orderer subpackage and examples).

pkg/configtx/orderer.go
```diff
@@ -17,11 +17,9 @@ import (
...
 	ob "github.com/hyperledger/fabric-protos-go/orderer"
 	eb "github.com/hyperledger/fabric-protos-go/orderer/etcdraft"
+	"github.com/hyperledger/fabric/pkg/configtx/orderer"
 )
 
-// ConsensusState defines the orderer mode of operation.
-type ConsensusState string
-
 // Orderer configures the ordering service behavior for a channel.
 type Orderer struct {
```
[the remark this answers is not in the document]
caod123: Matt suggested it would look better to namespace orderer subtypes in a separate package and I do agree it looks better now. The orderer type itself hasn't moved since it feels appropriate to namespace it as `configtx.Orderer` instead of `orderer.Orderer`. We're keeping a close look though on the upcoming work with chaining APIs and it may make sense to move more into the orderer subpackage as we progress through that but I think this is a good starting point for now

caod123 (approved):

PR #1138, opened 2020-04-21
pkg/configtx AnchorPeers() stop returning error when no anchor peers are defined

wlahti:
#### Type of change

- Bug fix

#### Description

Return nil slice and nil error when no anchor peers are defined (either because none ever existed or none remain after removing). 

#### Related issues

[FAB-17776](https://jira.hyperledger.org/browse/FAB-17776)

### DIFF
--- a/pkg/configtx/application_test.go
+++ b/pkg/configtx/application_test.go
@@ -552,7 +552,7 @@ func TestRemoveAnchorPeerFailure(t *testing.T) {
 	}
 }
 
-func TestAnchorPeer(t *testing.T) {
+func TestAnchorPeers(t *testing.T) {
 	t.Parallel()
 
 	gt := NewGomegaWithT(t)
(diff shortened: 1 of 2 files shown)


pkg/configtx/application_test.go
```diff
@@ -567,19 +567,36 @@ func TestAnchorPeer(t *testing.T) {
...
+	expectedAnchorPeer := Address{Host: "host1", Port: 123}
 	err = c.AddAnchorPeer("Org1", expectedAnchorPeer)
 	gt.Expect(err).NotTo(HaveOccurred())
 
-	anchorPeers, err := c.AnchorPeers("Org1")
+	anchorPeers, err = c.AnchorPeers("Org1")
 	gt.Expect(err).NotTo(HaveOccurred())
-	gt.Expect(len(anchorPeers)).To(Equal(1))
+	gt.Expect(anchorPeers).To(HaveLen(1))
 	gt.Expect(anchorPeers[0]).To(Equal(expectedAnchorPeer))
```
caod123, line 587: I'm a little hesitant with the basis for this test since it revolves around the fact that we use the same config for both the original and updated on line 570. In a real world scenario we could never actually do this since adding/removing an anchor peer only modifies the updated whereas the retriever only uses the original.

wlahti: I was working on changing just that (in what will now be a follow-on commit/PR) when I remembered that we wanted to update the functionality of this retrieval function, hence this JIRA and PR. :) I'll have that PR out later today.

caod123 (approved):
Merging under the assumption the tests will be updated in a followup PR

PR #1140, opened 2020-04-21
Improve lifecycle error message for committed sequence

denyeart:
#### Type of change

- Improvement to error message

#### Description

If you attempt to approveformyorg again for a committed chaincode definition sequence,
for example to update chaincode package on a peer, but you pass the incorrect
chaincode definition parameters, the error message was not clear.

This change updates the error message to something that the admin can likely
troubleshoot themselves.

Old message:
Error: proposal failed with status: 500 - failed to invoke backing implementation
of 'ApproveChaincodeDefinitionForMyOrg': attempted to define the current sequence
(2) for namespace marbles5, but: Version '3' != '4'

New message:
Error: proposal failed with status: 500 - failed to invoke backing implementation
of 'ApproveChaincodeDefinitionForMyOrg': attempted to redefine the current
committed sequence (3) for namespace marbles5 with different parameters:
expected Version '3' does not match passed Version '4'

### DIFF
--- a/core/chaincode/lifecycle/lifecycle.go
+++ b/core/chaincode/lifecycle/lifecycle.go
@@ -114,15 +114,15 @@ type ChaincodeParameters struct {
 func (cp *ChaincodeParameters) Equal(ocp *ChaincodeParameters) error {
 	switch {
 	case cp.EndorsementInfo.Version != ocp.EndorsementInfo.Version:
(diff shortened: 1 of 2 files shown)


wlahti (approved):
Definitely clearer to me what the cause of the error is with this wording. Thanks, Dave!

C0rWin (approved):

PR #1141, opened 2020-04-21
[FAB-13460] Add Support for TLS1.3

lindluni:
Add support for TLS1.3 and adds a tests to ensure these versions work on the comm stack.

### DIFF
--- a/internal/pkg/comm/client.go
+++ b/internal/pkg/comm/client.go
@@ -71,7 +71,9 @@ func (client *GRPCClient) parseSecureOptions(opts SecureOptions) error {
 
 	client.tlsConfig = &tls.Config{
 		VerifyPeerCertificate: opts.VerifyCertificate,
-		MinVersion:            tls.VersionTLS12} // TLS 1.2 only
+		MinVersion:            tls.VersionTLS12,
+		MaxVersion:            tls.VersionTLS13,
+	}
 	if len(opts.ServerRootCAs) > 0 {
 		client.tlsConfig.RootCAs = x509.NewCertPool()
 		for _, certBytes := range opts.ServerRootCAs {
--- a/internal/pkg/comm/creds.go
+++ b/internal/pkg/comm/creds.go
@@ -35,9 +35,9 @@ func NewServerTransportCredentials(
 	// NOTE: unlike the default grpc/credentials implementation, we do not
 	// clone the tls.Config which allows us to update it dynamically
 	serverConfig.config.NextProtos = alpnProtoStr
-	// override TLS version and ensure it is 1.2
+	// override TLS version and ensure it is 1.2 or 1.3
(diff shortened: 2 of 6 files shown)


internal/pkg/comm/client.go
```diff
@@ -71,7 +71,9 @@ func (client *GRPCClient) parseSecureOptions(opts SecureOptions) error {
 
 	client.tlsConfig = &tls.Config{
 		VerifyPeerCertificate: opts.VerifyCertificate,
-		MinVersion:            tls.VersionTLS12} // TLS 1.2 only
+		MinVersion:            tls.VersionTLS12,
+		MaxVersion:            tls.VersionTLS13,
```
ashku, line 75: NoOps here , but you could have 0 also as MaxVersion value. To me , it seems logical to have 0 consideration the situation when TLS moves to say , 1.4 , we do not have to change this field.

internal/pkg/comm/creds.go
```diff
@@ -35,9 +35,9 @@ func NewServerTransportCredentials(
 	// NOTE: unlike the default grpc/credentials implementation, we do not
 	// clone the tls.Config which allows us to update it dynamically
 	serverConfig.config.NextProtos = alpnProtoStr
-	// override TLS version and ensure it is 1.2
+	// override TLS version and ensure it is 1.2 or 1.3
 	serverConfig.config.MinVersion = tls.VersionTLS12
-	serverConfig.config.MaxVersion = tls.VersionTLS12
+	serverConfig.config.MaxVersion = tls.VersionTLS13
```
ashku, line 40: same comment as above.

ashku (changes requested):
One overall comment : The list of default Cipher Suite contains one Cipher more than what TLS 1.3 supports. Not knowing , how TLS 1.3 loads Cipher 

1. Does it load from its engine or
1. Honor defaultConfig setup in the code.

If it is latter , then it might throw error saying Cipher XXXXX not supported.

To be on safer side , we need to set cipher list in defaultCipherSuite matching TLS 1.3 cipher IMO.

internal/pkg/comm/client.go
btl5037: > One overall comment : The list of default Cipher Suite contains one Cipher more than what TLS 1.3 supports. Not knowing , how TLS 1.3 loads Cipher
> 
> 1. Does it load from its engine or
> 2. Honor defaultConfig setup in the code.
> 
> If it is latter , then it might throw error saying Cipher XXXXX not supported.
> 
> To be on safer side , we need to cipher list in defaultCipherSuite matching TLS 1.3 cipher IMO.

Yea, I'm working on that one now, I didn't run the tests locally, so I didn't get the failed cipher until it finished testing in CI

lindluni:
This is a bigger surface area than I at first thought, I am going to close for now and will reopen after the discussion in Jira.

PR #1143, opened 2020-04-21
Update pkg/configtx tests to use New to create ConfigTx

wlahti:
#### Type of change

- Bug fix
- Test update

#### Description

- Use New() to create ConfigTx in tests to ensure the functionality works as expected (i.e. getters retrieve from OriginalConfig() while setters set in UpdatedConfig() ). 
- Some miscellaneous cleanup.

#### Related issues

[FAB-17775](https://jira.hyperledger.org/browse/FAB-17775)

### DIFF
--- a/pkg/configtx/config.go
+++ b/pkg/configtx/config.go
@@ -317,6 +317,10 @@ func setValue(cg *cb.ConfigGroup, value *standardConfigValue, modPolicy string) 
 		return fmt.Errorf("marshaling standard config value '%s': %v", value.key, err)
 	}
 
+	if cg.Values == nil {
+		cg.Values = map[string]*cb.ConfigValue{}
+	}
+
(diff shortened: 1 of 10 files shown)


pkg/configtx/config.go
```diff
@@ -317,6 +317,10 @@ func setValue(cg *cb.ConfigGroup, value *standardConfigValue, modPolicy string)
 		return fmt.Errorf("marshaling standard config value '%s': %v", value.key, err)
 	}
 
+	if cg.Values == nil {
+		cg.Values = map[string]*cb.ConfigValue{}
+	}
```
wlahti, line 322: This was one oddity related to switching to New() for the config. It has to do with working with the UpdatedConfig() as derived by using proto.Clone on the original config. Empty maps are converted to nil and the code below would panic without this.

caod123 (approved):

PR #1144, opened 2020-04-22
[FAB-13460] Add Support for TLS1.3

lindluni:
Add support for TLS1.3 and adds a tests to ensure these versions work on the comm stack.

### DIFF
--- a/internal/pkg/comm/client.go
+++ b/internal/pkg/comm/client.go
@@ -71,7 +71,8 @@ func (client *GRPCClient) parseSecureOptions(opts SecureOptions) error {
 
 	client.tlsConfig = &tls.Config{
 		VerifyPeerCertificate: opts.VerifyCertificate,
-		MinVersion:            tls.VersionTLS12} // TLS 1.2 only
+		MinVersion:            tls.VersionTLS12,
+	}
 	if len(opts.ServerRootCAs) > 0 {
 		client.tlsConfig.RootCAs = x509.NewCertPool()
 		for _, certBytes := range opts.ServerRootCAs {
--- a/internal/pkg/comm/creds.go
+++ b/internal/pkg/comm/creds.go
@@ -35,9 +35,7 @@ func NewServerTransportCredentials(
 	// NOTE: unlike the default grpc/credentials implementation, we do not
 	// clone the tls.Config which allows us to update it dynamically
 	serverConfig.config.NextProtos = alpnProtoStr
-	// override TLS version and ensure it is 1.2
 	serverConfig.config.MinVersion = tls.VersionTLS12
-	serverConfig.config.MaxVersion = tls.VersionTLS12
 	return &serverCreds{
 		serverConfig: serverConfig,
 		logger:       logger}
--- a/internal/pkg/comm/server_test.go
+++ b/internal/pkg/comm/server_test.go
@@ -635,17 +635,33 @@ func TestNewSecureGRPCServer(t *testing.T) {
 	_, err = invokeEmptyCall(testAddress, grpc.WithTransportCredentials(creds))
 	assert.NoError(t, err, "client failed to invoke the EmptyCall service")
 
+	// Test TLS versions which should be valid
 	tlsVersions := map[string]uint16{
+		"TLS12": tls.VersionTLS12,
+		"TLS13": tls.VersionTLS13,
+	}
+	for name, tlsVersion := range tlsVersions {
+		tlsVersion := tlsVersion
+
+		t.Run(name, func(t *testing.T) {
+			creds := credentials.NewTLS(&tls.Config{RootCAs: certPool, MinVersion: tlsVersion, MaxVersion: tlsVersion})
+			_, err := invokeEmptyCall(testAddress, grpc.WithTransportCredentials(creds), grpc.WithBlock())
+			assert.NoError(t, err)
+		})
+	}
+
+	// Test TLS versions which should be invalid
+	tlsVersions = map[string]uint16{
 		"SSL30": tls.VersionSSL30,
 		"TLS10": tls.VersionTLS10,
 		"TLS11": tls.VersionTLS11,
 	}
-	for name, version := range tlsVersions {
-		version := version
+	for name, tlsVersion := range tlsVersions {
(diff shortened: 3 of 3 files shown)


internal/pkg/comm/client_test.go
```diff
@@ -253,10 +253,9 @@ func TestNewConnection(t *testing.T) {
 			serverTLS: &tls.Config{
 				Certificates: []tls.Certificate{testCerts.serverCert},
 				ClientAuth:   tls.RequireAndVerifyClientCert,
-				MaxVersion:   tls.VersionTLS12, // https://github.com/golang/go/issues/33368
```
sykesm: Pretty sure this is no longer testing what it really means to test. IIUC, the grpc client needs to be coerced into reading from the server.

internal/pkg/comm/client.go
```diff
@@ -71,7 +71,9 @@ func (client *GRPCClient) parseSecureOptions(opts SecureOptions) error {
 
 	client.tlsConfig = &tls.Config{
 		VerifyPeerCertificate: opts.VerifyCertificate,
-		MinVersion:            tls.VersionTLS12} // TLS 1.2 only
+		MinVersion:            tls.VersionTLS12,
+		MaxVersion:            tls.VersionTLS13,
```
sykesm: Why explicitly set the max version instead of relying on the judgement of the tls package's calculation of max?

internal/pkg/comm/creds.go
```diff
@@ -35,9 +35,9 @@ func NewServerTransportCredentials(
 	// NOTE: unlike the default grpc/credentials implementation, we do not
 	// clone the tls.Config which allows us to update it dynamically
 	serverConfig.config.NextProtos = alpnProtoStr
-	// override TLS version and ensure it is 1.2
+	// override TLS version and ensure it is 1.2 or 1.3
 	serverConfig.config.MinVersion = tls.VersionTLS12
-	serverConfig.config.MaxVersion = tls.VersionTLS12
+	serverConfig.config.MaxVersion = tls.VersionTLS13
```
sykesm: Why explicitly set the max version instead of relying on the judgement of the tls package's calculation of max?

orderer/consensus/kafka/config.go
```diff
@@ -57,7 +57,7 @@ func newBrokerConfig(
 			Certificates: []tls.Certificate{keyPair},
 			RootCAs:      rootCAs,
 			MinVersion:   tls.VersionTLS12,
-			MaxVersion:   0, // Latest supported TLS version
```
sykesm: Why explicitly set the max version instead of relying on the judgement of the tls package's calculation of max?

lindluni:
This was meant to be a draft, the test you commented on I'm aware of the problem I have in it. But I'll take the other comments under advisement. My opinion there was, keeping it as it was (limiting the supported TLS types) and setting the max version prevents unintended consequences when upgrading go versions that support new TLS versions in the future. But removing the default limitation also reduces the maintenance we have to do. I'm ok with either method, but both were used and I chose the former. If we agree the latter is the way we want to go, I'll remove the max version.

internal/pkg/comm/client.go
lindluni: I just preserved what was here, the fact the comment says "TLS 1.2 only" when tLS1.2 was the max, this meant setting the min version also forced the max inherently.

Should I remove the Max, it will keep us from having to perform upkeep on it in the future as new TLS versions become available, but it also means we will blindly accept the consequences of any new version of TLS?

I prefer the explicit declaration of a max supported TLS version, but I'm indifferent on what we decide.

internal/pkg/comm/creds.go
lindluni: I just preserved what was here. Should I remove the Max, it will keep us from having to perform upkeep on it in the future as new TLS versions become available, but it also means we will blindly accept the consequences of any new version of TLS?

I prefer the explicit declaration of a max supported TLS version, but I'm indifferent on what we decide.

internal/pkg/comm/client.go
mastersingh24: We should be able to remove the MaxVersion as gRPC  now works with TLS 1.3.
We can probably even remove the MinVersion as well given gRPC requires TLS 1.2 (it requires ALPN) but no harm explicitly setting the MinVersion just in case

mastersingh24 (changes requested):
> This was meant to be a draft, the test you commented on I'm aware of the problem I have in it. But I'll take the other comments under advisement. My opinion there was, keeping it as it was (limiting the supported TLS types) and setting the max version prevents unintended consequences when upgrading go versions that support new TLS versions in the future. But removing the default limitation also reduces the maintenance we have to do. I'm ok with either method, but both were used and I chose the former. If we agree the latter is the way we want to go, I'll remove the max version.

Removing the MaxVersion should be fine now

yacovm:
Correct me if I'm wrong, but - in TLS 1.3 the certificates are encrypted (unlike in 1.2).
Doesn't this make troubleshooting harder?

Can anyone explain the benefits of using TLS 1.3 in Fabric where we have long lasting TLS connections between nodes? The TLS handshake being faster now doesn't help us much...

mastersingh24 (approved):

mastersingh24:
> Correct me if I'm wrong, but - in TLS 1.3 the certificates are encrypted (unlike in 1.2).
> Doesn't this make troubleshooting harder?
> 
> Can anyone explain the benefits of using TLS 1.3 in Fabric where we have long lasting TLS connections between nodes? The TLS handshake being faster now doesn't help us much...

I don't think the troubleshooting you can do from within the Fabric context is any worse here.

There are really two reasons to allow TLS 1.3:
- it's supported by Go and gRPC so no reason not to allow it
- some users might have requirements to to TLS 1.3 in the near future

yacovm:
> I don't think the troubleshooting you can do from within the Fabric context is any worse here.

Well in TLS 1.2 I see the certificates in the clear so I can analyze traffic and catch validation chains that do not appear in the channel config, or mismatching SANs. With TLS 1.3 I don't have that because it's encrypted.

yacovm (approved):

PR #1147, opened 2020-04-22
push down the peer started log

hello2mao:
#### Type of change

#### Description

Push down the log statement until the peer has actually been started.

#### Additional details
NA

#### Related issues

NA

### DIFF
--- a/internal/peer/node/start.go
+++ b/internal/peer/node/start.go
@@ -832,8 +832,6 @@ func serve(args []string) error {
 		syscall.SIGTERM: func() { serve <- nil },
 	}))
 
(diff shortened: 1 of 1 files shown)


lindluni:
If anything this should be pushed down the function, until the peer has actually been started instead of removing it.

This log statement is part of the integration suites startup verification sequence which is why your tests are failing: https://github.com/hyperledger/fabric/blob/master/integration/nwo/network.go#L1220

sykesm:
@hello2mao what is motivating this change? 

Originally it was being removed and now it’s being moved.

The readiness message does have a purpose and is desired so I’m confused about the goal.

hello2mao:
> @hello2mao what is motivating this change?
> 
> Originally it was being removed and now it’s being moved.
> 
> The readiness message does have a purpose and is desired so I’m confused about the goal.

@sykesm I just want to put the log to the right place where peer has actually been started. if it is unnecessary, I will close the pr. Thanks.

PR #1155, opened 2020-04-23
suspend external builder bin release binary file Stat on windows

KevinXiong2018:
#### Type of change

#### Description

Under the windows platform, because os.Stat (release binary file) is forced on directory  bin/release when running the external builder release script, it is impossible to run the bin/release.exe or bin/release.com or bin/release.bat executable file . Unless an empty directory release is created under the bin directory to skip this detection.


lindluni:
Though I suspect there will be stronger opinions on this and whether it should be "fixed", this is definitely, in my opinion, not the right way to handle this, removing the check entirely to satisfy the side-effect of a single platform isn't the right way to go about this, as it functions as intended on all other platforms, and we don't technically support Windows for running fabric.

Adding an `if` to check for the Windows OS seems more reasonable, but we will see what the opinion on this is in general since it is on Windows

KevinXiong2018:
Yes, this is a very good idea, only suspend external builder bin release binary file Stat on windows, the code has been modified, thanks Brett.

sykesm (commented):
The peer node is not supported or tested on windows. I do not want to litter the code with windows specifics unless there is a focused line of work to support that platform, add tests, and explicitly document it.

KevinXiong2018:
Windows is a very commonly used client operating system, used by ordinary customers and computer engineers. Since fabric 2.0 released the new function of external builder, fabric has been very easy to run on the windows operating system, which is convenient for us to develop fabric cc on the windows platform and show blockchain applications to our commercial customers. If only simple code changes are needed here, and it obviously does not affect the fabric in other existing platforms, it can be friendly to support windows, which will be a very happy thing.

sykesm:
> Windows is a very commonly used client operating system, used by ordinary customers and computer engineers.

I understand but I will emphasize the word *client*. This implies support for invoking transactions and packaging chaincode. This is what we currently support on Windows.

> Since fabric 2.0 released the new function of external builder, fabric has been very easy to run on the windows operating system

This does not mean it's currently supported or tested environment to use the external builder.

> it can be friendly to support windows, which will be a very happy thing.

And focusing on the word *support*: This implies testing and documentation. Neither of those is currently present because Windows is not a supported platform for running the peer node. (We do support using the client commands related to chaincode and invoking transactions.)

If the community desires support for Windows, it should be done as part of a focused effort that delivers a test environment in CI. As it stands right now, we have only made changes to accommodate Windows in basic client side scenarios and there are exactly two places where we have Windows specific code:

- Attempt to test no permission in unit tests is skipped on windows.
- Signal handling reference to SIGUSR1 is conditionally omitted on Windows by build tag

Instead of making platform specific checks, please modify the code to use a platform agnostic interface that would work equally well in our tested environment and in your environment. I'd suggest evaluating adoption of `exec.LookPath` instead of `os.Stat`.

sykesm (changes requested):
Instead of making platform specific checks, please modify the code to use a platform agnostic interface that would work equally well in our tested environment and in your environment. I'd suggest evaluating adoption of exec.LookPath instead of os.Stat.

Also, the PR should consist of a single commit. This currently has 5 commits (4 of which are merge commits).

KevinXiong2018:
In core/container/externalbuilder/externalbuilder.go, we have three other functions: Detect, Build, and Run. They do not perform file  bin/detect,  bin/build, and  bin/run existence detection, and they can safely run on various platforms. I haven't found the need for the Release function to check the file  bin/release existence. If these code segments are removed, it can also be safely run on various platforms including windows.

PR #1156, opened 2020-04-23
[FAB-17790] Checking returned error is redundant when getting peer host by splitting coreConfig peerAddress

wuyingjun-lucky:
#### Description

Checking returned error is redundant when getting peer host by splitting coreConfig peerAddress.

It will return when  encountering errors in the process of getting coreConfig if the hostAddress is not formatted.

#### Related issues
https://jira.hyperledger.org/browse/FAB-17790


sykesm (changes requested):
Thank you for the contribution but we should always check errors.

wuyingjun-lucky:
@sykesm ok，thank you

PR #1157, opened 2020-04-23
[FAB-17728] Add 100ms delay to pkcs11 create session loop after failing to retrieve a session from the HSM session cache

caod123:
#### Type of change

#### Description

Previously there was no backoff when attempting to create a new session if one was not available in the HSM session cache. This PR introduces a hardcoded backoff after each attempt up to 10.

#### Related issues

[FAB-17728](https://jira.hyperledger.org/browse/FAB-17728)

### DIFF
--- a/bccsp/pkcs11/pkcs11.go
+++ b/bccsp/pkcs11/pkcs11.go
@@ -15,6 +15,7 @@ import (
 	"fmt"
 	"math/big"
 	"sync"
+	"time"
 
 	"github.com/miekg/pkcs11"
 	"go.uber.org/zap/zapcore"
@@ -93,6 +94,7 @@ func (csp *impl) getSession() (session pkcs11.SessionHandle) {
 func createSession(ctx *pkcs11.Ctx, slot uint, pin string) pkcs11.SessionHandle {
 	var s pkcs11.SessionHandle
 	var err error
+	// attempt 10 times to open a session with a 100ms delay after each attempt
 	for i := 0; i < 10; i++ {
 		s, err = ctx.OpenSession(slot, pkcs11.CKF_SERIAL_SESSION|pkcs11.CKF_RW_SESSION)
 		if err != nil {
@@ -100,6 +102,7 @@ func createSession(ctx *pkcs11.Ctx, slot uint, pin string) pkcs11.SessionHandle 
 		} else {
 			break
 		}
+		time.Sleep(100 * time.Millisecond)
 	}
 	if err != nil {
 		logger.Fatalf("OpenSession failed [%s]", err)


caod123:
@btl5037 do you think it's worth making this configurable? The number of retries is already hardcoded to 10 so we didn't really bother with any max limits since a 10th attempt maxes out with a sleep of 515ms (for a total of about 2.6s after 10 attempts) which seems reasonable. On that note do you think that amount of backoff should be enough to alleviate the blasting

bccsp/pkcs11/pkcs11.go
```diff
@@ -93,13 +95,16 @@ func (csp *impl) getSession() (session pkcs11.SessionHandle) {
...
+	// attempt 10 times to open a session with a minor exponential backoff after each attempt
+	// backoff is 1.2^attempt * 100 in milliseconds
 	for i := 0; i < 10; i++ {
 		s, err = ctx.OpenSession(slot, pkcs11.CKF_SERIAL_SESSION|pkcs11.CKF_RW_SESSION)
 		if err != nil {
 			logger.Warningf("OpenSession failed, retrying [%s]\n", err)
 		} else {
 			break
 		}
+		time.Sleep(time.Duration(math.Pow(1.2, float64(i))*100) * time.Millisecond)
```
sykesm: Do we really need to bring in a math library to do
```
delay := 100 * time.Millisecond
for ... {
    ...
    time.Sleep(delay)
    delay = delay * 6 / 5
}
```
And adding more configuration leads to more doc and more pain. Just waiting 100ms between iterations would be a huge step forward here.

sykesm (changes requested):
Can this be simplified? Either a simple wait (no backoff) or simpler logic to calculate the backoff?

bccsp/pkcs11/pkcs11.go
caod123: Makes sense, exponential probably doesn't make much sense given how small of a scale it is anyways.

bccsp/pkcs11/pkcs11.go
```diff
@@ -93,13 +94,15 @@ func (csp *impl) getSession() (session pkcs11.SessionHandle) {
 func createSession(ctx *pkcs11.Ctx, slot uint, pin string) pkcs11.SessionHandle {
 	var s pkcs11.SessionHandle
 	var err error
+	// attempt 10 times to open a session with a 100ms backoff after each attempt
```
sykesm: nit: it's a delay instead of a backoff.

sykesm (dismissed):
Seems the browser remember the *last* comment I made...  LGTM.

I'm pretty certain this issue was opened based on comments I made to a PR last week....

bccsp/pkcs11/pkcs11.go
caod123: I'll make the change since CI is still running anyways

sykesm (approved):

lindluni:
> @btl5037 do you think it's worth making this configurable? The number of retries is already hardcoded to 10 so we didn't really bother with any max limits since a 10th attempt maxes out with a sleep of 515ms (for a total of about 2.6s after 10 attempts) which seems reasonable. On that note do you think that amount of backoff should be enough to alleviate the blasting

I think fundamentally there are going to be bigger problems we will need to alleviate. While this addresses the immediate concern of the uncontrolled pummeling of the HSM, running at ~8tps we saw hundreds of sessions opened with the HSM, at even moderate TPS you are going to start hitting up against the HSM's resource limitations (and who knows what happens when that starts to occur).

We really need to get an integration/system test for HSM done so we can actually test functionality. It's on my backlog to do a HSM functional test, but actually getting around to doing it is another story

caod123:
> We really need to get an integration/system test for HSM done so we can actually test functionality. It's on my backlog to do a HSM functional test, but actually getting around to doing it is another story

Likewise I have a TODO to revisit the pkcs11 integration test at some point hopefully in the near future.

lindluni:

lindluni:

lindluni:

lindluni:

lindluni:

PR #1159, opened 2020-04-23
[FAB-17778] MSP.IsWellFormed: Only allow canonical CA signatures

yacovm:
This change makes MSP.IsWellFormed (not used in transaction validation, only for gossip)
for x509 based identities to check that the signature of the CA over an identity has canonical ASN1 encoding.
It also makes sure that the certificate is encoded in a single and minimal PEM block.

Change-Id: I7ba0db00b3eb58171bb05ee7680721a41ef1c802

### DIFF
--- a/msp/mspimpl.go
+++ b/msp/mspimpl.go
@@ -19,6 +19,7 @@ import (
 	"github.com/hyperledger/fabric/bccsp/factory"
 	"github.com/hyperledger/fabric/bccsp/signer"
 	"github.com/hyperledger/fabric/bccsp/sw"
+	"github.com/hyperledger/fabric/bccsp/utils"
 	"github.com/pkg/errors"
 )
 
@@ -825,10 +826,14 @@ func (msp *bccspmsp) sanitizeCert(cert *x509.Certificate) (*x509.Certificate, er
 // In this MSP implementation, well formed means that the PEM has a Type which is either
 // the string 'CERTIFICATE' or the Type is missing altogether.
 func (msp *bccspmsp) IsWellFormed(identity *m.SerializedIdentity) error {
-	bl, _ := pem.Decode(identity.IdBytes)
+	bl, rest := pem.Decode(identity.IdBytes)
 	if bl == nil {
 		return errors.New("PEM decoding resulted in an empty block")
 	}
+	if len(rest) > 0 {
+		return errors.Errorf("identity %s for MSP %s has trailing bytes", string(identity.IdBytes), identity.Mspid)
+	}
+
 	// Important: This method looks very similar to getCertFromPem(idBytes []byte) (*x509.Certificate, error)
 	// But we:
 	// 1) Must ensure PEM block is of type CERTIFICATE or is empty
@@ -837,6 +842,30 @@ func (msp *bccspmsp) IsWellFormed(identity *m.SerializedIdentity) error {
 	if bl.Type != "CERTIFICATE" && bl.Type != "" {
 		return errors.Errorf("pem type is %s, should be 'CERTIFICATE' or missing", bl.Type)
 	}
-	_, err := x509.ParseCertificate(bl.Bytes)
-	return err
+	cert, err := x509.ParseCertificate(bl.Bytes)
+	if err != nil {
+		return err
+	}
+
+	return isIdentitySignedInCanonicalForm(cert.Signature, identity.Mspid, identity.IdBytes)
+
+}
+
+func isIdentitySignedInCanonicalForm(sig []byte, mspID string, pemEncodedIdentity []byte) error {
+	r, s, err := utils.UnmarshalECDSASignature(sig)
+	if err != nil {
+		return err
+	}
+
+	expectedSig, err := utils.MarshalECDSASignature(r, s)
+	if err != nil {
+		return err
+	}
+
+	if !bytes.Equal(expectedSig, sig) {
+		return errors.Errorf("identity %s for MSP %s has a non canonical signature",
+			string(pemEncodedIdentity), mspID)
+	}
+
+	return nil
 }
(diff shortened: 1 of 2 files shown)


C0rWin (dismissed):

msp/mspimpl.go
```diff
@@ -825,10 +827,14 @@ func (msp *bccspmsp) sanitizeCert(cert *x509.Certificate) (*x509.Certificate, er
 // In this MSP implementation, well formed means that the PEM has a Type which is either
 // the string 'CERTIFICATE' or the Type is missing altogether.
 func (msp *bccspmsp) IsWellFormed(identity *m.SerializedIdentity) error {
-	bl, _ := pem.Decode(identity.IdBytes)
+	bl, rest := pem.Decode(identity.IdBytes)
```
ale-linux, line 829: Thanks for the investigation, Yacov. pem.Decode is also used in deserializeIdentityInternal, isn't there a risk that there could be principal multiplication by adding trailing bytes there too?

msp/mspimpl.go
```diff
@@ -837,6 +843,32 @@ func (msp *bccspmsp) IsWellFormed(identity *m.SerializedIdentity) error {
...
+	expectedSig, err := utils.MarshalECDSASignature(r, s)
+	if err != nil {
+		return err
+	}
+
+	if !bytes.Equal(expectedSig, sig) {
+		encoding := base64.StdEncoding
+		return errors.Errorf("identity %s for MSP %s has invalid CA signature: expected %s but got %s",
+			string(pemEncodedIdentity), mspID, encoding.EncodeToString(expectedSig),
+			encoding.EncodeToString(sig))
```
ale-linux: I'm not sure that logging those bytes is particularly helpful...

ale-linux (changes requested):

msp/mspimpl.go
yacovm: That's a good question, but I think there is no risk, because the MSP discards the PEM (and the DER) completely and just propagates the pure x509 certificate into the identity instance construction. 

Then, the identity identifier is computed over the raw DER of the certificate which is enclosed in the first PEM block so we should be safe.

The reason I am doing this for gossip, is while gossip uses the MSP to validate identities, it carries around the representations received from the network and the MSP never takes part in sanitizing them.
This also means that gossip can theoretically save in memory different representations of the same identity - the low and the high S.
However I am OK with that, because gossip supports having several identities mapped to the same endpoint(s), and there is no harm in having 2 different representations.

Additionally, we can't really sanitize these certs because we don't know the order of the curve's group (so we can't compute if we're below or above half the order... a high S in P256 is a low S in P384 isn't it?).
To my understanding, you cannot _easily_ deduce the group parameters of the signature from the certificate (and that's a [good thing](https://media.defense.gov/2020/Jan/14/2002234275/-1/-1/0/CSA-WINDOWS-10-CRYPT-LIB-20190114.PDF)!) but only from its parent, which we don't have access to in the gossip layer (only in the MSP)

The reason for this change is to prevent an infinite amount of representations which can flood the memory.

msp/mspimpl.go
yacovm: How so? under normal circumstances we should never ever see this error. 
If we do happen to see it, one can use it to investigate what's going on.

msp/mspimpl.go
yacovm: @adecaro I'd like to know what is your opinion.

msp/mspimpl.go
ale-linux: Yeah but those bytes don't help the investigation - the error message is sufficient, isn't it? What is anyone going to do with those bytes anyway?

msp/mspimpl.go
ale-linux: So, I tried this

https://pastebin.com/tqD0Jvfj

and it works, the identity is valid despite the trailing bytes. This isn't a huge problem, because the identifier of the certificate, which is used for cert dedup, is the hash of the part that the x.509 library digests, so the trailing bytes are just dropped. Still, it's possible that this is an issue elsewhere.. Your thoughts?

msp/mspimpl.go
yacovm: You would know who is the offending identity!

msp/mspimpl.go
yacovm: This is what I said above, we ignore the trailer when it is time to make access control decisions. Who cares if the identity has trailing bytes if you shave them off of it and forget about them?

`Still, it's possible that this is an issue elsewhere.. Your thoughts?`

Well what about the certificate header? We ignore it as well don't we? And I remember that in the past we never used "CERTIFICATE" to serialize the header, but used a blank string. 
Look at the comment in [IsWellFormed](https://github.com/hyperledger/fabric/blob/master/msp/mspimpl.go#L832-L836), it talks exactly about this. 
If you want to make a change without a capability you need to make sure it doesn't run anyone over.

msp/mspimpl.go
ale-linux: identity %s for MSP %s is what tells you who the offender is. The rest is useless, isn't it?

msp/mspimpl.go
ale-linux: RIght, I guess a deeper analysis of all places where pem decoding is invoked might be warranted. I've seen it uses in the raft code as well for instance. Perhaps you could create a jira so we don't forget? Thanks

msp/mspimpl.go
yacovm: OK I removed the prints you don't like

msp/mspimpl.go
yacovm: OK I opened FAB-17811.

ale-linux (approved):

PR #1160, opened 2020-04-23
[FAB-17778] MSP.IsWellFormed: Only allow canonical CA signatures

yacovm:
This change makes MSP.IsWellFormed (not used in transaction validation, only for gossip)
for x509 based identities to check that the signature of the CA over an identity has canonical ASN1 encoding.
It also makes sure that the certificate is encoded in a single and minimal PEM block.

Change-Id: I7ba0db00b3eb58171bb05ee7680721a41ef1c802

### DIFF
--- a/msp/mspimpl.go
+++ b/msp/mspimpl.go
@@ -18,6 +18,7 @@ import (
 	"github.com/hyperledger/fabric/bccsp"
 	"github.com/hyperledger/fabric/bccsp/factory"
 	"github.com/hyperledger/fabric/bccsp/signer"
+	"github.com/hyperledger/fabric/bccsp/utils"
 	m "github.com/hyperledger/fabric/protos/msp"
 	"github.com/pkg/errors"
 )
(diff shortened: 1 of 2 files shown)


msp/mspimpl.go
```diff
@@ -815,6 +821,27 @@ func (msp *bccspmsp) IsWellFormed(identity *m.SerializedIdentity) error {
...
-	return err
+	cert, err := x509.ParseCertificate(bl.Bytes)
+	if err != nil {
+		return err
+	}
+
+	r, s, err := utils.UnmarshalECDSASignature(cert.Signature)
+	if err != nil {
+		return nil
+	}
```
C0rWin: return err?

msp/mspimpl.go
```diff
@@ -815,6 +821,27 @@ func (msp *bccspmsp) IsWellFormed(identity *m.SerializedIdentity) error {
...
+
+	expectedSig, err := utils.MarshalECDSASignature(r, s)
+	if err != nil {
+		return nil
+	}
```
C0rWin: return err?

msp/mspimpl.go
```diff
@@ -815,6 +821,27 @@ func (msp *bccspmsp) IsWellFormed(identity *m.SerializedIdentity) error {
...
+
+	if !bytes.Equal(expectedSig, cert.Signature) {
+		encoding := base64.StdEncoding
+		return errors.Errorf("identity %s for MSP %s has invalid CA signature: expected %s but got %s",
+			string(identity.IdBytes), identity.Mspid, encoding.EncodeToString(expectedSig),
+			encoding.EncodeToString(cert.Signature))
+	}
```
C0rWin: probably externalize validation code into separate function? 

```
     return validateCertSig(cert.Signature)
}

func validateCertSig(certSig []byte) error {
	r, s, err := utils.UnmarshalECDSASignature(cert.Signature)
	if err != nil {
		return err
	}


	expectedSig, err := utils.MarshalECDSASignature(r, s)
	if err != nil {
		return err
	}

	if !bytes.Equal(expectedSig, cert.Signature) {
		encoding := base64.StdEncoding
		return errors.Errorf("identity %s for MSP %s has invalid CA signature: expected %s but got %s",
			string(identity.IdBytes), identity.Mspid, encoding.EncodeToString(expectedSig),
			encoding.EncodeToString(cert.Signature))
	}

       return nil
}
```

ale-linux (approved):

PR #1162, opened 2020-04-24
remove external builder bin release binary file Stat

KevinXiong2018:
Signed-off-by: kevinxiong <59443487@qq.com>

#### Type of change

#### Description

In core/container/externalbuilder/externalbuilder.go, we have three other functions: Detect, Build, and Run. They do not perform file bin/detect, bin/build, and bin/run existence detection, and they can safely run on various platforms. I haven't found the need for the Release function to check the file bin/release existence. If these code segments are removed, it can also be safely run on various platforms including windows.

reference: suspend external builder bin release binary file Stat on windows
https://github.com/hyperledger/fabric/pull/1155


sykesm (changes requested):
Release is optional and documented as such. I’ve already provided a recommendation on how to change the implementation to preserve the existing semantics.

sykesm:
See #1163.

PR #1163, opened 2020-04-24
Switch from os.Stat to exec.LookPath

sykesm:
Replace call to `os.Stat` to check for the presence of the `bin/release` script with `exec.LookPath`. The `LookPath` function, when provided with a relative path, will look for the presence of the executable and determine if it's executable. On non-unix platforms, it will also handle looking for executable suffixes as appropriate.

This supersedes #1155 and #1162 as it implements the suggested change.


lindluni (approved):

lindluni (approved):

KevinXiong2018:
This afternoon I once tried the above code. Under the windows platform, it can work normally, but on Linux and Mac platforms, they will go to $ PATH to find the executable file of the release, and it seems to not work.

PR #1174, opened 2020-04-24
[FAB-17748] Additional ConfigTx doc cleanup

stephyee:
#### Type of change

- Documentation update

#### Description

   - Adjust line length on godoc examples and constants to fix horizontal
    scroll
  - Reword constants for ConfigGroup keys

#### Related issues
[FAB-17748](https://jira.hyperledger.org/browse/FAB-17748)

### DIFF
--- a/pkg/configtx/constants.go
+++ b/pkg/configtx/constants.go
@@ -6,37 +6,35 @@ SPDX-License-Identifier: Apache-2.0
 
 package configtx
 
-import "math"
-
 const (
 	// These values are fixed for the genesis block.
 	msgVersion = 0
 	epoch      = 0
 
-	defaultHashingAlgorithm = "SHA256"
-
-	defaultBlockDataHashingStructureWidth = math.MaxUint32
-
-	// ConsortiumKey is the key for the common.ConfigValue for the Consortium message.
+	// ConsortiumKey is the key for the ConfigValue of a
+	// Consortium.
 	ConsortiumKey = "Consortium"
 
-	// HashingAlgorithmKey is the common.ConfigValue type key name for the HashingAlgorithm message.
+	// HashingAlgorithmKey is the key for the ConfigValue of a
+	// HashingAlgorithm.
 	HashingAlgorithmKey = "HashingAlgorithm"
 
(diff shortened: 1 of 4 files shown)


pkg/configtx/constants.go
```diff
@@ -6,37 +6,35 @@ SPDX-License-Identifier: Apache-2.0
...
+	// CapabiltiesKey can be used at the channel, application, and orderer levels.
 	CapabilitiesKey = "Capabilities"
 
-	// EndpointsKey is the common.ConfigValue key name for the Endpoints message in the OrdererOrgGroup.
+	// EndpointsKey is the key for the ConfigValue, Endpoints in
+	// a OrdererOrgGroup.
 	EndpointsKey = "Endpoints"
 
-	// MSPKey is the key for the MSP definition in orderer groups.
+	// MSPKey is the key for the ConfigValue MSP in orderer ConfigGroups.
```
caod123: Can you check if this is only used for Orderer? I feel like this is used for any org ConfigGroup

pkg/configtx/msp.go
```diff
@@ -71,8 +71,8 @@ type MSP struct {
 	// List of TLS intermediate certificates trusted by this MSP;
 	// They are returned by GetTLSIntermediateCerts.
 	TLSIntermediateCerts []*x509.Certificate
-	// fabric_node_ous contains the configuration to distinguish clients from peers from orderers
-	// based on the OUs.
+	// fabric_node_ous contains the configuration to distinguish clients
```
caod123: it's not fabric_node_ous, can you check the other comments to make sure they match with the field theyre describing?

caod123 (changes requested):

pkg/configtx/constants.go
stephyee: I know it can also be used for application at least.

caod123: MSP configuration is for an org in general. Off the top of my head I think this key is used for the MSP config value in which case it would apply to all orgs

pkg/configtx/msp.go
stephyee: 🧹

caod123 (approved):

PR #1177, opened 2020-04-26
Error strings should not be capitalized (unless beginning with proper nouns or acronyms) or end with punctuation

wuyingjun-lucky:
#### Type of change

- Test update

#### Description

Error strings should not be capitalized (unless beginning with proper nouns or acronyms) or end with punctuation


lindluni (approved):
Thank you for the correction!

PR #1181, opened 2020-04-27
[LEDGER] ensure peer panic does not occur when the index path is wrong

cendhu:
#### Type of change

- Bug fix

#### Description

While processing the CouchDB index files, we retrieve and store all directories in the index file path into an array in order to find whether it is a `chaincode` index or `collection` index. If it is a collection index, the collection name would be found using the same array. 

Currently, we access this array without checking the number of directories in the path (i.e., the length of the array) assuming that the user would pass the correct path. When the user passes a wrong path, the array access could result in a peer panic. Hence, in this PR, we check the length of the array before accessing it. UT is added too. 

#### Related issues

[FAB-17771](https://jira.hyperledger.org/browse/FAB-17771)


manish-sethi (approved):
Thanks @cendhu for keeping track and fixing this.

PR #1184, opened 2020-04-27
[LEDGER] cp couchdb config struct while testing couchdb failure scenarios

cendhu:
#### Type of change

#### Description

We make changes to `MaxRetries` and `Address` fields in `config` struct to test some failure scenarios. To ensure proper cleanup of the test environment, we revert those changes at the end of each UT. It looks hacky and hence, in this PR, we clone the config, manipulate the values in that clone and use it for testing failure scenarios.

#### Related issues

[FAB-17761](https://jira.hyperledger.org/browse/FAB-17761)


manish-sethi (approved):
Thanks @cendhu for fixing this.

PR #1185, opened 2020-04-27
[FAB-17813] Implement function to add consortium

stephyee:
#### Type of change

- New feature

#### Description

Support adding a new consortium to a consortiums group.

#### Related issues

[FAB-17813](https://jira.hyperledger.org/browse/FAB-17813)

### DIFF
--- a/pkg/configtx/consortiums.go
+++ b/pkg/configtx/consortiums.go
@@ -21,6 +21,22 @@ type Consortium struct {
 	Organizations []Organization
 }
 
+// AddConsortium adds a new consortium to a channel configuration.
+func (c *ConfigTx) AddConsortium(consortium Consortium) error {
+	consortiumsGroup := c.updated.ChannelGroup.Groups[ConsortiumsGroupKey]
+
+	consortiumsGroup.Groups[consortium.Name] = newConfigGroup()
+
+	for _, org := range consortium.Organizations {
+		err := c.SetConsortiumOrg(org, consortium.Name)
+		if err != nil {
+			return err
+		}
+	}
+
+	return nil
+}
+
 // RemoveConsortium removes a consortium from a channel configuration.
 // Removal will panic if the consortiums group does not exist.
 func (c *ConfigTx) RemoveConsortium(consortiumName string) {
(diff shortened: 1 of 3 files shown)


pkg/configtx/consortiums.go
```diff
@@ -21,6 +21,22 @@ type Consortium struct {
 	Organizations []Organization
 }
 
+// AddConsortium adds a new consortium to a channel configuration.
+func (c *ConfigTx) AddConsortium(consortium Consortium) error {
+	consortiumsGroup := c.updated.ChannelGroup.Groups[ConsortiumsGroupKey]
+
+	consortiumsGroup.Groups[consortium.Name] = newConfigGroup()
+
+	for _, org := range consortium.Organizations {
```
caod123, line 30: This logic implies that when adding a consortium the consumer will also include the orgs for the consortium. I'm fine with that and even if no orgs are specified they can be added later.

pkg/configtx/consortiums_test.go
```diff
@@ -900,6 +900,60 @@ func TestGetConsortiumOrg(t *testing.T) {
...
+	}
+
+	c := New(config)
+
+	newConsortium := consortiums[0]
+	newConsortium.Name = "Consortium2"
+
+	err = c.AddConsortium(newConsortium)
+	gt.Expect(err).NotTo(HaveOccurred())
+	// err = protolator.DeepMarshalJSON(os.Stdout, c.UpdatedConfig())
```
caod123, line 927: Any reason why we aren't testing the value was added?

pkg/configtx/example_test.go
```diff
@@ -564,6 +564,29 @@ func ExampleConfigTx_SetConsortiumMSP() {
...
+
+	org, err := c.ApplicationOrg("Org1")
+	if err != nil {
+		panic(err)
+	}
+
+	newConsortium := configtx.Consortium{
+		Name: "SampleConsortium2",
+		Organizations: []configtx.Organization{
+			org,
```
caod123, line 580: It may be improper as an example here (although technically fine) to add an application org as the consortium org for the example. Let's create a new org for the sole purpose of adding it to the consortium just for clarity.

caod123 (changes requested):

pkg/configtx/consortiums_test.go
stephyee: I missed that, l will add an assertion.

PR #1191, opened 2020-04-29
Update discovery-overview.rst

RobertGumair:
#### Type of change

- Bug fix

#### Description

"Service Discovery" was referenced as "Discovery Service" incorrectly.

#### Release Note


yacovm (changes requested):
This is not a mistake.

RobertGumair:
so service discovery is a discovery service.  Interesting.

PR #1193, opened 2020-04-29
[FAB-17819] Discovery returns user friendly errors

yacovm:
Currently, the discovery service filters out peers that don't have the chaincode installed
early on in the computation, and as a result - the service cannot distinguish from a case
where there are not enough alive peers to satisfy the endorsement policy, or that there are
enough peers but the chaincode is not installed on enough of them.

This change set defers the chaincode filtering to the end of the computation, so
the layouts and peer group mapping is creating without taking into account if the peers
have the chaincode installed on them, and if there is no layout that can be satisfied
without taking into account the chaincodes - the error that is returned now
is "no peer combination can satisfy the endorsement policy",
instead of "cannot satisfy any principal combination".

Afterwards, the layouts are being inspected once again, and then the layouts
that cannot be satisfied are filtered out, when the error returned
when no layout can be satisfied is now: "required chaincodes are not installed on sufficient peers".

Change-Id: I74eb29b30aec1a87842d220414c73872cdbc8304

### DIFF
--- a/discovery/endorsement/endorsement.go
+++ b/discovery/endorsement/endorsement.go
@@ -56,6 +56,11 @@ type gossipSupport interface {
 	Peers() Members
 }
 
+type membersChaincodeMapping struct {
+	members          Members
+	chaincodeMapping map[string]NetworkMember
+}
+
 type endorsementAnalyzer struct {
 	gossipSupport
 	principalEvaluator
@@ -77,13 +82,13 @@ type peerPrincipalEvaluator func(member NetworkMember, principal *msp.MSPPrincip
 
 // PeersForEndorsement returns an EndorsementDescriptor for a given set of peers, channel, and chaincode
 func (ea *endorsementAnalyzer) PeersForEndorsement(channelID common.ChannelID, interest *discovery.ChaincodeInterest) (*discovery.EndorsementDescriptor, error) {
-	chanMembership, err := ea.PeersAuthorizedByCriteria(channelID, interest)
+	membersAndCC, err := ea.peersByCriteria(channelID, interest, false)
 	if err != nil {
 		return nil, errors.WithStack(err)
 	}
-	channelMembersById := chanMembership.ByID()
+	channelMembersById := membersAndCC.members.ByID()
 	// Choose only the alive messages of those that have joined the channel
-	aliveMembership := ea.Peers().Intersect(chanMembership)
+	aliveMembership := ea.Peers().Intersect(membersAndCC.members)
 	membersById := aliveMembership.ByID()
 	// Compute a mapping between the PKI-IDs of members to their identities
 	identitiesOfMembers := computeIdentitiesOfMembers(ea.IdentityInfo(), membersById)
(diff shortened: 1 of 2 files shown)


discovery/endorsement/endorsement.go
```diff
@@ -150,23 +171,51 @@ func (ea *endorsementAnalyzer) computeEndorsementResponse(ctx *context) (*discov
...
 	}, nil
 }
 
+func filterOutUnsatisfiedLayouts(endorsersByGroup map[string]*discovery.Peers, layouts []*discovery.Layout) []*discovery.Layout {
+	// Iterate once again over all layouts and ensure every layout has enough peers in the EndorsersByGroups
+	// as required by the quantity in the layout.
+	var filteredLayouts []*discovery.Layout
+	for _, layout := range layouts {
+		var layoutInvalid bool
+		for group, quantity := range layout.QuantitiesByGroup {
```
ale-linux, line 205: Could layouts contain a nil layout, which would cause a nil ptr deref here?

discovery/endorsement/endorsement.go
```diff
@@ -150,23 +171,51 @@ func (ea *endorsementAnalyzer) computeEndorsementResponse(ctx *context) (*discov
...
 		Layouts:           layouts,
-		EndorsersByGroups: endorsersByGroup(criteria),
+		EndorsersByGroups: endorsersByGroup,
 	}, nil
 }
 
+func filterOutUnsatisfiedLayouts(endorsersByGroup map[string]*discovery.Peers, layouts []*discovery.Layout) []*discovery.Layout {
+	// Iterate once again over all layouts and ensure every layout has enough peers in the EndorsersByGroups
+	// as required by the quantity in the layout.
+	var filteredLayouts []*discovery.Layout
```
ale-linux: You could optimise here and pre-create the slice with 0 size and capacity equal to layouts

discovery/endorsement/endorsement.go
```diff
@@ -150,23 +171,51 @@ func (ea *endorsementAnalyzer) computeEndorsementResponse(ctx *context) (*discov
...
-		chanMemberById:  ctx.channelMembersById,
-		idOfMembers:     ctx.identitiesOfMembers,
+		possibleLayouts:  layouts,
+		satGraph:         satGraph,
+		chanMemberById:   ctx.channelMembersById,
+		idOfMembers:      ctx.identitiesOfMembers,
+		chaincodeMapping: ctx.chaincodeMapping,
+	}
+
+	endorsersByGroup := endorsersByGroup(criteria)
```
ale-linux: I personally found this confusing - variable with the same name as a function. Up to you whether you may want to rename this

ale-linux (changes requested):

discovery/endorsement/endorsement.go
yacovm: no.

discovery/endorsement/endorsement.go
yacovm: done

discovery/endorsement/endorsement.go
yacovm: done

discovery/endorsement/endorsement.go
```diff
@@ -150,23 +171,51 @@ func (ea *endorsementAnalyzer) computeEndorsementResponse(ctx *context) (*discov
...
 		Layouts:           layouts,
-		EndorsersByGroups: endorsersByGroup(criteria),
+		EndorsersByGroups: groupToEndorserListMapping,
 	}, nil
 }
 
+func filterOutUnsatisfiedLayouts(endorsersByGroup map[string]*discovery.Peers, layouts []*discovery.Layout) []*discovery.Layout {
+	// Iterate once again over all layouts and ensure every layout has enough peers in the EndorsersByGroups
+	// as required by the quantity in the layout.
+	filteredLayouts := make([]*discovery.Layout, len(layouts))
```
ale-linux: don't you have to create it as
`filteredLayouts := make([]*discovery.Layout, 0, len(layouts))`
? Otherwise you're just appending to a bunch of nils on line 214

ale-linux (changes requested):

discovery/endorsement/endorsement.go
yacovm: Do you see now why i prefer my slices, cut thin?

yacovm: done, thanks for noticing.

ale-linux (approved):

PR #1197, opened 2020-04-29
[FAB-17774] support orderer restart without genesis block

DereckLuo:

#### Type of change

- Bug fix

#### Description
* adding support for orderer to restart without providing genesis block.

#### Related issues
[FAB-17774](https://jira.hyperledger.org/browse/FAB-17774)

### DIFF
--- a/orderer/common/server/etcdraft_test.go
+++ b/orderer/common/server/etcdraft_test.go
@@ -30,10 +30,6 @@ func nextPort() int32 {
 func TestSpawnEtcdRaft(t *testing.T) {
 	gt := NewGomegaWithT(t)
 
-	// Set the fabric root folder for easy navigation to sampleconfig folder
-	fabricRootDir, err := filepath.Abs(filepath.Join("..", "..", ".."))
-	gt.Expect(err).NotTo(HaveOccurred())
-
 	// Build the configtxgen binary
 	configtxgen, err := gexec.Build("github.com/hyperledger/fabric/cmd/configtxgen")
 	gt.Expect(err).NotTo(HaveOccurred())
@@ -56,11 +52,11 @@ func TestSpawnEtcdRaft(t *testing.T) {
 
 	t.Run("Bad", func(t *testing.T) {
 		t.Run("Invalid bootstrap block", func(t *testing.T) {
-			testEtcdRaftOSNFailureInvalidBootstrapBlock(NewGomegaWithT(t), tempDir, orderer, fabricRootDir, configtxgen, cryptoPath)
+			testEtcdRaftOSNFailureInvalidBootstrapBlock(NewGomegaWithT(t), tempDir, orderer, configtxgen, cryptoPath)
 		})
 
 		t.Run("TLS disabled single listener", func(t *testing.T) {
-			testEtcdRaftOSNNoTLSSingleListener(NewGomegaWithT(t), tempDir, orderer, fabricRootDir, configtxgen, cryptoPath)
+			testEtcdRaftOSNNoTLSSingleListener(NewGomegaWithT(t), tempDir, orderer, configtxgen, cryptoPath)
 		})
 	})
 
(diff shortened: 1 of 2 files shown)


orderer/common/server/main.go
```diff
@@ -124,8 +124,13 @@ func Main() {
 	var clusterDialer *cluster.PredicateDialer
 	var clusterType, reuseGrpcListener bool
 	var serversToUpdate []*comm.GRPCServer
-	if conf.General.BootstrapMethod == "file" {
+	bootstrapMethod := conf.General.BootstrapMethod
+	if bootstrapMethod == "file" || bootstrapMethod == "none" {
 		bootstrapBlock := extractBootstrapBlock(conf)
+		// loop through all channels to obtain system channel when bootstrapmethod is set to none.
```
yacovm: The loop is in the function so let's remove this comment?

orderer/common/server/main.go
```diff
@@ -124,8 +124,13 @@ func Main() {
...
+		if bootstrapMethod == "none" {
```
yacovm: Why restrict to the "none" method? Why not also use the extraction from the ledger in the case the extractBootstrapBlock(conf) is nil?

orderer/common/server/etcdraft_test.go
```diff
@@ -249,3 +265,35 @@ func launchOrderer(gt *GomegaWithT, orderer, tempDir, genesisBlockPath, fabricRo
 	gt.Expect(err).NotTo(HaveOccurred())
 	return sess
 }
+
+func restartOrderer(gt *GomegaWithT, orderer, tempDir, genesisBlockPath, cryptoPath string) *gexec.Session {
```
yacovm: there is a "launch orderer" method above this one. Can't we use it somehow? do we have to copy all the env vars?

orderer/common/server/main.go
```diff
@@ -703,15 +727,16 @@ func initializeMultichannelRegistrar(
...
 	registrar := multichannel.NewRegistrar(*conf, lf, signer, metricsProvider, bccsp, callbacks...)
 
 	consenters := map[string]consensus.Consenter{}
 
 	var icr etcdraft.InactiveChainRegistry
-	if conf.General.BootstrapMethod == "file" && isClusterType(bootstrapBlock, bccsp) {
-		etcdConsenter := initializeEtcdraftConsenter(consenters, conf, lf, clusterDialer, bootstrapBlock, ri, srvConf, srv, registrar, metricsProvider, bccsp)
-		icr = etcdConsenter.InactiveChainRegistry
+	if conf.General.BootstrapMethod == "file" || conf.General.BootstrapMethod == "none" {
+		if bootstrapBlock != nil && isClusterType(bootstrapBlock, bccsp) {
```
yacovm, line 736: can the bootstrapBlock be nil here? We don't have the check on the left side, so how can it be nil? Wouldn't we have crashed because we didn't have this check in the past?

And let's say bootstrap block is nil, doesn't that mean that you don't initialize the etcdraft? Isn't that a problem?

orderer/common/server/etcdraft_test.go
```diff
@@ -120,11 +120,27 @@ func generateCryptoMaterials(gt *GomegaWithT, cryptogen, path string) string {
...
+func testEtcdRaftOSNRestart(gt *GomegaWithT, tempDir, configtxgen, orderer, cryptoPath string) {
+	genesisBlockPath := generateBootstrapBlock(gt, tempDir, configtxgen, "system", "SampleEtcdRaftSystemChannel")
+
+	// Launch the OSN
+	ordererProcess := launchOrderer(gt, orderer, tempDir, genesisBlockPath, cryptoPath)
+	defer func() { gt.Eventually(ordererProcess.Kill(), time.Minute).Should(gexec.Exit()) }()
+	gt.Eventually(ordererProcess.Err, time.Minute).Should(gbytes.Say("Beginning to serve requests"))
+	gt.Eventually(ordererProcess.Err, time.Minute).Should(gbytes.Say("becomeLeader"))
+
+	// Restart orderer with ORDERER_GENERAL_BOOTSTRAPMETHOD = none
```
yacovm, line 132: I am not sure this is enough... shouldn't we also delete the file?

orderer/common/server/main.go
DereckLuo: done

orderer/common/server/main.go
DereckLuo: good point, done.

orderer/common/server/etcdraft_test.go
DereckLuo: done

orderer/common/server/main.go
DereckLuo: bootstrapBlock should not be nil here, however, in test _TestInitializeMultichannelRegistrar/"registrar without a system channel"_, it directly calls **initializeMultichannelRegistrar**, and passed in bootstrapBlock as nil, and BootstrapMethod as "none". In this case, the test will pass the old code as it will only check cluster type if BootstrapMethod is "file", but not the new code.

yacovm: So why not **not** pass a nil bootstrap block then? we shouldn't change the production code just to fit tests...

DereckLuo: It's definitely good point, we should never change production code to just fit tests. 

However, currently orderer supports an action to startup without system channel defined ([FAB-15709](https://jira.hyperledger.org/browse/FAB-15709)), which sets **BootstrapMethod** to "none" and **bootstrapBlock** to be nil. 

And in this PR, we are supporting restarting orderer without providing genesisblock, which was being requested to set **BootstrapMethod** to "none", and obtain **bootstrapBlock** from system channel ([FAB-17774](https://lists.hyperledger.org/g/fabric/message/8137)). Hence, the nil check here is to distinguish the two cases.

We can also adding a new bootstrapMethod phrase here, instead of using "none" for both use cases. I'm opening to any suggestions.

yacovm (approved):

PR #1204, opened 2020-05-01
Remove GOPATH references in CI artifacts

sykesm:
With the move to modules in master, `GOPATH` should be a thing of the past. Remove the remaining references to prevent confusion and help ensure an implicit dependency does not return.

### DIFF
--- a/ci/azure-pipelines-release.yml
+++ b/ci/azure-pipelines-release.yml
@@ -8,8 +8,6 @@ pr: none
 
 variables:
   - group: credentials
-  - name: GOPATH
-    value: $(Agent.BuildDirectory)/go
   - name: GOVER
     value: 1.14.1
 
@@ -32,7 +30,7 @@ stages:
               TARGET: windows-amd64
         steps:
           - checkout: self
-            path: 'go/src/github.com/hyperledger/fabric'
+            path: 'fabric'
             displayName: Checkout Fabric Code
           - script: ./ci/scripts/create_binary_package.sh
             displayName: Compile Binary and Create Tarball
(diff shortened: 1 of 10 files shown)


ci/azure-pipelines-release.yml
```diff
@@ -32,7 +30,7 @@ stages:
               TARGET: windows-amd64
         steps:
           - checkout: self
-            path: 'go/src/github.com/hyperledger/fabric'
+            path: 'fabric'
```
sykesm, line 33: I'm a little concerned about this change since I don't know the full contents of the golang docker image. In particular whether or not `$(go env GOPATH)/bin` is on the path.

Then again, since this is just building, it may not matter.

lindluni: This should be alright, but I'll test it out of band against my own repo

lindluni (changes requested):
Thanks for doing this, but a small ask, the only reason we had to include the checkout step was because we had to put the repo on the GOPATH. Can you remove the checkout step completely, the pipeline by default handles checking out the code and putting it into the AZP working directory, you only need to include the checkout step if you need to do something not handled by the default functionality.

sykesm:
> the pipeline by default handles checking out the code and putting it into the AZP working directory

The doc implies it throws it into a directory named `s`. That just rubs me the wrong way.

lindluni:
> > the pipeline by default handles checking out the code and putting it into the AZP working directory
> 
> The doc implies it throws it into a directory named `s`. That just rubs me the wrong way.

It does, all of their default directories are named on that schema and I would expect it to have checked the code out just as any other standard checkout would have (into a folder named after the repo), so I can sympathize. But given that we don't have remote access to the machines, my preference is to cleanup the pipeline file rather than override the defaults. Especially since it's expected that you would use the AZP environment variable `Build.SourcesDirectory` to reference it rather than navigate to it yourself.

lindluni:
If you really have strong feelings against it, I'm willing to bend and take it as is, your call if you want to merge as is, or remove.

lindluni (dismissed):

sykesm:
I'll come back to this in a bit.

PR #1210, opened 2020-05-04
[LEDGER] reduce #arguments in a few kvledger methods

cendhu:
#### Type of change

#### Description

As the number of arguments passed to `newKVLedger()`, `initTxMgr()`, and `NewLockBasedTxMgr()` is high, we reduce it by passing the initializer struct itself.

#### Additional details

In subsequent PRs, we would think of moving the `BookeepingProvider` and `DB` to the `txmgr` pkg and instead create a `TxMgrProvider` in the kvLedger to reduce the number of arguments further by 1. Moreover, it is natural to initiate structs that are directly used by kvLedger in the kvledger pkg. For now, the  `BookeepingProvider` and `DB` are not directly used by the kvLedger.

#### Related issues

[FAB-17683](https://jira.hyperledger.org/browse/FAB-17683)

### DIFF
--- a/core/ledger/kvledger/kv_ledger.go
+++ b/core/ledger/kvledger/kv_ledger.go
@@ -54,43 +54,49 @@ type kvLedger struct {
 	isPvtstoreAheadOfBlkstore atomic.Value
 }
 
-func newKVLedger(
-	ledgerID string,
-	blockStore *blkstorage.BlockStore,
-	pvtdataStore *pvtdatastorage.Store,
-	versionedDB privacyenabledstate.DB,
-	historyDB *history.DB,
-	configHistoryMgr confighistory.Mgr,
-	stateListeners []ledger.StateListener,
-	bookkeeperProvider bookkeeping.Provider,
-	ccInfoProvider ledger.DeployedChaincodeInfoProvider,
-	ccLifecycleEventProvider ledger.ChaincodeLifecycleEventProvider,
-	stats *ledgerStats,
-	customTxProcessors map[common.HeaderType]ledger.CustomTxProcessor,
-	hasher ledger.Hasher,
-) (*kvLedger, error) {
+type lgrInitializer struct {
+	ledgerID                 string
+	blockStore               *blkstorage.BlockStore
+	pvtdataStore             *pvtdatastorage.Store
+	versionedDB              privacyenabledstate.DB
+	historyDB                *history.DB
+	configHistoryMgr         confighistory.Mgr
+	stateListeners           []ledger.StateListener
+	bookkeeperProvider       bookkeeping.Provider
+	ccInfoProvider           ledger.DeployedChaincodeInfoProvider
+	ccLifecycleEventProvider ledger.ChaincodeLifecycleEventProvider
+	stats                    *ledgerStats
+	customTxProcessors       map[common.HeaderType]ledger.CustomTxProcessor
+	hasher                   ledger.Hasher
+}
+
+func newKVLedger(initializer *lgrInitializer) (*kvLedger, error) {
+	ledgerID := initializer.ledgerID
 	logger.Debugf("Creating KVLedger ledgerID=%s: ", ledgerID)
 	l := &kvLedger{
 		ledgerID:        ledgerID,
-		blockStore:      blockStore,
-		pvtdataStore:    pvtdataStore,
-		historyDB:       historyDB,
+		blockStore:      initializer.blockStore,
+		pvtdataStore:    initializer.pvtdataStore,
+		historyDB:       initializer.historyDB,
 		blockAPIsRWLock: &sync.RWMutex{},
 	}
 
-	btlPolicy := pvtdatapolicy.ConstructBTLPolicy(&collectionInfoRetriever{ledgerID, l, ccInfoProvider})
+	btlPolicy := pvtdatapolicy.ConstructBTLPolicy(&collectionInfoRetriever{ledgerID, l, initializer.ccInfoProvider})
 
-	if err := l.initTxMgr(
-		versionedDB,
-		stateListeners,
-		btlPolicy,
-		bookkeeperProvider,
-		ccInfoProvider,
-		customTxProcessors,
-		hasher,
-	); err != nil {
+	txmgrInitializer := &lockbasedtxmgr.Initializer{
+		LedgerID:            ledgerID,
+		DB:                  initializer.versionedDB,
+		StateListeners:      initializer.stateListeners,
+		BtlPolicy:           btlPolicy,
+		BookkeepingProvider: initializer.bookkeeperProvider,
+		CCInfoProvider:      initializer.ccInfoProvider,
+		CustomTxProcessors:  initializer.customTxProcessors,
+		Hasher:              initializer.hasher,
+	}
+	if err := l.initTxMgr(txmgrInitializer); err != nil {
 		return nil, err
 	}
+
 	// btlPolicy internally uses queryexecuter and indirectly ends up using txmgr.
 	// Hence, we need to init the pvtdataStore once the txmgr is initiated.
 	l.pvtdataStore.Init(btlPolicy)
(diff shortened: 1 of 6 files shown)


manish-sethi:
@cendhu - I don't think that this change is in the right direction. The dependencies were being passed explicitly, as is typically desired.

lindluni:
> @cendhu - I don't think that this change is in the right direction. The dependencies were being passed explicitly, as is typically desired.

I don't have an extensive background on what is right or wrong here, but I tend to support this change. The simplification greatly reduced the complexity of following the code. The object being passed around (and it remains in the kvledger package, and in the same file) the same package until it is ultimately passed off to another package (the txmgr) where the necessary parameters are extracted from the object before being passed out of the package.

If I was tracing this code I have to track the same 5 parameters through multiple function calls, making it difficult to keep track of mentally. With Senthil's change I can simply follow the object until it's ultimately extracted within the current package and passed off for consumption to a single function. The context is much simpler.

It's also worth noting, that if you were to turn on the code smell feature of many of the most popular Golang linting tools, a function taking in this many number of arguments would NOT pass the smell tests as being unusually complicated and not being mentally digestible.

As is typically desired doesn't always mean mandatory. The reduction in complexity IMO is worth breaking from what might be typically desired.

core/ledger/kvledger/kv_ledger.go
```diff
@@ -78,24 +69,17 @@ func newKVLedger(
...
-	if err := l.initTxMgr(
-		versionedDB,
-		stateListeners,
-		btlPolicy,
-		bookkeeperProvider,
-		ccInfoProvider,
-		customTxProcessors,
-		hasher,
-	); err != nil {
+	err := l.initTxMgr(versionedDB, lgrProvider.stateListeners, btlPolicy, lgrProvider)
```
lindluni: you're already passing in the `lgrProvider` you can pull the `lgrProvider.stateListeners` off the object in the callee instead of passing it down explicitly.

manish-sethi: I am not against clubbing the parameters to reduce the number of parameters. My objection is to the way it is achieved here. i.e., pass down the `provider` instance much deeper in the code where we are supposed to have the code specific to one ledger instance. This is overusing the `provider`, which is not merely a holder of objects but has an associated behavior as well.

manish-sethi:
> > @cendhu - I don't think that this change is in the right direction. The dependencies were being passed explicitly, as is typically desired.
> 
> I don't have an extensive background on what is right or wrong here, but I tend to support this change. The simplification greatly reduced the complexity of following the code. The object being passed around (and it remains in the kvledger package, and in the same file) the same package until it is ultimately passed off to another package (the txmgr) where the necessary parameters are extracted from the object before being passed out of the package.
> 
> If I was tracing this code I have to track the same 5 parameters through multiple function calls, making it difficult to keep track of mentally. With Senthil's change I can simply follow the object until it's ultimately extracted within the current package and passed off for consumption to a single function. The context is much simpler.
> 
> It's also worth noting, that if you were to turn on the code smell feature of many of the most popular Golang linting tools, a function taking in this many number of arguments would NOT pass the smell tests as being unusually complicated and not being mentally digestible.
> 
> As is typically desired doesn't always mean mandatory. The reduction in complexity IMO is worth breaking from what might be typically desired.

I am not against clubbing the parameters to reduce the number of parameters. My objection is to the way it is achieved here. i.e., pass down the provider instance much deeper in the code where we are supposed to have the code specific to one ledger instance. This is overusing the provider, which is not merely a holder of objects but has an associated behavior as well.
For instance, the same problem was there when passing the dependencies from peer to ledger and they were clubbed in this struct - https://github.com/hyperledger/fabric/blob/7778c49212fe42cb55779fe9a089deb781ea0ef2/core/ledger/ledgermgmt/ledger_mgmt.go#L45

cendhu:
@manish-sethi As @btl5037 mentioned, having too many arguments make the code readability a bit hard. I have faced it multiple times when adding/making changes to kvledger package. Let me share the thought process I had before submitting this PR.

1. given that `newKVLedger()` uses many variables present in the `Provider`, I thought of moving `newKVLedger()` method under `Provider` struct itself and renaming it to `initKVLedger()` but that might break the current file organization and wasn't sure whether you would approve it. In the earlier days, `kv_ledger_provider.go` and `kv_ledger.go` confused me a lot and now I know about them. Anyway, I didn't want to do a bigger refactoring in this PR. 
2. when we call `newKVLedger()`, we do pass most of the arguments needed to construct the `kvLedger` struct but `txmgr`. Hence, I was thinking of directly creating `kvLedger` struct from the `Provider` and calling `kvLedger.init()` but again I wasn't sure whether newXXX only should create the sturct and hence, I dropped that option too. Moreover, we pass all needed items from Provider except txmgr which did not look correct to me. As can be seen in the PR description, I am planning to introduce something called txmgr provider.
3. finally I was thinking of introducing a struct called `ledgerConfig` and `txmgrConfig` but I thought this might collide with the ledgerConfig passed from start.go. 

I didn't think of the initializer struct. Are you suggesting to introduce something like `kvLedgerInitializer` struct and pass it to newKVLedger()? Similarly, we can introduce `txmgrInitializer` to reduce the number of arguments in `initTxmgr()`

manish-sethi:
> 1. given that `newKVLedger()` uses many variables present in the `Provider`, I thought of moving `newKVLedger()` method under `Provider` struct itself and renaming it to `initKVLedger()` 

I am fine with this approach, if you want to go that way.

> Moreover, we pass all needed items from Provider except txmgr which did not look correct to me. As can be seen in the PR description, I am planning to introduce something called txmgr provider.

There is a difference between the other items that are passed in and the txmgr. The other items are either passed-in dependencies from other packages (e.g., `DeployedChaincodeInfoProvider`) or they are shared stores (e.g., leveldbs), whereas, the txmgr is just an internal struct which is constructed one for each ledger - and hence constructed internally.
 
> I didn't think of the initializer struct. Are you suggesting to introduce something like `kvLedgerInitializer` struct and pass it to newKVLedger()? Similarly, we can introduce `txmgrInitializer` to reduce the number of arguments in `initTxmgr()`

Yes, that's what I meant.

So, either (or mix of both) of `provider.newKVLedger()` and `txmgrInitializer` works for me.

Finally, IMO, the bigger problem with this function is that some of the dependencies of initialization ordering needs a revisit... so, various `init`'s can be removed

mastersingh24:
@manish-sethi  is on to something here.  I'd say that this is worthy of a bigger refactor (although I do see the value in reducing the number of parameters where possible).

We really need to define the hierarchy of structs here and try to avoid side effects as best as possible.  For example, does a provider have multiple ledgers or do ledgers have a provider (where multiple ledgers can share a provider)

cendhu:
@mastersingh24 Yes, Manish was planning to refactor these dependencies ( some are ledger sub-packages and some are not) as part of validation refactoring epic (if I remember correctly). For now, we have decided to defer a few bigger refactoring work as we need to move to ledger checkpointing work. We would still do small refactoring while we code for the ledger checkpointing and defer the bigger refactoring which involves the following packages:

1. mvcc `validation` pkg. 
2. `kvledger` pkg
3. merge `kvledger` pkg with `ledger` pkg to reduce folder depth.
4. merge `lockbasedtxmgr` pkg with the `txmgr` pkg.
5. merge `privacyenabledstate` pkg with the `lockbasedtxmgr` pkg

and a few more.

mastersingh24:
> @mastersingh24 Yes, Manish was planning to refactor these dependencies ( some are ledger sub-packages and some are not) as part of validation refactoring epic (if I remember correctly). For now, we have decided to defer a few bigger refactoring work as we need to move to ledger checkpointing work. We would still do small refactoring while we code for the ledger checkpointing and defer the bigger refactoring which involves the following packages:
> 
> 1. mvcc `validation` pkg.
> 2. `kvledger` pkg
> 3. merge `kvledger` pkg with `ledger` pkg to reduce folder depth.
> 4. merge `lockbasedtxmgr` pkg with the `txmgr` pkg.
> 5. merge `privacyenabledstate` pkg with the `lockbasedtxmgr` pkg
> 
> and a few more.

Got it - makes sense

manish-sethi (approved):

core/ledger/kvledger/txmgmt/txmgr/lockbasedtxmgr/lockbased_txmgr.go
```diff
@@ -59,36 +59,46 @@ func (c *current) maxTxNumber() uint64 {
...
-	bookkeepingProvider bookkeeping.Provider,
-	ccInfoProvider ledger.DeployedChaincodeInfoProvider,
-	customTxProcessors map[common.HeaderType]ledger.CustomTxProcessor,
-	hasher ledger.Hasher,
-) (*LockBasedTxMgr, error) {
-
-	if hasher == nil {
+func NewLockBasedTxMgr(initializer *Initializer) (*LockBasedTxMgr, error) {
+
+	if initializer.Hasher == nil {
```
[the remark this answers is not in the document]
cendhu: I agree. At least we can move this check to kvledger package for now.

PR #1212, opened 2020-05-05
[FAB-17831] Use generic constant/var names for ledger data format

wenjianqiao:

#### Type of change

#### Description
Currently Version1x and Version20 are defined in common/ledger/dataformat.go.
They should be renamed to more generic names such as PreviousVersion and CurrentVersion.
Note that these versions are used for the data format in ledger databases. They would change only when the data format is changed. 

#### Related issues
https://jira.hyperledger.org/browse/FAB-17831

### DIFF
--- a/common/ledger/dataformat/dataformats.go
+++ b/common/ledger/dataformat/dataformats.go
@@ -8,35 +8,36 @@ package dataformat
 
 import "fmt"
 
+// The data format stored in ledger databases may be changed in a new release to
+// support new features. This file defines the constants to check whether or not
+// the data format on the ledger matches the CurrentFormat. If not matched, peer start
+// will fail and the ledger must be upgraded to the CurrentFormat. If a Fabric version
+// does not introduce a new data format, CurrentFormat will remain the same as the latest
+// format prior to the Fabric version.
 const (
-	// Version1x specifies the data format in version 1.x
-	Version1x = ""
+	// PreviousFormat specifies the data format in previous fabric version
+	PreviousFormat = ""
 
-	// Version20 specifies the data format in version 2.0
-	Version20 = "2.0"
+	// CurrentFormat specifies the data format in current fabric version
+	CurrentFormat = "2.0"
 )
 
-var (
-	// Version1xBytes defines the []byte for Version1x to be nil
-	Version1xBytes []byte = nil
-)
-
-// ErrVersionMismatch is returned if it is detected that the version of the format recorded in
+// ErrFormatMismatch is returned if it is detected that the version of the format recorded in
 // the internal database is different from what is specified in the `Conf` that is used for opening the db
-type ErrVersionMismatch struct {
-	DBInfo          string
-	ExpectedVersion string
-	Version         string
+type ErrFormatMismatch struct {
+	DBInfo         string
+	ExpectedFormat string
+	Format         string
 }
 
-func (e *ErrVersionMismatch) Error() string {
+func (e *ErrFormatMismatch) Error() string {
 	return fmt.Sprintf("unexpected format. db info = [%s], data format = [%s], expected format = [%s]",
-		e.DBInfo, e.Version, e.ExpectedVersion,
+		e.DBInfo, e.Format, e.ExpectedFormat,
 	)
 }
 
-// IsVersionMismatch returns true if err is an ErrVersionMismatch
+// IsVersionMismatch returns true if err is an ErrFormatMismatch
 func IsVersionMismatch(err error) bool {
-	_, ok := err.(*ErrVersionMismatch)
+	_, ok := err.(*ErrFormatMismatch)
 	return ok
 }
--- a/common/ledger/blkstorage/blockstore_provider.go
+++ b/common/ledger/blkstorage/blockstore_provider.go
@@ -65,8 +65,8 @@ type BlockStoreProvider struct {
 // NewProvider constructs a filesystem based block store provider
 func NewProvider(conf *Conf, indexConfig *IndexConfig, metricsProvider metrics.Provider) (*BlockStoreProvider, error) {
 	dbConf := &leveldbhelper.Conf{
-		DBPath:                conf.getIndexDir(),
-		ExpectedFormatVersion: dataFormatVersion(indexConfig),
+		DBPath:         conf.getIndexDir(),
+		ExpectedFormat: dataFormatVersion(indexConfig),
 	}
 
 	p, err := leveldbhelper.NewProvider(dbConf)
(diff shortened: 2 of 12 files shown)


cendhu:
@wenjianqiao Can you give a bit more context here? When we move to version 3, will we change the current version to 3.0 and previous version to 2.x?

manish-sethi:
> @wenjianqiao Can you give a bit more context here? When we move to version 3, will we change the current version to 3.0 and previous version to 2.x?

@cendhu - yes, that was the intent

common/ledger/dataformat/dataformats.go
```diff
@@ -9,16 +9,19 @@ package dataformat
 import "fmt"
 
 const (
-	// Version1x specifies the data format in version 1.x
-	Version1x = ""
+	// PreviousVersion specifies the data format in previous version (1.x for now)
+	PreviousVersion = ""
```
cendhu:
```suggestion
	LastMajorVersion = ""
```
This is because both 2.0 and 2.1 are called as two different versions. It would be good if we can be explicit.

common/ledger/dataformat/dataformats.go
```diff
@@ -9,16 +9,19 @@ package dataformat
...
 
-	// Version20 specifies the data format in version 2.0
-	Version20 = "2.0"
+	// CurrentVersion specifies the data format in current version (2.0 for now)
+	CurrentVersion = "2.0"
```
cendhu:
```suggestion
	CurrentMajorVersion = "2.0"
```

cendhu (commented):
Thanks, @manish-sethi . Does my suggestion make sense? Only you know the clear context here and I leave it to you.

common/ledger/dataformat/dataformats.go
manish-sethi: These versions are not as such related to fabric versions. As you see the package name is `dataformat`, so they would change only when the data format is changed.

wenjianqiao:
Hit a CI flake. Log is attached to FAB-17451.
FAIL: TestWithStaticDeliverClientLeader (46.00s)
gossip_service_test.go:626: Failed to establish full channel membership. Only 0 out of 2 peers have full membership

common/ledger/dataformat/dataformats.go
wenjianqiao: @cendhu Thank you for your comments. As @manish-sethi  said, these versions are for the data formats in ledger databases (e.g, dataformat.CurrentVersion). They may or may not change when fabric version is upgraded. Currently, we don't have the concept for "major" data format version.

cendhu: I see. Thanks. I think the confusion was due to the comment `CurrentVersion specifies the data format in the current version (2.0 for now)` and we have `CurrentVersion = "2.0"`. In that case, can we rename `CurrentVersion` to `CurrentDataVersion` and `PreviousVersion` to `PreviousDataVersion`.   Even some other name is fine but let's be more explicit to avoid hidden meaning. I think it is better to remove `2.0` as well but we cannot do that now, right? Some of the networks that have migrated to 2.0 would already have this format. Usually, `2.x` can be interpreted as 2 -- major version, x -- minor version. I assume we don't support major.minor version for the data format. When we change the data format in the future, we can use non-decimal number to avoid these confusions.

manish-sethi: I am fine with the names `CurrentDataVersion` and `PreviousDataVersion`, if @wenjianqiao is fine. 
Yes, version is just a string for us and we cannot change `2.0`. The intention was to keep track when this format was introduced than anything else. So, in a way, that a binary is expected to work with the greatest data format only where  `data format version <= binary version`. @wenjianqiao - may be some of this description, you may like to add in the data format file.

common/ledger/dataformat/dataformats.go
```diff
@@ -9,16 +9,19 @@ package dataformat
...
 )
 
 var (
-	// Version1xBytes defines the []byte for Version1x to be nil
-	Version1xBytes []byte = nil
+	// PreviousVersionBytes specifies the []byte for PreviousVersion
+	PreviousVersionBytes []byte = nil
+
+	// CurrentVersionBytes specifies the []byte for CurrentVersion
+	CurrentVersionBytes = []byte(CurrentVersion)
```
cendhu: Do we need these? Can't the consumer simply convert the version string to bytes when needed?

common/ledger/dataformat/dataformats.go
wenjianqiao: I am fine with the names CurrentDataVersion and PreviousDataVersion

common/ledger/dataformat/dataformats.go
wenjianqiao: The format is stored as []byte in a db. There was no format key in fabric v1.x, so we compare to "nil" for upgrade eligibility. However, we cannot convert the PreviousVersion string to nil bytes. CurrentVersionBytes can be converted from CurrentVersion string, but it is added to be same as PreviousVersionBytes. I don't have strong opinion to keep or remove CurrentVersionBytes.

cendhu: A general suggestion is to avoid package-level variables (for Fabric). I am not getting into that. My question is more about why we need to define variables such as **XXXbytes** to just represent a byte version of **XXX** const. I am fine with this usage when it is used within a function as local variables but certainly not as exported variables (when both **XXX** and **XXXbytes** are exported). 

Can't we use 
```
if bytes.Equal(format, []byte(dataformat.PreviousVersion)) 
if bytes.Equal(format, []byte(dataformat.CurrentVersion))
```
 instead of 
```
if bytes.Equal(format, dataformat.PreviousVersionBytes)
if bytes.Equal(format, dataformat.CurrentVersionBytes)
```
?

Even if the stored version of the data is nil, can't we compare it against `[]byte(dataformat.PreviousVersion)` given that the `dataformat.PreviousVersion` is `""`? 

The following bytes.Equal would return true. 
```
var dbVersion1x []byte = nil
var previousVersion = ""
fmt.Println(bytes.Equal(dbVersion1x, []byte(previousVersion)))
fmt.Println(bytes.Equal(nil, []byte(previousVersion)))
fmt.Println(bytes.Equal(nil, []byte("")))
fmt.Println(bytes.Equal([]byte(nil), []byte("")))
```
You can play with this code here -- https://play.golang.org/p/WywdtwFTGvo

Let me know if I am missing anything.

wenjianqiao: Thank you. Will remove XXXBytes.

common/ledger/dataformat/dataformats.go
wenjianqiao: Talked offline and agreed to change the constants to CurrentFormat and PreviousFormat.

manish-sethi (approved):

PR #1214, opened 2020-05-05
[FAB-17751] Fixed RAFT OSN panic with repeatedly adding/removing

DereckLuo:

#### Type of change

- Bug fix

#### Description

* fixed RAFT OSN panic with repeatedly adding/removing OSN from the
  channel.

#### Related issues

[FAB-17751](https://jira.hyperledger.org/browse/FAB-17751)

### DIFF
--- a/integration/raft/config_test.go
+++ b/integration/raft/config_test.go
@@ -80,10 +80,6 @@ var _ = Describe("EndToEnd reconfiguration and onboarding", func() {
 			network = nwo.New(nwo.MultiNodeEtcdRaft(), testDir, client, StartPort(), components)
 			network.GenerateConfigTree()
 			network.Bootstrap()
-
-			networkRunner := network.NetworkGroupRunner()
-			networkProcess = ifrit.Invoke(networkRunner)
-			Eventually(networkProcess.Ready(), network.EventuallyTimeout).Should(BeClosed())
 		})
 
 		// This tests:
@@ -93,6 +89,10 @@ var _ = Describe("EndToEnd reconfiguration and onboarding", func() {
 		// 3. raft orderer processes type A config updates and delivers the
 		//    config blocks to the peers.
 		It("executes an etcdraft network with 2 orgs and three orderer nodes", func() {
+			networkRunner := network.NetworkGroupRunner()
+			networkProcess = ifrit.Invoke(networkRunner)
+			Eventually(networkProcess.Ready(), network.EventuallyTimeout).Should(BeClosed())
+
 			orderer1 := network.Orderer("orderer1")
 			orderer2 := network.Orderer("orderer2")
 			orderer3 := network.Orderer("orderer3")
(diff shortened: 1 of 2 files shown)


yacovm:
How was this tested? description of what happened, why, etc? 
Isn't it possible to write an integration test that reproduces this?

C0rWin:
> How was this tested? description of what happened, why, etc?
> Isn't it possible to write an integration test that reproduces this?

Moreover, I wonder what was the root cause of this bug?

yacovm:
> > How was this tested? description of what happened, why, etc?
> > Isn't it possible to write an integration test that reproduces this?
> 
> Moreover, I wonder what was the root cause of this bug?

That's what I meant in: "description of what happened, why, etc?"

C0rWin (changes requested):
The new test you have introduced adds 80 second of artificial delay prolonging the execution of integration test suite, while I'm under impression one of the goals is to reduce and optimize tests execution time. Moreover it's not very clear why would you need so many `time.Sleep()`.

Could you also please add a short explanation to the description of what exactly you are trying to fix and what is the use you attempt to re-create?

integration/raft/config_test.go
```diff
@@ -136,6 +136,89 @@ var _ = Describe("EndToEnd reconfiguration and onboarding", func() {
...
+
+			By("assert Block reception")
+			assertBlockReception(map[string]int{
+				"testchannel":   0,
+				"systemchannel": 1,
+			}, []*nwo.Orderer{o1, o2, o3}, peer, network)
+
+			fmt.Fprintf(GinkgoWriter, "0. Leader node = %d\n", findLeader(ordererRunners)) // 1,2,3
+
+			extendNetwork(network)
```
yacovm, line 171: why do you need to do that? Can't you reproduce this by sticking to the existing cluster size?

integration/raft/config_test.go
```diff
@@ -136,6 +136,89 @@ var _ = Describe("EndToEnd reconfiguration and onboarding", func() {
...
+			certificatesOfOrderers := refreshOrdererPEMs(network)
+
+			By("1. Removing o1")
```
yacovm, line 174: 1) Are you sure we need to remove 2 consenters? Isn't removing a single one sufficient to reproduce the behavior?
2) When you remove a consenter - it might have been the leader, so the cluster needs to stabilize. 10 seconds might or might not be enough, so let's find another means to ensure the remaining consenters see the leader ?

DereckLuo:
> The new test you have introduced adds 80 second of artificial delay prolonging the execution of integration test suite, while I'm under impression one of the goals is to reduce and optimize tests execution time. Moreover it's not very clear why would you need so many `time.Sleep()`.
> 
> Could you also please add a short explanation to the description of what exactly you are trying to fix and what is the use you attempt to re-create?

@C0rWin Thanks for the review. The `time.Sleep()`was that added in the test is trying to make sure the cluster is stabilized after adding/removing orderers. Totally agree that this is not the best ways to do so, and  in addition as Yacov mentioned above, 10 seconds might not be enough, and I'll fix it with other ways to check cluster stabilization. 

The initial bug was opened in [FAB-17751](https://jira.hyperledger.org/browse/FAB-17751), where in `chain.Apply()` function, when `c.accDataSize >= c.sizeLimit ` ([code](https://github.com/hyperledger/fabric/blob/master/orderer/consensus/etcdraft/chain.go#L1050)) , the unmarshalled block might be an empty block and causes the access to block header number to panic. He provided a test with repeatly add/remove OSN which re-create this bug.

DereckLuo:
with further investigation, this bug seems to be only reproducible in commit `b03b3d9fe`, but cannot be re-created in both `master` and `release-1.4` branch. Going to close this PR for now.

yacovm:
How is that possible? It necessarily means we had a commit that introduced a regression and then it was fixed. Shouldn't we investigate what happened?

yacovm:
And, didn't you reproduce it with that integration test on the master branch?

DereckLuo:
> How is that possible? It necessarily means we had a commit that introduced a regression and then it was fixed. Shouldn't we investigate what happened?

Yes, I was not able to reproduce the error against release-1.4 for the past three days. Just closing this PR for now, but not the Jira. As I'm still looking into this issue, and try to reproduce the bug both against `release-1.4`, and `b03b3d9fe`. Sorry about the confusion on closing out this PR.

PR #1221, opened 2020-05-07
[LEDGER] tests for commit hashes with rollback and reset

cendhu:
#### Type of change

- Test update

#### Description
The `COMMIT_HASH` is being added to each block both in v1.1 and >= v1.4 As this hash has to be the same on all peers for a given block, it is necessary to start adding this hash from the same block on all peers (as it a cumulative hash, i.e., it includes the previous `COMMIT_HASH`). Hence, it is necessary to reset the existing channels to the genesis block inorder to add `COMMIT_HASH` feature. Otherwise, the peer would not add COMMIT_HASH to any blocks.

This PR adds the following tests  

1. ensure that the `COMMIT_HASH` is being added to the block once a peer has been reset to the genesis block. Otherwise, COMMIT_HASH would not be added.
2. assuming that `COMMIT_HASH `is already present, doing a rollback/reset and reprocessing blocks should not result in different `COMMIT_HASH` (as we assume no changes in the validation results). This test also helps to ensure that any changes to the code that computes the `COMMIT_HASH` do not break the backward compatibility.

#### Additional details

These tests are not new and had been added in v1.4 https://github.com/hyperledger/fabric/commit/7e65b51f0d21d3a317f129f031aa96b02b60275f However, it was never ported to v2.x Hence, I have cherry-picked them and made following changes

1. converted `sample_ledgers_with_commit_hashes/ledgersData` to `sample_ledgers_with_commit_hashes/ledgersData.zip`
2. removed `closeAndReopenLgr()`
3. changed `resetFun func(*testhelper)` to `resetFun func(*testhelper, string)`
4. used new env that does not exist in v1.4 along with the integration of UpgradeDBs command.

No changes are made to the actual test logic. 

#### Related issues

[FAB-17832](https://jira.hyperledger.org/browse/FAB-17832)


manish-sethi (approved):
Thanks @cendhu for digging this out - not sure, how I missed this porting to 2.x.

Merging this as is. However, this still does not suffice for verifying that the commit hash computation is backward compatible. The test in this patch checks thoroughly that the commit hash is generated when it is supposed to be. But, I was more concerned about the backward compatibility of the `commit hash value`.

As I mentioned before, in the recent PR (https://github.com/hyperledger/fabric/pull/1209) the UT is a decent safety check but not a good measure for e2e behavior. For instance, if in that PR, you would have changed the proto message tag number ordering, the UT would have passed but commit hash value would not have been backward compatible.

Given, this PR, it should be easy to enhance the existing test to include this verification. We need to add, perhaps here (https://github.com/hyperledger/fabric/pull/1221/files#diff-0449fd005d0e85d0b0e7772d9f134cd5R196), something of the following effect...

```
if preResetCommitHashExists {
	verify that the hashes in the block at the location `ledgerFSRoot` are the same before and after executing function `resetFunc(h, ledgerFSRoot)`
}
```

PR #1222, opened 2020-05-07
[FAB-17846] Replace sample scripts for external chaincode with /bin/s…

satota2:
…h based ones

#### Type of change

- Documentation update

#### Description

This patch replaces `/bin/bash` based sample scripts for external chaincode builder and launcher with `/bin/sh` based ones for the following reasons:
* Peer images use Alpine Linux, which does not contain bash.
* Scripts with bash shebang are not executed in the peers but it is not easy to detect it from the peer logs.

The patch includes some fixes to comply with the POSIX specifications.

#### Related issues

- https://jira.hyperledger.org/browse/FAB-17846

### DIFF
--- a/docs/source/cc_launcher.md
+++ b/docs/source/cc_launcher.md
@@ -8,6 +8,8 @@ Starting with Fabric 2.0, External Builders and Launchers address these limitati
 
 Note that if no configured external builder claims a chaincode package, the peer will attempt to process the package as if it were created with the standard Fabric packaging tools such as the peer CLI or node SDK.
 
+**Note:** This is an advanced feature that will likely require custom packaging of the peer image. For example, the following samples use `go` and `bash`, which are not included in the current official `fabric-peer` image.
+
 ## External builder model
 
 Hyperledger Fabric External Builders and Launchers are loosely based on Heroku [Buildpacks](https://devcenter.heroku.com/articles/buildpack-api). A buildpack implementation is simply a collection of programs or scripts that transform application artifacts into something that can run. The buildpack model has been adapted for chaincode packages and extended to support chaincode execution and discovery.
--- a/docs/source/cc_service.md
+++ b/docs/source/cc_service.md
@@ -13,6 +13,8 @@ The rest of this topic describes how to configure chaincode as an external servi
 * [Deploying the chaincode](#deploying-the-chaincode)
 * [Running the chaincode as an external service](#running-the-chaincode-as-an-external-service)
 
+**Note:** This is an advanced feature that will likely require custom packaging of the peer image. For example, the following samples use `jq` and `bash`, which are not included in the current official `fabric-peer` image.
+
 ## Packaging chaincode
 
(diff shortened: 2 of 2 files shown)


sykesm (changes requested):
These are examples; they are based on what we use in our own integration tests. The fact that they don't run in peer image that is published isn't relevant. This is especially true given we fully expect users of external builders to add a number of additional dependencies to their images.

Besides the fact that they don't work in Alpine, what's motivating this change?

sykesm:
Elaborating a bit further, in addition to `bash`, the peer images don't contain `jq` or `go` either yet the scripts use them. The goal is to show an example of the function of how data is presented to the external launcher programs and how those external programs *may* operate to generate the expected outputs.

I'll also say that the choice of `/bin/bash` is explicit given we frequently use bashisms in scripts. This includes scripts we run in CI, the `bootstrap.sh` script, and the vast majority of scripts in our samples and examples.

satota2:
@sykesm
Thank you for your comments.
 
The main motivation is to prevent developers using the original fabric peer image from being unable to detecting problems, because the current error messages doesn't seem to give information to distinguish between bash shebang not running scripts and other issues.
 
> the peer images don't contain jq or go either yet the scripts use them.
 
That is as you pointed out.
Based on your comments, I would like to add information to specify the prerequisites when using the peer image to the documents instead of replacing scripts.
Additionally, I would like to propose to improve log messages when scripts can not be executed on the external chaincode builder programs as another patch.
How about these ideas?

sykesm:
> The main motivation is to prevent developers using the original fabric peer image from being unable to detecting problems, because the current error messages doesn't seem to give information to distinguish between bash shebang not running scripts and other issues.

The "original fabric peer image" is not designed to support external builders and launchers. It is published as a convenience for simple deployments, samples, and test environments. It does not contain anything beyond the bare minimum requirements.

> Based on your comments, I would like to add information to specify the prerequisites when using the peer image to the documents instead of replacing scripts.

I really don't want to add documentation that implies in any way that the published peer image supports external builders and launchers. Instead of trying to imply it does, the doc section related to this feature should have a clear statement indicating that this is an advanced feature that will likely require custom packaging of Fabric.

Existing users are creating custom builds of the peer images that are extended with the tools necessary to support their external builders and launchers.

> Additionally, I would like to propose to improve log messages when scripts can not be executed on the external chaincode builder programs as another patch.

Messages can always be improved and I'd welcome work to help on that front provided the changes do not assume the peer or the external programs are running in the context of a container.

satota2:
@sykesm
Thank you for your comments.

I understand that original fabric peer image is not designed to support this feature and it is NOT good to add documentations that implies in any way that the published peer image supports the feature.

> Instead of trying to imply it does, the doc section related to this feature should have a clear statement indicating that this is an advanced feature that will likely require custom packaging of Fabric.

Based on the above comment, I updated the patch to add clear statements to the documents related to this feature.

> Messages can always be improved and I'd welcome work to help on that front provided the changes do not assume the peer or the external programs are running in the context of a container.

Also, I'm going to propose another patch to improve log messages.

sykesm (approved):
Thanks!

PR #1223, opened 2020-05-07
Validate TLS certs during raft consenter addition

wlahti:
#### Type of change

- New feature
- Test update

#### Description

- Validate the client/server TLS certs against the orderer organizations while adding a raft consenter.
- Also some test cleanup/reformatting in the first commit. 

#### Related issues

[FAB-17733](https://jira.hyperledger.org/browse/FAB-17733)


orderer/consensus/etcdraft/membership.go
```diff
@@ -105,6 +114,102 @@ func ComputeMembershipChanges(oldMetadata *etcdraft.BlockMetadata, oldConsenters
...
+		return errors.Wrap(err, "parsing tls server cert")
+	}
+
+	for _, org := range ordererConfig.Organizations() {
+		opts, err := createX509VerifyOptions(org.MSP())
+		if err != nil {
+			return errors.WithMessage(err, "creating x509 verify options")
+		}
+
+		// following the lead of msp/mspimplvalidate.go:
```
yacovm: This is for TLS, so we actually do care about the time. If the CA are expired or too early, the TLS handshake won't work.

wlahti: Ah, I see. Makes sense. I'll fix that. Thanks!

orderer/consensus/etcdraft/membership.go
```diff
@@ -105,6 +114,102 @@ func ComputeMembershipChanges(oldMetadata *etcdraft.BlockMetadata, oldConsenters
...
+	}
+	tlsIntermediates := x509.NewCertPool()
+	for _, cert := range intermediateCerts {
+		tlsIntermediates.AddCert(cert)
+	}
+
+	return x509.VerifyOptions{
+		Roots:         tlsRoots,
+		Intermediates: tlsIntermediates,
+		KeyUsages: []x509.ExtKeyUsage{
```
yacovm, line 167: what if we only have one of the two key usages? won't this fail?

orderer/consensus/etcdraft/membership.go
```diff
@@ -105,6 +114,102 @@ func ComputeMembershipChanges(oldMetadata *etcdraft.BlockMetadata, oldConsenters
...
+	clientCert, err := parseCertificateFromBytes(c.ClientTlsCert)
+	if err != nil {
+		return errors.Wrap(err, "parsing tls client cert")
+	}
+	serverCert, err := parseCertificateFromBytes(c.ServerTlsCert)
+	if err != nil {
+		return errors.Wrap(err, "parsing tls server cert")
+	}
+
+	for _, org := range ordererConfig.Organizations() {
```
yacovm: why do you need to loop around the organizations? Just construct the TLS CA cert pools from the union of all the CA certs over all MSPs, and give it a go. The x509 library should be able to find a validation path if such exists.

integration/raft/config_test.go
```diff
@@ -391,20 +384,41 @@ var _ = Describe("EndToEnd reconfiguration and onboarding", func() {
...
+				ServerTlsCert: thirdOrdererCertificate,
+				ClientTlsCert: thirdOrdererCertificate,
+				Host:          "127.0.0.1",
+				Port:          uint32(network.OrdererPort(orderer3, nwo.ClusterPort)),
+			})
 
-			By("Ensuring orderer3 detects leader loss")
-			leaderLoss := fmt.Sprintf("Raft leader changed: %d -> 0", leader)
-			Eventually(ordererRunners[2].Err(), network.EventuallyTimeout, time.Second).Should(gbytes.Say(leaderLoss))
+			By("removing orderer3's TLS root CA certificate")
```
yacovm: It's too early to remove that consenter. You just added it and you are immediately removing it from the config. 

We need to wait for the consenter to onboard, and make sure it participates in consensus and only then we need to remove it. 
Otherwise it might be a false positive test execution.

integration/raft/config_test.go
```diff
@@ -391,20 +384,41 @@ var _ = Describe("EndToEnd reconfiguration and onboarding", func() {
...
+			o3Runner := network.OrdererRunner(orderer3)
+			ordererRunners[2] = o3Runner
+			o3Proc = ifrit.Invoke(o3Runner)
+			Eventually(o3Proc.Ready(), network.EventuallyTimeout).Should(BeClosed())
+			ordererProcesses[2] = o3Proc
+
+			By("ensuring TLS handshakes fail with the other orderers")
+			for i, oRunner := range ordererRunners {
+				if i < 2 {
+					Eventually(oRunner.Err(), network.EventuallyTimeout).Should(gbytes.Say("TLS handshake failed with error tls: client didn't provide a certificate"))
```
yacovm, line 420: Is it because the rest of the OSNs don't want to connect to complete the handshake once they see our server certificate?

orderer/consensus/etcdraft/membership.go
wlahti: I did this due to the assumption that we'd want the client and server TLS certs to be valid under the same organization's MSP. I can change this if that assumption is wrong.

yacovm (changes requested):
Made a quick pass, gave some comments.

I'm also missing the integration test that checks that you are trying to add a consenter with a mismatching certificate and it rejects you.

I know you did a unit test, but IMO we have too much super-glue between different components (etcdraft chain, channel configtx, etc.) so we'd want to be on the safe side. 

I guess you can just **addConsenter** with an invalid cert and catch the return error in the config test.

orderer/consensus/etcdraft/membership.go
yacovm: In the end of the day, the check should validate the current behavior of TLS validation - which actually aggregates all TLS root/intermediate CA certificates and also adds the ones in the local config.

orderer/consensus/etcdraft/membership.go
wlahti: No, this accepts certificates as long as they have any of the key usages specified.

orderer/consensus/etcdraft/membership.go
wlahti: Combined into a single set of VerifyOptions with all TLS root/intermediate CA certs.

orderer/consensus/etcdraft/membership.go
wlahti: Done.

integration/raft/config_test.go
wlahti: I've updated this by waiting for orderer3 to see the leader but let me know if there's anything else we can/should do here.

integration/raft/config_test.go
wlahti: Not sure on the exact mechanics here but I can research/investigate.

wlahti:
Thanks for the review and insight into Raft/TLS, Yacov! Addressed most of the comments now. Let me see what I can do for the negative case in the integration test.

orderer/consensus/etcdraft/membership.go
wlahti: Done.

orderer/consensus/etcdraft/membership.go
wlahti: Thinking a little more while I'm working on the negative integration test: do/should we verify the key usages separately, i.e. check that the ServerTlsCert has the server key usage and the ClientTlsCert has the client key usage?

yacovm: I don't think so, there is a limit to which we need to coddle the users.

yacovm (approved):

PR #1224, opened 2020-05-08
Clarify the deliver access denied message

jyellick:
#### Type of change

#### Description

There are two scenarios where a deliver client could receive a
'FORBIDDEN' result when requesting blocks.  Either the client was not
authorized to connect to the channel initially, or, the client's
access was revoked after a successful connection by some later
configuration block.  In both cases, we log an identical error message
that "Client authorization revoked" when in fact, for the first case,
the client may never have had access, so claiming it was revoked is
misleading.


jyellick:
Because of deficiencies in github UI, I cannot comment on line 325, but this is where the other, duplicate log message occurs which is emitted when access is actually revoked.

yacovm (approved):

PR #1229, opened 2020-05-10
[FAB-17869] Allow TLS CAs with overlapping issuers

yacovm:
The client root TLS CA certificate pool construction doesn't allow different issuers
with the same subject name to exist in the CA cert pool.

This prevents use cases where different organizations have issued CAs with the same subject name.

There is little sense to prevent this in the first place, as there can be only 2 possible cases:
I) The certificate chain validation of a TLS handshake allows this: In this case, Fabric should not
   artificially prevent supporting such scenarios.
II) The certificate chain validation of a TLS handshake disallows this: In this case, there is little
    gained in skipping pluralities in subject name during TLS CA cert pool construction, as the
    chain validation would prevent it anyway. 

This change set removes the artificial constraint, and adjust one of the integration tests
that was built to adjust to the said constraint.

Change-Id: Ie93c34997ad3e134a0a04b805fb007cd036d9683

### DIFF
--- a/integration/raft/config_test.go
+++ b/integration/raft/config_test.go
@@ -8,10 +8,6 @@ package raft
 
 import (
 	"bytes"
-	"crypto/ecdsa"
-	"crypto/rand"
-	"crypto/x509"
-	"encoding/pem"
 	"fmt"
 	"io/ioutil"
 	"os"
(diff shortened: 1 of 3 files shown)


kopaygorodsky:
Should user validate his input and check for duplicates in certs list etc?

C0rWin (approved):

PR #1230, opened 2020-05-10
[FAB-17864] DropApplicationDBs should return error when createCouchInstance fails

wenjianqiao:

#### Type of change
- Bug fix

#### Description
Fix DropApplicationDBs to return error when getting an error from
createCouchInstance. Also set CouchDB config to limit the number
of retries.

#### Related issues
https://jira.hyperledger.org/browse/FAB-17864

### DIFF
--- a/core/ledger/kvledger/txmgmt/statedb/statecouchdb/couchdbutil_test.go
+++ b/core/ledger/kvledger/txmgmt/statedb/statecouchdb/couchdbutil_test.go
@@ -213,3 +213,11 @@ func TestDropApplicationDBs(t *testing.T) {
 	assert.NoError(t, err, "Error when retrieving application db names")
 	assert.Equal(t, 0, len(dbs), "Databases should be dropped")
 }
+
+func TestDropApplicationDBsWhenDBNotStarted(t *testing.T) {
+	config := testConfig()
+	config.MaxRetriesOnStartup = 1
+	config.Address = "127.0.0.1:5984"
+	err := DropApplicationDBs(config)
+	assert.EqualError(t, err, `unable to connect to CouchDB, check the hostname and port: http error calling couchdb: Get "http://127.0.0.1:5984/": dial tcp 127.0.0.1:5984: connect: connection refused`)
+}
--- a/core/ledger/kvledger/txmgmt/statedb/statecouchdb/couchdbutil.go
+++ b/core/ledger/kvledger/txmgmt/statedb/statecouchdb/couchdbutil.go
@@ -304,7 +304,7 @@ func DropApplicationDBs(config *ledger.CouchDBConfig) error {
 	logger.Info("Dropping CouchDB application databases ...")
 	couchInstance, err := createCouchInstance(config, &disabled.Provider{})
 	if err != nil {
-		return nil
+		return err
 	}
 	dbNames, err := couchInstance.retrieveApplicationDBNames()
 	if err != nil {


core/ledger/kvledger/txmgmt/statedb/statecouchdb/couchdbutil_test.go
```diff
@@ -213,3 +213,11 @@ func TestDropApplicationDBs(t *testing.T) {
 	assert.NoError(t, err, "Error when retrieving application db names")
 	assert.Equal(t, 0, len(dbs), "Databases should be dropped")
 }
+
+func TestDropApplicationDBsWhenDBNotStarted(t *testing.T) {
+	config := testConfig()
+	config.MaxRetriesOnStartup = 1
+	config.Address = "127.0.0.1:5984"
+	err := DropApplicationDBs(config)
+	assert.Error(t, err, `unable to connect to CouchDB, check the hostname and port: http error calling couchdb: Get "http://127.0.0.1:5984/": dial tcp 127.0.0.1:5984: connect: connection refused`)
```
cendhu:
```suggestion
	assert.EqualError(t, err, `unable to connect to CouchDB, check the hostname and port: http error calling couchdb: Get "http://127.0.0.1:5984/": dial tcp 127.0.0.1:5984: connect: connection refused`)
```

core/ledger/kvledger/upgrade_dbs.go
```diff
@@ -44,7 +44,10 @@ func UpgradeDBs(config *ledger.Config) error {
 	}
 
 	if config.StateDBConfig.StateDatabase == "CouchDB" {
-		if err = statecouchdb.DropApplicationDBs(config.StateDBConfig.CouchDB); err != nil {
+		couchdbConfig := config.StateDBConfig.CouchDB
+		couchdbConfig.MaxRetries = 2
+		couchdbConfig.MaxRetriesOnStartup = 2
```
cendhu: Similar to `peer node upgrade-dbs`, the `peer node start` also expects the CouchDB to be up. However, we don't manipulate/change the user configured `MaxRetries` and `MaxRetriesOnStartup` when we execute `peer node start`. I am not comfortable in manipulating the user-configured values only for the  `peer node upgrade-dbs`. It looks like a hack to me.

If the rationale is the following, I am still not comfortable: in v2.0, we allow CouchDB not to be up while executing `upgrade-dbs` but in v2.2, we need the CouchDB to be up as the peer itself drops CouchDB databases. As admin could have ignored noticing this change and might run this command without keeping CouchDB alive. Hence, we need to return from the `upgrade-dbs` earlier. 

The reason is that the `upgrade-dbs` command would be executed only once when the move is made from v1.4.x to v2.x. It is clear that this command would not be used frequently. If an admin makes a mistake for one-time execution, I don't see a strong reason to exit early by manipulating/changing the user-configured `MaxRetries` and `MaxRetriesOnStartup`.

cendhu (changes requested):

core/ledger/kvledger/upgrade_dbs.go
wenjianqiao: @cendhu thank you. This was what we discussed at our scrum. Let's talk again in today's scrum. I am fine to use the user-configured values instead of changing them.

wenjianqiao: Discussed at scrum and agreed to use the user-configured values instead of changing them in individual commands.

cendhu (approved):
Thanks, Wenjian, for the changes.

manish-sethi (approved):

PR #1232, opened 2020-05-11
FAB-17839 Ch.Part.API: List channels REST handler

tock-ibm:
Develop the following handler functionality.

List all channels:
`GET /participation/v1/channels`.

List a single channel:
`GET /participation/v1/channels/my-channel`.

Define the interface the Registrar will implement,
place skeleton methods in it.

Change-Id: I4ce65b406ccbbbe8e1dbd4a95680c3561f5fc7c9

#### Type of change

- New feature

#### Related issues
Story: https://jira.hyperledger.org/browse/FAB-15710
Epic: https://jira.hyperledger.org/browse/FAB-17712

### DIFF
--- a/orderer/common/channelparticipation/responses.go
+++ b/orderer/common/channelparticipation/responses.go
@@ -1,0 +1,34 @@
+/*
+Copyright IBM Corp. All Rights Reserved.
+
+SPDX-License-Identifier: Apache-2.0
+*/
+
+package channelparticipation
+
+type ErrorResponse struct {
+	Error string `json:"error"`
+}
+
+// ListAllChannels carries the response to a List of all channels.
+type ListAllChannels struct {
+	Size          int                `json:"size"`          // The number of channels
+	SystemChannel string             `json:"systemChannel"` // The system channel name, empty if doesn't exist
+	Channels      []ChannelInfoShort `json:"channels"`      // All channels
+}
+
+// The short info of a single channel.
+type ChannelInfoShort struct {
+	Name string `json:"name"` // The channel name
+	URL  string `json:"url"`  // The channel relative URL (no Host:Port)
+}
+
+// ChannelInfoFull carries the response to a List of a single channel.
+type ChannelInfoFull struct {
+	Name            string `json:"name"`            // The channel name
+	URL             string `json:"url"`             // The channel relative URL (no Host:Port)
+	ClusterRelation string `json:"clusterRelation"` // Whether it is a “member” or ”follower”, case insensitive
+	Status          string `json:"status"`          // ”onboarding” or ”active”, case insensitive
+	Height          uint64 `json:"height"`          // Current block height
+	BlockHash       []byte `json:"blockHash"`       // Last block hash
+}
--- a/orderer/common/channelparticipation/restapi.go
+++ b/orderer/common/channelparticipation/restapi.go
@@ -58,9 +58,11 @@ func NewHTTPHandler(config localconfig.ChannelParticipation, registrar ChannelMa
 		router:    mux.NewRouter(),
 	}
 
-	handler.router.HandleFunc(urlWithChannelIDKey, handler.serveListOne).Methods("GET")
-	handler.router.HandleFunc(urlWithChannelIDKey, handler.serveJoin).Methods("POST")
-	handler.router.HandleFunc(urlWithChannelIDKey, handler.serveRemove).Methods("DELETE")
+	handler.router.HandleFunc(urlWithChannelIDKey, handler.serveListOne).Methods(http.MethodGet)
+	handler.router.HandleFunc(urlWithChannelIDKey, handler.serveJoin).Methods(http.MethodPost).Headers(
+		"Content-Type", "application/json") // TODO support application/octet-stream & multipart/form-data
+	handler.router.HandleFunc(urlWithChannelIDKey, handler.serveBadContentType).Methods(http.MethodPost)
+	handler.router.HandleFunc(urlWithChannelIDKey, handler.serveRemove).Methods(http.MethodDelete)
 	handler.router.HandleFunc(urlWithChannelIDKey, handler.serveNotAllowed)
 
 	handler.router.HandleFunc(URLBaseV1Channels, handler.serveListAll).Methods("GET")
@@ -145,6 +147,11 @@ func (h *HTTPHandler) serveRemove(resp http.ResponseWriter, req *http.Request) {
 	//TODO
 	err = errors.Errorf("not implemented yet: %s %s", req.Method, req.URL.String())
 	h.sendResponseJsonError(resp, http.StatusNotImplemented, err)
+}
+
+func (h *HTTPHandler) serveBadContentType(resp http.ResponseWriter, req *http.Request) {
+	err := errors.Errorf("unsupported Content-Type: %s", req.Header.Values("Content-Type"))
+	h.sendResponseJsonError(resp, http.StatusBadRequest, err)
 }
 
 func (h *HTTPHandler) serveNotAllowed(resp http.ResponseWriter, req *http.Request) {
--- a/orderer/common/channelparticipation/restapi.go
+++ b/orderer/common/channelparticipation/restapi.go
@@ -9,21 +9,28 @@ package channelparticipation
 import (
 	"encoding/json"
 	"net/http"
+	"strings"
 
+	"github.com/hyperledger/fabric/common/configtx"
 	"github.com/hyperledger/fabric/common/flogging"
 	"github.com/hyperledger/fabric/orderer/common/localconfig"
 	"github.com/pkg/errors"
 )
 
 const URLBaseV1 = "/participation/v1/"
-
-type ErrorResponse struct {
-	Error string `json:"error"`
-}
+const URLBaseV1Channels = URLBaseV1 + "channels"
 
 //go:generate counterfeiter -o mocks/channel_management.go -fake-name ChannelManagement . ChannelManagement
 
 type ChannelManagement interface {
+	// ListAll returns the names of all channels (including the system channel), and the name of the system channel
+	// (empty if does not exist). The URL field is empty, and is to be completed by the caller.
+	ListAllChannels() ([]ChannelInfoShort, string)
+
+	// ListChannel provides extended status information about a channel.
+	// The URL field is empty, and is to be completed by the caller.
+	ListChannel(channelID string) (*ChannelInfoFull, error)
+
 	// TODO skeleton
 }
 
@@ -51,7 +58,43 @@ func (h *HTTPHandler) ServeHTTP(resp http.ResponseWriter, req *http.Request) {
 	}
 
 	switch req.Method {
-	case http.MethodGet, http.MethodPost, http.MethodDelete:
+	case http.MethodGet:
+		_, err := negotiateContentType(req) // Only application/json for now
+		if err != nil {
+			h.sendResponseJsonError(resp, http.StatusNotAcceptable, err)
+			return
+		}
+
+		channelID, all, err := detectResource(req.URL.Path)
+		if err != nil {
+			h.sendResponseJsonError(resp, http.StatusBadRequest, err)
+		}
+
+		// Asking for a list of all channels
+		if all {
+			var channels *ListAllChannels = &ListAllChannels{}
+			channels.Channels, channels.SystemChannel = h.registrar.ListAllChannels()
+			channels.Size = len(channels.Channels)
+			for i, info := range channels.Channels {
+				channels.Channels[i].URL = URLBaseV1Channels + `/` + info.Name
+			}
+			h.sendResponseOK(resp, channels)
+			return
+		}
+
+		// Asking for a single channel
+		if err = configtx.ValidateChannelID(channelID); err != nil {
+			h.sendResponseJsonError(resp, http.StatusBadRequest, errors.Wrap(err, "invalid channel ID"))
+			return
+		}
+		infoFull, err := h.registrar.ListChannel(channelID)
+		if err != nil {
+			h.sendResponseJsonError(resp, http.StatusNotFound, err)
+			return
+		}
+		h.sendResponseOK(resp, infoFull)
+
+	case http.MethodPost, http.MethodDelete:
 		//TODO skeleton
 		err := errors.Errorf("not implemented yet: %s %s", req.Method, req.URL.String())
 		h.sendResponseJsonError(resp, http.StatusNotImplemented, err)
@@ -62,11 +105,53 @@ func (h *HTTPHandler) ServeHTTP(resp http.ResponseWriter, req *http.Request) {
 	}
 }
 
+func negotiateContentType(req *http.Request) (string, error) {
+	acceptReq := req.Header.Get("Accept")
+	if len(acceptReq) == 0 {
+		return "application/json", nil
+	}
+
+	options := strings.Split(acceptReq, ",")
+	for _, opt := range options {
+		if strings.Contains(opt, "application/json") ||
+			strings.Contains(opt, "application/*") ||
+			strings.Contains(opt, "*/*") {
+			return "application/json", nil
+		}
+	}
+
+	return "", errors.New("response Content-Type is application/json only")
+}
+
+func detectResource(path string) (string, bool, error) {
+	nakedPath := strings.TrimSuffix(path, "/")
+
+	if !strings.HasPrefix(path, URLBaseV1Channels) {
+		return "", false, errors.Errorf("invalid path: %s", path)
+	}
+	channel := strings.TrimPrefix(nakedPath, URLBaseV1Channels)
+
+	if len(channel) == 0 {
+		return "", true, nil
+	}
+
+	return channel[1:], false, nil
+}
+
 func (h *HTTPHandler) sendResponseJsonError(resp http.ResponseWriter, code int, err error) {
 	encoder := json.NewEncoder(resp)
 	resp.WriteHeader(code)
 	resp.Header().Set("Content-Type", "application/json")
 	if err := encoder.Encode(&ErrorResponse{Error: err.Error()}); err != nil {
-		h.logger.Errorw("failed to encode payload", "error", err)
+		h.logger.Errorf("failed to encode error, err: %s", err)
+	}
+}
+
+func (h *HTTPHandler) sendResponseOK(resp http.ResponseWriter, content interface{}) {
+	encoder := json.NewEncoder(resp)
+	resp.WriteHeader(http.StatusOK)
+	resp.Header().Set("Content-Type", "application/json")
+	if err := encoder.Encode(content); err != nil {
+		h.logger.Errorf("failed to encode content, err: %s", err)
 	}
 }
--- a/orderer/common/channelparticipation/restapi.go
+++ b/orderer/common/channelparticipation/restapi.go
@@ -8,19 +8,24 @@ package channelparticipation
 
 import (
 	"encoding/json"
-	"github.com/hyperledger/fabric/orderer/common/types"
 	"net/http"
 	"path"
 	"strings"
 
+	"github.com/gorilla/mux"
 	"github.com/hyperledger/fabric/common/configtx"
 	"github.com/hyperledger/fabric/common/flogging"
 	"github.com/hyperledger/fabric/orderer/common/localconfig"
+	"github.com/hyperledger/fabric/orderer/common/types"
 	"github.com/pkg/errors"
 )
 
-const URLBaseV1 = "/participation/v1/"
-const URLBaseV1Channels = URLBaseV1 + "channels"
+const (
+	URLBaseV1           = "/participation/v1/"
+	URLBaseV1Channels   = URLBaseV1 + "channels"
+	channelIDKey        = "channelID"
+	urlWithChannelIDKey = URLBaseV1Channels + "/{" + channelIDKey + "}"
+)
 
 //go:generate counterfeiter -o mocks/channel_management.go -fake-name ChannelManagement . ChannelManagement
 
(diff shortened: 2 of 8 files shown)


orderer/common/channelparticipation/responses.go
```diff
@@ -0,0 +1,34 @@
...
+// The short info of a single channel.
+type ChannelInfoShort struct {
+	Name string `json:"name"` // The channel name
+	URL  string `json:"url"`  // The channel relative URL (no Host:Port)
+}
+
+// ChannelInfoFull carries the response to a List of a single channel.
+type ChannelInfoFull struct {
+	Name            string `json:"name"`            // The channel name
+	URL             string `json:"url"`             // The channel relative URL (no Host:Port)
```
yacovm, line 29: what's a URL? can you give an example?

orderer/common/multichannel/registrar.go
```diff
@@ -370,3 +371,13 @@ func (r *Registrar) NewChannelConfig(envConfigUpdate *cb.Envelope) (channelconfi
 func (r *Registrar) CreateBundle(channelID string, config *cb.Config) (channelconfig.Resources, error) {
 	return channelconfig.NewBundle(channelID, config, r.bccsp)
 }
+
+func (r *Registrar) ListAllChannels() ([]channelparticipation.ChannelInfoShort, string) {
```
yacovm: I'm not sure the registrar should return a slice of this type. This creates an unnecessary dependency on this package. Perhaps define a type in a leaf package and then have the registrar return a slice of that type?

orderer/common/multichannel/registrar.go
```diff
@@ -370,3 +371,13 @@ func (r *Registrar) NewChannelConfig(envConfigUpdate *cb.Envelope) (channelconfi
...
+	//TODO
+	return make([]channelparticipation.ChannelInfoShort,0), ""
+}
+
+func (r *Registrar) ListChannel(channelID string) (*channelparticipation.ChannelInfoFull, error) {
```
yacovm: Perhaps indicate the system channel string into the struct returned from this API as a boolean?

orderer/common/multichannel/registrar.go
tock-ibm: I was thinking the same way and originally had the `ListAllChannels` return  `([]string, string)`.... But then, when I crafted the 

`ListChannel(channelID string) (*channelparticipation.ChannelInfoFull, error)` 

I realized these two packages are going to be linked anyway... so I gave up on that direction. 
The way it is now we don't have to copy strings from one struct to another...

Putting these types in a leaf package to break the cycle will help.

orderer/common/multichannel/registrar.go
tock-ibm: We want to quickly identify whether we have one or not w/o scanning the entire list... this goes to the client so we want this efficient and clear. Note that the system channel may change from "" to "i-am-here" and back to "" as we can create and delete it with this API.

yacovm: Under what circumstance do we need to identify if we have a system channel or not, and not also anyway return all channels? 

The way this is used in the existing code in this PR is that you return all the channels anyway, so you are going to see who is the system channel when you iterate over it anyway, no?

orderer/common/channelparticipation/restapi.go
```diff
@@ -51,7 +58,43 @@ func (h *HTTPHandler) ServeHTTP(resp http.ResponseWriter, req *http.Request) {
...
 	switch req.Method {
-	case http.MethodGet, http.MethodPost, http.MethodDelete:
+	case http.MethodGet:
+		_, err := negotiateContentType(req) // Only application/json for now
+		if err != nil {
+			h.sendResponseJsonError(resp, http.StatusNotAcceptable, err)
+			return
+		}
+
+		channelID, all, err := detectResource(req.URL.Path)
```
C0rWin: this is not the way you should implement this logic, instead of manually extracting parameters you should template REST endpoint, for example:

```
  var channelsEndpoint = /participation/v1/channels/{channelID}
```

then you can use

```
   vars := mux.Vars(req)
   channelID := vars["channelID"]
```

orderer/common/channelparticipation/responses.go
```diff
@@ -0,0 +1,34 @@
...
+
+package channelparticipation
+
+type ErrorResponse struct {
+	Error string `json:"error"`
+}
+
+// ListAllChannels carries the response to a List of all channels.
+type ListAllChannels struct {
+	Size          int                `json:"size"`          // The number of channels
```
C0rWin: Why do you need this? Isn't it just length of the `Channels` slice?

orderer/common/channelparticipation/restapi.go
```diff
@@ -9,21 +9,28 @@ package channelparticipation
...
-	Error string `json:"error"`
-}
+const URLBaseV1Channels = URLBaseV1 + "channels"
 
 //go:generate counterfeiter -o mocks/channel_management.go -fake-name ChannelManagement . ChannelManagement
 
 type ChannelManagement interface {
+	// ListAll returns the names of all channels (including the system channel), and the name of the system channel
+	// (empty if does not exist). The URL field is empty, and is to be completed by the caller.
+	ListAllChannels() ([]ChannelInfoShort, string)
```
C0rWin: IMO, instead of returning a touple it's better to wrap it into struct.

orderer/common/channelparticipation/restapi.go
```diff
@@ -51,7 +58,43 @@ func (h *HTTPHandler) ServeHTTP(resp http.ResponseWriter, req *http.Request) {
...
+			h.sendResponseJsonError(resp, http.StatusBadRequest, err)
+		}
+
+		// Asking for a list of all channels
+		if all {
+			var channels *ListAllChannels = &ListAllChannels{}
+			channels.Channels, channels.SystemChannel = h.registrar.ListAllChannels()
+			channels.Size = len(channels.Channels)
+			for i, info := range channels.Channels {
+				channels.Channels[i].URL = URLBaseV1Channels + `/` + info.Name
```
C0rWin:
```
path.Join(URLBaseV1Channels, info.Name)
```

orderer/common/channelparticipation/restapi.go
C0rWin: Also I think that better name just to call it `Channels`

orderer/common/channelparticipation/restapi.go
```diff
@@ -9,21 +9,28 @@ package channelparticipation
...
 //go:generate counterfeiter -o mocks/channel_management.go -fake-name ChannelManagement . ChannelManagement
 
 type ChannelManagement interface {
+	// ListAll returns the names of all channels (including the system channel), and the name of the system channel
+	// (empty if does not exist). The URL field is empty, and is to be completed by the caller.
+	ListAllChannels() ([]ChannelInfoShort, string)
+
+	// ListChannel provides extended status information about a channel.
+	// The URL field is empty, and is to be completed by the caller.
+	ListChannel(channelID string) (*ChannelInfoFull, error)
```
C0rWin: `ChannelInfo`?

orderer/common/channelparticipation/restapi.go
```diff
@@ -51,7 +58,43 @@ func (h *HTTPHandler) ServeHTTP(resp http.ResponseWriter, req *http.Request) {
...
+			h.sendResponseOK(resp, channels)
+			return
+		}
+
+		// Asking for a single channel
+		if err = configtx.ValidateChannelID(channelID); err != nil {
+			h.sendResponseJsonError(resp, http.StatusBadRequest, errors.Wrap(err, "invalid channel ID"))
+			return
+		}
+		infoFull, err := h.registrar.ListChannel(channelID)
```
C0rWin: I think there is a way to register different path patterns to distinguish the desired function call, that is to separate logic of list all channels versus retrieving specific channel information.

C0rWin (changes requested):

orderer/common/channelparticipation/restapi.go
C0rWin: just realized we are not using `gorilla/mux` package to route REST requests.

orderer/common/channelparticipation/responses.go
tock-ibm: It is, but it goes to the client via JSON, so maybe it is useful there?

orderer/common/channelparticipation/restapi.go
tock-ibm: True! We have github.com/gorilla already vendored in Fabric... I wan't aware of that. I'll use that.

orderer/common/channelparticipation/restapi.go
tock-ibm: True! using `github.com/gorilla/mux`, I'll use that.

orderer/common/channelparticipation/restapi.go
tock-ibm: OK

orderer/common/channelparticipation/restapi.go
tock-ibm: OK

orderer/common/channelparticipation/restapi.go
tock-ibm: OK

orderer/common/multichannel/registrar.go
tock-ibm: Yes, but this structure is intended mainly for the client, who receives JSON.
The client may want to identify the system channel immediately, for example, when it needs to know if it is at all possible to create channels or only list them. If the system channel exists, the API can only serve GET and DELETE of the system channel. This way, the client can find that out without iterating.

orderer/common/channelparticipation/restapi.go
```diff
@@ -50,23 +78,122 @@ func (h *HTTPHandler) ServeHTTP(resp http.ResponseWriter, req *http.Request) {
...
+	} else {
+		resp.Header().Set("Allow", "GET")
+	}
+	resp.Header().Set("Content-Type", "application/json")
+	if err := encoder.Encode(&types.ErrorResponse{Error: err.Error()}); err != nil {
+		h.logger.Errorf("failed to encode error, err: %s", err)
+	}
+}
+
+func negotiateContentType(req *http.Request) (string, error) {
```
C0rWin, line 176: instead what you can do is to define headers you would accept while initializing gorilla/mux router, e.g.

```
    router.Headers("Content-Type", "application/json")
```

C0rWin: Or there is an option to handle regexps as well


```
    router.HeadersRegexp("Content-Type", "application/(text|json)")
```

orderer/common/channelparticipation/restapi.go
```diff
@@ -50,23 +78,122 @@ func (h *HTTPHandler) ServeHTTP(resp http.ResponseWriter, req *http.Request) {
...
-	case http.MethodGet, http.MethodPost, http.MethodDelete:
-		//TODO skeleton
-		err := errors.Errorf("not implemented yet: %s %s", req.Method, req.URL.String())
-		h.sendResponseJsonError(resp, http.StatusNotImplemented, err)
+	h.router.ServeHTTP(resp, req)
+}
+
+// List all channels
+func (h *HTTPHandler) serveListAll(resp http.ResponseWriter, req *http.Request) {
+	_, err := negotiateContentType(req) // Only application/json for now
```
C0rWin, line 89: routers allows to define matching headers hence you can just define content type there once avoiding repeating this functionality.

orderer/common/channelparticipation/restapi.go
```diff
@@ -50,23 +78,122 @@ func (h *HTTPHandler) ServeHTTP(resp http.ResponseWriter, req *http.Request) {
...
+		h.sendResponseJsonError(resp, http.StatusNotAcceptable, err)
+		return
+	}
 
-	default:
-		err := errors.Errorf("invalid request method: %s", req.Method)
-		h.sendResponseJsonError(resp, http.StatusBadRequest, err)
+	channelID, ok := mux.Vars(req)[channelIDKey]
+	if !ok {
+		h.sendResponseJsonError(resp, http.StatusInternalServerError, errors.Wrap(err, "missing channel ID"))
```
C0rWin, line 114: should be ` http.StatusBadRequest` cause `http.StatusInternalServerError` means fatal error, while here is just an empty or missing channelID

tock-ibm: Actually this has nothing to do with the request, it is an indication of bug - that the key used in the pattern and Vars map (const channelIDKey) is inconsistent in registration & retrieval. That would be a repeatable error internal to the server that the client cannot correct by changing the request.

orderer/common/channelparticipation/restapi.go
tock-ibm: Good observation. I will use it to ensure that the POST arrives with Content-type: application/json. (GET and DELETE have empty bodies, so there is no need to check.)

The Headers() matcher is good for filtering out requests that come in with bad Content-Type, for example, but here I am dealing with the Accept header. It needs special treatment that is not easy with regexp, and in particular it requires a response with http.StatusNotAcceptable and an error message in the body that describes what the server can deliver. In other cases it requires selecting the Mime type according to the q value of preferred types and so on. 

See also: https://github.com/gorilla/mux/issues/18

orderer/common/channelparticipation/restapi.go
tock-ibm: See above, content negotiation with Accept is not really supported by the mux Headers() matcher.

orderer/common/multichannel/registrar.go
tock-ibm: Moreover, this representation is shorter. 

```
{
  "size":3,
  "systemChannel":"a",
  "channels":[
     {"name":"a","url":"/api/channels/a"},
     {"name":"b","url":"/api/channels/b"},
     {"name":"c","url":"/api/channels/c"},
]}
```

is shorter than 


```
{
  "size":2,
  "channels":[
     {"name":"a","url":"/api/channels/a","systemChannel":"true"}, 
     {"name":"b","url":"/api/channels/b","systemChannel":"false"},
     {"name":"c","url":"/api/channels/c","systemChannel":"false"},
]}
```
even for a short list.

tock-ibm: The other alternative I see is:

```
{
  "systemChannel":{"name":"a","url":"/api/channels/a"},
  "channels":[
     {"name":"b","url":"/api/channels/b"},
     {"name":"c","url":"/api/channels/c"},
  "size":2,
]}
```
i.e. that the `channels` array carry only application channels.

orderer/common/types/channelinfo.go
```diff
@@ -0,0 +1,38 @@
...
+
+// ChannelInfo carries the response to an HTTP request to List a single channel.
+// This is marshaled into the body of the HTTP response.
+type ChannelInfo struct {
+	Name            string `json:"name"`            // The channel name
+	URL             string `json:"url"`             // The channel relative URL (no Host:Port)
+	ClusterRelation string `json:"clusterRelation"` // Whether it is a “member” or ”follower”, case insensitive
+	Status          string `json:"status"`          // ”onboarding” or ”active”, case insensitive
+	Height          uint64 `json:"height"`          // Current block height
+	BlockHash       []byte `json:"blockHash"`       // Last block hash
```
yacovm: What are clients expected to do with the block hash?

orderer/common/types/channelinfo.go
```diff
@@ -0,0 +1,38 @@
...
+// ErrorResponse carries the error response an HTTP request.
+// This is marshaled into the body of the HTTP response.
+type ErrorResponse struct {
+	Error string `json:"error"`
+}
+
+// ChannelList carries the response to an HTTP request to List all the channels.
+// This is marshaled into the body of the HTTP response.
+type ChannelList struct {
+	Size          int                `json:"size"`          // The number of channels
```
yacovm: I think having a size here is redundant as you can count the number of channels

orderer/common/channelparticipation/restapi.go
```diff
@@ -9,21 +9,35 @@ package channelparticipation
...
+	channelIDKey        = "channelID"
+	urlWithChannelIDKey = URLBaseV1Channels + "/{" + channelIDKey + "}"
+)
 
 //go:generate counterfeiter -o mocks/channel_management.go -fake-name ChannelManagement . ChannelManagement
 
 type ChannelManagement interface {
+	// ListAll returns the names of all channels (including the system channel), and the name of the system channel
+	// (empty if does not exist). The URL field is empty, and is to be completed by the caller.
+	ListAllChannels() *types.ChannelList
```
yacovm: The more idiomatic name is just **Channels()**. 
I also don't understand why the return type is a pointer type. 
Under what circumstance do we return nil? Why not return a non pointer type?

orderer/common/channelparticipation/restapi.go
```diff
@@ -9,21 +9,35 @@ package channelparticipation
...
+
+	// ListChannel provides extended status information about a channel.
+	// The URL field is empty, and is to be completed by the caller.
+	ListChannel(channelID string) (*types.ChannelInfo, error)
```
yacovm: why not non pointer type?

orderer/common/multichannel/registrar.go
yacovm: I like the last one the most, modulo the size which is redundant.

orderer/common/channelparticipation/restapi.go
C0rWin: I do not understand how missing channel id in the request path is internal server error. Can you elaborate?

orderer/common/multichannel/registrar.go
C0rWin: There is another option to return a map:

```
{
  "systemChannel": "a",
  "channels": {
      "a": "/api/channels/a",
      "b": "/api/channels/b",
      "c": "/api/channels/c",
   }
}
```

orderer/common/types/channelinfo.go
C0rWin: For example match it with other OSNs response?

yacovm: What do you expect to find, a fork in the Blockchain? 

And besides, the intended users of this interface is local administrator, you only have access to your own nodes anyway, so if a malicious leader is going to fork you, it is going to fork all your organizations nodes now that it knows you can compare hashes.

yacovm: By the way let's encode this in hexa or base64 otherwise the terminal might puke funny characters if it's a byte slice.

tock-ibm: It is marshaled into a JSON object as base64 by default. 
But I do have to implement a Stringer to make it log friendly, that is correct.

orderer/common/types/channelinfo.go
tock-ibm: I agree it is redundant from a program point of view. What I was thinking is that it would be useful for a human user that looks at the JSON object in an HTTP response, say in a browser, or using curl, to immediately tell how many channels there are. There could be many, and counting the stuff in field `Channels` could be confusing.

If you strongly object I will get rid of it.

orderer/common/channelparticipation/restapi.go
tock-ibm: The pointer type is because the struct can be big, having all those channel names in there, so this way a copy is avoided.

The names corresponds to the API:

List *          -> ListAllChannels()
List x          -> ListChannel(x)
Join x         -> JoinChannel(x)
Remove x  -> RemoveChannel(x)

That is the plan, at least.

orderer/common/channelparticipation/restapi.go
tock-ibm: This is borderline, I agree. A non-pointer would work, since it is a "smallish" struct. However, i do return nil when the channel is not found.

orderer/common/channelparticipation/responses.go
tock-ibm: See comment to @yacovm below...

orderer/common/channelparticipation/restapi.go
yacovm: But the Channels slice inside the struct, is only initialized once, and its elements are not going to be copied even if you use a non pointer type. 

Look at this [go playground sample](https://play.golang.org/p/Cn_BPDlIJm2).
Copying a slice is cheap because you only copy 2 pointers and not the elements inside

orderer/common/channelparticipation/restapi.go
yacovm: if you return an error in this case then it doesn't matter if you return nil if it's not found, or a zero value

orderer/common/types/channelinfo.go
yacovm: No, you have a point.

orderer/common/types/channelinfo.go
yacovm: OK but I'm still struggling to understand why you need the hash. It only makes the output more verbose, and I doubt you can do anything useful with this hash. 

Generally, I am reluctant adding information or functionality that I am not sure how they are going to be used.

tock-ibm: Adding the block-hash was a review comment I got on the [design document](https://docs.google.com/document/d/1kYoicS_Z59gNvmuuB7VFZ6WsOL31C34dWMeOFQiIygA/edit?usp=sharing), so I added it. I don't mind taking it off.

orderer/common/channelparticipation/restapi.go
tock-ibm: You are correct. I'll change that.

orderer/common/channelparticipation/restapi.go
tock-ibm: non- pointer type it is, then.

orderer/common/multichannel/registrar.go
tock-ibm: OK Lets's go with 
```
{
  "systemChannel":{"name":"a","url":"/api/channels/a"},
  "channels":[
     {"name":"b","url":"/api/channels/b"},
     {"name":"c","url":"/api/channels/c"},
]}
```
I want to keep the ChannelInfoShort struct just in case we'll need to add another field one day.

orderer/common/channelparticipation/responses.go
tock-ibm: Out it goes.

yacovm (dismissed):

orderer/common/channelparticipation/restapi.go
```diff
@@ -9,21 +9,36 @@ package channelparticipation
...
+	urlWithChannelIDKey = URLBaseV1Channels + "/{" + channelIDKey + "}"
+)
 
 //go:generate counterfeiter -o mocks/channel_management.go -fake-name ChannelManagement . ChannelManagement
 
 type ChannelManagement interface {
+	// ListAllChannels returns a slice of ChannelInfoShort containing all application channels (excluding the system
+	// channel), and ChannelInfoShort of the system channel (nil if does not exist).
+	// The URL fields are empty, and are to be completed by the caller.
+	ListAllChannels() types.ChannelList
```
C0rWin: I thought we agreed to rename APIs here?

tock-ibm: Actually we didn't, we just agreed on the return types, see here:
https://github.com/hyperledger/fabric/pull/1232#discussion_r423875968

C0rWin: I recall you have confirmed at least twice:

https://github.com/hyperledger/fabric/pull/1232#discussion_r422969830
https://github.com/hyperledger/fabric/pull/1232#discussion_r422969632

tock-ibm: I change ChannelInfoLong to ChannelInfo, and used the tuple suggestion as well, but not the name of the method...

tock-ibm: done

yacovm (approved):

C0rWin (approved):

PR #1233, opened 2020-05-11
[FAB-17844] Copy symlinks as-is in external builder output

sstone1:
Currently, the external builder code does not check
for symlinks in build output when copying them. This
results in the resolved files being copied as files
instead of symlinks.

Node.js chaincode relies on certain symlinks that npm
creates in the node_modules tree. There may be other
chaincode scenarios that rely on symlinks.

This commit changes the external builder code so that
it tests for symlinks, and copies them as symlinks
instead of copying them as files into the destination
directory.

Signed-off-by: Simon Stone <sstone1@uk.ibm.com>

#### Type of change

- Bug fix
- New feature
- Test update
- Documentation update

### DIFF
--- a/core/container/externalbuilder/copy.go
+++ b/core/container/externalbuilder/copy.go
@@ -10,47 +10,15 @@ import (
 	"io"
 	"os"
 	"path/filepath"
+	"strings"
 
 	"github.com/hyperledger/fabric/common/flogging"
 	"github.com/pkg/errors"
 )
 
-// MoveOrCopyDir attempts to copy src to dest by firstly trying to move, then copy upon failure.
-func MoveOrCopyDir(logger *flogging.FabricLogger, srcroot, destroot string) error {
-	mvErr := os.Rename(srcroot, destroot)
-	if mvErr == nil {
-		return nil
-	}
-
-	logger.Debugf("Failed to move %s to %s: %s, try copy instead", srcroot, destroot, mvErr)
-
-	info, err := os.Stat(srcroot)
-	if err != nil {
-		return errors.WithMessagef(err, "failed to stat dir: %s", srcroot)
-	}
-
-	if err = os.MkdirAll(destroot, info.Mode()); err != nil {
-		return errors.WithMessagef(err, "failed to make dir: %s", destroot)
-	}
-
-	cpErr := CopyDir(srcroot, destroot)
-	if cpErr == nil {
-		return nil
-	}
-
-	logger.Errorf("Failed to copy %s to %s: %s", srcroot, destroot, cpErr)
-
-	rmErr := os.RemoveAll(destroot)
-	if rmErr != nil {
-		logger.Errorf("Failed to clean targeting dir %s: %s", destroot, rmErr)
-	}
-
-	return errors.WithMessagef(cpErr, "failed to copy %s to %s", srcroot, destroot)
-}
-
 // CopyDir creates a copy of a dir
-func CopyDir(srcroot, destroot string) error {
-	return filepath.Walk(srcroot, func(path string, info os.FileInfo, err error) error {
+func CopyDir(logger *flogging.FabricLogger, srcroot, destroot string) error {
+	err := filepath.Walk(srcroot, func(path string, info os.FileInfo, err error) error {
 		if err != nil {
 			return err
 		}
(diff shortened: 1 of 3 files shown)


core/container/externalbuilder/copy.go
```diff
@@ -78,6 +88,14 @@ func CopyDir(srcroot, destroot string) error {
 	})
 }
 
+func copySymlink(srcpath, destpath string) error {
+	symlinkDest, err := os.Readlink(srcpath)
+	if err != nil {
+		return err
+	}
+	return os.Symlink(symlinkDest, destpath)
```
sykesm, line 96: The links must not point outside of the tree and must be relative. This does not appear to handle that.

sykesm (changes requested):
Please consider the security implications of links outside of the tree that we have created for source and build. While the current copy is incorrect, when we add explicit symlink handling, we should do the right thing.

sykesm:
I received this question via chat so I thought I'd move it over here.

>  If we want to reject absolute symlinks, and relative symlinks that point outside of the tree - we should do this for the simple moving the directory case as well right?

Yes, we should. In fact, I think we should remove the directory move path entirely. When the copy path was added to handle moving data across filesystems, the move path was left as an optimization. As we start policing symlinks, we'll have to walk all of the files anyway to make modifications and this is unnecessary complexity. I'd much rather have one correct path than maintain an optimization, especially for something that does not happen very frequently.

I do want to be careful, however, that we're not going too far down the rabbit hole. External builders and launchers are considered "trusted" extensions of the peer. They do not execute in a sandbox and we do little to nothing to prevent implementations from shooting themselves in the foot. The reason why the move path also needs symlink management is because, if fully qualified links are used, they can point to the location of the tree _prior_ to the move. This clearly won't work.

So, in summary:

- The optimized move path should be removed
- Symlink validation should ensure that the target of the source symlink lives within the root of the input tree
- Symlink management should ensure that the target of the output link is a relative path that resolves to a file within the output tree
- The file walk should only support regular files, directories, and symlinks with relative targets that resolve within the tree; any other file type (special, fifo, devices, character, whatnot) should result in an error

Make sense?

sstone1:
That's great, thanks @sykesm!

sstone1:
@sykesm I believe I've made the requested changes:

- Removed the fast path "move directory" option
- Now returning errors on files that are not regular files, directories, or symlinks
- Now returning errors on absolute symlinks and symlinks that point outside of the tree

Please can you re-review?

core/container/externalbuilder/copy_test.go
```diff
@@ -19,23 +20,35 @@ import (
...
+		logger                  *flogging.FabricLogger
+		srcRootDir, destRootDir string
+		err                     error
 	)
 
 	BeforeEach(func() {
 		srcRootDir, err = ioutil.TempDir("", "copy-test-")
 		Expect(err).NotTo(HaveOccurred())
-		srcSubDir, err = ioutil.TempDir(srcRootDir, "sub-")
+
```
sykesm, line 34: Looks like you've copied a pattern from the original code of making all file permissions 0777.

I'd prefer something more sane like 0644 for regular files (no execute and only writable by owner) and 0755 for directories. Also, the use of `os.ModePerm` hides the actual 0777 value via indirection and does not use the constant as intended - a mask to retrieve the permission bits.

core/container/externalbuilder/copy_test.go
```diff
@@ -45,51 +58,77 @@ var _ = Describe("copy", func() {
...
-		BeforeEach(func() {
-			destRootDir, err = ioutil.TempDir("", "dest-")
-			Expect(err).NotTo(HaveOccurred())
-			err = os.RemoveAll(destRootDir)
+	It("Copies all files and subdirectories", func() {
+		err = CopyDir(logger, srcRootDir, destRootDir)
+		Expect(err).NotTo(HaveOccurred())
+
+		_, err := os.Stat(destRootDir)
+		Expect(err).NotTo(HaveOccurred())
```
sykesm: What is this attempting to accomplish? If it's trying to assert the directory exists, `Expect(destRootDir).To(BeADirectory())` would be much clearer.

core/container/externalbuilder/copy_test.go
```diff
@@ -45,51 +58,77 @@ var _ = Describe("copy", func() {
...
+		_, err = os.Stat(filepath.Join(destRootDir, "file-in-root.txt"))
+		Expect(err).NotTo(HaveOccurred())
```
sykesm: Is this trying to ensure the file exists? `Expect(...).To(BeARegularFile())`? I think most of the places you're doing the `os.Stat` to get a side-effect of an error should be replaced with a [matcher](https://onsi.github.io/gomega/#working-with-files).

I'd also expect some assertions on file contents to ensure we didn't just do the `os.Create` and forget the copy. (Because it sure looks like I can comment out the `io.Copy` and all of the tests will still pass...

core/container/externalbuilder/copy_test.go
```diff
@@ -19,23 +20,35 @@ import (
 
 var _ = Describe("copy", func() {
 	var (
-		logger                             *flogging.FabricLogger
-		srcRootDir, srcSubDir, destRootDir string
-		srcRootFile, srcSubFile            *os.File
-		err                                error
+		logger                  *flogging.FabricLogger
+		srcRootDir, destRootDir string
+		err                     error
```
sykesm: Oh, please no. Don't declare an `err` var at the top level describe or at any scope that gets shared.

You didn't do this but since there will be changes, please address this as well.

core/container/externalbuilder/copy_test.go
```diff
@@ -45,51 +58,77 @@ var _ = Describe("copy", func() {
...
-			Expect(err).NotTo(HaveOccurred())
-			_, err = os.Stat(filepath.Join(destRootDir, filepath.Base(srcSubDir)))
-			Expect(err).NotTo(HaveOccurred())
-			f, err := os.Stat(filepath.Join(destRootDir, filepath.Base(srcSubDir), filepath.Base(srcSubFile.Name())))
-			Expect(err).NotTo(HaveOccurred())
-			Expect(f.Mode()).To(Equal(os.ModePerm))
+			Expect(destRootDir).NotTo(BeAnExistingFile())
 		})
 	})
+
```
sykesm: nit: unnecessary blank line

core/container/externalbuilder/copy_test.go
```diff
@@ -19,23 +20,35 @@ import (
...
+		err = ioutil.WriteFile(filepath.Join(srcRootDir, "file-in-root.txt"), []byte{}, os.ModePerm)
 		Expect(err).NotTo(HaveOccurred())
-		srcRootFile, err = ioutil.TempFile(srcRootDir, "file-")
+
+		err = os.Symlink("file-in-root.txt", filepath.Join(srcRootDir, "symlink-in-root.txt"))
 		Expect(err).NotTo(HaveOccurred())
-		srcSubFile, err = ioutil.TempFile(srcSubDir, "subfile-")
+
+		err = os.Mkdir(filepath.Join(srcRootDir, "subdir"), os.ModePerm)
```
sykesm: The repeated use of `filepath.Join(srcRootDir, "subdir")` means we should probably have a var.

sykesm (changes requested):

sykesm (approved):

sstone1:

PR #1235, opened 2020-05-11
[FAB-17863] Enhance rebuild-dbs to drop CouchDB state database

wenjianqiao:

#### Type of change

#### Description
The rebuild-dbs command should be able to drop CouchDB state db.

#### Related issues
https://jira.hyperledger.org/browse/FAB-17863


manish-sethi:
LGTM. However, to take a note of two related things
- A test can be added in the kvledger/tests level to check for error propagation for APIs such as upgrade and rebuild, when the couchdb is not running
- Just an orthogonal thought, we can also do the actual rebuild inside the functions "Rebuild and UpgradeDB" to actually justify the API names. That way, admin commands will be more aligned as well and the peer start next time would be no different from a regular peer start - wdyt?

manish-sethi (approved):

PR #1236, opened 2020-05-11
[FAB-17862] Allow upgrade-dbs to drop all dbs in multiple runs

wenjianqiao:

#### Type of change

#### Description
In some cases upgrade-dbs may be run multiple times; e.g., if admin forgot 
to pass the CouchDB variable CORE_LEDGER_STATE_STATEDATABASE,
it will not drop the CouchDB databases as part of upgrade-dbs.
Therefore, upgrade-dbs should be able to drop all dbs when
running multiple times.

#### Related issues
https://jira.hyperledger.org/browse/FAB-17862

### DIFF
--- a/core/ledger/kvledger/kv_ledger_provider.go
+++ b/core/ledger/kvledger/kv_ledger_provider.go
@@ -509,7 +509,7 @@ func (s *idStore) checkUpgradeEligibility() (bool, error) {
 		return false, err
 	}
 	if emptydb {
-		logger.Warnf("Ledger database %s is empty, nothing to upgrade.", s.dbPath)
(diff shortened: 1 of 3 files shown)


manish-sethi:
I think that it's worth adding a simple test at kvledger/tests level to verify that the upgrade can be invoked multiple times. Even a simple test with leveldb should suffice - wdyt?

wenjianqiao:
> I think that it's worth adding a simple test at kvledger/tests level to verify that the upgrade can be invoked multiple times. Even a simple test with leveldb should suffice - wdyt?

Thank you Manish. Agree. I plan to add additional upgrade tests (upgrade multiple times, couchdb not started, couchdb not configured at 1st time, etc) in a separate PR.

manish-sethi (approved):

PR #1239, opened 2020-05-12
Without checking for errors when creating the server asynchronously, …

duanmengkk:
…we should at least print the log

#### Type of change

- Bug fix
- New feature
- Test update
- Documentation update

### DIFF
--- a/core/operations/system.go
+++ b/core/operations/system.go
@@ -120,7 +120,12 @@ func (s *System) Start() error {
 	}
 	s.addr = listener.Addr().String()
 
-	go s.httpServer.Serve(listener)
+	go func() {
+		err := s.httpServer.Serve(listener)
+		if err != nil {
+			s.Log("http server serve failed")
+		}
+	}()
 
 	return nil
 }


sykesm (changes requested):
Thank you for your contribution.

I’m not sure what is motivating logging like this. 'Serve` *always* returns an error.

```
Serve accepts incoming connections on the Listener l, creating a new service goroutine for each. The service goroutines read requests and then call srv.Handler to reply to them.
...
Serve always returns a non-nil error and closes l. After Shutdown or Close, the returned error is ErrServerClosed..
```
The error that comes back is either because the server was shutdown, the listener was closed, or because accept failed with a non-network error. 

Two of these are normal during termination so a warning log does not seem appropriate. 

Also, logging requires context. The message does not provide that. It doesn’t include the error or describe which http server exited.

Finally, production code changes usually include tests. While that’s not always required for logging, this package does wire a test logger and make assertions on the messages. 

If there is a real concern motivating this change, please do not log warnings on the expected shutdown paths, add context to the message, and add tests.

Thanks!

duanmengkk:
ok，i will close it

PR #1241, opened 2020-05-12
[FAB-17865] Update fabric upgrade doc for dropping CouchDB

wenjianqiao:

#### Type of change
- Documentation update

#### Description
The "upgrade-dbs" peer command has been enhanced to automatically
drop the CouchDB state database as of v2.2.0. Update the upgrade
document accordingly.

#### Related issues
https://jira.hyperledger.org/browse/FAB-17865

### DIFF
--- a/docs/source/upgrade_to_newest_version.md
+++ b/docs/source/upgrade_to_newest_version.md
@@ -42,9 +42,9 @@ For more information, check out [Logging control](./logging-control.html#chainco
 
 ### Peer databases upgrade
 
-For information about how to upgrade peers, check out our documentation on [upgrading components](./upgrading_your_components.html). During the process for [upgrading your peers](./upgrading_your_components.html#upgrade-the-peers), you will need to perform one additional step to upgrade the peer databases. The databases of all peers (which include not just the state database but the history database and other internal databases for the peer) must be rebuilt using the v2.x data format as part of the upgrade to v2.x. To trigger the rebuild, the databases must be dropped before the peer is started. The instructions below utilize the `peer node upgrade-dbs` command to drop the local databases managed by the peer and prepare them for upgrade, so that they can be rebuilt the first time the v2.x peer starts. If you are using CouchDB as the state database, since it is not a local database, the peer cannot automatically drop this database. You must drop the CouchDB state database yourself.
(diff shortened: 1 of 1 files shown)


docs/source/upgrade_to_newest_version.md
```diff
@@ -42,9 +42,9 @@ For more information, check out [Logging control](./logging-control.html#chainco
...
+For information about how to upgrade peers, check out our documentation on [upgrading components](./upgrading_your_components.html). During the process for [upgrading your peers](./upgrading_your_components.html#upgrade-the-peers), you will need to perform one additional step to upgrade the peer databases. The databases of all peers (which include not just the state database but the history database and other internal databases for the peer) must be rebuilt using the v2.x data format as part of the upgrade to v2.x. To trigger the rebuild, the databases must be dropped before the peer is started. The instructions below utilize the `peer node upgrade-dbs` command to drop the local databases managed by the peer and prepare them for upgrade, so that they can be rebuilt the first time the v2.x peer starts. If you are using CouchDB as the state database, the peer has support to automatically drop this database as of v2.2. To leverage the support, you must configure the peer with CouchDB as the state database and start CouchDB before running the `upgrade-dbs` command. In v2.0 and v2.1, the peer does not automatically drop the CouchDB state database; therefore you must drop it yourself.
```
pamandrejko, line 45: I'm unclear what this means because v2.2. is not available yet:
| _If you are using CouchDB as the state database, the peer has support to automatically drop this database as of v2.2._

Do you mean this==>
If you are using CouchDB as the state database, you must manually drop the database before running the `upgrade-dbs` command. In  Fabric v2.0 and v2.1, the peer does not automatically drop the CouchDB state database; therefore you must drop it yourself. In a future release, the database will be dropped for you as part of the  `upgrade-dbs` command.

docs/source/upgrade_to_newest_version.md
```diff
@@ -54,7 +54,7 @@ docker run --rm -v /opt/backup/$PEER_CONTAINER/:/var/hyperledger/production/ \
             hyperledger/fabric-peer:2.0 peer node upgrade-dbs
 ```
 
-If you are using CouchDB as the state database, also drop the CouchDB database.  This can be done by removing the CouchDB /data volume directory.
+In v2.0 and v2.1, if you are using CouchDB as the state database, also drop the CouchDB database. This can be done by removing the CouchDB /data volume directory.
```
pamandrejko, line 57: In Fabric v2.0 and v2.1, if you are using CouchDB as the state database, you also need to drop the CouchDB database. This can be done by removing the CouchDB /data volume directory.

pamandrejko (dismissed):
A few minor clarifications.

docs/source/upgrade_to_newest_version.md
cendhu: @wenjianqiao We can be more explicit here. When the admin upgrades the peer from v1.4.x to either v2.0 or v2.1, admin needs not to keep the CouchDB container up before running upgradeDBs instead drop couchDB data manually. If the admin upgrades from v1.4.x to >2.1, the peer itself would drop the CouchDB data. Hence, the admin must start the CouchDB container before running upgradeDBs. Not the exact sentences but something similar to what is conveyed regarding the version.

wenjianqiao: @pamandrejko @cendhu Thank you for the comments. This doc update is meant for v2.2. I think we can wait for v2.2 instead of updating it now with "a future release".

denyeart (approved):

PR #1242, opened 2020-05-12
rm DB interface from privacyenabledstate pkg

cendhu:
#### Type of change

#### Description

As part of [FAB-10794](https://jira.hyperledger.org/browse/FAB-10794), we would be making changes to `LoadCommittedVersions()`. Before doing that, we remove `DB` interface from the `privacyenabledstate` pkg (which includes LoadCommittedVersions() API) as it has a single implementation only. Only when we need to have multiple implementations, we need an interface. Otherwise, it just adds unnecessary difficulties during code navigation and casting such as `db.(*privacyenabledstate.CommonStorageDB)` in the test.

As we remove `DB` interface, the filename `db.go` does not make sense. Hence, we rename it to `types.go` as it holds all private data related types. Further, `db_test.go` contains tests that
should be in `common_storage_db_test.go`. Hence, we also move those test to the correct location and rename `db_test.go` to `types_test.go`.

Further, we have 

1. renamed `CommonStorageDB` to `DB`
2. removed the `DBProvider` interface, 
3. removed the term `CommonStorage` from variables
4. renamed `common_storage_db.go` to `db.go` and `common_storage_db_test.go` to `db_test.go`

#### Additional details

Two commits have been pushed to ease the review process. 

#### Related issues

[FAB-17880](https://jira.hyperledger.org/browse/FAB-17880)


manish-sethi (changes requested):
Discussed offline with @cendhu for some further changes - so, holding this for now.

core/ledger/kvledger/kv_ledger_provider.go
```diff
@@ -319,7 +319,7 @@ func (p *Provider) open(ledgerID string) (ledger.PeerLedger, error) {
 	p.collElgNotifier.registerListener(ledgerID, pvtdataStore)
 
 	// Get the versioned database (state database) for a chain/ledger
-	vDB, err := p.vdbProvider.GetDBHandle(ledgerID)
+	db, err := p.dbProvider.GetDBHandle(ledgerID)
```
manish-sethi, line 322: In the lines of above comment, `statedb` may be a better choice, if not `vDB`. Plain `db` looses the context.

core/ledger/kvledger/kv_ledger_provider.go
```diff
@@ -65,7 +65,7 @@ type Provider struct {
 	idStore              *idStore
 	blkStoreProvider     *blkstorage.BlockStoreProvider
 	pvtdataStoreProvider *pvtdatastorage.Provider
-	vdbProvider          privacyenabledstate.DBProvider
+	dbProvider           *privacyenabledstate.DBProvider
```
manish-sethi, line 68: Given that we have more than one stores, I feel that the earlier name `vdbProvider` was clearer. If you want to change, then `statedbProvider` may be a better choice in order to retain the context.

manish-sethi (approved):
Just some nit comments on a couple variable names. But would not like to hold this PR for that.

cendhu:
Thanks, @manish-sethi I agree with the suggested changes and have created https://jira.hyperledger.org/browse/FAB-17896 to address the review comments.

PR #1220, opened 2020-05-07
[FAB-17433] Add _lifecycle function for querying approved chaincode

satota2:
This patch adds a function for querying the details of the chaincode definition approved by the organization itself to `_lifecycle`, so that Fabric peer can provide a function to query them.

#### Type of change

- New feature

#### Description

Currently, Fabric peer does not provide a function to query the details of approved chaincode definitions.

On the other hand, Fabric admins will need to confirm/use the information after approval for their operations.

So, we propose to add a function for querying the details of the approved chaincode definition to Fabric peer CLI (Refer to https://jira.hyperledger.org/browse/FAB-17401).

As a sub-task of FAB-17401, this patch adds a function for querying the details of the chaincode definition approved by the organization itself to `_lifecycle`, so that Fabric peer can provide a function to query them.

The following patch has been merged to extend lifecycle protos:
- https://github.com/hyperledger/fabric-protos/pull/25

Immediately after this patch has been merged into the fabric repository,  I will submit the following related patches:
- https://github.com/satota2/fabric/commit/d6f60009dce9f74caa4df5ac49951d0a0998aac0
- https://github.com/satota2/fabric/commit/fec797b75c40163e9dd92dc1e51acadd4756c7a3
- https://github.com/satota2/fabric/commit/f4d5fff151dc87a597a058184c203b9e9ce72629

#### Related issues
- Main: 
  - https://jira.hyperledger.org/projects/FAB/issues/FAB-17432

- Others: 
  - https://jira.hyperledger.org/projects/FAB/issues/FAB-17401
  - https://jira.hyperledger.org/projects/FAB/issues/FAB-17432
  - https://jira.hyperledger.org/projects/FAB/issues/FAB-17434
  - https://jira.hyperledger.org/projects/FAB/issues/FAB-17435
  - https://jira.hyperledger.org/projects/FAB/issues/FAB-17436

PR #1243, opened 2020-05-12
[FAB-17434] Add `queryapproved` command to query the details of the approved chaincode definition

satota2:
This patch adds `queryapproved` command for querying the details of the chaincode definition approved by the organization itself to peer CLI, so that Fabric peer can provide a function to query them.
Also, this patch fixes default value of `sequence` parameter to empty for several reasons.

#### Type of change

- New feature
- Bug fix

#### Description

Currently, Fabric peer does not provide a function to query the details of approved chaincode definitions.

On the other hand, Fabric admins will need to confirm/use the information after approval for their operations.

So, we propose to add a function for querying the details of the approved chaincode definition to Fabric peer CLI (Refer to https://jira.hyperledger.org/browse/FAB-17401).

As a sub-task of FAB-17401, this patch adds `queryapproved` command for querying the details of the chaincode definition approved by the organization itself to peer CLI, so that Fabric peer can provide a function to query them.

Also, this patch fixes default value of `sequence` parameter to `0` that means empty for the following reasons.

- According to `Validate()` in `approveformyorg.go`, this function is implemented assuming that `sequence` is `0`. This is inconsistent with the default value `1`.

- The current default value `1` enables Fabric users `approveformyorg` without specifying the sequence number. Not forcing the specification of a sequence number may cause operation mistakes (e.g, updating the package-id for sequence 1 by mistake) when running `approveformyorg` in the second and subsequent times.

- The default value is preferably `0` to reuse the same parameter `sequence` in CLI commands to query the details of approved chaincode definitions.

The following patches has been merged to extend lifecycle protos and to add `_lifecycle` function for querying approved chaincode:
- https://github.com/hyperledger/fabric-protos/pull/25
- https://github.com/hyperledger/fabric/pull/1220

Immediately after this patch has been merged into the fabric repository,  I will submit the following related patches:
- https://github.com/satota2/fabric/commit/f4d5fff151dc87a597a058184c203b9e9ce72629

#### Related issues
- Main: 
  - https://jira.hyperledger.org/projects/FAB/issues/FAB-17434

- Others: 
  - https://jira.hyperledger.org/projects/FAB/issues/FAB-17401
  - https://jira.hyperledger.org/projects/FAB/issues/FAB-17432
  - https://jira.hyperledger.org/projects/FAB/issues/FAB-17433
  - https://jira.hyperledger.org/projects/FAB/issues/FAB-17435
  - https://jira.hyperledger.org/projects/FAB/issues/FAB-17436

### DIFF
--- a/internal/peer/lifecycle/chaincode/chaincode.go
+++ b/internal/peer/lifecycle/chaincode/chaincode.go
@@ -46,6 +46,7 @@ func Cmd(cryptoProvider bccsp.BCCSP) *cobra.Command {
 	chaincodeCmd.AddCommand(QueryInstalledCmd(nil, cryptoProvider))
 	chaincodeCmd.AddCommand(GetInstalledPackageCmd(nil, cryptoProvider))
 	chaincodeCmd.AddCommand(ApproveForMyOrgCmd(nil, cryptoProvider))
+	chaincodeCmd.AddCommand(QueryApprovedCmd(nil, cryptoProvider))
 	chaincodeCmd.AddCommand(CheckCommitReadinessCmd(nil, cryptoProvider))
 	chaincodeCmd.AddCommand(CommitCmd(nil, cryptoProvider))
 	chaincodeCmd.AddCommand(QueryCommittedCmd(nil, cryptoProvider))
@@ -80,8 +81,8 @@ var (
 
 var chaincodeCmd = &cobra.Command{
 	Use:   "chaincode",
-	Short: "Perform chaincode operations: package|install|queryinstalled|getinstalledpackage|approveformyorg|checkcommitreadiness|commit|querycommitted",
-	Long:  "Perform chaincode operations: package|install|queryinstalled|getinstalledpackage|approveformyorg|checkcommitreadiness|commit|querycommitted",
+	Short: "Perform chaincode operations: package|install|queryinstalled|getinstalledpackage|approveformyorg|queryapproved|checkcommitreadiness|commit|querycommitted",
+	Long:  "Perform chaincode operations: package|install|queryinstalled|getinstalledpackage|approveformyorg|queryapproved|checkcommitreadiness|commit|querycommitted",
 	PersistentPreRun: func(cmd *cobra.Command, args []string) {
 		common.InitCmd(cmd, args)
 		common.SetOrdererEnv(cmd, args)
--- a/internal/peer/lifecycle/chaincode/queryapproved.go
+++ b/internal/peer/lifecycle/chaincode/queryapproved.go
@@ -1,0 +1,215 @@
+/*
+Copyright Hitachi America, Ltd. All Rights Reserved.
+
+SPDX-License-Identifier: Apache-2.0
+*/
+
+package chaincode
+
+import (
+	"context"
+	"fmt"
+	"io"
+	"os"
+	"strings"
+
+	"github.com/golang/protobuf/proto"
+	cb "github.com/hyperledger/fabric-protos-go/common"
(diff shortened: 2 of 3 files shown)


internal/peer/lifecycle/chaincode/chaincode.go
```diff
@@ -123,7 +124,7 @@ func ResetFlags() {
 	flags.DurationVar(&waitForEventTimeout, "waitForEventTimeout", 30*time.Second,
 		fmt.Sprint("Time to wait for the event from each peer's deliver filtered service signifying that the 'invoke' transaction has been committed successfully"))
 	flags.StringVarP(&packageID, "package-id", "", "", "The identifier of the chaincode install package")
-	flags.IntVarP(&sequence, "sequence", "", 1, "The sequence number of the chaincode definition for the channel")
+	flags.IntVarP(&sequence, "sequence", "", 0, "The sequence number of the chaincode definition for the channel")
```
jyellick: This is a change in the default value for this command.  I think I prefer this default, as it works with the defaults provided by the system chaincode (0 is 'unset').  But, @nikhil550 @denyeart will this break anything in our samples/docs etc.?

internal/peer/lifecycle/chaincode/queryapproved.go
```diff
@@ -0,0 +1,215 @@
...
+	Sequence     int64
+	OutputFormat string
+}
+
+// QueryApprovedCmd returns the cobra command for
+// querying the approved chaincode definition for the organization
+func QueryApprovedCmd(a *ApprovedQuerier, cryptoProvider bccsp.BCCSP) *cobra.Command {
+	chaincodeQueryApprovedCmd := &cobra.Command{
+		Use:   "queryapproved",
+		Short: fmt.Sprintf("Query the approved chaincode definition for my org."),
```
jyellick: This is no longer "for my org", it would be for any org, so perhaps more appropriately "Query an org's approved chaincode definition from its peer." or something like that.

jyellick (commented):
This generally looks good to me -- the only thing that gets me a little worried is changing the default of a command flag.  I think it's still probably for the best, but would like to gather some additional opinions.

internal/peer/lifecycle/chaincode/chaincode.go
nikhil550: None of the samples use the default value for the sequence parameter.

denyeart: I agree with the default change, we'll need to release note it with v2.2.
However I would expect the default change to come in a separate PR, and also the command line help text that mentions the default needs to be updated in the same PR:

`--sequence int                   The sequence number of the chaincode definition for the channel (default 1)`

satota2: @jyellick @nikhil550 @denyeart
Thank you for your reviews.

Based on the comments, I submitted a separate PR for this as below.
The PR also includes updates to the help docs.
https://github.com/hyperledger/fabric/pull/1262

Immediately after the separate PR has been merged, I will resubmit the updated patch to address the other comments.

internal/peer/lifecycle/chaincode/queryapproved.go
satota2: @jyellick Thank you for your review.
I will submit the updated patch to address the comments after #1262 is merged.

satota2:
@jyellick
Thank you for merging the PR (#1262).
I updated this patch to address the other comments.

jyellick (approved):

PR #1256, opened 2020-05-13
Provide more context in identity deserialization warning message

wlahti:
#### Type of change

#### Description

Improve MSP error message and include certificate subject/serial number in logged warning message. 

#### Related issues

[FAB-14099](https://jira.hyperledger.org/browse/FAB-14099)

### DIFF
--- a/common/policies/policy.go
+++ b/common/policies/policy.go
@@ -7,6 +7,8 @@ SPDX-License-Identifier: Apache-2.0
 package policies
 
 import (
+	"crypto/x509"
+	"encoding/pem"
 	"fmt"
 	"strings"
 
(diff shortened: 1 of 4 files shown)


common/policies/policy.go
```diff
@@ -393,3 +400,20 @@ func SignatureSetToValidIdentities(signedData []*protoutil.SignedData, identityD
...
 }
+
+func getCertFromSerializedIdentity(serializedIdentity []byte) (*x509.Certificate, error) {
+	id := &msp.SerializedIdentity{}
+	err := proto.Unmarshal(serializedIdentity, id)
+	if err != nil {
+		return nil, errors.Wrap(err, "unmarshaling serialized identity")
+	}
+	pemBlock, _ := pem.Decode(id.IdBytes)
+	if pemBlock == nil {
```
yacovm, line 411: A SerializedIdentity doesn't mean you **have to have** an x509 certificate inside it. 
I suggest we make a function that returns a string to be logged, instead of the certificate.
If the identity has a PEM, it will attempt to parse a certificate and upon success return these fields you want, as well as the MSP ID.
In any other case, either the identity not being a PEM, or just a failure, it will return the MSP ID and the identity bytes in base64 encoding.

yacovm (changes requested):
Please see comment.

yacovm (approved):

PR #1258, opened 2020-05-13
Update deps to fix dist/windows-amd64 build

sykesm:
Cross-compiling for Windows fails against master:

```
Building release/windows-amd64/bin/peer for windows-amd64
mkdir -p release/windows-amd64/bin
GOOS=windows GOARCH=amd64 go build -o release/windows-amd64/bin/peer -tags "" -ldflags "-X github.com/hyperledger/fabric/common/metadata.Version=2.2.0 -X github.com/hyperledger/fabric/common/metadata.CommitSHA=3d69ec0f3 -X github.com/hyperledger/fabric/common/metadata.BaseDockerLabel=org.hyperledger.fabric -X github.com/hyperledger/fabric/common/metadata.DockerNamespace=hyperledger" github.com/hyperledger/fabric/cmd/peer
# github.com/docker/docker/pkg/system
vendor/github.com/docker/docker/pkg/system/filesys_windows.go:111:24: cannot use uintptr(unsafe.Pointer(&sd[0])) (type uintptr) as type *"golang.org/x/sys/windows".SECURITY_DESCRIPTOR in assignment
make: *** [release/windows-amd64/bin/peer] Error 2
```

Fix this by moving to an older version of golang.org/x/sys.


lindluni:
Thank you for this, I hit this in the CA yesterday, and hadn't yet investigated the issue.

lindluni:
Tested locally, works well

lindluni (approved):

go.mod
```diff
@@ -2,6 +2,9 @@ module github.com/hyperledger/fabric
 
 go 1.14
 
+// https://github.com/golang/go/issues/34610
+replace golang.org/x/sys => golang.org/x/sys v0.0.0-20190920190810-ef0ce1748380
```
lindluni, line 6: Out of curiosity, why do the `replace` rather than pin it to the version in the `require` block

sykesm: I'm still learning the tao of modules so it may not be a best practice.

Putting the version in the require block gets us the right version but it doesn't prevent a `go get -u` from updating it automatically. So if, for example, someone wanted to update grpc, the version of x/sys would also get bumped because that package depends on a slightly newer version. By using a `replace` directive, we unilaterally pick the version.

I think if the change impacted CI or development on Darwin, I'm not sure I would bother with a replace because our regular builds would fail. Since it's just for Windows, I wanted to highlight it.

I can go either way.

lindluni: The question was really just for my curiosity and understanding. Like you I am just trying to understand everything with modules as well. While it's a step in the right direction, it doesn't totally behave live other software package managers I am used to (such as npm).

It makes sense to use the replace directive to prevent a transient dependency (such as gRPC) from updating the package to something that doesn't work for us.

Thanks for the clarification.

PR #1268, opened 2020-05-14
[FAB-17793] Generate db name to namespace mapping for state couchdb

wenjianqiao:

#### Type of change
- New feature

#### Description
This PR generates db name to namespace mapping for state couchdb and
stores the mapping data in channel’s metadata db. Due to couchdb's
length restriction on db names, the mapping is needed to drop all
the databases for a channel as well as snapshot support.

#### Additional details
When CouchDB is used as statedb, internal databases are created at a per-chaincode/collection basis. Because CouchDB has length limitation on db names, channel name may be truncated if the combined length of channel+namespace+collection is over the limit. Due to the truncation, the original namespaces are lot and we may not be able to re-construct the database names. 

#### Related issues
https://jira.hyperledger.org/browse/FAB-17793

### DIFF
--- a/core/ledger/kvledger/txmgmt/statedb/statecouchdb/couchdoc_conv.go
+++ b/core/ledger/kvledger/txmgmt/statedb/statecouchdb/couchdoc_conv.go
@@ -184,6 +184,17 @@ type couchSavepointData struct {
 	TxNum    uint64 `json:"TxNum"`
 }
 
+type channelMetadata struct {
+	ChannelName string `json:"ChannelName"`
+	// namespace to namespaceDBInfo mapping
+	NamespaceDBsInfo map[string]*namespaceDBInfo `json:"NamespaceDBsInfo"`
+}
+
+type namespaceDBInfo struct {
+	Namespace string `json:"Namespace"`
+	DBName    string `json:"DBName"`
+}
+
 func encodeSavepoint(height *version.Height) (*couchDoc, error) {
 	var err error
 	var savepointDoc couchSavepointData
@@ -207,6 +218,26 @@ func decodeSavepoint(couchDoc *couchDoc) (*version.Height, error) {
 		return nil, err
 	}
 	return &version.Height{BlockNum: savepointDoc.BlockNum, TxNum: savepointDoc.TxNum}, nil
+}
+
+func encodeChannelMetadata(metadataDoc *channelMetadata) (*couchDoc, error) {
+	metadataJSON, err := json.Marshal(metadataDoc)
+	if err != nil {
+		err = errors.Wrap(err, "failed to marshal channel metadata")
+		logger.Errorf("%+v", err)
+		return nil, err
+	}
+	return &couchDoc{jsonValue: metadataJSON, attachments: nil}, nil
+}
+
+func decodeChannelMetadata(couchDoc *couchDoc) (*channelMetadata, error) {
+	metadataDoc := &channelMetadata{}
+	if err := json.Unmarshal(couchDoc.jsonValue, &metadataDoc); err != nil {
+		err = errors.Wrap(err, "failed to unmarshal channel metadata")
+		logger.Errorf("%+v", err)
+		return nil, err
+	}
+	return metadataDoc, nil
 }
 
 type dataformatInfo struct {
--- a/core/ledger/kvledger/txmgmt/statedb/statecouchdb/statecouchdb.go
+++ b/core/ledger/kvledger/txmgmt/statedb/statecouchdb/statecouchdb.go
@@ -27,6 +27,10 @@ var logger = flogging.MustGetLogger("statecouchdb")
 const (
 	// savepointDocID is used as a key for maintaining savepoint (maintained in metadatadb for a channel)
 	savepointDocID = "statedb_savepoint"
+	// channelMetadataDocID is used as a key to store the channel metadata for a channel (maintained in the channel's metadatadb).
+	// Due to CouchDB's length restriction on db names, channel names and namepsaces may be truncated in db names.
+	// The metadata is used for dropping channel-specific databases and snapshot support.
+	channelMetadataDocID = "channel_metadata"
 	// fabricInternalDBName is used to create a db in couch that would be used for internal data such as the version of the data format
 	// a double underscore ensures that the dbname does not clash with the dbnames created for the chaincodes
 	fabricInternalDBName = "fabric__internal"
@@ -169,7 +173,8 @@ type VersionedDB struct {
 	couchInstance      *couchInstance
 	metadataDB         *couchDatabase            // A database per channel to store metadata such as savepoint.
 	chainName          string                    // The name of the chain/channel.
-	namespaceDBs       map[string]*couchDatabase // One database per deployed chaincode.
+	namespaceDBs       map[string]*couchDatabase // One database per namespace.
+	channelMetadata    *channelMetadata          // Store channel name and namespaceDBInfo
 	committedDataCache *versionsCache            // Used as a local cache during bulk processing of a block.
 	verCacheLock       sync.RWMutex
 	mux                sync.RWMutex
@@ -197,6 +202,21 @@ func newVersionedDB(couchInstance *couchInstance, redoLogger *redoLogger, dbName
 		redoLogger:         redoLogger,
 		cache:              cache,
 	}
+
+	vdb.channelMetadata, err = vdb.readChannelMetadata()
+	if err != nil {
+		return nil, err
+	}
+	if vdb.channelMetadata == nil {
+		vdb.channelMetadata = &channelMetadata{
+			ChannelName:      chainName,
+			NamespaceDBsInfo: make(map[string]*namespaceDBInfo),
+		}
+		if err = vdb.writeChannelMetadata(); err != nil {
+			return nil, err
+		}
+	}
+
 	logger.Debugf("chain [%s]: checking for redolog record", chainName)
 	redologRecord, err := redoLogger.load()
 	if err != nil {
@@ -241,6 +261,16 @@ func (vdb *VersionedDB) getNamespaceDBHandle(namespace string) (*couchDatabase, 
 	db = vdb.namespaceDBs[namespace]
 	if db == nil {
 		var err error
+		if _, ok := vdb.channelMetadata.NamespaceDBsInfo[namespace]; !ok {
+			logger.Debugf("[%s] add namespaceDBInfo for namespace %s", vdb.chainName, namespace)
+			vdb.channelMetadata.NamespaceDBsInfo[namespace] = &namespaceDBInfo{
+				Namespace: namespace,
+				DBName:    namespaceDBName,
+			}
+			if err = vdb.writeChannelMetadata(); err != nil {
+				return nil, err
+			}
+		}
 		db, err = createCouchDatabase(vdb.couchInstance, namespaceDBName)
 		if err != nil {
 			return nil, err
@@ -690,6 +720,19 @@ func (vdb *VersionedDB) Close() {
 	// no need to close db since a shared couch instance is used
 }
 
+// writeChannelMetadata saves channel metadata to metadataDB
+func (vdb *VersionedDB) writeChannelMetadata() error {
+	couchDoc, err := encodeChannelMetadata(vdb.channelMetadata)
+	if err != nil {
+		return err
+	}
+	if _, err := vdb.metadataDB.saveDoc(channelMetadataDocID, "", couchDoc); err != nil {
+		return err
+	}
+	_, err = vdb.metadataDB.ensureFullCommit()
+	return err
+}
+
 // ensureFullCommitAndRecordSavepoint flushes all the dbs (corresponding to `namespaces`) to disk
 // and Record a savepoint in the metadata db.
 // Couch parallelizes writes in cluster or sharded setup and ordering is not guaranteed.
@@ -767,6 +810,21 @@ func (vdb *VersionedDB) GetLatestSavePoint() (*version.Height, error) {
 		return nil, nil
 	}
 	return decodeSavepoint(couchDoc)
+}
+
+// readChannelMetadata returns channel metadata stored in metadataDB
+func (vdb *VersionedDB) readChannelMetadata() (*channelMetadata, error) {
+	var err error
+	couchDoc, _, err := vdb.metadataDB.readDoc(channelMetadataDocID)
+	if err != nil {
+		logger.Errorf("Failed to read db name mapping data %s", err.Error())
+		return nil, err
+	}
+	// ReadDoc() not found (404) will result in nil response, in these cases return nil
+	if couchDoc == nil || couchDoc.jsonValue == nil {
+		return nil, nil
+	}
+	return decodeChannelMetadata(couchDoc)
 }
 
 func (vdb *VersionedDB) GetFullScanIterator(skipNamespace func(string) bool) (statedb.FullScanIterator, byte, error) {
--- a/core/ledger/kvledger/txmgmt/statedb/statecouchdb/statecouchdb_test.go
+++ b/core/ledger/kvledger/txmgmt/statedb/statecouchdb/statecouchdb_test.go
@@ -1365,3 +1365,127 @@ func TestMissingRevisionRetrievalFromCache(t *testing.T) {
 	require.Equal(t, "rev1", revisions["key1"])
 	require.Equal(t, "rev2", revisions["key2"])
 }
+
+func TestChannelMetadata(t *testing.T) {
+	vdbEnv.init(t, sysNamespaces)
+	defer vdbEnv.cleanup()
+	channelName := "testchannelmetadata"
+
+	db, err := vdbEnv.DBProvider.GetDBHandle(channelName)
+	require.NoError(t, err)
+	vdb := db.(*VersionedDB)
+	expectedChannelMetadata := &channelMetadata{
+		ChannelName:      channelName,
+		NamespaceDBsInfo: make(map[string]*namespaceDBInfo),
+	}
+	savedChannelMetadata, err := vdb.readChannelMetadata()
+	require.NoError(t, err)
+	require.Equal(t, expectedChannelMetadata, savedChannelMetadata)
+	require.Equal(t, expectedChannelMetadata, vdb.channelMetadata)
+
+	// call getNamespaceDBHandle for new dbs, verify that new db names are added to dbMetadataMapping
+	namepsaces := make([]string, 10)
+	for i := 0; i < 10; i++ {
+		ns := fmt.Sprintf("nsname_%d", i)
+		_, err := vdb.getNamespaceDBHandle(ns)
+		require.NoError(t, err)
+		namepsaces[i] = ns
+		expectedChannelMetadata.NamespaceDBsInfo[ns] = &namespaceDBInfo{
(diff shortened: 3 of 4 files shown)


cendhu (changes requested):
Thanks, @wenjianqiao, for prioritizing this. I quickly went over the changes and have a high-level comment. Once the comment is addressed, I will do a detailed review. 

We have two requirements:

1. find all dbNames associated with a channel so that we can drop databases at per channel level while executing peer rollback and reset. 
2. find the mapping between a dbName and the namespace it handles. 
   - Given that we trim the namespace to create the associated dbName as per CouchDB dbName's length restriction and do not store this mapping anywhere, this mapping information is lost when considering standalone couchdb pkg. However, the mapping info is required for the snapshot story. 
   - As part of statedb snapshotting, in #1261, we have introduced a new API called `GetFullScanIterator(skipNamespace func(string) bool)` and is implemented only in the leveldb pkg for now. When we implement the same for the CouchDB, we need to know all databases associated with a channel (has been addressed in this PR) as well as the namespace associated with each database. This is because, to the `skipNamespace` function, the CouchDB would need to pass the namespace (not the internal dbName) to find out whether a namespace denotes a pvtdata so that it would be skipped. 

Currently, this PR satisfies only the requirement (1) .  Can you update this PR to address the requirement (2) as well? Basically, we need to map each internal dbName to its full namespace and store it per channel (in the metadataDB).

wenjianqiao:
@cendhu You are quick to response, Thank you. The subtask and story were only for requirement 1, but agree it makes sense to cover both requirements in a single PR.

core/ledger/kvledger/txmgmt/statedb/statecouchdb/statecouchdb.go
```diff
@@ -238,6 +258,9 @@ func (vdb *VersionedDB) getNamespaceDBHandle(namespace string) (*couchDatabase,
 	namespaceDBName := constructNamespaceDBName(vdb.chainName, namespace)
 	vdb.mux.Lock()
 	defer vdb.mux.Unlock()
+	if err := vdb.updateAndSaveNamespaceDBsInfo(namespace, namespaceDBName); err != nil {
+		return nil, err
+	}
```
cendhu: 1. Only after we get success from `createCouchDatabase()`, we sould update the channel metadata. Can you place these lines after line 271? 
2. We seem to continue the pattern of side-effect functions. The `createCouchDatabase()` is in fact `createCouchDatabaseIfNotExist()`. For code readers, it is natural to assume that the `createCouchDatabase()` creates a new database. Similarly, when I read `updateAndSaveNamespaceDBsInfo()`, I was wondering why we aren't checking whether the mapping already exists (as createCouchDatabase() does not create a new database always) before performing the update. When I looked at the `updateAndSaveNamespaceDBsInfo()`, I found that we repeat the hidden `if not-exist` pattern. Personally, I would prefer to have the check at the caller before calling the update function (which should blindly update it) similar to `Put()`. If only I find this pattern to add difficulty to code readability, please feel free to ignore this point.

core/ledger/kvledger/txmgmt/statedb/statecouchdb/statecouchdb.go
```diff
@@ -690,6 +713,40 @@ func (vdb *VersionedDB) Close() {
...
+		return err
+	}
+	dbResponse, err := vdb.metadataDB.ensureFullCommit()
+	if err != nil {
+		return err
+	}
+	if !dbResponse.Ok {
+		logger.Errorf("failed to perform full commit while saving couchdb metadata")
+		return errors.New("failed to perform full commit while saving couchdb metadata")
+	}
```
cendhu: 1. The `ensureFullCommit()` checks for `dbResponse.Ok` and return an error if it is false. Hence, we don't need to repeat the same check here. 
2. In a separate PR, we can refactor `ensureFullCommit()` as it returns the dbResponse unnecessarily which leads to such code duplication.

core/ledger/kvledger/txmgmt/statedb/statecouchdb/statecouchdb.go
```diff
@@ -769,6 +826,21 @@ func (vdb *VersionedDB) GetLatestSavePoint() (*version.Height, error) {
 	return decodeSavepoint(couchDoc)
 }
 
+// getLatestNamespaceDBsInfo returns db name mapping for the VersionedDB
+func (vdb *VersionedDB) getLatestNamespaceDBsInfo() ([]*namespaceDBInfo, error) {
```
cendhu:
```suggestion
func (vdb *VersionedDB) getChannelMetadata() (*channelMetadata, error) {
```

core/ledger/kvledger/txmgmt/statedb/statecouchdb/couchdoc_conv.go
```diff
@@ -184,6 +184,10 @@ type couchSavepointData struct {
 	TxNum    uint64 `json:"TxNum"`
 }
 
+type channelMetadata struct {
+	NamespaceDBsInfo []*namespaceDBInfo `json:"NamespaceDBsInfo"`
+}
+
```
cendhu:
```suggestion
type channelMetadata struct {
      ChannelName      string             `json:"ChannelName"`
	NamespaceDBsInfo []*namespaceDBInfo `json:"NamespaceDBsInfo"`
}
```
`ChannelName` filed helps in indireclty storing the channelName associated with a metadataDB. Otherwise, we can store the name explicty as well.

core/ledger/kvledger/txmgmt/statedb/statecouchdb/statecouchdb.go
```diff
@@ -167,16 +171,22 @@ func (provider *VersionedDBProvider) HealthCheck(ctx context.Context) error {
...
 type VersionedDB struct {
 	couchInstance      *couchInstance
-	metadataDB         *couchDatabase            // A database per channel to store metadata such as savepoint.
-	chainName          string                    // The name of the chain/channel.
-	namespaceDBs       map[string]*couchDatabase // One database per deployed chaincode.
-	committedDataCache *versionsCache            // Used as a local cache during bulk processing of a block.
+	metadataDB         *couchDatabase              // A database per channel to store metadata such as savepoint.
+	chainName          string                      // The name of the chain/channel.
+	namespaceDBs       map[string]*couchDatabase   // One database per namespace.
+	namespaceDBsInfo   map[string]*namespaceDBInfo // One namespaceDBInfo per namespace.
```
cendhu:
```suggestion
	namespaceDBs       map[string]*couchDatabase   // One database per namespace.
	channelMetadata    *channelMetadata .
```
`namespaceDBs` and `namespaceDBsInfo` look confusing.

core/ledger/kvledger/txmgmt/statedb/statecouchdb/statecouchdb.go
```diff
@@ -167,16 +171,22 @@ func (provider *VersionedDBProvider) HealthCheck(ctx context.Context) error {
...
 	mux                sync.RWMutex
 	redoLogger         *redoLogger
 	cache              *cache
 }
 
+type namespaceDBInfo struct {
+	Namespace string `json:"Namespace"`
+	DBName    string `json:"DBName"`
+}
+
```
cendhu: We can move `namespaceDBInfo` to the `couchdoc_conv.go` and place it next to the `channelMetadata`. This is because we store all types used to create a couchdoc in the `couchdoc_conv.go` and it is better to retain the consistency.

core/ledger/kvledger/txmgmt/statedb/statecouchdb/statecouchdb_test.go
```diff
@@ -1365,3 +1365,53 @@ func TestMissingRevisionRetrievalFromCache(t *testing.T) {
 	require.Equal(t, "rev1", revisions["key1"])
 	require.Equal(t, "rev2", revisions["key2"])
 }
+
+func TestGetAndUpdateNamespaceDBsInfo(t *testing.T) {
```
cendhu: We need to simulate peer failure scenarios in the test:
1. assume that the database is created and the peer fails before updating the channel metadata. When the peer comes up again and calls the `getNamespaceDBHandle()`, it should be reflected in the channel metadata.
2. a similar failure test for `newVersionedDB()` too.

cendhu (changes requested):
Thanks, @wenjianqiao for the changes. I have a few suggestions. 

Also, the current test ignores a few peer failure scenarios. I have specifically suggested a few in the code review. 

The following are other scenarios but I do not see any worrisome inconsistency issues (only a few minor ones which we need to keep in mind while updating peer rollback and adding snapshot code).  

1. A peer could fail after creating a database for a new channel but before committing the genesis block. When the peer comes online, the user might decide to make the peer not join the channel. In such a case, we might have an orphan database. If needed, we can remove orphan databases periodically but I assume that it is not a big issue for now. 
2. The namespace database is created at many stages:
     - **when the chaincode is deployed with indexes**: After the serializability validator, we invoke all the listeners. One of the listeners is the chaincode lifecycle mgmt and it calls `HandleDeployChaincode()` to deploy the index if the block has a deployment transaction. We create a database for the chaincode and collection namespace only when there is an index file associated with it. Otherwise, we don't. Assume that the peer fails or is killed by the user to run peer rollback or reset after creating a few new databases but before committing the block to the block store, we might have some extra databases that are not really associated with the channel as per the current block height. However, I assume that the peer rollback would just drop these extra databases and it isn't a problem for us to worry. 
    - **when the chaincode or collection is invoked first**: Assume that some chaincodes and collections do not have an index and hence, we wouldn't have created databases until the first access. As there is no database and no data associated with the chaincode/collection, ignoring them in peer rollback and snapshot isn't an issue. Further, if a database is empty, we might ignore it during the snapshot. Anyway, more thought would be given when we do CouchDB snapshot.

wenjianqiao:
@cendhu Thank you for your detailed comments. 
This PR was originally for deleting channel-specific data. Let's discuss the following at Monday's scrum to make sure it also meets the snapshot requirements.

1. Add ChannelName to channelMetadata
- This is not needed for dropping dbs. If needed for snapshot, we can add it.
2. Save channelMetadata before or after  createCouchDatabase
- The PR purposely saves channelMetadata BEFORE createCouchDatabase. If moving it after createCouchDatabase, in (rare) case peer crashes after the db is created and the user wants to unjoin the channel, the db name will be lost and cannot be deleted.
3. Add metadataDB to NamespaceDBsInfo
- When dropping dbs for a channel, we first delete all namespace dbs and at last metadataDB. Because if metadataDB is deleted before some namespace dbs and peer crashes, namespace db info will be lost. For this reason, the PR does not add metadataDB to NamespaceDBsInfo. Otherwise, we need to skip metadataDB until all other namespaceDBs are dropped.

core/ledger/kvledger/txmgmt/statedb/statecouchdb/statecouchdb.go
wenjianqiao: For 1, let's discuss the order offline. 
For 2, it is debatable. Personally I prefer the function is self-contained for this kind of check so that it doesn't require/depend on the caller's check.

core/ledger/kvledger/txmgmt/statedb/statecouchdb/statecouchdb.go
```diff
@@ -207,6 +219,14 @@ func newVersionedDB(couchInstance *couchInstance, redoLogger *redoLogger, dbName
...
 	}
 
+	dbsInfo, err := vdb.getLatestNamespaceDBsInfo()
+	if err != nil {
+		return nil, err
+	}
+	for _, dbInfo := range dbsInfo {
+		vdb.namespaceDBsInfo[dbInfo.Namespace] = dbInfo
+	}
+
```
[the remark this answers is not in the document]
wenjianqiao: Did you mean line 212? It has to be after VersionedDB is created.

cendhu:
> 1) Add ChannelName to channelMetadata
> This is not needed for dropping dbs. If needed for snapshot, we can add it.
> 3) Add metadataDB to NamespaceDBsInfo
> When dropping dbs for a channel, we first delete all namespace dbs and at last metadataDB. Because if metadataDB is deleted before some namespace dbs and peer crashes, namespace db info will be lost. For this reason, the PR does not add metadataDB to NamespaceDBsInfo. Otherwise, we need to skip metadataDB until all other namespaceDBs are dropped.

I don't think I have suggested 2. I have suggested 1 in this [comment](https://github.com/hyperledger/fabric/pull/1268#discussion_r426131028) which indirectly keeps the mapping between the metadataDB and channelName. Even if we go with 2, it is easy to delete the metadataDB last. 

Neither of _dropping databases_ and _snapshotting_ require the metadataDB to channelName mapping. However, some use-case that I have mentioned in this [comment](https://github.com/hyperledger/fabric/pull/1268#discussion_r426130096) does require it. I thought the goal of this PR is to preserve all the mapping between the trimmed name and the full name as we missed to do it in v1.0 and facing the trouble now. Once we preserve this mapping, we can easily enable the dropping of databases, snapshotting it, and also to satisfy future requirements. IMO, it is better to solve the problem for all cases at once rather than fixing it for one particular use-case. Anyway, I don't have a strong opinion on this and feel free to do what is best here. 
> 1. Save channelMetadata before or after createCouchDatabase
> The PR purposely saves channelMetadata BEFORE createCouchDatabase. If moving it after createCouchDatabase, in (rare) case peer crashes after the DB is created and the user wants to unjoin the channel, the db name will be lost and cannot be deleted.

If the peer crashes once the metadataDB is created, the genesis block itself would not have been committed. Hence, there is no peer unjoin. Further, the created metadataDB would become an orphan if the user decides to not join the channel again. There is certainly a tradeoff between inconsistent information in the channelMetadata and number of orphan databases and it is kind of complex as I mentioned in [comment](https://github.com/hyperledger/fabric/pull/1268#pullrequestreview-413057258). Addressing of consistency issue might be hard but it wouldn't create an issue for us I hope. If doing an update of channelMetadata before `createCouchDatabase()` can reduce the number of orphan databases while increasing the probability of inconsistency, I am fine with it as in the current form, it is hard to avoid inconsistency altogether unless we do a big refactoring or add some cleanup functions to handle panics and kill signal.

core/ledger/kvledger/txmgmt/statedb/statecouchdb/statecouchdb.go
```diff
@@ -197,6 +202,21 @@ func newVersionedDB(couchInstance *couchInstance, redoLogger *redoLogger, dbName
...
+	vdb.channelMetadata, err = vdb.getSavedChannelMetadata()
+	if err != nil {
+		return nil, err
+	}
+	if vdb.channelMetadata == nil {
+		vdb.channelMetadata = &channelMetadata{
+			ChannelName:      chainName,
+			NamespaceDBsInfo: make(map[string]*namespaceDBInfo),
+		}
+		if err = vdb.updateAndSaveNamespaceDBsInfo("", dbName); err != nil {
```
manish-sethi: This could lead to missing a mapping, as we allow the empty namespace as well (we write channel config in that). Do we really need to store the mapping for the channel metadata itself? We anyways, always need to know its name in order to get in to pull out the mapping.

manish-sethi (changes requested):
In addition to the in-place comments, can you please add a few more tests that can be added easily to verify that the errors are propagated.

core/ledger/kvledger/txmgmt/statedb/statecouchdb/statecouchdb.go
wenjianqiao: I don't understand why this could lead to missing a mapping. The empty namespace is same as metadataDB, which will not be missing since it has been added here. Agree we don't really need to store the metadataDB mapping. However, even if it is not added here, it will be added later when getNamespaceDBHandle is called with empty namespace.

manish-sethi: Unfortunately, there are two different functions for generating the names.
- For the channel metadatadb - [constructMetadataDBName](https://github.com/hyperledger/fabric/blob/e461141a2dfaa1214de60a47f3114c8b8b4805ef/core/ledger/kvledger/txmgmt/statedb/statecouchdb/couchdbutil.go#L186) 
- For the user namespace db - [constructNamespaceDBName](https://github.com/hyperledger/fabric/blob/e461141a2dfaa1214de60a47f3114c8b8b4805ef/core/ledger/kvledger/txmgmt/statedb/statecouchdb/couchdbutil.go#L205)

Logically, they can produce different outputs. So, if you always populate the entry in the map with an empty namespace with the output of the first function, what would happen when you call the second function with an empty namespace?

cendhu: The introduction of `channelName string` indirectly stores the mapping and hence, we don't need to store it explicitly. ~Though there won't be any collusion with the channelconfig namespace here, it adds a bit of confusion due to the usage of an empty namespace for metadataDBName.~ Further, the `getNamespaceDBHandle()` is not involved when accessing metadataDB.  I think the storage of metadataDBName mapping is forced here as the `updateAndSaveNamespaceDBsInfo()` has the following code
```
	if _, ok := vdb.channelMetadata.NamespaceDBsInfo[ns]; ok {
 		return nil
 	}

 	logger.Debugf("[%s] add namespaceDBInfo for namespace %s", vdb.chainName, ns)
 	vdb.channelMetadata.NamespaceDBsInfo[ns] = &namespaceDBInfo{
 		Namespace: ns,
 		DBName:    dbName,
 	}
```
If these line can be moved to `getNamespaceDBHandle()`, we can rename `updateAndSaveNamespaceDBsInfo()`  to `commitChannelMetadata()` and skip storing the mapping between metadatDBName and channel explicitly (given that channelName stores it implicitly). 

~Otherwise, we need to avoid this confusion using a comment if you are not comfortable in modifying the `updateAndSaveNamespaceDBsInfo()`~

cendhu: Ah. Correcting myself. A collision does happen on the namespace but not on the dbName.

wenjianqiao: Thank you @manish-sethi and @cendhu. Will remove metadataDB from mapping. 
A side note for metadataDB and namespaceDB. Based on existing implementation, there is dbName collision between metadataDB and namespaceDB when namespace is an empty string. However if channel name is too long and truncated, their dbNames are not same any more.

cendhu: I see. Nice find @wenjianqiao So the channelconfig block could be either in the metadataDB or in a separate DB based on the length of the channel name. We may have to live with this non-deterministic behaviour for now. Please add a comment stating this behaviour and create a JIRA task too. As part of upgradeDB or rebuildDB commands, we can think of addressing this issue. We can modify [constructMetadataDBName](https://github.com/hyperledger/fabric/blob/e461141a2dfaa1214de60a47f3114c8b8b4805ef/core/ledger/kvledger/txmgmt/statedb/statecouchdb/couchdbutil.go#L186) to return a name that cannot collide with empty namespace's dbName  and migrate the content from existing metadataDB to newMetadataDB in the `newVersionedDB() `

wenjianqiao:
@manish-sethi @cendhu I have addressed all the comments.

manish-sethi (approved):

PR #1282, opened 2020-05-17
[DocUpdate] basic chaincode package tutorial

davidkhala:
#### Type of change

- Documentation update

#### Description

- Remote the sentence
"It is not necessary for organizations to use the same package
label."
- speak out the source tar name required in package
- prune Chaincode-Package-Metadata legacy

#### Additional details

The point especially looking for your feedback is what is the meaning of
"It is not necessary for organizations to use the same package label"

This give me an impression that even we use different label, fabric could smartly detect two package installed on different orgs are same chaincode, meanwhile it suggest label is optional. 

But actually it is not the case. It make reader confused.

### DIFF
--- a/docs/source/chaincode_lifecycle.md
+++ b/docs/source/chaincode_lifecycle.md
@@ -82,7 +82,7 @@ automatically create a file in this format.
 - The chaincode needs to be packaged in a tar file, ending with a `.tar.gz` file
   extension.
 - The tar file needs to contain two files (no directory): a metadata file
-  "metadata.json" and another tar containing the chaincode files.
+  "metadata.json" and another tar "code.tar.gz" containing the chaincode files.
 - "metadata.json" contains JSON that specifies the
   chaincode language, code path, and package label. You can see an example of
   a metadata file below:


docs/source/chaincode_lifecycle.md
```diff
@@ -94,8 +94,7 @@ automatically create a file in this format.
 
 *The chaincode is packaged separately by Org1 and Org2. Both organizations use
 MYCC_1 as their package label in order to identify the package using the name
-and version. It is not necessary for organizations to use the same package
```
denyeart: Different orgs can indeed use different labels, and even different chaincodes. An organization associates a chaincode package with a chaincode when they do approveformyorg lifecycle step.

denyeart (changes requested):

docs/source/chaincode_lifecycle.md
davidkhala: Cool! it seems packageID linking up with chaincodeID is established during approveformyorg
I will revert this line
Thanks for your clarification, Dave,

denyeart (approved):

PR #1304, opened 2020-05-20
[FAB-17059]: Add missing mspID when init MembershipProvider

hello2mao:
#### Type of change

- Bug fix

#### Description

Add missing mspID when init MembershipProvider.
Otherwise, the mspid match check will not work as expect.

https://github.com/hyperledger/fabric/blob/e9c4f2858f8678d1a80d190595eb90b6267ffc72/core/common/privdata/membershipinfo.go#L38-L44

#### Additional details

NA

#### Related issues

See [FAB-17059](https://jira.hyperledger.org/browse/FAB-17059)

### DIFF
--- a/core/common/privdata/membershipinfo_test.go
+++ b/core/common/privdata/membershipinfo_test.go
@@ -23,18 +23,20 @@ func TestMembershipInfoProvider(t *testing.T) {
 		Signature: []byte{1, 2, 3},
 		Data:      []byte{4, 5, 6},
 	}
+	emptyPeerSelfSignedData := protoutil.SignedData{}
 
 	identityDeserializer := func(chainID string) msp.IdentityDeserializer {
 		return &mockDeserializer{}
 	}
 
-	// verify membership provider returns true
-	membershipProvider := NewMembershipInfoProvider(mspID, peerSelfSignedData, identityDeserializer)
+	// verify membership provider pass simple check returns true
+	membershipProvider := NewMembershipInfoProvider(mspID, emptyPeerSelfSignedData, identityDeserializer)
 	res, err := membershipProvider.AmMemberOf("test1", getAccessPolicy([]string{"peer0", "peer1"}))
 	assert.True(t, res)
 	assert.Nil(t, err)
 
-	// verify membership provider returns false
+	// verify membership provider fall back to default access policy evaluation returns false
+	membershipProvider = NewMembershipInfoProvider(mspID, peerSelfSignedData, identityDeserializer)
 	res, err = membershipProvider.AmMemberOf("test1", getAccessPolicy([]string{"peer2", "peer3"}))
 	assert.False(t, res)
 	assert.Nil(t, err)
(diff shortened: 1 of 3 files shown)


mastersingh24 (dismissed):
This looks right to me.  @caod123 ?

sykesm (changes requested):
The production code change looks fine but there should be a corresponding test update.

> Otherwise, the mspid match check will not work as expect.

That means we should have a failing test case.

yacovm:
> The production code change looks fine but there should be a corresponding test update.
> 
> > Otherwise, the mspid match check will not work as expect.
> 
> That means we should have a failing test case.

+1 let's please change the tests to reflect it, preferrably integration tests 

> The production code change looks fine but there should be a corresponding test update.
> 
> > Otherwise, the mspid match check will not work as expect.
> 
> That means we should have a failing test case.

My thoughts exactly. 

Seems like the integration test introduced in [this PR](https://github.com/hyperledger/fabric/pull/626/commits/504c27431c20c4a10165c2d274d5fee388b3706a) passes when it shouldn't have passed.

hello2mao:
@sykesm @yacovm 
Sorry. It's a little difficult for me to fix or add a new integration test.
Maybe @caod123 can help me with that. ^_^.

caod123:
I'll take a look and see if I can fix the integration test when I get time, opened https://jira.hyperledger.org/browse/FAB-17922 to track in the meantime. On that note there probably is still some improvement that could be made to the unit test here to check the case when mspID is passed in as empty (or rather any value that would cause the production code to fall back to the default membership evaluation logic) as well as an update test to verify it doesn't fall back to the default evaluation logic if the mspID passed in is in fact a member org.

caod123:
Ah I remember now, the integration tests actually pass because there's two paths that utilize the membership check. The `MembershipInfoProvider` here isn't actually used as part of private data dissemination/reconciliation hence why the integration test properly passes. For that logic it actually utilizes the mspID passed down to the [coordinator/privatedataprovider](https://github.com/hyperledger/fabric/blob/master/gossip/privdata/pvtdataprovider.go#L128-L137). If I remember correctly the `MembershipInfoProvider` path that this bug is referring to is actually only used for when [collection eligibility changes](https://github.com/hyperledger/fabric/blob/master/core/ledger/kvledger/coll_elg_notifier.go#L130-L137). There never was an integration test for this path because it's somewhat hard to test this. I think updating the unit tests as I suggested above might be enough to cover the testing requirement for this fix.

caod123:
@hello2mao do you think you could add the unit tests for the above?

hello2mao:
> @hello2mao do you think you could add the unit tests for the above?

@caod123 Thanks.  I've updated unit tests to verify these two case.

core/common/privdata/membershipinfo_test.go
```diff
@@ -23,18 +23,25 @@ func TestMembershipInfoProvider(t *testing.T) {
...
-	membershipProvider := NewMembershipInfoProvider(mspID, peerSelfSignedData, identityDeserializer)
+	// verify membership provider pass simple check returns true
+	membershipProvider := NewMembershipInfoProvider(mspID, emptyPeerSelfSignedData, identityDeserializer)
 	res, err := membershipProvider.AmMemberOf("test1", getAccessPolicy([]string{"peer0", "peer1"}))
 	assert.True(t, res)
 	assert.Nil(t, err)
 
-	// verify membership provider returns false
+	// verify membership provider fall back to default access policy evaluation returns true
+	membershipProvider = NewMembershipInfoProvider(mspID, peerSelfSignedData, identityDeserializer)
```
caod123: I'm not sure if this will actually fall back to the default access policy since the mspID (`peer0` in this case) should satisfy the member orgs of the policy on line 40. You may be able to just remove this specific case since you already test further below when passing an empty mspID it falls back and evaluates to true.

caod123 (changes requested):

core/common/privdata/membershipinfo_test.go
hello2mao: Remove this case. Thanks.

caod123 (approved):

sykesm (approved):

sykesm:
Thank you for the test updates.

PR #1305, opened 2020-05-20
FAB-17840 Ch.Part.API: Join channel REST handler

tock-ibm:
Implement a handler for a POST request to join a channel.

Here we support requests carrying application/json, support
for additional Content-Type will be added later.

Change-Id: I8d09e3cba09842f2adc47fb60161eee814e33d31

#### Type of change

- New feature
#### Related issues

Task: FAB-17840
Story: FAB-15710
Epic: FAB-17712


lindluni:
The command is `/ci-run`, sorry missed your question

orderer/common/channelparticipation/restapi.go
```diff
@@ -127,17 +132,67 @@ func (h *HTTPHandler) serveListOne(resp http.ResponseWriter, req *http.Request)
...
+	switch err {
+	case types.ErrSystemChannelExists:
+		h.sendResponseNotAllowed(resp, errors.Wrap(err, "cannot join"), http.MethodGet)
+	case types.ErrChannelAlreadyExists:
+		h.sendResponseNotAllowed(resp, errors.Wrap(err, "cannot join"), http.MethodGet, http.MethodDelete)
+	case types.ErrAppChannelsAlreadyExists:
+		h.sendResponseJsonError(resp, http.StatusForbidden, errors.Wrap(err, "cannot join"))
+	default:
+		h.sendResponseJsonError(resp, http.StatusBadRequest, errors.Wrap(err, "cannot join"))
+	}
```
C0rWin, line 246: I think that since you treat `ErrSystemChannelExists` And `ErrChannelAlreadyExists` you might want to consolidate it into

```
   case types.ErrSystemChannelExists, types.ErrChannelAlreadyExists: {
   } 
```

Also I think that in general it might be good to extract error treatment into separate function.

orderer/common/types/content.go
```diff
@@ -0,0 +1,14 @@
+/*
+Copyright IBM Corp. All Rights Reserved.
+
+SPDX-License-Identifier: Apache-2.0
+*/
+
+package types
+
+// JoinBody is encoded in the body of a Join POST request when Content-Type is application/json.
+type JoinBody struct {
```
C0rWin: Why do you need this?

orderer/common/channelparticipation/restapi.go
```diff
@@ -127,17 +132,67 @@ func (h *HTTPHandler) serveListOne(resp http.ResponseWriter, req *http.Request)
...
+
+	bodyBytes, err := ioutil.ReadAll(req.Body)
+	if err != nil {
+		h.logger.Debugf("Failed to read request body: %s", err)
+		h.sendResponseJsonError(resp, http.StatusBadRequest, errors.Wrap(err, "cannot read request body"))
+		return
+	}
+
+	body := types.JoinBody{}
+	err = json.Unmarshal(bodyBytes, &body)
```
C0rWin: That assumes you augment config block into json, while multipart content type allows you just to insert the binary without it. For instance in the way you suggesting to do it, it will be very hard to utilize curl or wget for instance to submit join channel request.

C0rWin (changes requested):

orderer/common/types/content.go
tock-ibm: This task implements POST with a body of type application/json, with structure:
```
{
   configBlock: "block content (binary, after marshaling to protobuf) in base64"
}
```
this structure helps marshal and unmarshal a []byte slice that contains the block bytes.

Support for Content-type application/octet-stream and multipart/form-data will be added later. The plan is to support all 3 types, see:
https://jira.hyperledger.org/browse/FAB-17891

I am starting with this one because it is easier to unit-test & integration-test (IMO), but eventually all 3 will be supported, as planed.

orderer/common/channelparticipation/restapi.go
tock-ibm: You are correct. Adding Content-type application/octet-stream and multipart/form-data will be added later. The plan is to support all 3 types, see:
https://jira.hyperledger.org/browse/FAB-17891

orderer/common/channelparticipation/restapi.go
tock-ibm: They are somewhat different, as the "Allow" header in the response is different:
* In the `ErrSystemChannelExists` case you can only GET the channel (whether it exists or not is not checked here).
* In the `ErrChannelAlreadyExists` case, the system channel does not exist, but the channel itself does. This means that what you can do on the channel is GET & DELETE.

```go
switch err {
	case types.ErrSystemChannelExists:
		h.sendResponseNotAllowed(resp, errors.Wrap(err, "cannot join"), 
                http.MethodGet)
	case types.ErrChannelAlreadyExists:
		h.sendResponseNotAllowed(resp, errors.Wrap(err, "cannot join"), 
                http.MethodGet, http.MethodDelete)
	case types.ErrAppChannelsAlreadyExists:
		h.sendResponseJsonError(resp, http.StatusForbidden, errors.Wrap(err, "cannot join"))
	default:
		h.sendResponseJsonError(resp, http.StatusBadRequest, errors.Wrap(err, "cannot join"))
	}
```

I will restructure the error handling to a separate function.

orderer/common/types/content.go
C0rWin: I do not think there is a need in augmenting binary into json as I also do not see particular reason to support all three.

tock-ibm: I added support for multipart/form-data.

C0rWin: in such case you do not need this `JoinBody` struct, right?

tock-ibm: I would like to keep both `multipart/form-data` & `application/json`.

C0rWin: I do not think we need to introduce two methods to achieve same goal, hence I think we better to start with one option only.

tock-ibm: OK. It is easy enough to add if someone needs it.

C0rWin (approved):

PR #1311, opened 2020-05-20
Fix OSN leader election failure

yeasy:
#### Type of change

- Bug fix

#### Description

Currently in fabric 1.4/2.x with raft consensus, when one OSN is deleted
from channel and rejoin later, it will be assigned a new raft id.
However, in some cases including that catchup is triggered or the OSN
is stopped during the re-configuration, the OSN still uses the old raft
id.

Other OSNs including the leader are using the latest view and believe
that the rejoined OSN will use a new raft id.  This will never be
recoved and may fail the leader election due to qurum lost. No new
transaction will be accepted, and no new configuration can be changed.

This patchset fixes the issue by halting the chain proactively and let
it obtain the correct raft id after tracking the latest config block.

Change-Id: I6e352b59e9feba85f2bbc0119b9eb47bce746adc

#### Additional details

Since this is a critical issue that will block the service, need to backport to v1.4 branch too.

#### Related issues

https://jira.hyperledger.org/browse/FAB-17875

### DIFF
--- a/integration/raft/config_test.go
+++ b/integration/raft/config_test.go
@@ -989,7 +989,7 @@ var _ = Describe("EndToEnd reconfiguration and onboarding", func() {
 
 			By("Launching the orderers")
 			for _, o := range orderers {
-				runner := network.OrdererRunner(o)
+				runner := network.OrdererRunner(o, "FABRIC_LOGGING_SPEC=orderer.consensus.etcdraft=debug:info")
 				ordererRunners = append(ordererRunners, runner)
 				process := ifrit.Invoke(runner)
 				ordererProcesses = append(ordererProcesses, process)
@@ -1076,6 +1076,80 @@ var _ = Describe("EndToEnd reconfiguration and onboarding", func() {
 			resp, err := nwo.Broadcast(network, orderers[survivor], env)
 			Expect(err).NotTo(HaveOccurred())
 			Expect(resp.Status).To(Equal(common.Status_SUCCESS))
+		})
+
+		When("an evicted node is added back while it's offline", func() {
+			It("can start with correct new raft id", func() {
+				o1 := network.Orderer("orderer1")
+				o2 := network.Orderer("orderer2")
+				o3 := network.Orderer("orderer3")
+
+				By("Waiting for them to elect a leader")
+				findLeader(ordererRunners)
+
+				By("Creating an application channel testchannel")
+				network.CreateChannel("testchannel", o1, peer)
+
+				assertBlockReception(map[string]int{
+					"testchannel":   0,
+					"systemchannel": 1,
+				}, []*nwo.Orderer{o1, o2, o3}, peer, network)
+
+				By("Killing the first orderer")
+				ordererProcesses[0].Signal(syscall.SIGTERM)
+				Eventually(ordererProcesses[0].Wait(), network.EventuallyTimeout).Should(Receive())
+
+				// We need to wait for stabilization, as we might have killed the leader OSN.
+				By("Waiting for the channel to stabilize after killing the orderer")
+				assertBlockReception(map[string]int{
+					"testchannel":   0,
+					"systemchannel": 1,
+				}, []*nwo.Orderer{o2, o3}, peer, network)
+
+				By("observing active nodes to shrink")
+				o2Runner := ordererRunners[1]
+				Eventually(o2Runner.Err(), network.EventuallyTimeout).Should(gbytes.Say("Current active nodes in cluster are: \\[2 3\\]"))
+
+				By("Removing the first orderer from the application channel")
+				server1CertBytes, err := ioutil.ReadFile(filepath.Join(network.OrdererLocalTLSDir(o1), "server.crt"))
+				Expect(err).To(Not(HaveOccurred()))
+				removeConsenter(network, peer, o2, "testchannel", server1CertBytes)
+
+				By("Adding the evicted orderer back to the application channel")
+				addConsenter(network, peer, o2, "testchannel", etcdraft.Consenter{
+					ServerTlsCert: server1CertBytes,
+					ClientTlsCert: server1CertBytes,
+					Host:          "127.0.0.1",
+					Port:          uint32(network.OrdererPort(o1, nwo.ClusterPort)),
+				})
+
+				By("Removing the first orderer from the application channel again")
+				removeConsenter(network, peer, o2, "testchannel", server1CertBytes)
+
+				By("Adding the evicted orderer back to the application channel again")
+				addConsenter(network, peer, o2, "testchannel", etcdraft.Consenter{
+					ServerTlsCert: server1CertBytes,
+					ClientTlsCert: server1CertBytes,
+					Host:          "127.0.0.1",
+					Port:          uint32(network.OrdererPort(o1, nwo.ClusterPort)),
+				})
+
+				By("Starting the first orderer again")
+				o1Runner := network.OrdererRunner(o1)
+				ordererProcesses[0] = ifrit.Invoke(o1Runner)
+				Eventually(ordererProcesses[0].Ready(), network.EventuallyTimeout).Should(BeClosed())
+
+				By("Submitting tx")
+				env := CreateBroadcastEnvelope(network, o2, "testchannel", []byte("foo"))
+				resp, err := nwo.Broadcast(network, o2, env)
+				Expect(err).NotTo(HaveOccurred())
+				Expect(resp.Status).To(Equal(common.Status_SUCCESS))
+
+				By("Waiting for the channel to stabilize")
+				assertBlockReception(map[string]int{
+					"testchannel": 5,
+				}, []*nwo.Orderer{o1, o2, o3}, peer, network)
+			})
 		})
 
 		It("notices it even if it is down at the time of its eviction", func() {
--- a/orderer/consensus/etcdraft/chain.go
+++ b/orderer/consensus/etcdraft/chain.go
@@ -460,6 +460,12 @@ func (c *Chain) Consensus(req *orderer.ConsensusRequest, sender uint64) error {
 		return fmt.Errorf("failed to unmarshal StepRequest payload to Raft Message: %s", err)
 	}
 
+	if stepMsg.To != c.raftID {
+		c.logger.Warnf("Received msg to %d, my ID is probably wrong due to out of date, cowardly halting", stepMsg.To)
+		c.Halt()
+		return nil
+	}
+
 	if err := c.Node.Step(context.TODO(), *stepMsg); err != nil {
 		return fmt.Errorf("failed to process Raft Step message: %s", err)
 	}
--- a/orderer/consensus/etcdraft/chain_test.go
+++ b/orderer/consensus/etcdraft/chain_test.go
@@ -193,7 +193,7 @@ var _ = Describe("Chain", func() {
 
 		campaign := func(c *etcdraft.Chain, observeC <-chan raft.SoftState) {
 			Eventually(func() <-chan raft.SoftState {
-				c.Consensus(&orderer.ConsensusRequest{Payload: protoutil.MarshalOrPanic(&raftpb.Message{Type: raftpb.MsgTimeoutNow})}, 0)
+				c.Consensus(&orderer.ConsensusRequest{Payload: protoutil.MarshalOrPanic(&raftpb.Message{Type: raftpb.MsgTimeoutNow, To: 1})}, 0)
 				return observeC
 			}, LongEventualTimeout).Should(Receive(StateEqual(1, raft.StateLeader)))
 		}
@@ -3749,7 +3749,7 @@ func (n *network) elect(id uint64) {
 
 	// Send node an artificial MsgTimeoutNow to emulate leadership transfer.
 	fmt.Fprintf(GinkgoWriter, "Send artificial MsgTimeoutNow to elect node %d\n", id)
-	candidate.Consensus(&orderer.ConsensusRequest{Payload: protoutil.MarshalOrPanic(&raftpb.Message{Type: raftpb.MsgTimeoutNow})}, 0)
+	candidate.Consensus(&orderer.ConsensusRequest{Payload: protoutil.MarshalOrPanic(&raftpb.Message{Type: raftpb.MsgTimeoutNow, To: id})}, 0)
 	Eventually(candidate.observe, LongEventualTimeout).Should(Receive(StateEqual(id, raft.StateLeader)))
 
 	n.Lock()


yacovm (dismissed):
1. Where is the integration test that reproduces the bug? My guess is that you reproduced this locally without giving the removed node the chance to detect its removal. Am I right? 
2. The code is taking what I think is the wrong approach. I don't think the chain should mutate its identifier during its runtime. When Raft detects it left the channel, it shuts its chain down, and begins monitoring for re-addition. Until that moment you should not remove it from the channel, as this would cause what you describe. Therefore, if you wish to re-introduce that orderer to the channel you need to wait until it recognizes it's not in the channel and only afterwards add it. This will ensure it will create the chain instance properly with the right ID, and the previous chain instance with the old ID would terminate prior to that.

yeasy:
> 1. Where is the integration test that reproduces the bug? My guess is that you reproduced this locally without giving the removed node the chance to detect its removal. Am I right?

yacov, currently, there is no integration testing covering such remove and rejoin scenario.
It can be reproduced easily. I do not know how to achieve what you said, because there is no API to let a node detect its removal. There is eviction suspicion, however, cannot block the issue's happening.

> 2. The code is taking what I think is the wrong approach. I don't think the chain should mutate its identifier during its runtime. When Raft detects it left the channel, it shuts its chain down, and begins monitoring for re-addition. Until that moment you should not remove it from the channel, as this would cause what you describe. Therefore, if you wish to re-introduce that orderer to the channel you need to wait until it recognizes it's not in the channel and only afterwards add it. This will ensure it will create the chain instance properly with the right ID, and the previous chain instance with the old ID would terminate prior to that.

That's only for very ideal cases. 
However, fabric's code does not guarantee a "Raft detects it left the channel, it shuts its chain down, and begins monitoring for re-addition.". And users will not know how long they should "wait until it recognizes it's not in the channel and only afterwards add it.". 
In this case, the problem will happen. If a problem can be triggered by users repeatedly,  it makes sense to consider to improve the software?

yacovm:
> yacov, currently, there is no integration testing covering such remove and rejoin scenario.
> It can be reproduced easily. I do not know how to achieve what you said, because there is no API to let a node detect its removal. There is eviction suspicion, however, cannot block the issue's happening.

We actually have tests that test that nodes detect their eviction, take a look [here](https://github.com/hyperledger/fabric/blob/5b63cef9a7a82775a58f5f530199b99a06750e3e/integration/raft/config_test.go#L1130) for an example. We also have tests that remove a node and then [later on re-add it](https://github.com/hyperledger/fabric/blob/5b63cef9a7a82775a58f5f530199b99a06750e3e/integration/raft/config_test.go#L1059-L1066). 
You claim there is a bug that can be reproduced easily, therefore we should make an integration test to ensure:
1. There is a way to actually make the bug manifest
2. The production code changes fix the bug.
3. Further production code changes will not re-introduce the bug.

> However, fabric's code does not guarantee a "Raft detects it left the channel, it shuts its chain down, and begins monitoring for re-addition."

It does guarantee it, we have [tests that test it](https://github.com/hyperledger/fabric/blob/5b63cef9a7a82775a58f5f530199b99a06750e3e/integration/raft/config_test.go#L1078-L1165). 

> And users will not know how long they should "wait until it recognizes it's not in the channel and only afterwards add it.".

If you want to improve the documentation, by all means please do so. We always welcome documentation improvements.

> In this case, the problem will happen. If a problem can be triggered by users repeatedly, it makes sense to consider to improve the software?

Improving software is a constant trade-off between risk and benefit. 
The Raft chain has a ton of state (fields) and maintaining their lifecycle is a delicate thing.
By making the chain reconfigure sensitive fields you introduce risk, which I think outweighs the benefit. 

> There is eviction suspicion, however, cannot block the issue's happening.

Eviction suspicion eventually confirms eviction or confirms non-eviction.

yeasy:
> tests that test it

Hi Yacov
If I understand it correctly, your suggestion here is to add some integration testing?  If so, we can think about adding the testing case later.

orderer/consensus/etcdraft/chain.go
```diff
@@ -921,10 +921,72 @@ func (c *Chain) catchUp(snap *raftpb.Snapshot) error {
...
+			c.logger.Panicf("Failed to configure communication: %s", err)
+		}
+	}
+
 	c.logger.Infof("Finished syncing with cluster up to and including block [%d]", b.Header.Number)
 	return nil
 }
 
+// detectSelfID will try to find the current osn's raft id from the given consenters.
+func (c *Chain) detectSelfID(consenters map[uint64]*etcdraft.Consenter) (uint64, error) {
```
yacovm: Why are you copying the [method](https://github.com/hyperledger/fabric/blob/master/orderer/consensus/etcdraft/consenter.go#L106-L128) from the consenter? You can instead refactor the code to a non-instance method and pass the logger as a parameter.

yeasy: Well, that's not to break the consenter part. 
But It's OK, and I can change the code to move this dtectSelfID() method into some utils category. How about that?

yacovm:
> > tests that test it
> 
> Hi Yacov
> If I understand it correctly, your suggestion here is to add some integration testing? If so, we can think about adding the testing case later.

Sure, you may do it at your leisure, no rush. 
However, should this pull request be merged, it will have the integration test in it.

orderer/consensus/etcdraft/chain.go
yeasy: Besides, the current Chain struct cannot restart with new consensus config, and cannot re-apply its config once created. That's also part of reasons we need to carefully update these config items at runtime. The other reason is to improve performance, because a recreation of chain always take more time than updating its parameters.

yeasy:
> > > tests that test it
> > 
> > 
> > Hi Yacov
> > If I understand it correctly, your suggestion here is to add some integration testing? If so, we can think about adding the testing case later.
> 
> Sure, you may do it at your leisure, no rush.
> However, should this pull request be merged, it will have the integration test in it.

OK, will do it later.

orderer/consensus/etcdraft/chain.go
yacovm: >  I can change the code to move this dtectSelfID() method into some utils category. How about that?

Sounds good.

orderer/consensus/etcdraft/chain.go
```diff
@@ -921,6 +921,40 @@ func (c *Chain) catchUp(snap *raftpb.Snapshot) error {
 		next++
 	}
 
+	if memberChanged { // There is a membership change, need to re-calculate local raft id and refresh communications
+		id, err := detectSelfID(c.opts.Consenters, c.opts.Cert, c.logger)
+		if err != nil { // means i'm not in the channel, halt it
```
yacovm: I still don't understand what we're trying to accomplish here.

1. If all you want is to rotate a certificate of a consenter, you can do it without changing the ID via removing the certificate and adding a new one like demonstrated [here](https://github.com/hyperledger/fabric/blob/master/integration/raft/config_test.go#L441-L578). In such a case, Raft will assign [the old ID](https://github.com/hyperledger/fabric/blob/e4c1fccb442a42ad1d3c9a603d1c9b4fec8c7984/orderer/consensus/etcdraft/membership.go#L81-L86) to the consenter with the rotated certificate.
2. If you want to remove a node then the chain will halt anyway, and you will never reach this code path unless you remove the node, it will not detect its removal and then you change your mind fast enough and re-add it again somehow. 

Since config changes require out of band signature assembly, it's unlikely that this scenario I think you are trying to make available (removal and instant re-addition) will be required, so I don't think the benefits of this change are more than the risk involved in having it.

jyellick:
If I can summarize my impression of this bug and fix.

**Bug:** If an OSN is removed from the consenter set, and, because he is not the leader he does not get the block which triggers his eviction.  Then, the network progresses some amount, and, the same OSN is added back.  This OSN now begins receiving Raft messages again, and detects that the network has moved beyond the point that simply playing the log forward is possible, and instead of committing the blocks serially (and realizing its been evicted), instead does a state transfer, looks at the latest block and says "great, we were reconfigured, but I'm still in the membership set".  Now, the node is operating with the wrong Raft ID and therefore isn't really participating in consensus.

**Question:**  Does the OSN continue to commit blocks at this point? Does the OSN recompute its ID properly if restarted?

**Fix:** I do think that if this is a bug, and I think that it is, that it should be fixed.  But, I think I agree with @yacovm, it is dangerous to be tweaking internal state like node IDs dynamically.  Instead, we should ensure that the Raft component shuts down entirely, and is restarted with the correct ID.  @guoger what do you think?

yeasy:
> If I can summarize my impression of this bug and fix.
> 
> **Bug:** If an OSN is removed from the consenter set, and, because he is not the leader he does not get the block which triggers his eviction. Then, the network progresses some amount, and, the same OSN is added back. This OSN now begins receiving Raft messages again, and detects that the network has moved beyond the point that simply playing the log forward is possible, and instead of committing the blocks serially (and realizing its been evicted), instead does a state transfer, looks at the latest block and says "great, we were reconfigured, but I'm still in the membership set". Now, the node is operating with the wrong Raft ID and therefore isn't really participating in consensus.
> 
> **Question:** Does the OSN continue to commit blocks at this point? Does the OSN recompute its ID properly if restarted?
> 
> **Fix:** I do think that if this is a bug, and I think that it is, that it should be fixed. But, I think I agree with @yacovm, it is dangerous to be tweaking internal state like node IDs dynamically. Instead, we should ensure that the Raft component shuts down entirely, and is restarted with the correct ID. @guoger what do you think?

@jyellick 

Yes, that is a bug, and we've repeated it many times locally. I'm working on creating a ginkgo based testing case, but seems it lacks some API to trigger the catchup(). It will be helpful if you know how to do that.

I'm fine with the solution that we restart the chain, but as i said, it will take more time comparing with just updating the related configuration items. It is not that dangerous during this catchup period, because at that time, the OSN cannot be a leader, and will not receive new msg. Make sense?

guoger (changes requested):
I do think this is a bug (thanks for catching it!), however simply recomputing raft id on-the-fly does not solve the problem.

What happens here is that when re-added node starts with stale ID, it does *not* catch up with leader via normal raft replication, because its ID is not among active nodes tracked by leader. On the other hand, since it still recognizes leader by receiving heartbeats, suspicion thread and forced catchup is not triggered. Therefore, it gets stuck in limbo.

We could:
- make suspicion more aggressive (even if a node is not leaderless, we still perform `confirmSuspicion` periodically, AND
- halts the chain if it finds itself evicted by a config block during `catchUp` [here](https://github.com/hyperledger/fabric/blob/c799dcc188d76c5d3f3e2cdab931f30a00273d17/orderer/consensus/etcdraft/chain.go#L900)

I have a working prototype locally and l'll demo it to you @yeasy

yacovm:
> I do think this is a bug (thanks for catching it!), however simply recomputing raft id on-the-fly does not solve the problem.
> 
> What happens here is that when re-added node starts with stale ID, it does _not_ catch up with leader via normal raft replication, because its ID is not among active nodes tracked by leader. On the other hand, since it still recognizes leader by receiving heartbeats, suspicion thread and forced catchup is not triggered. Therefore, it gets stuck in limbo.
> 
> We could:
> 
>     * make suspicion more aggressive (even if a node is not leaderless, we still perform `confirmSuspicion` periodically, AND
> 
>     * halts the chain if it finds itself evicted by a config block during `catchUp` [here](https://github.com/hyperledger/fabric/blob/c799dcc188d76c5d3f3e2cdab931f30a00273d17/orderer/consensus/etcdraft/chain.go#L900)
> 
> 
> I have a working prototype locally and l'll demo it to you @yeasy

If the leader knows the new ID of the node and it sends it heartbeats, then we can look at the ID of the node that leader tells us and if it's not our current ID then we can halt the chain and then forcely synchronize ourselves and restart the chain again somehow... @guoger  wdyt?

```
	if stepMsg.Type == raftpb.MsgHeartbeat {
		if c.raftID != stepMsg.To {
			c.HandleExoticBug()
		}
	}
```

guoger:
> > I do think this is a bug (thanks for catching it!), however simply recomputing raft id on-the-fly does not solve the problem.
> > What happens here is that when re-added node starts with stale ID, it does _not_ catch up with leader via normal raft replication, because its ID is not among active nodes tracked by leader. On the other hand, since it still recognizes leader by receiving heartbeats, suspicion thread and forced catchup is not triggered. Therefore, it gets stuck in limbo.
> > We could:
> > ```
> > * make suspicion more aggressive (even if a node is not leaderless, we still perform `confirmSuspicion` periodically, AND
> > 
> > * halts the chain if it finds itself evicted by a config block during `catchUp` [here](https://github.com/hyperledger/fabric/blob/c799dcc188d76c5d3f3e2cdab931f30a00273d17/orderer/consensus/etcdraft/chain.go#L900)
> > ```
> > 
> > 
> > I have a working prototype locally and l'll demo it to you @yeasy
> 
> If the leader knows the new ID of the node and it sends it heartbeats, then we can look at the ID of the node that leader tells us and if it's not our current ID then we can halt the chain and then forcely synchronize ourselves and restart the chain again somehow... @guoger wdyt?
> 
> ```
> 	if stepMsg.Type == raftpb.MsgHeartbeat {
> 		if c.raftID != stepMsg.To {
> 			c.HandleExoticBug()
> 		}
> 	}
> ```

a slightly different variation i tried was adding this to `chain.Consensus`:
```
	if stepMsg.To != c.raftID {
		c.logger.Errorf("Received msg to %d, my ID is probably wrong, cowardly halting", stepMsg.To)
		c.Halt()
		return nil
	}
```

however, i was concerned that halting the chain is too aggressive a reaction to such a message. Admittedly the code is much simpler, we will need a lengthy comment to explain the reason.

Anyway, i'm ok with either

yacovm:
> however, i was concerned that halting the chain is too aggressive a reaction to such a message

In Raft, we assume the leader is honest and correct. If this assumption breaks, Raft breaks apart.


>  make suspicion more aggressive (even if a node is not leaderless, we still perform confirmSuspicion periodically, AND halts the chain if it finds itself evicted by a config block during catchUp here

I'm not OK with doing eviction suspicion periodically, when everything is fine, just to fix a scenario that occurs due to incorrect use of Raft. 
This scenario is avoided if the user doesn't do something against his best interest. 
We should not change the normal common case just to cater to a rare failure case.

I suggest we proceed by halting the chain if we got a heartbeat with a **To** that is different than our identifier, and make an integration test that reproduces this as part of the change set.

guoger:
Sounds good to me. @yeasy [here](https://github.com/guoger/fabric/commit/c60d7548b6baf99ffba33cd22b68f0def70a2ec3)'s a quick draft of what i had in mind. If you'd like, you could start from there and work out a proper PR.

orderer/consensus/etcdraft/chain.go
```diff
@@ -921,6 +921,40 @@ func (c *Chain) catchUp(snap *raftpb.Snapshot) error {
...
+			c.opts.RaftID = id
+
+			c.Node.logger = lg
+			c.Node.config.ID = id
+			c.Node.config.Logger = lg
+			c.Node.metadata = c.opts.BlockMetadata
+			c.logger.Debugf("Updated to new ID=%d, will restart the raft.Node", c.channelID, id)
+
+			c.Node.Node.Stop()
+			c.Node.Node = raft.RestartNode(c.Node.config)
```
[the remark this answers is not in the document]
yeasy: @guoger 
Thanks, and during the catchup processing, the c.Node.Node should be OK as the OSN will not handle any other message. Am I correct?

yeasy:
> here

Sure, let me have a look and work with you on an appropriate patchset.

guoger:
@yeasy per our offline discussion, we've agreed that rejecting step message with incorrect node id is a better choice. Also, your concern around repeatedly adding/removing a node while the node itself is behind, is not a problem. It would pull the latest config and extract correct node id to start with. It can be easily proved by adding such operations to the IT i demonstrated (which should be added i think).

looking forward to your PR update

guoger:
you'll need to add `To` to the raft message at [this line](https://github.com/hyperledger/fabric/blob/c428e2a229b8424ab0b57072df373e1a0fee9a4b/orderer/consensus/etcdraft/chain_test.go#L3748)

integration/raft/config_test.go
```diff
@@ -1077,6 +1077,75 @@ var _ = Describe("EndToEnd reconfiguration and onboarding", func() {
...
+				Expect(err).To(Not(HaveOccurred()))
+
+				By("Adding the evicted orderer back to the application channel")
+				addConsenter(network, peer, o2, "testchannel", etcdraft.Consenter{
+					ServerTlsCert: certificatesOfOrderers[0].oldCert,
+					ClientTlsCert: certificatesOfOrderers[0].oldCert,
+					Host:          "127.0.0.1",
+					Port:          uint32(network.OrdererPort(orderers[0], nwo.ClusterPort)),
+				})
+
```
guoger, line 1125: Also, let's remove and re-add the same node here again, to further advance node id

orderer/consensus/etcdraft/util.go
```diff
@@ -353,3 +354,30 @@ func CreateConsentersMap(blockMetadata *etcdraft.BlockMetadata, configMetadata *
 	}
 	return consenters
 }
+
+// detectSelfID will try to find the current osn's raft id from the given consenters.
+func detectSelfID(consenters map[uint64]*etcdraft.Consenter, Cert []byte, logger *flogging.FabricLogger) (uint64, error) {
```
yacovm: There is no need for this refactoring to be part of the change set any longer

integration/raft/config_test.go
```diff
@@ -1077,6 +1077,75 @@ var _ = Describe("EndToEnd reconfiguration and onboarding", func() {
...
+				// We need to wait for stabilization, as we might have killed the leader OSN.
+				By("Waiting for the channel to stabilize after killing the orderer")
+				assertBlockReception(map[string]int{
+					"testchannel":   0,
+					"systemchannel": 1,
+				}, []*nwo.Orderer{o2, o3}, peer, network)
+
+				By("Removing the first orderer from the application channel")
+				extendNetwork(network)
```
yacovm: why do you need to call **extendNetwork** and **refreshOrdererPEMs** ? I thought we can just remove and re-add an existing orderer?

integration/raft/config_test.go
```diff
@@ -1077,6 +1077,86 @@ var _ = Describe("EndToEnd reconfiguration and onboarding", func() {
 			Expect(resp.Status).To(Equal(common.Status_SUCCESS))
 		})
 
+		When("an evicted node is added back while it's offline", func() {
+			FIt("can start with correct new raft id", func() {
```
guoger: s/FIt/It/g

integration/raft/config_test.go
yeasy: Done, but have issues to pass.

integration/raft/config_test.go
guoger: i think this was adhering to the pattern [here](https://github.com/hyperledger/fabric/blob/251e989da1a7fe9eb7b4507ba093b140a93660a6/integration/raft/config_test.go#L1110-L1111) to pick up node cert, which could be read directly instead.

guoger (changes requested):
LGTM overall, just some nits

guoger:
And you need to add `To` section to [this line](https://github.com/hyperledger/fabric/blob/251e989da1a7fe9eb7b4507ba093b140a93660a6/orderer/consensus/etcdraft/chain_test.go#L195) too, for the same reason above

integration/raft/config_test.go
```diff
@@ -1077,6 +1077,77 @@ var _ = Describe("EndToEnd reconfiguration and onboarding", func() {
...
+
+				By("Submitting tx")
+				env := CreateBroadcastEnvelope(network, o2, "testchannel", []byte("foo"))
+				resp, err := nwo.Broadcast(network, o2, env)
+				Expect(err).NotTo(HaveOccurred())
+				Expect(resp.Status).To(Equal(common.Status_SUCCESS))
+
+				By("Waiting for the channel to stabilize")
+				assertBlockReception(map[string]int{
+					"testchannel": 3,
```
guoger: i think this should be 5, no?

guoger (changes requested):

lindluni:
I got you

guoger:
there's a bug in current validation logic. I will submit a fix and let you know @yeasy 

https://jira.hyperledger.org/browse/FAB-17951

integration/raft/config_test.go
```diff
@@ -1078,6 +1078,77 @@ var _ = Describe("EndToEnd reconfiguration and onboarding", func() {
...
+				ordererProcesses[0].Signal(syscall.SIGTERM)
+				Eventually(ordererProcesses[0].Wait(), network.EventuallyTimeout).Should(Receive())
+
+				// We need to wait for stabilization, as we might have killed the leader OSN.
+				By("Waiting for the channel to stabilize after killing the orderer")
+				assertBlockReception(map[string]int{
+					"testchannel":   0,
+					"systemchannel": 1,
+				}, []*nwo.Orderer{o2, o3}, peer, network)
+
```
guoger, line 1108: could you please add one step here to assert/wait for ActiveNode metrics to be 2? this would allow us to deterministically reproduce the Node ID error, and assure that #1367 fixes it

guoger (changes requested):
Once #1367 goes green, you could rebase on top of it. Also, please add one more assertion in the IT (commented in line)

integration/raft/config_test.go
yeasy: #1367 is still in open.
Consider the severity, do you think we can reduce to one-round remove-add in the test case, and then update after #1367 got merged?

guoger:
@yeasy #1367 is now merged, i believe you could rebase this

integration/raft/config_test.go
```diff
@@ -1078,6 +1078,82 @@ var _ = Describe("EndToEnd reconfiguration and onboarding", func() {
...
+				By("observing active nodes to shrink")
+				// Enable debug log for orderer2 so we could assert its content later
+				o2Runner := network.OrdererRunner(o2, "FABRIC_LOGGING_SPEC=orderer.consensus.etcdraft=debug:info")
```
guoger: this has to be done at line 1084 (when the orderer runner was initially created).

integration/raft/config_test.go
```diff
@@ -1078,6 +1078,80 @@ var _ = Describe("EndToEnd reconfiguration and onboarding", func() {
...
+				By("Adding the evicted orderer back to the application channel again")
+				addConsenter(network, peer, o2, "testchannel", etcdraft.Consenter{
+					ServerTlsCert: server1CertBytes,
+					ClientTlsCert: server1CertBytes,
+					Host:          "127.0.0.1",
+					Port:          uint32(network.OrdererPort(o1, nwo.ClusterPort)),
+				})
+
+				By("Starting the first orderer to see if it will get the correct raft id")
+				o1Runner := ordererRunners[0]
```
guoger: to restart o1 (effectively start a new process), you need to create a new runner, instead of reusing the old one.

integration/raft/config_test.go
```diff
@@ -1078,6 +1078,80 @@ var _ = Describe("EndToEnd reconfiguration and onboarding", func() {
...
+
+				By("Adding the evicted orderer back to the application channel again")
+				addConsenter(network, peer, o2, "testchannel", etcdraft.Consenter{
+					ServerTlsCert: server1CertBytes,
+					ClientTlsCert: server1CertBytes,
+					Host:          "127.0.0.1",
+					Port:          uint32(network.OrdererPort(o1, nwo.ClusterPort)),
+				})
+
+				By("Starting the first orderer to see if it will get the correct raft id")
```
guoger: nit - In "By", we describe the actions instead of expectations. Plus, i don't see any assertion on raft id. This could simply be
> By("Starting the first orderer again")

orderer/consensus/etcdraft/chain.go
```diff
@@ -10,11 +10,12 @@ import (
 	"context"
 	"encoding/pem"
 	"fmt"
-	"github.com/hyperledger/fabric/orderer/common/types"
 	"sync"
 	"sync/atomic"
 	"time"
 
+	"github.com/hyperledger/fabric/orderer/common/types"
+
```
guoger: remove empty line

guoger (changes requested):
LGTM overall, you need to fix a minor problem in IT to get it work

guoger (approved):

PR #1319, opened 2020-05-22
Pass HashProvider to ledger instead of Hasher

manish-sethi:

#### Type of change
- Improvement

#### Description
- So far ledger used hash function for computing hash on small amount of data at a time that was kept in memory and hence the dependency passed from `bccsp`, [Hash](https://github.com/hyperledger/fabric/blob/85599979a087765d443a2d569511e52a675d6e05/core/ledger/ledger_interface.go#L697) was sufficient. However, for computing the hash of the data streaming while exporting the data (without caching in memory) for a snapshot, the access to to Hash object is needed. This commit makes the change in this dependency. However, no change is needed on the upper layer as the already passed down dependency already implements the desired function.

- Also makes the function signature for constructing a new Hash object uniform across snapshot-able components

### DIFF
--- a/common/ledger/blkstorage/blockindex.go
+++ b/common/ledger/blkstorage/blockindex.go
@@ -9,7 +9,6 @@ package blkstorage
 import (
 	"bytes"
 	"fmt"
-	"hash"
 	"path"
 	"unicode/utf8"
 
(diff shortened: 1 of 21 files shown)


ale-linux (changes requested):
I'm probably missing something, but what's wrong with hash.Hash? It embeds an io.Writer where you can keep adding data as it streams in. You can reuse it an arbitrary number of times by calling Reset(). Where does it fall short?

cendhu:
> I'm probably missing something, but what's wrong with hash.Hash? It embeds an io.Writer where you can keep adding data as it streams in. You can reuse it an arbitrary number of times by calling Reset(). Where does it fall short?

We use it as you mentioned above. Refer to https://github.com/hyperledger/fabric/blob/master/common/ledger/snapshot/file.go#L42 and https://github.com/hyperledger/fabric/blob/master/core/ledger/confighistory/mgr.go#L215

However, when we want to stream data to multiple files in parallel and want to compute hashes of the two streams separately, we need to create two hashers. Refer to https://github.com/hyperledger/fabric/blob/master/core/ledger/kvledger/txmgmt/privacyenabledstate/snapshot.go#L29 Here, we iterate over the db and stream public data to one file and private state hashes to another file. At the end, the function returns four hashes in total (two related to data files and two related to metadata file). 

If we make the kvledger to call `ExportXXX()` API present at block store, configHistoryDB, and stateDB in parallel, we would certainly need newHasher and cannot reuse `hash.Hash`.

ale-linux (approved):
Thanks for your thorough explanation! LGTM

PR #1322, opened 2020-05-22
Silence noise when running internal/peer/lifecycle/chaincode tests

sykesm:
The tests exercise a number of commands that result in error and usage messages. These pollute the output and make it difficult to see what’s really going on.

This change sets the `SilenceErrors` and `SilenceUsage` on the commands created in test. These flags prevent Cobra from printing messages.

The change also sets the `GinkgoWriter` as the default logging writer. This suppresses logging messages unless tests fail or are executed in _verbose_ mode.


ale-linux (approved):
Thanks Matt, these log lines were almost as irritating as the incessant bccsp init log lines ;)

mastersingh24 (approved):
Very nice

PR #1325, opened 2020-05-23
Lint work

sykesm:
Address a number of issues highlighted by linters.


lindluni (approved):

lindluni:
Is this everything caught by the extended linter? If so we shouldn't enable some of these check (like unchecked errors and dead code) in the linting script

lindluni:
Also thank you for this, as much as I know you love removing dead code, its still tedious work.

PR #1326, opened 2020-05-23
Update membership link

cabelotaina:
Changing link from .html to .md

#### Type of change

- Bug fix

#### Description

Reading the document I see the link, I clicked ..., and said wow this link don't works, why? And going to the folder I see a .md file and not a .html file and I resolve to change the link. And I changed...

#### Additional details

Nop

#### Related issues

Nop

### DIFF
--- a/docs/source/identity/identity.md
+++ b/docs/source/identity/identity.md
@@ -254,7 +254,7 @@ trusted members of a blockchain network. That's where a Membership Service Provi
 (MSP) comes into play --- **it identifies the parties who are the members of a
 given organization in the blockchain network**.
 
-To learn more about membership, check out the conceptual documentation on [MSPs](../membership/membership.html).
+To learn more about membership, check out the conceptual documentation on [MSPs](../membership/membership.md).
 
 <!---
 Licensed under Creative Commons Attribution 4.0 International License https://creativecommons.org/licenses/by/4.0/


lindluni:
Thank you for the change. We require a DCO signoff on all commits, you will need to checkout your branch locally and do a `git commit --signoff --amend --no-edit && git push origin patch-1 -f`

docs/source/identity/identity.md
```diff
@@ -254,7 +254,7 @@ trusted members of a blockchain network. That's where a Membership Service Provi
 (MSP) comes into play --- **it identifies the parties who are the members of a
 given organization in the blockchain network**.
 
-To learn more about membership, check out the conceptual documentation on [MSPs](../membership/membership.html).
+To learn more about membership, check out the conceptual documentation on [MSPs](../membership/membership.md).
```
sykesm, line 257: I'm pretty sure the old link is the correct one. This page is rendered [here](https://hyperledger-fabric.readthedocs.io/en/release-2.0/identity/identity.html) and the link becomes https://hyperledger-fabric.readthedocs.io/en/release-2.0/membership/membership.html.

If you're attempting to follow links in within the repository on GitHub, they won't work.

cabelotaina: Hi I'm reading the documentation to see if it is a good behaviour. If we can use, another configuration, to use .md path links on github and transform to .html for exhibit on the page, something like this.

cabelotaina: Looking in the [convention](https://github.com/hyperledger/fabric/tree/b3ce17d4a7c02f5f7f559ea2d289470d57378f91/docs#conventions), we can see this note: 

> Relative links are unfortunately not working on github when browsing through a RST file.

But [relative link](https://github.com/hyperledger/fabric/blob/b3ce17d4a7c02f5f7f559ea2d289470d57378f91/docs/source/getting_started.rst) works for another pages not inside a same page (anchors), and the project have some files in format .md - this one for example **docs/source/identity/identity.md**, why? This is some workaround for the problem with .rst files?

Good when I can I test this on my local machine, But if someone have the response say me.

thx

lindluni: Matt is correct. We use a tool called tox to render our webpages and a third-party tool to convert `.md` and `.rst` files into `.html` files. ReadTheDocs.com, where the doc in `/docs/source` is hosted utilizes our tox configuration to pick the correct pages and render them. So when you want to view the doc in the `docs/source` directory, it's not meant to be read locally, it has to be built using the tox engine and then you need to view it in its native `.html` format, not its `.md` or `.rst` format. So opening it in your IDE or browser without rendering it with tox is not supported.

If you want to view it locally, you need to install tox and run `tox -edocs` to build it locally and then you can view it in `/docs/source/build/_html`

lindluni (changes requested):

lindluni:
I'll also add, aside from reading the doc locally, it's also no meant to be read in GitHub, there are many things that will only render on ReadTheDocs or locally after building with tox

PR #1327, opened 2020-05-23
More changes to address issues flagged by linters

sykesm:
Address more issues flagged by linters.

### DIFF
--- a/bccsp/idemix/bridge/bridge_test.go
+++ b/bccsp/idemix/bridge/bridge_test.go
@@ -764,9 +764,9 @@ var _ = Describe("Idemix Bridge", func() {
 					IssuerPublicKey,
 					credRequest,
 					nil,
-					&bccsp.IdemixCredentialRequestSignerOpts{IssuerNonce: []byte("pine-aple-pine-apple-pinapple")},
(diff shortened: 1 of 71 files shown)


bccsp/idemix/bridge/bridge_test.go
```diff
@@ -764,9 +764,9 @@ var _ = Describe("Idemix Bridge", func() {
 					IssuerPublicKey,
 					credRequest,
 					nil,
-					&bccsp.IdemixCredentialRequestSignerOpts{IssuerNonce: []byte("pine-aple-pine-apple-pinapple")},
+					&bccsp.IdemixCredentialRequestSignerOpts{IssuerNonce: []byte("pin-apple-pine-apple-pineapple")},
```
lindluni, line 767: Did you mean to spell `pin` instead of `pine` in the first instance

lindluni (approved):

bccsp/idemix/bridge/bridge_test.go
sykesm: It just needed to be a word and the total length less than 32 characters.

PR #1328, opened 2020-05-25
Address more concerns highlighted by linters

sykesm:
These changes remove dead code, add error checks, and assign unused variables to the unnamed variable `_`.

### DIFF
--- a/core/ledger/confighistory/mgr.go
+++ b/core/ledger/confighistory/mgr.go
@@ -220,6 +220,9 @@ func (r *Retriever) ExportConfigHistory(dir string, newHashFunc snapshot.NewHash
 		return nil, err
 	}
 	metadataHash, err := metadataFileWriter.Done()
+	if err != nil {
+		return nil, err
+	}
 
 	return map[string][]byte{
 		snapshotDataFileName:     dataHash,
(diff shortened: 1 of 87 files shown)


core/ledger/confighistory/mgr.go
```diff
@@ -220,6 +220,9 @@ func (r *Retriever) ExportConfigHistory(dir string, newHashFunc snapshot.NewHash
 		return nil, err
 	}
 	metadataHash, err := metadataFileWriter.Done()
+	if err != nil {
```
ale-linux, line 223: Hmm, what am I missing here, why is it a good thing to ignore that error?

sykesm: Not sure what you mean. Looks like the typical error checking pattern

```go
if err != nil {
    return nil, err
}
```

There was no error check before and now there is.

ale-linux: Sorry, my bad, I checked out the changes, used meld to inspect the diff but inverted the branches so in that particular file it looked like you removed the error check :facepalm:

ale-linux (approved):

PR #1331, opened 2020-05-26
[FAB-17927] Add AllChaincodesInfo to DeployedChaincodeInfoProvider

wenjianqiao:

#### Type of change

#### Description
Add  AllChaincodesInfo to DeployedChaincodeInfoProvider to query ledger
and return chaincode info for all the deployed chaincode (both new lifecycle
and legacy chaincodes) on a channel. 

#### Additional details
This function is needed for ledger checkpointing and deletion of channel-specific databases in statecouchdb.

#### Related issues
https://jira.hyperledger.org/browse/FAB-17927


core/chaincode/lifecycle/deployedcc_infoprovider.go
```diff
@@ -103,6 +103,39 @@ func (vc *ValidatorCommitter) ChaincodeInfo(channelName, chaincodeName string, q
...
+	}
+	ccQuery := &ExternalFunctions{Resources: vc.Resources}
+	namespaceDefs, err := ccQuery.QueryNamespaceDefinitions(sqes)
+	if err != nil {
+		return nil, err
+	}
+
+	result := make([]*ledger.DeployedChaincodeInfo, 0)
+	for ccName := range namespaceDefs {
+		deployedccInfo, err := vc.ChaincodeInfo(channelName, ccName, sqe)
```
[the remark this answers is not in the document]
manish-sethi: `QueryNamespaceDefinitions()` returns listing only from `_lifecycle` namespace. So, `ChaincodeInfo()` would never execute line 94. 

Still, it may lead to duplicate entries for the chaincodes that are defined in both `_lifecycle` and `lscc` and hence, at line 132, the results form `legacyCCs` needs to be filtered out, if already appeared in `_lifecycle`

core/scc/lscc/deployedcc_infoprovider.go
```diff
@@ -84,6 +85,34 @@ func (p *DeployedCCInfoProvider) ChaincodeInfo(channelName, chaincodeName string
 	}, nil
 }
 
+// AllChaincodesInfo returns a list of legacy chaincodes info
+func (p *DeployedCCInfoProvider) AllChaincodesInfo(channelName string, qe ledger.SimpleQueryExecutor) ([]*ledger.DeployedChaincodeInfo, error) {
+	iter, err := qe.GetStateRangeScanIterator(lsccNamespace, "", "")
+	if err != nil {
+		return nil, err
+	}
```
manish-sethi, line 93: `defer iter.Release()`

core/chaincode/lifecycle/deployedcc_infoprovider.go
```diff
@@ -103,6 +103,39 @@ func (vc *ValidatorCommitter) ChaincodeInfo(channelName, chaincodeName string, q
 	}, nil
 }
 
+// AllChaincodesInfo returns a list of deployed chaincodes info, including both new lifecycle and legacy chaincodes
+func (vc *ValidatorCommitter) AllChaincodesInfo(channelName string, sqe ledger.SimpleQueryExecutor) ([]*ledger.DeployedChaincodeInfo, error) {
```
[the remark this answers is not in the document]
manish-sethi: IMO, at the level of `DeployedCCInfoProvider`, it seems good to return `DeployedChaincodeInfo` to be consistent with other functions.

core/scc/lscc/deployedcc_infoprovider.go
```diff
@@ -84,6 +85,34 @@ func (p *DeployedCCInfoProvider) ChaincodeInfo(channelName, chaincodeName string
...
+		if err != nil {
+			return nil, err
+		}
+		if entry == nil {
+			break
+		}
+
+		kv := entry.(*queryresult.KV)
+		if !privdata.IsCollectionConfigKey(kv.Key) {
+			deployedccInfo, err := p.ChaincodeInfo(channelName, kv.Key, qe)
```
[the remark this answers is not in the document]
manish-sethi: In case you have not already figured, the entries for the chaincodes and collection config could be interleaved. So, I feel that better to either keep this function as is for simplicity or otherwise maintain a map that keeps one entry for each chaincode. This map would get updated by either adding a new entry for the chaincodeInfo, or by updating the existing chaincodeInfo with its collection config (when encountered). I do not have a strong inclination either way, as this is not in the critical perf path.

core/chaincode/lifecycle/deployedcc_infoprovider_test.go
```diff
@@ -236,6 +240,114 @@ var _ = Describe("ValidatorCommitter", func() {
...
 	})
 
+	Describe("AllChaincodesInfo", func() {
+		var (
+			fakeStateIteratorKVs MapLedgerShim
+		)
+		BeforeEach(func() {
+			fakeStateIteratorKVs = MapLedgerShim(map[string][]byte{})
+			err := resources.Serializer.Serialize("namespaces", "cc-name", &lifecycle.ChaincodeDefinition{}, fakeStateIteratorKVs)
+			Expect(err).NotTo(HaveOccurred())
```
manish-sethi, line 250: Better to add a non-nil collection config in `lifecycle.ChaincodeDefinition`, because the function `AllChaincodesInfo` is expected to return that as well.

core/chaincode/lifecycle/deployedcc_infoprovider_test.go
```diff
@@ -236,6 +240,114 @@ var _ = Describe("ValidatorCommitter", func() {
...
+				}
+				return fakeResultsIterator, nil
+			}
+		})
+
+		It("returns deployed chaincodes info for new lifecycle chaincodes", func() {
+			res, err := vc.AllChaincodesInfo("channel-name", fakeQueryExecutor)
+			Expect(err).NotTo(HaveOccurred())
+			Expect(res).To(HaveLen(1))
+			// because ExplicitCollectionConfigPkg is a protobuf object, we have to compare individual fields
```
manish-sethi, line 271: leftover comment?

core/chaincode/lifecycle/deployedcc_infoprovider_test.go
```diff
@@ -236,6 +240,114 @@ var _ = Describe("ValidatorCommitter", func() {
...
+				fakePublicState["namespaces/fields/cc-name/ValidationInfo"] = []byte("garbage")
+			})
+
+			It("wraps and returns that error", func() {
+				_, err := vc.AllChaincodesInfo("channel-name", fakeQueryExecutor)
+				Expect(err).To(MatchError("could not get info about chaincode: could not deserialize chaincode definition for chaincode cc-name: could not unmarshal state for key namespaces/fields/cc-name/ValidationInfo: proto: can't skip unknown wire type 7"))
+			})
+		})
+	})
+
```
manish-sethi, line 381: As discussed in the [comment](https://github.com/hyperledger/fabric/pull/1331/files#r430630953), better to add a test where there is an overlap in the results from `_lifecycle` and `lscc`

manish-sethi (changes requested):

core/chaincode/lifecycle/deployedcc_infoprovider.go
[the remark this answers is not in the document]
wenjianqiao: As @manish-sethi said, the method returns DeployedChaincodeInfo to be consist with other functions. Plus ledger.DeployedChaincodeInfo is already used by ledger component.

core/scc/lscc/deployedcc_infoprovider.go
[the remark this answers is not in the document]
wenjianqiao: What @manish-sethi said was the exact reason that the PR skipped collection entries and called ChaincodeInfo for a simple implementation. Otherwise, we will have to assemble multiple entries.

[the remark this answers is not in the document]
cendhu: Okay. It just looked weird to me that the required key and value are present in this function but we call some other function to get them again. As we have provided it like a generic API, I cannot comment on the performance as I don't know where this API would be called in the future. 

Anyway, I am fine with the current usage as it helps to avoid the following lines of code. Note that it is always possible to reduce the number of lines of code by adding hacks. 

```
deployedCCInfo := make(map[string]*DeployedChaincodeInfo)
if  !privdata.IsCollectionConfigKey(kv.Key) {
        chaincodeData := &ccprovider.ChaincodeData{}
	if err := proto.Unmarshal(chaincodeDataBytes, chaincodeData); err != nil {
		return nil, errors.Wrap(err, "error unmarshalling chaincode state data")
	}
	deployedCCInfo[kv.Key] = &ledger.DeployedChaincodeInfo{Name: kv.Key, Hash: chaincodeData.Id, Version: chaincodeData.Version, IsLegacy: true}
        continue
}
ccName := privdata.GetCCNameFromCollectionConfigKey(kv.Key)
collectionConfigPkg := &peer.CollectionConfigPackage{}
if err := proto.Unmarshal(collectionConfigPkgBytes, collectionConfigPkg); err != nil {
	return nil, errors.Wrap(err, "error unmarshalling chaincode collection config pkg")
}
deployedCCInfo[ccName].ExplicitCollectionConfigPkg = collectionConfigPkg

deployedCCInfo := []*DeployedChaincodeInfo{}
for _, ccInfo := range deplyedCCInfo {
        append(deployedCCInfo, ccInfo)
}
```

core/chaincode/lifecycle/deployedcc_infoprovider.go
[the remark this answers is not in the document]
cendhu: DeployedChaincodeInfoProvider returns `[]string`, `ChaincodeLifecycleInfo`, `DeployedChaincodeInfo`, `peer.StaticCollectionConfig`, `peer.CollectionConfigPackage`. Some of these return types are defined in the ledger_interface.go. Hence, returning a new type called `DeployedChaincodesNamespaces` that is `type DeployedChaincodesNamespaces map[string]string` (to be defined in the ledger_interface.go) or some better name simplifies the whole operation and then we can remove this API later once we are done with our use-case. 

As I mentioned in the initial comment, I am also okay with retaining the current function as it can be reused by other components in the future.

core/chaincode/lifecycle/deployedcc_infoprovider.go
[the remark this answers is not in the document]
wenjianqiao: QueryNamespaceDefinitions() returns only namespaces using new lifecycle (_lifecycle namespace). What use case may lead lscc and _lifecycle namespaces to have duplicate entries? When a legacy chaincode is upgraded to a new lifecycle chaincode?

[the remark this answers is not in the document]
manish-sethi: yes

core/chaincode/lifecycle/deployedcc_infoprovider.go
ale-linux: Thanks Wenjian, could you please explain the difference between this function and 

https://github.com/hyperledger/fabric/blob/master/core/chaincode/lifecycle/cache.go#L64

It too delivers a comprehensive list of all chaincode definitions from both legacy and new lifecycle.

cendhu: **Background:** When CouchDB is used as the state database, for each chaincode within a channel and for each collection within a chaincode, we create a separate database.

**Scenario:** We want to delete all databases associated with a channel or export data of a particular channel for a ledger snapshot -- 

**Problem:** We don't explicitly store all databases name associated with each channel. Even by querying all databases present in a CouchDB, we cannot identify a set of databases associated with a channel. 

**Challenge:** If we know the names of all chaincodes and collections within a channel, we can find all databases associated with a channel. In order to find all existing chaincodes and collections belong to a channel, ledger needs to depend on the lifecycle pkg. 

**Solution:** For all existing channels/chaincodes/collections as well as for new ones, we plan to store the following mapping in CouchDB: [database name, channelName], [database name, chaincodeName] [database name,  collectionName]. 

In  #1268, we started to store the mapping for new databases (also for existing databases but only when they are opened -- quite complicated to explain -- refer to this [discussion](https://github.com/hyperledger/fabric/pull/1268#pullrequestreview-413057258)). 

To store the mapping for all existing databases, currently, we plan to use this API introduced by this PR when the CouchDB database is opened for an existing channel. In other words, when the peer starts up, it initializes various components. As you know, Peer also acquires `ledger.PeerLedger` for each existing channel. When the ledger associated with a channel is opened, we open respective databases too. After opening the database (approximately [here](https://github.com/hyperledger/fabric/blob/master/core/ledger/kvledger/txmgmt/privacyenabledstate/db.go#L98), we want to call the deployedCCInfo provider to get all deployed chaincode names and respective collection names. Once it is retrieved, the db would be asked to update the mapping it maintains (introduced in #1268). While this occurs, the lifecycle cache would not have been populated. This is because we initialize the lifecycle cache at https://github.com/hyperledger/fabric/blob/master/core/ledger/kvledger/kv_ledger.go#L166 once the database is opened for the ledger. 

Further, the `ledger.DeployedCCInfoProvider` is already passed to the kvledger initializer. We plan to reuse the existing dependency itself for our needs. It may not be good to add more dependency to the kvledger package. Note that this API is used once when the peer is upgraded to 2.2 or from old LTS to new LTS. Once we store the mapping of all existing channels/chaincodes/collection to database name, this API can be removed.

manish-sethi: > https://github.com/hyperledger/fabric/blob/master/core/chaincode/lifecycle/cache.go#L64
> 
> It too delivers a comprehensive list of all chaincode definitions from both legacy and new lifecycle.

As far as I recall, this cache only maintains data in the `_lifecycle` namespace. 

Moreover, the purpose of this cache was primarily to maintain some states so as to be able to capture and process the events related to lifecycle operations, as opposed to what the deployedccinfoprovider (in the same package) does - i.e., provide a stateless set of functions just to hide the details of the lifecycle data model from other components.

core/chaincode/lifecycle/deployedcc_infoprovider_test.go
wenjianqiao: The empty ChaincodeDefinition here is only used for GetStateRangeScanIteratorStub so that QueryNamespaceDefinitions can return a mocked map[string]string. The actual ChaincodeDefinition is from mocked GetState that returns fakeChaincodeDef (line 94). In the test below BeforeEach, it verifies the result against fakeChaincodeDef.

core/chaincode/lifecycle/deployedcc_infoprovider_test.go
wenjianqiao: It is not a leftover comment. It explains why we have to compare individual fields instead of entire object.

core/chaincode/lifecycle/deployedcc_infoprovider.go
cendhu: The lifecycleCache is being used by few other components too (to get endorsement info, I think).

ale-linux: thanks @cendhu for the very thorough explanation! But to @manish-sethi 's point: I'm pretty sure that
https://github.com/hyperledger/fabric/blob/master/core/chaincode/lifecycle/cache.go#L64
gives a list of {legacy U new} definitions, with shadowing rules (e.g. if a chaincode is defined in both, we'll get defintions from the new only). Couldn't we just make use of that and register a new listener there? My point is just that we already have a ton of these lifecycle info providers, and we should try to avoid adding new ones unless it's strictly mandatory.

wenjianqiao: @ale-linux thanks for your comment and pointing out cache.go. I see a few different usages between ledger.DeployedChaincodeInfoProvider and lifecycle.Cache.
- DeployedChaincodeInfoProvider is an existing interface used by ledger. This PR adds a new method that returns an existing data type. It is a simple extension and fits the existing usage pattern. On the other hand, ledger does not depend on lifecycle.Cache to retrieve chaincode info, so we will have to introduce new dependency to ledger.
- ledger needs to get deployed chaincodes, but lifecycle.Cache only has ListInstalledChaincodes. 
- ledger needs to get deployed chaincodes for both new lifecycle and legacy chaincodes. Looking at https://github.com/hyperledger/fabric/blob/master/internal/peer/node/start.go#L769, lifecycle cache only initializes new lifecycle chaincodes. In my test, I verified that lifecycle.Cache didn't include legacy ccs.
- ledger needs to get chaincode data during ledger initialization. The lifecycle cache is initialized after ledger init. Although we may be able to refactor the code, it will add more complexity. 

Given the above, I think it makes sense to use DeployedChaincodeinfoProvider.

ale-linux: Thanks @wenjianqiao !

manish-sethi (approved):

PR #1335, opened 2020-05-27
Remove txmgr interface

manish-sethi:

#### Type of change
- Improvement (Refactoring)

#### Description
This PR removes the interface `TxMgr`, as there is a single implementation - `LockBasedTxMgr`. All the dependent packages are able to use this single implementation directly. The only exception is the `ledger/validation` package that causes a circular dependency. 
The `ledger/validation` package uses only a single function from this interface - `NewTxSimulator` in order to allow for the execution a post-order transaction (only channel config transaction in the current code). For meeting this need, this commit now defines a
local interface with a single function in that package itself.

#### Additional details
Initially, there was a plan to provide an alternate implementation, a more efficient approach - lock free TxMgr (using a variation of timestamp based approach). Since, that is not planned for the near future, we remove this interface for a better code navigation/readability.


cendhu (approved):

denyeart (approved):

core/ledger/kvledger/txmgmt/txmgr/lockbased_txmgr.go
```diff
@@ -53,6 +52,26 @@ type pvtdataPurgeMgr struct {
 	usedOnce bool
 }
 
+// ErrUnsupportedTransaction is expected to be thrown if a unsupported query is performed in an update transaction
```
sykesm, line 55: https://github.com/golang/go/wiki/Errors

cendhu: This PR only removed the interface and moved around the existing code. These error types are not added newly. 

In terms of usage, golang has similar err types too https://github.com/golang/go/blob/c5058652fd724cb1ed8ea91a4e76b202d6910482/src/math/big/float.go#L77-L83

Anyway, I agree that a struct is not needed here as it does carry extra data with the error.  Instead, we can go with `var errType` as defined here https://github.com/hyperledger/fabric/blob/master/core/ledger/ledgermgmt/ledger_mgmt.go#L30 and
https://github.com/golang/go/blob/7067aa1c106d000e99b62e4804118c7bceeed680/src/internal/syscall/windows/registry/value.go#L41

sykesm: I should have chosen the section: https://github.com/golang/go/wiki/Errors#Naming

I’ve commented on this error naming convention in the ledger code before as it’s something I think should be addressed to adhere to the expected conventions. ErrXXX is for holding sentinel error values and XXXError is for defining new error types. `ErrUnsupposedTransaction` is a type and should be named `UnsupportedTransactionError`.

I’d also recommend reviewing the [Constant errors](https://dave.cheney.net/2016/04/07/constant-errors) blog post to see if that pattern makes sense.

manish-sethi: Thanks Matt. We can review all the custom errors defined in ledger code and fix the names in a single PR. In fact, I am sure that except a few, these need not be the custom errors.

PR #1337, opened 2020-05-27
[FAB-17471] Fix OrdererType line in configtx.yaml

naonishijima:
#### Type of change
- Bug fix

#### Description
This CR moves OrdererType to the correct line in sampleconfig/configtx.yaml. SampleInsecureKafka profile tries to change consensus type solo to Kafka using OrdererType. But it was written above "<<: *OrdererDefaults" line. That causes the overwrite consensus type again.
This CR can generate the correct ordererer type's genesis block.

### DIFF
--- a/sampleconfig/configtx.yaml
+++ b/sampleconfig/configtx.yaml
@@ -485,8 +485,8 @@ Profiles:
     SampleInsecureKafka:
         <<: *ChannelDefaults
         Orderer:
-            OrdererType: kafka
             <<: *OrdererDefaults
+            OrdererType: kafka
         Consortiums:
             SampleConsortium:
                 Organizations:


sykesm (changes requested):
Thank you for the PR.

While experience tells me that the yaml parser implementation commonly used in go will overwrite values with the last observed key, I believe the yaml specification [requires keys in mappings to be unique](https://yaml.org/spec/1.1/#mapping/information%20model).

> The content of a mapping node is an unordered set of key: value node pairs, with the restriction that each of the keys is unique.

In the case of merge keys, they actually specify that key value/pairs are added [*unless* the key already exists in it](https://yaml.org/type/merge.html).

> The “<<” merge key is used to indicate that all the keys of one or more specified maps should be inserted into the current map. If the value associated with the key is a single mapping node, each of its key/value pairs is inserted into the current mapping, unless the key already exists in it.

Given the current structure isn't really adhering to the spec, I'd rather avoid the ordering problem altogether and, instead, remove the duplicate keys.

For the sample config, that means removing `OrdererType` from the mapping declared by `OrdererDefaults` and explicitly specifying `OrdererType: solo` where it's necessary.

naonishijima:
Thanks for reviewing. I updated the patch.

sampleconfig/configtx.yaml
```diff
@@ -476,6 +477,8 @@ Profiles:
         <<: *ChannelDefaults
         Orderer:
             <<: *OrdererDefaults
+            OrdererType: solo
+            Organizations:
```
sykesm: Since there’s nothing associated with the key, I think we can omit it.

sampleconfig/configtx.yaml
```diff
@@ -500,6 +504,7 @@ Profiles:
         <<: *ChannelDefaults
         Orderer:
             <<: *OrdererDefaults
+            OrdererType: solo
             Organizations:
                 - <<: *SampleOrg
```
sykesm: I don’t want to balloon the size of this change so I think this is a separate issue, but, the `SampleOrgPolicies` reference has the same sort of issue where `Admins` is defined in the anchor and then defined again after the reference.

@jyellick - Do you have any thoughts here? It looks like the structure of the yaml is relying quite a bit on the implementation of the yaml parser in a way that is not strictly spec compliant. I’m not sure how far we want to go down the path of making this “correct”.

sykesm (commented):
LGTM; I’m okay with merging but I’d like to get @jyellick to have a look first.

jyellick:
I don't think there is a problem here, and in fact, this is a bit more nuanced than it first appears.

The merge operator `<<` (or `!!merge`) is not part of  the [v1.2 yaml spec](https://yaml.org/spec/1.2/spec.html), so, technically, `configtx.yaml` is not pure v1.2 yaml.

Instead, the merge operator is defined as part of of the [yaml types specification](https://yaml.org/type/) which is not required for yaml parsing, but is strongly recommended, to quote:

> The following is the list of language-independent YAML tags defined under the domain yaml.org. The use of these tags is not mandatory. However these tags represent types that are useful across a wide range of applications and it is strongly recommended they be used whenever appropriate to promote interoperability.

Referencing the 'merge' type [as defined in this recommended portion of the spec](https://yaml.org/type/merge.html), the behavior indicates that the merged elements should be inserted "unless the key already exists in it".  It seems that @sykesm's reading of this is that the key should be defined before the merge, but it seems the merge is always evaluated after the other keys.  We can look to the provided example.  Copy-pasting a subset from their example of equivalent maps:

```---
- &CENTER { x: 1, y: 2 }
- &LEFT { x: 0, y: 2 }
- &BIG { r: 10 }
- &SMALL { r: 1 }

# All the following maps are equal:

- # Explicit keys
  x: 1
  y: 2
  r: 10
  label: center/big

...

- # Override
  << : [ *BIG, *LEFT, *SMALL ]
  x: 1
  label: center/big
```

So, in this case, the merge operator specifies an anchor `LEFT` which includes the key `x` set to `0`, and then the key `x` is specified a second time as `1`.  The resulting map has `x` set to `1`.

So, as best as I can tell, although `configtx.yaml` depends on the types extension for merging, it is using the merge type appropriately and this PR is not necessary.

sykesm:
> So, as best as I can tell, although `configtx.yaml` depends on the types extension for merging, it is using the merge type appropriately and this PR is not necessary.

Except that it doesn’t actually work with the yaml parser implementation we are currently using. We either need to remove the duplicate keys or we need rely on a detail and reorder the anchor references to precede the overrides.

naonishijima:
@jyellick Thanks for reviewing.

The issue is that SampleInsecureKafka profile makes a wrong genesis block.
https://jira.hyperledger.org/browse/FAB-17471

The reason is that OrdererType is written on the not correct line.
https://github.com/hyperledger/fabric/blob/master/sampleconfig/configtx.yaml#L488

SampleInsecureKafka tries to change orderer type to Kafka using OrdererType key. But this key is written before "<<: OrdererDefaults". This will be orderer type overwrite to solo which is set in OrdererDefaults.

The first version patch moves OrdererType to after OrdererDefaults.

sykesm:
@naonishijima It looks like your first commit was probably perfectly fine. Apologies for the churn. Unless @jyellick has a different opinion, I’m happy to go back to what you started with and simply order the key after the reference.

jyellick:
> Except that it doesn’t actually work with the yaml parser implementation we are currently using.

I see.  It wasn't clear to me that this was a parsing problem in the yaml parser used by configtxgen, rather than a more general statement on the correctness of the yaml when processed by other tools.

> We either need to remove the duplicate keys or we need rely on a detail and reorder the anchor references to precede the overrides.

Unless we intend to remove all of the dependence on this behavior, which, as you point out there is a fair bit of it, I don't see removing a small subset provides much value.

> @naonishijima It looks like your first commit was probably perfectly fine. Apologies for the churn. Unless @jyellick has a different opinion, I’m happy to go back to what you started with and simply order the key after the reference.

Yes, my vote would be to simply fix the ordering to apply the merge before the key.

naonishijima:
I updated the patch.
@sykesm I learned a lot today about yaml. That was really helpful. Thanks!

sykesm (approved):

PR #1338, opened 2020-05-27
[FAB-17933] Fix cache update logic for installed chaincode info when …

satota2:
…an empty or uninstalled package ID is specified

This patch fixes cache update logic for installed chaincode information an empty or uninstalled package ID is specified.

#### Type of change

- Bug fix

#### Description

Current cache update logic for install chaincode info implemented in `core/chaincode/lifecycle/cache.go` does not work properly when an empty or uninstalled package ID is specified.

These wrong references may mislead Fabric admins.
So, we should fix the problem.

#### Related issues

- https://jira.hyperledger.org/browse/FAB-17933

### DIFF
--- a/core/chaincode/lifecycle/cache_test.go
+++ b/core/chaincode/lifecycle/cache_test.go
@@ -116,15 +116,17 @@ var _ = Describe("Cache", func() {
 						string(util.ComputeSHA256([]byte("namespaces/fields/chaincode-name#3/EndorsementInfo"))),
 						string(util.ComputeSHA256([]byte("namespaces/fields/chaincode-name#3/ValidationInfo"))),
 						string(util.ComputeSHA256([]byte("namespaces/fields/chaincode-name#3/Collections"))),
+						string(util.ComputeSHA256([]byte("chaincode-sources/fields/chaincode-name#3/PackageID"))),
 					},
 				},
 			},
 			InterestingHashes: map[string]string{
-				string(util.ComputeSHA256([]byte("namespaces/metadata/chaincode-name#3"))):               "chaincode-name",
-				string(util.ComputeSHA256([]byte("namespaces/fields/chaincode-name#3/Sequence"))):        "chaincode-name",
-				string(util.ComputeSHA256([]byte("namespaces/fields/chaincode-name#3/EndorsementInfo"))): "chaincode-name",
-				string(util.ComputeSHA256([]byte("namespaces/fields/chaincode-name#3/ValidationInfo"))):  "chaincode-name",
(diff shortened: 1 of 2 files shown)


core/chaincode/lifecycle/cache_test.go
```diff
@@ -1143,6 +1145,113 @@ var _ = Describe("Cache", func() {
 					Expect(installedCC.References["channel-id"]).To(HaveLen(1))
 				})
 			})
+
+			Context("when a chaincode definition is updated by an approve event with an empty package ID for existing chaincode name and sequence number", func() {
+
+				BeforeEach(func() {
```
ale-linux, line 1151: This BeforeEach has a considerable amount of identical code as the one below - could you please unify the contexts, have a common BeforeEach and highlight the differences in two nested contexts?

ale-linux (commented):
Thanks for your contribution! Just one minor nit on tests.
@manish-sethi, @jyellick: with this patch, it would now seem possible that if the admin approves with "packageID", then approves with empty package ID, then approves again with "packageID", we deliver "installed" events for the same package multiple times. Is this a problem for instance for couch DB indexes?

satota2:
@ale-linux
Thank you for your reviews.
I updated the patch to unify the contexts for the newly added tests to address your comments.

denyeart:
@manish-sethi @jyellick could you see Ale's question above?

jyellick:
@ale-linux I think that @manish-sethi can speak more specifically to the actual affects of multiply installing the same indexes.  I'm actually a little unclear what happens with indices when upgrading from one chaincode version to another.  My understanding of couchdb indexes is that if the documents do not match the indexes, then they are ignored.  So, perhaps indexes are always an additive procedure? I'm not sure.

I do not believe that we actually 'unset' indexes if a chaincode no longer requires them.  Though, I would imagine that that is what we should actually do.  Once we remove the legacy lifecycle, we should be able to significantly rework the "chaincode install event path" and along with it, we should make sure that we're handling these edge cases appropriately.

I guess the short answer is "I don't see that this manifests any problem paths that didn't exist before."  Though I do think @manish-sethi would have a more informed opinion.

manish-sethi:
@ale-linux - Sorry, I missed this earlier. Wish there was a way to receive a separate notification other than getting regular email like any other activity on github.

Regarding the question, no, there is no problem with applying the same indexes again. One of the existing [integration test](https://github.com/hyperledger/fabric/blob/a5411a18f578de1e50d7a22bd62571d2e4d1d0ed/integration/ledger/couchdb_indexes_test.go#L160) does something of this effect.

ale-linux (approved):
Thanks @manish-sethi and @jyellick! LGTM

PR #1343, opened 2020-05-27
Revert "Bump viper version to the last working commit"

lindluni:
This change unknowingly broke configuration. Values were ignored entirely during the viper parsing and they reverted to their default values. We were able to observe this locally in testing by enabling PKCS11 BCCSP provider and note that it still fell back to the default provider.

You can also test that this change breaks things by putting unintentional values in for certain parameters that should break configuration, but dont, such as putting in a bad hash value for BCCSP and note everything still starts just fine.


lindluni:
Waiting to see if we want to try and debug it or accept this revert

sykesm (approved):
Such is life. Nothing is free.

PR #1346, opened 2020-05-28
mv DBValue proto to statedb pkg -- needed for couchdb export

cendhu:
#### Type of change

#### Description

The `DBValue` proto encloses bytes of `version`, `value`, and `metadata`. This proto message has multiple uses. It helps in

1. storing the version, value, and metadata as a bytes slice in the DB.
2. returning the version, value, metadata associated with a key as a bytes slice.

Currently, (1) and (2) are applicable only for the stateleveldb and hence, the DBValue proto is present in the stateleveldb pkg. Now, we need (2) for the statecouchdb also--to export the data out of couchdb using `FullScanIterator`. Instead of duplicating the proto and encoding/decoding function, we move the DBValue proto from stateleveldb pkg to stated pkg which is a common location for both stateleveldb and statecouchdb pkg.  

#### Additional details

We do have a `statedb.VersioneKV` (a non-proto struct) that looks similar to `DBValue` proto except that version is not represented as bytes. If desired, in a separate PR, we can replace every use of `statedb.VersionedKV` with DBValue so that we don't need to maintain two similar looking types and also avoid copying references between these types.


manish-sethi:
I am not sure if this change is in the right direction. The db value format was supposed to be implementation specific and hence in the particular package. From couchdb point of view, the proto for snapshot may look different - with two fields only {`value`, `versionAndMetadata`} as the second field is already encoded. So, my preference would be either to have a separate couch specific proto, in couch package or bite the bullet on performance impact and use the existing versionedKV - because, for couchdb, anyway you would end up transforming for `DBValue` Format. wdyt?

cendhu:
I have pushed the WIP PR -- #1348 As I reuse existing scanner to retrieve all the data, I don't get `versionedAndMetadata` instead `statedb.VersionedKV`. If we implement another scanner, it would duplicate a lot of code from the existing scanner. In order to convert the version, value, and metadata, I have to define and use the same DBValue proto present in the stateleveldb pkg. To avoid duplication, I moved the DBValue proto to statedb. As a result, the goal was to remove the db specific encoding/decoding. In that case, we also don't need db value format. It kinds of simplify everything IMO.

cendhu:
> bite the bullet on performance impact and use the existing versionedKV - because, for couchdb, anyway you would end up transforming for DBValue Format. wdyt?

Are you suggesting to remove the []byte return from `statedb.FullScanIterator.Next()`, instead return the `VersionedKV` from both level and couch db and let the `privacyenabledstate` pkg to create a proto for the `VersionedKV` before writing to the file?

cendhu:
Manish suggested a simpler solution that makes a small change to the existing scanner implementation. As that solved this issue, I am closing this PR.

PR #1349, opened 2020-05-28
FAB-17890 Ch.Part.API: allow registrar to list a single channel

tock-ibm:
Implement the ChannelInfo(..) method in the registrar that looks up
a channel in the chains map and reports extended information on its
status. For this end we have to introduce the follower.Chain and
augment existing cluster type chains with means of reporting their
status.

- Introduce the skeleton a new type of consensus.Chain
  implementation: follower.Chain. This will be created and run when
  the orderer is required to follow the cluster and pull blocks from
  other orderers.

- The plan (for future commits) is for the follower.Chain to trigger
  the creation of an etcdraft.Chain when it discovers the orderer was
  added to the cluster, and vise versa; the etcdraft.Chain will replace
  itself with a follower.Chain when the orderer is removed from the
  cluster.

- Introduce a new interface that cluster-type chains implement, that
  allows them to report their relation to the cluster and their status.
  This is done because the registrar is not aware of the exact type of
  the chains it is keeping. The registrar cannot reflect on the type
  as well, as this will cause an import cycle (due to the etcdraft
  package importing multichannel).

Change-Id: Ia454f47f04a8ba3dcd76886a5919d1c734c01015

#### Type of change

- New feature

#### Related issues

Task: FAB-17980
Story: FAB-17824
Epic: FAB-17712

### DIFF
--- a/orderer/common/multichannel/registrar.go
+++ b/orderer/common/multichannel/registrar.go
@@ -401,8 +401,20 @@ func (r *Registrar) ChannelList() types.ChannelList {
 }
 
 func (r *Registrar) ChannelInfo(channelID string) (types.ChannelInfo, error) {
-	//TODO
-	return types.ChannelInfo{}, errors.New("Not implemented yet")
+	r.lock.RLock()
+	defer r.lock.RUnlock()
+
+	info := types.ChannelInfo{}
+	cs, ok := r.chains[channelID]
+	if !ok {
+		return info, types.ErrChannelNotExist
+	}
+
+	info.Name = channelID
+	info.Height = cs.Height()
+	info.ClusterRelation, info.Status = cs.StatusReport()
+
+	return info, nil
 }
 
 func (r *Registrar) JoinChannel(channelID string, configBlock *cb.Block) (types.ChannelInfo, error) {
--- a/orderer/common/types/channelinfo.go
+++ b/orderer/common/types/channelinfo.go
@@ -29,6 +29,23 @@ type ChannelInfoShort struct {
 	URL string `json:"url"`
 }
 
+type ClusterRelation string
+
+const (
+	ClusterRelationMember        ClusterRelation = "member"
+	ClusterRelationFollower      ClusterRelation = "follower"
+	ClusterRelationConfigTracker ClusterRelation = "config-tracker"
+	ClusterRelationNone          ClusterRelation = "none"
+)
+
+type Status string
+
+const (
+	StatusActive     Status = "active"
+	StatusOnBoarding Status = "onboarding"
+	StatusInactive   Status = "inactive"
+)
+
 // ChannelInfo carries the response to an HTTP request to List a single channel.
 // This is marshaled into the body of the HTTP response.
 type ChannelInfo struct {
@@ -39,11 +56,11 @@ type ChannelInfo struct {
 	// Whether the orderer is a “member” or ”follower” of the cluster, or "config-tracker" of the cluster, for this channel.
 	// For non cluster consensus types (solo, kafka) it is "none".
 	// Possible values:  “member”, ”follower”, "config-tracker", "none".
-	ClusterRelation string `json:"clusterRelation"`
+	ClusterRelation ClusterRelation `json:"clusterRelation"`
 	// Whether the orderer is ”onboarding”, ”active”, or "inactive", for this channel.
 	// For non cluster consensus types (solo, kafka) it is "active".
 	// Possible values:  “onboarding”, ”active”, "inactive".
-	Status string `json:"status"`
+	Status Status `json:"status"`
 	// Current block height.
 	Height uint64 `json:"height"`
 }
--- a/orderer/common/types/channelinfo.go
+++ b/orderer/common/types/channelinfo.go
@@ -29,21 +29,36 @@ type ChannelInfoShort struct {
 	URL string `json:"url"`
 }
 
+// ClusterRelation represents the relationship between the orderer and the channel's consensus cluster.
 type ClusterRelation string
 
 const (
-	ClusterRelationMember        ClusterRelation = "member"
-	ClusterRelationFollower      ClusterRelation = "follower"
+	// The orderer is a cluster member of a cluster consensus protocol (e.g. etcdraft) for a specific channel.
+	// That is, the orderer is in the consenters set of the channel.
+	ClusterRelationMember ClusterRelation = "member"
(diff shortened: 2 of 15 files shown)


orderer/common/multichannel/registrar.go
```diff
@@ -401,8 +401,20 @@ func (r *Registrar) ChannelList() types.ChannelList {
...
+	r.lock.RLock()
+	defer r.lock.RUnlock()
+
+	info := types.ChannelInfo{}
+	cs, ok := r.chains[channelID]
+	if !ok {
+		return info, types.ErrChannelNotExist
+	}
+
+	info.Name = channelID
```
yacovm, line 413: can't you initialize them in a single directive?

orderer/consensus/cluster_status.go
```diff
@@ -0,0 +1,30 @@
...
+// It allows the node to report its cluster relation and its status within that relation.
+// This information is used to generate the channelparticipation.ChannelInfo in response
+// to a "List" request on a particular channel.
+//
+// Not all chains must implement this, in particular non-cluster-type (solo, kafka) are
+// assigned a StaticStatusReporter at construction time.
+type StatusReporter interface {
+	// StatusReport provides the cluster relation and status.
+	// See:  channelparticipation.ChannelInfo for more details.
+	StatusReport() (clusterRelation, status string)
```
yacovm: I am not sure they should be string. Why not enums?

orderer/consensus/etcdraft/chain.go
```diff
@@ -1315,6 +1315,11 @@ func (c *Chain) ValidateConsensusMetadata(oldMetadataBytes, newMetadataBytes []b
 	return nil
 }
 
+// StatusReport returns the ClusterRelation & Status
+func (c *Chain) StatusReport() (string, string) {
+	return "member", "active"
```
yacovm: Perhaps instead of constructing the API results in a loose manner such as here, we define constants and return them?
That's also why I suggest using enums, so that you can have different types for "relation" and "status".  Also I am not sure why we need 2 different aspects. 
We have "member", and "follower" for a role, and a status.
Can't we flatten them with a Cartesian product of all possible combinations into a single enum? How many are there? I'm sure it's not too many...

tock-ibm: These go directly into the `types.ChannelInfo` struct which gets marshaled into json and sent in the http response. I do not see a point in having an intermediate type as an enum, and then translating to the strings that need to go into the json response.

As you can see in the documentation below, the "ClusterRelation" and "Status"  fields go into separate fields in that struct, which is what the API design calls for. So the internal representation follows the API to avoid unnecessary translations. 

Note also the we have, for backwards compatibility, not just "member" & "follower" but also "config-tracker" & "none". The status is "active", "onboarding" and "inactive". So I wouldn't want to create all the combinations (as these might also evolve in the future).

```
type ChannelInfo struct {
	// The channel name.
	Name string `json:"name"`
	// The channel relative URL (no Host:Port, only path), e.g.: "/participation/v1/channels/my-channel".
	URL string `json:"url"`
	// Whether the orderer is a “member” or ”follower” of the cluster, or "config-tracker" of the cluster, for this channel.
	// For non cluster consensus types (solo, kafka) it is "none".
	// Possible values:  “member”, ”follower”, "config-tracker", "none".
	ClusterRelation string `json:"clusterRelation"`
	// Whether the orderer is ”onboarding”, ”active”, or "inactive", for this channel.
	// For non cluster consensus types (solo, kafka) it is "active".
	// Possible values:  “onboarding”, ”active”, "inactive".
	Status string `json:"status"`
	// Current block height.
	Height uint64 `json:"height"`
}
```

What I can do, and thought of doing but sustained, is to create constants in the types package, e.g.
```go
const (
	ChannelInfoClusterRelationMember string = "member"
)
```
and so on.

orderer/consensus/cluster_status.go
tock-ibm: see below

orderer/common/multichannel/registrar.go
tock-ibm: I am not sure what you mean... like so:
```go
info.Name, info.Height = channelID, cs.Height()
```

versus

```go
info.Name = channelID
info.Height = cs.Height()
```

yacovm: no, like: info := &InfoObject{ .... }

orderer/consensus/etcdraft/chain.go
yacovm: Constants will do I guess :+1:

tock-ibm: or perhaps something like 

```go

type ClusterRelation string
const (
	ClusterRelationMember ClusterRelation = "member"
)
```

yacovm: Yes this is better

orderer/common/multichannel/registrar.go
tock-ibm: there is already an empty info object above that I return in case of error, here I just fill it up... not sure what else I can do. This saves only a few characters anyway.

orderer/common/types/channelinfo.go
```diff
@@ -29,6 +29,23 @@ type ChannelInfoShort struct {
 	URL string `json:"url"`
 }
 
+type ClusterRelation string
+
+const (
+	ClusterRelationMember        ClusterRelation = "member"
+	ClusterRelationFollower      ClusterRelation = "follower"
+	ClusterRelationConfigTracker ClusterRelation = "config-tracker"
```
yacovm, line 44: what's a config tracker? I don't see anything about it in the RFC, and what's none for?

tock-ibm: This indeed was not in the RFC because the RFC did not address how this API would behave on a "legacy" system that runs in the system channel model. We have to add it to the RFC some time for completeness. 

The ClusterRelation="config-tracker" is the inactive.Chain. It shows up in the "List" of channels, and we have to differentiate it from the etcdraft.Chain which is "member"/"active". 

The ClusterRelation="none" is for kafka and solo. I know they are deprecated, but they are still in the code-base, so I addressed it.

It is all written on the documentation of the respective ChannelInfo.ClusterRelation field, but I guess it needs to be on the constant as well.

yacovm: Makes sense, thanks

yacovm (approved):

PR #1350, opened 2020-05-28
[FAB-17935] Change unnecessary warning log to debug in gossip privdata

caod123:
#### Type of change

#### Description

There are some unnecessary logger.Warning lines in the gossip logs when processing private data from a block.

#### Related issues

[FAB-17935](https://jira.hyperledger.org/browse/FAB-17935)

### DIFF
--- a/gossip/privdata/coordinator.go
+++ b/gossip/privdata/coordinator.go
@@ -432,7 +432,7 @@ func getTxInfoFromTransactionBytes(envBytes []byte) (*txInfo, error) {
 
 	if chdr.Type != int32(common.HeaderType_ENDORSER_TRANSACTION) {
 		err := errors.New("header type is not an endorser transaction")
-		logger.Warningf("Invalid transaction type: %s", err)
+		logger.Debugf("Invalid transaction type: %s", err)
 		return nil, err
 	}
 


gossip/privdata/coordinator.go
```diff
@@ -407,63 +409,51 @@ func getTxInfoFromTransactionBytes(envBytes []byte) (*txInfo, error) {
 	txInfo := &txInfo{}
 	env, err := protoutil.GetEnvelopeFromBlock(envBytes)
 	if err != nil {
-		logger.Warningf("Invalid envelope: %s", err)
-		return nil, err
+		return nil, fmt.Errorf("invalid envelope: %s", err)
```
ale-linux: There's a convention of using errors.Errorf from github.com/pkg/errors for this kind of error - any specific reason to use fmt here?

ale-linux (commented):
Thanks Danny, just a nit.

gossip/privdata/coordinator.go
```diff
@@ -407,63 +409,51 @@ func getTxInfoFromTransactionBytes(envBytes []byte) (*txInfo, error) {
...
 	}
 
 	payload, err := protoutil.UnmarshalPayload(env.Payload)
 	if err != nil {
-		logger.Warningf("Invalid payload: %s", err)
```
yacovm: I don't understand the motivation behind this change at all. All these errors are not things that should appear during normal operation. Why are they logged at debug? 

The JIRA details a single warning about a transaction not being an endorser type. 

Why shouldn't all the other errors be logged with warnings?

yacovm (changes requested):
See comment

gossip/privdata/coordinator.go
caod123: My understanding was the warning logs in this specific path seem extraneous to show up during regular gossip flow, especially since we skip the transactions that we fail to retrieve private data from anyways. The thought here being it's a debug statement to mention we're skipping a particular transaction in the block. I'm indifferent though and can understand leaving the other lines at warning, let me know if you strongly believe they should be warning and I can change it back.

gossip/privdata/coordinator.go
caod123: I believe `github.com/pkg/errors` isn't supporting any new functionality as go2 errors design is in process. As such eventually the package will no longer be updated/supported. The [standard](https://github.com/golang/go/wiki/Errors) now seems to be moving towards the standard go library pkg `errors` which does not have an `Errorf` formatting method and instead [defers to `fmt.Errorf`](https://golang.org/pkg/errors/#example_New_errorf) for formatted errors.

gossip/privdata/coordinator.go
yacovm: > The thought here being it's a debug statement to mention we're skipping a particular transaction in the block. 

My problem with it, is that we are skipping the transaction due to something that should never happen under normal operations. If you have transactions that have malformed fields inside of them, you would want to know that and tell it to someone or ask someone even if you aren't effected by it, because someone else is. 

I usually think of logging levels as something like:

- **Debug**: Logs low level details of the operation flow, such as operations we are doing and operations that are done to us, etc.
- **Info**: Logs events the user should care about from a high level point of view. 
- **Warning**: Something that shouldn't happen happened, it may or may not be a problem.
- **Error**: Something happened and is surely a problem. 
- **Fatal**: Something really bad happened and we cannot continue operation.

> let me know if you strongly believe they should be warning and I can change it back.

This can be a good opportunity to look through these and re-classify them.

gossip/privdata/coordinator.go
ale-linux: Ah thanks Danny - good to know!

denyeart:
I agree with Yacov - any unexpected issue should be logged as a Warning so that admins are made aware.
I agree that the original Warning mentioned in the Jira should be turned into a Debug, since it is an expected situation (for non endorser transactions).

caod123:
Sounds good, I'll change them back to warning

yacovm (approved):

PR #1353, opened 2020-05-29
[FAB-17900] Fixes env variable override bug

stephyee:
#### Type of change

- Bug fix
- Documentation update

#### Description

- Update documentation examples to include quotes around the `Pin` key. 
- Use WeakDecode to fix bug when setting a numeric environment variable override

~Supports Pin being specified as quoted string or integer in the core.yaml or orderer.yaml. Previously If specified as `9876543` an error would occur because mapstructure package decodes the value as an integer while fabric expects a string. To avoid breaking current string this supports quoted strings and integers.~

#### Related issues

[FAB-17900](https://jira.hyperledger.org/browse/FAB-17900)

### DIFF
--- a/common/viperutil/config_util.go
+++ b/common/viperutil/config_util.go
@@ -93,6 +93,7 @@ func getKeysRecursively(base string, getKey viperGetter, nodeKeys map[string]int
 
 func unmarshalJSON(val interface{}) (map[string]string, bool) {
 	mp := map[string]string{}
+
 	s, ok := val.(string)
 	if !ok {
 		logger.Debugf("Unmarshal JSON: value is not a string: %v", val)
@@ -303,7 +304,7 @@ func bccspHook(f reflect.Type, t reflect.Type, data interface{}) (interface{}, e
 
 	config := factory.GetDefaultOpts()
 
-	err := mapstructure.Decode(data, config)
+	err := mapstructure.WeakDecode(data, config)
 	if err != nil {
 		return nil, errors.Wrap(err, "could not decode bcssp type")
 	}
--- a/docs/source/hsm.md
+++ b/docs/source/hsm.md
@@ -53,7 +53,7 @@ bccsp:
   default: PKCS11
   pkcs11:
     Library: /etc/hyperledger/fabric/libsofthsm2.so
-    Pin: 71811222
+    Pin: "71811222"
     Label: fabric
     hash: SHA2
     security: 256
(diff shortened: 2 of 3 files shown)


sykesm:
Why? It is a string. It should be read into a string field and it should be a string in yaml.

sykesm:
This is another example of how broken the configuration model is. The configuration structures all appear to have the pin properly typed as a string yet, because of the crazy mechanics of `EnhancedExactUnmarshal`, things break.

Instead of saying this enables a number or a string, the commit should say that this fixes a bug in handling the intermediate data during config processing. The pin defined as a string and should be configured as a string. That means, in particular, that the yaml rendering should require quotes around it.

When it comes to the environment variables, that’s when we get into the (broken) land of viper and all of its nonsense.

EDIT: Turns out this little to do with the intermediate processing. The initial value was read as a number instead of a string.

EDIT 2: There were two parts to this and the intermediate processing is getting in the way as any value loaded by viper from the environment is typed as a string.

stephyee:
@sykesm The other easy option considered is to document the requirement of quotes around the pin in the yaml.

sykesm:
> @sykesm The other easy option considered is to document the requirement of quotes around the pin in the yaml.

I believe that is the correct answer. Values need to be quoted when the type cannot be correctly inferred. This is true of numbers and bools in particular where `123 != “123”` and `”true” != true`.

I just verified that without these changes, when the value is quoted in the yaml, the value is properly treated as a string. This is also true when provided in an environment override without any decorations `ORDERER_GENERAL_BCCSP_PKCS11_PIN=33333`.

stephyee:
@sykesm I'll document where necessary then.

sykesm:
> > @sykesm I'll document where necessary then.
> 
> I may be wrong but I think it’s less documenting yaml than fixing all of the bad examples. The documentation has unquoted scalar pins all over the place.

I looked at the issue and it’s not just the pin; apparently attempting to set a numeric value in an environment variable override breaks. That looks like a real bug and I’ll assume that’s related to `viperutil`.

common/viperutil/config_util.go
```diff
@@ -303,7 +304,7 @@ func bccspHook(f reflect.Type, t reflect.Type, data interface{}) (interface{}, e
 
 	config := factory.GetDefaultOpts()
 
-	err := mapstructure.Decode(data, config)
+	err := mapstructure.WeakDecode(data, config)
```
sykesm, line 307: The one bit you may want to consider is whether or not to carry the `ErrorUnused` option here.

sykesm (approved):

common/viperutil/config_util.go
stephyee: 🤔I think of it as extra validation but not sure if it harms or helps.

caduellery:
Any chance this get backported to 1.4?

PR #1354, opened 2020-05-29
Remove s390x, powerpc64le from RELEASE_PLATFORMS

sykesm:
Things have changed with release-2.0. In particular,
- Release automation only creates amd64 binaries for darwin, linux, and windows.
- Continuous integration no longer runs on powerpc64le or s390x
Given we’re not producing release artifacts for those two platforms and we are not performing automated tests in CI for them, we can remove the references from the build process.

If we get to the point where we want to/need to release or test on these platforms again, this will force us to re-evaluate the required changes at that time. This is a good thing as it will prevent us from carrying baggage that we don’t understand.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -81,7 +81,7 @@ GO_TAGS ?=
 
 RELEASE_EXES = orderer $(TOOLS_EXES)
 RELEASE_IMAGES = baseos ccenv orderer peer tools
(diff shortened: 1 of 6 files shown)


lindluni:
do you want to keep the ppc64 build tags still?

sykesm:
> do you want to keep the ppc64 build tags still?

No. Just removed them.

lindluni (approved):

PR #1363, opened 2020-06-02
Add ExtractMSPIDsForApplicationOrgs util method to channelconfig

wenjianqiao:

#### Type of change

#### Description
Add a util method to extract MSPIDs for applications from a config block.

#### Additional details
This util method will be called by ledger to create checkpoint snapshots and remove per-channel databases when using  state couchdb.

#### Related issues
https://jira.hyperledger.org/browse/FAB-17942

### DIFF
--- a/common/channelconfig/util.go
+++ b/common/channelconfig/util.go
@@ -18,6 +18,7 @@ import (
 	"github.com/hyperledger/fabric-protos-go/orderer/etcdraft"
 	pb "github.com/hyperledger/fabric-protos-go/peer"
 	"github.com/hyperledger/fabric/bccsp"
+	"github.com/hyperledger/fabric/bccsp/factory"
 	"github.com/hyperledger/fabric/protoutil"
 	"github.com/pkg/errors"
 )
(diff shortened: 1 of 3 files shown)


common/channelconfig/util.go
```diff
@@ -241,51 +242,72 @@ func ACLValues(acls map[string]string) *StandardConfigValue {
...
 	configEnv := &cb.ConfigEnvelope{}
 	_, err = protoutil.UnmarshalEnvelopeOfType(envelopeConfig, cb.HeaderType_CONFIG, configEnv)
 	if err != nil {
-		return errors.Errorf("malformed configuration envelope: %s", err)
+		return nil, errors.Errorf("malformed configuration envelope: %s", err)
 	}
 
 	if configEnv.Config == nil {
-		return errors.New("nil config envelope Config")
+		return nil, errors.New("nil config envelope Config")
```
lindluni: Are there other types of config envelopes, if not, config is a stutter, should be `nil config envelope`, if there are can we rephrase it to `nil config envelope of type Config`

common/channelconfig/util.go
```diff
@@ -241,51 +242,72 @@ func ACLValues(acls map[string]string) *StandardConfigValue {
...
 	if configEnv.Config.ChannelGroup.Groups == nil {
-		return errors.New("no channel configuration groups are available")
+		return nil, errors.New("no channel configuration groups are available")
 	}
 
 	_, exists := configEnv.Config.ChannelGroup.Groups[ApplicationGroupKey]
 	if !exists {
-		return errors.Errorf("invalid configuration block, missing %s "+
+		return nil, errors.Errorf("invalid configuration block, missing %s "+
 			"configuration group", ApplicationGroupKey)
```
lindluni: We are already using Errorf, can we remove the concatenation

common/channelconfig/util.go
```diff
@@ -241,51 +242,72 @@ func ACLValues(acls map[string]string) *StandardConfigValue {
...
 
 	cc, err := NewChannelConfig(configEnv.Config.ChannelGroup, bccsp)
 	if err != nil {
-		return errors.Errorf("no valid channel configuration found due to %s", err)
-	}
-
-	// Check the channel top-level capabilities
-	if err := cc.Capabilities().Supported(); err != nil {
-		return err
+		return nil, errors.Errorf("no valid channel configuration found due to %s", err)
```
lindluni: Nit: this isn't a very idiomatic error statement, in general we would wrap the context around the error, not use phrasing like "due to"

lindluni (commented):
Couple nits on the logging statements

common/channelconfig/util.go
manish-sethi: Moreover, the use of `errors.Errorf` will cause loosing the stack trace of the actual underlying error. On a general pattern for using `"github.com/pkg/errors"`, we should use
- `New / Errorf`, when constructing the original error in the fabric code (so that the stack trace is generated)
- `Wrapf` when propagating the error received from an external library (so that stack trace is generated)
- Finally, for propagating the errors from fabric apis (as is the case here, and majority of the fabric code), we should either forward as is, if we don't have much to add to the context. If we want to add some context, we should use `WithMessagef` so that the full stack trace is not lost.

common/channelconfig/util.go
caod123: this could be something like `envelope missing config`

common/channelconfig/util.go
caod123: fwiw we may want to change this to `fmt.Errorf` simply for the reason that `github.com/pkg/errors` will no longer support new features and the shift is towards using golang's `errors` package and delegating formatted errors to `fmt.Errorf`. I'm indifferent though if we want to maintain the stack trace as Manish mentions below.

common/channelconfig/util.go
wenjianqiao: @btl5037 @manish-sethi Thank you. This method was simply split from the existing ValidateCapabilities method for code sharing. Therefore, all the error handlings were kept unchanged from the original code. Having said so, definitely agree with your comments. Will fix it.

common/channelconfig/util.go
wenjianqiao: @btl5037 thanks for pointing it out. This is existing code, but should be fixed since we are changing it anyway.
@caod123 thanks for pointing it out. In this PR, I prefer to keep errors.Errorf to be consistent with the rest of code. If it is decided to change to fmt.Errorf, we may want to have a separate PR for the change.

common/channelconfig/util.go
wenjianqiao: @btl5037 thanks for pointing it out. This is existing code, but should be fixed since we are changing the line anyway.

wenjianqiao:
Hit gossip UT flake: https://jira.hyperledger.org/browse/FAB-17451

caod123 (approved):

PR #1364, opened 2020-06-02
Documentation update: container name description

aeharvlee:
When execute fabric-samples/fabcar/startFabric.sh, it creates CA container which has name ca_org1, ca_org2, ca_orderer, not ca_peerOrg1.

#### Type of change

- Bug fix
- New feature
- Test update
- Documentation update

### DIFF
--- a/docs/source/write_first_app.rst
+++ b/docs/source/write_first_app.rst
@@ -163,7 +163,7 @@ Enrolling the admin user
 .. note:: The following two sections involve communication with the Certificate
           Authority. You may find it useful to stream the CA logs when running
           the upcoming programs by opening a new terminal shell and running
-          ``docker logs -f ca_peerOrg1``.
+          ``docker logs -f ca_org1``.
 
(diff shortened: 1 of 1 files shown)


lindluni:

This doesn't appear to be valid on the release-1.4 branch. I just checked and on 1.4 the ca is in fact named `ca_peerOrg1`. There is no mention of `ca_org1` in the entire fabric-samples repo on the 1.4 branch.

However, this change does appear to be valid for the 2.x branches where the fabric-samples Fabcar example has migrated to the `test-network` from the `first-network`, so `release-2.1`, `release-2.0` and `master`. This change should be resubmitted against `master` and then we will automate the backport.

You will also need to sign the commit using `git commit --signoff` before submitting the change. Due to the requirement of having a DCO signature, if you submit a change through the GitHub UI editor, you need to pull the Pull Request locally and manually add the signoff, or perform the change locally in the first place.

PR #1366, opened 2020-06-03
Sort JSON value to ensure deterministic stateDB snapshot hash across peers

cendhu:
#### Type of change

- New feature

#### Description

We need to ensure that the json document written to the snapshot file is the same across peers. Hence, this commit adds an utility function to sort json.

#### Additional details

#1348 will use this utility function while exporting the data out of CouchDB

#### Related issues

[FAB-17901](https://jira.hyperledger.org/browse/FAB-17901)

### DIFF
--- a/core/ledger/kvledger/txmgmt/statedb/statecouchdb/testdata/json_documents/6_unsorted.json
+++ b/core/ledger/kvledger/txmgmt/statedb/statecouchdb/testdata/json_documents/6_unsorted.json
@@ -1,0 +1,36 @@
+{
+  "ns_read_write_Set": [
+    {
+      "Namespace": "lscc",
+      "KVRWSet": {
+        "reads": [
+          {
+            "key": "mycc1"
+          }
+        ],
+        "writes": [
+          {
+            "key": "mycc1",
+            "value": "\n\u0005mycc1\u0012\u00010\u001a\u0004escc\"\u0004vscc*\u0017\u0012\b\u0012\u0006\b\u0001\u0012\u0002\b\u0000\u001a\u000b\u0012\t\n\u0007DEFAULT2D\n ��$˞�K݂o�pJ{��_��(\u0015ݮ�>\u001a�c�\tS�\u0012 \u0012V�1\u000e�\u000e��\u0010Q�i�$�VtV\\�\u0018L(�I����F�: E��ؿk�<\u0015�x��F\r�\u0005��C9\u0001�\t�bA\u000e\u0018�5�B\u0019\u0012\b\u0012\u0006\b\u0001\u0012\u0002\b\u0000\u001a\r\u0012\u000b\n\u0007DEFAULT\u0010\u0001"
+          }
+        ]
+      }
+    }
+  ],
+  "chaincode_spec": {
+    "type": 1,
+    "chaincode_id": {
+      "name": "lscc"
+    },
+    "input": {
+      "Args": [
+        "deploy",
+        "ch1",
+        "\nU\b\u0001\u0012I\n=github.com/hyperledger/fabric/examples/chaincode/go/marbles02\u0012\u0005mycc1\u001a\u00010\u001a\u0006\n\u0004init",
+        "",
+        "escc",
+        "vscc"
+      ]
+    }
+  }
+}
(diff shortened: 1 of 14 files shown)


core/ledger/kvledger/txmgmt/statedb/statecouchdb/testdata/json_documents/6_unsorted.json
```diff
@@ -0,0 +1,36 @@
...
+  "chaincode_spec": {
+    "type": 1,
+    "chaincode_id": {
+      "name": "lscc"
+    },
+    "input": {
+      "Args": [
+        "deploy",
+        "ch1",
+        "\nU\b\u0001\u0012I\n=github.com/hyperledger/fabric/examples/chaincode/go/marbles02\u0012\u0005mycc1\u001a\u00010\u001a\u0006\n\u0004init",
```
cendhu, line 29:
```suggestion
"\nU\b\x01\x12I\n=github.com/hyperledger/fabric/examples/chaincode/go/marbles02\x12\x05mycc1\x1a\x010\x1a\x06\n\x04init",
```
`-  Expected`
`+ Actual`

@manish-sethi This is the only case that fails. It is due to `strconv.Quote(v.String())` that I use in `sortJSON()` to escape `"`.  If you have some free time, can you please look into this and give suggestion? Otherwise, I can get back to this on Friday. This PR blocks #1348

manish-sethi: @cendhu - As far as I understand, JSON arrays are arrays and hence are supposed to maintain the original order. As we discussed, a simple canonical representation should suffice. I am not sure beforehand whether `encoding/json` will do the job or the other canonical marshal library that we saw would be needed. Moreover, I think that this is an ideal place to explore the usage of binary formatted JSON as, this exported data is not meant for human readable and we can gain from a binary JSON. Let's discuss when you are back.

cendhu: > As far as I understand, JSON arrays are arrays and hence are supposed to maintain the original order.

I too think so. Just to avoid assumptions as I couldn't verify this, I went with sorting array elements too. If this is not needed, just a simple ` json.Marshal()` does the job for us as used in #1348 [line](https://github.com/hyperledger/fabric/pull/1348/files#diff-83bacdb7062a79abb0348662caf133a8R117) 

` json.Marshal()` takes care of sorting the map keys [recursively](https://github.com/golang/go/blob/master/src/encoding/json/encode.go#L137-L139) and I have verified that using these test cases. 

The goal is not to export human readable content. Only for the ease of testing and code readability, I sort the human readable JSON documents. 

To ensure that the config block fetched from the DB is deterministic across peers, this will not help anyway and needs to bring one of the abandoned CR from https://github.com/cendhu/fabric-gerrit/blob/master/fabric/29678-29852/29725.md

manish-sethi: > > As far as I understand, JSON arrays are arrays and hence are supposed to maintain the original order.
> 
> I too think so. Just to avoid assumptions as I couldn't verify this, I went with sorting array elements too. If this is not needed...

Not only not needed but rather, it would be wrong. Because, in an application it is legitimate to access the data by array index, which would be changed by applying sorting at that level.

cendhu: I agree. Then, we will go with `json.Marshal()`. I can close this PR for now. We can do rigorous ledger-level integration testing in the `kvledger/tests` and can get back here if we find any issues.

manish-sethi: Regarding the other point in my original message, I feel that it's definitely worth considering (and exploring a bit) for Binary JSON as a replacement of using `json.Marshal()`. That would be highly efficient both in exporting and later in importing the data.

cendhu: Okay. For the time being, #1348 will use `json.Marshal()`. Once it is merged, I will explore the binary JSON to improvement the performance.

PR #1367, opened 2020-06-03
FAB-17951 fetch correct node id for validation

guoger:
Today, when a config update that modifies consenter set (add/remove
node), there's validator performing basics checks, including dangerous
configs that may result in quorum loss, e.g. removing active node from
a network with 2/3 alive nodes.

However, this validator always assumes node ID in consenter set to start
from 1. This is benign in most cases, as nodes are normally online, and
actual node ID doesn't really matter from validation pov. Although, in
certain special cases, this is problematic:
- we have [2, 3, 4] in consenter set
- [2, 3] are alive, and [4] is inactive and subject to remove
- if validator assume node starts from 1, then it would incorrectly conclude
  that [3] out of [1, 2, 3] is to be removed, while [2, 3] are the 2/3 alive
  nodes. Therefore, it would reject such request
- instead, we need to take actual node IDs into account

This commit fixes the problem and adds some UTs. IT will be added in a seperate
commit.

### DIFF
--- a/integration/nwo/network.go
+++ b/integration/nwo/network.go
@@ -1171,10 +1171,11 @@ func (n *Network) BrokerGroupRunner() ifrit.Runner {
 
 // OrdererRunner returns an ifrit.Runner for the specified orderer. The runner
 // can be used to start and manage an orderer process.
-func (n *Network) OrdererRunner(o *Orderer) *ginkgomon.Runner {
+func (n *Network) OrdererRunner(o *Orderer, env ...string) *ginkgomon.Runner {
 	cmd := exec.Command(n.Components.Orderer())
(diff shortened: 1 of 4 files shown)


sykesm (changes requested):
I can't comment on the first commit as it's totally not an area of expertise for me. On the second commit (the integration test), I think we can get rid of a bunch of boilerplate with a simple modification.

guoger:
@sykesm good suggestions, appreciated!

sykesm (approved):
LGTM but I really think @jyellick or @yacovm needs to review...

jyellick (approved):
Makes sense and looks good to me.

PR #1369, opened 2020-06-03
[FAB-17795] Build channel metadata retroactively if not present at peer start

wenjianqiao:

#### Type of change
- New feature

#### Description
Channel metadata has been added to statecouchdb to store the mapping of db namespaces
to db names. However, the channel metadata is not available in v2.0/v2.1 peers. This PR
adds support to retroactively build channel metadata at peer start if such metadata
is not present in statecouchdb.

When a state CouchDB is opened (GetDBHandle), it retroactively builds channel metadata if
the metadata is not present. A NamespaceProvider is implemented to build possible namespaces for the channel. The possible namespaces will be verified by the existing databases and only namespaces matching existing databases will be added to channel metadata.

#### Additional details
The channel metadata will be used to remove per-channel databases and support checkpoint snapshots (e.g., create snapshot, clean up after bootstrap failure)

#### Related issues
https://jira.hyperledger.org/browse/FAB-17795


core/ledger/kvledger/txmgmt/statedb/statecouchdb/statecouchdb.go
```diff
@@ -837,6 +848,32 @@ func (vdb *VersionedDB) GetFullScanIterator(skipNamespace func(string) bool) (st
...
+		dbsInfo[ns] = &namespaceDBInfo{
+			Namespace: ns,
+			DBName:    constructNamespaceDBName(vdb.chainName, ns),
+		}
+	}
+	vdb.channelMetadata = &channelMetadata{
+		ChannelName:      vdb.chainName,
+		NamespaceDBsInfo: dbsInfo,
+	}
+	return vdb.writeChannelMetadata()
```
cendhu: Shouldn't we check for the database existence and filter certain namespaces?

By default, we create a database for a chaincode or collection only when it is accessed first (applicable for implicit collection too). Hence, if we create mapping for all received namespaces and write the channelMetadata, it would end up an inconsistent data. To avoid this inconsistency, we need to check whether a db exist for a given namespace and only then add it to the channel metadata.

cendhu (commented):
Thanks, Wenjian. I have suggestions related to the overall approach.

core/ledger/kvledger/txmgmt/privacyenabledstate/namespace_helper.go
```diff
@@ -0,0 +1,121 @@
...
+
+	// add namespaces for implicit collections
+	mspids, err := channelInfoProvider.GetAllMSPIDs()
+	if err != nil {
+		return nil, err
+	}
+	implicitCollNames := deployedccInfoProvider.ImplicitCollectionNames(mspids)
+	for _, implicitCollName := range implicitCollNames {
+		retNamespaces = append(retNamespaces, derivePvtDataNs("_lifecycle", implicitCollName))
+		retNamespaces = append(retNamespaces, deriveHashedDataNs("_lifecycle", implicitCollName))
```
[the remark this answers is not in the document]
wenjianqiao: Thank you for pointing it out. Fixed.

core/ledger/kvledger/txmgmt/privacyenabledstate/namespace_helper.go
```diff
@@ -0,0 +1,121 @@
...
+		return nil, nil
+	}
+	versionedKV := queryResult.(*statedb.VersionedKV)
+	return &queryresult.KV{Namespace: versionedKV.Namespace, Key: versionedKV.Key, Value: versionedKV.Value}, nil
+}
+
+// Close implements method in interface ledger.ResultsIterator
+func (itr *resultsItr) Close() {
+	itr.dbItr.Close()
+}
```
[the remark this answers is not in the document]
wenjianqiao: Since statecouchdb returns *statedb.VersionedKV but the caller expects *queryresult.KV, the data type conversion is needed.

wenjianqiao:
@cendhu Thank you for your comments. As discussed in our scrum, I have refactored the code and moved channelMetadata rebuild to newVersionedDB. I have addressed your comments unless the code that has been removed due to refactoring. 
The SimpleQueryExecutor and results iterator wrapper structs are remained the same because it requires refactoring.

core/ledger/kvledger/txmgmt/statedb/statecouchdb/couchdb.go
```diff
@@ -459,6 +459,38 @@ func (couchInstance *couchInstance) retrieveApplicationDBNames() ([]string, erro
 	return applicationsDBNames, nil
 }
 
+// retrieveDBNamesMatchedForChannel returns existing db names whose channel segment matches
+// either the original channel name or the truncated channel name.
+// Note that false matching is possible if the channel name exceeds chainNameAllowedLength
+// because another channel may be truncated to the same name. Therefore, this method may
+// may return more than the channel's DB names.
+func (couchInstance *couchInstance) retrieveDBNamesMatchedForChannel(channelName string) (map[string]struct{}, error) {
```
cendhu: Is this function needed?

Instead of allowing the false matching while trying to approximately fetch DBs associated with a channel, can't we simply use `retrieveApplicationDBNames()` at the `initChannelMetadata()` and remove this function. The logic in `initChannelMetadata()` takes care of finding the correct mapping between the namespace and db. This additional complexity does not seem necessary.

core/ledger/kvledger/txmgmt/statedb/statecouchdb/statecouchdb.go
```diff
@@ -228,6 +215,10 @@ func newVersionedDB(couchInstance *couchInstance, redoLogger *redoLogger, dbName
 		return nil, err
 	}
 
+	if err = vdb.initChannelMetadata(savepoint == nil, nsProvider); err != nil {
```
cendhu: To improve the readability, we can have

```
isNewDB := savepoint == nil
if err = vdb.initChannelMetadata(isNewDB, nsProvider)
```

This helps in providing a context for `savepoint == nil`.

cendhu (commented):
Thanks, Wenjian, for the changes. I have a few more suggestions. Once they are addressed, I will look at the test.

core/ledger/kvledger/txmgmt/privacyenabledstate/namespace_provider.go
```diff
@@ -0,0 +1,139 @@
...
+	}
+
+	// iterate each chaincode, add namespaces for chaincodes, implicit collections, and explicit collections
+	retNamespaces := []string{}
+	for _, ccInfo := range chaincodesInfo {
+		ccName := ccInfo.Name
+		retNamespaces = append(retNamespaces, ccName)
+		for _, implicitCollName := range implicitCollNames {
+			retNamespaces = append(retNamespaces, derivePvtDataNs(ccName, implicitCollName))
+			retNamespaces = append(retNamespaces, deriveHashedDataNs(ccName, implicitCollName))
```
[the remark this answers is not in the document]
wenjianqiao: @cendhu thanks for the comment. Initially I thought it would require new dependency to check peer membership for a collection, but just found that ledger already had MembershipProvider interface. Will add code to check eligibility before adding private data implicit/explicit collections.

core/ledger/kvledger/txmgmt/statedb/statecouchdb/couchdb.go
wenjianqiao: Since we need to build a map to look up DBNames, this function builds the map with minimum entries to make dbname lookup more efficient. However, because this is one time call for v2.0/2.0 peers only, I agree we don't need the additional complexity.

core/ledger/kvledger/txmgmt/privacyenabledstate/namespace_provider.go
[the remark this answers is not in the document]
cendhu: For an implicit collection, this might work. For explicit collections, we only fetch the recent collection config using `DeployedCCInfoProvider.AllChaincodeInfo`. To find all members (including the past members) of a collections, we need to get all collection config from the collection config store. I am not sure whether this complexity is needed. Hence, I am okay if we can add a comment to avoid this confusion (as it is a temporary code).

core/ledger/kvledger/txmgmt/privacyenabledstate/namespace_provider.go
```diff
@@ -0,0 +1,146 @@
...
+	// add special namespaces: "" and lifecycle management namespaces (new and legacy)
+	retNamespaces = append(retNamespaces, "")
+	retNamespaces = append(retNamespaces, p.Namespaces()...)
+
+	return retNamespaces, nil
+}
+
+// simpleQueryExecutor implements ledger.SimpleQueryExecutor interface
+type simpleQueryExecutor struct {
+	vdb statedb.VersionedDB
```
manish-sethi: if you embed `statedb.VersionedDB`, you don't need to implement first and third functions below.

core/ledger/kvledger/txmgmt/privacyenabledstate/namespace_provider.go
```diff
@@ -0,0 +1,146 @@
...
+// only if the peer is a member of the collection or owns the implicit collection. However, this function
+// adopts a simple implementation that always adds private data namespace for a collection without checking
+// peer membership/ownership. As a result, it may return more private data namespaces that what should be.
+// However, it will not cause any inconsistent issue because the caller in statecouchdb will check if any
+// existing database matches the namespace and filter out all extra namespaces since no databases match them.
+// Checking peer membership is complicated because it requires retrieving all the collection configs from
+// the collection config store. Because this is a temporary function needed to retroactively build namespaces
+// when upgrading v2.0/2.1 peers to a newer v2.x version and because returning extra private data namespaces
+// does not cause inconsistence, it makes sense to use the simple implementation.
+func (p *NamespaceProvider) GetStateDBNamespaces(channelName string, vdb statedb.VersionedDB) ([]string, error) {
```
manish-sethi: In the connection with the comment in `kv_ledger_provider`, after looking at this code here, I feel that the logic in this funciton better suits in the package `kvledger`. That would prevent this high level logic to percolate much deeper in the code - wdyt?

manish-sethi (commented):
The logic looks good to me. However, I have a couple comments, mostly related to reorganizing the code a bit for a better separation of concerns.

core/ledger/kvledger/txmgmt/privacyenabledstate/namespace_provider.go
cendhu: Due to the dependency on `deriveHashedDataNs()` and `derivePvtDataNs()`, it may be bit challenging to move the whole `GetStateDBNamespaces()` to `kvledger` pkg.

wenjianqiao: This function has logic (e.g. derivePvtDataNs) specific to the `privacyenabledstate` package, therefore cannot be moved to kvledger. Will split it into two functions: one is moved to channelInfoProvider in the `kvledger` package and the other stays in `privacyenabledstate` package.

core/ledger/kvledger/txmgmt/privacyenabledstate/namespace_provider.go
wenjianqiao: Changed to embed statedb.VersionedDB. But 1st and 3rd functions are needed because they overwrite the same functions in VersionedDB.

core/ledger/kvledger/txmgmt/statedb/statedb.go
```diff
@@ -77,6 +78,12 @@ type VersionedDB interface {
 	Close()
 }
 
+// NamespaceProvider provides a mean to get all the namespaces for a channel. The intended use is for statecouchdb
+// to retroactively build channel metadata when it is missing (e.g., a statecouchdb from an older version)
+type NamespaceProvider interface {
+	GetStateDBNamespaces(channelName string, vdb VersionedDB) ([]string, error)
+}
+
```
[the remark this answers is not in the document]
wenjianqiao: Thank you @manish-sethi. Agree defining a function may be simpler in some cases. However, will keep the interface for this one, because we need to mock it multiple times in test code, and it is easier to mock an interface.

manish-sethi (approved):
Minor comments, if you agree and want to address in a separate PR.

PR #1370, opened 2020-06-03
Update "peer channel update" to allow disabling of signatures

wlahti:
#### Type of change

- New feature

#### Description

Allow disabling of automatic signatures in "peer channel update". This will allow us to truly verify
the functionality of the fabric-config/configtx library.

A follow-on PR (or commit) will be pushed to utilize this in the configtx integration test suite.

#### Related issues

[FAB-17948](https://jira.hyperledger.org/browse/FAB-17948)


ale-linux (commented):
I'd like to see what others think about this, but it seems to me that this PR introduces in our CLI a function that is only meaningful for tests. Rather than doing this, we could just "manually" do what peer channel update does (minus the signing) in the test package - there's a precedent for something similar to that for the lifecycle work.

sykesm:
> I'd like to see what others think about this, but it seems to me that this PR introduces in our CLI a function that is only meaningful for tests. Rather than doing this, we could just "manually" do what peer channel update does (minus the signing) in the test package - there's a precedent for something similar to that for the lifecycle work.

While I agree in principle, I do think we could come up with some valid scenarios where sending a configuration update does not require a signature from the submitter. Basically, I don't have a problem with this change even though it's being proposed to help with a test scenario.

ale-linux:
hmm, I can't really think of any legitimate scenario where we would be sending an unsigned config update. I'd rather invest in extending our integration test machinery with what the CLI would do with this change - send a blob to the ordering service. I did it once to manually test a corner case, it wasn't too painful. That said, if the consensus is that extending the CLI this way is a good idea, I won't stand in the way.

cendhu:
> hmm, I can't really think of any legitimate scenario where we would be sending an unsigned config update. I'd rather invest in extending our integration test machinery with what the CLI would do with this change - send a blob to the ordering service. I did it once to manually test a corner case, it wasn't too painful. That said, if the consensus is that extending the CLI this way is a good idea, I won't stand in the way.

If there is no valid scenario, I agree with you @ale-linux To enable testing of some other repo `fabric-config/configtx`, it is not good to change the `peer` CLI that is designated for the production use by the admin/user. Further, `--sign=true/false` also could create unnecessary confusion for the CLI user when there is no real use-case as you say.

sykesm:
> hmm, I can't really think of any legitimate scenario where we would be sending an unsigned config update.

My understanding is that it has been signed by other parties. If that's wrong, then I agree.

wlahti:
The goal is to allow signing the config update (creating a ConfigSignature) and the envelope using the configtx library instead relying on the CLI's automatic signing to create the ConfigSignature and envelope signature.

yacovm:
> The goal is to allow signing the config update (creating a ConfigSignature) and the envelope using the configtx library instead relying on the CLI's automatic signing to create the ConfigSignature and envelope signature.

What's so special about testing this that it requires testing it in an integration test?

wlahti:
> > The goal is to allow signing the config update (creating a ConfigSignature) and the envelope using the configtx library instead relying on the CLI's automatic signing to create the ConfigSignature and envelope signature.
> 
> What's so special about testing this that it requires testing it in an integration test?

We can't verify that the signatures produced by the configtx library are valid when the CLI automatically adds another config signature and envelope signature.

sykesm:
@wlahti it looks like most people would prefer to just leave the CLI as it is. I'd suggest closing this and implementing it as code in the test.

yacovm:
> > > The goal is to allow signing the config update (creating a ConfigSignature) and the envelope using the configtx library instead relying on the CLI's automatic signing to create the ConfigSignature and envelope signature.
> > 
> > 
> > What's so special about testing this that it requires testing it in an integration test?
> 
> We can't verify that the signatures produced by the configtx library are valid when the CLI automatically adds another config signature and envelope signature.

1. Why not? 
2. I'm not sure an integration test is the most suitable way of testing such low level property. This should be tested in the configtx library unit tests.

PR #1378, opened 2020-06-08
Update configtx dependency and integration test

wlahti:
#### Type of change

- Bug fix
- Test update

#### Description

- Bump fabric-config/configtx dependency to v0.0.4.
- Update configtx integration test to use nwo.Broadcast for submitting envelopes.
- Add tests to update orderer and other application configuration.

#### Related issues

[FAB-17956](https://jira.hyperledger.org/browse/FAB-17956)

### DIFF
--- a/integration/configtx/configtx_test.go
+++ b/integration/configtx/configtx_test.go
@@ -15,18 +15,18 @@ import (
 	"os"
 	"strconv"
 	"syscall"
+	"time"
 
 	docker "github.com/fsouza/go-dockerclient"
 	"github.com/golang/protobuf/proto"
 	"github.com/hyperledger/fabric-config/configtx"
+	"github.com/hyperledger/fabric-protos-go/common"
 	"github.com/hyperledger/fabric/integration/nwo"
 	"github.com/hyperledger/fabric/integration/nwo/commands"
 	"github.com/tedsuo/ifrit"
 
 	. "github.com/onsi/ginkgo"
 	. "github.com/onsi/gomega"
-	"github.com/onsi/gomega/gbytes"
-	"github.com/onsi/gomega/gexec"
 )
 
 var _ = Describe("ConfigTx", func() {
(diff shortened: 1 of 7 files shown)


integration/configtx/configtx_test.go
```diff
@@ -115,8 +114,9 @@ var _ = Describe("ConfigTx", func() {
 		}
 
 		channelID := "testchannel"
-		createChannelUpdate, err := configtx.NewCreateChannelTx(channel, channelID)
+		createChannelUpdate, err := configtx.NewMarshaledCreateChannelTx(channel, channelID)
```
sykesm, line 117: This seems to be associated with the previous commit, right? If we bump the dependency without this change, will the code build? If not, this should be pulled into the previous commit.

Our CI (for better or worse) does not build and test each commit independently so it's easy to have a broken commit that could bite us during bisects or other postmortem activities.

wlahti: Good to know. Squashed the two commits together.

integration/configtx/configtx_test.go
```diff
@@ -140,81 +140,190 @@ var _ = Describe("ConfigTx", func() {
...
+
+		By("updating orderer channel configuration")
+		o := c.Orderer()
+		oConfig, err := o.Configuration()
+		Expect(err).NotTo(HaveOccurred())
+		oConfig.BatchTimeout = 2 * time.Second
+		err = o.SetConfiguration(oConfig)
+		Expect(err).NotTo(HaveOccurred())
+		host, port := ordererHostPort(network, orderer)
+		err = o.Organization(orderer.Organization).SetEndpoint(configtx.Address{Host: host, Port: port + 1})
```
sykesm, line 159: Did you really want to use `port + 1` or get another reserved port from the network?

sykesm (approved):
LGTM; I'll let you decide if adding a port to a possibly active server is okay and rely on @caod123 to merge as appropriate.

integration/configtx/configtx_test.go
caod123: I believe the idea was that this port wasn't actually going to be used for anything hence no need to get another reserved port. I think this is fine as is.

PR #1380, opened 2020-06-08
[FAB-16846] Fix OU validation

stephyee:
#### Type of change

- Bug fix

#### Description

WIP,
- when OU enforcement is disabled validate specified identities  
- when OU enforcement is enabled validate identity to ensure only one special OU exists 

#### Related issues

[FAB-16846](https://jira.hyperledger.org/browse/FAB-16846)


msp/mspimplvalidate.go
```diff
@@ -170,18 +170,15 @@ func (msp *bccspmsp) validateIdentityOUsV1(id *identity) error {
 }
 
 func (msp *bccspmsp) validateIdentityOUsV11(id *identity) error {
-	// Run the same checks as per V1
```
yacovm, line 173: There are 2 distinct OU validations:
1. [ouEnforcement](https://github.com/hyperledger/fabric/blob/a282f7132565d509d3ba6d502cd9c07cc5e567f0/msp/mspimplsetup.go#L253) means that the MSP classifies node types (peer, orderer, etc.) according to OUs. 
2. OUs that classify this MSP as defined [here](https://github.com/hyperledger/fabric/blob/a282f7132565d509d3ba6d502cd9c07cc5e567f0/msp/mspimplsetup.go#L390-L419).

The function [validateIdentityOUsV1](https://github.com/hyperledger/fabric/blob/f2dbb66542cf0520391c3388d7d2f1fb9662ff5f/msp/mspimplvalidate.go#L142) performs the second check. 

Before your proposed change, all identities had to comply to the second validation rule.

By doing the change you are proposing we are essentially saying that the first rule replaces the second one, but they can coexist. 

Aside for this, we cannot change the semantics of past version MSP validation functions such as this, because they have the potential to fork the network:

Consider a peer **peer0** that exists in the network for a year, and that has decided that a certain transaction is invalid because it didn't pass **validateIdentityOUsV1**, but it would have passed the other checks down this function, and further assume that **ouEnforcement** is activated during the transaction validation of that transaction.

Now, imagine your commit is merged and the network is upgraded to the latest version of Fabric that includes your commit, and let **peer1** be a peer that joined to the network after this upgrade.
When **peer1** executes **validateIdentityOUsV11** for that transaction, the transaction would be deemed valid, and the state updates of that transaction would be applied on **peer1** in contrast to **peer0**, and we have reached a state fork.

yacovm (changes requested):

yacovm:
I don't see why the problem that the user reported in the JIRA is really a bug:

He laments: 

> consider I have define my own ou, the array msp.ouIdentifiers are not empty （only contain the OrganizationalUnits I've defined and not contain node ou).So 'admin' ou will not be found.

However if he defines an OU (or several), then it means (by definition) that every certificate the MSP recognizes should be among these OUs. 

He defined the node OUs as "client", "admin", etc. which obviously contradicts the MSP definition of:

```
OrganizationalUnitIdentifiers:
  - Certificate: "cacerts/cacert1.pem"
    OrganizationalUnitIdentifier: "zc"
```
I think what we may want to do, is to change our tooling such that it won't be possible to create a channel creation transaction (or a genesis block) with such a contradiction. 

Opinions? @adecaro  @ale-linux

ale-linux:
Just to repeat things here as well, @adecaro has looked into the jira and left some questions there - let's figure those out first before we proceed with this CR

stephyee:
For now, closing this draft as user error.

PR #1383, opened 2020-06-09
Fix use of connection profile for lifecycle CLI install operations

wlahti:
#### Type of change

- Bug fix

#### Description

Install operations (install, queryinstalled, getinstalledpackage) do not use the --channelID flag and thus the connection profile code did not work. Add a --targetPeer flag to specify the peer in a connection profile to use for these channel-less operations. 

#### Related issues

[FAB-17513](https://jira.hyperledger.org/browse/FAB-17513)

### DIFF
--- a/internal/peer/lifecycle/chaincode/client_connections.go
+++ b/internal/peer/lifecycle/chaincode/client_connections.go
@@ -39,6 +39,7 @@ type ClientConnectionsInput struct {
 	PeerAddresses         []string
 	TLSRootCertFiles      []string
 	ConnectionProfilePath string
+	TargetPeer            string
 	TLSEnabled            bool
 }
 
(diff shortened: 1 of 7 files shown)


internal/peer/lifecycle/chaincode/client_connections.go
```diff
@@ -155,6 +165,17 @@ func (c *ClientConnections) validatePeerConnectionParameters(input *ClientConnec
 	return nil
 }
 
+func (c *ClientConnectionsInput) appendPeerConfig(n *common.NetworkConfig, peer string) error {
+	peerConfig, ok := n.Peers[peer]
+	if !ok {
+		return errors.Errorf("peer '%s' doesn't have associated peer config", peer)
```
lindluni, line 180: `"peer '%s' does not have an associated peer config"`

internal/peer/lifecycle/chaincode/client_connections.go
```diff
@@ -118,17 +119,26 @@ func (c *ClientConnections) validatePeerConnectionParameters(input *ClientConnec
...
+		if input.ChannelID == "" {
+			if input.TargetPeer == "" {
+				return errors.New("--targetPeer must be specified for channel-less operation using connection profile")
+			}
+			err := input.appendPeerConfig(networkConfig, input.TargetPeer)
+			if err != nil {
+				return err
+			}
+		}
+
```
lindluni: It seems to me this should return, `return input.appendPeerConfig(networkConfig, input.TargetPeer)`, since you've validated there is no ChannelID, the check below is never going to be `>0`

wlahti: Right, but it does need to go through the validation below that block. I should probably refactor all of this code based on the numerous things I've learned over the years about go programming but I'm not sure the effort is truly worth it for this CLI code.

lindluni: Totally missed expanding the code, and all of the validation logic below, sorry Will

wlahti: I'm refactoring a little though anyways. :)

denyeart (dismissed):
LGTM, but I'll defer to @btl5037 and @suryalnvs to see if this meets their needs.

lindluni:
I'll try to test this out this weekend, sorry for the delay Will

mastersingh24 (commented):
LGTM but let's see that it meets the need(s) for which it was implemented.  Looks like it should

lindluni:
I'm testing this right now Will, sorry for the delay

lindluni:
Works perfectly, thanks Will!

lindluni (approved):

caod123:
@btl5037 @wlahti does this need to be backported? Original JIRA was targeted for 2.0 so might be worth backporting to 2.0-2.2?

PR #1384, opened 2020-06-09
Bump Go and Alpine Version

lindluni:
Bump Go to 1.14.4. This is also the first Go release to switch to alpine 3.12 with 1.14.3 being the last to support 3.11. Performing this update now will allow us to cross the alpine minor version barrier prior to our LTS release of 2.2 to give us time to test the stability of the release.


sykesm (approved):
Nit, but I think we should be updating the version in the vagrant box as well. While it's rarely used, it is a working, viable development environment.

PR #1390, opened 2020-06-11
add iter.Seek() to leveldb wrapper

cendhu:
#### Type of change

- New feature

#### Description

This PR adds `iter.Seek()`method  to the leveldb wrapper. The `Seek()` is already supported by leveldb iterator. However, to use this method correctly, this PR introduces a wrapper that adds `dbName` as a prefix to the key. 

#### Additional details

1. `Seek()` will be used in the code that exports data out of stateleveldb by skipping a few namespaces. 
2. This PR also adds test for `First()`, `Last()`, and `Prev()` methods on iterator.

### DIFF
--- a/common/ledger/util/leveldbhelper/leveldb_provider_test.go
+++ b/common/ledger/util/leveldbhelper/leveldb_provider_test.go
@@ -15,6 +15,7 @@ import (
 
 	"github.com/hyperledger/fabric/common/flogging"
 	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
 )
 
 func TestMain(m *testing.M) {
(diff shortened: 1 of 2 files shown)


common/ledger/util/leveldbhelper/leveldb_provider_test.go
```diff
@@ -51,6 +52,53 @@ func TestIterator(t *testing.T) {
...
+	defer itr4.Release()
+	require.True(t, itr4.Seek([]byte(createTestKey(10))))
+	require.Equal(t, []byte(createTestKey(10)), itr4.Key())
+	checkItrResults(t, itr4, createTestKeys(11, 19), createTestValues("db2", 11, 19))
+
+	itr5 := db2.GetIterator([]byte(createTestKey(11)), nil)
+	defer itr5.Release()
+	require.True(t, itr5.Seek([]byte(createTestKey(5))))
+	require.Equal(t, []byte(createTestKey(11)), itr5.Key())
+	checkItrResults(t, itr5, createTestKeys(12, 19), createTestValues("db2", 12, 19))
```
manish-sethi: Since all these tests are independent, it's better to run as a named sub-tests. For instance, this sub-test can be named as "seek-before-range"

manish-sethi (commented):
A minor comment in tests arrangement

common/ledger/util/leveldbhelper/leveldb_provider_test.go
cendhu: Overall, the `TestIterator` took <= 100 msecs. Given that we already run tests in parallel in CI build, I am not sure whether having parallel test here will help much. IMO, it just complicates the test as the seek needs to check for return `true` or `false`. If it is true, it needs to check the value at the current location. Further, I do use `First()`, `Prev()`, and `Last()` too. I am not convinced about this added complexity to the test.

cendhu: I think, I misread independent as a suggestion to use `t.Parallel` with the subtest. I think you meant to use only `t.Run` without `t.Parallel` when you mentioned independent.

manish-sethi (approved):

PR #1391, opened 2020-06-11
Integration test for channel participation API list channels

wlahti:
#### Type of change

- Test update

#### Related issues

[FAB-17982](https://jira.hyperledger.org/browse/FAB-17982)

### DIFF
--- a/integration/e2e/e2e_test.go
+++ b/integration/e2e/e2e_test.go
@@ -36,6 +36,8 @@ import (
 	. "github.com/onsi/gomega"
 	"github.com/onsi/gomega/gbytes"
 	"github.com/onsi/gomega/gexec"
+	. "github.com/onsi/gomega/gstruct"
+	"github.com/onsi/gomega/types"
 	"github.com/tedsuo/ifrit"
 	"github.com/tedsuo/ifrit/ginkgomon"
 )
@@ -95,6 +97,7 @@ var _ = Describe("EndToEnd", func() {
 			network = nwo.New(nwo.BasicSolo(), testDir, nil, StartPort(), components)
 			network.MetricsProvider = "statsd"
 			network.StatsdEndpoint = datagramReader.Address()
+			network.ChannelParticipationEnabled = true
 			network.Profiles = append(network.Profiles, &nwo.Profile{
 				Name:          "TwoOrgsBaseProfileChannel",
 				Consortium:    "SampleConsortium",
@@ -157,6 +160,7 @@ var _ = Describe("EndToEnd", func() {
 
 			By("setting up the channel")
 			network.CreateAndJoinChannel(orderer, "testchannel")
+			channelParticipationList(network, orderer, []string{"testchannel"}, "systemchannel")
 			nwo.EnableCapabilities(network, "testchannel", "Application", "V2_0", orderer, network.Peer("Org1", "peer0"), network.Peer("Org2", "peer0"))
 
 			By("attempting to install unsupported chaincode without docker")
@@ -213,6 +217,7 @@ var _ = Describe("EndToEnd", func() {
 		BeforeEach(func() {
 			network = nwo.New(nwo.BasicKafka(), testDir, client, StartPort(), components)
 			network.MetricsProvider = "prometheus"
+			network.ChannelParticipationEnabled = true
 			network.GenerateConfigTree()
 			network.Bootstrap()
 
@@ -256,6 +261,7 @@ var _ = Describe("EndToEnd", func() {
 			orderer := network.Orderer("orderer")
 
 			network.CreateAndJoinChannel(orderer, "testchannel")
+			channelParticipationList(network, orderer, []string{"testchannel"}, "systemchannel")
 			nwo.EnableCapabilities(network, "testchannel", "Application", "V2_0", orderer, network.Peer("Org1", "peer0"), network.Peer("Org2", "peer0"))
 
 			// package, install, and approve by org1 - module chaincode
@@ -342,6 +348,7 @@ var _ = Describe("EndToEnd", func() {
 
 		BeforeEach(func() {
 			network = nwo.New(nwo.MultiChannelEtcdRaft(), testDir, client, StartPort(), components)
+			network.ChannelParticipationEnabled = true
 			network.GenerateConfigTree()
 			for _, peer := range network.Peers {
 				core := network.ReadPeerConfig(peer)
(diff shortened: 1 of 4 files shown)


integration/e2e/e2e_test.go
```diff
@@ -914,3 +924,42 @@ func hashFile(file string) string {
...
+func channelParticipationList(n *nwo.Network, o *nwo.Orderer, expectedChannels []string, systemChannel ...string) {
+	authClient, unauthClient := OrdererOperationalClients(n, o)
+	listChannelsURL := fmt.Sprintf("https://127.0.0.1:%d/participation/v1/channels", n.OrdererPort(o, nwo.OperationsPort))
+
+	expectedChannelList := &types.ChannelList{}
+
+	channelToChannelInfoShort := func(channel string) types.ChannelInfoShort {
+		return types.ChannelInfoShort{
+			Name: channel,
+			URL:  fmt.Sprintf("/participation/v1/channels/%s", channel),
```
tock-ibm: Use the constant from:
https://github.com/hyperledger/fabric/blob/946016f7f415459447d1e77cf2f3f65eb9d4b78a/orderer/common/channelparticipation/restapi.go#L31

integration/e2e/e2e_test.go
```diff
@@ -914,3 +924,42 @@ func hashFile(file string) string {
 func chaincodeContainerNameFilter(n *nwo.Network, chaincode nwo.Chaincode) string {
 	return fmt.Sprintf("^/%s-.*-%s-%s$", n.NetworkID, chaincode.Label, hashFile(chaincode.PackageFile))
 }
+
+func channelParticipationList(n *nwo.Network, o *nwo.Orderer, expectedChannels []string, systemChannel ...string) {
```
tock-ibm, line 939: As we are going to use that a lot, how about creating a public utility method that returns the list, and another to match against expected values?

integration/e2e/e2e_test.go
wlahti: I don't think we should use the constant here. It will help us ensure we don't unintentionally make any breaking changes to the API.

integration/e2e/e2e_test.go
wlahti: I'd prefer to leave it private for now but I can change it if you insist. We can make it public as soon as we use it in another test.

integration/e2e/e2e_test.go
tock-ibm: OK, sounds good.

integration/e2e/e2e_test.go
tock-ibm: I think that the bulk of tests will not be located in the e2e package but in the raft package, or possibly in a package of their own. This is because most of the interesting functionality of of join and delete is only applicable to raft. But I have no problem in exposing that when we actually have a test that needs that.

integration/e2e/e2e_test.go
```diff
@@ -914,3 +925,98 @@ func hashFile(file string) string {
...
+type channelInfo struct {
+	Name            string `json:"name"`
+	URL             string `json:"url"`
+	Status          string `json:"status"`
+	ClusterRelation string `json:"clusterRelation"`
+}
+
+func channelParticipationListOne(n *nwo.Network, o *nwo.Orderer, channel channelInfoShort) {
+	authClient, _ := OrdererOperationalClients(n, o)
+	listChannelURL := fmt.Sprintf("https://127.0.0.1:%d/%s", n.OrdererPort(o, nwo.OperationsPort), channel.URL)
```
tock-ibm: I think the expected ChannelInfo should be passed in as an argument. This way we'll be able to reuse this method... It will not always be "none"+"active" and "member"+"active", there are more options. 

There should a `config-tracker` here somewhere, maybe in the raft package.... an orderer that is not a member of a certain channel.

integration/nwo/fabricconfig/orderer.go
```diff
@@ -121,3 +122,8 @@ type OrdererStatsd struct {
 	WriteInterval time.Duration `yaml:"WriteInterval,omitempty"`
 	Prefix        string        `yaml:"Prefix,omitempty"`
 }
+
+type ChannelParticipation struct {
+	Enabled       bool `yaml:"Enabled"`
+	RemoveStorage bool `yaml:"RemoveStorage"`
+}
```
tock-ibm: `RemoveStorage` is probably redundant right now, there is a (significant) change we won't support archiving, so probably better to leave it out for now.

integration/nwo/orderer_template.go
```diff
@@ -112,4 +112,7 @@ Metrics:
     WriteInterval: 5s
     Prefix: {{ ReplaceAll (ToLower Orderer.ID) "." "_" }}
 {{- end }}
+ChannelParticipation:
+  Enabled: {{ .ChannelParticipationEnabled }}
+  RemoveStorage: false
```
tock-ibm: Same here, probably better to leave it out for now.

integration/e2e/e2e_test.go
```diff
@@ -914,3 +925,98 @@ func hashFile(file string) string {
...
+}
+
+// for the current test scenarios, we only have these two cluster relations.
+// other possible values are "follower" and "config-tracker"
+func clusterRelation(consensusType string) string {
+	if consensusType == "etcdraft" {
+		return "member"
+	}
+	return "none"
+}
```
tock-ibm: This is correct for these particular tests, but there is no re-use possible, so as soon as we hit "config-tracker" and "follower".... probably best not to embed this in the `*ListOne` method...

integration/e2e/e2e_test.go
```diff
@@ -914,3 +925,98 @@ func hashFile(file string) string {
...
+		"URL":  Equal(fmt.Sprintf("/participation/v1/channels/%s", channel)),
+	})
+}
+
+type channelInfo struct {
+	Name            string `json:"name"`
+	URL             string `json:"url"`
+	Status          string `json:"status"`
+	ClusterRelation string `json:"clusterRelation"`
+}
```
tock-ibm: Where is the Height? 
And why repeat the definition of orderer/types/ChannelInfo?

wlahti: Mirroring the types here provides us an automated way to notice any breaking changes that will affect consumers. For this, I only included the fields I could test the exact values. We could include Height but it would require some additional logic to determine the expected value (or we could just check that it's set to a non-empty value).

caod123 (approved):
Going to just merge this as is for now to allow for updates as we add more integration tests to address the reusability of some of the functions/location of the tests.

PR #1392, opened 2020-06-11
Removing viper lib from viperutil/config_util.go

nbasker:
#### Type of change

- Improvement (improvement to code and performance by avoiding viper library)

#### Description

There is overlap between in the implementation of
`getKeysRecursively` and the viper.Get() implementation.
The viper does yaml.Unmarshal and stores in viper.config,
which is provided to getKeysRecursively and further sub-keys
are available readily and there is no need to fetch
through viper.Get(). Towards the overall objective of
moving away from viper, its usage here can be removed.

Updated the PR with changes to remove "Viper" library from following three modules. These are the modules that were using Viper along with EnhancedExactUnmarshal()/getKeysRecursively() that had overlapping functionality.
1) orderer/common/localconfig
2) internal/configtxgen/genesisconfig
3) common/viperutil

In the common/viperutil/config_util.go introduced a ConfigParser that tracks the config-files, paths, environment variables and unmarshalled-config. It may be better to create a new directory file called common/configparser/configparser.go. Based on review comments of this, shall take up this change.

#### Additional details

Updated unit test of the three modules

go test -race -count 1 -v -run Test github.com/hyperledger/fabric/orderer/common/localconfig | tail -2
PASS
ok      github.com/hyperledger/fabric/orderer/common/localconfig        5.437s
go test -race -count 1 -v -run Test github.com/hyperledger/fabric/internal/configtxgen/genesisconfig | tail -2
PASS
ok      github.com/hyperledger/fabric/internal/configtxgen/genesisconfig        5.884s
go test -race -count 1 -v -run Test github.com/hyperledger/fabric/common/viperutil | tail -2
PASS
ok      github.com/hyperledger/fabric/common/viperutil  5.507s
go test -race -count 1 -v -run Test github.com/hyperledger/fabric/orderer/consensus/etcdraft | tail -2
PASS
ok      github.com/hyperledger/fabric/orderer/consensus/etcdraft        59.218s
go test -race -count 1 -v -run Test github.com/hyperledger/fabric/orderer/consensus/etcdraft | grep SUCCESS!
SUCCESS! -- 105 Passed | 0 Failed | 0 Pending | 0 Skipped

#### Related issues

https://jira.hyperledger.org/browse/FAB-17508


nbasker:
Manually ran the same test case a few times and it passed. Shall resubmit with further cleanup after getting review feedback.

go test -race -count 1 -v -run Test github.com/hyperledger/fabric/orderer/consensus/etcdraft 
<--snip-->                                                                                                                                                      
••2020-06-12 09:10:02.116 IST [test] HandleChain -> INFO 02b EvictionSuspicion not set, defaulting to 10m0s
•                                                                                
Ran 105 of 105 Specs in 44.319 seconds                                           
SUCCESS! -- 105 Passed | 0 Failed | 0 Pending | 0 Skipped                
--- PASS: TestEtcdraft (44.32s)

denyeart:
Tagging @sykesm for review since the PR is related to a Jira he opened.

sykesm (commented):
Thanks for the PR. Sorry for the delay in reviewing. I've been focusing on other things.

Looks like the approach you've taken is to extract a bunch of code from viper and inlined it into the `viperutil` package. While normally I'd push back on that because we wouldn't benefit from the enhancements and fixes from that community, but, given we're stuck on an ancient version, it's probably good enough.

My goal is to do a detailed review over the next day or two. Based on my quick glance, I think there are a few things that were pulled over that we simply don't need and that should be removed (e.g. `SetConfigType`, `AutomaticEnv`), there are some idiomatic Go changes that I'd like to see, and there are tests that need to be backfilled.

common/viperutil/config_util.go
```diff
@@ -24,14 +28,230 @@ import (
...
+	if c.envKeyReplacer != nil {
+		envKey = c.envKeyReplacer.Replace(envKey)
+	}
+	return os.Getenv(envKey)
+}
+
+// Return the parsed config store
+func (c *ConfigParser) GetParsedConfig() map[string]interface{} {
+	return c.config
+}
```
sykesm, line 252: This function is only used by `EnhancedExactUnmarshal` and can probably be unexported.

common/viperutil/config_util.go
```diff
@@ -315,7 +543,7 @@ func bccspHook(f reflect.Type, t reflect.Type, data interface{}) (interface{}, e
 // EnhancedExactUnmarshal is intended to unmarshal a config file into a structure
 // producing error when extraneous variables are introduced and supporting
 // the time.Duration type
-func EnhancedExactUnmarshal(v *viper.Viper, output interface{}) error {
+func EnhancedExactUnmarshal(c *ConfigParser, output interface{}) error {
```
sykesm, line 546: This is no longer a function around viper so it should be converted into a method on `ConfigParser`.

orderer/common/localconfig/config.go
```diff
@@ -16,7 +16,7 @@ import (
 	"github.com/hyperledger/fabric/common/flogging"
 	"github.com/hyperledger/fabric/common/viperutil"
 	coreconfig "github.com/hyperledger/fabric/core/config"
-	"github.com/spf13/viper"
+	//"github.com/spf13/viper"
```
sykesm, line 19: please remove commented out code

sykesm (changes requested):
Given the approach is to stop using viper, we don't need to maintain the generalizations implied by viper.  That means a number of methods can be removed and/or unexported.

Also, since this is not tied to viper, the package should be renamed to something other than `viperutil`.  I'm not sure what the best name so it's probably best to address the other comments first.

Thanks.

nbasker:
> Thanks for the PR. Sorry for the delay in reviewing. I've been focusing on other things.
> 
> Looks like the approach you've taken is to extract a bunch of code from viper and inlined it into the `viperutil` package. While normally I'd push back on that because we wouldn't benefit from the enhancements and fixes from that community, but, given we're stuck on an ancient version, it's probably good enough.
> 
> My goal is to do a detailed review over the next day or two. Based on my quick glance, I think there are a few things that were pulled over that we simply don't need and that should be removed (e.g. `SetConfigType`, `AutomaticEnv`), there are some idiomatic Go changes that I'd like to see, and there are tests that need to be backfilled.

Shall be addressing the review comments and raising a PR in the next few days.

nbasker:
I still need to take care of a few more changes given, shall post once all the changes are taken care.

nbasker:
Addressed all the code review comments and updated the PR.
Regarding the tests that need to be backfilled, should it be around the ConfigParse methods.
Further, instead of viperutil, this can be moved into common/configparse/configparse.go and configparse_test.go to reflect that it is not viperutil anymore (or any other better name).
Since the dependency between viperutil.EnhancedExactUnmarshal() and viper is removed, updating the viper to latest can also be considered.

common/viperutil/config_util.go
```diff
@@ -24,14 +28,166 @@ import (
...
+	c.envPrefix = in
+}
+
+// Search for the existence of filename for all supported extensions
+func (c *ConfigParser) searchInPath(in string) (filename string) {
+	var supportedExts []string = []string{"yaml", "yml"}
+	logger.Debugf("Searching for config in ", in)
+	for _, ext := range supportedExts {
+		fullPath := filepath.Join(in, c.configName+"."+ext)
+		logger.Debugf("Checking for ", fullPath)
```
sykesm, line 163: Same here. I don't see the need but, if we have it, it should be the correct variant.

common/viperutil/config_util.go
```diff
@@ -24,14 +28,166 @@ import (
...
+		_, err := os.Stat(fullPath)
+		if err == nil {
+			logger.Debugf("Found: ", fullPath)
```
sykesm, line 166: Same here. I don't see the need but, if we have it, it should be the correct variant.

common/viperutil/config_util.go
```diff
@@ -24,14 +28,166 @@ import (
...
+// If "peer" is set here, all environment variable
+// searches shall be prefixed with "PEER_"
+func (c *ConfigParser) SetEnvPrefix(in string) {
+	c.envPrefix = in
+}
+
+// Search for the existence of filename for all supported extensions
+func (c *ConfigParser) searchInPath(in string) (filename string) {
+	var supportedExts []string = []string{"yaml", "yml"}
+	logger.Debugf("Searching for config in ", in)
```
sykesm, line 160: I don't believe we need the debug information but, if it's here, it should be the correct variant. Debugf uses a Printf style format string and value and that's not what's here.

Looks like I need to update the linter.

common/viperutil/config_util.go
```diff
@@ -24,14 +28,166 @@ import (
...
+			return fullPath
+		}
+	}
+	return ""
+}
+
+// Search for the configName in all configPaths
+func (c *ConfigParser) findConfigFile() string {
+
+	logger.Debugf("Searching for config in ", c.configPaths)
```
sykesm, line 176: I don't believe this debug is needed.

sykesm (commented):
It's looking a lot better. Let's get these comments taken care of and we can discuss other changes as follow-on PRs.

sykesm:
Looks like I forgot to request that you squash your commits. This change should be contained in a single commit. Thanks.

nbasker:
> It's looking a lot better. Let's get these comments taken care of and we can discuss other changes as follow-on PRs.

nbasker:
> Looks like I forgot to request that you squash your commits. This change should be contained in a single commit. Thanks.

Tried squashing my commits, but in between had done a merge from master to sync the bugfix branch, in the process messed up the squash. I am going to drop this PR and the branch and raise a fresh one. Shall include all the latest review comments including "file as the reader to ReadConfig" which got missed earlier.

PR #1393, opened 2020-06-11
[FAB-17963] Ch.Part.API: Validate join block

caod123:
#### Type of Change

* New Feature

#### Description

Introduce function for validating the join block received by the
registrar's JoinChannel. `validateJoinBlock` will determine whether the
block contains an application channel or system channel and also whether
the node is a member of the consenter set.

#### Related Issues

Task: [FAB-17963](https://jira.hyperledger.org/browse/FAB-17963)

### DIFF
--- a/orderer/common/channelparticipation/validator.go
+++ b/orderer/common/channelparticipation/validator.go
@@ -1,0 +1,54 @@
+/*
+Copyright IBM Corp. All Rights Reserved.
+
+SPDX-License-Identifier: Apache-2.0
+*/
+
+package channelparticipation
+
+import (
+	"errors"
+	"fmt"
+
+	cb "github.com/hyperledger/fabric-protos-go/common"
+	"github.com/hyperledger/fabric/bccsp/factory"
+	"github.com/hyperledger/fabric/common/channelconfig"
+	"github.com/hyperledger/fabric/protoutil"
+)
+
+// ValidateJoinBlock returns whether this block can be used as a join block for the channel participation API
+// and whether it is an system channel if it contains consortiums, or otherwise
+// and application channel if an application group exists.
+func ValidateJoinBlock(channelID string, configBlock *cb.Block) (isAppChannel bool, err error) {
+	if !protoutil.IsConfigBlock(configBlock) {
+		return false, errors.New("block is not a config block")
+	}
+
+	envelope, err := protoutil.ExtractEnvelope(configBlock, 0)
+	if err != nil {
+		return false, err
+	}
+
+	cryptoProvider := factory.GetDefault()
+	bundle, err := channelconfig.NewBundleFromEnvelope(envelope, cryptoProvider)
+	if err != nil {
+		return false, err
+	}
+
+	// Check channel id from join block matches
+	if bundle.ConfigtxValidator().ChannelID() != channelID {
+		return false, fmt.Errorf("config block channelID [%s] does not match passed channelID [%s]",
+			bundle.ConfigtxValidator().ChannelID(), channelID)
+	}
+
+	// Check channel type
+	_, isSystemChannel := bundle.ConsortiumsConfig()
+	if !isSystemChannel {
+		_, isAppChannel = bundle.ApplicationConfig()
+		if !isAppChannel {
+			return false, errors.New("invalid config: must have at least one of application or consortiums")
+		}
+	}
+
+	return isAppChannel, err
+}
--- a/orderer/common/multichannel/registrar.go
+++ b/orderer/common/multichannel/registrar.go
@@ -417,7 +417,7 @@ func (r *Registrar) ChannelInfo(channelID string) (types.ChannelInfo, error) {
 	return info, nil
 }
 
-func (r *Registrar) JoinChannel(channelID string, configBlock *cb.Block) (types.ChannelInfo, error) {
+func (r *Registrar) JoinChannel(channelID string, configBlock *cb.Block, isAppChannel bool) (types.ChannelInfo, error) {
 	r.lock.RLock()
 	defer r.lock.RUnlock()
 
--- a/orderer/common/multichannel/registrar_test.go
+++ b/orderer/common/multichannel/registrar_test.go
@@ -531,7 +531,7 @@ func TestRegistrar_JoinChannel(t *testing.T) {
 		registrar := NewRegistrar(localconfig.TopLevel{}, ledgerFactory, mockCrypto(), &disabled.Provider{}, cryptoProvider)
 		registrar.Initialize(mockConsenters)
 
-		info, err := registrar.JoinChannel("some-app-channel", &cb.Block{})
+		info, err := registrar.JoinChannel("some-app-channel", &cb.Block{}, false)
 		assert.EqualError(t, err, "system channel exists")
 		assert.Equal(t, types.ChannelInfo{}, info)
 	})
@@ -559,7 +559,7 @@ func TestRegistrar_JoinChannel(t *testing.T) {
 		registrar.CreateChain("my-channel")
 		assert.NotNil(t, registrar.GetChain("my-channel"))
 
-		info, err := registrar.JoinChannel("my-channel", &cb.Block{})
+		info, err := registrar.JoinChannel("my-channel", &cb.Block{}, false)
 		assert.EqualError(t, err, "channel already exists")
 		assert.Equal(t, types.ChannelInfo{}, info)
 	})
@@ -571,10 +571,12 @@ func TestRegistrar_JoinChannel(t *testing.T) {
 
 		ledgerFactory, _ := newLedgerAndFactory(tmpdir, "", nil)
 		mockConsenters := map[string]consensus.Consenter{"not-raft": &mockConsenter{}}
+
 		config := localconfig.TopLevel{}
 		config.General.BootstrapMethod = "none"
 		config.General.GenesisFile = ""
 		registrar := NewRegistrar(config, ledgerFactory, mockCrypto(), &disabled.Provider{}, cryptoProvider)
+
 		assert.Panics(t, func() { registrar.Initialize(mockConsenters) })
 	})
 }
--- a/orderer/common/channelparticipation/mocks/channel_management.go
+++ b/orderer/common/channelparticipation/mocks/channel_management.go
@@ -33,11 +33,12 @@ type ChannelManagement struct {
 	channelListReturnsOnCall map[int]struct {
 		result1 types.ChannelList
 	}
-	JoinChannelStub        func(string, *common.Block) (types.ChannelInfo, error)
+	JoinChannelStub        func(string, *common.Block, bool) (types.ChannelInfo, error)
 	joinChannelMutex       sync.RWMutex
 	joinChannelArgsForCall []struct {
 		arg1 string
 		arg2 *common.Block
+		arg3 bool
 	}
 	joinChannelReturns struct {
 		result1 types.ChannelInfo
@@ -178,17 +179,18 @@ func (fake *ChannelManagement) ChannelListReturnsOnCall(i int, result1 types.Cha
 	}{result1}
 }
 
-func (fake *ChannelManagement) JoinChannel(arg1 string, arg2 *common.Block) (types.ChannelInfo, error) {
+func (fake *ChannelManagement) JoinChannel(arg1 string, arg2 *common.Block, arg3 bool) (types.ChannelInfo, error) {
 	fake.joinChannelMutex.Lock()
 	ret, specificReturn := fake.joinChannelReturnsOnCall[len(fake.joinChannelArgsForCall)]
 	fake.joinChannelArgsForCall = append(fake.joinChannelArgsForCall, struct {
 		arg1 string
 		arg2 *common.Block
-	}{arg1, arg2})
-	fake.recordInvocation("JoinChannel", []interface{}{arg1, arg2})
+		arg3 bool
+	}{arg1, arg2, arg3})
+	fake.recordInvocation("JoinChannel", []interface{}{arg1, arg2, arg3})
 	fake.joinChannelMutex.Unlock()
 	if fake.JoinChannelStub != nil {
-		return fake.JoinChannelStub(arg1, arg2)
+		return fake.JoinChannelStub(arg1, arg2, arg3)
 	}
 	if specificReturn {
 		return ret.result1, ret.result2
@@ -203,17 +205,17 @@ func (fake *ChannelManagement) JoinChannelCallCount() int {
 	return len(fake.joinChannelArgsForCall)
 }
 
-func (fake *ChannelManagement) JoinChannelCalls(stub func(string, *common.Block) (types.ChannelInfo, error)) {
+func (fake *ChannelManagement) JoinChannelCalls(stub func(string, *common.Block, bool) (types.ChannelInfo, error)) {
 	fake.joinChannelMutex.Lock()
 	defer fake.joinChannelMutex.Unlock()
 	fake.JoinChannelStub = stub
 }
 
-func (fake *ChannelManagement) JoinChannelArgsForCall(i int) (string, *common.Block) {
+func (fake *ChannelManagement) JoinChannelArgsForCall(i int) (string, *common.Block, bool) {
 	fake.joinChannelMutex.RLock()
 	defer fake.joinChannelMutex.RUnlock()
 	argsForCall := fake.joinChannelArgsForCall[i]
-	return argsForCall.arg1, argsForCall.arg2
+	return argsForCall.arg1, argsForCall.arg2, argsForCall.arg3
 }
 
 func (fake *ChannelManagement) JoinChannelReturns(result1 types.ChannelInfo, result2 error) {
--- a/orderer/common/channelparticipation/restapi.go
+++ b/orderer/common/channelparticipation/restapi.go
@@ -49,7 +49,7 @@ type ChannelManagement interface {
 	ChannelInfo(channelID string) (types.ChannelInfo, error)
 
 	// JoinChannel instructs the orderer to create a channel and join it with the provided config block.
-	JoinChannel(channelID string, configBlock *cb.Block) (types.ChannelInfo, error)
+	JoinChannel(channelID string, configBlock *cb.Block, isAppChannel bool) (types.ChannelInfo, error)
 
 	// RemoveChannel instructs the orderer to remove a channel.
 	// Depending on the removeStorage parameter, the storage resources are either removed or archived.
@@ -163,7 +163,13 @@ func (h *HTTPHandler) serveJoin(resp http.ResponseWriter, req *http.Request) {
 		return
 	}
 
-	info, err := h.registrar.JoinChannel(channelID, block)
+	isAppChannel, err := ValidateJoinBlock(channelID, block)
+	if err != nil {
+		h.sendResponseJsonError(resp, http.StatusBadRequest, errors.Wrap(err, "invalid join block"))
+		return
+	}
+
+	info, err := h.registrar.JoinChannel(channelID, block, isAppChannel)
 	if err == nil {
 		info.URL = path.Join(URLBaseV1Channels, info.Name)
 		h.logger.Debugf("Successfully joined channel: %s", info)
--- a/orderer/common/channelparticipation/restapi_test.go
+++ b/orderer/common/channelparticipation/restapi_test.go
@@ -17,10 +17,12 @@ import (
 	"path"
 	"testing"
 
+	"github.com/hyperledger/fabric-protos-go/common"
 	"github.com/hyperledger/fabric/orderer/common/channelparticipation"
 	"github.com/hyperledger/fabric/orderer/common/channelparticipation/mocks"
 	"github.com/hyperledger/fabric/orderer/common/localconfig"
 	"github.com/hyperledger/fabric/orderer/common/types"
+	"github.com/hyperledger/fabric/protoutil"
 	"github.com/pkg/errors"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
@@ -245,7 +247,7 @@ func TestHTTPHandler_ServeHTTP_Join(t *testing.T) {
 		fakeManager.JoinChannelReturns(info, nil)
 
 		resp := httptest.NewRecorder()
-		req := genJoinRequestFormData(t, []byte{})
+		req := genJoinRequestFormData(t, validBlockBytes("ch-id"))
 		h.ServeHTTP(resp, req)
 		assert.Equal(t, http.StatusCreated, resp.Result().StatusCode)
 		assert.Equal(t, "application/json", resp.Result().Header.Get("Content-Type"))
@@ -260,7 +262,7 @@ func TestHTTPHandler_ServeHTTP_Join(t *testing.T) {
 		fakeManager, h := setup(config, t)
 		fakeManager.JoinChannelReturns(types.ChannelInfo{}, types.ErrSystemChannelExists)
 		resp := httptest.NewRecorder()
-		req := genJoinRequestFormData(t, []byte{})
+		req := genJoinRequestFormData(t, validBlockBytes("ch-id"))
 		h.ServeHTTP(resp, req)
 		checkErrorResponse(t, http.StatusMethodNotAllowed, "cannot join: system channel exists", resp)
 		assert.Equal(t, "GET", resp.Result().Header.Get("Allow"))
@@ -270,7 +272,7 @@ func TestHTTPHandler_ServeHTTP_Join(t *testing.T) {
 		fakeManager, h := setup(config, t)
 		fakeManager.JoinChannelReturns(types.ChannelInfo{}, types.ErrChannelAlreadyExists)
 		resp := httptest.NewRecorder()
-		req := genJoinRequestFormData(t, []byte{})
+		req := genJoinRequestFormData(t, validBlockBytes("ch-id"))
 		h.ServeHTTP(resp, req)
 		checkErrorResponse(t, http.StatusMethodNotAllowed, "cannot join: channel already exists", resp)
 		assert.Equal(t, "GET, DELETE", resp.Result().Header.Get("Allow"))
@@ -280,7 +282,7 @@ func TestHTTPHandler_ServeHTTP_Join(t *testing.T) {
 		fakeManager, h := setup(config, t)
 		fakeManager.JoinChannelReturns(types.ChannelInfo{}, types.ErrAppChannelsAlreadyExists)
 		resp := httptest.NewRecorder()
-		req := genJoinRequestFormData(t, []byte{})
+		req := genJoinRequestFormData(t, validBlockBytes("ch-id"))
 		h.ServeHTTP(resp, req)
 		checkErrorResponse(t, http.StatusForbidden, "cannot join: application channels already exist", resp)
 	})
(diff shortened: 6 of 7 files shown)


orderer/common/multichannel/registrar.go
```diff
@@ -446,3 +453,65 @@ func (r *Registrar) RemoveChannel(channelID string, removeStorage bool) error {
...
+
+	etcdRaftMetadata := &eb.ConfigMetadata{}
+	err = proto.Unmarshal(ordererConfig.ConsensusMetadata(), etcdRaftMetadata)
+	if err != nil {
+		return false, false, errors.Wrap(err, "failed unmarshaling etcdraft metadata")
+	}
+
+	isMember = false
+	for _, c := range etcdRaftMetadata.Consenters {
+		if r.config.General.ListenAddress == c.Host &&
```
yacovm: It's better to compare the certificates. The reason is that you might have a matching host and port, but if all certificates in the consenter set are different than yours, no consenter will talk to you due to TLS pinning.

orderer/common/multichannel/registrar_test.go
```diff
@@ -510,13 +513,13 @@ func TestBroadcastChannelSupport(t *testing.T) {
 }
 
 func TestRegistrar_JoinChannel(t *testing.T) {
```
tock-ibm: Add a test case that tests that the ChannelId in the channel header matches the method parameter.
Add a test case that exercises the certificate matching, and the isMemeber return value.
Add a test case that exercises bad etcdraft metadata detection.

orderer/common/multichannel/registrar.go
```diff
@@ -446,3 +453,64 @@ func (r *Registrar) RemoveChannel(channelID string, removeStorage bool) error {
...
+		return false, false, errors.Wrap(err, "failed unmarshaling etcdraft metadata")
+	}
+
+	isMember = false
+	for _, c := range etcdRaftMetadata.Consenters {
+		if r.config.General.TLS.Certificate == string(c.ClientTlsCert) {
+			isMember = true
+			break
+		}
+	}
```
tock-ibm: Not sure this is correct. See how they do it here:
etcdraft/consenter.go detectSelfID()
https://github.com/hyperledger/fabric/blob/946016f7f415459447d1e77cf2f3f65eb9d4b78a/orderer/consensus/etcdraft/consenter.go#L106

orderer/common/multichannel/registrar.go
```diff
@@ -446,3 +453,64 @@ func (r *Registrar) RemoveChannel(channelID string, removeStorage bool) error {
...
+	consensusType := ordererConfig.ConsensusType()
+	if consensusType != "etcdraft" {
+		return false, false, fmt.Errorf("join block consensus type is not configured to etcdraft, found type %s", consensusType)
+	}
+
+	etcdRaftMetadata := &eb.ConfigMetadata{}
+	err = proto.Unmarshal(ordererConfig.ConsensusMetadata(), etcdRaftMetadata)
+	if err != nil {
+		return false, false, errors.Wrap(err, "failed unmarshaling etcdraft metadata")
+	}
```
tock-ibm: Need to verify that the etcdRaftMetadata won't cause HandleChain to err, so better to use this as well:
etcdraft/utils.go CheckConfigMetadata() etc,
https://github.com/hyperledger/fabric/blob/946016f7f415459447d1e77cf2f3f65eb9d4b78a/orderer/consensus/etcdraft/util.go#L200

tock-ibm (changes requested):

tock-ibm:
How about adding a method to 

https://github.com/hyperledger/fabric/blob/946016f7f415459447d1e77cf2f3f65eb9d4b78a/orderer/consensus/consensus.go#L19

Say, `ValidateJoinBlock(block) err`?

This way the etcdraft consenter will have access to all the good stuff in etcdraft/util.go, and we won't have to maintain etcdraft specific logic in two places? In addition, solo and kafka consenters simple return an error "not supported" or something like that. 

So in the JoinChannel(), after doing all that we can that is non-specific to a consensus-type, we'll retrieve the right consenter, and invoke this method?

orderer/common/multichannel/registrar_test.go
caod123: Originally I had just added the tests for the registrars `JoinChannel` since that's the exported function that consumers would use instead of the `validateJoinBlock` helper function. Because this PR doesnt actually process the returned values for `isMember` and `isAppChannel` in `JoinChannel` I only had the error cases tested. I can change it to just unit test the helper function though until we actually implement the rest of the `JoinChannel` behavior.

orderer/common/multichannel/registrar.go
caod123: Ah thanks, I was looking for something like this and wasn't sure if what I had was correct.

caod123:
> How about adding a method to
> 
> https://github.com/hyperledger/fabric/blob/946016f7f415459447d1e77cf2f3f65eb9d4b78a/orderer/consensus/consensus.go#L19
> 
> Say, `ValidateJoinBlock(block) err`?
> 
> This way the etcdraft consenter will have access to all the good stuff in etcdraft/util.go, and we won't have to maintain etcdraft specific logic in two places? In addition, solo and kafka consenters simple return an error "not supported" or something like that.
> 
> So in the JoinChannel(), after doing all that we can that is non-specific to a consensus-type, we'll retrieve the right consenter, and invoke this method?

Sounds good, I was hoping to see if we could simplify as much of the validation logic as possible to avoid having to dance around the import cycle, but given the additional validations that are required it probably makes sense to utilize the util functions as you've suggested.

orderer/common/multichannel/registrar.go
```diff
@@ -446,3 +453,64 @@ func (r *Registrar) RemoveChannel(channelID string, removeStorage bool) error {
 	//TODO
 	return errors.New("Not implemented yet")
 }
+
+// validateJoinBlock returns whether this block can be used as a join block for the channel participation API
+// as well as whether the orderer is a member of the consenter set.
+// A join block is a block of either a system channel containing a ConsoritumsConfig, or an application channel
```
wlahti:
```suggestion
// A join block is a block of either a system channel containing a ConsortiumsConfig, or an application channel
```
Super nit: typo

orderer/consensus/etcdraft/util.go
```diff
@@ -196,6 +196,43 @@ func ConsensusMetadataFromConfigBlock(block *common.Block) (*etcdraft.ConfigMeta
...
+	if err != nil {
+		return nil, errors.Wrap(err, "failed to extract payload from envelope")
+	}
+
+	configEnvelope, err := configtx.UnmarshalConfigEnvelope(payload.Data)
+	if err != nil {
+		return nil, errors.Wrap(err, "failed to extract config envelope from payload")
+	}
+
+	if configEnvelope.Config != nil && configEnvelope.Config.ChannelGroup != nil {
```
wlahti: It's a style preference, but it looks weird to me now to see things nested like this (and all the fall-out errors after the expected path). Any reason to not invert the logic to return out early for the errors to keep it clear at a quick glance which error goes with which path of the code? 

    if configEnvelope.Config == nil && configEnvelope.Config.ChannelGroup == nil {
		return nil, errors.New("config block missing channel group")
	}

	ordererConfigGroup, ok := configEnvelope.Config.ChannelGroup.Groups["Orderer"]
	if !ok {
		return nil, errors.New("config block missing orderer group")
	}

	val, ok := ordererConfigGroup.Values["ConsensusType"]
	if !ok {
		return nil, errors.New("config block missing consensus type value")
	}

	return MetadataFromConfigValue(val)

wlahti:
One thing that isn't clear to me (but it's probably just my lack of familiarity with the orderer) is the true role of the registrar's validateJoinBlock() v. the consenter's ValidateJoinBlock(). I assume one of them handles this part of the JIRA: "Make sure the join block does not crash the orderer" but I can't really tell which. It might be good to clarify the comments to help ignorant people like myself get up to speed on some of these parts of the orderer. :)

caod123:
@wlahti the etcdraft consenter's `ValidateJoinBlock` handles all of the validation related to etcdraft metadata and determining if the node is a member of the consenter set. The higher level registrar's `validateJoinBlock` handles the remaining validation around the block itself (checking for groups, etc) and later calls the consenter's `ValidateJoinBlock` to handle the etcdraft metadata portion. Agree the comments could be improved though :)

orderer/consensus/etcdraft/util.go
caod123: Good catch, I had originally written this logic while I was still figuring out which portions to check for which ended up resulting in the nested checks. I meant to change this afterwards but it slipped through after I had forgotten about it. Thanks!

orderer/common/multichannel/registrar.go
```diff
@@ -446,3 +453,64 @@ func (r *Registrar) RemoveChannel(channelID string, removeStorage bool) error {
...
+// containing an ApplicationConfig.
+func (r *Registrar) validateJoinBlock(channelID string, configBlock *cb.Block) (isAppChannel bool, isMember bool, err error) {
+	if configBlock == nil {
+		return false, false, errors.New("nil block")
+	}
+
```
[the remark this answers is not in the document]
caod123: Done with the ConfigtxValidator on the bundle since we extract the channelID from the header to there already.

orderer/common/multichannel/registrar.go
```diff
@@ -446,3 +451,67 @@ func (r *Registrar) RemoveChannel(channelID string, removeStorage bool) error {
...
+	// Check channel type
+	_, isSystemChannel := bundle.ConsortiumsConfig()
+	_, isAppChannel = bundle.ApplicationConfig()
+
+	// A valid system channel has consortiums but no application group
+	// A valid application channel has an application group but no consortiums
+	if isSystemChannel == isAppChannel {
+		return false, false, errors.New("invalid channel type: must contain only one of consortiums or application group")
+	}
```
tock-ibm: I think this part is stricter than what fabric actually enforces. To the best of my knowledge, there is no piece of code where a system channel config is strictly not allowed to have an Application group. It will ignore it completely, but will not reject it or crash because of it. So I think something like:
```go
_, isSystemChannel := bundle.ConsortiumsConfig()
if !isSystemChannel {
   _, isAppChannel = bundle.ApplicationConfig()
  if !isAppChannel {
    return false, false, error
}
}
```
is more appropriate.

orderer/common/multichannel/registrar_test.go
```diff
@@ -563,4 +566,144 @@ func TestRegistrar_JoinChannel(t *testing.T) {
...
+		assert.NoError(t, err)
+		ledger.Append(genesisBlockAppSolo)
+
+		info, err := registrar.JoinChannel("wrong-channel-id", genesisBlockSysSolo)
+
+		assert.EqualError(t, err, "config block channelID sys-channel does not match passed channelID wrong-channel-id")
+		assert.Equal(t, types.ChannelInfo{}, info)
+	})
+
+	t.Run("Reject join when join block has both application group and consortiums", func(t *testing.T) {
```
tock-ibm: In light of the previous comment, this test may need to be tuned.

orderer/consensus/etcdraft/consenter.go
```diff
@@ -103,6 +104,49 @@ func (c *Consenter) ReceiverByChain(channelID string) MessageReceiver {
...
+// ValidateJoinBlock implements the JoinBlockValidator interface and validates the etcdraft metadata
+// in the join block passed to the channel participation API. It performs checks to ensure the orderer
+// won't crash when calling the consenter's HandleChain.
+// It also determines whether the node is a member of the consenter set defined in the join block.
+func (c *Consenter) ValidateJoinBlock(configBlock *common.Block) (isClusterMember bool, err error) {
+	// Fetch consenter metadata
+	metadata, err := protoutil.GetConsenterMetadataFromBlock(configBlock)
+	if err != nil {
+		return false, errors.Wrap(err, "failed to get consenter metadata from config block")
+	}
```
tock-ibm: The metadata can be nil (migration or genesis blocks) 

https://github.com/hyperledger/fabric/blob/5fd497e7bb33987b1f029f0a2d51c9aa9f277e10/orderer/consensus/etcdraft/consenter.go#L141

can you add a test to check that it works ok?

orderer/consensus/etcdraft/util.go
```diff
@@ -196,6 +196,48 @@ func ConsensusMetadataFromConfigBlock(block *common.Block) (*etcdraft.ConfigMeta
 	return MetadataFromConfigUpdate(configUpdate)
 }
 
+// ConsensusMetadataFromConfigBlockConfig reads consensus metadata from the configuration block's Config
+func ConsensusMetadataFromConfigBlockConfig(block *common.Block) (*etcdraft.ConfigMetadata, error) {
+	if block == nil {
```
tock-ibm: This name is funny... `ConsensusMetadataFromConfigBlock` is only called in one place so I guess it won't be too much to rename it to `ConsensusMetadataFromConfigUpdateBlock` and use that name here?

Or maybe something like `ExtractConsensusTypeMetadata` ?

orderer/common/multichannel/registrar.go
caod123: Done

orderer/consensus/etcdraft/consenter.go
caod123: Done

orderer/consensus/etcdraft/util.go
caod123: Went with the first suggestion

orderer/common/multichannel/registrar_test.go
caod123: Done

orderer/common/multichannel/registrar.go
```diff
@@ -430,6 +430,11 @@ func (r *Registrar) JoinChannel(channelID string, configBlock *cb.Block) (types.
 		return types.ChannelInfo{}, types.ErrChannelAlreadyExists
 	}
 
+	_, _, err := r.validateJoinBlock(channelID, configBlock)
```
yacovm: we are coupling here validation and future implementation of joining a block.

Why is this here in the so called "back end" and not closer to the request processing?

orderer/common/multichannel/registrar.go
```diff
@@ -438,3 +443,66 @@ func (r *Registrar) RemoveChannel(channelID string, removeStorage bool) error {
 	//TODO
 	return errors.New("Not implemented yet")
 }
+
+// validateJoinBlock returns whether this block can be used as a join block for the channel participation API
+// as well as whether the orderer is a member of the consenter set.
+// A join block is a block of either a system channel containing a ConsortiumsConfig, or an application channel
+// containing an ApplicationConfig.
+func (r *Registrar) validateJoinBlock(channelID string, configBlock *cb.Block) (isAppChannel bool, isMember bool, err error) {
```
yacovm: I don't understand why you're implementing config block validation inside the registrar. 

Can't we decouple it into its own component?

orderer/common/multichannel/registrar.go
```diff
@@ -438,3 +443,66 @@ func (r *Registrar) RemoveChannel(channelID string, removeStorage bool) error {
...
+		return false, false, err
+	}
+
+	bundle, err := channelconfig.NewBundleFromEnvelope(envelope, r.bccsp)
+	if err != nil {
+		return false, false, err
+	}
+
+	// Check channel id from join block matches
+	if bundle.ConfigtxValidator().ChannelID() != channelID {
```
yacovm: instead of doing this check why can't we just always use the channel ID from within the config block? I mean - to remove the ability to specify a channel ID in the join API altogether and just derive it from the config block

orderer/common/multichannel/registrar.go
```diff
@@ -438,3 +443,66 @@ func (r *Registrar) RemoveChannel(channelID string, removeStorage bool) error {
...
+		return false, false, fmt.Errorf("join block consensus type is not configured to etcdraft, found type %s", consensusType)
+	}
+
+	// Retrieve etcdraft consenter so we can validate consensus metadata
+	consenter, ok := r.consenters["etcdraft"]
+	if !ok {
+		return false, false, errors.New("cannot find consenter of type etcdraft")
+	}
+
+	joinBlockValidator, ok := consenter.(consensus.JoinBlockValidator)
```
yacovm: Ideally we'll have an object which we can inject this thing into.

What we're doing here is always assuming we have etcdraft consenter, yet at the same time we are trying to be generic by using a JoinblockValidation.

Let's either be generic or not be generic, but this looks to me like something in between which is neither.

orderer/consensus/consensus.go
```diff
@@ -26,6 +26,16 @@ type Consenter interface {
 	HandleChain(support ConsenterSupport, metadata *cb.Metadata) (Chain, error)
 }
 
+// JoinBlockValidator performs validation on a join block for the channel participation API.
+type JoinBlockValidator interface {
```
yacovm: If we could push the processing of this validation closer to the client input processing then maybe we wouldn't have to extend this interface.

orderer/consensus/etcdraft/consenter.go
```diff
@@ -104,6 +106,53 @@ func (c *Consenter) ReceiverByChain(channelID string) MessageReceiver {
...
+
+	// Check if member of consenters in join block
+	blockMetadata, err := ReadBlockMetadata(metadata, etcdRaftMetadata)
+	if err != nil {
+		return false, errors.Wrapf(err, "failed to read Raft metadata")
+	}
+
+	consenters := CreateConsentersMap(blockMetadata, etcdRaftMetadata)
+
+	_, err = c.detectSelfID(consenters)
```
yacovm: I think instead it's better to just initialize a  ConsenterCertificate  instance and invoke its IsConsenterOfChannel . 

All you need to know is if you're in the channel or not, you don't need the ID. 

Furthermore, you don't need the consenter metadata and there is no real "validation" of it, you just check it's a valid protobuf which isn't much.

orderer/common/multichannel/registrar.go
tock-ibm: The "front-end" in the channelparticipation  package has no knowledge of the channels that exist. We therefore want to check that the operation is permitted before looking deeper into the block. The boundary we decided on is that the front-end just unmarshals the []byte into a *common.Block, and the "back-end" looks into the block.
The next PR will show how we use the result of this one. You can look at a draft here:
https://github.com/tock-ibm/fabric/tree/draft-join-member-no-onboarding

yacovm (changes requested):
See some comments about high level placement and structure

orderer/common/multichannel/registrar.go
tock-ibm: because then the channelparticipation api would have to look into the block...

yacovm:
> @wlahti the etcdraft consenter's `ValidateJoinBlock` handles all of the validation related to etcdraft metadata and determining if the node is a member of the consenter set. The higher level registrar's `validateJoinBlock` handles the remaining validation around the block itself (checking for groups, etc) and later calls the consenter's `ValidateJoinBlock` to handle the etcdraft metadata portion. Agree the comments could be improved though :)

Unless my memory betrays me, at genesis there is no consenter metadata. 
In fact, the method **ReadBlockMetadata** is faking it by looking if there is no consenter metadata and then creating one dynamically:
```
	m := &etcdraft.BlockMetadata{
		NextConsenterId: 1,
		ConsenterIds:    make([]uint64, len(configMetadata.Consenters)),
	}
	// need to read consenters from the configuration
	for i := range m.ConsenterIds {
		m.ConsenterIds[i] = m.NextConsenterId
		m.NextConsenterId++
	}
```

orderer/common/multichannel/registrar.go
yacovm: > We therefore want to check that the operation is permitted before looking deeper into the block.

I don't think processing time optimization is a justification to couple together block validation and channel mapping. 

You can not check if the channel already exists and then return that it exists in the registrar, but in my opinion block validation should be checked closer to the input processing. 

In general I am not in favor of doing validation in the backend because it makes testing more complex. If you have a layer that only does validation checks then you can isolate it nicely and test it easier.

orderer/common/multichannel/registrar.go
yacovm: Right, so I think it makes more sense to have some kind of abstract validation object which the participation API would delegate the validation to, and it would be done there

orderer/common/multichannel/registrar.go
tock-ibm: I don't necessarily think it is a time optimization, but rather a division of work, and the logical flow of processing.  We can do some of the validation that is consenter agnostic and non-state related in the front-end, for sure. For example, checking the consistency of the channelID in the block with the url, and such. 

However, other parts of validation - or shall I say processing - those that are related to the state and are specific to a consenter should be done later, in the Registrar. The Registrar actually delegates part of the validation to the consenter, to keep the consenter specific (raft, in this case) logic close together, and because the registrar cannot call etcdraft functions because of an import cycle.

BTW, the import cycle can be broken quite easily, but still, I think the style of this code is  to prefer that the multichannel package will not depend on a specific consensus package but use the indirection of the Consenter abstraction.

orderer/common/multichannel/registrar.go
tock-ibm: The link between the channel participation api to the rest of the server state is through the registrar only. As mentioned above, some validation can be performed earlier, but not all.

orderer/common/multichannel/registrar.go
tock-ibm: By generic do you mean to have the kafka and solo consenters implement this and return error("join unsupported") or something? we can do that.

orderer/common/multichannel/registrar.go
yacovm: Yes, I understand that the flow goes like this:
**channelParticipationAPI --> registrar --> consenter**

But if you observe, you don't really need the consenter at all. You're using a method of the consenter without a real need for it and you can replace it with **ConsenterCertificate** in the etcdraft package like I hinted in the CR. 

So, I am saying that going through the registrar and then into the consenter is un-needed and instead you can just create an object which does block validation and wire it into the **channelParticipationAPI** at its creation time, according to the consenter type. 

You can do it in 2 ways: 
1) Have another interface that just validates join requests and invoke it in the channel Participation API.
2) Have a proxy struct which embeds the [ChannelManagement](https://github.com/hyperledger/fabric/blob/master/orderer/common/channelparticipation/restapi.go#L41) and implements only the **JoinChannel** and adds validation checks to it, and then calls further into the next **ChannelManagement** (which is really the registrar) so you'll have:
**channelManagementAPI --> validation layer ---> registrar** and this way you can keep the registrar doing what it is supposed to do and not put additional responsibility to it.

orderer/common/multichannel/registrar.go
yacovm: I am saying the opposite - why have "JoinBlockValidator" for all consenters if it makes sense only to have it for cluster types?

orderer/common/multichannel/registrar.go
tock-ibm: The way to the ConsenterCertificate goes through the consenter, because it is the one initialized with all this material. The way to the consenter goes through the registrar. So i don't see any value in complicating the flow:

api -> abstract-validate(really regeistrar->consenter-validate) -> api -> join -> registrar -> consenter -> handle-chain

instead of 

 api -> join -> registrar ->consenter-validate -> consenter->handle-chain

Note that the type of checks performed in the first part of the current validate method (i.e. non-consenter specific) is not something foreign to the registrar... it does essentially the same [here, checking resources](https://github.com/tock-ibm/fabric/blob/73e5a537deeacc55dff3b352b478a4ae4e45a625/orderer/common/multichannel/registrar.go#L284), and here [parsing config](https://github.com/tock-ibm/fabric/blob/73e5a537deeacc55dff3b352b478a4ae4e45a625/orderer/common/multichannel/registrar.go#L129) ...

orderer/common/multichannel/registrar.go
tock-ibm: You said either or...
Anyway, we chose introspection in order to avoid changing the solo and kafka consenters. But the change is minimal really, 3 lines each. Having it in all would allow us to reject solo and kafka without even looking explicitly at the consensusType and spelling "etcdraft" in the registrar. One may argue it is cleaner. 

The `JoinBlockValidator` was "inspired" so to speak by the [MetadataValidator](https://github.com/tock-ibm/fabric/blob/73e5a537deeacc55dff3b352b478a4ae4e45a625/orderer/consensus/consensus.go#L32) interface that also targets raft only without binding to it directly... maybe its the wrong pattern...

orderer/consensus/etcdraft/consenter.go
caod123: I think the idea here was we wanted to mimic as closely as possible what's actually called in `HandleChain` since that's where things would potentially break if not valid. Because `HandleChain` calls `detectSelfID` we do something similar in the validate. I can see it reaching the same conclusion though so I'd be indifferent with whichever approach we want to use.

yacovm: Checking the certificate is byte-wise equal is a more strict check than what is done in the **detectSelfID**, so it's not possible that your earlier (byte-wise) check passes, but the new one fails.

I mean, you are a server which started up and you are running so your TLS certificate is obviously well formatted, right? 

If it's equal to the one in the consenter set, then the consenter set is also well formatted.

orderer/common/multichannel/registrar.go
yacovm: I said either or because I meant that either we are 100% generic or we are not, but what we have in this code is an attempt to be generic in the global structure but actually when you look deep enough you expect to have only a single type of consenter so we're not generic, hence delegating this consensus specific validation makes little sense because there is only a single implementation of it anyway.

What you're trying to do here is to just follow the existing philosophy of the orderer where everything in the multichannel invokes the consensus support interface, but i think the interface is really overly complex at this moment and is doing way too many stuff than it really should. 

Although it's convenient to extend this pattern i think we should avoid it here because there is really a much easier way to achieve what we want, and that is just abstract out what we need from the consensus validation checks (which i believe is a single function) and inject it into a validation object which the ParticipationAPI would delegate all validation of join channel to.

This has the benefit of reducing the responsibility of the registrar

orderer/common/multichannel/registrar.go
yacovm: > The link between the channel participation api to the rest of the server state is through the registrar only.

I don't see any harm in either adding another interface to it which will validate join blocks, or just proxying the registrar with a struct that performs validation checks.

orderer/common/multichannel/registrar.go
yacovm: > The way to the ConsenterCertificate goes through the consenter, because it is the one initialized with all this material. 

It has the TLS certificate in the consenter, true - but you can also acquire it in the same place where the ParticipationAPI is initialized (the main). 

> 
>  The way to the consenter goes through the registrar. So i don't see any value in complicating the flow:
> 
> api -> abstract-validate(really regeistrar->consenter-validate) -> api -> join -> registrar -> consenter -> handle-chain
> 
> instead of
> 
> api -> join -> registrar ->consenter-validate -> consenter->handle-chain

That's not exactly what I proposed.

I proposed to move all validation logic to a separate struct which would only do that, and not couple the consenter object to validation at all. 
If you need something from the consensus/etcdraft package you can inject a function/interface from where that validation object will be initialized (the same place as the ParticipationAPI will).


> Note that the type of checks performed in the first part of the current validate method (i.e. non-consenter specific) is not something foreign to the registrar... it does essentially the same here, checking resources, and here parsing config ...

This is different because the resource checking is done to initialize a bundle so all errors along the way are fatal errors. 
Similar thing for obtaining the last config from the ledger - it's not a **check**, it is **parsing** without recovery resolution.

In contrast, what you have in your case is validation checks which I do not believe are the responsibility of the registrar.

caod123:
I'll go ahead with @yacovm's comments and move the `ValidateJoinBlock` to a validator in the `channelparticipation` package as well as remove the etcdraft metadata validation from the consenter.

caod123:
After some discussion with @tock-ibm we've decided to only handle the block structure validation at the validation layer after the rest API. JoinChannel will then delegate to a new consenter function in a followup PR that handles the etcdraft metadata level validation.

orderer/common/channelparticipation/validator.go
```diff
@@ -0,0 +1,54 @@
...
+	if !protoutil.IsConfigBlock(configBlock) {
+		return false, errors.New("block is not a config block")
+	}
+
+	envelope, err := protoutil.ExtractEnvelope(configBlock, 0)
+	if err != nil {
+		return false, err
+	}
+
+	cryptoProvider := factory.GetDefault()
```
yacovm, line 32: unrelated to this PR - I think we should look into why we really need BCCSP just to know whether we have consortium config and application config in the Bundle.

yacovm (approved):

PR #1397, opened 2020-06-13
use iter.Seek() instead of creating a new iterator while skipping namespace

cendhu:
#### Type of change

#### Description

Instead of creating a new iterator whenever we want to skip a namespace during a statedb export, it is natural to use Seek() method on iterator. Hence, this PR make that change.


sijocherian:
lgtm

manish-sethi (approved):

PR #1410, opened 2020-06-16
rm no-op error from transient store

cendhu:
#### Type of change

#### Description

We always return a nil error from `transientstore.GetTxPvtRWSetByTxid()`. Hence, this PR remove that no-op error to simplify a few test.

### DIFF
--- a/core/transientstore/store.go
+++ b/core/transientstore/store.go
@@ -153,7 +153,7 @@ func (s *Store) Persist(txid string, blockHeight uint64,
 
 // GetTxPvtRWSetByTxid returns an iterator due to the fact that the txid may have multiple private
 // write sets persisted from different endorsers.
-func (s *Store) GetTxPvtRWSetByTxid(txid string, filter ledger.PvtNsCollFilter) (RWSetScanner, error) {
+func (s *Store) GetTxPvtRWSetByTxid(txid string, filter ledger.PvtNsCollFilter) RWSetScanner {
 
 	logger.Debugf("Getting private data from transient store for transaction %s", txid)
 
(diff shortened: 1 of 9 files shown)


core/transientstore/store.go
```diff
@@ -153,7 +153,7 @@ func (s *Store) Persist(txid string, blockHeight uint64,
 
 // GetTxPvtRWSetByTxid returns an iterator due to the fact that the txid may have multiple private
 // write sets persisted from different endorsers.
-func (s *Store) GetTxPvtRWSetByTxid(txid string, filter ledger.PvtNsCollFilter) (RWSetScanner, error) {
+func (s *Store) GetTxPvtRWSetByTxid(txid string, filter ledger.PvtNsCollFilter) RWSetScanner {
```
manish-sethi, line 156: The existing signature was good but in the function the error is not being checked. In the leveldb iterator, you have to check for the error explicitly - [see usage here](https://github.com/hyperledger/fabric/blob/2fdbafbed9fc1213a28e5980b8c334ce04c27dcb/core/ledger/kvledger/txmgmt/statedb/stateleveldb/stateleveldb.go#L296)

Perhaps, we should handle this right in the beginning in the leveldbhelper package and return error from there, so not to miss this in the upper layers.

manish-sethi (changes requested):

core/transientstore/store.go
cendhu: Ah. So we have an existing bug in the transient store then. Will fix and update the PR.

cendhu: We have `missing error check` at the `stateleveldb` pkg too except the newly introduced `GetFullDBScanner()`.

cendhu:
Closing this as the proper fix is underway that touches many files. Hence, I don't want to keep this PR and instead push a new one.

PR #1412, opened 2020-06-16
[FAB-17993] Add Support For CouchDB 3.x

lindluni:
Couch 3.x requires the user provide an admin user at startup. It also prevents the DB user from modifying the `_users` security objects. This affects the peer startup behavior where it checks if the `_users` table exists and creates the table. Since the user no longer has permission to access the `_users` security object a check was added to confirm the DB version before attempting to modify the `_users` table.

Tests were updated to included an admin user and password for couchDB in the DB configs as well as on the CouchDB runner

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -49,7 +49,7 @@ BASE_VERSION = 2.2.0
 
 # 3rd party image version
 # These versions are also set in the runners in ./integration/runners/
-COUCHDB_VER ?= 2.3
+COUCHDB_VER ?= 3.1
 KAFKA_VER ?= 5.3.1
 ZOOKEEPER_VER ?= 5.3.1
 
--- a/core/ledger/kvledger/tests/v1x_test.go
+++ b/core/ledger/kvledger/tests/v1x_test.go
@@ -381,8 +381,8 @@ func startCouchDBWithV13Data(t *testing.T, ledgerFSRoot string) (*ledger.CouchDB
 	// set required config data to use state couchdb
 	couchdbConfig := &ledger.CouchDBConfig{
 		Address:             couchAddress,
-		Username:            "",
-		Password:            "",
+		Username:            "admin",
+		Password:            "adminpw",
 		MaxRetries:          3,
 		MaxRetriesOnStartup: 3,
 		RequestTimeout:      10 * time.Second,
--- a/core/ledger/kvledger/txmgmt/privacyenabledstate/test_exports.go
+++ b/core/ledger/kvledger/txmgmt/privacyenabledstate/test_exports.go
@@ -156,8 +156,8 @@ func (env *CouchDBTestEnv) Init(t testing.TB) {
 			StateDatabase: "CouchDB",
 			CouchDB: &ledger.CouchDBConfig{
 				Address:             env.couchAddress,
-				Username:            "",
-				Password:            "",
+				Username:            "admin",
+				Password:            "adminpw",
 				MaxRetries:          3,
 				MaxRetriesOnStartup: 20,
 				RequestTimeout:      35 * time.Second,
--- a/core/ledger/kvledger/txmgmt/statedb/statecouchdb/couchdb_test.go
+++ b/core/ledger/kvledger/txmgmt/statedb/statecouchdb/couchdb_test.go
@@ -40,8 +40,8 @@ var assetJSON = []byte(`{"asset_name":"marble1","color":"blue","size":"35","owne
 func testConfig() *ledger.CouchDBConfig {
 	return &ledger.CouchDBConfig{
 		Address:               "",
-		Username:              "",
-		Password:              "",
+		Username:              "admin",
+		Password:              "adminpw",
 		MaxRetries:            3,
 		MaxRetriesOnStartup:   20,
 		RequestTimeout:        35 * time.Second,
@@ -52,8 +52,8 @@ func testConfig() *ledger.CouchDBConfig {
 func TestDBBadConnectionDef(t *testing.T) {
 	config := &ledger.CouchDBConfig{
 		Address:             badParseConnectURL,
-		Username:            "",
-		Password:            "",
+		Username:            "admin",
+		Password:            "adminpw",
 		MaxRetries:          3,
 		MaxRetriesOnStartup: 3,
 		RequestTimeout:      35 * time.Second,
@@ -114,8 +114,8 @@ func TestBadCouchDBInstance(t *testing.T) {
 	badCouchDBInstance := couchInstance{
 		conf: &ledger.CouchDBConfig{
 			Address:             badParseConnectURL,
-			Username:            "",
-			Password:            "",
+			Username:            "admin",
+			Password:            "adminpw",
 			MaxRetries:          3,
 			MaxRetriesOnStartup: 10,
 			RequestTimeout:      30 * time.Second,
@@ -319,8 +319,8 @@ func TestDBBadConnection(t *testing.T) {
 	//Limit the maxRetriesOnStartup to 3 in order to reduce time for the failure
 	config := &ledger.CouchDBConfig{
 		Address:             badConnectURL,
-		Username:            "",
-		Password:            "",
+		Username:            "admin",
+		Password:            "adminpw",
 		MaxRetries:          3,
 		MaxRetriesOnStartup: 3,
 		RequestTimeout:      35 * time.Second,
@@ -854,7 +854,6 @@ func TestCouchDBVersion(t *testing.T) {
 
 	err = checkCouchDBVersion("0.0.0.0")
 	require.Error(t, err, "Error should have been thrown for invalid version")
-
 }
 
 func TestIndexOperations(t *testing.T) {
--- a/core/ledger/kvledger/txmgmt/statedb/statecouchdb/statecouchdb_test.go
+++ b/core/ledger/kvledger/txmgmt/statedb/statecouchdb/statecouchdb_test.go
@@ -52,8 +52,8 @@ func (env *testVDBEnv) init(t *testing.T, sysNamespaces []string) {
 	}
 	config := &ledger.CouchDBConfig{
 		Address:             env.couchDBEnv.couchAddress,
-		Username:            "",
-		Password:            "",
+		Username:            "admin",
+		Password:            "adminpw",
 		InternalQueryLimit:  1000,
 		MaxBatchUpdateSize:  1000,
 		MaxRetries:          3,
@@ -1122,6 +1122,8 @@ func testFormatCheck(t *testing.T, dataFormat string, dataExists bool, expectedE
 	defer os.RemoveAll(redoPath)
 	config := &ledger.CouchDBConfig{
 		Address:             vdbEnv.couchDBEnv.couchAddress,
+		Username:            "admin",
+		Password:            "adminpw",
 		MaxRetries:          3,
 		MaxRetriesOnStartup: 20,
 		RequestTimeout:      35 * time.Second,
--- a/integration/nwo/network.go
+++ b/integration/nwo/network.go
@@ -1212,6 +1212,8 @@ func (n *Network) PeerRunner(p *Peer, env ...string) *ginkgomon.Runner {
 		commands.NodeStart{PeerID: p.ID()},
 		"",
 		fmt.Sprintf("FABRIC_CFG_PATH=%s", n.PeerDir(p)),
+		fmt.Sprintf("CORE_LEDGER_STATE_COUCHDBCONFIG_USERNAME=admin"),
+		fmt.Sprintf("CORE_LEDGER_STATE_COUCHDBCONFIG_PASSWORD=adminpw"),
 	)
 	cmd.Env = append(cmd.Env, env...)
 
--- a/integration/runner/couchdb.go
+++ b/integration/runner/couchdb.go
@@ -23,7 +23,9 @@ import (
 	"github.com/tedsuo/ifrit"
 )
 
-const CouchDBDefaultImage = "couchdb:2.3"
+const CouchDBDefaultImage = "couchdb:3.1"
+const CouchDBUsername = "admin"
+const CouchDBPassword = "adminpw"
 
 // CouchDB manages the execution of an instance of a dockerized CounchDB
 // for tests.
@@ -96,7 +98,11 @@ func (c *CouchDB) Run(sigCh <-chan os.Signal, ready chan<- struct{}) error {
 			Name: c.Name,
 			Config: &docker.Config{
 				Image: c.Image,
-				Env:   []string{"_creator=" + c.creator},
+				Env: []string{
+					fmt.Sprintf("_creator=%s", c.creator),
+					fmt.Sprintf("COUCHDB_USER=%s", CouchDBUsername),
+					fmt.Sprintf("COUCHDB_PASSWORD=%s", CouchDBPassword),
+				},
 			},
 			HostConfig: hostConfig,
 		},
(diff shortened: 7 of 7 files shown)


Makefile
```diff
@@ -49,7 +49,8 @@ BASE_VERSION = 2.2.0
 
 # 3rd party image version
 # These versions are also set in the runners in ./integration/runners/
-COUCHDB_VER ?= 2.3
+COUCHDB_VER ?= 3.1
+COUCHDB_LEGACY_VER ?= 2.3
```
cendhu: As per the CouchDB doc (which we may not trust 😄 based on what we found out), [the upgrade from 2.3 to 3.1](https://docs.couchdb.org/en/stable/install/upgrading.html#standalone-single-node-upgrades) looks very similar to how the peers are being upgraded.  Another option is to simply start a new CouchDB instance and the peer will build the DB automatically.

Is there a reason for supporting both 2.3 & 3.1? something related to release version or handling risk or avoid statedb rebuild?

For CI and unit-test/integration test, should we need to repeat the run on both CouchDB 2.3 and CouchDB 3.1 to ensure we aren't breaking any backward compatibility in the ledger code?

core/ledger/kvledger/txmgmt/statedb/statecouchdb/couchdb_test_export.go
```diff
@@ -22,6 +22,16 @@ func StartCouchDB(t *testing.T, binds []string) (addr string, stopCouchDBFunc fu
 	return couchDB.Address(), func() { couchDB.Stop() }
 }
 
+// StartCouchDB starts the CouchDB if it is not running already
+func StartLegacyCouchDB(t *testing.T, binds []string) (addr string, stopCouchDBFunc func()) {
```
cendhu: As it is present in the `couchdb_test_export.go`, I have to assume that the test would be run only using CouchDB 2.3. I don't know what problem we would face/introduce due to not testing with both CouchDB 2.3 and 3.1.

core/ledger/kvledger/txmgmt/statedb/statecouchdb/couchdbutil.go
```diff
@@ -134,19 +146,50 @@ func createCouchDatabase(couchInstance *couchInstance, dbName string) (*couchDat
 	return &couchDBDatabase, nil
 }
 
+// legacy versions of Couch prior to 3.0.0 did not require a userID and password to be configured when starting
+// the database. If the _users table did not exist in Couch < 3.0 the peer startup process created the _users
+// table. In 3.0.0 the _users table cannot be modified directly. This function can be used to determine if legacy (< 3.0.0)
```
cendhu: It might be possible to modify _users if appropriate config is set -- as per the depreciation notice.

I am fine with the current code as it helps to remove having an assumption on the config.

core/ledger/kvledger/txmgmt/statedb/statecouchdb/couchdbutil.go
```diff
@@ -134,19 +146,50 @@ func createCouchDatabase(couchInstance *couchInstance, dbName string) (*couchDat
...
+func checkLegacyCouchDBVersion(couchVersion string) (bool, error) {
+	couchSemVersion, err := version.NewSemver(couchVersion)
+	if err != nil {
+		return false, errors.Wrap(err, "error parsing CouchDB semver")
+	}
+	gateSemVersion, err := version.NewSemver("3.0.0")
+	if err != nil {
+		return false, errors.Wrap(err, "error parsing default CouchDB semver")
+	}
+	return couchSemVersion.LessThan(gateSemVersion), nil
```
cendhu: Do we need hash-graph's version matching pkg? I thought we would always support one CouchDB version. At the worst case, we might support the previous and current version. Can't we simply compare strings (`"3.1.0", "2.3"`) rather than using a new dependency and making this code generic? Further, LessThan might return true even if the deployment uses `2.0` CouchDB. Hence, I am confused a bit. If there is a plan to support multiple CouchDB versions (more than 2) officially, it would create complication for us whenever we touch `statecouchdb` pkg.

lindluni:
@manish-sethi @cendhu 

I added a unit-test to confirm legacy `_users` functionality still works using a 2.3  couchdb instance.

Do you think I should add a legacy 2.3 E2E integration test? I can run a quick E2E smoke test across a network with couchdb 2.3 just to confirm functionality. Or I can add a second peer to one of the existing integration tests that uses a 2.3 couch instance for its StateDB.

core/ledger/kvledger/txmgmt/statedb/statecouchdb/couchdbutil.go
```diff
@@ -134,19 +146,50 @@ func createCouchDatabase(couchInstance *couchInstance, dbName string) (*couchDat
...
+// legacy versions of Couch prior to 3.0.0 did not require a userID and password to be configured when starting
+// the database. If the _users table did not exist in Couch < 3.0 the peer startup process created the _users
+// table. In 3.0.0 the _users table cannot be modified directly. This function can be used to determine if legacy (< 3.0.0)
+// operations can and should be performed.
+func checkLegacyCouchDBVersion(couchVersion string) (bool, error) {
+	couchSemVersion, err := version.NewSemver(couchVersion)
+	if err != nil {
+		return false, errors.Wrap(err, "error parsing CouchDB semver")
+	}
+	gateSemVersion, err := version.NewSemver("3.0.0")
```
cendhu: Shouldn't it be `3.1.0`?
Screenshot 2020-06-17 at 9 04 34 AM

core/ledger/kvledger/txmgmt/statedb/statecouchdb/couchdbutil_test.go
```diff
@@ -68,17 +68,40 @@ func TestCreateCouchDBSystemDBs(t *testing.T) {
...
 
 	//Retrieve the info for the new database and make sure the name matches
 	dbResp, _, errdb := db.getDatabaseInfo()
-	assert.NoError(t, errdb, "Error when trying to retrieve _users database information")
-	assert.Equal(t, "_users", dbResp.DbName)
+	assert.NoError(t, errdb, "Error when trying to retrieve _replicator database information")
+	assert.Equal(t, "_replicator", dbResp.DbName)
 
-	db = couchDatabase{couchInstance: couchInstance, dbName: "_replicator"}
+	db = couchDatabase{couchInstance: couchInstance, dbName: "_global_changes"}
```
cendhu: Not related to changes made in this PR. We need to understand the last bullet in https://docs.couchdb.org/en/latest/whatsnew/3.0.html#performance better and first bullet in https://docs.couchdb.org/en/latest/whatsnew/3.0.html#deprecated-feature-removal.

cendhu (commented):
Thanks, Brett. I have few doubts which I have raised in the PR. Once I understand the scope of this PR, I will do an in-depth review.

cendhu:
> Do you think I should add a legacy 2.3 E2E integration test? I can run a quick E2E smoke test across a network with couchdb 2.3 just to confirm functionality. Or I can add a second peer to one of the existing integration tests that uses a 2.3 couch instance for its StateDB.

I was reviewing the PR and saw this after submitting my reviews. I had raised a similar question in one my comment. 

We do use CouchDB at the `core/ledger/kvledger/tests` pkg which is a ledger level integration test for us to ensure backward compatibility since v1.0. In every major release, we add complete ledgerData folder to ensure we aren't breaking anything. If we plan to support two versions of CouchDB, it may be necessary to run these test on both (brute force approach) or the CouchDB version may not be matter (if we think carefully and analyse the complete impact).

core/ledger/kvledger/txmgmt/statedb/statecouchdb/couchdbutil.go
lindluni: With this we would support 3.1.x and 2.3.x. Today in fabric we actually test against 2.3 (though there is no actual 2.3, its whatever the latest 2.3.x version is), so it's whatever is latest, no idea until runtime what the patch level is. And I actually only chose to switch to the semver package because it seemed less error prone than doing string splits and comparisons. I also only elected to use it as the package is already part of fabrics dep tree.

If there is other dissent on this I'll switch it to a string matcher

core/ledger/kvledger/txmgmt/statedb/statecouchdb/couchdbutil.go
lindluni: I'll clarify the comment, and point to the deprecation notice.

core/ledger/kvledger/txmgmt/statedb/statecouchdb/couchdb_test_export.go
lindluni: The proposal here is all testing is done using 3.1, the only testing done on 2.3 is the auto-creation of the `_users` table. Unless we think it should be reversed, all testing on 2.3, with only the _users case tested on 3.1. But my opinion is that we should move forward and 3.1 should become our new standard.

Makefile
lindluni: We could drop support for 2.3 altogether in the 2.{2,next} stream, and then a lot of these changes go away. TBH, I would prefer that in 2.{2,next} we say the minimum required level of Couch is 3.1. But Couch 2 has been around for nearly 10 years (so getting users to migrate may not be easy, as 3.x is not as tested), and our code is not backwards compatible if we remove support for 2.3, so that would require a major version change to rip out the creation of the `_users` table (which is required for removing support for 2.3).

lindluni:
> > Do you think I should add a legacy 2.3 E2E integration test? I can run a quick E2E smoke test across a network with couchdb 2.3 just to confirm functionality. Or I can add a second peer to one of the existing integration tests that uses a 2.3 couch instance for its StateDB.
> 
> I was reviewing the PR and saw this after submitting my reviews. I had raised a similar question in one my comment.
> 
> We do use CouchDB at the `core/ledger/kvledger/tests` pkg which is a ledger level integration test for us to ensure backward compatibility since v1.0. In every major release, we add complete ledgerData folder to ensure we aren't breaking anything. If we plan to support two versions of CouchDB, it may be necessary to run these test on both (brute force approach) or the CouchDB version may not be matter (if we think carefully and analyse the complete impact).

There are quite a few integration scenarios I can think through here. We will need to have a discussion on a plan at large for what we want to do, wrt integration tests and version support

Makefile
ale-linux: I'm not a ledger expert so I'll defer to @cendhu, @manish-sethi and @denyeart. However, my recommendation would be
- to only upgrade couch versions with a good reason (e.g. a new feature of the new version of couch is necessary; or the old couch version will no longer receive security updates, or it has reached EoL)
- to only upgrade couch version across (major) releases
- to never support more than one couch version per release

With this in mind, I would recommend (once again, deferring to the expert) not to go into the business of supporting multiple versions at the same time.

cendhu: @ale-linux Agree. There is a bug (http://danluu.com/fsyncgate/) (https://github.com/apache/couchdb/commit/3505281559513e2922484ebf0996a8846dcc0a34) that is fixed only in CouchDB 3.x and we do need them. At the same time, we are not sure about the stability of CouchDB 3.x (new release). We haven't decided yet on supporting multiple CouchDB versions (mostly, we shouldn't).

denyeart:
I don't think we need all the complexity of separate v2.x and v3.x support.
I think we can keep the peer code generic between couchdb 2.x and 3.x. In 3.x admin can still create _users table, I think we just need to suppress the update on _users table that creates the security object. This can be suppressed on both 2.x and 3.x deployments. Basically just add one if statement to skip for _users table in https://github.com/hyperledger/fabric/blob/master/core/ledger/kvledger/txmgmt/statedb/statecouchdb/couchdb.go#L1426

lindluni:
> I don't think we need all the complexity of separate v2.x and v3.x support.
> I think we can keep the peer code generic between couchdb 2.x and 3.x. In 3.x admin can still create _users table, I think we just need to suppress the update on _users table that creates the security object. This can be suppressed on both 2.x and 3.x deployments. Basically just add one if statement to skip for _users table in https://github.com/hyperledger/fabric/blob/master/core/ledger/kvledger/txmgmt/statedb/statecouchdb/couchdb.go#L1426

I refactored `createDatabaseIfNotExist()` to have a single entrypoint into `applyDatabasePermissions()` (as opposed to the existing three entrypoints) so I could perform the check there, rather than doing a check and no-op in  `createDatabaseIfNotExist()`

lindluni:
I will add an integration test yet that does an e2e smoke test against 2.3

core/ledger/kvledger/txmgmt/statedb/statecouchdb/couchdb.go
```diff
@@ -232,73 +231,45 @@ func (dbclient *couchDatabase) createDatabaseIfNotExist() error {
...
+	if dbclient.dbName != "_users" {
+		errSecurity := dbclient.applyDatabasePermissions()
+		if errSecurity != nil {
+			return errSecurity
+		}
 	}
-
-	logger.Infof("Created state database %s", dbclient.dbName)
-
+	logger.Debugf("[%s] Database already exists", dbclient.dbName)
```
denyeart: I think this debug is misplaced... it will print this message even if the database gets created in this function.

lindluni:
Removed all references to the legacy 2.3 Couch. So this now fully supports Couch 3.1 and does all testing on 3.1

core/ledger/kvledger/txmgmt/statedb/statecouchdb/couchdb.go
```diff
@@ -232,73 +231,47 @@ func (dbclient *couchDatabase) createDatabaseIfNotExist() error {
 		}
 	}
 
-	//If the dbInfo returns populated and status code is 200, then the database exists
-	if dbInfo != nil && couchDBReturn.StatusCode == 200 {
+	if dbInfo == nil || couchDBReturn.StatusCode != 200 {
```
manish-sethi: Since you inverted the condition, I find that explicitly checking `dbInfo == nil && couchDBReturn.StatusCode == 404` makes it more readable.

Just to take a note, the both parts of the condition are actually same but it seems that they were always like that.

core/ledger/kvledger/txmgmt/statedb/statecouchdb/couchdb.go
```diff
@@ -232,73 +231,47 @@ func (dbclient *couchDatabase) createDatabaseIfNotExist() error {
...
+		if err != nil {
+			// Check to see if the database exists
+			// Even though handleRequest() returned an error, the
+			// database may have been created and a false error
+			// returned due to a timeout or race condition.
+			// Do a final check to see if the database really got created.
+			dbInfo, couchDBReturn, dbInfoErr := dbclient.getDatabaseInfo()
+			//If there is no error, then the database exists,  return without an error
+			if dbInfoErr != nil || dbInfo == nil || couchDBReturn.StatusCode != 200 {
+				return err
```
manish-sethi: You reversed the logic of "suppressing the error" into returning the error. However, the comment is outdated now...

core/ledger/kvledger/txmgmt/statedb/statecouchdb/couchdb.go
lindluni: Done

core/ledger/kvledger/txmgmt/statedb/statecouchdb/couchdb.go
lindluni: removed the comment

lindluni:
opened #1430 to remove the 3.1 migration and only skip the security object modification. Review should move there. This change will automatically rebase itself on master once that change is merged (GitHub now autorebases when commits are merged from other PR's)

lindluni:
@Mergifyio rebase

denyeart (approved):
I like this much better than the dual support of 2.3 and 3.1 PR :-)

PR #1428, opened 2020-06-20
check leveldb iterator error

cendhu:
#### Type of change

- Bug fix

#### Description

Goleveldb iterator does not explicitly return an error. We need to call the `itr.Error()` method on the iterator to check for the presence of any error. As the check is missing at a few places, this PR make the `leveldbhelper` pkg itself to return an error by explicitly calling `itr.Error()`
#### Additional details

In the subsequent PR, we would supply the fix for iter.Next() as well.

#### Related issues

### DIFF
--- a/common/ledger/util/leveldbhelper/leveldb_provider.go
+++ b/common/ledger/util/leveldbhelper/leveldb_provider.go
@@ -12,6 +12,7 @@ import (
 	"sync"
 
 	"github.com/hyperledger/fabric/common/ledger/dataformat"
+	"github.com/pkg/errors"
 	"github.com/syndtr/goleveldb/leveldb"
 	"github.com/syndtr/goleveldb/leveldb/iterator"
 )
@@ -172,7 +173,7 @@ func (h *DBHandle) WriteBatch(batch *UpdateBatch, sync bool) error {
 // GetIterator gets an handle to iterator. The iterator should be released after the use.
 // The resultset contains all the keys that are present in the db between the startKey (inclusive) and the endKey (exclusive).
 // A nil startKey represents the first available key and a nil endKey represent a logical key after the last available key
-func (h *DBHandle) GetIterator(startKey []byte, endKey []byte) *Iterator {
+func (h *DBHandle) GetIterator(startKey []byte, endKey []byte) (*Iterator, error) {
 	sKey := constructLevelKey(h.dbName, startKey)
 	eKey := constructLevelKey(h.dbName, endKey)
 	if endKey == nil {
@@ -180,7 +181,12 @@ func (h *DBHandle) GetIterator(startKey []byte, endKey []byte) *Iterator {
 		eKey[len(eKey)-1] = lastKeyIndicator
 	}
 	logger.Debugf("Getting iterator for range [%#v] - [%#v]", sKey, eKey)
-	return &Iterator{h.dbName, h.db.GetIterator(sKey, eKey)}
+	itr := h.db.GetIterator(sKey, eKey)
+	if err := itr.Error(); err != nil {
+		itr.Release()
+		return nil, errors.Wrapf(err, "internal leveldb error while obtaining db iterator")
+	}
+	return &Iterator{h.dbName, itr}, nil
 }
 
 // UpdateBatch encloses the details of multiple `updates`
--- a/core/ledger/pvtdatastorage/store.go
+++ b/core/ledger/pvtdatastorage/store.go
@@ -586,7 +586,10 @@ func (s *Store) GetPvtDataByBlockNum(blockNum uint64, filter ledger.PvtNsCollFil
 	}
 	startKey, endKey := getDataKeysForRangeScanByBlockNum(blockNum)
 	logger.Debugf("Querying private data storage for write sets using startKey=%#v, endKey=%#v", startKey, endKey)
-	itr := s.db.GetIterator(startKey, endKey)
+	itr, err := s.db.GetIterator(startKey, endKey)
+	if err != nil {
+		return nil, err
+	}
 	defer itr.Release()
 
 	var blockPvtdata []*ledger.TxPvtData
@@ -658,7 +661,10 @@ func (s *Store) GetMissingPvtDataInfoForMostRecentBlocks(maxBlock int) (ledger.M
 	lastCommittedBlock := atomic.LoadUint64(&s.lastCommittedBlock)
 
 	startKey, endKey := createRangeScanKeysForEligibleMissingDataEntries(lastCommittedBlock)
-	dbItr := s.db.GetIterator(startKey, endKey)
+	dbItr, err := s.db.GetIterator(startKey, endKey)
+	if err != nil {
+		return nil, err
+	}
 	defer dbItr.Release()
 
 	for dbItr.Next() {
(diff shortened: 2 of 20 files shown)


ale-linux (commented):
Thanks @cendhu. Could you please clarify the implications of this change to the committer path (if any)? Just to make sure we don't introduce any change that may lead to a fork. Thanks!

cendhu:
Good question, @ale-linux. Sorry for the long answer but I had to be comprehensive to convey that this PR is not adding a fork possibilities.

**Answer**
In this PR, the function signature has been changed to return an error ONLY for the following ledger APIs
- `func (h *DBHandle) GetIterator(startKey []byte, endKey []byte) (*Iterator, error) {`
- `func newMetadataHint(bookkeeper *leveldbhelper.DBHandle) (*metadataHint, error) {`
- `func (d *db) getNamespaceIterator(ns string) (*leveldbhelper.Iterator, error) {`

Given that all callers of these function already return an error, this PR does not introduce a new error in general and should not introduce a fork as per my analysis. In fact, only a very few functions missed to check `iter.Error()`.

Only in a goroutine in the ledger, this PR introduces a panic and is debatable. 
https://github.com/hyperledger/fabric/pull/1428/files#diff-19d675f588314393c7a74b5a587c1939R828-R831

This goroutine receives collection membership change event and moves ineligible missing data entries to eligible entries (as the peer has become a member now). If any issue occurs while reading entries from the db, the peer would panic as this error can occur mainly due to corrupted filesystem. Further, a single leveldb instance is used for all db. Hence, it is not possible to abort only a particular chain as well. 

**Additional Details** 
Any LevelDB iterator error might occur due to either 
- a corrupted filesystem or corrupted LevelDB files or 
- a closed DB (this should never occur unless there is a bug in the code). 

Hence, an iterator would return an error in a rare case only and the caller needs to decide whether to 

1.  Propagate the error to whoever called it and so on. 
2. Panic the peer. https://github.com/hyperledger/fabric/blob/c4d38024326e69bb85139ae4704e618f0f865d7a/gossip/state/state.go#L581
3. Stop processing the chain. https://github.com/hyperledger/fabric/blob/c4d38024326e69bb85139ae4704e618f0f865d7a/gossip/state/state.go#L576-L579
4. Log a warning message and continue. 
https://github.com/hyperledger/fabric/blob/c4d38024326e69bb85139ae4704e618f0f865d7a/gossip/privdata/pvtdataprovider.go#L297-L307

I am not going into the details of why the above functionalities result in a panic or an error or a warning as this response is already a big one 😄 (in short, it depends on the severity. For example, any error happens while fetching a pvtdata from other peers need not to panic or stop processing a chain). 

I am also aware of what happened with vscc and couchdb error but this PR wouldn't make a similar issue to reoccur. 

If I am missing something, please let me know.

core/ledger/pvtdatastorage/store.go
```diff
@@ -816,7 +825,10 @@ func (s *Store) processCollElgEvents() {
 	s.purgerLock.Lock()
 	defer s.purgerLock.Unlock()
 	collElgStartKey, collElgEndKey := createRangeScanKeysForCollElg()
-	eventItr := s.db.GetIterator(collElgStartKey, collElgEndKey)
+	eventItr, err := s.db.GetIterator(collElgStartKey, collElgEndKey)
+	if err != nil {
+		panic(err)
```
manish-sethi: Here and below, I would replace `panic` by an error return and then log the error in the caller. `panic` would be too harsh for a situation that neither causes all the functionality to be unavailable nor causes some data corruption. If we want to take more noticeable step, we could cause the channel to be declared as unusable but that may require more changes than what is in the scope of this PR.
So, I would just look at this PR something that improves the current code by logging the error (otherwise, the current behavior is not changed at all).

manish-sethi (changes requested):

common/ledger/util/leveldbhelper/leveldb_provider.go
```diff
@@ -172,15 +173,19 @@ func (h *DBHandle) WriteBatch(batch *UpdateBatch, sync bool) error {
...
 	eKey := constructLevelKey(h.dbName, endKey)
 	if endKey == nil {
 		// replace the last byte 'dbNameKeySep' by 'lastKeyIndicator'
 		eKey[len(eKey)-1] = lastKeyIndicator
 	}
 	logger.Debugf("Getting iterator for range [%#v] - [%#v]", sKey, eKey)
-	return &Iterator{h.dbName, h.db.GetIterator(sKey, eKey)}
+	itr := h.db.GetIterator(sKey, eKey)
+	if err := itr.Error(); err != nil {
+		return nil, errors.Wrapf(err, "internal leveldb error while obtaining db iterator")
```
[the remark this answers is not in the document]
cendhu: Agree.

core/ledger/pvtdatastorage/store.go
cendhu: I do not have a strong opinion on this. I agree that only functionality that gets affected is that the new collection member would not get the private data. This could leads to endorsement failures due to missing pvtdata in the stateDB. When that happens, it is an operational challenge to see the error from the log and then use a different disk or new peer (depending on the root cause of the error). Hence, I thought it is better to panic here to ease operational task. If you prefer to only log errors, I will update the code.

manish-sethi: What you mentioned above is the current behavior as well. No change in that as such, except that the error would now be logged and the task would not be cleared off.
If you want to change the behavior, IMO, the error state should be recorded such that the private data retrieval function should report the error to the caller. If you want to go that far, I would encourage to deal with that in a separate PR. But, as I mentioned before, `panic` is too brutal for this error.

cendhu: As this is a goroutine, the error propagation is quite hard to do. A larger refactoring is required as you mentioned. For now, I have reverted to the old behaviour + error logs.

core/ledger/pvtdatastorage/store.go
```diff
@@ -801,22 +810,31 @@ func (s *Store) retrieveExpiryEntries(minBlkNum, maxBlkNum uint64) ([]*expiryEnt
 
 func (s *Store) launchCollElgProc() {
 	go func() {
-		s.processCollElgEvents() // process collection eligibility events when store is opened - in case there is an unprocessed events from previous run
+		if err := s.processCollElgEvents(); err != nil {
+			// process collection eligibility events when store is opened -
+			// in case there is an unprocessed events from previous run
+			logger.Errorf("failed to process collection eligibility event [%s]", err)
+		}
```
manish-sethi: [%s] -> %+v (print stack trace as well). Also, it may be good to start using `logger.Errorw` that would print the error as a separate field.

core/ledger/pvtdatastorage/store.go
```diff
@@ -801,22 +810,31 @@ func (s *Store) retrieveExpiryEntries(minBlkNum, maxBlkNum uint64) ([]*expiryEnt
...
 		for {
 			logger.Debugf("Waiting for collection eligibility event")
 			s.collElgProcSync.waitForNotification()
-			s.processCollElgEvents()
+			if err := s.processCollElgEvents(); err != nil {
+				logger.Errorf("failed to process collection eligibility event [%s]", err)
```
manish-sethi: same as above

core/ledger/pvtdatastorage/store.go
```diff
@@ -801,22 +810,31 @@ func (s *Store) retrieveExpiryEntries(minBlkNum, maxBlkNum uint64) ([]*expiryEnt
 
 func (s *Store) launchCollElgProc() {
 	go func() {
-		s.processCollElgEvents() // process collection eligibility events when store is opened - in case there is an unprocessed events from previous run
+		if err := s.processCollElgEvents(); err != nil {
+			// process collection eligibility events when store is opened -
+			// in case there is an unprocessed events from previous run
+			logger.Errorw("failed to process collection eligibility events", "err", err)
+		}
```
manish-sethi, line 817: Thanks @cendhu. It will be good to add a test for this log event, as this is kind of part of the behavior that we count on. I am fine merging this PR, has it has wide spread change. You can do this in a separate PR.

manish-sethi (dismissed):

core/ledger/pvtdatastorage/store.go
cendhu: Done.

core/ledger/pvtdatastorage/store.go
cendhu: Done.

core/ledger/pvtdatastorage/store.go
cendhu: Done.

manish-sethi (approved):

PR #1430, opened 2020-06-20
[FAB-17993] Skip modifying _users security object

lindluni:
Introduces a check to skip modifying the security object of the `_users` table. The latest versions of CouchDB protect from this operation being performed and it isn't necessary for the peer to actually modify the object to perform its functionality

### DIFF
--- a/core/ledger/kvledger/txmgmt/statedb/statecouchdb/couchdb.go
+++ b/core/ledger/kvledger/txmgmt/statedb/statecouchdb/couchdb.go
@@ -222,7 +222,6 @@ func closeResponseBody(resp *http.Response) {
 
 //createDatabaseIfNotExist method provides function to create database
 func (dbclient *couchDatabase) createDatabaseIfNotExist() error {
-
 	logger.Debugf("[%s] Entering CreateDatabaseIfNotExist()", dbclient.dbName)
 
 	dbInfo, couchDBReturn, err := dbclient.getDatabaseInfo()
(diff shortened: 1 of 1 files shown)


denyeart:
LGTM and this does simplify it somewhat, but would also like @manish-sethi or @cendhu to look before merging though.

core/ledger/kvledger/txmgmt/statedb/statecouchdb/couchdb.go
```diff
@@ -232,73 +231,46 @@ func (dbclient *couchDatabase) createDatabaseIfNotExist() error {
...
-	defer closeResponseBody(resp)
 
-	errSecurity := dbclient.applyDatabasePermissions()
-	if errSecurity != nil {
-		return errSecurity
+	if dbclient.dbName != "_users" {
+		errSecurity := dbclient.applyDatabasePermissions()
+		if errSecurity != nil {
+			return errSecurity
+		}
```
cendhu, line 269: Can't we just move the dbName check to the `applyDatabasePermissions()` itself? This ensures that this concept of applying permission to non `_user` table is kept close to each other.

cendhu: This will also help in returning early as per line 262 if db exists already?

lindluni: It's generally bad practice for the first check of a function to be a singular condition on which the function simply returns without an error. Its like passing a value that could potentially be nil into a function, it's the caller's responsibility to check the state of the objects being invoked on, not the callee.

manish-sethi: I am fine with this as is.

manish-sethi (approved):

core/ledger/kvledger/txmgmt/statedb/statecouchdb/couchdb.go
cendhu: I see. Thanks, Brett. Make sense for the caller to do it instead of callee.

btl5037:

PR #1431, opened 2020-06-21
FAB-17910 Ch.Pat.API: join member w/o onboarding

tock-ibm:
This commit handles the following scenario:
* The orderer is a member of the cluster - the config block given at
  join has the self ID of the orderer.
* The config block given at join is number 0 - no onboarding needed.

In this case the join block is first appended to the ledger as a
genesis block, and the multichannel.ChainSupport is created in the
normal fashion, reading the last-block from the ledger and constructing
the consensus.Chain accordingly.

In addition this commit prepares the ground for joining with a join block with number >0, which requires on-boarding. For that purpose a new method is added to the consensus.Consenter interface.

This commit also handles the case of an orderer rejecting a join to the system channel when there are already application channels (FAB-17913), as it is just two lines of code (and a test).

Change-Id: I0b2bd4e4da901c6b9c53585bffc6dc0dd5573948

#### Type of change

- New feature
#### Related issues

Tasks: FAB-17910 FAB-17913
Story: FAB-15711
Epic: FAB-17712

### DIFF
--- a/orderer/common/multichannel/registrar.go
+++ b/orderer/common/multichannel/registrar.go
@@ -165,7 +165,10 @@ func (r *Registrar) Initialize(consenters map[string]consensus.Consenter) {
 		if configTx == nil {
 			logger.Panic("Programming error, configTx should never be nil here")
 		}
-		ledgerResources := r.newLedgerResources(configTx)
+		ledgerResources, err := r.newLedgerResources(configTx)
+		if err != nil {
+			logger.Panicf("Error creating ledger resources: %s", err)
+		}
 		channelID := ledgerResources.ConfigtxValidator().ChannelID()
 
 		if _, ok := ledgerResources.ConsortiumsConfig(); ok {
@@ -173,7 +176,7 @@ func (r *Registrar) Initialize(consenters map[string]consensus.Consenter) {
 				logger.Panicf("There appear to be two system channels %s and %s", r.systemChannelID, channelID)
 			}
 
-			chain := newChainSupport(
+			chain, err := newChainSupport(
 				r,
 				ledgerResources,
 				r.consenters,
@@ -181,6 +184,9 @@ func (r *Registrar) Initialize(consenters map[string]consensus.Consenter) {
 				r.blockcutterMetrics,
 				r.bccsp,
 			)
+			if err != nil {
+				logger.Panicf("Error creating chain support: %s", err)
+			}
 			r.templator = msgprocessor.NewDefaultTemplator(chain, r.bccsp)
 			chain.Processor = msgprocessor.NewSystemChannel(
 				chain,
@@ -209,7 +215,7 @@ func (r *Registrar) Initialize(consenters map[string]consensus.Consenter) {
 			defer chain.start()
 		} else {
 			logger.Debugf("Starting channel: %s", channelID)
-			chain := newChainSupport(
+			chain, err := newChainSupport(
 				r,
 				ledgerResources,
 				r.consenters,
@@ -217,6 +223,9 @@ func (r *Registrar) Initialize(consenters map[string]consensus.Consenter) {
 				r.blockcutterMetrics,
 				r.bccsp,
 			)
+			if err != nil {
+				logger.Panicf("Error creating chain support: %s", err)
+			}
 			r.chains[channelID] = chain
 			chain.start()
 		}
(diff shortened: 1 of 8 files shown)


orderer/common/multichannel/registrar.go
```diff
@@ -430,8 +450,42 @@ func (r *Registrar) JoinChannel(channelID string, configBlock *cb.Block) (types.
 		return types.ChannelInfo{}, types.ErrChannelAlreadyExists
 	}
 
-	//TODO
-	return types.ChannelInfo{}, errors.New("Not implemented yet")
+	if !isAppChannel && len(r.chains) > 0 {
+		return types.ChannelInfo{}, types.ErrAppChannelsAlreadyExists
```
yacovm, line 454: shouldn't this be "system channel already exists" ?

orderer/common/multichannel/registrar.go
```diff
@@ -430,8 +450,42 @@ func (r *Registrar) JoinChannel(channelID string, configBlock *cb.Block) (types.
...
+	}
+
+	configEnv, err := protoutil.ExtractEnvelope(configBlock, 0)
+	if err != nil {
+		return types.ChannelInfo{}, errors.Wrap(err, "failed extracting config envelope from block")
+	}
+
+	//TODO save the join-block in the file repo to make this action crash tolerant.
```
yacovm, line 462: do we have an API that defines how to do that? If yes can we put this here?

orderer/consensus/consensus.go
```diff
@@ -24,6 +24,12 @@ type Consenter interface {
 	// the last block committed to the ledger of this Chain. For a new chain, or one which is migrated,
 	// this metadata will be nil (or contain a zero-length Value), as there is no prior metadata to report.
 	HandleChain(support ConsenterSupport, metadata *cb.Metadata) (Chain, error)
+
+	// JoinChain should create and return a reference to a Chain when the channel is started with a join-block
+	// that is not a genesis block (i.e. join-block.number>0), and a ledger with height <= than join-block.number.
+	// This means that there is a gap between the join-block and the last block in the ledger that requires
+	// on-boarding. In contrast, HandleChain creates a Chain based on the last block found in the ledger.
+	JoinChain(support ConsenterSupport, joinBlock *cb.Block) (Chain, error)
```
yacovm, line 32: I don't understand why we should have a different initialization flow for the chain in this case? Why don't we always call **HandleChain** anyway?

orderer/common/multichannel/registrar.go
tock-ibm: No, This is for when a user is trying to create a system channel (`isAppChannel=false`), but the  orderer is already working with app. channel only mode (i.e., with participation API). This not supported. We only (will) allow the creation of the system channel wit a join when there are no channels at all.

orderer/common/multichannel/registrar.go
tock-ibm: We are working on that in parallel, it is defined in tasks:
https://jira.hyperledger.org/browse/FAB-17960,
https://jira.hyperledger.org/browse/FAB-18009,
https://jira.hyperledger.org/browse/FAB-18008.

After a discussion with the ledger team & Jason Yellick, the plan is to use an "atomic file repo" to save the join blocks. The repo will be scanned on startup to reconcile the join state of chains with their ledger. 

BTW, the same file repo code will be used to make ledger removal CFT, see here:
https://jira.hyperledger.org/browse/FAB-17868

orderer/consensus/consensus.go
tock-ibm: `HandleChain` starts a chain based on the last block in the ledger, which it gets from the ledger resources.

What we need in order to join a chain in cases when on-boarding is needed (see special case below) is to start a `follower.Chain`, with a join-block, when the ledger is empty. The follower chain does not need many of the resources that an etcdraft.Chain needs, as it only pulls blocks (for example, it never signs and never uses the BlockWriter) . Moreover, some things it cannot even create, like a BlockWriter, which need to be based on a tip of a ledger (which we do not have at this point). Rather then stud HandleChain with many "ifs and buts", and extend its interface with the join-block, I think it will be much clearer and safer (not introducing bugs to HandleChain) to define a new method that does exactly the job of joining a channel. 

The way I want to think about it, conceptually, HandleChain starts from a valid the ledger, so a failure means that something is fatally wrong (e.g. corruption of the ledger). In contrast, JoinChain is started with input from the user, i.e. the join-block, and errors in processing his request should be propagated back to the user. 

A special case is when the join-block has number=0, and then it is appended to the ledger and `HandleChain` is called. If the orderer is a member (`detectSelfID()` no error), an `etcdraft.Chain` is created. Otherwise, a follower.Chain be created. However, this follower.Chain is past on-boarding and is just pulling blocks and observing the config.

When considering CFT, an orderer might fail in the middle of on-boarding a channel. When we recover, we'll find a join-block.Number >= ledger.Height, which indicates a gap between the ledger tip and the join-block. In this case we'll also call JoinChain rather than HandleChain. 

In summary, I think that all these special cases and deviations from what HandleChain currently does merit a new method.

yacovm: What I am missing here is to understand the complete flow - once the **follower.Chain** has replicated blocks and decides that it should mutate into a fully fledged **etcdraft.Chain**, how is this done? Is it via a mechanism similar to the **inactive.Chain**  ? 

> The follower chain does not need many of the resources that an etcdraft.Chain needs, as it only pulls blocks (for example, it never signs and never uses the BlockWriter) . Moreover, some things it cannot even create, like a BlockWriter, which need to be based on a tip of a ledger (which we do not have at this point). Rather then stud HandleChain with many "ifs and buts", and extend its interface with the join-block, I think it will be much clearer and safer (not introducing bugs to HandleChain) to define a new method that does exactly the job of joining a channel.

1. If we need less things, it's easy to just not use what we don't need in the follower chain
2. There is already an "if" in the **HandleChain** which makes it return an **Inactive.Chain** , shouldn't it be pretty easy to classify when you need to spawn a follower chain?

>  When considering CFT, an orderer might fail in the middle of on-boarding a channel. When we recover, we'll find a join-block.Number >= ledger.Height, which indicates a gap between the ledger tip and the join-block. In this case we'll also call JoinChain rather than HandleChain.

What do you mean? Do we call JoinChain synchronously from the user-end ? Or do we decide autonomously?

I guess you mean autonomously and then i don't see why we can't do this check in the **HandleChain**.

tock-ibm: Here is the flow of a join request with `join-block.Number >0`. 

- The `JoinChain' is called synchronously from the client request go-routine. If it fails, the request fails. If it succeeds, a follower.Chain is created and started (it has a go-routine). 
- The follower.Chain is given a function pointer to a `createChain` function just like in the `inactive.Chain` and the haltCallback in the `etcdraft.Chain`. 
- The follower.Chain starts pulling blocks. It is in status `on-boarding` until it reaches `join-block.Number`, then it changes to `active`.
- If at that point  the orderer is in the consenters set of the  join-block ("a member"), the call to `createChain` in invoked, triggering the transition between follower.Chain -> etcdraft.Chain. This happens with a call to HandleChain. 
- HandleChain has a switch to create a follower.Chain when there is no sys-chan, and an inactive.Chain when there is a sys-chan.
- If the orederer is not  in the consenters set of the  join-block, it will continue to pull blocks and watch config blocks.
- Note that once the ledger catches up with the join-block, the join-block is deleted.
- When a config block arrives in which the orderer is a member, `createChain` in invoked, again, triggering the transition between follower,Chain -> etcdraft.Chain.  This happens with a call to HandleChain. 

Conversely, here is the flow of an etcdraft.Chain evicted:
- Eviction is detected, halt is called (as usual)
- The haltCallback is initialized to create a chain. Calls HandleChain. HandleChain creates a follower.Chain without a join-block.

So, `JoinChain` is normally called  by the client request go-routine, and always when there is a join-block. HandleChain is called by internal go-routines, and when the chain is started off the ledger. 

The `JoinChain` call behaves differently from `HandleChain` only when  `join-block.Number >0`, i.e. there is a need for on-boarding. Otherwise, if it is  `join-block.Number==0`, it just appends the block to the ledger and delegates to `HandleChain`. That is what's addressed in this commit, all the rest is future tasks.

I hope that answers your question. I will also document this in the Jira.

yacovm: Thanks for the detailed descriptions. I have to say that I still don't like the idea that we have 2 different disjoint APIs for chain initialization and in 1 of them we actually have a fork (inactive + regular).

However, I think that an obstacle in having the follower chain being initialized in the regular **HandleChain** is that we would need to see if the join block exists or not, which as I understand, would require us to rappel down yet another API through the consenter support which is already a mess... 

Let's via the current approach for now but I think for the future there would be a need for tidying up the place.

yacovm (approved):

PR #1435, opened 2020-06-22
Remove call to couchdb API _ensure_full_commit

manish-sethi:

#### Type of change

#### Description
This PR removes the invocation to the couchdb API "_ensure_full_commit".

#### Additional details
We recommended `delayed_commits=true` in couchdb config for better performance and then at the block commit time, we invoke `_ensure_full_commit`. This arrangement was under the assumption that the `_ensure_full_commit` fsyncs the data as per the couchdb docs https://docs.couchdb.org/en/2.3.1/api/database/compact.html#db-ensure-full-commit. However, after talking to couchdb experts, we figured that this was true only till couchdb 1.x. Since then, this is more or less a no-op call as far as `fsync` is concerned.

#### Related issues
https://jira.hyperledger.org/browse/FABB-151
https://jira.hyperledger.org/browse/FAB-17999

### DIFF
--- a/core/ledger/kvledger/txmgmt/statedb/statecouchdb/statecouchdb.go
+++ b/core/ledger/kvledger/txmgmt/statedb/statecouchdb/statecouchdb.go
@@ -122,19 +122,8 @@ func writeDataFormatVersion(couchInstance *couchInstance, dataformatVersion stri
 	if err != nil {
 		return err
 	}
-	if _, err := db.saveDoc(dataformatVersionDocID, "", doc); err != nil {
-		return err
-	}
-	dbResponse, err := db.ensureFullCommit()
-
-	if err != nil {
-		return err
-	}
-	if !dbResponse.Ok {
-		logger.Errorf("failed to perform full commit while writing dataformat version")
-		return errors.New("failed to perform full commit while writing dataformat version")
-	}
-	return nil
+	_, err = db.saveDoc(dataformatVersionDocID, "", doc)
+	return err
 }
 
 // GetDBHandle gets the handle to a named database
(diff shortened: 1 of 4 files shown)


core/ledger/kvledger/txmgmt/statedb/statecouchdb/statecouchdb.go
```diff
@@ -685,8 +674,22 @@ func (vdb *VersionedDB) postCommitProcessing(committers []*committer, namespaces
...
+				go db.runWarmIndexAllIndexes()
+				db.indexWarmCounter = 0
+			}
+			db.indexWarmCounter++
+		}
+	}
+
 	// Record a savepoint at a given height
-	if err := vdb.ensureFullCommitAndRecordSavepoint(height, namespaces); err != nil {
+	if err := vdb.recordSavepoint(height, namespaces); err != nil {
```
cendhu: Out of scope for this PR. Given that `recordSavepoint` is the time-consuming operation in this function, we can run it first (i mean in the beginning of this function) in a goroutine in parallel with the cache update and warmIndex.

core/ledger/kvledger/txmgmt/statedb/statecouchdb/statecouchdb.go
```diff
@@ -719,54 +722,14 @@ func (vdb *VersionedDB) Close() {
...
-	select {
-	case err := <-errsChan:
-		logger.Errorf("Failed to perform full commit")
-		return errors.Wrap(err, "failed to perform full commit")
-	default:
-		logger.Debugf("All changes have been flushed to the disk")
-	}
-
+// recordSavepoint records a savepoint in the metadata db for the channel.
+func (vdb *VersionedDB) recordSavepoint(height *version.Height, namespaces []string) error {
```
cendhu: `namespaces []string` is unused and not needed.

cendhu (commented):
LGTM. A very minor comment.

core/ledger/kvledger/txmgmt/statedb/statecouchdb/statecouchdb.go
manish-sethi: Good catch. Let me fix this.

core/ledger/kvledger/txmgmt/statedb/statecouchdb/statecouchdb.go
manish-sethi: Yes, I would keep any further optimization out of scope of this. BTW, warmIndex stuff already happens in a separate goroutine.

cendhu (approved):

PR #1441, opened 2020-06-23
[FAB-17960] Ch.Part.API: Implement filerepo

caod123:
#### Type of change

- New feature

#### Description

Introduce a general purpose file repo for persisting join block data
as well as supporting CFT on join and remove operations from the
registrar and file ledger.

#### Related issues

Task: [FAB-17960](https://jira.hyperledger.org/browse/FAB-17960)

### DIFF
--- a/orderer/common/filerepo/filerepo.go
+++ b/orderer/common/filerepo/filerepo.go
@@ -1,0 +1,178 @@
+/*
+Copyright IBM Corp. All Rights Reserved.
+
+SPDX-License-Identifier: Apache-2.0
+*/
+
+package filerepo
+
+import (
+	"fmt"
+	"io/ioutil"
+	"os"
+	"path/filepath"
+	"strings"
+	"sync"
+
+	"github.com/hyperledger/fabric/internal/fileutil"
+	"github.com/pkg/errors"
+)
+
+const (
+	repoFilePermPrivateRW      os.FileMode = 0600
+	defaultTransientFileMarker             = "~"
+)
+
+// Repo manages filesystem operations for saving files marked by the fileSuffix
+// in order to support crash fault tolerance for components that need it by maintaining
+// a file repo structure storing intermediate state.
+type Repo struct {
+	mu                  sync.Mutex
+	fileRepoDir         string
+	fileSuffix          string
+	transientFileMarker string
+}
+
+// New initializes a new file repo at repoParentDir/fileSuffix.
+// All file system operations on the returned file repo are thread safe.
+func New(repoParentDir, fileSuffix string) (*Repo, error) {
+	if err := validateFileSuffix(fileSuffix); err != nil {
+		return nil, err
+	}
+
+	fileRepoDir := filepath.Join(repoParentDir, fileSuffix)
+
+	if _, err := fileutil.CreateDirIfMissing(fileRepoDir); err != nil {
+		return nil, err
+	}
+
+	if err := fileutil.SyncDir(repoParentDir); err != nil {
+		return nil, err
+	}
+
+	files, err := ioutil.ReadDir(fileRepoDir)
+	if err != nil {
+		return nil, err
+	}
+
+	// Remove existing transient files in the repo
+	transientFilePattern := "*" + fileSuffix + defaultTransientFileMarker
+	for _, f := range files {
+		isTransientFile, err := filepath.Match(transientFilePattern, f.Name())
+		if err != nil {
+			return nil, err
+		}
+		if isTransientFile {
+			if err := os.Remove(filepath.Join(fileRepoDir, f.Name())); err != nil {
+				return nil, errors.Wrapf(err, "error cleaning up transient files")
+			}
+		}
+	}
+
+	if err := fileutil.SyncDir(fileRepoDir); err != nil {
+		return nil, err
+	}
+
+	return &Repo{
+		transientFileMarker: defaultTransientFileMarker,
+		fileSuffix:          fileSuffix,
+		fileRepoDir:         fileRepoDir,
+	}, nil
+}
+
+// Save atomically persists the content to suffix/baseName+suffix file by first writing it
+// to a tmp file marked by the transientFileMarker and then moves the file to the final
+// destination indicated by the FileSuffix.
+func (r *Repo) Save(baseName string, content []byte) error {
+	r.mu.Lock()
+	defer r.mu.Unlock()
+
+	fileName := r.baseToFileName(baseName)
+	dest := r.baseToFilePath(baseName)
+
+	if _, err := os.Stat(dest); err == nil {
+		return fmt.Errorf("file already exists at %s", dest)
+	}
+
+	tmpFileName := fileName + r.transientFileMarker
+	if err := fileutil.CreateAndSyncFileAtomically(r.fileRepoDir, tmpFileName, fileName, content, repoFilePermPrivateRW); err != nil {
+		return err
+	}
+
+	return fileutil.SyncDir(r.fileRepoDir)
+}
+
+// Remove removes the file associated with baseName from the file system.
+func (r *Repo) Remove(baseName string) error {
+	r.mu.Lock()
+	defer r.mu.Unlock()
+
+	filePath := r.baseToFilePath(baseName)
+
+	if err := os.RemoveAll(filePath); err != nil {
+		return err
+	}
+
+	return fileutil.SyncDir(r.fileRepoDir)
+}
+
+// Read reads the file in the fileRepo associated with baseName's contents.
+func (r *Repo) Read(baseName string) ([]byte, error) {
+	r.mu.Lock()
+	defer r.mu.Unlock()
+
+	filePath := r.baseToFilePath(baseName)
+	return ioutil.ReadFile(filePath)
+}
+
+// List parses the directory and produce a list of file names, filtered by suffix.
+func (r *Repo) List() ([]string, error) {
+	r.mu.Lock()
+	defer r.mu.Unlock()
+
+	var repoFiles []string
+
+	files, err := ioutil.ReadDir(r.fileRepoDir)
+	if err != nil {
+		return nil, err
+	}
+
+	for _, f := range files {
+		isFileSuffix, err := filepath.Match("*"+r.fileSuffix, f.Name())
+		if err != nil {
+			return nil, err
+		}
+		if isFileSuffix {
+			repoFiles = append(repoFiles, f.Name())
+		}
+	}
+
+	return repoFiles, nil
+}
+
+// FileToBaseName strips the suffix from the file name to get the associated channel name.
+func (r *Repo) FileToBaseName(fileName string) string {
+	baseFile := filepath.Base(fileName)
+
+	return strings.TrimSuffix(baseFile, "."+r.fileSuffix)
+}
+
+func (r *Repo) baseToFilePath(baseName string) string {
+	return filepath.Join(r.fileRepoDir, r.baseToFileName(baseName))
+}
+
+func (r *Repo) baseToFileName(baseName string) string {
+	return fmt.Sprintf("%s.%s", baseName, r.fileSuffix)
+}
+
+func validateFileSuffix(fileSuffix string) error {
+	if len(fileSuffix) == 0 {
+		return errors.New("fileSuffix illegal, cannot be empty")
+	}
+
+	if strings.Contains(fileSuffix, string(os.PathSeparator)) {
+		return fmt.Errorf("fileSuffix [%s] illegal, cannot contain os path separator", fileSuffix)
+	}
+
+	return nil
+}
--- a/orderer/common/filerepo/filerepo_test.go
+++ b/orderer/common/filerepo/filerepo_test.go
@@ -1,0 +1,185 @@
+/*
+Copyright IBM Corp. All Rights Reserved.
+
+SPDX-License-Identifier: Apache-2.0
+*/
+
+package filerepo_test
+
+import (
+	"fmt"
+	"io/ioutil"
+	"os"
+	"path/filepath"
+	"testing"
+
+	"github.com/hyperledger/fabric/orderer/common/filerepo"
+	"github.com/stretchr/testify/require"
+)
+
+func TestNewFileRepo(t *testing.T) {
+	// Write a temporary file
+	tempFilePath := filepath.Join("testdata", "joinblock", "mychannel.joinblock~")
+	tempFile, err := os.Create(tempFilePath)
+	require.NoError(t, err)
+	defer func() {
+		tempFile.Close()
+		os.Remove(tempFilePath)
+	}()
+
+	// Check tempfile exists prior to creating a new file repo
+	_, err = os.Stat(tempFilePath)
+	require.NoError(t, err)
+
+	_, err = filerepo.New("testdata", "joinblock")
+	require.NoError(t, err)
+
+	// Check tempfile was cleared out upon creating a new file repo
+	require.NoFileExists(t, tempFilePath)
+}
+
+func TestNewFileRepoFailure(t *testing.T) {
+	tests := []struct {
+		testName    string
+		fileSuffix  string
+		expectedErr string
+	}{
+		{
+			testName:    "invalid fileSuffix",
+			fileSuffix:  "join/block",
+			expectedErr: "fileSuffix [join/block] illegal, cannot contain os path separator",
+		},
+		{
+			testName:    "empty fileSuffix",
+			fileSuffix:  "",
+			expectedErr: "fileSuffix illegal, cannot be empty",
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.testName, func(t *testing.T) {
+			_, err := filerepo.New("testdata", tt.fileSuffix)
+			require.EqualError(t, err, tt.expectedErr)
+		})
+	}
+}
+
+func TestFileRepo_Save(t *testing.T) {
+	tests := []struct {
+		testName string
+		content  []byte
+	}{
+		{
+			testName: "Non-empty bytes",
+			content:  []byte("block-bytes"),
+		},
+		{
+			testName: "Empty bytes",
+			content:  []byte{},
+		},
+		{
+			testName: "Nil bytes",
+			content:  nil,
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.testName, func(t *testing.T) {
+			r, err := filerepo.New("testdata", "joinblock")
+			require.NoError(t, err)
+
+			filePath := filepath.Join("testdata", "joinblock", "newchannel.joinblock")
+
+			err = r.Save("newchannel", tt.content)
+			defer os.Remove(filePath)
+			require.NoError(t, err)
+
+			// Check file bytes
+			bytes, err := ioutil.ReadFile(filePath)
+			require.NoError(t, err)
+			if tt.content != nil {
+				require.Equal(t, tt.content, bytes)
+			} else {
+				require.Equal(t, []byte{}, bytes)
+			}
+
+			// Check tempfile doesn't exist
+			require.NoFileExists(t, filePath+"~")
+		})
+	}
+}
+
+func TestFileRepo_SaveFailure(t *testing.T) {
+	r, err := filerepo.New("testdata", "joinblock")
+	require.NoError(t, err)
+
+	filePath := filepath.Join("testdata", "joinblock", "mychannel.joinblock")
+
+	err = r.Save("mychannel", []byte{})
+	require.EqualError(t, err, fmt.Sprintf("file already exists at %s", filePath))
+}
+
+func TestFileRepo_Remove(t *testing.T) {
+	// Write a temporary file
+	tempFilePath := filepath.Join("testdata", "joinblock", "channel2.joinblock")
+	tempFile, err := os.Create(tempFilePath)
+	require.NoError(t, err)
+	defer func() {
+		tempFile.Close()
+		os.Remove(tempFilePath)
+	}()
+
+	r, err := filerepo.New("testdata", "joinblock")
+	require.NoError(t, err)
+
+	err = r.Remove("channel2")
+	require.NoError(t, err)
+
+	require.NoFileExists(t, tempFilePath)
+}
+
+func TestFileRepo_Read(t *testing.T) {
+	t.Run("Successful read, non-empty bytes", func(t *testing.T) {
+		r, err := filerepo.New("testdata", "joinblock")
+		require.NoError(t, err)
+
+		bytes, err := r.Read("mychannel")
+		require.NoError(t, err)
+		require.Equal(t, "dummy-data", string(bytes))
+	})
+
+	t.Run("Successful read, empty bytes", func(t *testing.T) {
+		r, err := filerepo.New("testdata", "remove")
+		require.NoError(t, err)
+
+		bytes, err := r.Read("mychannel")
+		require.NoError(t, err)
+		require.Equal(t, []byte{}, bytes)
+	})
+
+	t.Run("Failed read, invalid file", func(t *testing.T) {
+		r, err := filerepo.New("testdata", "joinblock")
+		require.NoError(t, err)
+
+		_, err = r.Read("invalidfile")
+		require.EqualError(t, err, "open testdata/joinblock/invalidfile.joinblock: no such file or directory")
+	})
+}
+
+func TestFileRepo_List(t *testing.T) {
+	r, err := filerepo.New("testdata", "joinblock")
+	require.NoError(t, err)
+
+	files, err := r.List()
+	require.NoError(t, err)
+	require.Equal(t, []string{"mychannel.joinblock"}, files)
+}
+
+func TestFileRepo_FileToBaseName(t *testing.T) {
+	r, err := filerepo.New("testdata", "joinblock")
+	require.NoError(t, err)
+
+	filePath := filepath.Join("testdata", "joinblock", "mychannel.joinblock")
+	channelName := r.FileToBaseName(filePath)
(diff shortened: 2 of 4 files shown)


orderer/common/filerepo/filerepo.go
```diff
@@ -0,0 +1,95 @@
...
+	TransientFilePrefix string
+	FileRepoDir         string
+}
+
+// Save atomically persists the content to the channelName+suffix file by first writing it
+// to a tmp file prefixed by the TransientFilePrefix and then moves the file to the final
+// destination indicated by the FileSuffix.
+func (r *Repo) Save(channelName string, content []byte) error {
+	fileRepoDir := r.Options.FileRepoDir
+	dest := filepath.Join(fileRepoDir, channelName+r.Options.FileSuffix)
```
yacovm: is it possible to have the channel name a string that contains the OS file separator? If yes then maybe we need to ensure it doesn't contain such?

caod123: I remember there are some fairly strict channel name rules, but I don't remember what characters are allowed in a channel name. I would imagine there should be some validation at the API level since you'd need to pass the channel name in via the join endpoint anyways. I do agree though that there probably could be some additional checks around the channel name.

caod123: From https://hyperledger-fabric.readthedocs.io/en/release-2.0/build_network.html#create-join-channel: 
```
Channel names must be all lower case, less than 250 characters long and match the regular expression [a-z][a-z0-9.-]*
```

yacovm: the code above which i missed, does just that.

orderer/common/filerepo/filerepo_test.go
```diff
@@ -0,0 +1,116 @@
...
+		TransientFilePrefix: "~",
+		FileRepoDir:         "testdata",
+	}
+
+	r := filerepo.Repo{
+		Options: options,
+	}
+
+	t.Run("Successful load", func(t *testing.T) {
+		bytes, err := r.Load("testdata/mychannel.joinblock")
```
cendhu: for a common `filerepo` pkg, do we really need two 17 KB files in the testdata? Is it not possible to use a simple file with a known data so that we can compare the content easily? Please correct me if I am missing something.

orderer/common/filerepo/filerepo_test.go
```diff
@@ -0,0 +1,116 @@
...
+
+	t.Run("Successful save", func(t *testing.T) {
+		err := r.Save("newchannel", []byte("block-bytes"))
+		defer os.Remove("testdata/newchannel.joinblock")
+		assert.NoError(t, err)
+
+		// Check file bytes
+		bytes, err := ioutil.ReadFile("testdata/newchannel.joinblock")
+		require.NoError(t, err)
+		assert.Equal(t, bytes, []byte("block-bytes"))
```
cendhu: why are we using a mix of `require` and `assert`? can't we use `require` consistently?

orderer/common/filerepo/filerepo.go
```diff
@@ -0,0 +1,102 @@
...
+
+	if err := tmpFile.Close(); err != nil {
+		return err
+	}
+
+	return os.Rename(tmpFile.Name(), dest)
+}
+
+// Scan parses the directory and produce a list of file names, filtered by suffix.
+func (r *Repo) Scan() ([]string, error) {
```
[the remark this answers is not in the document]
cendhu: Nit. `List()` may be more expressive.

orderer/common/filerepo/filerepo.go
```diff
@@ -0,0 +1,102 @@
...
+		if isFileSuffix {
+			repoFiles = append(repoFiles, f.Name())
+		}
+	}
+
+	return repoFiles, nil
+}
+
+// Load reads the file contents.
+func (r *Repo) Load(fileName string) ([]byte, error) {
```
cendhu: while the `Save()` receives a channelName, shouldn't load also receive the channel name? Maybe I am asking this because I don't have clarity on how the consumer would use it.

orderer/common/filerepo/filerepo_test.go
```diff
@@ -0,0 +1,116 @@
...
+		TransientFilePrefix: "~",
+		FileRepoDir:         "testdata",
+	}
+
+	r := filerepo.Repo{
+		Options: options,
+	}
+
+	channelName := r.FileToChannelName("testdata/mychannel.joinblock")
+	assert.Equal(t, channelName, "mychannel")
```
cendhu: Shouldn't the order of argument be `expected` and then `actual` value?

orderer/common/filerepo/filerepo_test.go
```diff
@@ -0,0 +1,116 @@
...
+	}
+
+	r := filerepo.Repo{
+		Options: options,
+	}
+
+	t.Run("Successful scan", func(t *testing.T) {
+		files, err := r.Scan()
+		assert.NoError(t, err)
+		assert.Equal(t, files, []string{"mychannel.joinblock"})
```
cendhu: Whenever testing a function with a loop, I would always prefer to have at least two iterations.

orderer/common/filerepo/filerepo.go
```diff
@@ -0,0 +1,102 @@
...
+	"bufio"
+	"io/ioutil"
+	"os"
+	"path/filepath"
+	"strings"
+
+	"github.com/hyperledger/fabric/common/configtx"
+)
+
+type Repo struct {
```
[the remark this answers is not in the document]
caod123: Yes we've deferred Remove for now until it is necessary. Although given that once an orderer has fully onboarded, the join block is no longer needed and can be safely removed so I can probably include the `Remove` on this PR as well.

orderer/common/filerepo/filerepo.go
```diff
@@ -0,0 +1,102 @@
...
+	Options Options
+}
+
+type Options struct {
+	FileSuffix          string
```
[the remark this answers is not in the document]
caod123: the file suffix is the extension or suffix for the files we are saving in the file repo. For joinblocks this will likely be something like `.joinblock` so all joinblocks are saved as `<filerepo dir>/<channel>.joinblock`

orderer/common/filerepo/filerepo.go
```diff
@@ -0,0 +1,102 @@
...
+	TransientFilePrefix string
```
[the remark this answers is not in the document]
caod123: There may be some use case for configuring the prefix, but currently it will probably always be `~` so we can probably remove it entirely from the options and leave it as an internal package constant.

orderer/common/filerepo/filerepo_test.go
caod123: Probably not, I wanted to test with a proper block but because we don't care about block validation at the file repo level you're probably right and these can just be simple files with dummy data.

orderer/common/filerepo/filerepo_test.go
caod123: I saw the slew of PRs recently for standardizing the ledger level tests to use `require` over `assert`. While inherently this doesn't introduce any problems, I feel mildly apprehensive that we may be overusing `require` when `assert` might be the proper usage. I think this may come down to what the intended intentions and proper usages of the `assert` and `require` packages were for initially. I'm fine changing all of these tests to using `require` exclusively though since it doesn't affect anything.

orderer/common/filerepo/filerepo.go
caod123: I believe the idea for the consumer would be to scan/list the joinblocks in the file repo and then load as needed the filenames returned from the `Scan/List` function. I thought about this initially but after thinking about it more I'm not sure if we should prefer taking a channel name or file name inherently. The `filerepo` could ideally be used for more general purpose applications outside of channel participation related usages; however taking a file name in for the `Save` function would require changing some of the logic around the temp file path. Maybe for now for standardization purposes we can use the `filerepo` exclusively for channel participation and modify `Load` to receive the channel name as well. Alternatively we could leave it as is for now and refactor it on the fly as we come up with a better picture for how the consumer will use these functions.

orderer/common/filerepo/filerepo_test.go
cendhu: `require package is very similar to assert, except for it terminates test if assertion is not successful.` https://goinbigdata.com/testing-go-code-with-testify/ assert was introduced first and then require. 

@btl5037 also mentioned that a similar change happened at the SDK side a few days/weeks back. I do not have a strong opinion on the orderer pkg but my point was that this test uses a mix of both `require` and `assert` which look odd to me.

orderer/common/filerepo/filerepo_test.go
caod123: I'm so used to gomega matchers which is the opposite 😅 having the actual come first.

orderer/common/filerepo/filerepo_test.go
caod123: I believe it already iterates twice, once to add `mychannel.joinblock` and the other to filter out `mychannel.block` which doesn't match the file suffix pattern.

orderer/common/filerepo/filerepo.go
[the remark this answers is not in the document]
caod123: I don't think these functions would ever be called concurrently as saving occurs at the rest API level on receiving a join channel request; scan and load are called in order on orderer startup/registrar initialization to trigger onboarding for each join block.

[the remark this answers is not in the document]
cendhu: I see. I thought of this package as a generic file repo that has no assumptions about how the caller would use it. With your explanation, my assumption no longer holds and I don't know whether non-generic pkgs can reside in the common.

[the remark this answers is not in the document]
yacovm: If the API doesn't state it is thread safe the assumption of the user should be that it's not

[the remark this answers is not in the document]
caod123: Hmm good point. I think we may need to take a step back and evaluate whether this should be designed around general purpose usage or specifically for channel participation right now as a lot of the discussions below seem to hinge on this nuance. There's always the option of keeping it as specific to channel participation for now and in the future if we ever need to make it more general purpose, some small changes here and there could be included at that time. It's easy to say this could be used for other purposes, but right now I'm blanking on thinking of exactly what else we could use this for.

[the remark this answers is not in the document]
cendhu: Adding a lock and making it thread-safe is very easy to do it for this pkg. Correct me If I am missing any challenges. That's why I suggested and I believe that making these kinds of packages generic would help in a bug-free code (in the future). Just take it as my philosophical view and I would better leave it to the orderer squad to choose what is best. At least in the ledger, it helps us and we don't need to check everytime whether the call we make is thread safe or not :-)

orderer/common/filerepo/filerepo.go
```diff
@@ -0,0 +1,102 @@
...
+
+	if err := w.Flush(); err != nil {
+		return err
+	}
+
+	if err := tmpFile.Close(); err != nil {
+		return err
+	}
+
+	return os.Rename(tmpFile.Name(), dest)
```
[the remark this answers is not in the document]
cendhu: Wait for this pkg to be merged -- https://github.com/hyperledger/fabric/pull/1457 It has the required utility function.

orderer/common/filerepo/filerepo.go
```diff
@@ -0,0 +1,102 @@
...
+	FileSuffix          string
+	TransientFilePrefix string
+	FileRepoDir         string
+}
+
+// Save atomically persists the content to the channelName+suffix file by first writing it
+// to a tmp file prefixed by the TransientFilePrefix and then moves the file to the final
+// destination indicated by the FileSuffix.
+func (r *Repo) Save(channelName string, content []byte) error {
+	err := configtx.ValidateChannelID(channelName)
```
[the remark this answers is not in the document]
caod123: I figured this was the case. Will remove thanks!

orderer/common/filerepo/filerepo.go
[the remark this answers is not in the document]
caod123: I was originally thinking of having separate repo dir's defined in the orderer.yaml, but your first suggestion produces less configuration which may be simpler.

orderer/common/filerepo/filerepo.go
```diff
@@ -0,0 +1,102 @@
...
+
+	tmpFile, err := ioutil.TempFile(fileRepoDir, r.Options.TransientFilePrefix+channelName)
+	if err != nil {
+		return err
+	}
+
+	defer func() {
+		if err != nil {
+			tmpFile.Close()
+			os.Remove(tmpFile.Name())
```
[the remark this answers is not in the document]
caod123: This is best effort mainly. As Yoav mentioned, we can clear out temp files in another place as well.

caod123:
Pushed an initial update addressing the comments from above. Will update once #1457 is merged to use the fileutil functions for some of the file system operations.

orderer/common/filerepo/filerepo.go
```diff
@@ -0,0 +1,198 @@
...
+	// Remove existing transient files in the repo
+	for _, f := range files {
+		isTransientFile, err := filepath.Match("*"+repo.transientFileMarker, f.Name())
+		if err != nil {
+			return nil, err
+		}
+		if isTransientFile {
+			os.Remove(filepath.Join(repo.fileRepoDir, f.Name()))
+		}
+	}
```
tock-ibm: Either validate that the fileSuffix does not end with the transientFileMarker, or match against suffix+marker, rather than just the marker... otherwise this is going to remove legitimate files.

orderer/common/filerepo/filerepo.go
```diff
@@ -0,0 +1,199 @@
...
+	mu sync.Mutex
+}
+
+type repoOption func(r *Repo)
+
+func WithTransientFileMarker(marker string) repoOption {
+	return func(r *Repo) {
+		r.transientFileMarker = marker
+	}
+}
```
cendhu: Function pointers are not good for readability IMO. It is good when we want to add hooks (heavily used in linux kernel). I yet to read the updated code again since I reviewed a few days back. Is there no way to remove this?

caod123: I could change the `New` function to just take an optional variatic `transientFileMarker` list instead and read the first element in that list if it exists to override the default `transientFileMarker`. I was mainly trying to avoid just accepting a `transientFileMarker` argument and ignoring it if it was an empty string since the default would probably be to not pass an override.

orderer/common/filerepo/filerepo.go
```diff
@@ -0,0 +1,199 @@
...
+	fileRepoDir         string
+	fileSuffix          string
+	transientFileMarker string
+
+	mu sync.Mutex
+}
+
+type repoOption func(r *Repo)
+
+func WithTransientFileMarker(marker string) repoOption {
```
sykesm: Why do we keep doing this? An exported function should accept and return exported types.

Please export `RepoOption`.

caod123: I had initially left it unexported to leave defining functions that returned repoOptions to only be definable within this package. I guess that might have been too restrictive or unnecessary.

orderer/common/filerepo/filerepo.go
sykesm: This is using the OptionFunc pattern for extensible construction. I'm not sure why that's considered difficult to read. It's a flexible, pragmatic, and relatively common way of handling object construction in a friendly, backwards compatible way.

https://dave.cheney.net/2014/10/17/functional-options-for-friendly-apis

The same pattern can be used in place of the "builder" pattern to minimize error handling to one call while building up a bunch of changes.

orderer/common/filerepo/filerepo.go
```diff
@@ -0,0 +1,199 @@
...
+	repoFilePermPrivateRW os.FileMode = 0600
+	repoDirPermPrivateRWX os.FileMode = 0700
+)
+
+type Repo struct {
+	fileRepoDir         string
+	fileSuffix          string
+	transientFileMarker string
+
+	mu sync.Mutex
```
sykesm: nit: grouping the lock with the fields it protects is a useful convention. We generally expect to see the lock declared and the fields is protects just below it.

orderer/common/filerepo/filerepo.go
sykesm: May also want to consider returning an error from the function since the constructor returns an error.

sykesm: > I had initially left it unexported to leave defining functions that returned repoOptions to only be definable within this package. I guess that might have been too restrictive or unnecessary.

It's unnecessary. The reason you use options is so the implementation can access the internal details of the thing that's being constructed. The implementation of the function acting on `*Repo` must live in this package to do anything meaningful.

If you don't export the type, it's an inconsistent. It's no different than having a public constructor that returns an unexported implementation.

orderer/common/filerepo/filerepo.go
cendhu: Whenever there is a function pointer and it's usage, I always had to spend some extra time to understand the code. If there is a simple way like `NewServer(config)` or `NewServer(config1, confg2)`, I spend less time to read and understand. This is what I meant regarding the readability. If this is a common pattern and used widely, then I might probably improve my readability skill once I find and read many such patterns.

As I mentioned in my first set of review, I do not see a strong reason to configure the `transientFileMarker`. It is just an internal logic to the `filerepo`. Given that, I am not convinced that we need a function pointer or configurable option.

caod123:
Updated to use `fileutil` package

orderer/common/filerepo/filerepo.go
```diff
@@ -0,0 +1,194 @@
...
+func New(repoParentDir, fileSuffix string, opts ...RepoOption) (*Repo, error) {
+	// default transientFileMarker to ~, can modify via opts
+	repo := &Repo{
+		transientFileMarker: "~",
+		fileSuffix:          fileSuffix,
+		fileRepoDir:         filepath.Join(repoParentDir, fileSuffix),
+	}
+
+	for _, opt := range opts {
+		if err := opt(repo); err != nil {
```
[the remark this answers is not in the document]
caod123: I'll remove the configurability for now as I don't see an immediate need to override the default value. We can always revisit this down the line if an override is needed.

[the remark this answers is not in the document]
sykesm: In this instance, fine. In general, this is the wrong way.

When it comes to constructing objects, we have:

1. Zero values
2. Struct literals that require nothing more than wiring
3. Constructors that take no arguments
4. Constructors that take specific arguments
5. Constructors that can evolve

Option 1 is best. It's the "make the zero value useful" goal of of the standard library.

Option 2 is good when evolving from option 1. This allows us to create a default behavior with the zero value and add customizability after the fact by injecting something to override behavior. When we set a field, we have an explicit name at the call site so we know what we're providing. We see a lot of this in the http package.

Option 3 is okay but limiting. It means that we're constructing something and there's no way to control it. That implies we end up having to modify the state after construction to make things useful or build global state around it to influence it.

Option 4 is rigid and confusing. The rigidity comes from the fact that there's never a backward compatible way to add or remove constructor arguments without impacting callers. It's confusing when we get long lists of arguments that share the same type. In this particular case, it looks like we have `n` strings. So, at the call site, I need to figure out the order and what they mean.

Option 5 is good. In this model we can easy extend the functionality of constructors. Each option is named so when we make the call, we know what we're providing. The options allow better encapsulation of the object as even the names of the fields we use to hold dependencies is not exposed to the caller.

Given how prevalent the option func pattern is in libraries, we should all be able to recognize it for what it is and embrace the benefits it provides when it comes to evolving a code base.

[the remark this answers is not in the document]
cendhu: Thanks, @caod123 for removing this configuration option.

@sykesm Thanks for the nice summary. Especially what you mentioned in option 4 and 5 are nice -- no need to break backward compatibility when we add new configuration option -- didn't strike me. While optionalal func pattern might add a bit of overhead while reading it for the first time (as in my case), we will adjust to it as we see more patterns. I hope you don't mind that I am copying your suggestions to the [google doc](https://docs.google.com/document/d/1mU4S4U1dDpBk9AA180RMRB7g3OH887nqbcEPGD3px3Q/edit?usp=sharing) that I maintain so that these suggestions/recommendations are not lost within the PR comments.

caod123:
@cendhu could you do another pass of this and let me know if it looks good to you? There hasn't been much activity on this since last week.

orderer/common/filerepo/filerepo.go
```diff
@@ -0,0 +1,169 @@
...
+	return strings.TrimSuffix(baseFile, "."+r.fileSuffix)
+}
+
+func validateFileSuffix(fileSuffix string) error {
+	if len(fileSuffix) <= 0 {
+		return errors.New("fileSuffix illegal, cannot be empty")
+	}
+
+	if strings.Contains(fileSuffix, string(os.PathSeparator)) {
+		return errors.Errorf("fileSuffix [%s] illegal, cannot contain os path separator", fileSuffix)
```
cendhu: Nit.`fmt.Errorf`and `errors` pkg instead of `github.com/pkg/errors`

orderer/common/filerepo/filerepo_test.go
```diff
@@ -0,0 +1,168 @@
...
+	// Check tempfile exists prior to creating a new file repo
+	_, err = os.Stat(tempFilePath)
+	require.NoError(t, err)
+
+	_, err = filerepo.New("testdata", "joinblock")
+	require.NoError(t, err)
+
+	// Check tempfile was cleared out upon creating a new file repo
+	_, err = os.Stat(tempFilePath)
+	require.True(t, os.IsNotExist(err))
```
cendhu: Nit. `require.NoFileExists` can be used here.

orderer/common/filerepo/filerepo_test.go
```diff
@@ -0,0 +1,168 @@
...
+			require.NoError(t, err)
+			if tt.content != nil {
+				require.Equal(t, tt.content, bytes)
+			} else {
+				require.Equal(t, []byte{}, bytes)
+			}
+
+			// Check tempfile doesn't exist
+			_, err = os.Stat(filePath + "~")
+			require.True(t, os.IsNotExist(err))
```
cendhu: Nit. `require.NoFileExists` can be used here.

orderer/common/filerepo/filerepo.go
```diff
@@ -0,0 +1,169 @@
...
+	}
+
+	for _, f := range files {
+		isFileSuffix, err := filepath.Match("*"+r.fileSuffix, f.Name())
+		if err != nil {
+			return nil, err
+		}
+		if isFileSuffix {
+			repoFiles = append(repoFiles, f.Name())
+		}
```
cendhu, line 147: Is this necessary? Given that methods on `Repo` are thread-safe and `New()` removes all transientFiles, do we need to perform the match based on fileSuffix. If it is necessary, please update the test to reflect this. Otherwise, remove this part of the code.

cendhu (commented):
Danny, Most of my comments are nit and hence, I leave these to you to decide. 

Once you add the missing test, I think the PR can be merged.

orderer/common/filerepo/filerepo.go
```diff
@@ -0,0 +1,169 @@
...
+
+// New initializes a new file repo at repoParentDir/fileSuffix.
+// All file system operations on the returned file repo are thread safe.
+func New(repoParentDir, fileSuffix string) (*Repo, error) {
+	// Check that fileSuffix doesn't contain the os path separator
+	if err := validateFileSuffix(fileSuffix); err != nil {
+		return nil, err
+	}
+
+	fileRepoDir := filepath.Join(repoParentDir, fileSuffix)
```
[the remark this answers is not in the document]
caod123: Having separate subfolders helps to isolate different file repos from operating on the same folder. The suffix is still useful for filtering in case other files end up in the folder through other means.

orderer/common/filerepo/filerepo.go
caod123: It's still useful to ensure other files are filtered out if they somehow end up there. The tests already reflect this by filtering out `testdata/joinblock/mychannel.block` and only returning `testdata/joinblock/mychannel.joinblock`

orderer/common/filerepo/filerepo.go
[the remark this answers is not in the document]
cendhu: > Having separate subfolders helps to isolate different file repos from operating on the same folder. The suffix is still useful for filtering in case other files end up in the folder through other means.

What other means? You mean admin creates some files manually?

orderer/common/filerepo/filerepo.go
cendhu: Probably, I missed that. If test is there, then it is fine. 

> if they somehow end up there

I do not understand. Are you referring to some attack scenario?

cendhu (approved):
Thanks, Danny. I am still not sure about the following two points 

1. having a fileSuffix as both folder name and the file suffix
2. having a filter logic in the List()

If required, it can be fixed in a separate PR after the discussion and need not to block this PR. 

In future, please try to not rebase in the middle of reviews. Otherwise, it is quite hard to find out what changed compared to the initial commits.

orderer/common/filerepo/filerepo.go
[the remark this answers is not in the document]
caod123: Sure, I don't see it really happening but in the off chance the admin may backup a joinblock to the same directory or for some other reason accidentally move a separate file to the folder. In the end it doesn't really hurt to have the suffix in addition to the subfolder even if it introduces some minor redundancy. As for the subfolder existence itself, Yoav pointed out in a previous comment that having separate subfolders would speed up the scan and avoid filesystem sync issues if we use two file-repo's in the same process with different suffixes

PR #1442, opened 2020-06-23
[FAB-18021] Improve log messages in `msp/configbuilder.go`

satota2:
This patch improves log messages in `loadCertificateAt()` in `msp/configbuilder.go` from `warn` to `debug` in case `Certificate` for each `OUIdentifier` is not set in `msp/config.yaml`.

#### Type of change

#### Description

This patch improves log messages in `loadCertificateAt()` in `msp/configbuilder.go` from `warn` to `debug` in case `Certificate` for each `OUIdentifier` is not set in `msp/config.yaml`.

The reason of improvement is as follows:
- According to the document, setting `Certificate` is optional and it is typical NOT to configure a specific certificate. Despite the typical setting, it may be noisy for users to always get the WARN Level log message.
- It is desirable to be able to distinguish between the case where `Certificate` is not set and the case where the file is not found in the specific path.

#### Related issues

- https://jira.hyperledger.org/browse/FAB-18021

### DIFF
--- a/msp/configbuilder.go
+++ b/msp/configbuilder.go
@@ -363,6 +363,12 @@ func getMspConfig(dir string, ID string, sigid *msp.SigningIdentityInfo) (*msp.M
 }
 
 func loadCertificateAt(dir, certificatePath string, ouType string) []byte {
+
+	if certificatePath == "" {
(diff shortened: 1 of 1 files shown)


ale-linux (approved):
Thanks for your contribution

PR #1460, opened 2020-06-25
bootstrap collection config history store from snapshot

cendhu:
#### Type of change

- New feature

#### Description

This PR adds an `ImportXXX` method to the config history manager to allow the bootstrapping of collection config history for a given ledgerID.

#### Additional details

`Remove()` method would be added in a subsequent PR and will be used during failure and recovery.

#### Related issues

https://jira.hyperledger.org/browse/FAB-18003

### DIFF
--- a/core/ledger/confighistory/mgr_test.go
+++ b/core/ledger/confighistory/mgr_test.go
@@ -13,7 +13,7 @@ import (
 	"io/ioutil"
 	"math"
 	"os"
-	"path"
+	"path/filepath"
 	"testing"
 
 	"github.com/golang/protobuf/proto"
@@ -268,7 +268,6 @@ func TestWithImplicitColls(t *testing.T) {
 
 type testEnvForSnapshot struct {
 	mgr             *Mgr
-	retriever       *Retriever
 	testSnapshotDir string
 	cleanup         func()
 }
@@ -276,13 +275,11 @@ type testEnvForSnapshot struct {
 func newTestEnvForSnapshot(t *testing.T) *testEnvForSnapshot {
 	dbPath, err := ioutil.TempDir("", "confighistory")
 	require.NoError(t, err)
-	p, err := newDBProvider(dbPath)
+	mgr, err := NewMgr(dbPath, &mock.DeployedChaincodeInfoProvider{})
 	if err != nil {
 		os.RemoveAll(dbPath)
-		t.Fatalf("Failed to create new leveldb provider: %s", err)
+		t.Fatalf("Failed to create new config history manager: %s", err)
 	}
-	mgr := &Mgr{dbProvider: p}
-	retriever := mgr.GetRetriever("ledger1", nil)
 
 	testSnapshotDir, err := ioutil.TempDir("", "confighistorysnapshot")
 	if err != nil {
@@ -291,7 +288,6 @@ func newTestEnvForSnapshot(t *testing.T) *testEnvForSnapshot {
 	}
 	return &testEnvForSnapshot{
 		mgr:             mgr,
-		retriever:       retriever,
 		testSnapshotDir: testSnapshotDir,
 		cleanup: func() {
 			os.RemoveAll(dbPath)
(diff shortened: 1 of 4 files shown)


core/ledger/confighistory/mgr_test.go
```diff
@@ -300,100 +296,210 @@ func newTestEnvForSnapshot(t *testing.T) *testEnvForSnapshot {
...
 	}
-	verifyExportedConfigHistory(t, env.testSnapshotDir, fileHashes, expectedCollectionConfigs)
+
+	cleanup := func() {
+		require.NoError(t, os.RemoveAll(filepath.Join(env.testSnapshotDir, snapshotDataFileName)))
+		require.NoError(t, os.RemoveAll(filepath.Join(env.testSnapshotDir, snapshotMetadataFileName)))
+	}
+
+	t.Run("confighistory is empty", func(t *testing.T) {
+		retriever := env.mgr.GetRetriever("ledger1", nil)
```
manish-sethi: Not sure if this could cause a flake... can you change "ledger1" to "ledger0". As, the subtests are expected to be independent, better to make sure that even some reordering does not affect these.

core/ledger/confighistory/mgr_test.go
```diff
@@ -300,100 +296,210 @@ func newTestEnvForSnapshot(t *testing.T) *testEnvForSnapshot {
...
+		verifyExportedConfigHistory(t, env.testSnapshotDir, fileHashes, storedKVs)
+		cleanup()
+	})
+
+	t.Run("import confighistory error cases", func(t *testing.T) {
+		setup("ledger6")
+		retriever := env.mgr.GetRetriever("ledger6", nil)
+		_, err := retriever.ExportConfigHistory(env.testSnapshotDir, testNewHashFunc)
+		require.NoError(t, err)
+
```
manish-sethi, line 460: Not sure what is the significance of these lines in this test

core/ledger/confighistory/mgr_test.go
```diff
@@ -300,100 +296,210 @@ func newTestEnvForSnapshot(t *testing.T) *testEnvForSnapshot {
...
+		dataFileWriter, err = snapshot.CreateFile(filepath.Join(env.testSnapshotDir, snapshotMetadataFileName), snapshotFileFormat, testNewHashFunc)
+		require.NoError(t, err)
+		defer dataFileWriter.Close()
+		require.NoError(t, dataFileWriter.EncodeBytes([]byte("junk")))
+		err = env.mgr.ImportConfigHistory("ledger8", env.testSnapshotDir)
+		require.Contains(t, err.Error(), "error while reading from the snapshot file")
+		require.Contains(t, err.Error(), "confighistory.metadata: EOF")
+		cleanup()
+	})
```
manish-sethi, line 492: Can you add to this list one more test when the db returns an error while importing

manish-sethi (changes requested):
Overall LGTM. A couple minor comments related to tests

core/ledger/confighistory/mgr_test.go
manish-sethi: Also, in the similar lines, better to call `cleanup()` in `defer` as opposed to at the end. Else, one failed could cascade to others and the purpose of subtests (as opposed to a single monolithic) is defeated.

cendhu: I will make both changes. It wouldn't cause a flake because `t.Run()` is executed serially one after the other unless we specify `t.Parallel()` inside the `func()` present in the `t.Run()`.

core/ledger/confighistory/mgr_test.go
cendhu: I don't understand. I am exporting ledger 6 so that I don't need to create some random snapshot file which would otherwise give me `error reading snapshot file.... EOF`

core/ledger/confighistory/mgr_test.go
cendhu: I added it initially to check the db writebatch. Once I added error for the `iter.Next()`, I couldn't test the writebatch error. I, in fact, made the `maxBatchSize` as var instead of constant so that the test could covered both writebatches (inside and outside the loop). I will add a test for the db error (but it would return at `isEmpty()` check itself)

core/ledger/confighistory/mgr_test.go
manish-sethi: Understood. I found it confusing because of following two reasons
- This was not needed for the immidiate following test.
- As you follow through the test code, one has to keep track of how the content in the file is getting changed. 

I think that these error tests deserve to be separate subtests but I leave that to you. Otherwise, if you can produce the error causing condition explicitly (like you did at line 485) each time a fresh, that may be more consumable. In a related [test](https://github.com/hyperledger/fabric/blob/327765d59a9dd095e0282eda18c8e6306594daf1/common/ledger/blkstorage/snapshot_test.go#L342), I found this pattern simpler to maintain. But as I said, it's not a blocker for me and leave to you.

core/ledger/confighistory/mgr_test.go
manish-sethi: Not sure what does `t.Parallel()` has to do with my comment. What I mentioned is that I am not certain if the ordering of the subtests is guaranteed? If yes, then what order is guaranteed - in the order they appear in the file or in some other order (like sort order on the name of the subtest). Even if this is guaranteed, it would be fragile such that if you swap the subtest order in the code, it would break the tests

core/ledger/confighistory/mgr_test.go
manish-sethi: I understand that it's hard to skip isEmpty but whatever can be added to test, is good.

manish-sethi (approved):
@cendhu - with the latest update, only one minor comment is pending but I am good to merge this and if you like to care about the pending comment, you can address later.

core/ledger/confighistory/mgr_test.go
cendhu: Agree on independence and rearrangement. My response was mainly related to the flakes issue that you raised. The `t.Run()` is blocked call and it runs in the same order it occurs in the main test. Only If we put `t.Parallel()`, the execution would be made unblocked and sub-tests can interleave. This is what I wanted to convey to mention that it wouldn't result in a flake.

core/ledger/confighistory/mgr_test.go
cendhu: I understood your confusion now. When I add `Remove()` function, I will refactor this sub-test too.

PR #1467, opened 2020-06-26
Revert "rm old proto related code from transient store"

cendhu:
Reverts hyperledger/fabric#1464

After discussing with @manish-sethi , we found that there is a rare-case scenario for which we needs to maintain the old proto still. 

For e.g., say that the peer has been upgraded from v1.1 to v1.2. The entries in transient store that used old proto can stay for 1000 more blocks in the worst case (i.e., the default `maxBlockRetained`). If the peer is continued to be upgraded immediately to v1.2, v1.3, v1.4, .... till v2.2, we still needs to maintain the old proto or migrate the old proto to new one (as suggested by @manish-sethi). 

Hence, we are reverting this PR.

### DIFF
--- a/core/transientstore/store.go
+++ b/core/transientstore/store.go
@@ -10,6 +10,7 @@ import (
 	"errors"
 
 	"github.com/golang/protobuf/proto"
+	"github.com/hyperledger/fabric-protos-go/ledger/rwset"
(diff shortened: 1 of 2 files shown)


manish-sethi (approved):
Thanks @cendhu

PR #1471, opened 2020-06-26
[FAB-17129] Configure peer and orderer to use PKCS#11 as BCCSP in int test

stephyee:
#### Type of change

Test update

#### Description

Configures and peer and orderer using PKCS#11 as BCCSP and creates and joins a channel 

#### Related issues

[FAB-17129](https://jira.hyperledger.org/browse/FAB-17129)

### DIFF
--- a/integration/pkcs11/pkcs11_suite_test.go
+++ b/integration/pkcs11/pkcs11_suite_test.go
@@ -1,0 +1,54 @@
+/*
+Copyright IBM Corp. All Rights Reserved.
+
+SPDX-License-Identifier: Apache-2.0
+*/
+
+package pkcs11
+
+import (
+	"encoding/json"
+	"testing"
+
+	bpkcs11 "github.com/hyperledger/fabric/bccsp/pkcs11"
+	"github.com/hyperledger/fabric/integration"
+	"github.com/hyperledger/fabric/integration/nwo"
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+)
+
+func TestPKCS11(t *testing.T) {
+	RegisterFailHandler(Fail)
+	lib, pin, label := bpkcs11.FindPKCS11Lib()
+	if lib == "" || pin == "" || label == "" {
+		t.Skip("Skipping PKCS11 Suite: Required ENV variables not set")
+	}
+	RunSpecs(t, "PKCS11 Suite")
+}
+
+var (
+	buildServer *nwo.BuildServer
+	components  *nwo.Components
+)
+
+var _ = SynchronizedBeforeSuite(func() []byte {
+	buildServer = nwo.NewBuildServer("-tags=pkcs11")
+	buildServer.Serve()
+
+	components = buildServer.Components()
+	payload, err := json.Marshal(components)
+	Expect(err).NotTo(HaveOccurred())
+	return payload
+}, func(payload []byte) {
+	err := json.Unmarshal(payload, &components)
+	Expect(err).NotTo(HaveOccurred())
+})
+
+var _ = SynchronizedAfterSuite(func() {
+}, func() {
+	buildServer.Shutdown()
+})
+
+func StartPort() int {
+	return integration.PKCS11Port.StartPortForNode()
+}
--- a/integration/pkcs11/pkcs11_test.go
+++ b/integration/pkcs11/pkcs11_test.go
@@ -1,0 +1,391 @@
+/*
+Copyright IBM Corp. All Rights Reserved.
+
+SPDX-License-Identifier: Apache-2.0
+*/
+package pkcs11
+
+import (
+	"crypto/ecdsa"
+	"crypto/elliptic"
+	"crypto/rand"
+	"crypto/sha256"
+	"crypto/x509"
+	"crypto/x509/pkix"
+	"encoding/asn1"
+	"encoding/hex"
+	"encoding/pem"
+	"fmt"
+	"io/ioutil"
+	"math/big"
+	"os"
+	"path/filepath"
+	"syscall"
+	"time"
+
+	docker "github.com/fsouza/go-dockerclient"
+	bpkcs11 "github.com/hyperledger/fabric/bccsp/pkcs11"
+	"github.com/hyperledger/fabric/integration/nwo"
+	"github.com/hyperledger/fabric/integration/nwo/fabricconfig"
+	"github.com/miekg/pkcs11"
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+	"github.com/tedsuo/ifrit"
+)
+
+var _ = Describe("Configures PKCS#11 for peer and orderer", func() {
+	var (
+		tempDir string
+		network *nwo.Network
+		client  *docker.Client
+		process ifrit.Process
+		orderer *nwo.Orderer
+	)
+
+	BeforeEach(func() {
+		var err error
+		tempDir, err = ioutil.TempDir("", "p11")
+		Expect(err).NotTo(HaveOccurred())
+
+		client, err = docker.NewClientFromEnv()
+		Expect(err).NotTo(HaveOccurred())
+
+		network = nwo.New(nwo.BasicSolo(), tempDir, client, StartPort(), components)
+		network.GenerateConfigTree()
+
+		orderer = network.Orderer("orderer")
+
+		lib, pin, label := bpkcs11.FindPKCS11Lib()
+		myBCCSP := &fabricconfig.BCCSP{
+			Default: "PKCS11",
+			PKCS11: &fabricconfig.PKCS11{
+				Security: 256,
+				Hash:     "SHA2",
+				Pin:      pin,
+				Label:    label,
+				Library:  lib,
+			},
+		}
+
+		By("Updating bccsp peer config")
+		for _, peer := range network.Peers {
+			peerConfig := network.ReadPeerConfig(peer)
+			peerConfig.Peer.BCCSP = myBCCSP
+			network.WritePeerConfig(peer, peerConfig)
+		}
+
+		By("Updating bccsp orderer config")
+		ordererConfig := network.ReadOrdererConfig(orderer)
+		ordererConfig.General.BCCSP = myBCCSP
+		network.WriteOrdererConfig(orderer, ordererConfig)
+
+		network.Bootstrap()
+
+		ctx, sess := setupPKCS11Ctx(lib, label, pin)
+		Expect(err).NotTo(HaveOccurred())
+		defer ctx.CloseSession(sess)
+
+		configurePeerPKCS11(ctx, sess, network)
+		configureOrdererPKCS11(ctx, sess, network)
+
+		By("Starting fabric processes")
+		networkRunner := network.NetworkGroupRunner()
+		process = ifrit.Invoke(networkRunner)
+		Eventually(process.Ready(), network.EventuallyTimeout).Should(BeClosed())
+	})
+
+	AfterEach(func() {
+		process.Signal(syscall.SIGTERM)
+		Eventually(process.Wait(), network.EventuallyTimeout).Should(Receive())
+
+		if network != nil {
+			network.Cleanup()
+		}
+		os.RemoveAll(tempDir)
+	})
+
+	It("Creates and join channel successfully", func() {
+		orderer := network.Orderer("orderer")
+		network.CreateAndJoinChannels(orderer)
+	})
+})
+
+func configurePeerPKCS11(ctx *pkcs11.Ctx, sess pkcs11.SessionHandle, network *nwo.Network) {
+	for _, peer := range network.Peers {
+		orgName := peer.Organization
+
+		peerPubKey, peerCSR, peerSerial := createCSR(ctx, sess, orgName, "peer")
+		adminPubKey, adminCSR, adminSerial := createCSR(ctx, sess, orgName, "admin")
+
+		domain := network.Organization(orgName).Domain
+
+		// Retrieves org CA cert
+		orgCAPath := network.PeerOrgCADir(network.Organization(orgName))
+		caBytes, err := ioutil.ReadFile(filepath.Join(orgCAPath, fmt.Sprintf("ca.%s-cert.pem", domain)))
+		Expect(err).NotTo(HaveOccurred())
+
+		By("Updating the peer signcerts")
+
+		newOrdererPemCert := buildCert(network, caBytes, orgCAPath, peerCSR, peerSerial, peerPubKey)
+		updateMSPFolder(network.PeerLocalMSPDir(peer), fmt.Sprintf("peer.%s-cert.pem", domain), newOrdererPemCert)
+
+		By("Updating the peer admin user signcerts")
+		newAdminPemCert := buildCert(network, caBytes, orgCAPath, adminCSR, adminSerial, adminPubKey)
+
+		orgAdminMSPPath := network.PeerUserMSPDir(peer, "Admin")
+		updateMSPFolder(orgAdminMSPPath, fmt.Sprintf("Admin@%s-cert.pem", domain), newAdminPemCert)
+	}
+}
+
+func configureOrdererPKCS11(ctx *pkcs11.Ctx, sess pkcs11.SessionHandle, network *nwo.Network) {
+	orderer := network.Orderer("orderer")
+	orgName := orderer.Organization
+	domain := network.Organization(orgName).Domain
+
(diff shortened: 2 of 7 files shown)


scripts/run-integration-tests.sh
```diff
@@ -30,4 +30,9 @@ for ((i = "$agent_number"; i <= "${#test_dirs[@]}"; )); do
 done
 
 printf "\nRunning the following test suites:\n\n%s\n\nStarting tests...\n\n" "$(echo "${dirs[@]}" | tr -s ' ' '\n')"
-ginkgo -keepGoing --slowSpecThreshold 60 "${dirs[@]}"
+
+if [ -z "${PKCS11_LIB-}" ] || [ -z "${PKCS11_LABEL-}" ] || [ -z "${PKCS11_PIN-}" ]; then
```
sykesm: Can we avoid this? The suite can be skipped within the integration test if these aren't set.

scripts/run-integration-tests.sh
```diff
@@ -30,4 +30,9 @@ for ((i = "$agent_number"; i <= "${#test_dirs[@]}"; )); do
...
+   ginkgo -keepGoing --slowSpecThreshold 60 "${dirs[@]}"
+else
+   ginkgo -tags pkcs11 -keepGoing --slowSpecThreshold 60 "${dirs[@]}"
```
sykesm: Why would the tags be here instead of passed as arguments to the build server? Combined with the above check we wouldn't be building with pkcs11 tags if we're not running the tests.

integration/pkcs11/pkcs11_test.go
```diff
@@ -0,0 +1,420 @@
...
+	By("Updating the orderer admin user signcerts")
+	newAdminPemCert := buildCert(network, caBytes, orgCAPath, adminCSR, adminSerial, adminPubKey)
+
+	orgAdminMSPPath := network.OrdererUserMSPDir(orderer, "Admin")
+	updateMSPFolder(orgAdminMSPPath, "Admin@example.com-cert.pem", newAdminPemCert)
+
+}
+
+// Creates pkcs11 context and session
+func setupPKCS11Ctx(lib, label, pin string) (*pkcs11.Ctx, pkcs11.SessionHandle, error) {
```
sykesm: Why do we need to return an error? You're already asserting on errors.

integration/pkcs11/pkcs11_test.go
```diff
@@ -0,0 +1,420 @@
...
+	// Login
+	err = ctx.Login(sess, pkcs11.CKU_USER, pin)
+	Expect(err).NotTo(HaveOccurred())
+
+	return ctx, sess, err
+
+}
+
+// Identifies pkcs11 slot using specified label
+func findPKCS11Slot(ctx *pkcs11.Ctx, label string) (uint, error) {
```
sykesm: Here too. No need to return an error with an assertion in the body.

integration/pkcs11/pkcs11_suite_test.go
```diff
@@ -0,0 +1,49 @@
...
+
+	"github.com/hyperledger/fabric/integration"
+	"github.com/hyperledger/fabric/integration/nwo"
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+)
+
+func TestPKCS11(t *testing.T) {
+	RegisterFailHandler(Fail)
+	RunSpecs(t, "PKCS11 Suite")
```
sykesm: This is where we should be checking if we want to skip the tests because we don't have the PKCS11 environment variables setup.

integration/pkcs11/pkcs11_suite_test.go
```diff
@@ -0,0 +1,49 @@
...
+}
+
+var (
+	buildServer *nwo.BuildServer
+	components  *nwo.Components
+)
+
+var _ = SynchronizedBeforeSuite(func() []byte {
+	buildServer = nwo.NewBuildServer("-tags=pkcs11")
```
sykesm, line 35: And this is why we don't need tags when executing ginkgo.

integration/pkcs11/pkcs11_test.go
```diff
@@ -0,0 +1,420 @@
+// +build pkcs11
```
sykesm: There's no need for a tag here.

integration/pkcs11/pkcs11_test.go
```diff
@@ -0,0 +1,420 @@
...
+	bpkcs11 "github.com/hyperledger/fabric/bccsp/pkcs11"
+
+	"github.com/miekg/pkcs11"
+
+	docker "github.com/fsouza/go-dockerclient"
+	"github.com/hyperledger/fabric/integration/nwo/fabricconfig"
+	"github.com/tedsuo/ifrit"
+
+	"github.com/hyperledger/fabric/integration/nwo"
+	. "github.com/onsi/ginkgo"
```
sykesm: We can drop the blank lines.

integration/pkcs11/pkcs11_test.go
```diff
@@ -0,0 +1,420 @@
...
+
+		network.Bootstrap()
+
+		ctx, sess, err := setupPKCS11Ctx(lib, label, pin)
+		Expect(err).NotTo(HaveOccurred())
+
+		configurePeerPKCS11(ctx, sess, network)
+		configureOrdererPKCS11(ctx, sess, network)
+
+		ctx.CloseSession(sess)
```
sykesm: This should probably be a defer because of the assertions happening out of the config functions.

integration/pkcs11/pkcs11_test.go
```diff
@@ -0,0 +1,420 @@
...
+			foundSlot = true
+			slot = s
+			break
+		}
+	}
+
+	if !foundSlot {
+		Fail("Could not find slot with label")
+		return 0, err
+	}
```
sykesm: Why isn't this Expect(foundSlot).To(BeTrue(), "Could not find slot with label %s", label)

integration/pkcs11/pkcs11_test.go
```diff
@@ -0,0 +1,420 @@
...
+
+	return slot, nil
+
+}
+
+// Creates CSR for provided organization and organizational unit
+func createCSR(ctx *pkcs11.Ctx, sess pkcs11.SessionHandle, org, ou string) (*ecdsa.PublicKey, *x509.CertificateRequest, *big.Int) {
```
sykesm, line 231: We have multiple copies of a lot of this code floating around. We're at the point where we should consider consolidating them. I know there are versions in common and in comm.

integration/pkcs11/pkcs11_suite_test.go
```diff
@@ -0,0 +1,49 @@
...
+})
+
+var _ = SynchronizedAfterSuite(func() {
+}, func() {
+	buildServer.Shutdown()
+})
+
+func StartPort() int {
+	return integration.MSPPort.StartPortForNode()
+}
```
sykesm: Please don't reuse port ranges. Reserve a new range in integration/ports.go.

integration/pkcs11/pkcs11_test.go
```diff
@@ -0,0 +1,420 @@
...
+		client, err = docker.NewClientFromEnv()
+		Expect(err).NotTo(HaveOccurred())
+
+		network = nwo.New(nwo.BasicSolo(), tempDir, client, StartPort(), components)
+		network.GenerateConfigTree()
+
+		peer = network.Peers[0]
+		orderer = network.Orderer("orderer")
+
+		lib, pin, label := bpkcs11.FindPKCS11Lib()
```
sykesm, line 58: If this is the only reference you have to bccsp and the only reason you're adding the build tag, I'd *strongly* recommend just getting the environment variables we use in test.

integration/pkcs11/pkcs11_test.go
```diff
@@ -0,0 +1,411 @@
...
+
+	orgAdminMSPPath := network.OrdererUserMSPDir(orderer, "Admin")
+	updateMSPFolder(orgAdminMSPPath, "Admin@example.com-cert.pem", newAdminPemCert)
+
+}
+
+// Creates pkcs11 context and session
+func setupPKCS11Ctx(lib, label, pin string) (*pkcs11.Ctx, pkcs11.SessionHandle) {
+
+	ctx := pkcs11.New(lib)
```
sykesm: nit: more vertical white space

integration/pkcs11/pkcs11_test.go
```diff
@@ -0,0 +1,411 @@
...
+
+	ctx.Initialize()
```
sykesm: Error check?

integration/pkcs11/pkcs11_test.go
```diff
@@ -0,0 +1,411 @@
...
+	Expect(err).ToNot(HaveOccurred())
+
+	return pem.EncodeToMemory(&pem.Block{Bytes: signedCert, Type: "CERTIFICATE"})
+}
+
+// Overwrites existing cert and removes private key from keystore folder
+func updateMSPFolder(path, certName string, cert []byte) {
+	// Overwrite existing certificate with new certificate
+	signCertsPath := filepath.Join(path, "signcerts")
+	err := ioutil.WriteFile(fmt.Sprintf("%s/%s", signCertsPath, certName), cert, 0644)
```
sykesm: `filepath.Join` is preferred to `fmt.Sprintf`

integration/pkcs11/pkcs11_test.go
```diff
@@ -0,0 +1,411 @@
...
+		}
+	}
+
+	if !foundSlot {
+		Expect(foundSlot).To(BeTrue(), "Could not find slot with label %s", label)
+	}
+
+	return slot
+
+}
```
sykesm: I think this can be simplified a lot by returning immediately when the slot is found. After the for loop, we can have a simple `Fail(...)`.

integration/pkcs11/pkcs11_test.go
```diff
@@ -0,0 +1,411 @@
...
+
+// Generating key pair in HSM, convert, and return keys
+func generateKeyPair(ctx *pkcs11.Ctx, sess pkcs11.SessionHandle) (*ecdsa.PublicKey, *P11ECDSAKey) {
+	publabel, privlabel := "BCPUB7", "BCPRV7"
+
+	curve := asn1.ObjectIdentifier{1, 2, 840, 10045, 3, 1, 7} // secp256r1 Curve
+
+	marshaledOID, err := asn1.Marshal(curve)
+	Expect(err).ToNot(HaveOccurred())
```
sykesm: s/ToNot/NotTo/g

integration/pkcs11/pkcs11_test.go
```diff
@@ -0,0 +1,411 @@
...
+	pubK, privK, err := ctx.GenerateKeyPair(
+		sess,
+		[]*pkcs11.Mechanism{pkcs11.NewMechanism(pkcs11.CKM_EC_KEY_PAIR_GEN, nil)},
+		pubAttrs,
+		privAttrs,
+	)
+	Expect(err).ToNot(HaveOccurred())
+
+	ecpt := ecPoint(ctx, sess, pubK)
+	Expect(err).NotTo(HaveOccurred())
```
sykesm: What `err` are you asserting on?

integration/pkcs11/pkcs11_test.go
```diff
@@ -0,0 +1,411 @@
...
+
+	hash := sha256.Sum256(ecpt)
+	ski := hash[:]
```
sykesm, line 330: `ski := sha256.Sum256(ecpt)[:]`

integration/pkcs11/pkcs11_test.go
```diff
@@ -0,0 +1,411 @@
...
+
+	for _, a := range attr {
+		if a.Type == pkcs11.CKA_EC_POINT {
+			if ((len(a.Value) % 2) == 0) && (byte(0x04) == a.Value[0]) && (byte(0x04) == a.Value[len(a.Value)-1]) {
+				ecpt = a.Value[0 : len(a.Value)-1] // Trim trailing 0x04
+			} else if byte(0x04) == a.Value[0] && byte(0x04) == a.Value[2] {
+				ecpt = a.Value[2:len(a.Value)]
+			} else {
+				ecpt = a.Value
+			}
```
sykesm: if/else if/else should be rewritten as a switch

integration/pkcs11/pkcs11_test.go
```diff
@@ -0,0 +1,411 @@
...
+		pkcs11.NewAttribute(pkcs11.CKA_EC_PARAMS, nil),
+	}
+
+	attr, err := pkcs11lib.GetAttributeValue(session, key, template)
+	if err != nil {
+		Expect(err).ToNot(HaveOccurred(), "PKCS11: get(EC point)")
+	}
+
+	for _, a := range attr {
+		if a.Type == pkcs11.CKA_EC_POINT {
```
sykesm: we should flatten the loop by reversing the check and calling `continue`

integration/pkcs11/pkcs11_test.go
```diff
@@ -0,0 +1,411 @@
...
+			} else if byte(0x04) == a.Value[0] && byte(0x04) == a.Value[2] {
+				ecpt = a.Value[2:len(a.Value)]
+			} else {
+				ecpt = a.Value
+			}
+		}
+	}
+	if ecpt == nil {
+		Expect(ecpt).ToNot(BeNil(), "CKA_EC_POINT not found")
+	}
```
sykesm: This should be flattened by returning `ecpt` when processed above.

integration/ports.go
```diff
@@ -36,6 +36,7 @@ const (
 	PrivateDataBasePort
 	RaftBasePort
 	SBEBasePort
+	PKCS11Port
```
sykesm: These ports are actually supposed to be in alphabetical order.

sykesm (changes requested):
Looks good. Comments are mostly non-functional but would help with maintenance and readability.

integration/pkcs11/pkcs11_test.go
sykesm: Oh irony. I asked you to assert on `foundSlot` but stepping back, you don't need that variable so...

integration/pkcs11/pkcs11_test.go
caod123: I don't think arrays can be sliced into until they're actually allocated storage, the compiler will throw a `slice of unaddressable value` error if you try to one-line it. https://golang.org/ref/spec#Slice_expressions

integration/pkcs11/pkcs11_test.go
```diff
@@ -0,0 +1,411 @@
...
+		orderer := network.Orderer("orderer")
+		network.CreateAndJoinChannels(orderer)
+
+	})
+
+})
+
+func configurePeerPKCS11(ctx *pkcs11.Ctx, sess pkcs11.SessionHandle, network *nwo.Network) {
+
+	peer := network.Peers[0]
```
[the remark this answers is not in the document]
stephyee: Initially I thought there was only 1 🛠️

integration/pkcs11/pkcs11_test.go
```diff
@@ -0,0 +1,395 @@
...
+		Expect(err).NotTo(HaveOccurred())
+		defer ctx.CloseSession(sess)
+
+		configurePeerPKCS11(ctx, sess, network)
+		configureOrdererPKCS11(ctx, sess, network)
+
+		By("Starting fabric processes")
+		networkRunner := network.NetworkGroupRunner()
+		process = ifrit.Invoke(networkRunner)
+		Eventually(process.Ready(), time.Second*5).Should(BeClosed())
```
sykesm: Network.EventuallyTimeout ?

integration/pkcs11/pkcs11_test.go
sykesm: I can't argue with the compiler.

integration/pkcs11/pkcs11_test.go
```diff
@@ -0,0 +1,395 @@
...
+		if tokInfo.Label == label {
+			slot = s
+			break
+		}
+	}
+
+	Expect(slot).Should(BeNumerically(">", 0), "Could not find slot with label %s", label)
+
+	return slot
+}
```
sykesm: It's not a big deal but I still don't understand why we don't just return from inside the loop when we have the slot.

sykesm (dismissed):
PTAL at the ready timeout that was added.

integration/pkcs11/pkcs11_test.go
stephyee: 🔧

sykesm (approved):

PR #1476, opened 2020-06-27
[FAB-18028] Create new ExternalBuilder PropagateEnvironment key

lindluni:
Creates a new EnvPropagationList to replace the existing key. If the new key is used, it supersedes the existing key.

### DIFF
--- a/core/peer/config.go
+++ b/core/peer/config.go
@@ -37,7 +37,11 @@ import (
 // ExternalBuilder represents the configuration structure of
 // a chaincode external builder
 type ExternalBuilder struct {
-	EnvironmentWhitelist []string `yaml:"environmentWhitelist"`
+	// TODO: Remove Environment in 3.0
+	// Deprecated: Environment is retained for backwards compatibility.
+	// New deployments should use the new PropagateEnvironment field
+	Environment          []string `yaml:"environmentWhitelist"`
+	PropagateEnvironment []string `yaml:"propagateEnvironment"`
 	Name                 string   `yaml:"name"`
 	Path                 string   `yaml:"path"`
 }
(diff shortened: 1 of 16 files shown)


pamandrejko (dismissed):
Thanks for doing this Brett.

core/peer/config.go
```diff
@@ -37,9 +37,13 @@ import (
...
 // a chaincode external builder
 type ExternalBuilder struct {
-	EnvironmentWhitelist []string `yaml:"environmentWhitelist"`
-	Name                 string   `yaml:"name"`
-	Path                 string   `yaml:"path"`
+	// TODO: Remove EnvironmentList in 3.0
+	// Deprecated: EnvironmentList is retained for backwards compatibility.
+	// New deployments should use the new EnvPropagationList field
+	EnvironmentList    []string `yaml:"environmentWhitelist"`
+	EnvPropagationList []string `yaml:"envPropagationList"`
```
sykesm: No need for a list suffix on a list. `env` as an abbreviation in code is okay but I believe using the full name in the config key is more clear.

lindluni: I amended it, the only issue I had with the proposed name in Jira `propagateEnvironment` is that I felt it implied a boolean rather than a slice.

core/peer/config_test.go
```diff
@@ -354,6 +354,59 @@ func TestGlobalConfigDefault(t *testing.T) {
...
+		},
+		{
+			Name:                 "testName",
+			PropagateEnvironment: []string{"KEY=VALUE"},
+			Path:                 "/testPath",
+		},
+		{
+			Name:                 "testName",
+			Environment:          []string{"KEY=VALUE"},
+			PropagateEnvironment: []string{"KEY=VALUE"},
```
sykesm: Shouldn't these be different so we can determine which took precedence?

sykesm (approved):
LGTM other than the test. Can be fixed on top if needed.

core/peer/config_test.go
lindluni: done

lindluni:

PR #1481, opened 2020-06-28
FAB-18035 Ch.Part.API: consenter JoinChain creates a follower.Chain

tock-ibm:
Implement the JoinChain() method in the consenter, and create a skeleton implementation of a follower.Chain.

Create a cluster.BlockPuller that is initiated from the join-block.

Create a SelfMembershipPredicate for the follower.Chain.

The JoinChain method tries to be strict and check the inputs before chain creation, in order to minimize the risk of errors after the chain is started.

Change-Id: I94043a66c73488e32ac2a49606886e1f94be5cf0

#### Type of change
- New feature

#### Related issues

Task: FAB-18035
Story: FAB-15711
Epic: FAB-17712


orderer/consensus/follower/follower_chain.go
```diff
@@ -51,16 +137,121 @@ func (*Chain) Errored() <-chan struct{} {
...
-	return types.ClusterRelationFollower, status
+	if (c.joinBlock != nil) && (c.joinBlock.Header.Number >= c.support.Height()) {
+		status = types.StatusOnBoarding
+	}
+
+	return clusterRelation, status
+}
+
+func (c *Chain) run() {
+	c.logger.Debug("Enter")
```
guoger: not a big fan of this log pattern, especially when this is intended to be a long-running func... it would be pretty hard to pair the `Exit` with its `Enter`, no?

guoger (commented):
question, what would happen if the join-block used has a height lower than the currently ledger height on orderer node? IIUC, this should be an invalid request?

orderer/consensus/follower/follower_chain.go
sykesm: Additionally, imagine what happens when the calling function is removed from the record. What's the context? What does it mean? Log formats are customizable.

tock-ibm:
@guoger A `Join` request will continue to the consenter's `JoinChain()`  only if the ledger is empty and the channel does not exist. If the channel already exists the request is rejected by the `Registrar` prior to calling this method.

orderer/consensus/follower/follower_chain.go
tock-ibm: I am open to suggestions... the intent here is not really to pair the `Enter` and `Exit`, but to be able to detect that the go-routine actually started after say, a Join event; or conversely, actually exited after on-boarding has completed, for example.
These maybe far separated in time so normally there is no need to pair them, but one would like to know that the go-routine had started or ended in correlation to other meaningful events.
 
@sykesm do you mean you prefer a longer message with the full context? like `follower.Chain puller go-routine is starting`?

tock-ibm: Done

orderer/consensus/etcdraft/consenter.go
```diff
@@ -160,13 +160,32 @@ func (c *Consenter) HandleChain(support consensus.ConsenterSupport, metadata *co
...
 			return &inactive.Chain{Err: errors.Errorf("channel %s is not serviced by me", support.ChannelID())}, nil
 		} else {
+			// There is no system channel, follow this application channel.
 			//TODO fully construct a follower chain
-			return &follower.Chain{Err: errors.Errorf("orderer is a follower of channel %s", support.ChannelID())}, nil
+			consenterCertificate := &ConsenterCertificate{
+				ConsenterCertificate: c.Cert,
+				CryptoProvider:       c.BCCSP,
+			}
+			followerChain, err := follower.NewChain(
```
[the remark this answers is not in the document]
tock-ibm: Done

guoger:
@tock-ibm thx for reply! according to rfc, when a node is deleted from a channel, the deletion of resource storage (i assume ledger?) is optional and controlled by `options.removeStorage`. Is this channel still being tracked by node's `Registrar`? If not, then we might have the case where a join block used is behind the tip, no? (equivalent to resubmission of the join block)

tock-ibm:
@guoger This does not directly answer you question, but let me provide some context. Currently, the "Remove" story concentrates on the complete deletion of resources (i.e. removeStorage=true). So a removal of a channel removes the ledger and its entry in the Registrar, so there is no issue there. The "archiving" option in the remove story is currently low priority, so its implementation will come later. In the archiving option, storage is moved to a location away from the ledger and the entry in the registrar is removed. The issue really is how to restore from archive. Just using a "join" will be equivalent to re-pulling the entire ledger from scratch (note that it is even possible to create a different channel altogether by the same name). Currently the RFC does not specify an explicit restore technique. The way I am currently thinking about it, it is a kind of an emergency manual restore operation done by the admin. If the admin moves the storage back into place and restarts the orderer, then either a follower.Chain or an etcdraft.Chain will be created on top of the old ledger. The chain will then pull the rest of the ledger as normal. (There is a problem with rebuilding the indexes though).

That might present a problem only if the config of the channel changed so much such that the restored node no longer has the ability to pull... The simple, but possibly costly solution, is to join from scratch with a recent config block. However, a better solution is to supply the chain with a more recent config block to the channel in order to allow it to catch-up. (This can also happen when a node is down for a long time). **That is a point that is indeed currently missing in the RFC and the epic stories.** So - a great question because it uncovers a hole in our design. This would probably have to be implemented with another REST verb like a `PUT` (vs POST for Join) to perform an `Update`. 

More specifically to your question - if by any chance a follower chain finds a join block lower than the ledger height, it discards the join block and recreates the chain (via the CreateChain callback). This will spawn either an etcdraft.Chain or a follower.Chain, depending on the last config. This can happen only after a crash. You'll see that in the next PR.

orderer/consensus/follower/follower_chain.go
```diff
@@ -7,41 +7,133 @@ SPDX-License-Identifier: Apache-2.0
...
+
+		chain.logger.Infof("Created with join-block number: %d, ledger height: %d", joinBlock.Header.Number, support.Height())
+	}
+
+	return chain, nil
 }
 
 func (c *Chain) Order(_ *common.Envelope, _ uint64) error {
-	return c.Err
+	return errors.Errorf("orderer is a follower of channel %s", c.support.ChannelID())
```
yacovm, line 128: unrelated to this PR: It can be nice if the follower can send the transaction to one of the consenters via the endpoint in the consenter metadata, and it will forward onwards the transaction. 

Then, we get some sort of opaqueness to the user as it doesn't care who is a follower and who is not.

orderer/consensus/follower/follower_chain.go
```diff
@@ -51,16 +143,128 @@ func (*Chain) Errored() <-chan struct{} {
...
 func (c *Chain) Halt() {
-	//TODO skeleton
+	c.halt()
+
+	select {
+	case <-c.doneChan:
+	}
+}
+
+func (c *Chain) halt() {
```
yacovm, line 169: Why not just lock the chain and do:
```

select {
     case <- c.doneChan:
       return
    default: 
       close(c.doneChan)
}
```

?

I think the boolean stopped is redundant in this case

orderer/consensus/follower/follower_chain.go
```diff
@@ -51,16 +143,128 @@ func (*Chain) Errored() <-chan struct{} {
...
 
-// StatusReport returns the ClusterRelation & Status
+// StatusReport returns the ClusterRelation & Status.
 func (c *Chain) StatusReport() (types.ClusterRelation, types.Status) {
+	c.mutex.Lock()
+	defer c.mutex.Unlock()
+
+	clusterRelation := types.ClusterRelationFollower
+	if c.joinBlock != nil {
+		if err := c.amIInChannel(c.joinBlock); err == nil {
```
yacovm, line 189: what if you get a block that joins you into the channel? you are no longer a follower, right? In this case i guess this chain no longer exists anymore so this is not relevant right?

then why can we return a member if we shouldn't have the instance of a follower anymore?

orderer/consensus/follower/follower_chain.go
```diff
@@ -51,16 +143,128 @@ func (*Chain) Errored() <-chan struct{} {
...
+	if (c.joinBlock != nil) && (c.joinBlock.Header.Number >= c.support.Height()) {
+		status = types.StatusOnBoarding
+	}
+
+	return clusterRelation, status
+}
+
+func (c *Chain) run() {
+	c.logger.Debug("The follower.Chain puller goroutine is starting")
+	pullRetryTicker := time.NewTicker(500 * time.Millisecond) // TODO move constant to options
```
yacovm, line 204: In the next PR you just do time.After so the TODO isn't relevant i guess right?

orderer/consensus/follower/follower_chain.go
```diff
@@ -51,16 +143,128 @@ func (*Chain) Errored() <-chan struct{} {
...
+
+	err := c.pull()
+
+runLoop:
+	for err != nil {
```
yacovm, line 214: I really dislike labels and goto :( 

an alternative simpler solution would be:
```

defer func() {
	pullRetryTicker.Stop()
	close(c.doneChan)
	c.logger.Debug("The follower.Chain puller goroutine is exiting")
}

for {
  select {
    case   <- pullRetryTicker.C:
       // do nothing
   case <- stopChain:
           return
   }
    if err = c.pull(); err != nil {
           return
}
```

orderer/consensus/follower/follower_chain.go
```diff
@@ -51,16 +143,128 @@ func (*Chain) Errored() <-chan struct{} {
...
+	close(c.doneChan)
+	c.logger.Debug("The follower.Chain puller goroutine is exiting")
+}
+
+// pull blocks from other orderers, until the join block, or until a config block the indicates the orderer has become
+// a member pf the cluster. When the follower.Chain's job is done, this method halts, triggers the creation of a new
+// consensus.Chain, and returns nil.
+func (c *Chain) pull() error {
+	if c.joinBlock != nil {
+		err := c.pullUpToJoin()
```
yacovm, line 233: consider to just have a function pointer as a field and just invoke the function pointer.
The function pointer can be set at initialization time.

orderer/consensus/follower/follower_chain.go
```diff
@@ -51,16 +143,128 @@ func (*Chain) Errored() <-chan struct{} {
...
+		if err != nil {
+			return errors.Wrap(err, "failed to pull after join block")
+		}
+	}
+
+	// Trigger creation of a new chain. This will regenerate 'support' from the tip of the ledger, and will
+	// start either a follower.Chain or an etcdraft.Chain, depending on whether the orderer is a follower or a
+	// member of the cluster.
+	c.halt()
+	if c.chainCreationCallback != nil {
```
yacovm, line 250: consider checking at initialization and just assining a no-op if nil and anyway invoke it here anyway

orderer/consensus/follower/follower_chain.go
tock-ibm: That is an interesting idea. However for now I think the main reference scenario is that we create a follower for a channel even before we admit it to the channel endpoints and consenters set, and only after on-boarding a config update admits it into both endpoints and consenters set.

orderer/consensus/follower/follower_chain.go
tock-ibm: The `done` channel is closed only by the internal go-routine, as it exits. 
The `stop` channel is closed to signal the internal routine to exit. 

Note that the internal `halt` is called both by external and internal threads. The external comes through the `Halt`, and need to "wait" for the internal routine to end (if it was started), whereas the internal call to halt does not need to wait (because it is the one terminating). 
Note that the `stop` boolean also protects against a channel being started after it is stopped, in the `Start`; i.e. someone calling `Halt` then `Start`.

orderer/consensus/follower/follower_chain.go
tock-ibm: The follower.Chain also performs all on-boarding to a channel, even if the join-block indicates membership. So in case an orderer gets a join-block for which `c.amIInChannel(c.joinBlock)==nil`, it will still start a follower.Chain, which will pull blocks until the join-block, and then invoke `CreateChannel` and mutate into a `etcdraft.Chain`. (see next PR).

orderer/consensus/follower/follower_chain.go
tock-ibm: correct, this is completely revised, so the TODO is taken care of in the next PR.

orderer/consensus/follower/follower_chain.go
yacovm: Right, but - until we have fully morphed into a real chain I think we should return that we are a follower, no? Otherwise we are reporting a future state.

orderer/consensus/follower/follower_chain.go
yacovm: What do you think about changing the done channel to a waitgroup? I think it's simpler to reason about.

orderer/consensus/follower/follower_chain.go
tock-ibm: What you wrote in the example does not work. When the ticker hits, we want to pull again, not do nothing, and when we pull and get error we want to try again after the next ticker hits, not return. 

However I got your intention, one can certainly put the tail of the function in a defer and return early when the stop signal hits. 

And for the record, the "tags" are encouraged in go for[ nested structures that use break](https://forum.golangbridge.org/t/is-using-continue-label-or-break-label-good-practice/8345) , but I agree it is redundant here.

orderer/consensus/follower/follower_chain.go
tock-ibm: I really dislike function pointers for flow control. I think it makes the code harder to read and maintain. I think it is justified for decoupling components, like in the "CreateChain", and even there, I would have preferred an interface.

orderer/consensus/follower/follower_chain.go
tock-ibm: That would just add lines to the code while at the same time making it less readable... I mean the `if c.chainCreationCallback != nil {...}` is still there, but the noop now has to be defined.... what is the benefit?
I prefer simplicity...

orderer/consensus/follower/follower_chain.go
yacovm: I put the bracket in the wrong place... I updated the example.

tock-ibm: I understand how to eliminate the tag... 
In your corrected example the first pull happens only after a tick... and the return needs to be `if err==nil`...
But again, got your intention and will correct it.

tock-ibm: Done

orderer/consensus/follower/follower_chain.go
tock-ibm: The fact that it is a transient state is carried by the "status=onboarding", I think. Think about the user - he gave the system a block he knows indicates this orderer is a member... 
I think he expects to see "member, onboarding" -> "member, active"

The think the fact that we internally "re-use" the follower chain to do on-boarding does not interest him. That is how I think of it.

orderer/consensus/follower/follower_chain.go
tock-ibm: This will conflict with later PRs where the done channel is used in a select. Lets revisit this after we see later PRs.

yacovm (approved):

PR #1490, opened 2020-06-29
PKCS11 test environment variables take precedence

sykesm:
The implementation of FindPKCS11Lib would not honor PKCS11_PIN or PKCS11_LABEL if the location of the library was not specified. This change enables the override for test when needed.


caod123 (approved):

caod123:
Been seeing this CI failure a lot more frequently recently: https://jira.hyperledger.org/browse/FAB-18038

PR #1491, opened 2020-06-29
Add Check For CouchDB 3.1

lindluni:
Adds a warning to the log if the peer is using a version of CouchDB prior to 3.1, as this is the version we officially support.

### DIFF
--- a/core/ledger/kvledger/txmgmt/statedb/statecouchdb/couchdbutil.go
+++ b/core/ledger/kvledger/txmgmt/statedb/statecouchdb/couchdbutil.go
@@ -98,18 +98,16 @@ func createCouchInstance(config *ledger.CouchDBConfig, metricsProvider metrics.P
 	return couchInstance, nil
 }
 
-//checkCouchDBVersion verifies CouchDB is at least 2.0.0
 func checkCouchDBVersion(version string) error {
-
-	//split the version into parts
-	majorVersion := strings.Split(version, ".")
-
-	//check to see that the major version number is at least 2
-	majorVersionInt, _ := strconv.Atoi(majorVersion[0])
-	if majorVersionInt < 2 {
-		return errors.Errorf("CouchDB must be at least version 2.0.0. Detected version %s", version)
+	couchVersion := strings.Split(version, ".")
+	majorVersion, _ := strconv.Atoi(couchVersion[0])
+	minorVersion, _ := strconv.Atoi(couchVersion[1])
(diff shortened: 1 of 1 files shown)


lindluni:

core/ledger/kvledger/txmgmt/statedb/statecouchdb/couchdbutil.go
```diff
@@ -98,18 +98,16 @@ func createCouchInstance(config *ledger.CouchDBConfig, metricsProvider metrics.P
...
-	if majorVersionInt < 2 {
-		return errors.Errorf("CouchDB must be at least version 2.0.0. Detected version %s", version)
+	couchVersion := strings.Split(version, ".")
+	majorVersion, _ := strconv.Atoi(couchVersion[0])
+	minorVersion, _ := strconv.Atoi(couchVersion[1])
+	if majorVersion < 2 {
+		return errors.Errorf("CouchDB v%s detected. CouchDB must be at least version 2.0.0", version)
+	}
+	if majorVersion != 3 && minorVersion < 1 {
+		logger.Warnf("CouchDB v%s detected. CouchDB versions before 3.1.0 are unsupported.", version)
```
caod123, line 109: Is this ok to just log a warning instead of returning an error?

lindluni: Technically 2.x Couch is highly tested in previous releases, and users can continue to use it, it is just not our officially supported version anymore. We've actually modified the peer to support either version of Couch. Moving forward however we are only testing on 3.x Couch so we've decided to just make our official position you should be on Couch3, but we cant stop you if you are on Couch2.

If we were to error out, we would break backward compatibility of a peer thats migrated from 2.1 to 2.2 fabric. We can't error out until Fabric 3.0 to support backwards compatibility

lindluni: This will come with a strong statement of recommendation for users to migrate to Couch 3 in the release notes

caod123: On that note this check won't actually log a warning for 3.0.x or 2.1.x or for that matter any 2.x version where x >= 1. I think you may want to use an `||` here instead.

caod123 (approved):

denyeart (approved):

PR #1495, opened 2020-06-30
Only canonize ECDSA signatures in MSP:IsWellFormed

yacovm:
Currently, the MSP IsWellFormed function expects any signature to be
a valid ECDSA signature, however the certificate can be signed by
a non-ECDSA algorithm which will then yield a false negative.

This change set ensures the check only applies if the signature is ECDSA.

Change-Id: I0b14e3e9b87e860a3ca29cc233dc4810de1768ab


C0rWin (approved):

yacovm:

PR #1505, opened 2020-06-30
[FAB-18009] Ch.Part.API: Initialize file repos during registrar initialization

caod123:
#### Type of change

- New feature

#### Description

* Only used when channel participation API is enabled
* Initializes a joinblock file repo for the channel participation api

#### Related issues

Task: [FAB-18009](https://jira.hyperledger.org/browse/FAB-18009)

### DIFF
--- a/integration/e2e/e2e_test.go
+++ b/integration/e2e/e2e_test.go
@@ -120,6 +120,7 @@ var _ = Describe("EndToEnd", func() {
 				core.VM = nil
 				network.WritePeerConfig(peer, core)
 			}
+
 			network.Bootstrap()
 
 			networkRunner := network.NetworkGroupRunner()
--- a/orderer/common/multichannel/registrar.go
+++ b/orderer/common/multichannel/registrar.go
@@ -11,8 +11,7 @@ package multichannel
 
 import (
 	"fmt"
-	"github.com/hyperledger/fabric/orderer/common/cluster"
-	"github.com/hyperledger/fabric/orderer/common/follower"
+	"path/filepath"
 	"sync"
 
 	cb "github.com/hyperledger/fabric-protos-go/common"
@@ -25,6 +24,9 @@ import (
 	"github.com/hyperledger/fabric/common/metrics"
 	"github.com/hyperledger/fabric/internal/pkg/identity"
 	"github.com/hyperledger/fabric/orderer/common/blockcutter"
+	"github.com/hyperledger/fabric/orderer/common/cluster"
+	"github.com/hyperledger/fabric/orderer/common/filerepo"
+	"github.com/hyperledger/fabric/orderer/common/follower"
 	"github.com/hyperledger/fabric/orderer/common/localconfig"
 	"github.com/hyperledger/fabric/orderer/common/msgprocessor"
 	"github.com/hyperledger/fabric/orderer/common/types"
@@ -58,6 +60,8 @@ type Registrar struct {
 	callbacks          []channelconfig.BundleActor
 	bccsp              bccsp.BCCSP
 	clusterDialer      *cluster.PredicateDialer
+
+	joinBlockFileRepo *filerepo.Repo
 }
 
 // ConfigBlock retrieves the last configuration block from the given ledger.
@@ -101,7 +105,25 @@ func NewRegistrar(
 		clusterDialer:      clusterDialer,
 	}
 
+	if config.ChannelParticipation.Enabled {
+		r.initializeJoinBlockFileRepo()
+	}
+
 	return r
+}
+
+// initialize the channel participation API joinblock file repo. This creates
+// the fileRepoDir on the filesystem if it does not already exist.
+func (r *Registrar) initializeJoinBlockFileRepo() {
+	fileRepoDir := filepath.Join(r.config.FileLedger.Location, "filerepo")
+	logger.Infof("Channel Participation API enabled, registrar initializing with file repo %s", fileRepoDir)
+
+	joinBlockFileRepo, err := filerepo.New(fileRepoDir, "joinblock")
+	if err != nil {
+		logger.Panicf("Error initializing join block file repo: %s", err)
+	}
+
+	r.joinBlockFileRepo = joinBlockFileRepo
 }
 
 func (r *Registrar) Initialize(consenters map[string]consensus.Consenter) {
--- a/orderer/common/multichannel/registrar_test.go
+++ b/orderer/common/multichannel/registrar_test.go
@@ -11,6 +11,7 @@ import (
 	"io/ioutil"
 	"os"
 	"path"
+	"path/filepath"
 	"testing"
 
 	"github.com/golang/protobuf/proto"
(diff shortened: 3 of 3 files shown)


caod123:
This PR is dependent on #1441

orderer/common/multichannel/registrar.go
```diff
@@ -237,6 +244,30 @@ func (r *Registrar) Initialize(consenters map[string]consensus.Consenter) {
...
 	}
+
+	if r.config.ChannelParticipation.Enabled {
+		// Initialize channel-participation API file repos
+		fileRepoDir := r.config.ChannelParticipation.FileRepoDir
+		logger.Infof("Channel Participation API enabled, registrar initializing with file repo %s", fileRepoDir)
+
+		if err := validateFileRepoDir(fileRepoDir); err != nil {
+			logger.Panicf("Error initializing file repo dir: %s", err)
+		}
```
tock-ibm: This is a good direction, but maybe it belongs within the file-repo constructor?

orderer/common/filerepo/filerepo.go
```diff
@@ -176,3 +180,24 @@ func validateFileSuffix(fileSuffix string) error {
 
 	return nil
 }
+
+// validateFileRepoDir checks if the directory exists and is readable/writable
+func validateFileRepoDir(dir string) error {
+	// Check dir exists
+	if _, err := os.Stat(dir); os.IsNotExist(err) {
+		return err
+	}
```
cendhu: Nit. Can use https://github.com/hyperledger/fabric/blob/1ae54cf37d456582472488da221624ca217c9053/internal/fileutil/fileutil.go#L120

sampleconfig/orderer.yaml
```diff
@@ -361,6 +361,9 @@ ChannelParticipation:
     # The maximum size of the request body when joining a channel.
     MaxRequestBodySize: 1 MB
 
+    # Defines the file repo directory where channel partipation resources are stored.
+    FileRepoDir: /var/hyperledger/production/orderer/filerepo
```
cendhu: Do we really need to make it configurable? What is the advantage of doing this? Too many configurations affect the usability IMO. 

In the peer, we have a single file path configuration `fileSystemPath` and create all necessary internal folders there. For the ledger snapshot, a new filePath configuration would be introduced so that the user can decide to keep the snapshot on different storage especially when they create a snapshot to perform archival as the  `fileSystemPath` has no free capacity. 

If there is no such reasoning for making this configurable, I would prefer to remove this. Moreover, one needs to understand the internal of orderer before making a configuration decision on `FileRepoDir`, etc... IMO, this is internal to the orderer logic.

caod123: I think I agree with you here. The initial decision on configuring the file repo dir was from before when there was consideration for multiple file repo dirs (ie joinblock and remove). But after some refactoring in the file repo PR we've switched to subfolders of a root file repo parent dir (also reduces the number of configurations for needing to have one for each file repo dir). I can go ahead and remove this as it also simplifies some of the logic around requiring the parent repo dir to exist if we just create it internally as part of the orderer startup and default to an internal orderer subfolder.

caod123: @cendhu do you have any objections with reusing the file ledger's location so we don't have to create additional paths? Ie based off the default for the file ledger's location we could internally create a folder at `/var/hyperledger/production/orderer/filerepo` which is prefixed by the file ledger location.

caod123: The consensus `WALDir` and `SnapDir` seem to be configurable but the defaults are based off of the `FileLedger.Location` so that's probably the closest to a `fileSystemPath` for the orderer

caod123: Ah nevermind it seems the `FileLedger.Location` isn't the best choice here as it creates a temporary folder on orderer restart if the location is not configured. @cendhu what are your thoughts on the orderer's nonexistent `fileSystemPath` for internal folders. There doesn't seem to be a concept for one in the existing configuration. It would feel odd to add one just for usage by the file repo.

caod123: I think mirroring the peer here and adding a configurable `fileSystemPath` probably makes the most sense.

cendhu: As discussed offline, @manish-sethi, @caod123 and I are in favor of removing the temp file logic related to 
`FileLedger.Location`

https://github.com/hyperledger/fabric/blob/1ae54cf37d456582472488da221624ca217c9053/sampleconfig/orderer.yaml#L157

integration/nwo/network.go
```diff
@@ -138,18 +138,19 @@ type Profile struct {
...
+	DockerClient                    *docker.Client
+	ExternalBuilders                []fabricconfig.ExternalBuilder
+	NetworkID                       string
+	EventuallyTimeout               time.Duration
+	SessionCreateInterval           time.Duration
+	MetricsProvider                 string
+	StatsdEndpoint                  string
+	ClientAuthRequired              bool
+	ChannelParticipationEnabled     bool
+	ChannelParticipationFileRepoDir string
```
sykesm: Why are these on network? These are not network concepts but ordering feature flags. They should be moved to an appropriate position.

integration/e2e/e2e_test.go
```diff
@@ -213,7 +220,13 @@ var _ = Describe("EndToEnd", func() {
 		BeforeEach(func() {
 			network = nwo.New(nwo.BasicKafka(), testDir, client, StartPort(), components)
 			network.MetricsProvider = "prometheus"
+
 			network.ChannelParticipationEnabled = true
+			fileRepoDir := filepath.Join(testDir, "filerepo")
+			network.ChannelParticipationFileRepoDir = fileRepoDir
+			err := os.MkdirAll(fileRepoDir, 0755)
+			Expect(err).NotTo(HaveOccurred())
+
```
sykesm: So, does this mean we've replaced our `non-participation` tests with ones that enable it? Where are the legacy tests?

integration/e2e/e2e_test.go
```diff
@@ -213,7 +220,13 @@ var _ = Describe("EndToEnd", func() {
 		BeforeEach(func() {
 			network = nwo.New(nwo.BasicKafka(), testDir, client, StartPort(), components)
 			network.MetricsProvider = "prometheus"
+
 			network.ChannelParticipationEnabled = true
+			fileRepoDir := filepath.Join(testDir, "filerepo")
+			network.ChannelParticipationFileRepoDir = fileRepoDir
```
sykesm: Why is there a single file repository _for the network_? It does not seem that this is a network-wide location.

integration/nwo/orderer_template.go
```diff
@@ -120,4 +120,5 @@ Metrics:
 ChannelParticipation:
   Enabled: {{ .ChannelParticipationEnabled }}
   MaxRequestBodySize: 1 MB
+  FileRepoDir: {{ .ChannelParticipationFileRepoDir }}
```
sykesm: All orderers share the same location? Seems like a bad idea. Can you explain?

sampleconfig/orderer.yaml
wlahti: https://github.com/hyperledger/fabric/pull/1598 was recently merged to address this discussion. I'll update this PR soon to incorporate that and remove this FileRepoDir configuration option.

integration/nwo/network.go
wlahti: These have been moved/removed. We'll no longer have the FileRepoDir configuration and instead will store the file repo relative to the Orderer.FileLedger.Location.

integration/e2e/e2e_test.go
wlahti: These tests enable the channel participation API, but given that the system channel exists, it can only be used for listing channels. These tests still join using "legacy" functionality and simply verify that listing channels is possible.

integration/e2e/e2e_test.go
wlahti: Done. Fixed by moving the file repo directory under the configured file ledger location, which is different per orderer node.

integration/nwo/orderer_template.go
wlahti: Done. Fixed by moving the file repo directory under the configured file ledger location, which is different per orderer node.

orderer/common/multichannel/registrar.go
```diff
@@ -237,6 +241,29 @@ func (r *Registrar) Initialize(consenters map[string]consensus.Consenter) {
 			logger.Panicf("Error initializing without a system channel: failed to find an etcdraft consenter")
 		}
 	}
+
+	if r.config.ChannelParticipation.Enabled {
+		// Initialize channel-participation API file repos
+		r.initializeFileRepos()
```
tock-ibm: Maybe move this into the `New`

wlahti: Done

orderer/common/multichannel/registrar.go
```diff
@@ -101,9 +105,27 @@ func NewRegistrar(
...
 
+// initialize the channel participation API joinblock file repo. This creates
+// the fileRepoDir on the filesystem if it does not already exist.
+func (r *Registrar) initializeJoinBlockFileRepo() {
+	fileRepoDir := filepath.Join(r.config.FileLedger.Location, "filerepo")
+	logger.Infof("Channel Participation API enabled, registrar initializing with file repo %s", fileRepoDir)
+
+	joinBlockFileRepo, err := filerepo.New(fileRepoDir, "joinblock")
+	if err != nil {
+		logger.Panicf("Error initializing join block file repo: %s", err)
```
cendhu, line 123: Shouldn't panic happen at the `main()` rather than here? Given that the existing code in this file already has some 15 panic statement, I am not going to block this PR on that. However, please consider returning the err to the caller and take the panic decision at the `main()`

orderer/common/multichannel/registrar_test.go
```diff
@@ -254,6 +255,38 @@ func TestNewRegistrar(t *testing.T) {
 	})
 }
 
+func TestNewRegistrarWithFileRepo(t *testing.T) {
```
cendhu, line 258: Not related to this PR alone. In general, this package has no coverage for errors or panics. Please consider covering that if possible in the future PRs.

cendhu (approved):
Please consider fixing the two comments in sub-sequent PRs. As it blocks other PRs, I am merging it for now.

PR #1507, opened 2020-06-30
Use directly leveldb batch

manish-sethi:

#### Type of change

#### Description
This PR improves the `leveldbhelper.UpdateBatch` by replacing the map that it maintains by the actual leveldb batch.
This reduces the memory consumed and a sorting of key-values is avoided if the data added is already in sorted order (such as loading the data from snapshot files)

### DIFF
--- a/common/ledger/util/leveldbhelper/leveldb_provider.go
+++ b/common/ledger/util/leveldbhelper/leveldb_provider.go
@@ -204,21 +204,20 @@ func (h *DBHandle) DeleteAll() error {
 	return nil
 }
 
+// NewUpdateBatch returns a new UpdateBatch that can be used to update the db
+func (h *DBHandle) NewUpdateBatch() *UpdateBatch {
+	return &UpdateBatch{
+		dbName: h.dbName,
+		Batch:  &leveldb.Batch{},
+	}
+}
+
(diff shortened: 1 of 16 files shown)


common/ledger/util/leveldbhelper/leveldb_provider.go
```diff
@@ -204,21 +204,20 @@ func (h *DBHandle) DeleteAll() error {
...
+	return &UpdateBatch{
+		dbName: h.dbName,
+		Batch:  &leveldb.Batch{},
+	}
+}
+
 // WriteBatch writes a batch in an atomic way
 func (h *DBHandle) WriteBatch(batch *UpdateBatch, sync bool) error {
-	if len(batch.KVs) == 0 {
+	if batch == nil || batch.Len() == 0 {
```
cendhu, line 217: Not related to this PR. I was checking whether `leveldb.WriteBatch` returns err when the batch is nil (it does). You might be aware of the following. Then, I found that they use transaction (`OpenTransaction()`, `Commit()`) when the batch size is large to skip writing to journal. I remember once you said that we need to look at disabling the journal writes as we have a block store.

cendhu (approved):

cendhu:

PR #1508, opened 2020-06-30
Fix chaincode package example path

denyeart:
#### Type of change

- Documentation update

#### Description

With the Fabric shift to Go modules, chaincode packaging no longer
honors local GOPATH when resolving the chaincode path. User needs
to utilize a fully qualified path or a path relative to PWD.

This change updates the doc example that made an assumption about GOPATH usage.

#### Related issues

https://github.com/hyperledger/fabric/pull/1064

#### Release Note

A release note will be added in v2.2 to make it clear that GOPATH is no longer utilized.

### DIFF
--- a/docs/source/commands/peerlifecycle.md
+++ b/docs/source/commands/peerlifecycle.md
@@ -361,11 +361,13 @@ A chaincode needs to be packaged before it can be installed on your peers.
 This example uses the `peer lifecycle chaincode package` command to package
 a Go chaincode.
 
+  * Use the `--path` flag to indicate the location of the chaincode.
+    The path must be a fully qualified path or a path relative to your present working directory.
   * Use the `--label` flag to provide a chaincode package label of `myccv1`
     that your organization will use to identify the package.
 
     ```
-    peer lifecycle chaincode package mycc.tar.gz --path github.com/hyperledger/fabric-samples/chaincode/abstore/go/ --lang golang --label myccv1
+    peer lifecycle chaincode package mycc.tar.gz --path $CHAINCODE_DIR --lang golang --label myccv1
     ```
 
 ### peer lifecycle chaincode install example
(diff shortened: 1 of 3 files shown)


denali49 (approved):
lgtm.

lindluni:
Technically the side-effect we are trying to fix here is a user operating in `GO111MODULE=on` mode where their chaincode is not on their GOPATH. Which we don't force them to do. Not sure what the best path of documenting this is.

Does the fully qualified path and the relative path work when you are operating in `GO111MODULE=off` mode? If it does, I suppose this change would work for either case?

denyeart:
@btl5037 The chaincode is on my local GOPATH. Regardless of whether `GO111MODULE=on` is set or not, I see the following behavior after #1064 was merged:

- Fully qualified path always works
- Relative path wrt PWD always works
- Relative path wrt GOPATH never works (it worked prior to #1064)

Therefore it seemed pretty straightforward to me - document that you must have fully qualified path or relative path wrt PWD as of v2.2.

Let's see if this is the behavior that @sykesm expected.

docs/source/commands/peerlifecycle.md
```diff
@@ -361,11 +361,13 @@ A chaincode needs to be packaged before it can be installed on your peers.
 This example uses the `peer lifecycle chaincode package` command to package
 a Go chaincode.
 
+  * Use the `--path` flag to indicate the location of the chaincode.
```
wlahti, line 364: FYI: you need to make these changes in docs/wrappers/peer_lifecycle_chaincode_postscript.md and then run "make clean help-docs". Otherwise, the next time someone submits a PR with an updated command reference this change will be lost.

wlahti (changes requested):

docs/source/commands/peerlifecycle.md
denyeart: @wlahti That's how I did it :-)
You'll see that both files are updated in this PR.

wlahti: Oops, guess I was reviewing before my brain fully woke up. Carry on. :)

wlahti (approved):

lindluni:
> * Fully qualified path always works
> * Relative path wrt PWD always works

If the FQP and the relative path both work then I think this is a better way to document it anyway. Using the GOPATH version was always kind of magic, when I was new to Fabric and Go, I really didn't understand how that ever worked.

lindluni (approved):

denyeart:

PR #1510, opened 2020-07-01
[FAB-17374] Integration test for Chaincode as an External Service

psingh-io:
**Type of change**
Test update

**Details**
Fixes the issues reported in PR for Chaincode as External Service. The original PR with comments is here:
https://github.com/hyperledger/fabric/pull/552

The test is in integration/e2e/chaincode_service_test.go file

### DIFF
--- a/integration/e2e/chaincode_service_test.go
+++ b/integration/e2e/chaincode_service_test.go
@@ -1,0 +1,195 @@
+/*
+Copyright IBM Corp All Rights Reserved.
+
+SPDX-License-Identifier: Apache-2.0
+*/
+
+package e2e
+
+import (
+	"encoding/json"
+	"fmt"
+	"io/ioutil"
+	"os"
+	"os/exec"
+	"path/filepath"
+	"syscall"
+	"time"
+
+	"github.com/hyperledger/fabric/common/crypto/tlsgen"
+	"github.com/hyperledger/fabric/core/container/externalbuilder"
+	"github.com/hyperledger/fabric/integration/nwo"
+	"github.com/hyperledger/fabric/integration/nwo/fabricconfig"
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+	"github.com/tedsuo/ifrit"
+	"github.com/tedsuo/ifrit/ginkgomon"
+)
+
+var _ = Describe("ChaincodeAsExternalService", func() {
+	var (
+		testDir                 string
+		network                 *nwo.Network
+		extcc                   nwo.Chaincode
+		chaincodeServerAddrress string
+		certFiles               []string
+		process                 ifrit.Process
+		extbldr                 fabricconfig.ExternalBuilder
+		ccserver                ifrit.Process
+	)
+
+	BeforeEach(func() {
+		var err error
+		testDir, err = ioutil.TempDir("", "e2e-chaincode-service")
+		Expect(err).NotTo(HaveOccurred())
+		extcc = nwo.Chaincode{
+			Name:            "mycc",
+			Version:         "0.0",
+			Path:            components.Build("github.com/hyperledger/fabric/integration/chaincode/extcc"),
+			Lang:            "extcc",
+			PackageFile:     filepath.Join(testDir, "extcc.tar.gz"),
+			Ctor:            `{"Args":["init","a","100","b","200"]}`,
+			SignaturePolicy: `AND ('Org1MSP.member','Org2MSP.member')`,
+			Sequence:        "1",
+			Label:           "my_extcc_chaincode",
+		}
+
+		extbldr = fabricconfig.ExternalBuilder{
+			Path: filepath.Join("..", "externalbuilders", "extcc"),
+			Name: "extcc",
+		}
+
+		network = nwo.New(nwo.BasicSolo(), testDir, nil, StartPort(), components)
+
+		chaincodeServerAddrress = fmt.Sprintf("127.0.0.1:%d", network.ReservePort())
+
+		tlsCA, err := tlsgen.NewCA()
+		Expect(err).NotTo(HaveOccurred())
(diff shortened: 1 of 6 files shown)


integration/e2e/chaincode_service_test.go
```diff
@@ -0,0 +1,275 @@
...
+func generateCertKeysAndConnectionFiles(tlsCA tlsgen.CA, chaincodeServerAddrress string, tempDir string, chaincodeID string) []string {
+	certsDir := chaincodeCertsDir(tempDir, chaincodeID)
+
+	// Generate key files for chaincode server
+	err := os.MkdirAll(certsDir, 0755)
+	Expect(err).NotTo(HaveOccurred())
+
+	serverKeyFile := filepath.Join(certsDir, "key.pem")
+	serverCertFile := filepath.Join(certsDir, "cert.pem")
+	clientCAFile := filepath.Join(certsDir, "clientCA.pem")
```
sykesm, line 202: error checks?

integration/nwo/deploy.go
```diff
@@ -84,6 +84,31 @@ func DeployChaincode(n *Network, channel string, orderer *Orderer, chaincode Cha
 	}
 }
 
+// DeployChaincodeWithoutInitialization is a helper that will install chaincode to all peers that
+// are connected to the specified channel, approve the chaincode on one of the
+// peers of each organization in the network, commit the chaincode definition
+// on the channel using one of the peers, and wait for the chaincode commit to
+// complete on all of the peers.
+// This method does not invoke initialization of chaincode. This is useful for chaincode as extenal service
+// where package code is required for stating the external chaincode service.
```
sykesm: This condition is for exactly one scenario. It doesn't warrant duplicating `DeployChaincode` in the same package that hods it.

integration/e2e/chaincode_service_test.go
```diff
@@ -0,0 +1,275 @@
...
+	return filepath.Join(tempDir, "certs", chaincodeID)
+}
+
+func chaincodeConnectionDir(tempDir string, chaincodeID string) string {
+	return filepath.Join(tempDir, "chaincode-connections", chaincodeID)
+}
+
+func chaincodeConnectionPath(tempDir string, chaincodeID string) string {
+	return filepath.Join(chaincodeConnectionDir(tempDir, chaincodeID), "connection.json")
+}
```
sykesm: Why functions for these instead of vars in the test setup?  They should be calculated once.

integration/e2e/chaincode_service_test.go
```diff
@@ -0,0 +1,275 @@
...
+				Name:          "TwoOrgsBaseProfileChannel",
+				Consortium:    "SampleConsortium",
+				Orderers:      []string{"orderer"},
+				Organizations: []string{"Org1", "Org2"},
+			})
+			network.Channels = append(network.Channels, &nwo.Channel{
+				Name:        "baseprofilechannel",
+				Profile:     "TwoOrgsBaseProfileChannel",
+				BaseProfile: "TwoOrgsOrdererGenesis",
+			})
```
sykesm: I don't understand why we need to augment the network before generating the config. I don't see anything in here referring to `baseprofilechannel`. Can you please elaborate on why this is needed or modify the test to use the out of the box `BasicSolo` config?

integration/e2e/chaincode_service_test.go
```diff
@@ -0,0 +1,275 @@
...
+			// package connection.json
+			extcc.CodeFiles = map[string]string{
+				chaincodeConnectionPath(testDir, extcc.Name): "connection.json",
+			}
+
+			for _, peer := range network.PeersWithChannel("testchannel") {
+				core := network.ReadPeerConfig(peer)
+				core.VM = nil
+				network.WritePeerConfig(peer, core)
+			}
```
sykesm: Why  do we need to do this? We already have other tests that assert we can operate a peer without a docker configuration.

sykesm (changes requested):
Thank you for reviving this.

There's still a lot of extra boilerplate and customization that should not be necessary. I'm also very keen on isolating the changes to the `nwo` package to what is necessary and reusable. In particular, the addition of `DeployWithoutInitialization` isn't necessary because we can simply pass the chaincode struct to `DeployChaincode` after disabling `InitRequired`.

When it comes to error checking within tests, we do want assertions on errors to make sure things go as expected. There are a lot of places where that checking is missing. These need to be addressed.

Before merging this, we should also get the duration marshaling bug fixed.

I'll be on the lookout for changes.

psingh-io:
> Thank you for reviving this.
> 
> There's still a lot of extra boilerplate and customization that should not be necessary. I'm also very keen on isolating the changes to the `nwo` package to what is necessary and reusable. In particular, the addition of `DeployWithoutInitialization` isn't necessary because we can simply pass the chaincode struct to `DeployChaincode` after disabling `InitRequired`.
> 
> When it comes to error checking within tests, we do want assertions on errors to make sure things go as expected. There are a lot of places where that checking is missing. These need to be addressed.
> 
> Before merging this, we should also get the duration marshaling bug fixed.
> 
> I'll be on the lookout for changes.

@sykesm Thanks for reviewing and your feedback. I will update the PR with required changes in a day or two.

sykesm:
@psingh-io Two small things.

1. Looks like there's a problem highlighted by the linter.
2. We'd prefer to see pull requests without merge commits.

Can you please address the problem and rebase on top of master instead of merging master?

psingh-io:
Thanks for reviewing. I will make these updates today.

Pancham

sykesm:
@psingh-io May I fixup your commits and push so I can merge this in short order? We just need to squash and I think it's good to merge.

psingh-io:
Yes, please.

I did rebase and cleaned up commit history by squashing in one. But, there
were 2 commits still showing. When I tried to squash the last two, I was
getting merges in commit history and hence had reverted.

Thanks for your help.
Pancham

sykesm (approved):

PR #1514, opened 2020-07-01
Update go-kit to v0.10.0

wlahti:
Trying to see if this helps with the statsd flake: [FAB-18038](https://jira.hyperledger.org/browse/FAB-18038l)


wlahti:
No failures the first time. Let's try it again. :)

wlahti:
Success two times in a row!

lindluni:
At least this flake wasn't because of the metrics provider :)

lindluni:
/azp run

Testing again

wlahti:
Ah, but there it is, which sadly means this is not the fix. I've been trying to investigate the root cause but no luck yet.

PR #1516, opened 2020-07-01
Import state from snapshotted data in stateleveldb

manish-sethi:

#### Type of change
- New feature

#### Description

This commit enables `statelevedb` for importing the state data for a channel from an iterator that is expected to read from previously snapshotted state data

#### Related issues
FAB-18042

### DIFF
--- a/common/ledger/util/leveldbhelper/leveldb_provider.go
+++ b/common/ledger/util/leveldbhelper/leveldb_provider.go
@@ -204,6 +204,21 @@ func (h *DBHandle) DeleteAll() error {
 	return nil
 }
 
+// IsEmpty returns true if no data exists for the DBHandle
+func (h *DBHandle) IsEmpty() (bool, error) {
+	itr, err := h.GetIterator(nil, nil)
+	if err != nil {
+		return false, err
+	}
+	defer itr.Release()
+
+	if err := itr.Error(); err != nil {
+		return false, errors.WithMessagef(itr.Error(), "internal leveldb error while obtaining next entry from iterator")
+	}
+
+	return !itr.Next(), nil
+}
+
 // NewUpdateBatch returns a new UpdateBatch that can be used to update the db
 func (h *DBHandle) NewUpdateBatch() *UpdateBatch {
 	return &UpdateBatch{
--- a/common/ledger/util/leveldbhelper/leveldb_provider_test.go
+++ b/common/ledger/util/leveldbhelper/leveldb_provider_test.go
@@ -372,6 +372,79 @@ func TestClose(t *testing.T) {
 	require.Equal(t, map[string]*DBHandle{}, p.dbHandles)
 }
 
+func TestIsEmpty(t *testing.T) {
+	var env *testDBProviderEnv
+	var db1, db2 *DBHandle
+
+	setup := func() {
+		env = newTestProviderEnv(t, testDBPath)
+		p := env.provider
+		db1 = p.GetDBHandle("db1")
+		db2 = p.GetDBHandle("db2")
+	}
+
+	cleanup := func() {
+		env.cleanup()
+	}
+
+	t.Run("both the dbs are empty", func(t *testing.T) {
+		setup()
+		defer cleanup()
+
+		empty, err := db1.IsEmpty()
+		require.NoError(t, err)
+		require.True(t, empty)
+
+		empty, err = db2.IsEmpty()
+		require.NoError(t, err)
+		require.True(t, empty)
+	})
+
+	t.Run("only one db is empty", func(t *testing.T) {
+		setup()
+		defer cleanup()
+
+		db1.Put([]byte("key"), []byte("value"), false)
+		empty, err := db1.IsEmpty()
+		require.NoError(t, err)
+		require.False(t, empty)
+
+		empty, err = db2.IsEmpty()
+		require.NoError(t, err)
+		require.True(t, empty)
+	})
+
+	t.Run("both the dbs contain data", func(t *testing.T) {
+		setup()
+		defer cleanup()
+
+		db1.Put([]byte("key"), []byte("value"), false)
+		db2.Put([]byte("key"), []byte("value"), false)
+
+		empty, err := db1.IsEmpty()
+		require.NoError(t, err)
+		require.False(t, empty)
+
+		empty, err = db2.IsEmpty()
+		require.NoError(t, err)
+		require.False(t, empty)
+	})
+
+	t.Run("iter error", func(t *testing.T) {
+		setup()
+		defer cleanup()
+
+		env.provider.Close()
+		empty, err := db1.IsEmpty()
+		require.EqualError(t, err, "internal leveldb error while obtaining db iterator: leveldb: closed")
+		require.False(t, empty)
+
+		empty, err = db2.IsEmpty()
+		require.EqualError(t, err, "internal leveldb error while obtaining db iterator: leveldb: closed")
+		require.False(t, empty)
+	})
+}
+
 func testFormatCheck(t *testing.T, dataFormat, expectedFormat string, dataExists bool, expectedErr *dataformat.ErrFormatMismatch) {
 	assert.NoError(t, os.RemoveAll(testDBPath))
 	defer func() {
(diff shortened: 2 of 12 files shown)


core/ledger/kvledger/txmgmt/statedb/stateleveldb/stateleveldb.go
```diff
@@ -220,6 +221,41 @@ func (vdb *versionedDB) GetFullScanIterator(skipNamespace func(string) bool) (st
...
+				return err
+			}
+		}
+	}
+	return vdb.db.WriteBatch(dbBatch, true)
+}
+
+// IsEmpty return true if the statedb does not have any content
+func (vdb *versionedDB) IsEmpty() (bool, error) {
+	return vdb.db.IsEmpty()
```
cendhu, line 258: 1. Will the `privacyenabledstate` pkg call this before calling the `ImportState()` so that it can decide on whether to call `Remove()`? 
2. For `statecouchdb`, does it translate to whether metadata DB exist? -- assuming that metadata DB is the first DB to be created during the `ImportState()`?

core/ledger/kvledger/txmgmt/statedb/stateleveldb/stateleveldb.go
```diff
@@ -220,6 +221,41 @@ func (vdb *versionedDB) GetFullScanIterator(skipNamespace func(string) bool) (st
...
+		if compositeKey == nil {
+			break
+		}
+		dataKey := encodeDataKey(compositeKey.Namespace, compositeKey.Key)
+		batchSize += len(dataKey) + len(dbValue)
+		dbBatch.Put(dataKey, dbValue)
+		if batchSize >= maxDataImportBatchSize {
+			if err := vdb.db.WriteBatch(dbBatch, true); err != nil {
+				return err
+			}
```
cendhu, line 248: missed to reset the batchSize to 0.

core/ledger/kvledger/txmgmt/statedb/stateleveldb/stateleveldb.go
```diff
@@ -220,6 +221,41 @@ func (vdb *versionedDB) GetFullScanIterator(skipNamespace func(string) bool) (st
...
+			return err
+		}
+		if compositeKey == nil {
+			break
+		}
+		dataKey := encodeDataKey(compositeKey.Namespace, compositeKey.Key)
+		batchSize += len(dataKey) + len(dbValue)
+		dbBatch.Put(dataKey, dbValue)
+		if batchSize >= maxDataImportBatchSize {
+			if err := vdb.db.WriteBatch(dbBatch, true); err != nil {
```
cendhu, line 246: missed to reset the leveldb batch.

core/ledger/kvledger/txmgmt/statedb/stateleveldb/stateleveldb.go
```diff
@@ -220,6 +221,41 @@ func (vdb *versionedDB) GetFullScanIterator(skipNamespace func(string) bool) (st
...
 
+// ImportState implements method in VersionedDB interface. The function is expected to be used
+// for importing the state from a previously snapshotted state. The parameter itr provides access to
+// the snapshotted state.
+func (vdb *versionedDB) ImportState(itr statedb.FullScanIterator, dbValueFormat byte) error {
+	if dbValueFormat != fullScanIteratorValueFormat {
+		return errors.Errorf("value format [%x] not supported. Expected value format [%x]",
+			dbValueFormat, fullScanIteratorValueFormat)
+	}
+	dbBatch := vdb.db.NewUpdateBatch()
```
cendhu, line 232: When we reset the leveldb batch, would it be better to expose leveldb `batch.Reset()` to reuse the allocated memory? If I understood the leveldb implementation of `batch.grow()`, if we use `batch.Reset()`, we might not use more than 4 MB (i.e., `maxDataImportBatchSIze`) in total. 

This is not particular to this PR as we create new batches in other places too. I can also take this up and make changes in all places at once if you agree.

core/ledger/kvledger/txmgmt/statedb/commontests/test_common.go
```diff
@@ -1014,40 +1013,58 @@ func TestFullScanIterator(
...
-			})
 		}
 
-		expectedNamespacesInResults := stringset{"", "ns1", "ns2", "ns3", "ns4"}.minus(skipNamespaces)
 		expectedResults := []*statedb.VersionedKV{}
-		for _, ns := range expectedNamespacesInResults {
+		for _, ns := range expectedNamespacesInDestinationDB {
 			expectedResults = append(expectedResults, generateSampleData(ns)...)
 		}
 		require.Equal(t, expectedResults, results)
```
cendhu: can be replaced wtih `require.Equal(t, generateSampleData(expectedNamespacesInDestinationDB...), results)`

common/ledger/util/leveldbhelper/leveldb_provider.go
```diff
@@ -204,6 +204,16 @@ func (h *DBHandle) DeleteAll() error {
...
 }
 
+// IsEmpty returns true if no data exists for the DBHandle
+func (h *DBHandle) IsEmpty() (bool, error) {
+	itr, err := h.GetIterator(nil, nil)
+	if err != nil {
+		return false, err
+	}
+	defer itr.Release()
+	return !itr.Next(), errors.WithMessagef(itr.Error(), "internal leveldb error while obtaining next entry from iterator")
```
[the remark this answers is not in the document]
manish-sethi: sure, I can change this.

common/ledger/util/leveldbhelper/leveldb_provider_test.go
```diff
@@ -372,6 +372,79 @@ func TestClose(t *testing.T) {
...
 
+func TestIsEmpty(t *testing.T) {
+	var env *testDBProviderEnv
+	var db1, db2 *DBHandle
+
+	setup := func() {
+		env = newTestProviderEnv(t, testDBPath)
+		p := env.provider
+		db1 = p.GetDBHandle("db1")
+		db2 = p.GetDBHandle("db2")
```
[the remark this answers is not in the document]
manish-sethi: Since underlying we share a single leveldb instance, its better to check that the state of one db is not affecting the other.

core/ledger/kvledger/txmgmt/statedb/commontests/test_common.go
manish-sethi: yes, this makes it shorter

core/ledger/kvledger/txmgmt/statedb/stateleveldb/stateleveldb.go
manish-sethi: It is already exposed, as I had used the levedb Batch as an embedded field. Yes, it should be possible to reuse the buffer but not with any guarantee. The challenge being how we compute the batch size outside of this. Because, leveldb internally add additional bytes (between 3 and 11) per kv. Guaranteed reuse can come only at the cost of some ugly looking code at this layer, where we also account for additional max possible bytes per kv. But yes, in general, the reuse frequency may increase with the reset.

core/ledger/kvledger/txmgmt/statedb/stateleveldb/stateleveldb.go
manish-sethi: yes, good catch

core/ledger/kvledger/txmgmt/statedb/stateleveldb/stateleveldb.go
manish-sethi: yes, In fact I was thinking of moving this size field inside batch so we could call `batch.Size` and `batch.Reset` but eventually missed at both the places :-)

core/ledger/kvledger/txmgmt/statedb/stateleveldb/stateleveldb.go
manish-sethi: yes, to both (except that for the couch case, the requirement of existence of metadatadb may be relaxed, so long it does not contain any `namespace-dbname` mapping data yet)

cendhu (approved):
Thanks, Manish. LGTM.

PR #1520, opened 2020-07-01
Try doubling datagram reader's buffer size

wlahti:
Seeing if the buffer size is cutting off data sent by the peer, leading to the metrics flake: [FAB-18038](https://jira.hyperledger.org/browse/FAB-18038)

### DIFF
--- a/integration/e2e/e2e_suite_test.go
+++ b/integration/e2e/e2e_suite_test.go
@@ -8,6 +8,7 @@ package e2e
 
 import (
 	"encoding/json"
+	"fmt"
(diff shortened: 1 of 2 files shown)


wlahti:
Worked once. Trying again.

wlahti:
Twice.

wlahti:
Thrice. Added a bit of debug output and an intentional fail to see if this is just dumb luck again. :)

wlahti:
Yep, just dumb luck. On to the next guess.

PR #1523, opened 2020-07-01
Bump viper and mapstructure

sykesm:
Bump viper to v1.1.1 and mapstructure to 1.3.2.


cendhu:
Given that `go build` and `go test` automatically downloads the dependencies, do we need to explicitly vendor them? As the [go module mirror](https://blog.golang.org/module-mirror-launch) takes care of disappearing dependencies too (at least that's what they claim), I would like to understand our stands and reasoning for not using the [GOPROXY](https://proxy.golang.org/)? 

Our vendor folder size is 32 MB and the sum of the rest of the folders is ~20 MB. The following are some of the projects that don't seem to use a vendor folder and instead download dependencies on demand

1. https://github.com/grafana/grafana
2. https://github.com/influxdata/influxdb
3. https://github.com/DataDog/datadog-agent
4. https://github.com/google/gvisor
5. https://github.com/istio/istio

I found this FAQ: Is this the reason for not using this feature in Fabric?

> Why did a previously available module become unavailable in the mirror?
> 
> proxy.golang.org does not save all modules forever. There are a number of reasons for this, but one reason is if proxy.golang.org is not able to detect a suitable license. In this case, only a temporarily cached copy of the module will be made available, and may become unavailable if it is removed from the original source and becomes outdated. The checksums will still remain in the checksum database regardless of whether or not they have become unavailable in the mirror.

But many other articles mention that go mirror is immutable (maybe it denotes checksum only?)

1. [The Go module proxy is by default caching and storing all the dependencies forever (in immutable storage). This means you don’t have to use any vendor/ folder anymore.](https://arslan.io/2019/08/02/why-you-should-use-a-go-module-proxy/)
2. [In GoCenter (non-google proxy), Go Modules are immutable and always available.](https://jfrog.com/blog/gocenter-journey/)

This [article](https://www.ardanlabs.com/blog/2020/02/modules-04-mirros-checksums-athens.html) suggests keeping the dependency tree closer to the project using a vendor folder if it is practical to do so.

> I believe every project should vendor their dependencies until it’s not reasonable or practical to do so. Projects like Docker and Kubernetes can’t vendor their dependencies because there are just too many dependencies to vendor. However, for most of us this is not the case. In version v1.14, there is great support for vendoring and modules. I will talk about this in a different post.
> 
> I am bringing up vendoring for an important reason. I have heard people using Athens or a private module mirror as a substitute for vendoring. I think this is a mistake. One has nothing to do with the other. You could argue a module mirror vendors dependencies since the code for the modules are persisted, but the code is still far away from the project that is depending on it. Even if you believe in the resilience of your module mirror, I believe there is no substitute for your project owning all the source code it needs and not depending on anything else but the project itself to build the code.

The above article is wrong in the sense by quoting kubernetes and docker cannot vendor their dependencies but they do. Given that you might have already done the research, I would love to hear before doing more studies (modules mirror = centralized trusted authority + immutabiliy 😉)

mastersingh24 (approved):
glad to see we are actually able to start bumping versions and using some of the newer features

PR #1536, opened 2020-07-05
Fab-17943 Ch.Part.API: follower pulls blocks until join-block

tock-ibm:
- New feature

#### Description

Join with a join-block number >0, onboarding needed.

Start a follower that pulls blocks until the join-block, then mutates into an etcdraft.Chain.

The follower.Chain pulls blocks up to the join block, without inspecting intermediate config blocks.

After the join-block is pulled, the follower chain is stopped and the ChannelCreate function is invoked.

#### Related issues

Task: Fab-17943 
Epic: FAB-17712


orderer/consensus/follower/block_puller.go
```diff
@@ -49,6 +50,20 @@ func BlockPullerFromJoinBlock(
...
+			return errors.New("buffer is empty")
+		}
+		if blocks[0] == nil {
+			return errors.New("first block is nil")
+		}
+		if blocks[0].Header == nil {
+			return errors.New("first block header is nil")
+		}
+		if blocks[0].Header.Number == 0 {
+			blocksAfterGenesis := blocks[1:]
```
yacovm, line 63: why are you doing this? Is it the signature? If it is, then I think there are other ways of not verifying the signature of block 0, but what you're doing here also forces not to validate the hash pointer of block 1, no?

orderer/consensus/follower/follower_chain.go
```diff
@@ -199,31 +235,69 @@ func (c *Chain) StatusReport() (types.ClusterRelation, types.Status) {
 	return clusterRelation, status
 }
 
+func (c *Chain) IsRunning() bool {
+	c.mutex.Lock()
+	defer c.mutex.Unlock()
+
+	if c.started {
```
yacovm, line 252: under what circumstance is a follower chain re-started after shutdown?

orderer/consensus/follower/follower_chain.go
```diff
@@ -199,31 +235,69 @@ func (c *Chain) StatusReport() (types.ClusterRelation, types.Status) {
...
-		c.logger.Debugf("Pull failed, going to try again in %v; error: %s", pullRetryTicker, err)
+		c.logger.Debugf("Pull failed, going to try again in %v; error: %s", c.retryInterval, err)
 
 		select {
 		case <-c.stopChan:
 			c.logger.Debug("Received a stop signal")
 			break runLoop
 
-		case <-pullRetryTicker.C:
+		case <-time.After(c.retryInterval):
```
yacovm: why the change? the previous version is more testable no?

orderer/consensus/follower/follower_chain.go
```diff
@@ -199,31 +235,69 @@ func (c *Chain) StatusReport() (types.ClusterRelation, types.Status) {
...
+		}
 	}
 
-	pullRetryTicker.Stop()
 	close(c.doneChan)
 	c.logger.Debug("The follower.Chain puller goroutine is exiting")
 }
 
+func (c *Chain) increaseRetryInterval() {
+	if c.retryInterval == c.options.PullRetryMaxInterval {
```
yacovm, line 293: perhaps it should be >= ?

orderer/consensus/follower/follower_chain.go
```diff
@@ -257,14 +331,92 @@ func (c *Chain) pull() error {
...
+	if err != nil {
+		c.logger.Errorf("Error creating block puller: %s", err)
+		return errors.Wrap(err, "error creating block puller")
+	}
+	defer puller.Close()
+
+	var nextBlock *common.Block
+	var actualPrevHash []byte
+
+	// Pull the first block
```
yacovm: I don't get it, what's so special about the 1st block? why can't we pull it along the rest below?

orderer/consensus/follower/follower_chain.go
```diff
@@ -257,14 +331,92 @@ func (c *Chain) pull() error {
...
+			c.logger.Warnf("Stopped before pulling all the blocks: pulled blocks from %d until %d, target was %d",
+				firstBlockToPull, nextBlock.Header.Number, targetHeight)
+			return errors.New("stopped before pulling all the blocks")
+		default:
+			nextBlock = puller.PullBlock(seq)
+			if nextBlock == nil {
+				return errors.Wrapf(cluster.ErrRetryCountExhausted, "failed to pull block %d", seq)
+			}
+			reportedPrevHash := nextBlock.Header.PreviousHash
+			if !bytes.Equal(reportedPrevHash, actualPrevHash) {
```
yacovm: I thought the block puller verifies the hash chain on its own? Why do you also do it here?

orderer/consensus/follower/block_puller.go
tock-ibm: Yes, the only reason I am doing this is to avoid verifying the (missing) signature on block 0. 
You are correct - there is a problem here, we need to verify the data hash of block 0.
Any suggestions? 
Maybe just add verification code that applies to block 0?

orderer/consensus/follower/follower_chain.go
tock-ibm: A follower chain is started only once and stopped only once. Subsequent starts have no effect. I do not expect `start` to be called more than once in correct code. However,  I am protecting against it nevertheless. 

The `IsRunning` method is used only for testing so far, and checks whether the internal go-routine is running.

Note that when a follower chain shuts down "internally", it can happen for two reasons: either the follower finished on -boarding (reached join-block), or it is following the cluster (after the join block), and discovers it was added to the cluster by a config block (see next PR). In either case it then calls the "CreateChain" callback, (function pointer to the registrar's create chain), which will create either a new follower.Chain or an etcdraft.Chain.

- The first scenario is when the follower chain is doing on boarding and then continues to follow the cluster;
- The second scenario is when the follower chain is either: (1) doing on boarding and then mutates into an etcdraft chain, or (2) following the cluster and then add to the consenters set.

Note also that it is important to recreate the follower.Chain after on-boarding is finished, in order to recreate "support" from the tip of the ledger.

orderer/consensus/follower/follower_chain.go
tock-ibm: The change is in order to be able to reset the timeout after a good block is received and not get stuck in a long retry interval after a sequence of failures that eventually ends.

orderer/consensus/follower/follower_chain.go
tock-ibm: If it is `>` I want to reset it to the max... only if it had reached the max in a previous round I want to return early.

orderer/consensus/follower/follower_chain.go
tock-ibm: We need to compare the prevHash of the first block to the header hash of the previous block from the ledger, that is all.

orderer/consensus/follower/follower_chain.go
tock-ibm: The block puller verifies only batches, there is no continuity between them. I am following the pattern from here, which does the same:

https://github.com/hyperledger/fabric/blob/5362a5701d0118629d6b974327555e4769e4232b/orderer/common/cluster/replication.go#L212

orderer/consensus/follower/block_puller.go
tock-ibm: The hash pointer on block 1 will be validated, though, since in the code that calls the puller, we compute the actual header hash of the previous block (0), and compare it to the expected hash in the header of the current block (1).

orderer/consensus/follower/follower_chain.go
yacovm: But we verify the signature, don't we? So how can the signature be good, but the hash not be good? 

The pattern you listed is to pull the system channel, which has no signature validation at all...

orderer/consensus/follower/block_puller.go
yacovm: Nevermind, forget it. I thought that we can skip signature validation for genesis in another way.

orderer/consensus/follower/follower_chain.go
yacovm: Can we at least extract the time.After function to the options so it'll be injectable and more testable?

orderer/consensus/follower/follower_chain.go
yacovm: Why? it seems to me you are pulling and checking your ledger is consistent with what you're going to pull next, i don't understand why you're doing this for the 1st block? 

The same problem exists in the next block since you can pull it from a different orderer

orderer/consensus/follower/follower_chain.go
tock-ibm: Actually... this internal function is used by 
https://github.com/hyperledger/fabric/blob/84396aaa4164a0d992b3b3e61e55b6ad7a9878c8/orderer/common/cluster/replication.go#L182

which pulls application channels as well as the system channel. I addition, verification is orthogonal (pluggable) and the type of verifier normally used
https://github.com/hyperledger/fabric/blob/84396aaa4164a0d992b3b3e61e55b6ad7a9878c8/orderer/common/cluster/util.go#L202

does not verify all the signatures.

I think the long term solution is to define a "cluster.StrictVerifyBlocks()" that does everything we need. However, I would like to do this under a separate story, after the dust settles on the main join story.

orderer/consensus/follower/follower_chain.go
tock-ibm: I think I understand what you mean. Basically I want to initialize the actualPrevHash before the main loop. So I only need to either:
- pull the last block from the ledger and compute its hash, or
- pull the genesis block, compute its hash, and append it.
There is no need to pull the "next" block as I do there. I will correct that.

orderer/consensus/follower/follower_chain.go
tock-ibm: Yes, I'll do that, and add a unit test as well.

orderer/consensus/follower/follower_chain.go
```diff
@@ -16,31 +18,49 @@ import (
...
+	Cert                 []byte
 }
 
 // CreateBlockPuller is a function to create BlockPuller on demand.
 // It is passed into chain initializer so that tests could mock this.
 type CreateBlockPuller func() (ChannelPuller, error)
 
+//go:generate counterfeiter -o mocks/puller_creator.go -fake-name BlockPullerCreator . BlockPullerCreator
+type BlockPullerCreator interface {
+	CreateBlockPuller() (ChannelPuller, error)
```
jyellick, line 57: Somewhat related to the comment I added in #1536, but it seems like we're introducing additional indirection in order to escape the consensus plugin box, and access resources in the rest of the orderer codebase that would not otherwise be accessible inside of a consensus plugin.

In general, I'd love to see this sort of indirection removed, have the block pulling resources created directly as real types, with a clear lifecycle and dependencies.  My expectation would be that the bulk of the logic in this PR could be preserved, but if we pulled it out of a consensus plugin, we could make a lot of the wiring much simpler -- not only for the new follower logic, but also for the inactive logic.

orderer/consensus/follower/follower_chain.go
tock-ibm: Done

orderer/consensus/follower/follower_chain.go
tock-ibm: Done

orderer/consensus/follower/follower_chain.go
tock-ibm: Indeed following a channel is largely independent of consensus type, in the same way that the peer pulls blocks from endpoints. The dependency comes in when a follower needs to decide whether it is a member of the cluster. That is where the particular consensus implementation is needed. The cluster package solves that problem by indirectly receiving at creation a function pointer to a "AmIInChannel" method that is actually implemented by the etcdraft package. I also agree that casting the follower (and the inactive.Chain) as a chain is pretty artificial as it does not implement the important methods of the interface (Order, Configure). 

An alternative to the current composition would be to have a "follower" as a separate entity managed by the registrar, something that is not a consensus.Chain (atually - even better, not a multichannel.ChainSupport). It would reside in a package at the top level, like multichannel, etc. We would then have two maps in the registrar - chains (i.e. ChainSupport) and followers, mutually exclusive. A follower would know how to trigger the creation of a chain, and vise versa. Does that capture your intention?

jyellick: Yes, I think this is pretty close.  I would think we basically want to group our chains into three buckets.

1. Chains that this orderer is a member of the consenter set and is actively consenting on.
2. Chains that this orderer is not a member of the consenter set, but is tracking but might become a member of in the future. (inactive)
3. Chains that this orderer is not a member of the consenter set, but is actively replicating. (follower)

So, we will need a way for a consensus plugin to report, based on channel config, whether it is a member of the consenter set (the cluster service already does this via a function pointer, you say), but we could directly add it to the consenter interface if we wanted.

We will also need a way for an active consensus plugin to return control to the multichannel code indicating "I'm no longer active" and the multichannel code will need to decide whether the chain becomes inactive, or follower.  (I believe based on the RFC, we want this to be follower now?  Or perhaps that is only for the system-channel-less case?)

I would also think, that this could be a good place to fix the problem associated with state transfer and node removal, where the node is removed from the consenter set, and re-added say while offline.  If we coordinate the block puller, it could detect the removal, and stop the consensus plugin before completing the replication and restarting it.

tock-ibm: Note that the "I am no longer active" part is implemented by the etcdraft.Chain calling a "chain-creation" call-back, that is function pointer to the multichannel registrar "CreateChain" method. The same happens in the soon-to-be follower (when it wants to activate the etcdraft.Chain)  and the inactive-registry. So that is not a big issue.

Note that the "follower" bucket and the "inactive" bucket are not active at the same time, as we do not allow the system to operate in a mixed mode. The "follower"-bucket will be for the system-channel-less (ch. part. API)  mode, and the "inactive"-bucket for the "legacy" system-channel mode. 

My concern/dilema is a bit pragmatic. I am hesitant to make these two "open-heart" operations at once. I rather complete the follower work along these lines first, not touching the code paths that run in the system channel model, and only then overhaul the "system channel" paths to the same architectural pattern. There are a few reasons for that. First, caution: I don't want to break the old before the new is in place... too many changes at the same time I think. Second, as we move to the "new" composition, we might want to rethink how a "tracker" of the system channel and inactive chains work, possibly reusing some of the pieces we build for the follower.

jyellick: > My concern/dilema is a bit pragmatic. I am hesitant to make these two "open-heart" operations at once. I rather complete the follower work along these lines first, not touching the code paths that run in the system channel model, and only then overhaul the "system channel" paths to the same architectural pattern. There are a few reasons for that. First, caution: I don't want to break the old before the new is in place... too many changes at the same time I think. Second, as we move to the "new" composition, we might want to rethink how a "tracker" of the system channel and inactive chains work, possibly reusing some of the pieces we build for the follower.

I would say no need to do them "at once".  First, I would modify the 'inactive' concept to reside outside of the consensus plugin, and, once that's set and working (should be fairly easy, we've already got integration tests around the operations you'll want to test), then move on to the follower code.

I realize that this gets you to the participation API less quickly, but it means eliminating, rather than stacking debt.

To me, it comes down to a fairly simple question: Do you think having a follower, and inactive consensus plugin is the right eventual model for managing channel participation lifecycle?  If the answer is 'no', then why would we build it this way?  If the answer is 'yes', then let's discuss why.

tock-ibm: I would say that long term, considering only the channel participation API (no system channel), there is no need for the "inactive" concept. We only need a follower. I also agree that it does not belong to a consensus plugin, but rather at a higher level. I think we are on the same page on this. 

When I am considering the system channel mode, I think we need 3 concepts: a **chain** which runs consensus, a **tracker** which follows the last config block of every channel using the system channel, and an **on-boarding** component. When the tracker discovers a channel that the orderer was added to, it triggers the on-boarding component to complete the gap between the last block on the ledger, and the last config block that indicates membership. When on boarding is done - the consesus.Chain is started. 

Note that the latter two concepts are implemented by the InactiveChainRegistry in the cluster package. The inactive.Chain is just a place holder which does nothing. Its only use is to keep an entry in the registrar.chains map. I don't like this approach at all. Moreover, both tracking and on-boarding are implemented by a single go-routine, which is not a good idea at all.

Note that the "on-boarding" part of the story could be executed by a follower component - this is exactly what the current PR is about. So, the way I see it, getting the building blocks of the channel participation API in place will allow us to upgrade the implementation of the "system-channel" tracking & on-boarding more easily. I don't see it as stacking debt, because implementing the follower correctly in the right place will not create debt, but rather pave the way for the upgrade of the "system-channel" tracking & on-boarding.

tock-ibm: > I would say no need to do them "at once". First, I would modify the 'inactive' concept to reside outside of the consensus plugin, and, once that's set and working (should be fairly easy, we've already got integration tests around the operations you'll want to test), then move on to the follower code.

Just to clarify a bit more: moving the inactive.Chain out of the consensus package into separate a "container" or a map in the registrar is the easy part. The real complications in wiring and operations are in the InactiveChainRegistry and the go-routine that it runs.

tock-ibm:
Closed due to design change.

orderer/consensus/follower/follower_chain.go
tock-ibm: Opened https://jira.hyperledger.org/browse/FAB-18104 for this issue

PR #1537, opened 2020-07-05
 FAB-17914 Ch.Part.API: follower pulls blocks after join-block

tock-ibm:
#### Type of change

- New feature

#### Description

The follower.Chain pulls blocks after the join block and  invokes the ChannelCreate function when a config that indicates membership is found.

#### Related issues

Task:  FAB-17914
Epic: FAB-17712


orderer/consensus/etcdraft/consenter.go
```diff
@@ -160,13 +160,30 @@ func (c *Consenter) HandleChain(support consensus.ConsenterSupport, metadata *co
...
 			return &inactive.Chain{Err: errors.Errorf("channel %s is not serviced by me", support.ChannelID())}, nil
 		} else {
+			// There is no system channel, follow this application channel.
 			//TODO fully construct a follower chain
-			return &follower.Chain{Err: errors.Errorf("orderer is a follower of channel %s", support.ChannelID())}, nil
+			consenterCertificate := &ConsenterCertificate{
+				ConsenterCertificate: c.Cert,
+				CryptoProvider:       c.BCCSP,
+			}
+			return follower.NewChain(
```
jyellick: I'm a little concerned with the approach of returning yet another implementation of consensus.Chain here.

I would have though the much cleaner approach would be to remove the inactive.Chain, not implement a follower.Chain, and instead push all this logic back into the common/multichannel package.

We had that bug, where, if you were in an etcdraft consenter set, then removed, and then added back, that etcdraft did not appropriately pick this up (because we were never starting the inactive chain and its assorted tracking).

It seems like we are doing some backflips here to inject all sorts of logic to support the block puller, and we will still have the challenge of switching from one chain implementation to another, as we're forced to do it inside the individual consensus plugin, rather than at the top level.

Why not take all of the chain tracking logic, push it out of the consensus plugin, and then let this top level logic manage starting and stopping consensus implementations as needed, as well as handling the existing inactive chain and new follower logic.

tock-ibm: See here: https://github.com/hyperledger/fabric/pull/1536#discussion_r451504634

yacovm: > We had that bug, where, if you were in an etcdraft consenter set, then removed, and then added back, that etcdraft did not appropriately pick this up (because we were never starting the inactive chain and its assorted tracking).

That is not a bug Jason. The user is using Fabric wrong. When you remove you need to make sure the orderer detects the removal and then halts the chain. If you do it too quickly, then you get a new ID and everything goes haywire because you do not know it.

 
> Why not take all of the chain tracking logic, push it out of the consensus plugin, and then let this top level logic manage starting and stopping consensus implementations as needed, as well as handling the existing inactive chain and new follower logic.

I like this idea in general, but:

- It needs to be aware of the consensus types, and to be able to classify them. So far, the "everything consensus related is under orderer/consensus" has been proven to not work well. 
- We really don't have any consensus but Raft at this point. Mir, which significantly changes Fabric fundamental assumptions, will probably make Fabric into something it is not. So, is there any point to this package layout anymore? 
- If we are to do this, then we need to do it for the existing Raft inactive chain and this raises the question- do we need to do it now? Can we not do it later after we have the functionality?

jyellick: > That is not a bug Jason. The user is using Fabric wrong. When you remove you need to make sure the orderer detects the removal and then halts the chain. If you do it too quickly, then you get a new ID and everything goes haywire because you do not know it.

You and I are referring to different 'bugs' Yacov.  I know you feel strongly that adding a Raft consenter to a channel, removing it, and re-adding it is not a valid use case.  But it is one we've seen hit by many users who are attempting to learn how to manage a Raft cluster.

The bug I am referring to was FAB-13552 fixed in https://github.com/hyperledger/fabric/commit/cce22474d79dd6cd59955ae6990f94a65c914a35  / https://github.com/hyperledger/fabric/commit/eb65b8ec458cc451632b9075c90226ef3e65cf79

> It needs to be aware of the consensus types, and to be able to classify them. So far, the "everything consensus related is under orderer/consensus" has been proven to not work well.

There's nothing to say we can't extend the consensus interface to require consensus plugins to be able to report their own membership in a configuration.  In fact, we already have something similar.

> We really don't have any consensus but Raft at this point. Mir, which significantly changes Fabric fundamental assumptions, will probably make Fabric into something it is not. So, is there any point to this package layout anymore?

Regardless of the future of Fabric, the idea of trying to internally swap between consensus types by having the consensus plugin stop and replace itself with another plugin implementation seems clunky and unreliable to me.  The fact that we've encountered bugs around leaving and rejoining a channel reinforces this clunkiness.  The basic problem is that in Fabric v1.0.0-v1.4.0 there was no notion of actually stopping a consensus plugin, so the only lifecycle for a consensus was to start a chain, and this was inadequate for Raft.  At the time, we dodged this by implementing the inactive consensus type, so that we could 'start' a mostly empy fake consensus type, and eventually hand it over to Raft.  But, I think the 'right way' to have done this would have been to implement a way for a started consensus plugin to indicate "I'm done" or "I'm not interested" and hand control back to the upper layer who can make more informed decisions.

> If we are to do this, then we need to do it for the existing Raft inactive chain and this raises the question- do we need to do it now? Can we not do it later after we have the functionality?

If we were not implementing the follower type chain, then I wouldn't say changing the inactive chain is a high priority.  But, implementing the follower type chain is extending a problematic pattern and adding more debt.  We've gotten ourselves into lots of trouble in the past by extending debt, rather than fixing the underlying problem.  It would simply mean that when we finally do need to implement a real chain lifecycle, we have both inactive chain and follower chain to reimplement, rather than just inactive.

yacovm: [FAB-13552](https://jira.hyperledger.org/browse/FAB-13552) is not really a bug, as I opened an issue during development in the beginning of 2019 and it was only addressed near the end of 2019 after the release. So, we have known about the issue and that wasn't a surprise, hence I'm not sure that it can be classified as a bug.

> There's nothing to say we can't extend the consensus interface to require consensus plugins to be able to report their own membership in a configuration. In fact, we already have something similar.

And I am saying this approach is overly complex when the only consensus you have is Raft. 

> Regardless of the future of Fabric, the idea of trying to internally swap between consensus types by having the consensus plugin stop and replace itself with another plugin implementation seems clunky and unreliable to me.


> But, I think the 'right way' to have done this would have been to implement a way for a started consensus plugin to indicate "I'm done" or "I'm not interested" and hand control back to the upper layer who can make more informed decisions.


> The fact that we've encountered bugs around leaving and rejoining a channel reinforces this clunkiness.
Even assuming they are indeed bugs - I am not sure why is having the code in the multi-channel package makes it inherently less error prone and more stable? 

I am missing something here. How would that work thread-model wise? Currently, the consensus plugin understands it needs to terminate right after it committed a block. Then, it calls a function which is above the consensus.

How would the upper layer learn to do the informed decisions if not via a synchronous method invocation from within the consensus plugin? If it is still the case then how is it different than what we're doing today? 
We terminate ourselves and then add it to the chain tracker which is a layer above.

jyellick: > FAB-13552 is not really a bug, as I opened an issue during development in the beginning of 2019 and it was only addressed near the end of 2019 after the release. So, we have known about the issue and that wasn't a surprise, hence I'm not sure that it can be classified as a bug.

Sure, we can call it whatever we want.  But, it was behavior that was not desirable, that we chose to change.

> Even assuming they are indeed bugs - I am not sure why is having the code in the multi-channel package makes it inherently less error prone and more stable?

I guess the simple argument would be that a clear separation of responsibilities makes code less buggy.  Today, the multichannel package is responsible for:
* Initializing chains at startup, including instantiating their consensus implementation.
* Dealing with creating new chain resources when channels are created
* Giving access to the consensus plugin to the dependent services (Deliver, Broadcast).

Because we've hacked into the consensus plugin "track inactive chains", we end up with some weird side effects.  For instance, when a chain is 'inactive' on an orderer, the multichannel package needs to fulfill the last bullet there and we need to tell Deliver and Broadcast that we can't perform operations on that channel.

But from the multichannel's packages' perspective, the channel is there, so as a hack, instead, the 'inactive' consenter always returns a consenter error, which is then bubbled back to the Broadcast and or Deliver services.  These services then in turn return a 'SERVICE_UNAVAILABLE'.  The 'SERVICE_UNAVAILABLE' errors are supposed to be transient, but in the event that this consenter is not a member of the consenter set, then a more appropriate error would be 'NOT_FOUND'.

There are other weird side affects as well, we open ledgers for inactive chains, which we don't need to do.  (Which can do things like generate metrics which don't make sense).

There are lots of other weird incongruencies, like the fact that the inactive "consensus plugin" performs no consensus, and returns static errors or has empty implementations.

As to whether it's less error prone, I think simply things that are easier to reason about, are less likely to have missed corner cases.

> I am missing something here. How would that work thread-model wise? Currently, the consensus plugin understands it needs to terminate right after it committed a block. Then, it calls a function which is above the consensus.

There are a number of ways we could address this.  I would still suggest that in the case that the consensus plugin is active, it would still be the one to surrender control and exit.  And, once the consensus plugin has terminated, we could then perform the related channel lifecycle operation.

> How would the upper layer learn to do the informed decisions if not via a synchronous method invocation from within the consensus plugin? If it is still the case then how is it different than what we're doing today?
We terminate ourselves and then add it to the chain tracker which is a layer above.

I don't see why it needs to be synchronous? The consensus plugin terminates, the upper layer detects that termination, and then performs its next steps.  In the case of the system channel, it would move this chain to the inactive registry, and begin tracking its config changes.  When the consenter is added back, it simply removes it from the inactive registry, and starts a new consensus instance.  In the case of the non-system channel, it would move this chain to the follow registry, keep replicating it, and then either start the consensus plugin back up once it's in the consenter set, or, wait for the participation API to instruct it to stop following and or delete.

yacovm: > I don't see why it needs to be synchronous? The consensus plugin terminates, the upper layer detects that termination, and then performs its next steps. 

If it detects via the consensus plugin telling it, then it's synchronous.
Else, it needs to poll the consensus plugin periodically, and we don't have such a thing, unless you somehow connect the inactive chain tracker to the consensus plugin only to poll it to discover it stopped.

jyellick: > If it detects via the consensus plugin telling it, then it's synchronous.

Well, you could potentially poll on a channel as well, I don't see how the consensus plugin signaling inherently makes things synchronous or asynchronous.  If the multichannel go routine is off interacting with the participation API, for instance, it could return later and find a consensus plugin stopped.  But I'm really not sure what this synchronous/asynchronous conversation is driving at.

tock-ibm:
Closed due to a design change.

PR #1538, opened 2020-07-06
Add org name to consortium error message

denyeart:
Improve troubleshooting by including org name in the error message:
"Attempted to include a member which is not in the consortium"

New message:
"Attempted to include member BadOrgName which is not in the consortium"

This will help users that have an incorrect config transaction pinpoint the reason for failure.


caod123 (approved):

denyeart:

PR #1540, opened 2020-07-06
Channel Participation API: Orderer with system channel rejects application channel removal

wlahti:
#### Type of change

- New feature

#### Description

When a system channel is defined, the channel participation API should reject requests to remove application channels.

#### Related issues

[FAB-18011](https://jira.hyperledger.org/browse/FAB-18011)

### DIFF
--- a/orderer/common/multichannel/registrar.go
+++ b/orderer/common/multichannel/registrar.go
@@ -523,6 +523,8 @@ func (r *Registrar) joinAsFollower(ledgerRes *ledgerResources, clusterConsenter 
 // RemoveChannel instructs the orderer to remove a channel.
 // Depending on the removeStorage parameter, the storage resources are either removed or archived.
 func (r *Registrar) RemoveChannel(channelID string, removeStorage bool) error {
-	//TODO
+	if r.SystemChannelID() != "" && channelID != r.SystemChannelID() {
+		return types.ErrSystemChannelExists
+	}
 	return errors.New("Not implemented yet")
 }
--- a/orderer/common/multichannel/registrar_test.go
+++ b/orderer/common/multichannel/registrar_test.go
@@ -669,6 +669,31 @@ func TestRegistrar_JoinChannel(t *testing.T) {
 	})
 }
 
+func TestRegistrar_RemoveChannel(t *testing.T) {
+	// system channel
+	confSys := genesisconfig.Load(genesisconfig.SampleInsecureSoloProfile, configtest.GetDevConfigDir())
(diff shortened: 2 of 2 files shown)


orderer/common/multichannel/registrar_test.go
```diff
@@ -647,6 +647,29 @@ func TestRegistrar_JoinChannel(t *testing.T) {
 	})
 }
 
+func TestRegistrar_RemoveChannel(t *testing.T) {
+	// system channel
+	confSys := genesisconfig.Load(genesisconfig.SampleInsecureSoloProfile, configtest.GetDevConfigDir())
+	genesisBlockSys := encoder.New(confSys).GenesisBlockForChannel("sys-channel")
+
+	cryptoProvider, err := sw.NewDefaultSecurityLevelWithKeystore(sw.NewDummyKeyStore())
+	assert.NoError(t, err)
```
caod123: There's been a push to standardize using `require` over `assert` in some of the related code paths. I know the rest of this file still mixes both but we may want to standardize starting here.

caod123 (commented):
nit on usage of `require` over `assert`

orderer/common/multichannel/registrar.go
```diff
@@ -495,6 +495,8 @@ func (r *Registrar) JoinChannel(channelID string, configBlock *cb.Block, isAppCh
 // RemoveChannel instructs the orderer to remove a channel.
 // Depending on the removeStorage parameter, the storage resources are either removed or archived.
 func (r *Registrar) RemoveChannel(channelID string, removeStorage bool) error {
-	//TODO
+	if r.SystemChannelID() != "" && channelID != r.SystemChannelID() {
+		return errors.Errorf("cannot remove application channel %s because the system channel exists", channelID)
+	}
```
tock-ibm: The REST API checks for 2 explicit error codes:
```go
	case types.ErrSystemChannelExists:
		h.sendResponseNotAllowed(resp, errors.Wrap(err, "cannot remove"), http.MethodGet)
	case types.ErrChannelNotExist:
		h.sendResponseJsonError(resp, http.StatusNotFound, errors.Wrap(err, "cannot remove"))
```
see [link](https://github.com/hyperledger/fabric/blob/f27912f2f419c3b35d2c1df120f19585815eceb0/orderer/common/channelparticipation/restapi.go#L270).

So you need to return a specific error in order for the REST response error code & headers to be correct.

wlahti: Thanks for pointing that out. Not sure how I missed it. :) Done.

yacovm (approved):

PR #1550, opened 2020-07-07
[FAB-18054] Remove default value for top level orderer addresses if unset in config

caod123:
#### Type of change

- Bug fix

#### Description

Top level orderer addresses are deprecated in 2.2 so this commit
removes the default value for consumers that pass a config with no
orderer addresses value

#### Related issues

[FAB-18054](https://jira.hyperledger.org/browse/FAB-18054)

### DIFF
--- a/internal/peer/chaincode/common_test.go
+++ b/internal/peer/chaincode/common_test.go
@@ -175,8 +175,7 @@ func TestGetOrdererEndpointFromConfigTx(t *testing.T) {
 	ordererEndpoints, err := common.GetOrdererEndpointOfChain(mockchain, signer, mockEndorserClient, cryptoProvider)
 	assert.NoError(t, err, "GetOrdererEndpointOfChain from genesis block")
 
-	assert.Equal(t, len(ordererEndpoints), 1)
-	assert.Equal(t, ordererEndpoints[0], "127.0.0.1:7050")
+	assert.Equal(t, len(ordererEndpoints), 0)
 }
 
 func TestGetOrdererEndpointFail(t *testing.T) {
--- a/common/channelconfig/realconfig_test.go
+++ b/common/channelconfig/realconfig_test.go
@@ -33,6 +33,7 @@ func TestWithRealConfigtx(t *testing.T) {
 func TestOrgSpecificOrdererEndpoints(t *testing.T) {
 	t.Run("Without_Capability", func(t *testing.T) {
 		conf := genesisconfig.Load(genesisconfig.SampleDevModeSoloProfile, configtest.GetDevConfigDir())
+		conf.Orderer.Addresses = []string{"127.0.0.1:7050"}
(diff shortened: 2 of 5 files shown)


caod123:
Turns out a number of tests were relying on the default because they weren't actually passing a value for the orderer addresses when it was still required.

internal/peer/chaincode/common_test.go
```diff
@@ -175,8 +175,7 @@ func TestGetOrdererEndpointFromConfigTx(t *testing.T) {
 	ordererEndpoints, err := common.GetOrdererEndpointOfChain(mockchain, signer, mockEndorserClient, cryptoProvider)
 	assert.NoError(t, err, "GetOrdererEndpointOfChain from genesis block")
 
-	assert.Equal(t, len(ordererEndpoints), 1)
-	assert.Equal(t, ordererEndpoints[0], "127.0.0.1:7050")
+	assert.Equal(t, len(ordererEndpoints), 0)
```
caod123, line 178: This function is currently broken and will be addressed in [FAB-18055](https://jira.hyperledger.org/browse/FAB-18055)

denyeart:
This looks good to me. I've tested e2e and scanned if there are other areas of the code that may be impacted.
I did find one place where peer CLI uses the deprecated Orderer.Addresses. That should probably be changed to use the newer org level OrdererEndpoint at some point, but not urgent. Opened https://jira.hyperledger.org/browse/FAB-18056

caod123:
> This looks good to me. I've tested e2e and scanned if there are other areas of the code that may be impacted.
> I did find one place where peer CLI uses the deprecated Orderer.Addresses. That should probably be changed to use the newer org level OrdererEndpoint at some point, but not urgent. Opened https://jira.hyperledger.org/browse/FAB-18056

This is actually what I opened [FAB-18055](https://jira.hyperledger.org/browse/FAB-18055) for

denyeart:
> > This looks good to me. I've tested e2e and scanned if there are other areas of the code that may be impacted.
> > I did find one place where peer CLI uses the deprecated Orderer.Addresses. That should probably be changed to use the newer org level OrdererEndpoint at some point, but not urgent. Opened https://jira.hyperledger.org/browse/FAB-18056
> 
> This is actually what I opened [FAB-18055](https://jira.hyperledger.org/browse/FAB-18055) for

Cool, we both found the same thing... I've closed FAB-18056 as duplicate.

caod123:

denyeart (approved):
Approving, but will let Jason take a final look...

internal/peer/chaincode/common_test.go
jyellick: Would be nicer to actually have this in the commit rather than GH, but, not the end of the world.

jyellick (approved):

PR #1551, opened 2020-07-07
Remove {inmem,dummy}keystore config from bccsp

sykesm:
Explicitly configuring a Dummy key store or an In-Memory key store will not result in a working configuration. Remove the ability to configure them and remove the unused in-memory key implementation.


lindluni (approved):
I've looked at this a million times and never put a single thought into what it actually was supposed to do.

PR #1552, opened 2020-07-07
[FAB-18053] Make values for ACL policies in sample/test configs consi…

satota2:
…stent with default values in source

#### Type of change

- Documentation update

#### Description

Values for some ACL polices used in sample/test configurations are different from default values in source code (`core/aclmgmt/defaultaclprovider.go`).

For example, the following gaps from source code can be found in sampleconfig/configtx.yaml:
- _lifecycle/QueryChaincodeDefinition: /Channel/Application/Readers
- _lifecycle/QueryChaincodeDefinitions: /Channel/Application/Readers
- peer/ChaincodeToChaincode: /Channel/Application/Readers

In `core/aclmgmt/defaultaclprovider.go`, they are set to `/Channel/Application/Writers` as the default values.

Although the values in the sample configuration may be just examples,  Fabric users may expect them as default values as there is no documentation on ACL default values.

To prevent misunderstandings, if there is no specific intention, I think that it is desirable to have consistent default values.

So, this patch makes the values used in samples consistent with default values in source code.

#### Related issues

- https://jira.hyperledger.org/browse/FAB-18053

### DIFF
--- a/docs/source/enable_cc_lifecycle.md
+++ b/docs/source/enable_cc_lifecycle.md
@@ -116,10 +116,10 @@ Note that the `enable_lifecycle.json` uses sample values, for example `org1Polic
 			"policy_ref": "/Channel/Application/Writers"
 		},
 		"_lifecycle/QueryChaincodeDefinition": {
-			"policy_ref": "/Channel/Application/Readers"
+			"policy_ref": "/Channel/Application/Writers"
 		},
 		"_lifecycle/QueryChaincodeDefinitions": {
-			"policy_ref": "/Channel/Application/Readers"
+			"policy_ref": "/Channel/Application/Writers"
 		}
    }
 }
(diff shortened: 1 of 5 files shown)


denali49 (approved):
lgtm, thank you for your contribution! - APPROVAL FOR DOC REVIEW ONLY.

jyellick (approved):
I'm a bit hesitant to change sample values, which, are in effect default values for many, if not most users.  At the same time, because the propose API requires 'Writer' privileges, and it's very unusual for an identity to be a Writer, but not a Reader, almost all ACLs effectively degenerate to 'Writer'.  With that in mind, since this does improve consistency, and is unlikely to manifest as a change for that vast majority of deployments, I think we're okay to merge this.

WaleedMortaja:
@denyeart @jyellick Should this be backported to release-2.2?

denyeart:
I don't think it needs to be back-ported.

PR #1555, opened 2020-07-08
Improve peer CLI error message when no orderer address passed

denyeart:
For peer CLI commands that require an orderer address, the peer
command does best effort attempt to retrieve orderer address from channel
config if the orderer address is not passed (it currently only
checks the deprecated Orderer.Addresses, not the org level
OrdererEndpoints). The best effort attempt may be removed in
a future release, so for now improve the error message
to indicate that orderer address should be passed using
-o flag.


caod123 (approved):

denyeart:

PR #1535, opened 2020-07-03
[FAB-18046] IssuerRevocationPublicKey -> RevocationPublicKey

davidkhala:
IssuerRevocationPublicKey -> RevocationPublicKey

Update related idemix document

Signed-off-by: david <david-khala@hotmail.com>
Change-Id: Id9273e1243fa42194a8e7deb6e2ae5d5ea6abbff

#### Type of change
- Documentation update

#### Description

IssuerRevocationPublicKey -> RevocationPublicKey

Only the later is recognizable by configtxgen

#### Related issues

FAB-18046

PR #1557, opened 2020-07-08
Change RevocationPublicKey to IssuerRevocationPublicKey

lindluni:
The Fabric-CA implements this certificate as IssuerRevocationPublicKey.
This updates the name to reflect the actual file name, as it cannot be used
in its current state.

#### Type of change

- Bug fix
- New feature
- Test update
- Documentation update


lindluni:
@ale-linux @C0rWin @yacovm 

IdeMixer and the Fabric CA don't match up on this implementation. CA spits out a cert named `IssuerRevocationPublicKey` while idemix expects it to be `RevocationPublicKey`. We either need to update Fabric (this PR), or update the CA (and our doc per #1535)

lindluni:
@Mergifyio rebase

denyeart:
@adecaro Could you take a look?

mastersingh24:
Even though this would not really affect runtime operations at this point given we can't use idemix for local MSPs,  we should probably change this on the fabric ca side.  But I can be convinced to do it here as this is not widely used.

davidkhala:
？It is closed instead of merged because Gari prefer to do it on FabricCA side, is it? @mastersingh24 @btl5037

lindluni:
Nope. I accidentally deleted the branch while cleaning up my fabric repo. Will reopen

PR #1575, opened 2020-07-10
Print channel name in learnAnchorPeers

yacovm:
This change set makes all logging in learnAnchorPeers() print
also the channel name.

Change-Id: I83c4f47a183bc59bc237d779b487dcf1042eb03d


yacovm:
Ugh the code from 1.4 and 2.x diverges too much so i can't cherry pick it. Will take a different approach

mastersingh24 (approved):

PR #1578, opened 2020-07-10
Move everything from testify/assert to require

sykesm:
It’s time we leave `assert` behind and make a complete transition to `require`. I haven’t yet run into a situation where we explicitly want a test to proceed after an assertion failure.

- apply `gofmt` rewriting rules for assert -> require
- replace non-asserting `assert.ObjectsAreEqual` with alternative assertions
- update doc to call out `testify/require` instead of `testify/assert`


btl5037:
image

lindluni:
Wow. This was as aweful reviewing as it looks like it would be. But thank you for doing this. Will actually make debugging failures easier as they won't be wrapped in output of other tests.

lindluni (approved):

core/common/ccprovider/ccinfocache_test.go
```diff
@@ -241,8 +241,8 @@ func TestSetChaincodesPath(t *testing.T) {
 	defer SetChaincodesPath(cip)
 
 	f, err := ioutil.TempFile(dir, "chaincodes")
-	assert.NoError(t, err)
-	assert.Panics(t, func() {
+	require.NoError(t, err)
+	require.Panics(t, func() {
```
cendhu, line 245: Unrelated to this PR. Line 249 to 263 needs to be fixed in a separate PR.

PR #1586, opened 2020-07-10
Add length check for dirsDepth

scovetta:
This is a simple addition of a few bounds checks to ensure that the `dirsDepth` array is not accessed at an index beyond its size.

#### Type of change

#### Description

This change adds index checks before `dirDepths[collectionIndexDirDepth]` and `dirDepths[collectionNameDepth]` to avoid indexing beyond the end of the array. We *could* just check for the `collectionIndexDirDepth` and avoid the other two, but that would be more brittle in case the constants defined a few lines earlier were reordered. We could also just calculate `len(dirsDepth)` once as a small perf improvement.

#### Additional details

I have never used Hyperledger, and did not test this change in any way. I also don't know how likely it would be for `dirDepths` to short enough to cause the panic.

#### Related issues

This is related to the [suggestion](hyperledger#1080 (review)) from @manish-sethi in hyperledger#1080.

### DIFF
--- a/core/ledger/kvledger/txmgmt/privacyenabledstate/db.go
+++ b/core/ledger/kvledger/txmgmt/privacyenabledstate/db.go
@@ -426,7 +426,9 @@ func getIndexInfo(indexPath string) *indexInfo {
 		indexInfo.hasIndexForChaincode = true
 	case len(dirsDepth) > collectionDirDepth &&
 		dirsDepth[collectionDirDepth] == "collections" &&
-		dirsDepth[collectionIndexDirDepth] == "indexes":
+		len(dirsDepth) > collectionIndexDirDepth &&
+		dirsDepth[collectionIndexDirDepth] == "indexes" &&
+		len(dirsDepth) > collectionNameDepth:
 		indexInfo.hasIndexForCollection = true
 		indexInfo.collectionName = dirsDepth[collectionNameDepth]
 	}


core/ledger/kvledger/txmgmt/privacyenabledstate/db.go
```diff
@@ -426,7 +426,9 @@ func getIndexInfo(indexPath string) *indexInfo {
 		indexInfo.hasIndexForChaincode = true
 	case len(dirsDepth) > collectionDirDepth &&
 		dirsDepth[collectionDirDepth] == "collections" &&
-		dirsDepth[collectionIndexDirDepth] == "indexes":
+		len(dirsDepth) > collectionIndexDirDepth &&
+		dirsDepth[collectionIndexDirDepth] == "indexes" &&
+		len(dirsDepth) > collectionNameDepth:
```
cendhu, line 431: Instead, can you update the first condition in the line 427? `len(dirsDepth) > collectionDirDepth` to `len(dirsDepth) > collectionIndexDirDepth`

cendhu (changes requested):
Thank you for your contribution. I have made one comment. Other than that, please address the following too

- Update the unit-test at https://github.com/hyperledger/fabric/blob/89ab0bfcc2f66bd17475c9781cb784ef212de71f/core/ledger/kvledger/txmgmt/privacyenabledstate/db_test.go#L51 to ensure that this case is not missed in the future.
- Need to sign off the PR using `git commit --amend -s`

sykesm:
@scovetta We appreciate your time and the effort you've invested in opening this PR. Unfortunately it's been a month since we've heard from you, we're unable to merge commits without a DCO signoff, and there are pending comments so I'm going to close this out. If you find that you have some time to to revisit this in the future, please feel free to reopen.

Thanks.

PR #1588, opened 2020-07-10
Export additional info in the snapshot metadata

manish-sethi:

Export statedb-type and previous-block-hash in snapshot metadata

#### Type of change

### DIFF
--- a/core/ledger/kvledger/snapshot.go
+++ b/core/ledger/kvledger/snapshot.go
@@ -15,6 +15,7 @@ import (
 	"os"
 	"path/filepath"
 
+	"github.com/hyperledger/fabric/core/ledger"
 	"github.com/hyperledger/fabric/internal/fileutil"
 	"github.com/pkg/errors"
 )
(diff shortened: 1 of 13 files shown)


core/ledger/kvledger/snapshot_test.go
```diff
@@ -99,15 +117,42 @@ func TestGenerateSnapshot(t *testing.T) {
...
+func TestSnapshotDBTypeCouchDB(t *testing.T) {
+	conf, cleanup := testConfig(t)
+	defer cleanup()
+	provider := testutilNewProvider(conf, t, &mock.DeployedChaincodeInfoProvider{})
+	defer provider.Close()
+	lgr, err := provider.open("testLedger")
+	require.NoError(t, err)
+	kvlgr := lgr.(*kvLedger)
+
+	// artificially set the db type
```
cendhu, line 147: Maybe we should move the snapshot test to `kvledger/tests` so that we can run the test with both `GoLevelDB` and `CouchDB`. If we agree to do this, it can be done in a separate PR.

cendhu (commented):
LGTM, Manish, I have made two comments.

core/ledger/kvledger/snapshot.go
```diff
@@ -114,12 +118,19 @@ func (l *kvLedger) generateSnapshotMetadataFiles(
 	if err != nil {
 		return err
 	}
+
+	stateDBType := l.config.StateDBConfig.StateDatabase
+	if stateDBType != ledger.CouchDB {
+		stateDBType = simpleKeyValueDB
+	}
```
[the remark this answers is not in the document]
manish-sethi: - We could have been explicit, if we fix upfront when we parse the configs, to make sure that these two are the choices. Unfortunately, in the current code, the [check](https://github.com/hyperledger/fabric/blob/7fd80f54fd9fe6ee5f7460558bf884cbb2ba2996/core/ledger/kvledger/txmgmt/privacyenabledstate/db.go#L67) is like this only. Which means that any value (practically, default value) other than "CouchDB" for this string would result in leveldb as a statedb.
- Yes, it's nothing to do with the data format of export. The difference it important to make sure that we do not end up loading leveldb data into couchdb (which is more restrictive)
- As mentioned in point 1 above, exact value could very well be empty string on one peer while "goleveldb" being specified on the other (theoretically, it could be anything)

core/ledger/kvledger/snapshot_test.go
manish-sethi: Yes, I believe that we should retire all the tests in the `kvledger` package (they are just some form of duplicate) and have tests only at one place. I plan to move this into `kvledger/tests` when add function for bootstrap at this level. BTW, Ideally, sometime we should  think of moving `kvledger/tests` into `kvledger` folder but specify in `kvledger_test` that way they will serve both the purposes of external API testing and counting towards coverage.

core/ledger/kvledger/snapshot.go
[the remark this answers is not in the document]
cendhu: I see. Make sense. Because of the existing mechanism, this looks fine.

core/ledger/kvledger/snapshot_test.go
cendhu: Agree.

cendhu (approved):

PR #1590, opened 2020-07-13
Correct the explanation for signcerts in Membership section

manindra23:
Signed-off-by: Manindra Singh <manindra23@yahoo.co.in>

Correcting the explanation for signcert folder under MSP Structure in Membership section

#### Type of change

- Documentation update

#### Description
The explanation for signcert folder under MSP Structure is incorrect. Most of the explanation is for keystore folder i.e. about private key and not about public key which is part of signcert folder.


docs/source/membership/membership.md
```diff
@@ -172,11 +172,11 @@ A local MSP folder contains the following sub-folders:
 
   **Note:** If you are using a [Hardware Security Module(HSM)](../hsm.html) for key management, this folder is empty because the private key is generated by and stored in the HSM.
 
-* **signcert:** For a peer or orderer node (or in a client's local MSP) this folder contains the node's **signing key**. This key matches cryptographically the node's identity included in **Node Identity** folder and is used to sign data --- for example to sign a transaction proposal response, as part of the endorsement phase.
+* **signcert:**  For a peer or orderer node (or in a client’s local MSP) this folder contains the node’s certificate issued by CA. This certificate contains the node's **public key**. While private key is used to digitally sign a transaction, public key is used to verify that the signature attached to the transaction is valid.
```
denali49: See original explanation of the signature of the transaction proposal response, not just the blanket 'transaction'.

docs/source/membership/membership.md
```diff
@@ -172,11 +172,11 @@ A local MSP folder contains the following sub-folders:
...
 
-  This folder is mandatory for local MSPs, and must contain exactly one public key. Obviously, access to this folder must be limited only to the identities of users who have  administrative responsibility on the peer.
+  This folder is mandatory for local MSPs, and must contain exactly one public key. Obviously, access to this folder must be with all the members of a channel.
 
-  Configuration of a **channel MSP** does not include this folder, as channel MSPs solely aim to offer identity validation functionalities and not signing abilities.
+  Configuration of a **channel MSP** includes this folder, as channel MSPs offer identity validation functionalities.
```
denali49: See how channel MSP's are defined in the channel configuration (ref. configtx.yaml).

docs/source/membership/membership.md
denali49: Is used to sign a transaction proposal response for example as in the original explanation, not just a blanket 'transaction'.

docs/source/membership/membership.md
```diff
@@ -172,11 +172,11 @@ A local MSP folder contains the following sub-folders:
...
 
-* **signcert:** For a peer or orderer node (or in a client's local MSP) this folder contains the node's **signing key**. This key matches cryptographically the node's identity included in **Node Identity** folder and is used to sign data --- for example to sign a transaction proposal response, as part of the endorsement phase.
+* **signcert:**  For a peer or orderer node (or in a client’s local MSP) this folder contains the node’s certificate issued by CA. This certificate contains the node's **public key**. While private key is used to digitally sign a transaction, public key is used to verify that the signature attached to the transaction is valid.
 
-  This folder is mandatory for local MSPs, and must contain exactly one public key. Obviously, access to this folder must be limited only to the identities of users who have  administrative responsibility on the peer.
+  This folder is mandatory for local MSPs, and must contain exactly one public key. Obviously, access to this folder must be with all the members of a channel.
```
denali49: Why do you think access to this folder must be with all members of a channel and not limited to admin identities?

denali49 (commented):
Thank you for your efforts in contributing to make the Fabric docs better.  See inline comments.

docs/source/membership/membership.md
manindra23: I went through the original explanation of the signature of the transaction proposal response(https://hyperledger-fabric.readthedocs.io/en/latest/peers/peers.html?highlight=transaction proposal response#phase-1-proposal). It says "A peer endorses a proposal response by adding its digital signature, and signing the entire payload using its private key." As per my understanding(reference: Digital Signature definition at https://github.com/bitcoinbook/bitcoinbook/blob/develop/ch06.asciidoc), although digital signature uses both private and public keys, essentially in this scenario, private key is always used to sign and public key is always used to verify the signature. 
So, when the original text says "This key..and is used to sign data --- for example to sign a transaction proposal response, as part of the endorsement phase.", it seems incorrect since signcert folder contains public key and is meant to verify the signature and not sign the data.

Please let me know if my understanding is incorrect.

manindra23: I went through the original explanation of the signature of the transaction proposal response(https://hyperledger-fabric.readthedocs.io/en/latest/peers/peers.html?highlight=transaction proposal response#phase-1-proposal). It says "A peer endorses a proposal response by adding its digital signature, and signing the entire payload using its private key." As per my understanding(reference: Digital Signature definition at https://github.com/bitcoinbook/bitcoinbook/blob/develop/ch06.asciidoc), although digital signature uses both private and public keys, essentially in this scenario, private key is always used to sign and public key is always used to verify the signature. 
So, when the original text says "This key..and is used to sign data --- for example to sign a transaction proposal response, as part of the endorsement phase.", it seems incorrect since signcert folder contains public key and is meant to verify the signature and not sign the data.

Please let me know if my understanding is incorrect.

docs/source/membership/membership.md
manindra23: Checked it. I don't see the signcert folder in channel MSP in the given examples but I wrote this line based on the understanding that since Channel MSP aim to offer identity validation functionalities, public certificates of channel members would be needed. Please correct me if my understanding is incorrect.

docs/source/membership/membership.md
manindra23: Since every transaction within a block is validated at each peer during Phase 3 -Validation and Commit of Transaction Flow, each peer on the channel must have access to public key inside signcert folder to be able to verify the transactions.

Please correct me if my understanding is incorrect.

docs/source/membership/membership.md
jyellick: You are absolutely correct, in that the signercert folder does _not_ contain the private key, it contains a certificate, corresponding to the client's identity.

However, I fear that talking about public vs private keys is a little too technical here.  It is standard cryptography, that an X.509 certificate contains a public key, as well as information about signing algorithms, as well as a issuing information.  Certificates are considered to be 'public' information and may be shared, the private key is private information, corresponding to the certificate, which is secret, and is used to prove that a message purporting to be from the certificate owner is the certificate owner.

Personally, I would simply say that the certificate in the signcert directory represents the node's identity, and that that certificate's corresponding private key can be used to generate signatures which may be verified by anyone with a copy of that certificate.

jyellick: Yes, private keys are used to sign, and public keys are used to verify the signature, but I fear we are being a bit too specific here.

From a client, there is a signature over the proposal, there is a signature over the transaction. From the peer, there is a signature over the proposal response (the endorsement), as well as over requests to the orderer to retrieve blocks.  From the orderer there are signatures over the blocks themselves.

I also think the emphasis here should be that the certificate, which does embed a public key, is attached to node actions, like creating a transaction, or an endorsement, or a block, and it is this embedded certificate which is used to verify the signature.

docs/source/membership/membership.md
jyellick: This is incorrect.

Standard X.509 certificates operate based on a chain of trust.  Loosely speaking:

A node generates a public/private key pair, and creates a 'certificate signing request' (CSR) containing the public key and some other information.  It sends this request to the CA, who creates a certificate for the node, and signs it.  This certificate refer to the 'issuing CA', and includes the CA's signature.

When it comes time to validate a transaction (or any other action secured by a cryptographic signature), the verifier first extracts the certificate/identity from the message, then checks to see if the issuing CA is known to and trusted by the verifier.  If so, the verifier confirms that the signature from the issuing CA is valid.  At this point we know that the certificate itself is valid, then we can verify that the signature over the message is in fact from this certificate.

So, in general no, the signcert folder is not shared/known to the whole network.  Instead, the signcert certificate is generally embedded along with the signature, and then the chain of trust from the issuing CAs is used to validate the certificate's authenticity.

docs/source/membership/membership.md
jyellick: Channel configuration does not and should not include sign-certs.  Because the configuration for the local MSP and channel MSP shares the same data structure, you may erroneously see a signcert field in the channel configuration, but this should never be populated.

docs/source/membership/membership.md
manindra23: Understood. Yes, my sentence seems too technical. Your sentence seems much better. Thanks.

docs/source/membership/membership.md
manindra23: Got it. Thanks.

docs/source/membership/membership.md
manindra23: Agree with you. Thanks.

docs/source/membership/membership.md
manindra23: Understood. Thanks for the explanation.

manindra23:
@denali49 : What should be done now? Should I update the pull request with new commit by replacing my explanation with yours and correcting other sentences as per your comments or you would reject this pull request and update the explanation yourself ?

denali49:
> @denali49 : What should be done now? Should I update the pull request with new commit by replacing my explanation with yours and correcting other sentences as per your comments or you would reject this pull request and update the explanation yourself ?

Thank you again for contributing, please update your pull request based on the feedback in the comments.  If you are not familiar with the process for squashing your commits or amending, you may find the procedure here: https://hyperledger-fabric.readthedocs.io/en/latest/github/github.html
Scroll down to the Updating a Pull Request section and follow the procedure to either squash the commit, OR amend the previous commit.  Let me know if you have any difficulties and I will try to assist.

docs/source/membership/membership.md
manindra23: Sorry for late response. While re-checking this sentence, I got 2 more doubts:
(1)When above sentence says "...as channel MSPs solely aim to offer identity validation functionalities and not signing abilities.", it seems to imply that if signcert folder had been included, it would have provided signing abilities which seems incorrect. 

(2) How does channel MSP then perform identity validation? Will they extract certificate from the incoming message and do identify validation?

docs/source/membership/membership.md
```diff
@@ -172,9 +172,9 @@ A local MSP folder contains the following sub-folders:
 
   **Note:** If you are using a [Hardware Security Module(HSM)](../hsm.html) for key management, this folder is empty because the private key is generated by and stored in the HSM.
 
-* **signcert:** For a peer or orderer node (or in a client's local MSP) this folder contains the node's **signing key**. This key matches cryptographically the node's identity included in **Node Identity** folder and is used to sign data --- for example to sign a transaction proposal response, as part of the endorsement phase.
+* **signcert:**  For a peer or orderer node (or in a client’s local MSP) this folder contains the node’s certificate issued by CA. The certificate represents the node's identity, and that that certificate's corresponding **private key** can be used to generate signatures which may be verified by anyone with a copy of that certificate.
```
joealewine: The signcert folder does not contain the private key. That's what the keystore folder is for.

manindra23: Yes, agreed. The line "The certificate represents the node's identity, and that that certificate's corresponding **private key** can be.." only speaks about corresponding private key of the certificate in signcert folder; it does not tell that private key resides inside signcert folder.

manindra23:
@denali49, @jyellick, @joealewine : I have updated the pull request as per the review comments. Please check.

denyeart (approved):

denyeart:
Thanks, good doc fix.

denyeart:

PR #1595, opened 2020-07-13
check if desired version of samples exists

nasermirzaei89:
#### Type of change

#### Description

`bootstrap.sh` script uses `VERSION` of fabric repo in checkout fabric-samples. So, we need to check if it exists in fabric-samples.

### DIFF
--- a/scripts/bootstrap.sh
+++ b/scripts/bootstrap.sh
@@ -51,16 +51,23 @@ cloneSamplesRepo() {
     # version to the binaries and docker images to be downloaded
     if [ -d first-network ]; then
         # if we are in the fabric-samples repo, checkout corresponding version
-        echo "===> Checking out v${VERSION} of hyperledger/fabric-samples"
-        git checkout v${VERSION}
+        echo "==> Already in fabric-samples repo"
     elif [ -d fabric-samples ]; then
         # if fabric-samples repo already cloned and in current directory,
-        # cd fabric-samples and checkout corresponding version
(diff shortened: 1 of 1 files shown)


lindluni:
The correct behavior of this script when a non-existing version of fabric-samples is used is for the checkout to fail. This will cause the latest master branch to be used (which is the intended behavior). These versions should always line up except for when master should be used (which is the current case until we tag v2.2.0 in the next week or so)

If we are going to do anything we should do a `git ls-remote --tags origin refs/tags/v<version>` to check if the tag exists, and do a checkout if it exists, otherwise print a message that its `fabric-samples v<version> does not exist, defaulting master`

nasermirzaei89:
@btl5037 signed off

ale-linux:
@nasermirzaei89 could you pls address the comments from @btl5037? Or have they already been addressed to your satisfaction @btl5037?

lindluni (approved):

WaleedMortaja:
@lindluni @denyeart should this be backported to release-2.2?

denyeart:
@WaleedMortaja No need to backport since the download doc for all versions directs users to the bootstrap.sh script in main branch.
In general, when a PR is merged the maintainers consider whether it should be backported at that time. There is no need to identify and ask about every PR that is not backported since the maintainers have already considered the backport and made a decision. That being said, if you have a strong need for something to be backported to a certain release you are welcome to open a backport PR.

PR #1596, opened 2020-07-13
Fix latest tag on ccenv

lindluni:
`latest` no longer exists. Defaulting to the standard tags

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -62,6 +62,7 @@ BUILD_DIR ?= .build
 
 EXTRA_VERSION ?= $(shell git rev-parse --short HEAD)
 PROJECT_VERSION=$(BASE_VERSION)-snapshot-$(EXTRA_VERSION)
+TWO_DIGIT_VERSION = $(shell echo $(BASE_VERSION) | cut -d '.' -f 1,2)
 
 PKGNAME = github.com/$(PROJECT_NAME)
 CGO_FLAGS = CGO_CFLAGS=" "
@@ -308,6 +309,7 @@ $(BUILD_DIR)/image/%/$(DUMMY): Makefile $(BUILD_DIR)/image/%/payload $(BUILD_DIR
 	$(DBUILD) -t $(DOCKER_NS)/fabric-$(TARGET) $(@D)
 	docker tag $(DOCKER_NS)/fabric-$(TARGET) $(DOCKER_NS)/fabric-$(TARGET):$(DOCKER_TAG)
 	docker tag $(DOCKER_NS)/fabric-$(TARGET) $(DOCKER_NS)/fabric-$(TARGET):$(ARCH)-latest
+	docker tag $(DOCKER_NS)/fabric-$(TARGET) $(DOCKER_NS)/fabric-$(TARGET)
 	@touch $@
 
 $(BUILD_DIR)/gotools.tar.bz2: $(BUILD_DIR)/docker/gotools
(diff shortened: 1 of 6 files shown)


camillemarini:
Any idea on the timing of this review ? This is blocking us from using hlf in a production environment.

lindluni:
@camillemarini  You shouldn't be blocked, you should be setting `chaincode.builder` and `chaincode.<language>.runtime` to valid docker images in `core.yaml` or overriding them with `CORE_CHAINCODE_BUILDER` and `CORE_CHAINCODE_<language>_RUNTIME` environment variables

The only thing this PR is doing is setting those values (to a valid tag) for you, which is something you can do on your own.

camillemarini:
@btl5037 ok thank you for your quick answer :-)

Makefile
```diff
@@ -308,6 +308,7 @@ $(BUILD_DIR)/image/%/$(DUMMY): Makefile $(BUILD_DIR)/image/%/payload $(BUILD_DIR
 	$(DBUILD) -t $(DOCKER_NS)/fabric-$(TARGET) $(@D)
 	docker tag $(DOCKER_NS)/fabric-$(TARGET) $(DOCKER_NS)/fabric-$(TARGET):$(DOCKER_TAG)
 	docker tag $(DOCKER_NS)/fabric-$(TARGET) $(DOCKER_NS)/fabric-$(TARGET):$(ARCH)-latest
+	docker tag $(DOCKER_NS)/fabric-$(TARGET) $(DOCKER_NS)/fabric-$(TARGET):$(ARCH)-$(BASEIMAGE_RELEASE)
```
denyeart: I'm trying to understand why this one is BASEIMAGE_RELEASE rather than BASE_VERSION.

lindluni: You're right, the right thing to do here would be to hard code the tag like we do in release-2.x branches and tag the image proper in the `ccenv` target. I'll make that change

Makefile
```diff
@@ -308,6 +309,7 @@ $(BUILD_DIR)/image/%/$(DUMMY): Makefile $(BUILD_DIR)/image/%/payload $(BUILD_DIR
 	$(DBUILD) -t $(DOCKER_NS)/fabric-$(TARGET) $(@D)
 	docker tag $(DOCKER_NS)/fabric-$(TARGET) $(DOCKER_NS)/fabric-$(TARGET):$(DOCKER_TAG)
 	docker tag $(DOCKER_NS)/fabric-$(TARGET) $(DOCKER_NS)/fabric-$(TARGET):$(ARCH)-latest
+	docker tag $(DOCKER_NS)/fabric-$(TARGET) $(DOCKER_NS)/fabric-$(TARGET)
```
denyeart, line 312: Why is this extra tag needed?

denyeart (commented):
Overall I think this is the right approach. Just one question...

Makefile
lindluni: The `dockerutil` actually uses https://github.com/hyperledger/fabric/blob/master/common/metadata/metadata.go#L10 to derive the tag. So during testing it uses the `latest` tag. When the binaries are actually built metadata.go is linked using `ldflags` and the value is actually propogated from the Makefile. But this doesn't occur in testing. 

We had to do the same thing in the master branches and release-2.x branches. But Matt actually gave us a workaround that removed the dependency on all docker images for unit tests. Unfortunately it's not as straight-forward for release-1.4 so simply adding an additional tag to the make target seems simplest as it doesn't change any of the underlying code or functionality.

So what used to happen (and we didn't realize) was we were actually pulling ccenv/baseos from Dockerhub during testing instead of using the local build.

denyeart (approved):

PR #1598, opened 2020-07-13
Require orderer file ledger location to be set

wlahti:
#### Type of change

- New feature

#### Description

The orderer currently allows starting with a configuration that does not specify the Orderer.FileLedger.Location. When this occurs, the orderer uses the Orderer.FileLedger.Prefix and creates a temporary directory using this prefix. In order to handle crash fault tolerance, the orderer needs a configured directory and should not generate a temporary directory. The orderer now panics if a location is not provided.

#### Related issues

[FAB-18066](https://jira.hyperledger.org/browse/FAB-18066)

#### Release Note
Orderer.FileLedger.Location must now be set. The orderer will panic if it is not.

### DIFF
--- a/integration/nwo/fabricconfig/orderer.go
+++ b/integration/nwo/fabricconfig/orderer.go
@@ -72,7 +72,6 @@ type OrdererTopic struct {
 
 type FileLedger struct {
 	Location string `yaml:"Location,omitempty"`
-	Prefix   string `yaml:"Prefix,omitempty"`
 }
 
 type Kafka struct {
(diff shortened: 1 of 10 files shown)


jyellick (approved):
Looks good to me -- can you please make sure this gets added to the release notes as well?

lindluni:
This broke our deployments, when we upgraded to this code. Because our existing configs had the `Prefix` property in them, the strong marshaling of the Orderer config prevents the orderer.yaml from being marshaled while containing this key. The fix was to mount all of our volumes in kubernetes to secondary pods and remove the key. This is kind of breaking backwards compatibility, and would be a pain for anyone with a large deployment.

PR #1599, opened 2020-07-14
FAB-18065 Ch.Part.API: move follower to orderer/common

tock-ibm:
Following code review, it was decided to move the "follower" package 
from orderer/consensus/follower to orderer/common/follower.

The follower.Chain no longer implements consenus.Chain and is not
treated as a consensus plug-in.

Instead of a consensus.Consenter implemeting `JoinChain`, the registrar 
creates a follower.Chain and saves it in a separate map from
consensus.Chain implementations. Since the follower is not in the
chains map of the registrar, it is not found by broadcast requests
until it turns into an etcdraft.Chain.

Interface consensus.Consenter now defines a new interface
`IsChannelMember` that allows the registrar to detect whether a
join-block implies the orderer is a member or follower of the channel.

Change-Id: I12bd147961d70c6667aa93934ad9e936e2b4f03c

#### Type of change

#### Related issues

Task: FAB-18065
Epic: FAB-17712


orderer/common/multichannel/registrar.go
```diff
@@ -464,34 +478,148 @@ func (r *Registrar) JoinChannel(channelID string, configBlock *cb.Block, isAppCh
...
-	r.chains[channelID] = joinSupport
-	joinSupport.start()
+	r.followers[channelID] = fChain
+	fChain.Start()
 
+	logger.Infof("Joining channel: %v", info)
 	return info, nil
 }
 
+//func (c *Consenter) JoinChain(support consensus.ConsenterSupport, joinBlock *common.Block) (consensus.Chain, error) {
```
yacovm: :(

tock-ibm: leftovers...  oops

tock-ibm: done

orderer/common/multichannel/follower_resources.go
```diff
@@ -0,0 +1,84 @@
...
+	"github.com/hyperledger/fabric/common/channelconfig"
+	"github.com/hyperledger/fabric/common/ledger/blockledger"
+	"github.com/hyperledger/fabric/common/policies"
+	"github.com/hyperledger/fabric/internal/pkg/identity"
+	"github.com/hyperledger/fabric/orderer/consensus"
+	"github.com/hyperledger/fabric/protoutil"
+	"github.com/pkg/errors"
+)
+
+// FollowerResources encapsulates all the resources the follower needs to do its job:
```
jyellick: I realize this is an iterative change, so I understand where this is coming from.  But let's not copy the monolithic interface style from the consenter package, or the embedding of the ledger resources to implement things.

We can have a plain struct with fields each satisfying an interface (or, better yet, as a literal if we can get away with it), that way we can more sanely compose things and it's clearer where our implementations are coming from.

orderer/consensus/etcdraft/consenter_test.go
```diff
@@ -117,6 +120,52 @@ var _ = Describe("Consenter", func() {
...
+			confAppRaft.Consortiums = nil
+			confAppRaft.Consortium = ""
+			serverCertificates = generateCertificates(confAppRaft, tlsCA, dataDir)
+			bootstrapper, err := encoder.NewBootstrapper(confAppRaft)
+			Expect(err).NotTo(HaveOccurred())
+			genesisBlockApp = bootstrapper.GenesisBlockForChannel("my-raft-channel")
+			Expect(genesisBlockApp).NotTo(BeNil())
+		})
+
+		It("identifies a bad block", func() {
```
jyellick: All of these are not really written as a ginkgo-style test.  You might instead want to look at the [table extension](https://godoc.org/github.com/onsi/ginkgo/extensions/table).

orderer/consensus/kafka/consenter.go
```diff
@@ -103,8 +103,8 @@ func (consenter *consenterImpl) HandleChain(support consensus.ConsenterSupport,
 	return ch, nil
 }
 
-func (c *consenterImpl) JoinChain(support consensus.ConsenterSupport, joinBlock *cb.Block) (consensus.Chain, error) {
-	return nil, errors.New("the Kafka orderer does not support JoinChain")
+func (c *consenterImpl) IsChannelMember(joinBlock *cb.Block) (bool, error) {
```
jyellick: We could instead of making `IsChannelMember` a part of the consenter interface, having a separate `ClusterConsenter` interface that consensus types can optionally implement.  It would be a simple type assertion on the other side, and would make it easier to differentiate between a 'not supported' type error, and a genuine one.

orderer/common/multichannel/follower_resources.go
```diff
@@ -0,0 +1,84 @@
...
+		ledgerResources:  ledgerResources,
+		SignerSerializer: signer,
+		BCCSP:            bccsp,
+		consenter:        consenter,
+	}
+
+	return fRes
+}
+
+// TODO Move this method to ledgerResources struct, so it can be shared with the ChainSupport struct.
```
jyellick, line 42: The ledgerResources might be the right place to hang this, but, please don't be afraid to break up and or refactor the 'multichannel.configReources' type and its children, that structure was poorly thought, hard to follow, and now might be the time to refactor it.

jyellick (commented):
This definitely looks like we're headed in the right direction.  I do have a few comments.  We can either address them here, or, if you'd like to tackle them later in the chain, that's fine too.

orderer/common/follower/follower_chain.go
```diff
@@ -82,34 +100,39 @@ type Chain struct {
...
 	joinBlock *common.Block,
 	options Options,
-	createBlockPullerFunc CreateBlockPuller, // A method that creates a block puller on demand
+	createBlockPullerFunc CreateBlockPuller,
 	chainCreationCallback func(),
 	cryptoProvider bccsp.BCCSP,
-	amIInChannel cluster.SelfMembershipPredicate,
 ) (*Chain, error) {
+
+	// Check it once
```
wlahti: nit: what is "it" here? If we're going to have a comment, I think it should be more helpful for someone coming into the code without any knowledge of this area

orderer/consensus/consensus.go
```diff
@@ -25,11 +25,11 @@ type Consenter interface {
...
 	HandleChain(support ConsenterSupport, metadata *cb.Metadata) (Chain, error)
 
-	// JoinChain should create and return a reference to a Chain when the channel is started with a join-block
-	// that is not a genesis block (i.e. join-block.number>0), and a ledger with height <= than join-block.number.
-	// This means that there is a gap between the join-block and the last block in the ledger that requires
-	// on-boarding. In contrast, HandleChain creates a Chain based on the last block found in the ledger.
-	JoinChain(support ConsenterSupport, joinBlock *cb.Block) (Chain, error)
+	// IsChannelMember inspects the join block and detects whether it implies that this orderer is a member of the
+	// channel. It returns true if thus orderer is a member of the consenters set, and false if it is. The method
+	// also inspects the consensus type metadata for validity. It returns an error of membership cannot be determined
```
wlahti:
```suggestion
	// also inspects the consensus type metadata for validity. It returns an error if membership cannot be determined
```
Nit

orderer/common/follower/follower_chain.go
```diff
@@ -186,19 +191,24 @@ func (c *Chain) StatusReport() (types.ClusterRelation, types.Status) {
 
 	clusterRelation := types.ClusterRelationFollower
 	if c.joinBlock != nil {
-		if err := c.amIInChannel(c.joinBlock); err == nil {
+		isMem, _ := c.resources.IsChannelMember(c.joinBlock)
```
wlahti: Just wondering why we don't care about the error here. Is this code path not reachable for solo/kafka orderers?

orderer/common/multichannel/follower_resources.go
tock-ibm: Yes, I agree completely. The "support" pattern is a bad one and I am eager to get rid of it as soon as possible (at least for the follower at this point in time) . I will prepare the ground in this PR and complete that in the following PRs.

orderer/consensus/kafka/consenter.go
tock-ibm: Good idea.

orderer/common/multichannel/follower_resources.go
tock-ibm: I agree. I would like to untangle the `ledgerResources` step by step, in a PR of its own, for safety, as it affects all the other chains.

orderer/common/follower/follower_chain.go
tock-ibm: IsChannelMember() is only relevant for cluster-type consenters (follower, etcdraft) and is checked against errors when the follower is created. If it errs, the join is rejected since it indicates the join-block is bad. Here, there is virtually no chance of an error, and if it errs, we just treat it as a non member, which does no harm.

tock-ibm: PS - in the next commit to this PR IsChannelMember will not be implemented by solo & kafka at all.

orderer/common/follower/follower_chain.go

orderer/consensus/consensus.go
tock-ibm: Done

orderer/common/multichannel/follower_resources.go
tock-ibm: I plan to do this in the next task: https://jira.hyperledger.org/browse/FAB-18070

orderer/consensus/etcdraft/consenter_test.go
tock-ibm: Done

orderer/common/multichannel/follower_resources.go
jyellick: Sounds good to me.

orderer/common/multichannel/follower_resources.go
jyellick: Great, thanks.

orderer/common/multichannel/registrar.go
```diff
@@ -464,31 +478,97 @@ func (r *Registrar) JoinChannel(channelID string, configBlock *cb.Block, isAppCh
...
+
+	clusterConsenter, ok := consenter.(consensus.ClusterConsenter)
+	if !ok {
+		return types.ChannelInfo{}, errors.New("clusterConsenter is not a consensus.ClusterConsenter")
+	}
+	isMember, err := clusterConsenter.IsChannelMember(configBlock)
 	if err != nil {
-		//TODO remove join block, clean ledger resources
-		return types.ChannelInfo{}, errors.Wrap(err, "failed creating chain support for join")
+		return types.ChannelInfo{}, errors.Wrap(err, "failed to determine cluster membership from join-block ")
```
jyellick, line 500: Do you think this is the right approach?

The consensus type on a chain could change over time, and it seems like there's no explicit reason not to allow following a channel which the orderer cannot join?  I would have thought that we simply follow the channel, and, whenever a config change commits, look up the consensus plugin, check to see if it is a cluster type, and if so, check to see if we've been added to the membership (and if so, return to begin consenting).

Basically instead of passing a particular clusterConsenter in, why not pass in a map from consensus types to cluster consenter, and let it resolve the particular checker internally?

jyellick (approved):
I think this looks good enough to proceed, one comment I would like you to consider.

orderer/common/multichannel/registrar.go
tock-ibm: Do you mean change over time like in `kafka -> raft` or `raft -> something-new` ? (e.g. `a -> b`).
I haven't considered this scenario.

So we have a prefix of a chain with type `a` and a suffix of type `b`.
1. Let's assume we supply the orderer with a type `b` block (member or follower, doesn't matter).  All should work well. The orderer will pull all blocks from current ledger height until join block, without looking into config blocks at all (see next PR). Then it will create a `b` type chain or a new follower. The on-boarding follower has a link to a type `b` consenter, so it will be able to detect membership when the time comes. The "following" follower, the one that pulls blocks past the join block (2 PRs from here) will also have a link to a type `b` consenter, so it will be able to detect membership as well.

2. Now let's assume we give it a type `a` block that indicates non-membership. (If it is a member then the type `a` chain should support the migration procedure). The follower has a link to a type `a` consenter, and suddenly receives the first config block of type `b`. The follower should detect that, avoid processing the type `b` consensus metadata, and append the block. Here usually we'll have a re-start (at least on `kafka->raft` we have it), or the follower will recreate itself from the tip of the ledger; in both cases this will create a new follower with a link to a type `b` consenter. Your proposal also make sense - the follower will scan through the clusterConsenter map to find a type that fits and will continue from there.

We currently do not support `kafka` and `solo` at all; since they are deprecated. However, going forward, I don't think supporting this kind of behavior will be a problem. I prefer to keep it simple right now, and wait until we have a type after raft, and only then plan a migration procedure that includes the follower... it should not be too difficult.

With respect to what happens here: this error happens when an `etcdraft` consenter fails to process the consensus metadata of a block with `etcdraft` consensus-type (as `IsChannelMember` also checks the metadata). I think we should throw an error here and reject the join-block. Maybe the error message is a bit misleading - I guess a better one would be "bad join-block, error trying to bla bla...". That is, to indicate that the failure is on the part of the user for supplying a bad block, not the system for failing to process it.

PR #1600, opened 2020-07-14
Raft: Check suspect info once per suspect interval

jyellick:
#### Type of change

- Bug fix

#### Description

Today's existing suspect logic has a periodic checker, which checks
every 10s if the Raft cluster still has quorum.  If the cluster has lost
quorum, it marks the time this event begins, then, every 10s checks to
see if 'enough' time has elapsed since the quorum was lost to suspect
that the OSN has been evicted.

If the OSN has not been evicted, or cannot determine its eviction
status, then every 10s the OSN attempts to re-check its suspicion
status, which can lead to large volumes of network traffic, especially
in significiantly multichannel environments.

This commit modifies the logic to track the number of times that the
suspect checking logic has actually executed, to ensure that we check no
more than once every suspect interval (by default every 10m, instead of
every 10s).
-->

### DIFF
--- a/orderer/consensus/etcdraft/eviction.go
+++ b/orderer/consensus/etcdraft/eviction.go
@@ -26,6 +26,7 @@ type PeriodicCheck struct {
 	CheckInterval       time.Duration
 	Condition           func() bool
 	Report              func(cumulativePeriod time.Duration)
+	ReportCleared       func()
 	conditionHoldsSince time.Time
 	once                sync.Once // Used to prevent double initialization
 	stopped             uint32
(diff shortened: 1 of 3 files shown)


jyellick:
@yacovm @guoger @tock-ibm Could you take a look? Assuming this looks good, I think we'll want to create backports.

orderer/consensus/etcdraft/eviction.go
```diff
@@ -60,6 +61,9 @@ func (pc *PeriodicCheck) check() {
 }
 
 func (pc *PeriodicCheck) conditionNotFulfilled() {
+	if pc.ReportCleared != nil && pc.conditionHoldsSince != (time.Time{}) {
```
yacovm: maybe instead of:
`pc.conditionHoldsSince != (time.Time{}`
do:
`pc.conditionHoldsSince.IsZero()` ?

jyellick: Sure -- that is quite a bit more graceful, will fix.

yacovm (approved):

jyellick:

jyellick:

jyellick:

jyellick:

PR #1607, opened 2020-07-15
[FAB-17906] Add snapshot management APIs

wenjianqiao:

#### Type of change
- New feature

#### Description
This PR adds management APIs to submit/cancel snapshot requests and
call generateSnapshot after block commit and at peer startup.

#### Additional details
The remaining work includes: 
1) implement all ListSnapshots/DeleteSnapshot APIs
2) add tests under kvledger/tests

#### Related issues

### DIFF
--- a/core/ledger/kvledger/kv_ledger.go
+++ b/core/ledger/kvledger/kv_ledger.go
@@ -45,18 +45,20 @@ var (
 // kvLedger provides an implementation of `ledger.PeerLedger`.
 // This implementation provides a key-value based data model
 type kvLedger struct {
-	ledgerID               string
-	blockStore             *blkstorage.BlockStore
-	pvtdataStore           *pvtdatastorage.Store
-	txmgr                  *txmgr.LockBasedTxMgr
-	historyDB              *history.DB
-	configHistoryRetriever *confighistory.Retriever
-	bookkeepingProvider    *bookkeeping.Provider
-	blockAPIsRWLock        *sync.RWMutex
-	stats                  *ledgerStats
-	commitHash             []byte
-	hashProvider           ledger.HashProvider
-	config                 *ledger.Config
+	ledgerID                  string
+	blockStore                *blkstorage.BlockStore
+	pvtdataStore              *pvtdatastorage.Store
+	txmgr                     *txmgr.LockBasedTxMgr
+	historyDB                 *history.DB
+	configHistoryRetriever    *confighistory.Retriever
+	snapshotRequestBookkeeper *snapshotRequestBookkeeper
+	blockAPIsRWLock           *sync.RWMutex
+	snapshotGenerationRWLock  sync.RWMutex
+	snapshotRequestLock       sync.Mutex
+	stats                     *ledgerStats
+	commitHash                []byte
+	hashProvider              ledger.HashProvider
+	config                    *ledger.Config
 	// isPvtDataStoreAheadOfBlockStore is read during missing pvtData
 	// reconciliation and may be updated during a regular block commit.
 	// Hence, we use atomic value to ensure consistent read.
@@ -84,14 +86,13 @@ func newKVLedger(initializer *lgrInitializer) (*kvLedger, error) {
 	ledgerID := initializer.ledgerID
 	logger.Debugf("Creating KVLedger ledgerID=%s: ", ledgerID)
 	l := &kvLedger{
-		ledgerID:            ledgerID,
-		blockStore:          initializer.blockStore,
-		pvtdataStore:        initializer.pvtdataStore,
-		historyDB:           initializer.historyDB,
-		bookkeepingProvider: initializer.bookkeeperProvider,
-		hashProvider:        initializer.hashProvider,
-		config:              initializer.config,
-		blockAPIsRWLock:     &sync.RWMutex{},
+		ledgerID:        ledgerID,
+		blockStore:      initializer.blockStore,
+		pvtdataStore:    initializer.pvtdataStore,
+		historyDB:       initializer.historyDB,
+		hashProvider:    initializer.hashProvider,
+		config:          initializer.config,
+		blockAPIsRWLock: &sync.RWMutex{},
 	}
 
 	btlPolicy := pvtdatapolicy.ConstructBTLPolicy(&collectionInfoRetriever{ledgerID, l, initializer.ccInfoProvider})
@@ -152,6 +153,16 @@ func newKVLedger(initializer *lgrInitializer) (*kvLedger, error) {
 		return nil, err
 	}
 	l.configHistoryRetriever = initializer.configHistoryMgr.GetRetriever(ledgerID, l)
+
+	dbHandle := initializer.bookkeeperProvider.GetDBHandle(l.ledgerID, bookkeeping.SnapshotRequest)
+	l.snapshotRequestBookkeeper, err = newSnapshotRequestBookkeeper(dbHandle)
+	if err != nil {
+		return nil, err
+	}
+
+	if err = l.recoverMissingSnapshot(); err != nil {
+		return nil, err
+	}
 
 	l.stats = initializer.stats
 	return l, nil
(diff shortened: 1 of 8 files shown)


core/ledger/kvledger/snapshot_mgmt.go
```diff
@@ -0,0 +1,213 @@
...
+)
+
+// SubmitSnapshotRequest submits a snapshot request.
+// If the height is bigger than the block height, the request will be stored in the bookkeeper db to be triggered later.
+// If the height is equal to the block height, it will generate a snapshot.
+// If the height is smaller than the block height, it will return an error.
+// When height is 0, it will generate a snapshot at the current block height.
+func (l *kvLedger) SubmitSnapshotRequest(height uint64) error {
+	l.blockAPIsRWLock.Lock()
+	defer l.blockAPIsRWLock.Unlock()
```
manish-sethi: see lock related comment above

core/ledger/kvledger/snapshot_mgmt.go
```diff
@@ -0,0 +1,213 @@
...
+
+	// call l.blockStore directly since it already acquires the blockAPIsRWLock
```
manish-sethi: Also, this can be simplified with the lock related suggestion above.

manish-sethi (changes requested):
Some high level comments

core/ledger/kvledger/snapshot_mgmt.go
```diff
@@ -0,0 +1,213 @@
...
+	return k.dbHandle.Get(util.EncodeOrderPreservingVarUint64(height))
+}
+
+func (k *snapshotRequestBookkeeper) add(height uint64) error {
+	key := util.EncodeOrderPreservingVarUint64(height)
+	request := &msgs.SnapshotRequest{Status: msgs.SnapshotStatus_PENDING}
+	requestBytes, err := proto.Marshal(request)
+	if err != nil {
+		return err
+	}
```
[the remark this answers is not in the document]
wenjianqiao: Currently the protobuf msg is only a placeholder and not used yet. Will remove it.

core/ledger/kvledger/snapshot_mgmt.go
```diff
@@ -0,0 +1,342 @@
...
+// returns an error if no such a snapshot exists
+func (l *kvLedger) DeleteSnapshot(height uint64) error {
+	return errors.Errorf("not implemented")
+}
+
+// generateSnapshotIfNeeded generates a snapshot if the block height matches smallest request height.
+// It locks snapshotGenerationRWLock to synchronize between commit and snapshot generation.
+// It starts a go-routine to call generateSnapshot if needed and will keep the snapshotGenerationRWLock
+// until generateSnapshot is finished.
+func (l *kvLedger) generateSnapshotIfNeeded() error {
```
manish-sethi: You need to take `snapshotRequestLock` here, else, following are the two side effects
- Access to the variable `l.snapshotRequestBookkeeper.smallestRequestHeight` in this function is unprotected
- This function can can concurrently execute with the function `SubmitSnapshotRequest` and this in a corner case can cause launching the  `generateSnapshot` goroutine for the same snapshot twice.

I think that it would be much easier to understand the idea of post commit snapshot generation, if you don't have this function altogether and simply move the code at the end of commit function. There, you already have the block number and moreover, this function is not specific to snapshot mgmt as such.

core/ledger/kvledger/kv_ledger.go
```diff
@@ -480,6 +502,14 @@ func (l *kvLedger) CommitLegacy(pvtdataAndBlock *ledger.BlockAndPvtData, commitO
 	elapsedBlockProcessing := time.Since(startBlockProcessing)
 
 	startBlockstorageAndPvtdataCommit := time.Now()
+
+	logger.Debugf("[%s] Committing pvtdata and block [%d] to storage", l.ledgerID, blockNo)
+	// need to unlock snapshotGenerationRWLock before each return depending on code paths
+	l.snapshotGenerationRWLock.Lock()
+	defer l.snapshotGenerationRWLock.Unlock()
```
manish-sethi: Not sure what you meant by this code comment.

Also, can you make following two minor changes 
- Move this lock acquisition in the commitLegacy function before calling the commit function (just to avoid side-effects such as here, this may cause `elapsedBlockstorageAndPvtdataCommit` to be a very high value, while the commit may actually be waiting for background snapshot generation to finish

- Keep some flag if a background snapshot generation is ON and print a log message (like - commit waiting on .... in case a user may wonder what's happening with the commit). Feel free to tackle this in a separate PR, if you will

core/ledger/kvledger/snapshot.go
```diff
@@ -44,14 +44,27 @@ type snapshotAdditionalInfo struct {
...
+// - when a ledger is opened, if the smallest snapshot request height matches the block height
+//   and such a snapshot has not been generated.
+// In all cases, further commits should not be resumed till this function finishes, which is synchronized
+// by snapshotGenerationRWLock. The caller should acquire snapshotGenerationRWLock.RLock and this function
+// needs to unlock it before it returns. The corresponding snapshot request will be deleted no matter
+// if the snapshot generation is successful or not.
 func (l *kvLedger) generateSnapshot() error {
+	defer func() {
+		l.deleteSnapshotRequest()
+		l.snapshotGenerationRWLock.RUnlock()
```
manish-sethi: I feel that it will make it more maintainable if you move this line to the goroutine function that invokes `generateSnapshot`. It's always easy to understand lock acquire/release pair at a single place, Then, you won't need the changes caused in the file snapshot_test.go

manish-sethi (changes requested):
A few comments, I am still to look at the test file...

core/ledger/kvledger/snapshot_mgmt.go
```diff
@@ -0,0 +1,329 @@
...
+	}
+	if exists {
+		return errors.Errorf("duplicate snapshot request for height %d", height)
+	}
+
+	if err = l.snapshotRequestBookkeeper.add(height); err != nil {
+		return err
+	}
+
+	if height == bcInfo.Height {
```
cendhu, line 73: revert the condition and return early.

core/ledger/kvledger/snapshot_mgmt.go
```diff
@@ -0,0 +1,329 @@
...
+// until generateSnapshot is finished.
+func (l *kvLedger) SubmitSnapshotRequest(height uint64) error {
+	l.snapshotRequestLock.Lock()
+	defer l.snapshotRequestLock.Unlock()
+
+	// before calling GetBlockchainInfo, locks snapshotGenerationRWLock to synchronize with commit.
+	// It gets the block height and invokes generateSnapshot if the submitted height matches
+	// the block height. The snapshotGenerationRWLock will be held and unlocked inside of generateSnapshot.
+	keepLock := false
+	l.snapshotGenerationRWLock.RLock()
```
cendhu, line 40: Would it make sense to use `blockCommit.RLock()` instead of `snapshotGeneration.RLock()`?

core/ledger/kvledger/snapshot_mgmt.go
```diff
@@ -0,0 +1,329 @@
...
+// and CancelSnapshotRequest calls.
+// It also locks snapshotGenerationRWLock to synchronize with commit before getting
+// blockchain info so that it will not cancel a under-processing request.
+func (l *kvLedger) CancelSnapshotRequest(height uint64) error {
+	l.snapshotRequestLock.Lock()
+	defer l.snapshotRequestLock.Unlock()
+
+	l.snapshotGenerationRWLock.RLock()
+	defer l.snapshotGenerationRWLock.RUnlock()
+
```
cendhu, line 101: Isn't it okay to cancel the request when the block commit happens?

core/ledger/kvledger/snapshot_mgmt.go
```diff
@@ -0,0 +1,329 @@
...
+	// read db to get smallest request height
+	itr, err := dbHandle.GetIterator(nil, nil)
+	if err != nil {
+		return nil, err
+	}
+	defer itr.Release()
+
+	if err = itr.Error(); err != nil {
+		return nil, errors.Wrapf(err, "internal leveldb error while iterating for snapshot requests")
+	}
```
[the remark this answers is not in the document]
wenjianqiao: @cendhu thank you for pointing it out. I didn't realize that dbHandler.GetIterator already performs this check.

wenjianqiao:
Close the PR due to redesign.

PR #1608, opened 2020-07-15
Channel participation API: make joins crash fault tolerant

wlahti:
#### Type of change

- New feature

#### Description

First commit: 
- save joinblock on JoinChannel

Second commit:
- process joinblocks in filerepo during registrar init

#### Related issues

[FAB-18008](https://jira.hyperledger.org/browse/FAB-18008)

### DIFF
--- a/orderer/common/multichannel/registrar.go
+++ b/orderer/common/multichannel/registrar.go
@@ -14,6 +14,7 @@ import (
 	"path/filepath"
 	"sync"
 
+	"github.com/golang/protobuf/proto"
 	cb "github.com/hyperledger/fabric-protos-go/common"
 	"github.com/hyperledger/fabric/bccsp"
 	"github.com/hyperledger/fabric/common/channelconfig"
@@ -105,7 +106,10 @@ func NewRegistrar(
 	}
 
 	if config.ChannelParticipation.Enabled {
-		r.initializeJoinBlockFileRepo()
+		err := r.initializeJoinBlockFileRepo()
+		if err != nil {
+			logger.Panicf("Error initializing joinblock file repo: %s", err)
+		}
 	}
 
 	return r
@@ -113,16 +117,18 @@ func NewRegistrar(
 
 // initialize the channel participation API joinblock file repo. This creates
 // the fileRepoDir on the filesystem if it does not already exist.
-func (r *Registrar) initializeJoinBlockFileRepo() {
+func (r *Registrar) initializeJoinBlockFileRepo() error {
 	fileRepoDir := filepath.Join(r.config.FileLedger.Location, "filerepo")
 	logger.Infof("Channel Participation API enabled, registrar initializing with file repo %s", fileRepoDir)
 
 	joinBlockFileRepo, err := filerepo.New(fileRepoDir, "joinblock")
 	if err != nil {
-		logger.Panicf("Error initializing join block file repo: %s", err)
+		return err
 	}
 
 	r.joinBlockFileRepo = joinBlockFileRepo
+
+	return nil
 }
 
 func (r *Registrar) Initialize(consenters map[string]consensus.Consenter) {
@@ -255,7 +261,9 @@ func (r *Registrar) initAppChannels(existingChannels []string, channelsWithJoinB
 			if _, _, err := r.createAsMember(ledgerRes, joinBlock, channelID); err != nil {
 				logger.Panicf("Failed to createAsMember, error: %s", err)
 			}
-			// TODO remove the join block
+			if err := r.removeJoinBlock(channelID); err != nil {
+				logger.Panicf("Failed to remove join-block, channel: %s, error: %s", channelID, err)
+			}
 		} else {
 			if _, _, err = r.createFollower(ledgerRes, clusterConsenter, joinBlock, channelID); err != nil {
 				logger.Panicf("Failed to createFollower, error: %s", err)
@@ -512,22 +520,25 @@ func (r *Registrar) createNewChain(configtx *cb.Envelope) *ChainSupport {
 // SwitchFollowerToChain creates a consensus.Chain from the tip of the ledger, and removes the follower.
 // It is called when a follower detects a config block that indicates cluster membership and halts, transferring
 // execution to the consensus.Chain.
-func (r *Registrar) SwitchFollowerToChain(chainName string) {
+func (r *Registrar) SwitchFollowerToChain(channelID string) {
 	r.lock.Lock()
 	defer r.lock.Unlock()
 
-	lf, err := r.ledgerFactory.GetOrCreate(chainName)
+	lf, err := r.ledgerFactory.GetOrCreate(channelID)
 	if err != nil {
-		logger.Panicf("Failed obtaining ledger factory for %s: %v", chainName, err)
+		logger.Panicf("Failed obtaining ledger factory for channel %s: %v", channelID, err)
 	}
 
-	if _, chainExists := r.chains[chainName]; chainExists {
-		logger.Panicf("Programming error, chain already exists: %s", chainName)
+	if _, chainExists := r.chains[channelID]; chainExists {
+		logger.Panicf("Programming error, channel already exists: %s", channelID)
 	}
 
-	delete(r.followers, chainName)
-	logger.Debugf("Removed follower for channel %s", chainName)
+	delete(r.followers, channelID)
+	logger.Debugf("Removed follower for channel %s", channelID)
 	cs := r.createNewChain(configTx(lf))
+	if err := r.removeJoinBlock(channelID); err != nil {
+		logger.Panicf("Failed removing join-block for channel: %s: %v", channelID, err)
+	}
 	cs.start()
 	logger.Infof("Created and started channel %s", cs.ChannelID())
 }
@@ -639,7 +650,7 @@ func (r *Registrar) ChannelInfo(channelID string) (types.ChannelInfo, error) {
 
 // JoinChannel instructs the orderer to create a channel and join it with the provided config block.
 // The URL field is empty, and is to be completed by the caller.
-func (r *Registrar) JoinChannel(channelID string, configBlock *cb.Block, isAppChannel bool) (types.ChannelInfo, error) {
+func (r *Registrar) JoinChannel(channelID string, configBlock *cb.Block, isAppChannel bool) (info types.ChannelInfo, err error) {
 	r.lock.RLock()
 	defer r.lock.RUnlock()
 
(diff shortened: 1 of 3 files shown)


orderer/common/multichannel/registrar.go
```diff
@@ -239,6 +293,24 @@ func (r *Registrar) Initialize(consenters map[string]consensus.Consenter) {
...
+
+	joinBlockFileRepo, err := filerepo.New(fileRepoDir, "joinblock")
+	if err != nil {
+		logger.Panicf("Error initializing join block file repo: %s", err)
+	}
+
+	r.joinBlockFileRepo = joinBlockFileRepo
+
+	// TODO removeblock file repo
+}
```
tock-ibm: This does not belong here... the registar only manages the join-block repo. The remove repo will be internal to the file-ledger.

orderer/common/multichannel/registrar.go
```diff
@@ -498,3 +589,11 @@ func (r *Registrar) RemoveChannel(channelID string, removeStorage bool) error {
...
 	return errors.New("Not implemented yet")
 }
+
+func (r *Registrar) removeJoinBlock(channelID string) error {
+	if err := r.joinBlockFileRepo.Remove(channelID); err != nil {
+		return errors.Wrapf(err, "failed removing joinblock for channel %s", channelID)
+	}
+
+	return nil
+}
```
tock-ibm, line 903: We should have a version of this exported and protected with a mutex, so that when a follower finishes on boarding it will be able to call and remove the join block.

orderer/common/multichannel/registrar.go
```diff
@@ -239,6 +293,24 @@ func (r *Registrar) Initialize(consenters map[string]consensus.Consenter) {
 	}
 }
 
+// initialize the channel participation API file repos. This creates the
+// joinblock (and TODO removeblock) file repos for the orderer, creating
+// the fileRepoDir on the filesystem if it does not already exist.
+func (r *Registrar) initializeFileRepos() {
```
tock-ibm: return an error, not panic

orderer/common/multichannel/registrar.go
```diff
@@ -155,8 +160,57 @@ func NewRegistrar(
 func (r *Registrar) Initialize(consenters map[string]consensus.Consenter) {
 	r.consenters = consenters
 	existingChannels := r.ledgerFactory.ChannelIDs()
+	existingChannelsMap := map[string]struct{}{}
+	for _, channel := range existingChannels {
+		existingChannelsMap[channel] = struct{}{}
+	}
+
+	if r.config.ChannelParticipation.Enabled {
+		r.initializeFileRepos()
```
tock-ibm: This probably belongs in the `NewRegistrar`, returning an error, not panicking.

orderer/common/multichannel/registrar.go
```diff
@@ -155,8 +160,57 @@ func NewRegistrar(
 func (r *Registrar) Initialize(consenters map[string]consensus.Consenter) {
 	r.consenters = consenters
 	existingChannels := r.ledgerFactory.ChannelIDs()
+	existingChannelsMap := map[string]struct{}{}
+	for _, channel := range existingChannels {
+		existingChannelsMap[channel] = struct{}{}
+	}
```
tock-ibm: Following the code review we had, the registrar struct and registrar.Initialize are going to change a lot even before this PR,  so I would advise waiting until the dust settles. 

Among other things, we're gonna have a separate new map for followers and are going to need to discover if the system channel exists first.

orderer/common/multichannel/registrar.go
```diff
@@ -155,8 +160,57 @@ func NewRegistrar(
...
+				r.signer,
+				r.blockcutterMetrics,
+				r.bccsp,
+			)
+			if err != nil {
+				logger.Panicf("Error creating chain support: %s", err)
+			}
+
+			r.chains[channel] = chain
+			chain.start()
```
tock-ibm: we should not start until we have all the chains initialized,  or we run the risk of a follower thread accessing the chains map before we finish populating it.

orderer/common/multichannel/registrar.go
```diff
@@ -463,18 +535,37 @@ func (r *Registrar) JoinChannel(channelID string, configBlock *cb.Block, isAppCh
...
-		//TODO remove join block, clean ledger resources
-		return types.ChannelInfo{}, errors.Wrap(err, "failed creating chain support for join")
+		err = errors.Wrap(err, "failed creating chain support for join")
+
+		if err2 := r.removeJoinBlock(channelID); err2 != nil {
+			// Stack file repo error on top of chain support creation error
+			err = errors.WithStack(err2)
+		}
+
+		return types.ChannelInfo{}, err
```
tock-ibm: at this point we have to clean the ledger as well

wlahti:
I haven't addressed any comments yet. Just rebased this onto master and adjusted the code as necessary.

wlahti:
Just another rebase to spread out the effort of keeping up with master.

orderer/common/multichannel/registrar.go
wlahti: Removed

orderer/common/multichannel/registrar.go
wlahti: This was handled via @tock-ibm 's refactor of the registrar initialization.

orderer/common/multichannel/registrar.go
wlahti: Done

orderer/common/multichannel/registrar.go
wlahti: This was handled via @tock-ibm 's refactor of the registrar initialization.

orderer/common/multichannel/registrar.go
wlahti: Is there a particular reason the follower should handle removing the join block and not the registrar itself? For now, I've added the remove logic into SwitchFollowerToChain.

orderer/common/multichannel/registrar.go
wlahti: Done

orderer/common/multichannel/registrar.go
wlahti: Done

orderer/common/multichannel/registrar.go
tock-ibm: When the ledger progresses past the join block, there is no reason to keep it around... there is no harm in that, but it still feels a bit weird. Note that the follower can live for quite some time past the join-block... 

Nevertheless, this is acceptable. 

Another thing we could do in support of lazy removal is to use `loadJoinBlocks()` and remove a join block if the ledger's `height > join-block.Number`.

orderer/common/multichannel/registrar.go
```diff
@@ -651,7 +670,19 @@ func (r *Registrar) createAsMember(ledgerRes *ledgerResources, configBlock *cb.B
 		r.bccsp,
 	)
 	if err != nil {
-		return nil, types.ChannelInfo{}, errors.Wrap(err, "error creating chain support")
+		err = errors.Wrap(err, "error creating chain support")
+
+		if err2 := r.ledgerFactory.Remove(channelID); err2 != nil {
+			// Stack ledger removal error on top of chain support creation error
+			err = errors.WithStack(err2)
+		}
```
tock-ibm: you can put this in a defer that checks the return `err` right after the ledger is created,

orderer/common/multichannel/registrar.go
```diff
@@ -651,7 +670,19 @@ func (r *Registrar) createAsMember(ledgerRes *ledgerResources, configBlock *cb.B
...
+
+		if err2 := r.removeJoinBlock(channelID); err2 != nil {
+			// Stack file repo error on top of ledger removal + chain support creation error
+			err = errors.WithStack(err2)
+		}
```
tock-ibm: you can put this in a defer that checks the return `err` right after the join-block is saved

orderer/common/multichannel/registrar.go
```diff
@@ -609,8 +620,14 @@ func (r *Registrar) JoinChannel(channelID string, configBlock *cb.Block, isAppCh
...
-	//TODO save the join-block in the file repo to make this action crash tolerant.
-	//TODO remove join block & ledger if things go bad below, (using defer that identifies an error?)
+	blockBytes, err := proto.Marshal(configBlock)
+	if err != nil {
+		return types.ChannelInfo{}, errors.Wrap(err, "failed marshaling joinblock")
+	}
+
+	if err := r.joinBlockFileRepo.Save(channelID, blockBytes); err != nil {
+		return types.ChannelInfo{}, errors.Wrapf(err, "failed saving joinblock to file repo for channel %s", channelID)
+	}
```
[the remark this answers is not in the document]
wlahti: Done

orderer/common/multichannel/registrar.go
wlahti: Done

orderer/common/multichannel/registrar.go
wlahti: Done

orderer/common/multichannel/registrar.go
```diff
@@ -609,8 +620,14 @@ func (r *Registrar) JoinChannel(channelID string, configBlock *cb.Block, isAppCh
 		return types.ChannelInfo{}, err
 	}
 
-	//TODO save the join-block in the file repo to make this action crash tolerant.
-	//TODO remove join block & ledger if things go bad below, (using defer that identifies an error?)
+	blockBytes, err := proto.Marshal(configBlock)
+	if err != nil {
+		return types.ChannelInfo{}, errors.Wrap(err, "failed marshaling joinblock")
+	}
```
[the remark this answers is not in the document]
wlahti: Done

orderer/common/multichannel/registrar.go
```diff
@@ -754,7 +796,12 @@ func (r *Registrar) removeMember(channelID string, cs *ChainSupport) error {
 func (r *Registrar) removeFollower(channelID string, follower *follower.Chain) error {
 	follower.Halt()
 
-	// TODO if follower is onboarding, remove the joinblock from file repo
+	_, status := follower.StatusReport()
+	if status == types.StatusOnBoarding {
+		if err := r.removeJoinBlock(channelID); err != nil {
```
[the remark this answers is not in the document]
wlahti: Changed this to always attempt removing the join block.

tock-ibm:
Looks good to me.

tock-ibm:
Looks good to me.

orderer/common/multichannel/registrar.go
```diff
@@ -608,19 +619,41 @@ func (r *Registrar) JoinChannel(channelID string, configBlock *cb.Block, isAppCh
 	if err != nil {
 		return types.ChannelInfo{}, err
 	}
+	defer func() {
```
wlahti, line 673: @tock-ibm What about the case where the initLedgerResourcesClusterConsenter fails? That will leave around the ledger for the channel. Do you think newLedgerResources should be updated to remove the ledger if the call to ledgerFactory.GetOrCreate() fails?

orderer/common/multichannel/registrar.go
```diff
@@ -800,3 +840,11 @@ func (r *Registrar) loadJoinBlocks() map[string]*cb.Block {
 
 	return channelToBlockMap
 }
+
+func (r *Registrar) removeJoinBlock(channelID string) error {
+	if err := r.joinBlockFileRepo.Remove(channelID); err != nil {
+		return errors.Wrapf(err, "failed removing joinblock for channel %s", channelID)
```
jyellick, line 899: Why are we wrapping an error we generate? Why not simply 'WithMessage', looking up the tree, it looks like these errors are already generated with a stack.

jyellick (commented):
Looks generally good to me, one nit noted below.  Will wait to approve until other conversations have resolved.

orderer/common/multichannel/registrar.go
wlahti: I guess I've just been a little lazy with differentiating between Wrap and WithMessage lately... plus it's what the rest of the file (and maybe package?) does with the errors. I think it would be worth a single pass through to go with the "standard" we set forth for fabric: use Wrap for errors that come from an external library and use WithMessage for errors that fabric already generated using pkg/errors.

orderer/common/multichannel/registrar.go
tock-ibm: That is a good point. `GetOrCreate` is the last thing in `newLedgerResources` so I am not too worried about that. Hopefully `GetOrCreate`  does not leave an empty ledger behind if it fails. `initLedgerResourcesClusterConsenter` is used in other places as well, in which the ledger is not empty, so we cant effort to cleanup within it as well.
However if  `initLedgerResourcesClusterConsenter` fails after `newLedgerResources`, that requires cleanup in this method, that is correct. 

Here is an idea: how about putting the `defer func()`  from L622 just before `initLedgerResourcesClusterConsenter` (to L617)? 
It might try to remove a non-existing ledger, but that is not an error according to the blockstore_provider.go#L131

wlahti: Done

tock-ibm (approved):

jyellick (approved):

PR #1609, opened 2020-07-16
Extend orderer fileledger for removing channels

wlahti:
#### Type of change

- New feature

#### Description

Extend the orderer's fileledger for removing channels. This uses the recently implemented Drop function on the BlockStoreProvider.

#### Related issues

[FAB-17868](https://jira.hyperledger.org/browse/FAB-17868)


common/ledger/blockledger/fileledger/factory.go
```diff
@@ -26,39 +28,54 @@ type fileLedgerFactory struct {
...
+
+	err := f.blkstorageProvider.Drop(channelID)
+	if err != nil {
+		return err
+	}
+
+	delete(f.ledgers, channelID)
+
+	return nil
+}
```
tock-ibm: We need to make the `Remove` operation crash fault tolerant. The `blkstorageProvider.Drop(channelID)` may take a while, removing the index and then the folder, and if in that time there is a crash there is no recovery from `blkstorageProvider`. We need to mark the channel for removal with the file-repo at the start of this method, and unmark at the end. If we restart and find a channel marked for removal, we deed to retry the `Drop()`. 

This can be done in this PR or under a separate task, as you prefer. I think that since the file-repo init PR is still pending, it makes more sense to do it in a separate task. This way we'll be able to integrate the `Remove` into the Registrar earlier, hardening the implementation later on.

common/ledger/blockledger/fileledger/factory.go
```diff
@@ -26,39 +28,54 @@ type fileLedgerFactory struct {
...
 
-// Close releases all resources acquired by the factory
-func (flf *fileLedgerFactory) Close() {
-	flf.blkstorageProvider.Close()
+// Close releases all resources acquired by the factory.
+func (f *fileLedgerFactory) Close() {
+	f.blkstorageProvider.Close()
 }
 
 // New creates a new ledger factory
```
tock-ibm, line 82: I think the `New(...)` is the right place to initialize the file-repo for removals, and scan and retry `Drop`, thus making `Remove`  CFT.

common/ledger/blockledger/fileledger/factory_test.go
```diff
@@ -7,74 +7,83 @@ SPDX-License-Identifier: Apache-2.0
...
-	_, err = flf.GetOrCreate("testchannelid")
+	_, err = f.GetOrCreate("testchannelid")
 	require.NoError(t, err, "Error GetOrCreate channel")
-	require.Equal(t, 1, len(flf.ChannelIDs()), "Expected 1 channel")
-	flf.Close()
+	require.Equal(t, 1, len(f.ChannelIDs()), "Expected 1 channel")
+	f.Close()
 
-	flf, err = New(dir, metricsProvider)
+	f, err = New(dir, metricsProvider)
```
tock-ibm, line 66: What happens if `New` is called twice without `Close` in the middle?

common/ledger/blockledger/ledger.go
```diff
@@ -17,6 +17,9 @@ type Factory interface {
 	// or creates it if it does not
 	GetOrCreate(channelID string) (ReadWriter, error)
 
+	// Remove removes an existing ledger
+	Remove(channelID string) error
```
tock-ibm, line 21: Add some documentation that this operation is blocking and may take some time, as it cleans the index and then removes the folder.

tock-ibm: We may need to run some benchmarks that will estimate how long this may take for different ledger sizes. This will allow us to decide whether the upper layer will spawn a go-routine to do this or it is enough to do it on the request go-routine.
Moreover, since this takes a lock, we may need to rethink the concurrency scheme here, if it blocks for a long time. Otherwise, we wont be able to GetOrCreate new ledgers during that time.

common/ledger/blockledger/fileledger/factory.go
```diff
@@ -26,39 +28,54 @@ type fileLedgerFactory struct {
...
+	}
+
+	delete(f.ledgers, channelID)
+
+	return nil
+}
+
+// ChannelIDs returns the channel IDs the factory is aware of.
+func (f *fileLedgerFactory) ChannelIDs() []string {
+	channelIDs, err := f.blkstorageProvider.List()
```
tock-ibm, line 70: I recall that the ledger team reported that the blkstorageProvider may report empty ledgers in the `List`. This may happen if the orderer crashes after it created the ledger but before it appends the genesis block to it.  If that is the case, the `New()` below would be a good place to weed those out and remove them.

Unrelated to this PR - I am not sure this should panic... maybe leave the decision to an upper layer?

common/ledger/blockledger/fileledger/factory.go
wlahti: Opened [FAB-18082](https://jira.hyperledger.org/browse/FAB-18082)

common/ledger/blockledger/fileledger/factory_test.go
wlahti: We get a panic:

panic: Error opening leveldb: resource temporarily unavailable

common/ledger/blockledger/ledger.go
wlahti: Updated the comment. The other portion seems worthy of a follow-up task once we're able to actually perform the RemoveChannel operation.

common/ledger/blockledger/fileledger/factory.go
cendhu: Yes. The block store does not handle any failure/recovery for the consumer as it is common to both peer and orderer. Hence, it is responsibility of the consumer (in this case, orderer) to take care of failure/recovery. 

As Yoav mentioned, the `List()` cannot be trusted to return all active ledgers. To avoid misuse of `List()` in the future, we will consider to remove it altogether. A failure could happen either during channel creation or during the removal as well. In both cases, it is the responsibility of the block store consumer to perform cleanup during the reboot. In order to do that, we need to have bookkeeper at the orderer and keep track of ledger status such as `ACTIVE`, `UNDER_CONSTRUCTION`, and `UNDER_DESTRUCTION` either using the existing file repo or a new levelDB instance. During the orderer bootup, a necessary cleanup needs to be performed. We do this at peer here to delete all folders/files/data including the blockstore https://github.com/hyperledger/fabric/blob/add66376d90559d44a45bf61c6185079ac8c37ee/core/ledger/kvledger/kv_ledger_provider.go#L122

wlahti: @cendhu Do you propose we handle that in this PR or would it be fine to do as a follow-up?

cendhu: @wlahti 
```
func (flf *fileLedgerFactory) ChannelIDs() []string {
	channelIDs, err := flf.blkstorageProvider.List()
```
is an existing code and not introduced by this PR. As the goal of the current PR is to add `Remove()` function, I am okay to have a separate PR to handle the failure of both ledger creation & deletion.

common/ledger/blockledger/fileledger/factory_test.go
```diff
@@ -7,74 +7,83 @@ SPDX-License-Identifier: Apache-2.0
...
-
-func (mbsp *mockBlockStoreProvider) Open(ledgerid string) (*blkstorage.BlockStore, error) {
-	return mbsp.blockstore, mbsp.error
-}
-
-func (mbsp *mockBlockStoreProvider) List() ([]string, error) {
-	return mbsp.list, mbsp.error
-}
+func TestBlockStoreProviderErrors(t *testing.T) {
+	mockBlockStore := &mock.BlockStoreProvider{}
```
cendhu, line 22: Nit. It is possible to directly use the block store rather than using a mock but not related to this PR.

common/ledger/blockledger/fileledger/factory_test.go
```diff
@@ -7,74 +7,83 @@ SPDX-License-Identifier: Apache-2.0
...
+
+	f, err = New(dir, metricsProvider)
+	require.NoError(t, err)
+	err = f.Remove("bar")
+	require.NoError(t, err, "Error removing channel")
+	require.Equal(t, 2, len(f.ChannelIDs()))
+	err = f.Remove("this-isnt-an-existing-channel")
+	require.NoError(t, err, "Error removing channel")
+	require.Equal(t, 2, len(f.ChannelIDs()))
+	f.Close()
```
cendhu, line 88: I am not sure whether this test fits in `TestMultiReinitialization()`. Maybe `TestRemove()`. Please fix this in the subsequent PR.

cendhu (approved):

PR #1617, opened 2020-07-19
FAB-18070 Ch.Part.API: refactor to avoid support pattern

tock-ibm:
Avoid using the "support" pattern with the follower.
That is, do not pass a monolithic structure that embeds all
the dependencies of the follower. Instead, pass dependencies
as concrete types that implement individual interfaces that
the follower needs.

Some methods that are needed by both the follower and
consensus.Chains are promoted from `ChainSupport` to
`ledgerResources` along with respective tests.

Change-Id: I09e49951ce60e897b08499ef7ef1bf2a4652f834

#### Type of change

#### Related issues

Task: FAB-18070
Epic: FAB-17712


jyellick (approved):
A significant improvement, thanks!

PR #1619, opened 2020-07-20
FAB-17943 Ch.Part.API: follower pulls blocks until join-block

tock-ibm:
Follower pulls blocks until join-block.

#### Type of change

- New feature

#### Description

Join with a join-block number >0, onboarding is needed.

Start a follower that pulls blocks until the join-block, then mutates into an etcdraft.Chain or a follower.Chain.

The follower.Chain pulls blocks up to the join block, without inspecting intermediate config blocks.
After the join-block is pulled, the follower chain is stopped and either the CreateChain or CreateFollower function is invoked, depending on the channel membership indicated by the join-block.

#### Related issues

Task: FAB-17943
Epic: FAB-17712


orderer/common/follower/follower_chain.go
```diff
@@ -20,7 +24,12 @@ import (
 
 //TODO skeleton
 
-//go:generate mockery -dir . -name LedgerResources -case underscore -output mocks
+const (
+	defaultPullRetryMinInterval time.Duration = 10 * time.Millisecond
```
jyellick: An initial try of 10ms seems a little aggressive to me?  With an exponential backoff base of 1.2, that means we're going to have 17 retries within the first second.

orderer/common/follower/follower_chain.go
```diff
@@ -20,7 +24,12 @@ import (
...
+	defaultPullRetryMaxInterval time.Duration = 30 * time.Second
```
jyellick: I'd also think perhaps 30 seconds is a little short for a max? I'm less concerned here than the 10ms above, but maybe extend this to be 60s?

orderer/common/follower/follower_chain_test.go
```diff
@@ -83,26 +94,321 @@ func TestFollowerNewChain(t *testing.T) {
...
+		}
+		assert.Equal(t, 1, chainCreator.CreateChainCallCount())
+		assert.Equal(t, 0, followerCreator.CreateFollowerCallCount())
+
+		assert.Equal(t, 50, timeAfterCount.getInvocations())
+		assert.Equal(t, int64(5000), timeAfterCount.getMaxDelay())
+	})
+}
+
+type memoryBlockChain struct {
```
jyellick, line 363: For this type, and the one below, can we not have counterfeiter generate stubs, and then bind specific behavior on top? Not a huge savings in code, but some, and it will handle locking for us, give us call counts, etc.

orderer/common/follower/follower_chain_test.go
```diff
@@ -83,26 +94,321 @@ func TestFollowerNewChain(t *testing.T) {
...
+		chain, err = follower.NewChain(mockResources, nil, &common.Block{Header: &common.BlockHeader{}}, options, createBlockPuller, nil, nil, nil)
 		require.EqualError(t, err, "block data is nil")
 		require.Nil(t, chain)
 	})
 
 	t.Run("without join block", func(t *testing.T) {
 		mockResources := &mocks.LedgerResources{}
-		mockResources.On("ChannelID").Return("my-channel")
-		mockResources.On("Height").Return(uint64(5))
+		mockResources.ChannelIDReturns("my-channel")
```
jyellick, line 113: +1 on the conversions from mockery to counterfeiter here!

orderer/common/follower/block_puller.go
```diff
@@ -27,8 +27,43 @@ type ChannelPuller interface {
 	Close()
 }
 
+//
+//// BlockPullerFromJoinBlock creates a new block puller from a join-block.
```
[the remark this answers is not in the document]
tock-ibm: Done

orderer/common/follower/follower_chain.go
tock-ibm: Well, if the concern is the 17 steps from 10ms to 1s, then the min interval is not the only problem, but the base is... so I'll increase both to 50ms and 1.5. This will get us from 50 ms to 60s in ~17 steps.

orderer/common/follower/follower_chain.go

orderer/common/follower/follower_chain_test.go
tock-ibm: I got rid of the `timeAfterCounter` for a counterfeiter mock.

With the `memoryBlockChain` I want to keep it because it gets slightly more complicated in the next PR, and making it mock only will make the code unreadable, I think. Lets revisit that then, if we see it can be converted to purely a mock operated thing I'll do it then.

Note that `memoryBlockChain`  is operated by mocks:  the counterfeiter mocks call two instances of `memoryBlockChain`:

The ledger resources mock calls the instance (`resultChain`) modeling the local blockchain:
```go
		mockSupport = &mocks.LedgerResources{}
		mockSupport.ChannelIDReturns("my-channel")
		mockSupport.HeightCalls(resultChain.Height)
		mockSupport.BlockCalls(resultChain.Block)
		mockSupport.AppendCalls(func(block *common.Block) error {
			resultChain.Append(block)
			return nil
		})
```

whereas the block puller mock calls an instance (`blockchain`) that models the remote blockchain we're pulling from:
```go
	pullerCreator = &mocks.BlockPullerCreator{}
		puller = &mocks.ChannelPuller{}
		puller.PullBlockCalls(func(i uint64) *common.Block { return blockchain.Block(i) })
		pullerCreator.CreateBlockPullerReturns(puller, nil)
```
(lines 150-180)

I renamed them `localBlockchain` and `remoteBlockchain` respectively.

jyellick (approved):
No need to address it in this PR, but it really looks to me like we could simplify things considerably by eliminating the thread management from inside this package, and instead pushing it back out to the caller.

As I read the structure, it looks like this:

1. Caller creates 'NewChain' with a bunch of resources, including two callbacks, one to 'CreateChain', and one to 'CreateFollower'.
2. Caller invokes `Start()` which, spawns a go routine, which is managed inside the follower.
3. The go routine falls into the `run()` loop, which does the task of pulling blocks up to join, then invokes `CreateChain` or `CreateFollower`, which calls back into the original caller from (1).  And in the `CreateFollower` case, calls back into this code again.

If we instead had the caller essentially spawn the go routine that drops into the `run()` loop, it could simply return, with some status, and the caller could appropriately create the chains, or the follower.

I'd even question whether we need to recreate the follower, or if the follower could simply unconditionally replicate blocks through the join block, then continue pulling until the consensus set changed to include it.

In general, I'm just wary of these opaque function pointers which create cyclical dependencies.

```
a calls b calls a
```

is much harder to read than

```
a calls b, b returns and control continues in a
```

orderer/common/follower/follower_chain_test.go
jyellick: Sure, if this makes more sense in a later PR, we can defer and discuss later if necessary.

tock-ibm:
I agree that opaque function pointers make it hard to read. I also prefer using well defined interfaces for the decoupling of cyclical dependencies. That is something I can easily improve. The reason I am using them right now is because I am mimicking some of the existing usage patterns, treating some of the components as "black boxes" (e.g. BlockPuller). As I learn the ropes around this area of the code more options come to mind and there is certainly room for improvement. Another area that I agree can be simplified is that the follower can transition from onboarding to following past the join-block without recreating itself. It turned out this way because of the gradual layered approach to development - first add one functionality (onboarding) then the next (following past). I can harmonize between them for sure, but I prefer to do so after they are both in place. There is one additional PR following this one that completes the refactoring needed by the move from consensus/follower to common follower, and then refactoring the interaction patterns will be much easier.

PR #1624, opened 2020-07-22
FAB-17914 Ch.Part.API: follower pulls blocks after join-block

tock-ibm:
The follower.Chain pulls blocks after the join block and invokes the ChannelCreate function when a config that indicates membership is found.

#### Type of change

- New feature

#### Related issues

Task: FAB-17914
Epic: FAB-17712


orderer/common/follower/follower_chain.go
```diff
@@ -426,6 +426,129 @@ func (c *Chain) pullUpToJoin() error {
...
+		}
+		c.logger.Debugf("Orderer endpoint %s has the biggest ledger height: %d", endpoint, latestHeight)
+
+		if latestHeight <= c.ledgerResources.Height() {
+			c.logger.Debugf("My height: %d, latest height: %d; going to wait %v for latest height to grow",
+				c.ledgerResources.Height(), latestHeight, c.retryInterval)
+			select {
+			case <-c.stopChan:
+				return nil, errors.New("stopped while waiting for latest height to grow")
+			case <-c.timeAfter(c.retryInterval):
```
jyellick: This pattern really doesn't make sense to me here.

If we all the OSNs are up to date for this channel, say... there's no traffic on it, then we begin re-checking with the network for updates every 100ms by default, with an exponential backoff to 60s?  Why the exponential backoff? Why not simply pick a sensible refresh interval?

I'd also note, that the call to `cluster.LatestHeightAndEndpoint` is not necessarily a light one.  This hits all of the OSNs.

This is also a case where creating a common block puller path might make more sense.  Orderers are designed to deliver streams of blocks to peers as they are created, but here, we end up polling the network to look for updates.  We might want to poll periodically to make sure we're not connected to someone who is lagging, but generally, we should just be able to let the blocks arrive, like a peer.

orderer/common/follower/follower_chain.go
```diff
@@ -426,6 +426,129 @@ func (c *Chain) pullUpToJoin() error {
...
+		case <-c.stopChan:
+			c.logger.Warnf("Stopped while pulling blocks: from %d until %d, last pulled %d",
+				firstBlockToPull, latestHeight, nextBlock.Header.Number)
+			return nil, errors.New("stopped while pulling blocks")
+		default:
+			nextBlock = puller.PullBlock(seq)
+			if nextBlock == nil {
+				return nil, errors.Wrapf(cluster.ErrRetryCountExhausted, "failed to pull block %d", seq)
+			}
+			reportedPrevHash := nextBlock.Header.PreviousHash
```
jyellick, line 493: I'm not strictly opposed to these checks, but I'm trying to understand the adversarial model.

If the adversary is malicious, then it's trivial to report a correct previous block hash, and get past this check.

If the adversary is not malicious, and this is a simple namespace intersection, then I would generally expect the signature checking to fail.

The ledger code will not allow blocks which do not chain together to be persisted (it results in a panic).

orderer/common/follower/follower_chain.go
```diff
@@ -426,6 +426,129 @@ func (c *Chain) pullUpToJoin() error {
...
+			if !bytes.Equal(reportedPrevHash, actualPrevHash) {
+				return nil, errors.Errorf("block header mismatch on sequence %d, expected %x, got %x",
+					nextBlock.Header.Number, actualPrevHash, reportedPrevHash)
+			}
+			actualPrevHash = protoutil.BlockHeaderHash(nextBlock.Header)
+			if err := c.ledgerResources.Append(nextBlock); err != nil {
+				return nil, errors.Wrapf(err, "failed to append block %d to the ledger", nextBlock.Header.Number)
+			}
+			c.resetRetryInterval()
```
jyellick: It seems odd to sprinkle timer resets throughout code which does not otherwise interact with the timer.  Would the caller not be a better place to do this?

orderer/common/follower/follower_chain.go
```diff
@@ -426,6 +426,129 @@ func (c *Chain) pullUpToJoin() error {
...
+
+	// Pull until the latest height or a config block
+	for seq := firstBlockToPull; seq < latestHeight; seq++ {
+		select {
+		case <-c.stopChan:
+			c.logger.Warnf("Stopped while pulling blocks: from %d until %d, last pulled %d",
+				firstBlockToPull, latestHeight, nextBlock.Header.Number)
+			return nil, errors.New("stopped while pulling blocks")
+		default:
+			nextBlock = puller.PullBlock(seq)
```
jyellick: Out of curiosity, how is this implemented? I hope that it's not a separate API call per block and leverages an underlying stream instead?

orderer/common/follower/follower_chain.go
tock-ibm: Let's start from the beginning... at the risk of iterating things you most probably already know... 
The cluster.BlockPuller available in the orderer was designed to do on-boarding or state transfer, that is, to efficiently pull blocks that we know exist. So for on-boarding until the join-block, it is good. However it was not designed to wait for a stream of blocks as the delivery client in the peer does. I am aware of that. There were 3 possibilities:
1. write a simple - maybe not so efficient - but working solution on top of cluster.BlockPuller
2. augment the cluster.BlockPuller to behave (also) like the delivery client in the peer
3. write a unified delivery client for both orderer and peer.

I chose (1) because I think that for the main use case we describe in the RFC, the period in which the follower pulls blocks past the join-block is limited in time. An admin will detect that on-boarding had finished (status changes from "onboarding" to "active"), and will re-configure the channel to include the orderer in the consenters set. This period is expected to be limited in time and relatively short. 

The scenario in which an orderer is following the cluster for a long period of time, like a "hot backup", is low priority at the moment, I think. If it turns out that it is a frequent use case, we may need to go to (2) or (3).

Now, with respect to the polling frequency. What is a reasonable frequency? It depends on the rate of block arrival, and on how costly it is to poll. We can certainly put a constant "reasonable" number there, say Tpoll=500ms. The idea behind the backoff is to automatically explore and adapt to the approximate right number. Note that the retryInterval increases on consecutive failed attempts, but resets to the minimum on a successful attempt. On a channel with little traffic, say a block every T seconds, and a minimal polling interval of Tmin, we'll get a success every ~O(log(T/Tmin)), rather than ~(T/Tpoll). What we get is a kind of a saw-tooth  behavior that straddles the right interval. (Another option is to decrease on success, either dividing by the base or subtracting a constant).

orderer/common/follower/follower_chain.go
tock-ibm: I had this discussion with @yacovm  in the previous PR, see:
https://github.com/hyperledger/fabric/pull/1536#discussion_r450500840

Let me sum it up. The BlockPuller works by fetching a batch of blocks from another orderer and verifying that batch. It verifies the hash chain within a batch, but there is no check for the continuity of hashes between batches. Moreover, not all the signatures in a batch are verified. 

I think this needs to be fixed at the cluster package under a different story (see link above for more details). 

Exactly because an orderer panics, we want to be careful with data we get from external parties before we submit it to the ledger. We better have a follower (channel) not making progress than crashing the entire orderer.

orderer/common/follower/follower_chain.go
yacovm: Internally, the block puller sends a seek info of:
```
        orderer.SeekInfo{
		Start:         &orderer.SeekPosition{Type: &orderer.SeekPosition_Specified{Specified: &orderer.SeekSpecified{Number: startSeq}}},
		Stop:          &orderer.SeekPosition{Type: &orderer.SeekPosition_Specified{Specified: &orderer.SeekSpecified{Number: math.MaxUint64}}},
		Behavior:      orderer.SeekInfo_BLOCK_UNTIL_READY,
		ErrorResponse: orderer.SeekInfo_BEST_EFFORT,
	}
```
It then reads blocks from the stream, until its internal in-memory buffer (20MB by default) is full, and then returns the first block read. Subsequent calls will just pop blocks from the buffer until it is empty.

orderer/common/follower/follower_chain.go
tock-ibm: The resets are not "sprinkled"... they are placed after a successful pull, just after Append.
The retryInterval is increased on successive failed attempts in the outer method and reset on the first successful pull. Successful pulls remain in this method (loop), so there is no chance for the outer one to reset in the right time.

orderer/common/follower/follower_chain.go
tock-ibm: On the first BlockPull(n), the BlockPuller will connect to some orderer that has at least block n, and will inquire what is its latest block as well. It will then try to fill a buffer with with as many blocks between [n,latest], up to a maximal buffer size. On the next BlockPull(n+1), it will pop from the buffer, until the buffer is empty. When the buffer is empty, it all starts again. For pulling a batch of blocks, it uses a stream. 

See https://github.com/hyperledger/fabric/blob/92ac4658642c1199862162d02d42b2730cd1d509/orderer/common/cluster/deliver.go#L87

For pulling blocks that we know exist (on-boarding), this is pretty efficient. For following past the join-block... not so much... it really depends on the block rate and how much we wait before a retry. The saw-tooth type backoff actually helps here, as we eventually get a larger batch of blocks rather than pull them one by one.

orderer/common/follower/follower_chain.go
tock-ibm: I added different min & max interval defaults for the case when the follower pulls past the join block. Grows from 500ms to 10s with base 1.5. I think this is a reasonable compromise between not polling too fast and staying responsive to incoming blocks.

orderer/common/follower/follower_chain.go
jyellick: I think part of the issue I'm having with this function is that it acts both by return value, and by side effect. 

You can see up on line 459:

```
		￼configBlock, err = c.pullUntilLatestOrConfig(puller, latestHeight)
￼		if err != nil {
￼			return nil, errors.Wrap(err, "failed to pull until latest height or config")
￼		}
￼		c.resetRetryInterval()
```

Which, implies to me that `pullUntilLatestOrConfig` returns a config block, or nil, or an error.  And, based on the success or failure of this function, we reset the timer, or not.

But, it's actually not possible to return without error without already having reset the timer in the function itself.  So, why are we resetting it again on the outside?

Similarly, inside, we are setting the member `c.lastConfig`, and then we are also returning what we just set as the first return value of this function.

I suspect it's an artifact of starting from the consenter interface, but the follower to me seems like it should have very simple control flow, but it's very hard for me to get it from this code.

1. Pull blocks, until the latest known height.  On error, try again with some backoff logic.
2. Check to see if we're in the consenter set, and if we are, exit.
3. Recheck the block height periodically, until it advances beyond where we are, go to 1.

It seems to me that there's a lot of other behavior here.  We've got a retry timer that's used in the main run loop, which is manipulated throughout the functions it calls into, but not in ways that are necessarily clear.  If an error is returned, is the timer reset? Maybe.  The selected block pulling function is done by function pointer, the timeouts are adjusted based on the pulling type.  To me, this feels like it should be considerably less and much more clear code.

jyellick (commented):
So I spent quite a bit of time, digging through this, trying to grok exactly how things work, and I think I have a reasonable understanding of the control flow, and I'm really not loving the complexity I see here.

I'd much rather see a simple control loop, one that pulls blocks until a target (with an exponential backoff), checks for membership, and then periodically checks to see if it should re-loop back to pulling blocks again.

I think the pieces are here, but, it's all obfuscated behind function pointers, side effects, timer re-use, and some additional behavior.  For instance, why do we care about returning when we see a config block? Why not just finish replicating before we check membership?  The replication will need to be done regardless before we begin consenting.

orderer/common/follower/follower_chain.go
yacovm: I think the problem Yoav is trying to protect against, is that an orderer has a block with intact block data and header, but incorrect signatures due to storage corruption. 
This orderer isn't aware the signatures of its block are corrupt and can silently replicate this block to other orderers via either Raft replication or onboarding.
Further assume that all orderers with intact signatures on that block are removed from the channel, and you get a channel with only orderers that have corrupt signatures on that block, and then peers that join the network and only replicate from orderers, are stuck.

One way of solving this, is to check signature in every block. 
Another way would be to have the storage layer (the common ledger) use methods to detect corruption of the block when it reads it from the file system.

tock-ibm:
I update the commit:

- Simplified the control flow of block pulling after the join block in line with suggestions above.
- Pulling until the join block now flows past the join block if it is not a member, and continues to pull until membership.
- The callback to create a follower is eliminated. 
- The retry interval while pulling after the join block is increased exponentially on errors or when the latest height does not grow, and decreases linearly upon a successful pull until the latest block. This will approximate the rate of block arrival more closely.

orderer/common/follower/follower_chain.go
jyellick: @yacovm I don't think this check addresses anything with signatures of metadata? It is only checking the block header hash, which, is unconnected to the metadata?

Assuming that we have some correct signature over a future block which chains to this one, how do we get a hash mismatch to what's on our chain? The only way I see, is if someone accidentally re-used crypto material and had a channel ID conflict.  Basically, two blockchains with the same orderers, with the same name, but different contents.  We can protect against this... just trying to understand the point of the check.

orderer/common/follower/follower_chain.go
```diff
@@ -423,9 +413,130 @@ func (c *Chain) pullUpToJoin() error {
...
+		c.decreaseRetryInterval() // After a successful pull, we slowly decrease the retry interval.
+	}
+
+	return nil
+}
+
+// pullUntilLatestOrMembershipConfig is given a height-target and exits without an error when:
+// (a) it reaches that target, returning "false"; or
+// (b) it encounters a config block that indicates channel membership, returning "true".
+func (c *Chain) pullUntilLatestOrMembershipConfig(latestHeight uint64) (bool, error) {
```
jyellick: Why are we pulling only until a membership config? Why not simply until a target block height?  Obviously the membership could change along the way, but why stop replicating?  Replication happens no matter what before we can begin consenting.

orderer/common/follower/follower_chain.go
```diff
@@ -242,20 +210,28 @@ func (c *Chain) StatusReport() (types.ClusterRelation, types.Status) {
 	c.mutex.Lock()
 	defer c.mutex.Unlock()
 
-	clusterRelation := types.ClusterRelationFollower
+	return c.cRel, c.status
+}
+
+func (c *Chain) updateStatusReport() {
```
jyellick: Unfortunately, I don't think this is the solution to the control flow problem.  We're still invoking this status update in three different spots I count in the flow.

What I'm really looking for is a top level control loop, that invokes an action, waits for that action to return, and then sets a status based on the result.

I would expect something that looks like this:

```
targetHeight := c.joinBlock.SeqNo
for {
    if c.ledger.Height() < targetHeight {
        if err := replicateThrough(targetHeight); err != nil {
            // exponential backoff
            continue
        }
        // reset backoff
    }

    // at this point, we are at our replication target, set status
    // check membership, and exit replication if needed

    // at this point, we are still replicating, so wait:
    for {
        networkHeight, err := c.getNetworkHeight()
        if err != nil || networkHeight <= c.ledger.Height() {
            // If we really wanted to be ambitious, we could exponentially backoff in the error case
            time.Sleep(sometime)
            continue
        }

        targetHeight = networkHeight
        break
    }
}
```

jyellick (commented):
Perhaps I'm getting hung up on this particular PR, and we could iteratively improve things, but I still don't care for the structure of the follower control flow.

orderer/common/follower/follower_chain.go
yacovm: > @yacovm I don't think this check addresses anything with signatures of metadata? It is only checking the block header hash, which, is unconnected to the metadata?

I was never referring to the code, but to what Yoav wrote in his comment:
>  Moreover, not all the signatures in a batch are verified.

> Assuming that we have some correct signature over a future block which chains to this one, how do we get a hash mismatch to what's on our chain?

We don't get a hash mismatch... I am saying that if you don't check every signature then if there is an orderer that has a non-crash type of error (corrupt signature but intact block), under certain conditions and execution, this corrupt signature will propagate to all nodes in the system.

jyellick: Okay.  So, we are discussing two things here:

1. In the code that this thread references, we are checking that the previous block hash matches the block that is committed onto our ledger.
2. In the block replication code, for efficiency, we are not checking every signature, but rather only checking the last signature, and then verifying the hash chain.  But, this will not detect corruption in metadata.

For (1), the situation this code is defending against _should not_ ever occur.  As it implies two different valid blocks of the same sequence number (as they must both be covered by a valid signature, either directly or by chaining).  And we have a fork in our blockchain.  It might still be worth defending against, because we would rather not have the whole orderer crash, but importantly, this is not a security check, as security guarantees have already been violated if this check fails.  Instead, this is more of a sanity check in case of operator incompetence.

For (2), this does seem like a problem of sorts.  In the CFT case, it requires corruption, but in the eventual BFT case, obviously we'd need something more robust.  But as @tock-ibm says, an issue to be handled separately from this series.

yacovm: For (2) I suggest that we solve this problem universally in the common ledger by appending [CRC](https://golang.org/pkg/hash/crc32/) to every block we write into the filesystem and when we retrieve it, we validate the tag matches the CRC of the block.

orderer/common/follower/follower_chain.go

orderer/common/follower/follower_chain.go
tock-ibm: The height polling policy and the pull retry policy were separated in the last commit. 
All retries are now in function scope for readability.

orderer/common/follower/follower_chain.go
tock-ibm: opened https://jira.hyperledger.org/browse/FAB-18104 for this issue.

orderer/common/follower/follower_chain.go
tock-ibm: Addressed. Never dropping blocks.

orderer/common/follower/follower_chain.go
tock-ibm: Addressed in latest commit.

tock-ibm:
@jyellick Following our discussion on the life-cycle of the follower.Chain, I opened this story: https://jira.hyperledger.org/browse/FAB-18106

orderer/common/follower/follower_chain.go
tock-ibm: Addressed in the last commit.

orderer/common/follower/follower_chain.go
```diff
@@ -286,146 +255,277 @@ func (c *Chain) run() {
...
+		if isMember {
+			c.setStatusReport(types.ClusterRelationMember, types.StatusActive)
+			return nil
+		}
+
+		// Poll for latest network height to advance beyond ledger height.
+		var latestNetworkHeight uint64
+	heightPollLoop:
+		for {
+			endpoint, networkHeight, errHeight := cluster.LatestHeightAndEndpoint(c.blockPuller)
```
jyellick, line 397: So I suppose this is why we are creating the block puller externally, instead of inside of the `pullUntilLatest` with retry.  But I'm not convinced that we need this call here.

Why not have `pullUntilLatest` simply pull until the latest, in both paths?  As I've mentioned in other places in the review, I absolutely don't see why it's critical that we stop pulling after we get to the 'Join Block'.  We should replicate the chain up to its highest height, and then decide if we are members.

Then essentially, in the 'pullUntilJoin' path, you would create a block puller from the join block, and let it go.  In the 'pullAfterJoin' path, you would look at the latest config, create the block puller, and let it go.

orderer/common/follower/follower_chain.go
```diff
@@ -286,146 +255,277 @@ func (c *Chain) run() {
...
 					nextBlock.Header.Number, actualPrevHash, reportedPrevHash)
 			}
 			actualPrevHash = protoutil.BlockHeaderHash(nextBlock.Header)
-			if err = c.ledgerResources.Append(nextBlock); err != nil {
-				return errors.Wrapf(err, "failed to append block %d to the ledger", nextBlock.Header.Number)
+			if err := c.ledgerResources.Append(nextBlock); err != nil {
+				return n, errors.Wrapf(err, "failed to append block %d to the ledger", nextBlock.Header.Number)
+			}
+
+			if inspectConfig && protoutil.IsConfigBlock(nextBlock) {
```
jyellick, line 503: To confirm my understanding of why this is safe:

The block puller is created only using the orderer endpoints, as defined in the channel config.  Internally, the block puller through, magic or other means, uses the correct channel config for performing block signature validation.

Here, we are performing a secondary check, that, if the block we just retrieved changes the channel config, we are using this new channel config in order to generate a new block puller with more recent updates.

In the pullFromJoin path, we set `inspectConfig` to false, we already have 'more up to date' endpoints, but in the `pullAfterJoin` path, we set `inspectConfig` to true, to ensure that we re-construct the block puller with the most up to date endpoints.

Assuming that's all correct, what I'd like some clarification on, is if this is strictly necessary, and if this has any performance implications.  If we just got a config block, from the set of endpoints we know about, then it would imply to me, that the set of endpoints we currently know about is adequate / working.  In particular, I assume that the block puller, once connected, is pulling from a single endpoint, and because we just received a block, if we were to continue without recreating the block puller, nothing would break.  Is this because the block puller internally has error handling that will seek out and fail over to another node if this one fails?

jyellick (commented):
Thanks for the updates Yoav, I think we are getting closer, but I still see some points of potential simplification.

orderer/common/follower/follower_chain.go
yacovm: The current onboarding or Raft replication doesn't re-create the block puller upon [ingestion of a config block](https://github.com/hyperledger/fabric/blob/master/orderer/common/cluster/util.go#L535-L561). This is not necessary because the block puller [knows to update its internal bundle and PolicyManager](https://github.com/hyperledger/fabric/blob/87491e68c289b18f871527113c356afdaef36484/orderer/common/cluster/util.go#L617-L635) in-flight and autonomously. 
When you create a new block puller, it takes the bundle and PolicyManager that is the latest persisted on disk.
Then, as it eats config blocks, [it replaces this bundle and policy manager](https://github.com/hyperledger/fabric/blob/87491e68c289b18f871527113c356afdaef36484/orderer/common/cluster/util.go#L593-L606) reference.

orderer/common/follower/follower_chain.go
```diff
@@ -286,146 +255,277 @@ func (c *Chain) run() {
...
+	//assuming this will never overflow int64, as upperLimit cannot be over MaxInt64/2
+	*retryInterval = time.Duration(1.5 * float64(*retryInterval))
+	if *retryInterval > upperLimit {
+		*retryInterval = upperLimit
+	}
+	c.logger.Debugf("retry interval increased to: %v", *retryInterval)
+}
+
+func (c *Chain) resetRetryInterval(retryInterval *time.Duration, lowerLimit time.Duration) {
+	if *retryInterval == lowerLimit {
```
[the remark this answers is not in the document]
tock-ibm: Yes I only want to emit the log message if the value changes...
If I unconditionally set the retryInterval to the lower limit, how do I then detect that it changed? I'd have to introduce a local variable and check against that. I don't see how that is any better... I don't see a problem with this method...

orderer/common/follower/follower_chain.go
```diff
@@ -286,146 +255,277 @@ func (c *Chain) run() {
...
-			c.followerCreationCallback()
+			return nil
 		}
-	} else {
-		c.logger.Info("Block pulling finished successfully, invoking chain creation callback")
-		c.chainCreationCallback()
+
+		c.setStatusReport(types.ClusterRelationFollower, types.StatusActive)
+		c.joinBlock = nil
+		err = c.loadLastConfig()
```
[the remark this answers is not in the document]
tock-ibm: config is needed on entry to the pullAfterJoin method. It is initialized in the constructor in case we start the follower without a join block (say after restart) and then the only other place it needs to be set is during  the transition from until-join to after-join.

orderer/common/follower/follower_chain.go
```diff
@@ -286,146 +255,277 @@ func (c *Chain) run() {
...
 // pullUpToJoin pulls blocks up to the join-block height without inspecting membership on fetched config blocks.
 // It checks whether the chain was stopped between blocks.
 func (c *Chain) pullUpToJoin() error {
 	targetHeight := c.joinBlock.Header.Number + 1
-	firstBlockToPull := c.ledgerResources.Height()
-	c.logger.Debugf("first block to pull: %d, target height: %d", firstBlockToPull, targetHeight)
-	if firstBlockToPull >= targetHeight {
-		c.logger.Infof("Target height (%d) is <= to our ledger height (%d), skipping pulling", targetHeight, firstBlockToPull)
+	if c.ledgerResources.Height() >= targetHeight {
+		c.logger.Infof("Target height (%d) is <= to our ledger height (%d), skipping pulling", targetHeight, c.ledgerResources.Height())
```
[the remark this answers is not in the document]

orderer/common/follower/follower_chain.go
```diff
@@ -286,146 +255,277 @@ func (c *Chain) run() {
...
 		return nil
 	}
 
-	puller, err := c.createBlockPullerFunc()
+	var err error
+	c.blockPuller, err = c.createBlockPullerFunc() // TODO replace with a factory that has a join block as input
```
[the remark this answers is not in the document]
tock-ibm: See answer below.

orderer/common/follower/follower_chain.go
tock-ibm: Both paths call `pullUntilLatestWithRetry` only. `pullUntilLatestWithRetry` calls `pullUntilLatest`. I could have consolidated both these function in to one, but that would have created deep nesting and a long function, longer than my screen can show. 

The two paths require slightly different block pullers, see full comment below.

orderer/common/follower/follower_chain.go
tock-ibm: Your understanding is mostly correct. Let me shed some additional light and answer some of the questions.

Here is the overall desired behavior:
1. When pulling until the join block, the puller should take the most updated endpoints (from the join block), but use a block signature verification policy from the config block just prior to the block is it trying to verify. We cannot verify block `m` with the channel config from join-block `n` if `m<n` (it might work for degenerate cases, like our current integration tests, but it's wrong). Let's call this component the `blockVerifier`. This means that the `blockVerifier` need to be built from the genesis block up, until we reach the join block, but the endpoints are just from the join-block.
2. When pulling past the join block, we have to update the `blockVerifier` with every new channel config, and as we do so, we might as well do the same for the endpoints.

Your comment about the "magic" by which the block puller updates the channel config for signature verification sent me checking (again) on how this is currently implemented in components using the block puller in fabric. 
The conclusion is, that this does not happen "internally" to the block puller but rather on an upper layer.

- The etcdraft.Chain, in the [catchUp code](https://github.com/hyperledger/fabric/blob/9c705811e9d2dbda6a7876d0065d65a369a62ca5/orderer/consensus/etcdraft/chain.go#L879), provides the "support" to the block puller. This has the `ledgerResources` which has the `VerifyBlockSignature` that the `blockVerifier` uses, but also `mutableConfig`, which exposes the `Update(bundle)`. The chain then calls the block-writer `WriteConfigBlock` on every new config, which internally updates the channel config after commit. Note that the block puller is built from an opaque function pointer, which is in fact a closure defined in the consenter.
- The `onboarding` package that implements the `InactiveChainReplicator` does it differently - it maintains a separate `blockVerifier` for every channel, and updates it after every `Append`. (see the [verifierLoader](https://github.com/hyperledger/fabric/blob/9c705811e9d2dbda6a7876d0065d65a369a62ca5/orderer/common/onboarding/onboarding.go#L358)  and the [VerifierFactory](https://github.com/hyperledger/fabric/blob/9c705811e9d2dbda6a7876d0065d65a369a62ca5/orderer/common/cluster/util.go#L493) and the [BlockVerifierAssembler](https://github.com/hyperledger/fabric/blob/9c705811e9d2dbda6a7876d0065d65a369a62ca5/orderer/common/cluster/util.go#L587) from the `cluster` package) .

One of the main differences is that the `etcdraft.Chain` has a `support`  with the `ledgerResources` & `blockWriter` in it, whereas the `onboarding.Inactive*` does not. 

My original plan was to use different block pullers, constructed by the Registrar, depending on whether we construct the follower pre- or post-join. Hence the callback to recreate the follower to a new follower in the previous version. Specifically, after join, we can use the ledgerSources and its ability to `Update` to keep track of channel config. For pulling before the join, my plan was to use the block puller more like the `onboarding.Inactive*`, since we build the ledgerResources from the join block, but pull config blocks from the past. However, your comment about continuing directly from pre-join to post-join changed that.

Currently, all that wiring is not entirely in place. The `blockVerifier` given to the puller is based on the join block and is not updated. This will work is tests when channel config does not significantly change, but break otherwise. My plan is to put all that in place in the next [task / PR](https://jira.hyperledger.org/browse/FAB-18086), which I have already started. 

With this in mind, the way I would like to proceed is the following. I propose to express the rest of the comments on this PR in a Jira (say ["Improving follower flow control"](https://jira.hyperledger.org/browse/FAB-18108)), and address the issues around [wiring](https://jira.hyperledger.org/browse/FAB-18086) first. After those issues are nailed down, and the puller interface and usage patterns had stabilized, it will be much easier to implement the suggestions for flow control consolidation brought here. 

Attending to the "external" functionality first would not only simplify the internal refactoring later, it would also allow the rest of the team to execute in parallel some of the tasks that are blocked by this one, for example integration testing. Refactoring internals with tests in place is also a good idea.

orderer/common/follower/follower_chain.go
```diff
@@ -286,146 +255,277 @@ func (c *Chain) run() {
...
-		}
-	}
-
-	// Trigger creation of a new chain. This will regenerate 'ledgerResources' from the tip of the ledger, and will
-	// start either a follower.Chain or an etcdraft.Chain, depending on whether the orderer is a follower or a
-	// member of the cluster.
-	c.halt()
 
-	if c.joinBlock != nil {
+		// The join block never returns an error. This is checked before the follower is started.
```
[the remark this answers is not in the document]
tock-ibm: see comment below

orderer/common/follower/follower_chain.go
jyellick: Yes, sloppy commenting on my part "`pullUntilLatest` with retry" should have been `pullUntilLatestWithRetry`.  No qualms with splitting that particular function into two pieces.  I was simply suggesting that there was no need to parameterize the function, that we could have it simply query the latest height, and then pull until this height.

orderer/common/follower/follower_chain.go
jyellick: > Attending to the "external" functionality first would not only simplify the internal refactoring later, it would also allow the rest of the team to execute in parallel some of the tasks that are blocked by this one, for example integration testing. Refactoring internals with tests in place is also a good idea.

My feeling is still that we could drastically simplify this now.  But, I don't want to block progress longer than I already have.  I'll approve so that we can get things moving along, but I would like to make sure we revisit this in the future.

jyellick (approved):

PR #1628, opened 2020-07-22
Separate internal halt from Halt for etcdraft.Chain

wlahti:
#### Type of change

- New feature

#### Description

Stops a chain but does not call the haltCallback function, which is used to transfer responsibility to a follower or
inactive chain. This will be used by the channel participation API when removing a channel.

#### Related issues

[FAB-18069](https://jira.hyperledger.org/browse/FAB-18069)


orderer/consensus/etcdraft/chain.go
```diff
@@ -434,6 +436,14 @@ func (c *Chain) Halt() {
 	}
 }
 
+// HaltExternal stops the chain but does not call the haltCallback function.
+// This is intended to be used by the channel participation API when removing
+// a channel.
+func (c *Chain) HaltExternal() {
+	c.haltCallback = nil
```
yacovm: Isn't this callback an empty callback in case of a channel participation API?

wlahti: That does appear to currently be the case, yes, but it's my understanding that will change shortly to allow it to switch to a follower/inactive chain as applicable.

yacovm: But I thought the long term goal is to not have any inactive chain? @tock-ibm  ?

tock-ibm:
This approach looks good.

jyellick (approved):
I'd like to note for the record, that this control flow still seems very awkward to me.  The registrar is calling into the consenter, which as it halts is calling back into the registrar.  Any time we have bidirectional binding like this is a red flag in my opinion.

I'd much rather see this implemented in one of two ways:

1. Leave the internal run state management to the consenter plugins as is done today, and expose a channel or similar mechanism to allow the the registrar to detect when the consenter has exited.
2. Remove the 'start' / 'halt' bits of the consenter API, and instead require that the caller inject a go routine, and simply have control flow return once the consenter exits, indicating that a shutdown occurred.

I think (2) is probably easier to read and more natural, while (1) is less of an API change and less code.  In either case, I think both are better options than an opaque "I'm stopping" callback.

wlahti:
> I'd like to note for the record, that this control flow still seems very awkward to me. The registrar is calling into the consenter, which as it halts is calling back into the registrar. Any time we have bidirectional binding like this is a red flag in my opinion.
> 
> I'd much rather see this implemented in one of two ways:
> 
> 1. Leave the internal run state management to the consenter plugins as is done today, and expose a channel or similar mechanism to allow the the registrar to detect when the consenter has exited.
> 2. Remove the 'start' / 'halt' bits of the consenter API, and instead require that the caller inject a go routine, and simply have control flow return once the consenter exits, indicating that a shutdown occurred.
> 
> I think (2) is probably easier to read and more natural, while (1) is less of an API change and less code. In either case, I think both are better options than an opaque "I'm stopping" callback.

Sorry for the delayed response here. Opened https://jira.hyperledger.org/browse/FAB-18152 to address this.

PR #1629, opened 2020-07-23
[FAB-17906] Add snapshot mgmt APIs to ledger_interface

wenjianqiao:

#### Type of change

- New feature

#### Description
Add snapshot mgmt APIs to ledger_interface.

#### Additional details
Because some packages depend on ledger.PeerLedger interface, we have to re-generate the mock files whenever PeerLedger interface is changed. This is inefficient because we would not know which mocks need to be updated until running tests or basic-checks. To avoid the problem, ideally we should define interface at the caller side when a mock is needed. This requires refactoring in multiple components and will be handled in a separate PRs.

#### Related issues

### DIFF
--- a/core/chaincode/mock/peer_ledger.go
+++ b/core/chaincode/mock/peer_ledger.go
@@ -11,6 +11,17 @@ import (
 )
 
 type PeerLedger struct {
+	CancelSnapshotRequestStub        func(uint64) error
+	cancelSnapshotRequestMutex       sync.RWMutex
+	cancelSnapshotRequestArgsForCall []struct {
+		arg1 uint64
+	}
+	cancelSnapshotRequestReturns struct {
+		result1 error
+	}
+	cancelSnapshotRequestReturnsOnCall map[int]struct {
+		result1 error
+	}
 	CloseStub        func()
 	closeMutex       sync.RWMutex
 	closeArgsForCall []struct {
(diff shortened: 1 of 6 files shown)


wenjianqiao:
Hit a gossip flake: https://jira.hyperledger.org/browse/FAB-17451

manish-sethi:
@wenjianqiao - I am not sure if this piecemeal refactoring is buying us anything. For instance, for the validation package you refactored, it seems that though the real dependency is only one function to check the presence of duplicate transaction but in the tests, the real ledger is being used and hence other functions and in particular tx simulator (which is also an interface) etc. also gets dragged in. Any change in those interfaces will take back to the problem that you were trying to solve. BTW, not sure why you needed to generate the mocks for tx simulator when the tests in this package are using real impl.

I suggest that we take up this larger change separately later with removing the ledger interface and directly exposing KVLedger implementation... like we did with other interfaces - though, in the case of ledger interface, it's going to take lot more relatively.

wenjianqiao:
Discussed in slack with @manish-sethi and @cendhu. Because PeerLedger interface is passed to multiple components and PeerLedger itself can return other interfaces, it is hard for users to define local interfaces when mocks are needed for testing. One proposal is to use KVLedger struct instead of PeerLedger interface all over the components, which may also require code refactoring. This improvement will be addressed separately.

manish-sethi (approved):

PR #1633, opened 2020-07-23
Stop pulling Builder and Runtime Images

lindluni:
Since Builder and Runtime images now diverge
on versions, we will allow the the Peer
to pull them at runtime using the tag set in
core.yaml

### DIFF
--- a/scripts/bootstrap.sh
+++ b/scripts/bootstrap.sh
@@ -104,12 +104,8 @@ pullDockerImages() {
     if [ "${NODOCKER}" == 0 ]; then
         FABRIC_IMAGES=(peer orderer ccenv tools)
         case "$VERSION" in
-        1.*)
-            FABRIC_IMAGES+=(javaenv)
-            shift
-            ;;
         2.*)
-            FABRIC_IMAGES+=(nodeenv baseos javaenv)
+            FABRIC_IMAGES+=(baseos)
             shift
             ;;
         esac


scripts/bootstrap.sh
```diff
@@ -103,16 +103,6 @@ pullDockerImages() {
     NODOCKER=$?
     if [ "${NODOCKER}" == 0 ]; then
         FABRIC_IMAGES=(peer orderer ccenv tools)
-        case "$VERSION" in
-        1.*)
-            FABRIC_IMAGES+=(javaenv)
-            shift
-            ;;
-        2.*)
-            FABRIC_IMAGES+=(nodeenv baseos javaenv)
```
denyeart, line 112: Since baseos is built from fabric repo, we can keep that one here. It is used as the basis for Go chaincode images.

denyeart (changes requested):

scripts/bootstrap.sh
lindluni: I'll add it back, but in general I just don't see the point. The tag is set by us in core.yaml, there is no harm in just letting the peer pull it, since this is the same behavior for the other two languages, it made sense to me to keep the behavior consistent.

lindluni: Added it back

denyeart (approved):

denyeart (approved):

WaleedMortaja:
@denyeart should this be backported to release-2.2?

denyeart:
You can backport if you have a need. But typically we don't make updates to bootstrap.sh in back releases since the download instructions always point you to bootstrap.sh in the main branch.

WaleedMortaja:
Then, there is no need to backport. Okay.

PR #1640, opened 2020-07-25
Update images to remove reference to 'marble' for new samples

denali49:
Signed-off-by: Chris Gabriel <chris_gabriel_98@yahoo.com>

With the new samples, the images in the tutorial should now reference a generic asset named 'asset1' or 'asset', not 'marble'.

#### Type of change

- Documentation update

#### Description

Update docs per work agreed in the fabric-samples working group.


nikhil550 (approved):

nikhil550:
If you could check in the power point (if it is not already there), in a future pr, that would help users and writers make future edits to the images.

denali49:
Will do, I agree it is best practice include the full pptx to make future edits easier.

PR #1643, opened 2020-07-27
FAB-18086 Ch.Part.API: Improve wiring between Registrar and follower

tock-ibm:
#### Type of change
- New feature

#### Description

Currently several opaque function pointers are used to decouple the multichannel.Registrar and the follower.Chain. In this commit:
    
Convert opaque function pointers to well defined interfaces that are easy to follow and mock for tests.
    
The BlockPuller creation mechanisms is converted to a factory struct that receives all the prerequisites except the config-block, and exposes a method that builds a new BlockPuller given a config block. This allows us to use the same mechanisms for on-boarding, until the join block, and for following past the join-block.
    
The follower's function pointers that are used to create a consensus.Chain and a follower.Chain are replaced by an interface that defines these methods. This interface is implemented by the registrar and and is set into the new follower instance at construction time.
    
The method in the registrar to create a follower from the tip of the ledger is defined and implemented. This method is used in the follower life-cycle (transition from on-boarding to following) and when initializing follower chains after restart.

#### Related issues

Task: FAB-18086


tock-ibm:
Replaced by #1693

PR #1650, opened 2020-07-27
[FAB-17906] Synchronize commits and snapshots with channels and events

wenjianqiao:

#### Type of change
- New feature

#### Description
Implement a processEvents function to synchronize block commits, snapshot generations, and snapshot requests
using channels and events. When a ledger is created/opened, it invokes processEvents in a separate go-routine
so that it will run in background to process the commit and snapshot events one by one.

#### Related issues

### DIFF
--- a/core/ledger/kvledger/snapshot_mgmt.go
+++ b/core/ledger/kvledger/snapshot_mgmt.go
@@ -8,16 +8,40 @@ package kvledger
 
 import (
 	"math"
+	"os"
 
 	"github.com/hyperledger/fabric/common/ledger/util"
 	"github.com/hyperledger/fabric/common/ledger/util/leveldbhelper"
 	"github.com/pkg/errors"
 )
 
+const (
+	commitStart   = "commitStart"
+	commitDone    = "commitDone"
+	requestAdd    = "requestAdd"
+	requestCancel = "requestCancel"
+	snapshotDone  = "snapshotDone"
+)
+
 var (
 	snapshotRequestKeyPrefix        = []byte("s")
 	defaultSmallestHeight    uint64 = math.MaxUint64
 )
+
+type snapshotMgr struct {
+	snapshotRequestBookkeeper *snapshotRequestBookkeeper
+	events                    chan *event
+	commitProceed             chan struct{}
+	requestResponses          chan *requestResponse
+}
+
+type event struct {
+	typ    string
+	height uint64
+}
+type requestResponse struct {
+	err error
+}
 
 // SubmitSnapshotRequest submits a snapshot request for the specified height.
 // The request will be stored in the ledger until the ledger's block height is equal to
@@ -26,7 +50,9 @@ var (
 // It returns an error if the specified height is smaller than the ledger's block height
 // or the requested height already exists.
 func (l *kvLedger) SubmitSnapshotRequest(height uint64) error {
-	return errors.New("not implemented")
+	l.snapshotMgr.events <- &event{requestAdd, height}
+	response := <-l.snapshotMgr.requestResponses
+	return response.err
 }
 
 // CancelSnapshotRequest cancels the previously submitted request.
@@ -36,13 +62,14 @@ func (l *kvLedger) SubmitSnapshotRequest(height uint64) error {
 // It also locks snapshotGenerationRWLock to synchronize with commit before getting
 // blockchain info so that it will not cancel a under-processing request.
 func (l *kvLedger) CancelSnapshotRequest(height uint64) error {
-	return errors.New("not implemented")
+	l.snapshotMgr.events <- &event{requestCancel, height}
+	response := <-l.snapshotMgr.requestResponses
+	return response.err
 }
 
 // PendingSnapshotRequests returns a list of heights for the pending (or under processing) snapshot requests.
 func (l *kvLedger) PendingSnapshotRequests() ([]uint64, error) {
-	return nil, errors.New("not implemented")
-
+	return l.snapshotMgr.snapshotRequestBookkeeper.list()
 }
 
 // ListSnapshots returns the information for available snapshots.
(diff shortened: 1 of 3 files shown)


core/ledger/kvledger/snapshot_mgmt_test.go
```diff
@@ -110,3 +117,198 @@ func TestSnapshotRequestBookKeeperErrorPaths(t *testing.T) {
...
+
+	// commit blocks until block height is 10
+	lastBlock := testutilCommitBlocks(t, l, bg, 10, gbHash)
+	// wait until snapshot height=5 is generated
+	snapshotExists := func() bool {
+		exists, err := kvledger.snapshotExists(5)
+		require.NoError(t, err)
+		return exists
+	}
+	require.Eventually(t, snapshotExists, time.Minute, 3*time.Second)
```
manish-sethi: I didn't get this... why it needs to be `Eventually`? when the call on line 141 returns that would mean 10 blocks have been committed which indirectly means that the snapshot at height 5 should be guaranteed to be generated.

core/ledger/kvledger/snapshot_mgmt_test.go
```diff
@@ -110,3 +117,198 @@ func TestSnapshotRequestBookKeeperErrorPaths(t *testing.T) {
...
+
+	// Test 4: commit blocks, verify snapshots are generated and pending snapshot requests are updated
+	testutilCommitBlocks(t, l, bg, 20, protoutil.BlockHeaderHash(lastBlock.Header))
+	// wait until snapshots for height=15 and height=20 are generated
+	snapshotExists = func() bool {
+		exists, err := kvledger.snapshotExists(15)
+		require.NoError(t, err)
+		return exists
+	}
+	require.Eventually(t, snapshotExists, time.Minute, 3*time.Second)
```
manish-sethi: same as above.

core/ledger/kvledger/snapshot_mgmt_test.go
```diff
@@ -110,3 +117,198 @@ func TestSnapshotRequestBookKeeperErrorPaths(t *testing.T) {
...
+
+	// wait until previous snapshotDone events are processed
+	requestsUpdated = func() bool {
+		requests, err := l.PendingSnapshotRequests()
+		require.NoError(t, err)
+		return equal(requests, []uint64{50, 100})
+	}
+	require.Eventually(t, requestsUpdated, time.Minute, time.Second)
```
manish-sethi: same as above

core/ledger/kvledger/snapshot_mgmt_test.go
```diff
@@ -110,3 +117,198 @@ func TestSnapshotRequestBookKeeperErrorPaths(t *testing.T) {
...
+	snapshotExists = func() bool {
+		exists, err := kvledger.snapshotExists(10)
+		require.NoError(t, err)
+		return exists
+	}
+	require.Eventually(t, snapshotExists, time.Minute, 3*time.Second)
+
+	// Test 3: submit multiple requests in parallel, verify PendingSnapshotRequest and smallest request height
+	for _, height := range []uint64{100, 15, 50, 20} {
+		go l.SubmitSnapshotRequest(height)
```
manish-sethi: `require.NoError`

core/ledger/kvledger/snapshot_mgmt_test.go
```diff
@@ -110,3 +117,198 @@ func TestSnapshotRequestBookKeeperErrorPaths(t *testing.T) {
...
+	// wait until previous snapshotDone events are processed
+	requestsUpdated = func() bool {
+		requests, err := l.PendingSnapshotRequests()
+		require.NoError(t, err)
+		return equal(requests, []uint64{50, 100})
+	}
+	require.Eventually(t, requestsUpdated, time.Minute, time.Second)
+
+	// Test 5: reopen the provider and ledger and verify snapshot requests are correct
+	provider.Close()
```
manish-sethi: A test for recovery code can be added (perhaps, a separate test). For this, you can separately open the bookkeeper and add a request equal to commit height and then re-open the provider.

core/ledger/kvledger/snapshot_mgmt_test.go
wenjianqiao: Because a separate goroutine processes snapshotDone event and deletes the corresponding request, we should use eventually here.

wenjianqiao:
@manish-sethi Thank you. I have addressed your comments and pushed a new commit.

core/ledger/kvledger/snapshot_mgmt.go
```diff
@@ -63,6 +90,154 @@ func (l *kvLedger) DeleteSnapshot(height uint64) error {
...
+			if toStartCommitHeight == committedHeight && requestedHeight == committedHeight {
+				snapshotInProgress = true
+				go func() {
+					if err := l.generateSnapshot(); err != nil {
+						logger.Errorw("Failed to generate snapshot", "height", requestedHeight, "error", err)
+					}
+					events <- &event{snapshotDone, requestedHeight}
+				}()
+				requestResponses <- &requestResponse{}
+				continue
```
manish-sethi, line 196: Not needed

core/ledger/kvledger/snapshot_mgmt.go
```diff
@@ -63,6 +90,154 @@ func (l *kvLedger) DeleteSnapshot(height uint64) error {
...
+			}
+			// there is an in-progress commit or the requested height is higher than block height,
+			// we will evaluate the request at commitDone event
```
manish-sethi, line 199: if you remove the above lines, this comment can go off as well. With the revised if condition, this is clearer now anyways.

manish-sethi (approved):
Thanks @wenjianqiao - a pretty minor comment but I do not want to hole his PR for that.

PR #1653, opened 2020-07-29
Log Error Returned by PKC11 BCCSP

lindluni:
We return a series of errors if we are unable to
find a key in an HSM. We explicitly check if the error
was nil, but never log the error for debugging before
falling back to software.

### DIFF
--- a/bccsp/pkcs11/impl.go
+++ b/bccsp/pkcs11/impl.go
@@ -162,13 +162,14 @@ func (csp *impl) KeyImport(raw interface{}, opts bccsp.KeyImportOpts) (k bccsp.K
 // the Subject Key Identifier ski.
 func (csp *impl) GetKey(ski []byte) (bccsp.Key, error) {
 	pubKey, isPriv, err := csp.getECKey(ski)
-	if err == nil {
-		if isPriv {
-			return &ecdsaPrivateKey{ski, ecdsaPublicKey{ski, pubKey}}, nil
-		}
-		return &ecdsaPublicKey{ski, pubKey}, nil
+	if err != nil {
+		logger.Warnf("Key not found using PKCS11: %v", err)
+		return csp.BCCSP.GetKey(ski)
 	}
-	return csp.BCCSP.GetKey(ski)
+	if isPriv {
+		return &ecdsaPrivateKey{ski, ecdsaPublicKey{ski, pubKey}}, nil
+	}
+	return &ecdsaPublicKey{ski, pubKey}, nil
 }
 
 // Sign signs digest using key k.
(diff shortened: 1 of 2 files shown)


bccsp/pkcs11/impl.go
```diff
@@ -162,13 +162,14 @@ func (csp *impl) KeyImport(raw interface{}, opts bccsp.KeyImportOpts) (k bccsp.K
 // the Subject Key Identifier ski.
 func (csp *impl) GetKey(ski []byte) (bccsp.Key, error) {
 	pubKey, isPriv, err := csp.getECKey(ski)
-	if err == nil {
-		if isPriv {
-			return &ecdsaPrivateKey{ski, ecdsaPublicKey{ski, pubKey}}, nil
-		}
-		return &ecdsaPublicKey{ski, pubKey}, nil
+	if err != nil {
+		logger.Errorf("Key not found using PKCS11: %v", err)
```
ale-linux: It seems to me that this shouldn't be logged as an error but as a WARN, given that we continue with the execution

ale-linux (changes requested):
Thanks a lot for this CR - just a minor nit

bccsp/pkcs11/impl.go
lindluni: Fixed

bccsp/pkcs11/pkcs11.go
```diff
@@ -136,27 +137,27 @@ func (csp *impl) getECKey(ski []byte) (pubKey *ecdsa.PublicKey, isPriv bool, err
...
 	_, err = asn1.Unmarshal(marshaledOid, curveOid)
 	if err != nil {
-		return nil, false, fmt.Errorf("Failed Unmarshaling Curve OID [%s]\n%s", err.Error(), hex.EncodeToString(marshaledOid))
+		return nil, false, fmt.Errorf("failed Unmarshaling Curve OID [%s]\n%s", err.Error(), hex.EncodeToString(marshaledOid))
 	}
 
 	curve := namedCurveFromOID(*curveOid)
 	if curve == nil {
-		return nil, false, fmt.Errorf("Cound not recognize Curve from OID")
+		return nil, false, fmt.Errorf("could not recognize Curve from OID")
```
caod123, line 156: Not part of your change, but I find it odd how changing these error messages didn't require a test change. I'm guessing none of these errors were ever tested.

lindluni: Yea, just checked, there is no coverage on any of these error paths in this function, except for one. And it is only checking that there was an error, not the message returned.

ale-linux: Good point - @btl5037, it would be great to add test coverage for the error conditions - not necessarily in this PR tho

ale-linux (approved):

PR #1656, opened 2020-07-30
fix: modify certificate alternate DNS duplication

zhj0811:
Signed-off-by: zhj0811 <zhaojian@peerfintech.cn>

### DIFF
--- a/internal/cryptogen/ca/ca.go
+++ b/internal/cryptogen/ca/ca.go
@@ -147,7 +147,7 @@ func (ca *CA) SignCertificate(
 		if ip != nil {
 			template.IPAddresses = append(template.IPAddresses, ip)
 		} else {
-			template.DNSNames = append(template.DNSNames, alternateNames...)
+			template.DNSNames = append(template.DNSNames, san)
 		}
 	}
 
--- a/internal/cryptogen/ca/ca_test.go
+++ b/internal/cryptogen/ca/ca_test.go
@@ -228,12 +228,14 @@ func TestGenerateSignCertificate(t *testing.T) {
 	require.Contains(t, cert.Subject.OrganizationalUnit, ous[1])
 
 	// make sure sans are correctly set
-	sans := []string{testName2, testIP}
+	sans := []string{testName2, testName3, testIP}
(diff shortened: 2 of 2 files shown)


ale-linux (commented):
Thanks! Could you please provide some more context here - what is the issue? Is there a JIRA?

sykesm (changes requested):
Thanks for the PR.

The change looks correct to me but there  needs to be a corresponding test update. Since the current code isn't working as intended, there's clearly a test gap that needs to be addressed.

`ca_test.go` contains the tests for `SignCeritificate` and I'd expect new assertions around line 230.

zhaojianpeerfintech:
> Thanks! Could you please provide some more context here - what is the issue? Is there a JIRA?

image

It is a tiny PR. The  X509v3 Subject Alternative Name result contains duplicate alternateNames when alternateNames using multi domain

zhaojianpeerfintech:
> Thanks for the PR.
> 
> The change looks correct to me but there needs to be a corresponding test update. Since the current code isn't working as intended, there's clearly a test gap that needs to be addressed.
> 
> `ca_test.go` contains the tests for `SignCeritificate` and I'd expect new assertions around line 230.

The PR does not affect the ca_test.go.  It is negligible.

sykesm:
> > Thanks for the PR.
> 
> > 
> 
> > The change looks correct to me but there needs to be a corresponding test update. Since the current code isn't working as intended, there's clearly a test gap that needs to be addressed.
> 
> > 
> 
> > `ca_test.go` contains the tests for `SignCeritificate` and I'd expect new assertions around line 230.
> 
> 
> 
> The PR does not affect the ca_test.go.  It is negligible.

I'll try to be more clear. The fix is fine but the pull request *must* include a corresponding update to the unit tests to ensure the issue is resolved and is not regressed.

To quote our contribution guidelines:  
> Include unit and integration tests (or changes to existing tests) with every change.

Without a test update, the PR as written will not be merged.

Thanks.

sykesm (approved):

sykesm:

PR #1666, opened 2020-08-01
fix glossary.rst by reverting commit 21fa6480b161

nhoriguchi:
#### Type of change

- Documentation update

#### Description

Commit 21fa6480b161 converted the sentence "the latter would tell the
former about ``peer0.orgA``".  But this paragraph explains with example
that the info about peer0.orgA is transferred to peer1.orgB (referred to
as "the former") via peer0.orgC (referred to as "the latter").  So the
original text is totally correct. Let's revert the commit.

Signed-off-by: Naoya Horiguchi <naoya.horiguchi@nec.com>


lindluni (approved):

lindluni:

lindluni:

PR #1669, opened 2020-08-03
Reuse PKCS11 objects for keys

mastersingh24:
#### Type of change

- Improvement

#### Description

Prior to this change, each invocation of Sign or
Verify actually made a PKCS11 call to find the
key object.  This is actually unnecessary as the
object handle is valid across multiple sessions.

This change adds a handle field to both the private
and public key objects and uses these handles across
multiple invocations of Sign and Verify for any given
key pair.

#### Additional details

This change should improve overall performance when using network HSMs as each Sign or Verify invocation now just make a single call to the HSM.

### DIFF
--- a/bccsp/pkcs11/ecdsa.go
+++ b/bccsp/pkcs11/ecdsa.go
@@ -14,8 +14,8 @@ import (
 	"github.com/hyperledger/fabric/bccsp/utils"
 )
 
-func (csp *impl) signECDSA(k ecdsaPrivateKey, digest []byte, opts bccsp.SignerOpts) ([]byte, error) {
-	r, s, err := csp.signP11ECDSA(k.ski, digest)
+func (csp *impl) signECDSA(k *ecdsaPrivateKey, digest []byte, opts bccsp.SignerOpts) ([]byte, error) {
+	r, s, err := csp.signP11ECDSA(k.handle, digest)
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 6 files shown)


bccsp/pkcs11/ecdsa.go
```diff
@@ -46,6 +46,7 @@ func (csp *impl) verifyECDSA(k ecdsaPublicKey, signature, digest []byte, opts bc
 	if csp.softVerify {
 		return ecdsa.Verify(k.pub, digest, r, s), nil
 	}
-	return csp.verifyP11ECDSA(k.ski, digest, r, s, k.pub.Curve.Params().BitSize/8)
+	fmt.Printf("handle: %d\n", k.handle)
```
C0rWin: a left over?

mastersingh24: yeah ... force pushed a fix

C0rWin (approved):

lindluni:
Ran it through e2e functional tests, everything looks good.

lindluni (approved):

bccsp/pkcs11/ecdsakey.go
```diff
@@ -21,12 +21,15 @@ import (
 	"errors"
 	"fmt"
 
+	"github.com/miekg/pkcs11"
+
```
lindluni, line 25: nit: Blank line and imports out of order, I'll clean it up in a separate PR so you don't have to waste time repushing

PR #1681, opened 2020-08-05
[FAB-18208] Do not sign gossip message if membership is empty

yacovm:
This change set makes gossip discovery and channel modules only sign (and gossip)
alive messages and state info messages if the membership is non empty.

Change-Id: Id4de507cdccaa5860ba4b67cf059defc98545bd9

### DIFF
--- a/gossip/discovery/discovery_impl.go
+++ b/gossip/discovery/discovery_impl.go
@@ -232,16 +232,7 @@ func (d *gossipDiscoveryImpl) InitiateSync(peerNum int) {
 		return
 	}
 	var peers2SendTo []*NetworkMember
-	m, err := d.createMembershipRequest(true)
-	if err != nil {
-		d.logger.Warningf("Failed creating membership request: %+v", errors.WithStack(err))
-		return
-	}
-	memReq, err := m.NoopSign()
-	if err != nil {
-		d.logger.Warningf("Failed creating SignedGossipMessage: %+v", errors.WithStack(err))
-		return
-	}
+
 	d.lock.RLock()
 
 	n := d.aliveMembership.Size()
@@ -267,6 +258,22 @@ func (d *gossipDiscoveryImpl) InitiateSync(peerNum int) {
 	}
 
 	d.lock.RUnlock()
+
+	if len(peers2SendTo) == 0 {
+		d.logger.Debugf("No peers to send to, aborting membership sync")
+		return
+	}
+
+	m, err := d.createMembershipRequest(true)
+	if err != nil {
+		d.logger.Warningf("Failed creating membership request: %+v", errors.WithStack(err))
+		return
+	}
+	memReq, err := m.NoopSign()
+	if err != nil {
+		d.logger.Warningf("Failed creating SignedGossipMessage: %+v", errors.WithStack(err))
+		return
+	}
 
 	for _, netMember := range peers2SendTo {
 		d.comm.SendToPeer(netMember, memReq)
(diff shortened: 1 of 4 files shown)


mastersingh24 (commented):
So rather than having a global setting to disable gossip entirely (which would be a larger change), this essentially allows for minimal compute cycles required when the peer finds itself with an empty membership list.

yacovm:
> So rather than having a global setting to disable gossip entirely (which would be a larger change)

I am not sure it will be a larger change, but, it will be a more intrusive for sure. 
I don't like the "disable gossip" option because it introduces a config parameter and then everyone that wants to support it, will need to update their deployment scripts and what-not. 

> this essentially allows for minimal compute cycles required when the peer finds itself with an empty membership list.
Right:

- In the membership module we now only sign (and gossip) if the message is expected to be sent to anyone.
 Note, that we [will still sign whenever we want to send a membership request](https://github.com/hyperledger/fabric/blob/master/gossip/discovery/discovery_impl.go#L659), because it has to have a fresh (and monotonous increasing) logical time. 
The latter essentially "initiates" the gossip membership reconciliation and without it we will have a chicken and egg problem because we will wait for the periodical process to sign but our membership will be empty so it won't sign.
- In the channel module, we now always defer all metadata (ledger height, chaincodes installed, left channel) updates to the periodical time where we actually both sign the message and also gossip it, when both are only if the membership that is sampled isn't empty. 
This also has a side effect where when we are given a state info request (to get back a snapshot of our state info messages), we might give back a state info response that is empty (without our message and without anyone else's) in case we just booted. Previously, we would always include our message in the response. 
This is fine, because gossip handles empty membership responses gracefully and once we know at least a single peer that is not us, we will also send back a membership response that includes us, and if we don't know about any peer other than ourselves then we will probably never get a membership request in the first place (unless something is really broken in the membership layer, and then we don't want ourselves to participate in the channel anyway because we are basically a black hole that sucks in all messages and never forwards them).

denyeart:
Perhaps this change should be made in master, but I would hesitate from merging into release-1.4. Since there are many production users on release-1.4, we only make critical backports in release-1.4 in order to avoid regressions.
From my testing, I can suppress the gossip signings with a simple config update - just set peer.gossip.pullInterval and peer.gossip.aliveTimeInterval to a high value. Since there is a way to suppress the signings via a simple peer config update, do we really need a code change in release-1.4?

WDYT @yacovm @mastersingh24 ?

yacovm:
> From my testing, I can suppress the gossip signings with a simple config update - just set peer.gossip.pullInterval and peer.gossip.aliveTimeInterval to a high value. Since there is a way to suppress the signings via a simple peer config update, do we really need a code change in release-1.4?

But whenever we commit a block, we update the ledger height that gossip publishes to peers and it [involves a signature](https://github.com/hyperledger/fabric/blob/master/gossip/gossip/channel/channel.go#L941). 

> Perhaps this change should be made in master, but I would hesitate from merging into release-1.4. Since there are many production users on release-1.4, we only make critical backports in release-1.4 in order to avoid regressions.

I'm fine with withholding this from 1.4 and only targeting it to master unless user requirement to backport it will arise.

denyeart:
@yacovm You're right, this fix would still be needed to suppress the signing of ledger height upon block commit. 
peer.gossip.pullInterval and peer.gossip.aliveTimeInterval settings wouldn't suppress that one.
In this case, I'm open to merging into master and release-1.4.

C0rWin (approved):

PR #1683, opened 2020-08-05
[FAB-18117] Implement Snapshot grpc service

wenjianqiao:

#### Type of change
- New feature

#### Description
Implement Snapshot grpc service that delegates to ledger to submit, cancel, and list snapshot request.

#### Related issues
https://jira.hyperledger.org/browse/FAB-18117


core/ledger/ledgermgmt/snapshot_service.go
```diff
@@ -0,0 +1,94 @@
...
+
+// Snapshot Service implements SnapshotServer grpc interface
+type SnapshotService struct {
+	LedgerGetter LedgerGetter
+}
+
+// LedgerGetter gets the PeerLedger associated with a channel.
+type LedgerGetter interface {
+	GetLedger(cid string) ledger.PeerLedger
+}
```
manish-sethi: This looks odd to me that this service impl is in ledgermgmt package in which LegderMgr keeps track of all the ledgers and a dependency is being introduced for getting access to a ledger. This round-about dependency where the single package is create and opens ledgers and at the same time needs the access to the same object via an external dependency may cause confusions. I would prefer either
- Have this code in a separate package (snapshotgrpc?) and then have it as is
- If in this package, then better to implement these functions directly on LedgerMgr

I tend to inclined towards the former as this function is not directly related to ledgermgmt.

core/ledger/ledgermgmt/snapshot_service.go
```diff
@@ -0,0 +1,94 @@
...
+
+// Generate generates a snapshot request.
+func (s *SnapshotService) Generate(ctx context.Context, signedRequest *pb.SignedSnapshotRequest) (*empty.Empty, error) {
```
yacovm: Pardon my interest, but, you have here 
```
Generate(SignedSnapshotRequest) returns (google.protobuf.Empty) {}
Cancel(SignedSnapshotRequest) returns (google.protobuf.Empty) {}
QueryPendings(SignedSnapshotRequest) returns (QueryPendingSnapshotsResponse) {}
```

Why don't you make some dispatcher to handle all of them? like in the [discovery service](https://github.com/hyperledger/fabric/blob/23466eb45de730aabdd5491d8ba5e170091d4a52/discovery/service.go#L106-L119) ?

This way you can then add authentication checks and sanity checks only in a single place, and not spread them across this file, and after all checks pass, it will jump to a "function pointer" and return either empty or the response

manish-sethi: The handler approach causes unnecessary indirection. Which forces you to use `oneof` in  proto request message. IMO, the handler based approach would have been useful if we had a single handler at the peer level that performs these common tasks and new services require just to add new function. While writing full service, I see little value and instead invoke a common function at the beginning of each function sounds to me a good way to avoid indirection.

internal/peer/node/start.go
```diff
@@ -872,6 +873,10 @@ func serve(args []string) error {
 	// Register the Endorser server
 	pb.RegisterEndorserServer(peerServer.Server(), auth)
 
+	// register the snapshot server
+	snapshotSvc := &snapshotgrpc.SnapshotService{LedgerGetter: peerInstance}
+	pb.RegisterSnapshotServer(peerServer.Server(), snapshotSvc)
```
cendhu, line 878: Instead of `peerInstance`, it should pass the `LedgerMgr`.  The `peerInstance` should not be passed to the lower level components.

core/ledger/snapshotgrpc/snapshot_service.go
```diff
@@ -0,0 +1,94 @@
...
+	}
+	if request.ChannelId == "" {
+		return nil, errors.New("missing channel ID")
+	}
+
+	lgr := s.LedgerGetter.GetLedger(request.ChannelId)
+	if lgr == nil {
+		return nil, errors.Errorf("cannot find ledger for channel %s", request.ChannelId)
+	}
+
```
cendhu: as this part of the code is repeating thrice, we can move it into a separate function.

core/ledger/snapshotgrpc/mock/peer_ledger.go
```diff
@@ -0,0 +1,1716 @@
+// Code generated by counterfeiter. DO NOT EDIT.
+package mock
```
cendhu: 1. We decided to avoid mock within the ledger pkg as much as possible. Are there any difficulties in testing the services using the actual ledger pkg?
2. Even if we need, we need to create a local interface with the exact required methods and then create a mock. For example, why do we need to mock `CommitPvtDataOfOldBlocks()`?

Given that we already have a few mocks to be removed (TODO), I would recommend not to add more mocks.

core/ledger/snapshotgrpc/snapshot_service.go
```diff
@@ -0,0 +1,94 @@
...
+// Snapshot Service implements SnapshotServer grpc interface
+type SnapshotService struct {
+	LedgerGetter LedgerGetter
+}
+
+// LedgerGetter gets the PeerLedger associated with a channel.
+type LedgerGetter interface {
+	GetLedger(cid string) ledger.PeerLedger
+}
```
cendhu, line 27: Let's avoid these interfaces. We can directly use the object, i.e., `LedgerMgr`.

core/ledger/snapshotgrpc/snapshot_service.go
```diff
@@ -0,0 +1,94 @@
+/*
+Copyright IBM Corp. All Rights Reserved.
+
+SPDX-License-Identifier: Apache-2.0
+*/
+
+package snapshotgrpc
```
cendhu, line 7: IMO, having a bigger but smaller number of packages help in navigation, management and code readability over having a smaller but larger number of packages. On one side, we tried to reduce the number of packages and on the other side, we are increasing it.

cendhu: I tried to reduce both. Having a bigger package like `kvledger` helps in memorability and hence the productivity. I can just look at the filenames to understand what it has. There is a simple mental map of what `kvledger` has in my mind and helped many times. Having a package called `snapshotgrpc` and then file called `snapshot_service.go`, I don't know what we gain from this and personally, it annoys me while coding to move between folders, etc...

manish-sethi: > I don't know what we gain from this ...

I will again refer to my first [comment](https://github.com/hyperledger/fabric/pull/1683#discussion_r466001346) where I mentioned what looked odd to me when I looked at this code.

internal/peer/node/start.go
cendhu: Do you mean to say that we can implement these APIs on the `LedgerMgr` or use the ledger object maintained locally in the `ledgermgmt` pkg while implementing these APIs as a method on `SnapshotService` struct?

manish-sethi: The latter would again lead to a very clumsy cycle, as there is only one instance of the `LedgerMgr` and peer package will pass that instance back to the `SnapshotService` from the same package. So, I mentioned only the former among the two choices. Further, for the former, though it may bring simplicity of dependencies, hardly makes sense making `LedgerMgr` as a `SnapshotService`. That's the detail of my rational to tending to incline for a separate package.

core/ledger/ledgermgmt/snapshot_service.go
yacovm: That's not what I am saying. I am not saying to one-of into a single gRPC method, but to just use internally a dispatching pattern, and attach the function pointer in the gRPC function.

core/ledger/snapshotgrpc/snapshot_service.go
wenjianqiao: The code is not exactly repeated because the proto msg is different for QueryBindings (SnapshotQuery vs. SnapshotRequest). After unmarshalling, there are two parts of common code: channelID validation and getting ledger. We can move them into a common func like the following.

func (s *SnapshotService) getLedger(channelID string) (ledger.PeerLedger, error) {
	if channelID == "" {
		return nil, errors.New("missing channel ID")
	}
	lgr := s.LedgerGetter.GetLedger(channelID)
	if lgr == nil {
		return nil, errors.Errorf("cannot find ledger for channel %s", channelID)
	}
	return lgr, nil
}

core/ledger/ledgermgmt/snapshot_service.go
wenjianqiao: @yacovm thank you for your comment. We (ledger team) discussed and decided to use multiple rpc methods to avoid using oneof and make each call more expressive.

core/ledger/snapshotgrpc/snapshot_service.go
cendhu: Yeah, so there is code repetition that can be avoided. That's what I mentioned.

internal/peer/node/start.go
cendhu: As we discussed offline, it requires a larger refactoring to fix current implementation of `start.go` and how all gRPC services are implemented/called/their entry-point. Hence, it is out of scope for this PR. I agree with your assessment and suggestion.

wenjianqiao:
@manish-sethi @cendhu I have addressed the comments per our slack discussion. Please review again.
- moved the grpc server to a separate `snapshotgrpc` directory
- use real ledger for PeerLedger

core/ledger/snapshotgrpc/snapshot_service_test.go
```diff
@@ -0,0 +1,151 @@
...
+type ledgerGetter interface {
+	LedgerGetter
+}
+
+func TestSnapshot(t *testing.T) {
+	testDir, err := ioutil.TempDir("", "snapshotgrpc")
+	if err != nil {
+		t.Fatalf("Failed to create ledger directory: %s", err)
+	}
+
```
manish-sethi, line 34: could use `require.NoError`

core/ledger/snapshotgrpc/snapshot_service_test.go
```diff
@@ -0,0 +1,151 @@
...
+	fakeLedgerGetter.GetLedgerReturns(nil)
+	for _, test := range tests {
+		t.Run(test.name, func(t *testing.T) {
+			signedRequest = test.signedRequest
+			if signedRequest == nil {
+				// create a SignedQuery for all requests because height does not matter in these error tests
+				signedRequest = createSignedQuery(test.channelID)
+			}
```
manish-sethi, line 108: This is an unnecessary distraction. Better, have the requests explicitly part of the testcase.

manish-sethi (approved):
Thanks @wenjianqiao. I have left pretty minor comments in the test file but would not like to hold this from merging.

PR #1664, opened 2020-07-31
[FAB-18098] Update 1.4 docs for devmode

DereckLuo:
Signed-off-by: Chongxin Luo <Chongxin.Luo@ibm.com>

#### Type of change

- Documentation update

#### Description

Documentation updates for `Using dev mode`.

#### Related issues

[FAB-18098](https://jira.hyperledger.org/browse/FAB-18098)

PR #1684, opened 2020-08-05
[FAB-18073] Add integration test for dev mode

stephyee:
#### Type of change

- Test update

#### Description

Tests functionality of dev mode in 1.4 

#### Related issues

[FAB-18073](https://jira.hyperledger.org/browse/FAB-18073)

### DIFF
--- a/integration/devmode/devmode_suite_test.go
+++ b/integration/devmode/devmode_suite_test.go
@@ -1,0 +1,45 @@
+/*
+Copyright IBM Corp All Rights Reserved.
+
+SPDX-License-Identifier: Apache-2.0
+*/
+
+package devmode
+
+import (
+	"encoding/json"
+	"testing"
+
+	"github.com/hyperledger/fabric/integration/nwo"
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+)
+
+func TestDevMode(t *testing.T) {
+	RegisterFailHandler(Fail)
+	RunSpecs(t, "Devmode Suite")
+}
+
+var components *nwo.Components
+
+var _ = SynchronizedBeforeSuite(func() []byte {
+	components = &nwo.Components{}
+	components.Build()
+
+	payload, err := json.Marshal(components)
+	Expect(err).NotTo(HaveOccurred())
+
+	return payload
+}, func(payload []byte) {
+	err := json.Unmarshal(payload, &components)
+	Expect(err).NotTo(HaveOccurred())
+})
+
+var _ = SynchronizedAfterSuite(func() {
+}, func() {
+	components.Cleanup()
+})
+
+func BasePort() int {
+	return 39000 + 1000*GinkgoParallelNode()
+}
--- a/integration/devmode/devmode_test.go
+++ b/integration/devmode/devmode_test.go
@@ -1,0 +1,210 @@
+/*
+Copyright IBM Corp All Rights Reserved.
+
+SPDX-License-Identifier: Apache-2.0
+*/
+
+package devmode
+
+import (
+	"io/ioutil"
+	"os"
+	"os/exec"
+	"strconv"
+	"syscall"
+	"time"
+
+	docker "github.com/fsouza/go-dockerclient"
+	"github.com/hyperledger/fabric/integration/nwo"
+	"github.com/hyperledger/fabric/integration/nwo/commands"
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+	"github.com/onsi/gomega/gbytes"
+	"github.com/onsi/gomega/gexec"
+	"github.com/tedsuo/ifrit"
+	"github.com/tedsuo/ifrit/ginkgomon"
+)
+
+var _ = Describe("Devmode", func() {
+	var (
+		testDir          string
+		client           *docker.Client
+		network          *nwo.Network
+		process          ifrit.Process
+		chaincode        nwo.Chaincode
+		chaincodeRunner  *ginkgomon.Runner
+		chaincodeProcess ifrit.Process
+		org1peer0        *nwo.Peer
+	)
+
+	BeforeEach(func() {
+		var err error
+		testDir, err = ioutil.TempDir("", "devmode")
+		Expect(err).NotTo(HaveOccurred())
+
+		client, err = docker.NewClientFromEnv()
+		Expect(err).NotTo(HaveOccurred())
+
+		chaincode = nwo.Chaincode{
+			Name:    "mycc",
+			Version: "0.0",
+			Path:    "github.com/hyperledger/fabric/integration/chaincode/simple/cmd",
+			Ctor:    `{"Args":["init","a","100","b","200"]}`,
+			Policy:  `AND ('Org1MSP.member')`,
+		}
+
+		network = nwo.New(devModeSolo, testDir, client, BasePort(), components)
+		network.TLSEnabled = false
+		org1peer0 = network.Peer("Org1", "peer0")
+		org1peer0.DevMode = true
+
+		network.GenerateConfigTree()
+		network.Bootstrap()
+
+		networkRunner := network.NetworkGroupRunner()
+		process = ifrit.Invoke(networkRunner)
+		Eventually(process.Ready(), network.EventuallyTimeout).Should(BeClosed())
+	})
+
+	AfterEach(func() {
+		if process != nil {
+			process.Signal(syscall.SIGTERM)
+			Eventually(process.Wait(), network.EventuallyTimeout).Should(Receive())
+		}
+
+		if chaincodeProcess != nil {
+			chaincodeProcess.Signal(syscall.SIGTERM)
+		}
+
+		if network != nil {
+			network.Cleanup()
+		}
+
+		os.RemoveAll(testDir)
+	})
+
+	It("executes chaincode in dev mode", func() {
+		channelName := "testchannel"
+		chaincodeID := chaincode.Name + ":" + chaincode.Version
+		orderer := network.Orderer("orderer")
+
+		By("setting up the channel")
+		network.CreateAndJoinChannel(orderer, channelName)
+
+		By("building the chaincode")
+		chaincodePath, err := gexec.Build(chaincode.Path)
+		Expect(err).ToNot(HaveOccurred())
+
+		By("running the chaincode")
+		peerChaincodeAddress := network.PeerAddress(org1peer0, nwo.ChaincodePort)
+		cmd := exec.Command(chaincodePath, "--peer.address", peerChaincodeAddress)
+		cmd.Env = append(cmd.Env, "CORE_CHAINCODE_ID_NAME="+chaincodeID, "DEVMODE_ENABLED=true")
+		chaincodeRunner = ginkgomon.New(ginkgomon.Config{
+			Name:              "chaincode",
+			Command:           cmd,
+			StartCheck:        `starting up in devmode...`,
+			StartCheckTimeout: 15 * time.Second,
+		})
+		chaincodeProcess = ifrit.Invoke(chaincodeRunner)
+		Eventually(chaincodeProcess.Ready(), network.EventuallyTimeout).Should(BeClosed())
+
+		By("installing the chaincode")
+		nwo.InstallChaincode(network, chaincode, org1peer0)
+
+		By("Instantiating the chaincode")
+		nwo.InstantiateChaincode(network, channelName, orderer, chaincode, org1peer0, org1peer0)
+
+		By("Querying and invoking chaincode built and started manually")
+		RunQueryInvokeQuery(network, orderer, org1peer0, channelName, 100)
+		Eventually(chaincodeRunner).Should(gbytes.Say("invoking in devmode"))
+	})
+})
+
+func RunQueryInvokeQuery(n *nwo.Network, orderer *nwo.Orderer, peer *nwo.Peer, channel string, queryValue int) {
+	By("querying the chaincode")
+	sess, err := n.PeerUserSession(peer, "User1", commands.ChaincodeQuery{
+		ChannelID: channel,
+		Name:      "mycc",
+		Ctor:      `{"Args":["query","a"]}`,
+	})
+	Expect(err).NotTo(HaveOccurred())
+	Eventually(sess, n.EventuallyTimeout).Should(gexec.Exit(0))
+	Expect(sess).To(gbytes.Say(strconv.Itoa(queryValue)))
+
+	By("invoke the chaincode")
+	sess, err = n.PeerUserSession(peer, "User1", commands.ChaincodeInvoke{
+		ChannelID: channel,
+		Orderer:   n.OrdererAddress(orderer, nwo.ListenPort),
+		Name:      "mycc",
+		Ctor:      `{"Args":["invoke","a","b","10"]}`,
+		PeerAddresses: []string{
+			n.PeerAddress(n.Peer("Org1", "peer0"), nwo.ListenPort),
+		},
+		WaitForEvent: true,
+	})
+	Expect(err).NotTo(HaveOccurred())
+	Eventually(sess, n.EventuallyTimeout).Should(gexec.Exit(0))
+	Expect(sess.Err).To(gbytes.Say("Chaincode invoke successful. result: status:200"))
+
+	By("querying the chaincode again")
+	sess, err = n.PeerUserSession(peer, "User1", commands.ChaincodeQuery{
+		ChannelID: channel,
+		Name:      "mycc",
+		Ctor:      `{"Args":["query","a"]}`,
+	})
+	Expect(err).NotTo(HaveOccurred())
+	Eventually(sess, n.EventuallyTimeout).Should(gexec.Exit(0))
+	Expect(sess).To(gbytes.Say(strconv.Itoa(queryValue - 10)))
+}
+
+var devModeSolo = &nwo.Config{
+	Organizations: []*nwo.Organization{{
+		Name:          "OrdererOrg",
+		MSPID:         "OrdererMSP",
+		Domain:        "example.com",
+		EnableNodeOUs: false,
+		Users:         0,
+		CA:            &nwo.CA{Hostname: "ca"},
+	}, {
+		Name:          "Org1",
+		MSPID:         "Org1MSP",
+		Domain:        "org1.example.com",
+		EnableNodeOUs: true,
+		Users:         2,
+		CA:            &nwo.CA{Hostname: "ca"},
+	}},
+	Consortiums: []*nwo.Consortium{{
+		Name: "SampleConsortium",
+		Organizations: []string{
+			"Org1",
+		},
+	}},
+	Consensus: &nwo.Consensus{
+		Type: "solo",
+	},
+	SystemChannel: &nwo.SystemChannel{
+		Name:    "systemchannel",
+		Profile: "OneOrgOrdererGenesis",
+	},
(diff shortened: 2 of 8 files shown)


integration/devmode/devmode_test.go
```diff
@@ -0,0 +1,184 @@
...
+		network.GenerateConfigTree()
+		network.Bootstrap()
+
+		networkRunner := network.NetworkGroupRunner()
+		process = ifrit.Invoke(networkRunner)
+		Eventually(process.Ready(), network.EventuallyTimeout).Should(BeClosed())
+	})
+
+	AfterEach(func() {
+		By("Cleanup")
```
wlahti: nit: remove this?

integration/devmode/devmode_test.go
```diff
@@ -0,0 +1,184 @@
...
+		}
+
+		if network != nil {
+			network.Cleanup()
+		}
+
+		os.RemoveAll(testDir)
+	})
+
+	It("executing chaincode in dev mode", func() {
```
wlahti:
```suggestion
	It("executes chaincode in dev mode", func() {
```
Nit: these should read as a sentence: "It executes chaincode in devmode" (applies here and below)

integration/devmode/devmode_test.go
```diff
@@ -0,0 +1,184 @@
...
+		network.CreateAndJoinChannel(orderer, channelName)
+
+		By("building chaincode")
+		cmd := exec.Command("go", "build", "-o", filepath.Join(testDir, "simple"), chaincode.Path)
+		err := cmd.Run()
+		Expect(err).ToNot(HaveOccurred())
+
+		By("running chaincode")
+		peerChaincodeAddress := network.PeerAddress(org1peer0, nwo.ChaincodePort)
+		setPeerEnv(peerChaincodeAddress, chaincodeID)
```
wlahti: This setting and unsetting at the OS level seems wrong to me. Can't we just set the flag/env variables for the command and not the entire OS?

integration/devmode/devmode_test.go
```diff
@@ -0,0 +1,184 @@
...
+	})
+
+	It("executing chaincode in dev mode", func() {
+		org1peer0 := network.Peer("Org1", "peer0")
+		orderer := network.Orderer("orderer")
+
+		By("setting up the channel")
+		network.CreateAndJoinChannel(orderer, channelName)
+
+		By("building chaincode")
```
wlahti: Not entirely sure but can't we use components.Build() here to build the chaincode?

wlahti (commented):
Yay devmode integration test! Thanks for figuring this out. :) A few questions/suggestions and a few nits.

integration/devmode/devmode_test.go
stephyee: 🧹

integration/devmode/devmode_test.go
stephyee: 🧹

integration/devmode/devmode_test.go
stephyee: We could but I'm not sure we should because we don't want to build this chaincode every time the `Build()` is called which is in multiple suite files.

integration/devmode/devmode_test.go
stephyee: 🧹

integration/devmode/devmode_test.go
wlahti: Ah, my bad. components.Build() in release-1.4 is used differently than it is in master. This is probably fine here but I think we should use it when we port this test to master.

wlahti: Actually though... for this PR, why don't we use gexec.Build() to keep it consistent with how we build the other binaries? :)

stephyee: 🧹

wlahti:
Looks pretty good to me at this point. 

@sykesm Can you take a look when you get a chance? Thanks!

integration/devmode/devmode_test.go
```diff
@@ -0,0 +1,171 @@
...
+		peers := network.PeersWithChannel(channelName)
+		nwo.InstallChaincode(network, chaincode, peers...)
+		By("Instantiating the chaincode")
+		nwo.InstantiateChaincode(network, channelName, orderer, chaincode, org1peer0, peers...)
+		By("Querying and invoking the chaincode")
+		RunQueryInvokeQuery(network, orderer, org1peer0, channelName, 100)
+
+		By("killing the chaincode process")
+		chaincodeProcess.Signal(syscall.SIGKILL)
+		Eventually(chaincodeProcess.Wait(), network.EventuallyTimeout).Should(Receive(MatchError("exit status 137")))
```
sykesm: Why do we care what the error is here? The chaincode was killed and we expect it to terminate and that's the assertion we want.

In other words: we don't need or want the `MatchError`.

integration/devmode/devmode_suite_test.go
```diff
@@ -0,0 +1,45 @@
...
+import (
+	"encoding/json"
+	"testing"
+
+	"github.com/hyperledger/fabric/integration/nwo"
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+)
+
+func TestEndToEnd(t *testing.T) {
```
sykesm: Why is this EndToEnd instead of DevMode?

integration/devmode/devmode_test.go
```diff
@@ -0,0 +1,171 @@
...
+
+		By("building the chaincode")
+		cmd, err := gexec.Build(chaincode.Path)
+		Expect(err).ToNot(HaveOccurred())
+
+		By("running the chaincode")
+		peerChaincodeAddress := network.PeerAddress(org1peer0, nwo.ChaincodePort)
+		chaincodeRunner = network.ChaincodeRunner(cmd, buildPeerEnv(peerChaincodeAddress, chaincodeID)...)
+		chaincodeProcess = ifrit.Invoke(chaincodeRunner)
+		Eventually(chaincodeProcess.Ready(), network.EventuallyTimeout).Should(BeClosed())
```
sykesm: I'm confused. We're running a multi-peer network, starting an instance of the chaincode, and pointing it to a peer. Later, we install and instantiate the chaincode across two peers.

What asserts that our "dev" chaincode is being used?
How is the "dev" chaincode possibly getting used by the second instance of the peer given we've started it with a reference to the first?
How common is it for "dev" mode to require multiple endorsements?

sykesm (changes requested):
Many questions, some comments.

integration/nwo/network.go
```diff
@@ -1023,11 +1026,25 @@ func (n *Network) OrdererGroupRunner() ifrit.Runner {
...
+	cmd.Env = append(cmd.Env, env...)
+	return ginkgomon.New(ginkgomon.Config{
+		AnsiColorCode:     n.nextColor(),
+		Name:              "chaincode",
+		Command:           cmd,
+		StartCheck:        `starting up ...`,
+		StartCheckTimeout: 15 * time.Second,
+	})
+}
+
```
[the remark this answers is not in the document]
DereckLuo: removed the function, done.

integration/nwo/network.go
```diff
@@ -1065,7 +1082,9 @@ func (n *Network) peerCommand(command Command, env ...string) *exec.Cmd {
 	cmd := NewCommand(n.Components.Peer(), command)
 	cmd.Env = append(cmd.Env, env...)
 	if ConnectsToOrderer(command) {
-		cmd.Args = append(cmd.Args, "--tls")
+		if n.TLSEnabled {
+			cmd.Args = append(cmd.Args, "--tls")
+		}
```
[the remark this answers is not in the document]
DereckLuo: done.

integration/nwo/commands/peer.go
```diff
@@ -19,6 +23,7 @@ func (n NodeStart) SessionName() string {
 func (n NodeStart) Args() []string {
 	return []string{
 		"node", "start",
+		fmt.Sprintf("--peer-chaincodedev=%s", strconv.FormatBool(n.DevMode)),
```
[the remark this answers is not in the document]
DereckLuo: thanks, done.

integration/devmode/devmode_suite_test.go
```diff
@@ -0,0 +1,45 @@
...
+	Expect(err).NotTo(HaveOccurred())
+})
+
+var _ = SynchronizedAfterSuite(func() {
+}, func() {
+	components.Cleanup()
+})
+
+func BasePort() int {
+	return 39000 + 1000*GinkgoParallelNode()
```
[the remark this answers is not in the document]
DereckLuo: Integration test in 1.4 does not have ports.go, changed this to keep it consistent with other integration tests

integration/devmode/devmode_test.go
DereckLuo: done.

integration/devmode/devmode_suite_test.go
DereckLuo: fixed.

integration/devmode/devmode_test.go
```diff
@@ -0,0 +1,171 @@
...
+		By("Querying and invoking the chaincode")
+		RunQueryInvokeQuery(network, orderer, org1peer0, channelName, 90)
+	})
+})
+
+func buildPeerEnv(peerChaincodeAddress, chaincodeID string) []string {
+	envs := []string{"CORE_PEER_TLS_ENABLED=false"}
+	envs = append(envs, fmt.Sprintf("CORE_PEER_ADDRESS=%s", peerChaincodeAddress))
+	envs = append(envs, fmt.Sprintf("CORE_CHAINCODE_ID_NAME=%s", chaincodeID))
+	return envs
```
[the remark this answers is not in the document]
DereckLuo: Agree that disable TLS is not necessary here (fixed).

By following the 1.4 [DevMode doc](https://hyperledger-fabric.readthedocs.io/en/release-1.4/peer-chaincode-devmode.html?highlight=dev%20mode), `CORE_PEER_ADDRESS` needs to be set to `chaincode address` in order to launch the chaincode, and also doubled checked by setting it to `peer listen address` will not work.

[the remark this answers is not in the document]
sykesm: > By following the 1.4 [DevMode doc](https://hyperledger-fabric.readthedocs.io/en/release-1.4/peer-chaincode-devmode.html?highlight=dev%20mode), `CORE_PEER_ADDRESS` needs to be set to `chaincode address` in order to launch the chaincode, and also doubled checked by setting it to `peer listen address` will not work.

`CORE_PEER_ADDRESS` is not *needed* for the chaincode. It needs to know the chaincode address. There are two ways to provide that: [viper](https://github.com/hyperledger/fabric/blob/6361bff7dc0d2f2c5ee2c5e66aa6ca7af319bee7/core/chaincode/shim/chaincode.go#L295-L304) or [command line flag](https://github.com/hyperledger/fabric/blob/6361bff7dc0d2f2c5ee2c5e66aa6ca7af319bee7/core/chaincode/shim/chaincode.go#L83). When the peer spawns the chaincode, it [passes the `-peer.address' flag](https://github.com/hyperledger/fabric/blob/6361bff7dc0d2f2c5ee2c5e66aa6ca7af319bee7/core/chaincode/container_runtime.go#L164-L165).

The documentation can be updated to indicate both work and we should lean towards the flag.

integration/devmode/devmode_test.go
```diff
@@ -0,0 +1,185 @@
...
+		By("running the chaincode")
+		peerChaincodeAddress := network.PeerAddress(org1peer0, nwo.ChaincodePort)
+		envs := buildPeerEnv(peerChaincodeAddress, chaincodeID)
+		cmd := exec.Command(chaincodePath)
+		cmd.Env = append(cmd.Env, envs...)
+		chaincodeRunner = ginkgomon.New(ginkgomon.Config{
+			Name:              "chaincode",
+			Command:           cmd,
+			StartCheckTimeout: 15 * time.Second,
+		})
```
sykesm: You seemed to drop the start check so it's ready as soon as it's started. That's going to be flaky.

integration/devmode/devmode_test.go
```diff
@@ -0,0 +1,185 @@
...
+		Eventually(chaincodeProcess.Wait(), network.EventuallyTimeout).Should(Receive())
+
+		By("restarting the chaincode process")
+		cmd = exec.Command(chaincodePath)
+		cmd.Env = append(cmd.Env, envs...)
+		chaincodeRunner = ginkgomon.New(ginkgomon.Config{
+			Name:              "chaincode",
+			Command:           cmd,
+			StartCheckTimeout: 15 * time.Second,
+		})
```
sykesm: You seemed to drop the start check.

integration/devmode/devmode_test.go
```diff
@@ -0,0 +1,185 @@
...
+		By("setting up the channel")
+		network.CreateAndJoinChannel(orderer, channelName)
+
+		By("building the chaincode")
+		chaincodePath, err := gexec.Build(chaincode.Path)
+		Expect(err).ToNot(HaveOccurred())
+
+		By("running the chaincode")
+		peerChaincodeAddress := network.PeerAddress(org1peer0, nwo.ChaincodePort)
+		envs := buildPeerEnv(peerChaincodeAddress, chaincodeID)
```
sykesm: `buildPeerEnv` still seems misnamed. You're building the environment for the chaincode (not the peer) and the only item that is required is the the chaincode ID. Since the function is replacing and not appending the existing runtime environment (I don't know if that's what you wanted or not), you can just set the command environment to contain the chaincode ID.

integration/devmode/devmode_test.go
stephyee: 🧹

integration/devmode/devmode_test.go
stephyee: 🧹

integration/devmode/devmode_test.go
stephyee: 🧹

integration/devmode/devmode_test.go
```diff
@@ -0,0 +1,181 @@
...
+		nwo.InstallChaincode(network, chaincode, peers...)
+		By("Instantiating the chaincode")
+		nwo.InstantiateChaincode(network, channelName, orderer, chaincode, org1peer0, peers...)
+		By("Querying and invoking the chaincode")
+		RunQueryInvokeQuery(network, orderer, org1peer0, channelName, 100)
+
+		By("killing the chaincode process")
+		chaincodeProcess.Signal(syscall.SIGKILL)
+		Eventually(chaincodeProcess.Wait(), network.EventuallyTimeout).Should(Receive())
+
```
wlahti: This seems like a good place to address Matt's comment "What asserts that our "dev" chaincode is being used?" by attempting an invoke of the chaincode and getting a failure since we stopped the chaincode.

integration/devmode/devmode_test.go
```diff
@@ -0,0 +1,181 @@
...
+			StartCheckTimeout: 15 * time.Second,
+		})
+		chaincodeProcess = ifrit.Invoke(chaincodeRunner)
+		Eventually(chaincodeProcess.Ready(), network.EventuallyTimeout).Should(BeClosed())
+
+		By("installing the chaincode")
+		peers := network.PeersWithChannel(channelName)
+		nwo.InstallChaincode(network, chaincode, peers...)
+		By("Instantiating the chaincode")
+		nwo.InstantiateChaincode(network, channelName, orderer, chaincode, org1peer0, peers...)
```
wlahti: I don't know enough about devmode but maybe we could/should also check that docker images haven't been created/started? Just brainstorming here...

integration/devmode/devmode_test.go
```diff
@@ -0,0 +1,181 @@
...
+		By("building the chaincode")
+		chaincodePath, err := gexec.Build(chaincode.Path)
+		Expect(err).ToNot(HaveOccurred())
+
+		By("running the chaincode")
+		peerChaincodeAddress := network.PeerAddress(org1peer0, nwo.ChaincodePort)
+		envs := []string{fmt.Sprintf("CORE_PEER_ADDRESS=%s", peerChaincodeAddress), fmt.Sprintf("CORE_CHAINCODE_ID_NAME=%s", chaincodeID)}
+		cmd := exec.Command(chaincodePath)
+		cmd.Env = append(cmd.Env, envs...)
+		chaincodeRunner = ginkgomon.New(ginkgomon.Config{
```
yacovm, line 102: This might seem like an odd question, but - why are you running the chaincode in a separate process? Can't you run the chaincode directly from within this test? You'll have better control of it, and by taking control from the test you will ensure that it is your chaincode that is connecting to the peer, by (for example) making the chaincode (in the test) do a `PutState` to a key with a random value and then doing a query afterward the block commits that will ensure this worked end to end.

integration/devmode/devmode_test.go
stephyee: 🧹 
- Reduced the scope to a single peer and modified the endorsement policy for the chaincode
- Asserting the "dev" chaincode is being used by confirming no docker containers exist during the first query

integration/devmode/devmode_test.go
stephyee: 🧹 See updates

integration/devmode/devmode_test.go
stephyee: 🧹 See updates

integration/devmode/devmode_test.go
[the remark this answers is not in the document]
stephyee: 🧹 https://github.com/hyperledger/fabric/pull/1664

wlahti:
Looks good to me at this point!

@sykesm Can you review this PR again when you get a chance? Thanks!

integration/devmode/devmode_test.go
```diff
@@ -0,0 +1,159 @@
...
+		Expect(err).ToNot(HaveOccurred())
+
+		By("running the chaincode")
+		peerChaincodeAddress := network.PeerAddress(org1peer0, nwo.ChaincodePort)
+		cmd := exec.Command(chaincodePath, "--peer.address", peerChaincodeAddress)
+		cmd.Env = append(cmd.Env, "CORE_CHAINCODE_ID_NAME="+chaincodeID, "devmode_enabled=true")
+		chaincodeRunner = ginkgomon.New(ginkgomon.Config{
+			Name:              "chaincode",
+			Command:           cmd,
+			StartCheck:        `starting up...`,
```
wlahti: Can we update the chaincode to issue a startup message ("chaincode starting in devmode" or something) when the devmode_enabled environment variable is set and check that here? With that in place, we'll have a consistent start check for release-1.4 and master.

integration/devmode/devmode_test.go
```diff
@@ -0,0 +1,211 @@
...
+
+		chaincode = nwo.Chaincode{
+			Name:    "mycc",
+			Version: "0.0",
+			Path:    "github.com/hyperledger/fabric/integration/chaincode/simple/cmd",
+			Ctor:    `{"Args":["init","a","100","b","200"]}`,
+			Policy:  `AND ('Org1MSP.member')`,
+		}
+
+		chaincodeID = chaincode.Name + ":" + chaincode.Version
```
wlahti: Nit: similar to the PR on master, can we define this down where we use it instead of up here?

wlahti (commented):
Other than the one small nit, I think we've addressed all of the comments from @sykesm on the master branch PR. There are some minor differences in the environment variables for the command but any change would just be stylistic instead of functional.

sykesm (approved):

PR #1686, opened 2020-08-06
Init dbname in couchdb redolog

GrapeBaBa:
Signed-off-by: grapebaba <281165273@qq.com>

#### Type of change

- Bug fix

#### Description

The redolog in the couchdb missing dbname assignment when initialization.

### DIFF
--- a/core/ledger/kvledger/txmgmt/statedb/statecouchdb/redolog.go
+++ b/core/ledger/kvledger/txmgmt/statedb/statecouchdb/redolog.go
@@ -15,13 +15,12 @@ import (
 	"github.com/hyperledger/fabric/core/ledger/kvledger/txmgmt/statedb"
 )
 
-var redologKeyPrefix = []byte{byte(0)}
+var redoLogKey = []byte{byte(0)}
 
 type redoLoggerProvider struct {
 	leveldbProvider *leveldbhelper.Provider
 }
 type redoLogger struct {
-	dbName   string
 	dbHandle *leveldbhelper.DBHandle
 }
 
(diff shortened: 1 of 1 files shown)


core/ledger/kvledger/txmgmt/statedb/statecouchdb/redolog.go
```diff
@@ -40,6 +40,7 @@ func newRedoLoggerProvider(dirPath string) (*redoLoggerProvider, error) {
 
 func (p *redoLoggerProvider) newRedoLogger(dbName string) *redoLogger {
 	return &redoLogger{
+		dbName:   dbName,
 		dbHandle: p.leveldbProvider.GetDBHandle(dbName),
```
manish-sethi: Thanks @GrapeBaBa.

I'll have to recall but it appears that I had a field dbName but then later decided directly to maintain the db specific dbHandle and forgot to remove the field `dbName`.

On the usage, I noticed that it is being referred to [one place](https://github.com/hyperledger/fabric/blob/c6e71a7baa6c0a6d2b1482f2d1419b677296ca8f/core/ledger/kvledger/txmgmt/statedb/statecouchdb/redolog.go#L69) - which means that the keys won't have this value appended - which anyways is not needed because we are using db specific handle. Though, no one should have been using this (as this is applicable only if there would be some post-order transaction), still, since this refers to the persisted data, we should not change the encoded keys. So, ideally, we should rather remove this field. But I would like to take some time to go through the history and make sure of this...

cendhu: Yes, Manish. I had the same thought. We need to remove the `dbName`. Btw, the config tx is a post order tx and get stored in the `redoLogger`, right?

GrapeBaBa: @manish-sethi @cendhu Cool, what is your suggested change, maybe I can help to implement in this PR.

manish-sethi: @cendhu - yes, this gets used in the config tx path but since config tx is in a separate block and is a single key update transaction, it cannot cause inconsistency. This support was mainly added when the work on tokens was using the approach of post order execution and there was a possibility of committing a partial updates of a block before a crash happens. In fact, this can be removed now and has no value as such as the above condition does not exist anymore.

@GrapeBaBa sure, you can continue on this. The fix should be to remove the field `dbName` altogether from the struct `redoLogger`.

manish-sethi (changes requested):
Holding this in the current form

core/ledger/kvledger/txmgmt/statedb/statecouchdb/redolog.go
```diff
@@ -50,27 +48,21 @@ func (p *redoLoggerProvider) close() {
 }
 
 func (l *redoLogger) persist(r *redoRecord) error {
-	k := encodeRedologKey(l.dbName)
 	v, err := encodeRedologVal(r)
 	if err != nil {
 		return err
 	}
-	return l.dbHandle.Put(k, v, true)
+	return l.dbHandle.Put(redologKeyPrefix, v, true)
```
manish-sethi: Please rename `redologKeyPrefix` to `redologKey`, as this is the key now.

manish-sethi (commented):
One minor comment. Also, please update the title of the PR

manish-sethi:
Also, with you next push, squash all the commits into one.

manish-sethi (approved):
Thanks @GrapeBaBa for your contribution

PR #1693, opened 2020-08-09
FAB-18086 Ch.Part.API: Improve wiring between Registrar and follower

tock-ibm:
Currently several opaque function pointers are used to decouple the
multichannel.Registrar and the follower.Chain. In this commit:

Convert opaque function pointers to well defined interfaces that are
easy to follow and mock for tests.

The BlockPuller creation mechanisms is converted to a factory struct
that receives all the prerequisites except the config-block, and exposes
a method that builds a new BlockPuller given a config block. This allows
us to use the same mechanism for on-boarding, until the join block, and
for following past the join-block. The factory also maintains a link to
a block signature verifier that is built from the latest config, and
exposes a method that allows it to update the block signature verifier
on every incoming config.

The block-puller was augmented with a method to update the endpoints.
The endpoints are updated on every incoming config past the join-block.

The follower's function pointer that was used to create a consensus.Chain
is replaced by an interface that defines this method.
This interface is implemented by the registrar and is set into the new
follower instance at construction time.

Change-Id: I3fa6e662c43ca5f566ae1644b8e61695da0ba2d4

#### Type of change

- New feature
#### Related issues

Task: FAB-18086
Epic: FAB-17712


orderer/common/follower/follower_chain.go
```diff
@@ -313,7 +329,7 @@ func (c *Chain) pull() error {
 			// Trigger creation of a new consensus.Chain.
 			c.logger.Info("On-boarding finished successfully, invoking chain creation callback")
```
wlahti: Nit: shouldn't we update this log message to not reference the chain creation callback?

orderer/common/follower/follower_chain.go
```diff
@@ -335,7 +351,7 @@ func (c *Chain) pull() error {
 	// Trigger creation of a new consensus.Chain.
 	c.logger.Info("Block pulling finished successfully, invoking chain creation callback")
```
wlahti: Nit: also here, shouldn't we update this log message to not reference the chain creation callback?

orderer/common/follower/block_puller.go
```diff
@@ -24,73 +24,152 @@ import (
...
+		Dialer:              factory.stdDialer,
 	}
 
 	return bp, nil
 }
+
+// UpdateVerifierFromConfigBlock creates a new block signature verifier from the config block and updates the internal
+// link to said verifier.
+func (factory *BlockPullerCreator) UpdateVerifierFromConfigBlock(configBlock *common.Block) error {
+	configEnv, errC := cluster.ConfigFromBlock(configBlock)
```
wlahti: Style nit: why have errC and errV here instead of just err? Doesn't seem to serve a purpose.

orderer/common/follower/follower_chain.go
```diff
@@ -154,6 +166,14 @@ func NewChain(
 			return nil, errors.New("block data is nil")
 		}
 
+		// Check the block puller creation function once before we start the follower. This ensures we can extract
+		// the endpoints from the join-block.
+		puller, errBP := blockPullerFactory.BlockPuller(joinBlock)
```
wlahti: Style nit here as well: why errBP?

orderer/common/follower/follower_chain.go
```diff
@@ -313,7 +329,7 @@ func (c *Chain) pull() error {
 			// Trigger creation of a new consensus.Chain.
 			c.logger.Info("On-boarding finished successfully, invoking chain creation callback")
 			c.halt()
-			c.chainCreationCallback()
+			c.chainCreator.SwitchFollowerToChain(c.ledgerResources.ChannelID())
```
wlahti, line 331: +1 I like how it's now clear what's happening instead of the generic "chainCreationCallback". :)

orderer/common/follower/follower_chain.go
```diff
@@ -500,14 +521,18 @@ func (c *Chain) pullUntilTarget(targetHeight uint64, inspectConfig bool) (uint64
...
-				// Re-create the block puller to apply the new config, which may update the endpoints.
-				c.blockPuller.Close()
-				var err error
-				if c.blockPuller, err = c.createBlockPullerFunc(); err != nil { // TODO replace with a factory that has a config block as input
-					return n, errors.Wrapf(err, "failed to re-create block puller from last config,  block number: %d", nextBlock.Header.Number)
+				if errVer := c.blockPullerFactory.UpdateVerifierFromConfigBlock(nextBlock); errVer != nil {
+					return n, errors.Wrapf(errVer, "failed to update verifier from last config,  block number: %d", nextBlock.Header.Number)
+				}
+				if updateEndpoints {
+					endpoints, errEP := cluster.EndpointconfigFromConfigBlock(nextBlock, c.cryptoProvider)
```
wlahti: Style nits here also: why errVer and errEP?

orderer/common/multichannel/registrar_test.go
```diff
@@ -781,6 +786,49 @@ func TestRegistrar_JoinChannel(t *testing.T) {
...
+		fChain := registrar.GetFollower("my-raft-channel")
+		require.NotNil(t, fChain)
+		fChain.Halt()
+
+		// Let's assume the follower appended a block
+		genesisBlockAppRaft.Header.Number = 0
+		_ = newLedger(ledgerFactory, "my-raft-channel", genesisBlockAppRaft)
+
+		// Now Switch => a chain is created and the follower removed
+		assert.NotPanics(t, func() { registrar.SwitchFollowerToChain("my-raft-channel") })
```
wlahti: Why assert and not require here?

orderer/common/multichannel/registrar_test.go
```diff
@@ -634,8 +640,7 @@ func TestRegistrar_JoinChannel(t *testing.T) {
 		setup(t)
 		defer cleanup()
 
-		ledgerFactory.Close()
-		ledgerFactory, _ = newLedgerAndFactory(tmpdir, "sys-raft-channel", genesisBlockSysRaft)
+		_ = newLedger(ledgerFactory, "sys-raft-channel", genesisBlockSysRaft)
```
wlahti:
```suggestion
		newLedger(ledgerFactory, "sys-raft-channel", genesisBlockSysRaft)
```
If we just want to create the ledger and append the block, why the "_ ="?

orderer/common/multichannel/registrar_test.go
```diff
@@ -781,6 +786,49 @@ func TestRegistrar_JoinChannel(t *testing.T) {
...
+
+		// After creating the follower.Chain, it not in the chains map, it is in the followers map.
+		require.Nil(t, registrar.GetChain("my-raft-channel"))
+		fChain := registrar.GetFollower("my-raft-channel")
+		require.NotNil(t, fChain)
+		fChain.Halt()
+
+		// Let's assume the follower appended a block
+		genesisBlockAppRaft.Header.Number = 0
+		_ = newLedger(ledgerFactory, "my-raft-channel", genesisBlockAppRaft)
```
wlahti:
```suggestion
		newLedger(ledgerFactory, "my-raft-channel", genesisBlockAppRaft)
```
Same as above.

orderer/common/follower/block_puller.go
```diff
@@ -24,73 +24,152 @@ import (
...
+func NewBlockPullerFactory(
 	channelID string,
+	logger *flogging.FabricLogger,
 	signer identity.SignerSerializer,
-	blockVerifier cluster.BlockVerifier,
 	baseDialer *cluster.PredicateDialer,
 	clusterConfig localconfig.Cluster,
 	bccsp bccsp.BCCSP,
-) (*cluster.BlockPuller, error) {
+	alternativeVerifyBlocksFunc ...ClusterVerifyBlocksFunc,
```
jyellick: This strikes me as really odd.

I understand that by making this a varargs, you can omit it and things work 'with defaults', but it's a hack, and misleading, as you can't meaningfully specify more than one.  I've seen this done as a hack in places where we really don't want to break the API because it would require a cascade of changes or it's a public API, but in this case, we're already breaking it by adding the logger, so I don't understand the value.

It would actually seem much more natural to me to do away with the constructor, and instead have a struct with public members that are set, and give you the 'expected' defaults when they are omitted.  This has the bonus of making the construction easier to read as it becomes clear what each argument is meant to do.  Barring a more radical change like the above, I would certainly get rid of the varargs.

orderer/common/follower/block_puller.go
```diff
@@ -24,73 +24,152 @@ import (
...
+) (BlockPullerFactory, error) {
 	if baseDialer == nil {
 		return nil, errors.New("base dialer is nil")
```
jyellick: I realize this is not changed here, but checking arguments for nil is also a bit odd.  Generally speaking nil is not a valid parameter unless explicitly documented to be so.  Looking at the stdlib you can find lots of places that take pointers and simply crash if they are invalid.

orderer/common/follower/follower_chain.go
```diff
@@ -107,44 +111,52 @@ func NewChain(
...
+		logger:             options.Logger.With("channel", ledgerResources.ChannelID()),
+		timeAfter:          options.TimeAfter,
+		blockPullerFactory: blockPullerFactory,
+		chainCreator:       chainCreator,
+		cryptoProvider:     cryptoProvider,
 	}
 
-	if joinBlock == nil {
-		chain.status = types.StatusActive
+	if clusterConsenter == nil {
```
jyellick: These checks are also redundant to me.  We are the only ones calling this, is there some way we accidentally inject a nil? Wouldn't this be caught in the most basic of integration tests if any of these could be nil?

IMO the only time we should be testing for nil is when we either:
1. Expect the caller to pass in nil as a valid parameter value.
2. Are sanitizing inputs (e.g. reading a protobuf message, user input, etc.)

jyellick (commented):
Overall, like the direction here.  Replacing the opaque pointers with interfaces, and making it more clear when different components are updated is all good to me.

My feeling is that we could still unwind some of the abstractions and make the wiring a little more direct (and in fact, probably get rid of some redundant code).

In particular, we already have a notion of a 'verifier' which is kept up to date based on the state of the ledger.  One that is automatically updated whenever a config block is committed to the blockchain, and one that is leveraged for other policy work like access control in the Broadcast and Deliver services -- the chainSupport.  Maybe this is what's somehow happening under the covers, but it seems like it should be as simple as fetching the latest bundle from the chain support and utilizing this for the block verification.  As I read this it seems like we are instead parsing the channel config and building a new one.  I'm not that concerned from a performance perspective, but seems natural to me to leverage a "the committed state of the blockchain" verifier if one already exists.

As I was reviewing this, a question came to mind.  Perhaps it is a better question for @yacovm but when, in the system channel model, we are polling for the latest channel config for a channel, how do we verify that it's correct? Do we attempt to walk the config blocks back to our current one? Do we simply assume that since the OSN is in the system channel they will be honest about any channel we inquire about?  The thing that sparked this question in my mind was "When do we ever need a block verifier which does not correspond to exactly what's committed on chain?"

orderer/common/follower/follower_chain.go
tock-ibm: done

orderer/common/follower/follower_chain.go
tock-ibm: done

orderer/common/follower/block_puller.go
tock-ibm: done

orderer/common/follower/follower_chain.go
tock-ibm: done

orderer/common/follower/follower_chain.go
tock-ibm: done

orderer/common/multichannel/registrar_test.go
tock-ibm: done

orderer/common/multichannel/registrar_test.go
tock-ibm: done

orderer/common/follower/block_puller.go
tock-ibm: I got rid of the varargs.

I would like to keep the constructor because there is a bit of processing going on in it when building a `BlockPullerCreator` that I would like to hide from the follower creation method in the multichannel package.

yacovm:
> As I was reviewing this, a question came to mind. Perhaps it is a better question for @yacovm but when, in the system channel model, we are polling for the latest channel config for a channel, how do we verify that it's correct? Do we attempt to walk the config blocks back to our current one? Do we simply assume that since the OSN is in the system channel they will be honest about any channel we inquire about? The thing that sparked this question in my mind was "When do we ever need a block verifier which does not correspond to exactly what's committed on chain?"

You never verify the latest block is correct because you never commit it right away, you only poll to see the height but you always commit in-order.

orderer/common/follower/block_puller.go
tock-ibm: done

orderer/common/follower/follower_chain.go
tock-ibm: done

orderer/common/multichannel/registrar_test.go
tock-ibm: done

tock-ibm:
> In particular, we already have a notion of a 'verifier' which is kept up to date based on the state of the ledger. One that is automatically updated whenever a config block is committed to the blockchain, and one that is leveraged for other policy work like access control in the Broadcast and Deliver services -- the **chainSupport**. Maybe this is what's somehow happening under the covers, but it seems like it should be as simple as fetching the latest bundle from the chain support and utilizing this for the block verification.

In the beginning the `follower.Chain` was under the `consensus` package and had (or was inside) a `ChainSupport`. (... and the follower was a formless void and darkness covered the face of the maintainers...). We then decided to move it away from there and dismantle the "support" pattern, which I am glad we did. The `ChainSupport` has so much stuff we don't need here:
```go
type ChainSupport struct {
	*ledgerResources
	msgprocessor.Processor
	*BlockWriter
	consensus.Chain
	cutter blockcutter.Receiver
	identity.SignerSerializer
	BCCSP bccsp.BCCSP
	consensus.MetadataValidator
	consensus.StatusReporter
}
```
The follower does not have that object. The follower does not need to verify TXs, client signatures, filters, block-writer, etc, since it does not process TXs... All we need is a fraction of the capabilities hidden within `multichannel.ledgerResources`, that is, ledger access methods, and a way to validate blocks. The `identity.SignerSerializer` & `bccsp.BCCSP` are passed directly to the follower. I thought I would use the `multichannel.ledgerResources` for block validation past the join block. However, since we moved to a follower that flows through pull-before-join to pull-after-join I can't do that, as we don't have the genesis block when we start the follower, just the join-block. Therefore, I took the pattern from the `cluster` & `onboarding` packages and used that. In there, a verifier is maintained and evolves through the block-pulling, updated when config blocks are committed. This verifier retains a fraction of the channel-config that can be extracted from a config block. It keeps only what it needs.

Nevertheless, your comment resonates with something we discussed in an earlier review, that the `multichannel.ledgerResources` probably need some refactoring. What I think could be done is to split the current `multichannel.ledgerResources` in two:
- "True" ledger resources, i.e. some sugar around `blockledger.ReadWriter`, and 
- a public version of  `multichannel.configResources`, that can be constructed from within the `follower` package. (maybe relocated to a leaf package?)

The current `multichannel.ledgerResources` could then be a composition rather than an embedding of these two (or can stay the same). 

This will allow the follower to construct a (new) `ConfigResources` from either the genesis block as it arrives (late, after follower construction), or the tip of the ledger (if we restart the follower), and `Update()` it from there on when a config block is committed. The  (new) `ConfigResources` could implement 
> `func (lr *ledgerResources) VerifyBlockSignature(sd []*protoutil.SignedData, envelope *common.ConfigEnvelope) error`

I think that if we do this correctly, the cluster package would also be able to reuse that, instead of re-implementing it here:
> `func (bv *BlockValidationPolicyVerifier) VerifyBlockSignature(sd []*protoutil.SignedData, envelope *common.ConfigEnvelope) error`

Naturally, I think it is better to do this in a separate task...

Thoughts?

jyellick:
> We then decided to move it away from there and dismantle the "support" pattern, which I am glad we did. The ChainSupport has so much stuff we don't need here:

100% agree that we want to dismantle the support pattern.

> The follower does not have that object. The follower does not need to verify TXs, client signatures, filters, block-writer, etc, since it does not process TXs... All we need is a fraction of the capabilities hidden within multichannel.ledgerResources, that is, ledger access methods, and a way to validate blocks.

I guess what I suggesting here is that if we refactor the chainsupport, such that we can expose:

1. A way to access the current channel configuration.
2. A way to append blocks to the ledger
3. A way to peak at the head of what's committed to the ledger.

Then, we can keep an up to date copy of the channel configuration, based on what's committed to the ledger, leverage this in the block verifier, and, share a commit path with the normal blockwriter path.  When it's a consensus plugin that writes a block, we attach the normal metadata, sign, and then update the assorted resources like channel config during commit.  When it's the follower code, we take the fully formed block, don't muck with the metadata, but trigger that same channel config update.

> a public version of multichannel.configResources, that can be constructed from within the follower package. (maybe relocated to a leaf package?)

I don't quite follow this.  In order to validate a block, you need from the channel configuration:

1. The MSP Manager (as this is a requirement to evaluate policies).
2. The parsed Policy tree (as this is a requirement to be able to validate the block validation policy)

So, you need all of the MSP definitions, and all of the policy definitions, which, is more or less 90+% of the channel configuration.  We're missing a few things like the consenter metadata, anchor peers, endpoints, etc.  But these are, by weight, relatively insignificant relative to the MSP and policy definitions.

The way we generate the channel config also could really use some restructuring (and vast simplifications of its own), but the reality is that short of re-implementing the channel config parsing, you'll end up generating a full channel config any time you need to check a block validation policy.

> ```func (lr *ledgerResources) VerifyBlockSignature(sd []*protoutil.SignedData, envelope *common.ConfigEnvelope) error```

When would we ever need to verify block signatures for the block that is not under the currently committed channel configuration?

Basically, I think it's neat how the block puller optimistically pulls blocks, and only checks one signature and then gets assurances that things are okay based on hash chaining etc.  But... you can still do this using a single verifier that stays in sync with your committed channel configuration so long as your block range does not include a config block (which seems like a fine thing not to optimize for).

> Naturally, I think it is better to do this in a separate task...

Certainly agreed.

tock-ibm:
>The way we generate the channel config also could really use some restructuring (and vast simplifications of its own), but the reality is that short of re-implementing the channel config parsing, you'll end up generating a full channel config any time you need to check a block validation policy.

I agree. I am not suggesting to re-implement that. I don't think saving space or performance matters here, what I would like to achieve is a unified, simple way, for both the follower, raft chain, and possibly cluster to be able to use an object that encapsulates the channel config, and has the union of all the methods they require. The problem now is really of initialization and visibility. I think currently, the functionality we need is here in the `multichannel`:
```go
type mutableResources interface {
	channelconfig.Resources
	Update(*channelconfig.Bundle)
}

type configResources struct {
	mutableResources
	bccsp bccsp.BCCSP
}
```
([see here](https://github.com/hyperledger/fabric/blob/4c2ad49cb3d02655059f25e1d0bfa3e1901269fe/orderer/common/multichannel/ledger_resources.go#L42))

However, one problem is that this is embedded in the `ledgerResources`, which in turn is embedded in `ChainSupport`. The second problem is that this can be initialized only by `multichannel`, at the time of follower or chain creation. With a chain there is no problem because it always starts with a non-empty ledger. With the follower it's different because the ledger is empty. What I am suggesting is to make it possible to perform late initialization, so that a follower can instantiate an object that maintains the channel-config after it pulls the genesis block.

With respect to these points, 

>1. A way to access the current channel configuration.
>2. A way to append blocks to the ledger
>3. A way to peak at the head of what's committed to the ledger.

The above covers (1) - lets call it "config".

The functionality of (2+3) - lets call it "ledger" - is currently added on top of `ledgerResources`, which embeds the `blockledger.ReadWriter` as well. So the name `ledgerResources` is really misleading - it is "ledger-resources-on-top-of-channel-config-resources". What I am suggesting is to isolate and expose (1) and (2+3) independent from each other. 

We'll then have the flexibility to use either "ledger-resources" or "config-resources", and, if we choose to do so, build an object "ledger-resources-AND-config-resources".

tock-ibm:
>When would we ever need to verify block signatures for the block that is not under the currently committed channel configuration?

The block-puller pulls a range of blocks into memory, a range that may include config blocks. It then verifies the range, starting from the channel config that represents the tip of the ledger. Every time it encounters a config block it generates a new (private) verifier that allows it to continue. (More precisely, it uses 

>func (lr *ledgerResources) VerifyBlockSignature(sd []*protoutil.SignedData, envelope *common.ConfigEnvelope) error

with the second parameter to the previous config from the batch, but that is equivalent.) 

> Basically, I think it's neat how the block puller optimistically pulls blocks, and **only checks one signature** and then gets assurances that things are okay based on hash chaining 

I think that is a premature optimization. You could miss corrupt signatures and commit the block. The hash chain does not protect against that. I'd rather see every block signature verified...  Moreover, we have the issue of [hash chain continuity across batches](https://jira.hyperledger.org/browse/FAB-18104).

> etc. But... you can still do this using a single verifier that stays in sync with your committed channel configuration so long as your block range does not include a config block (which seems like a fine thing not to optimize for).

The block puller verifies an in memory batch that may include configs. The verification happens within the block-puller. However, commit to the ledger happens at a higher level, every time  it invokes `PullBlock()` and receives a config-block.  `PullBlock()` pops blocks from said batch.

I do agree that an alternative design of just stopping the batch validation at config blocks would have possible as well... one could mark in the batch the frontier of validation, deliver the config to the upper layer, the upper layer commits, updates channel config, and so on. I assume the performance implications would have been negligible.

tock-ibm:
>>Naturally, I think it is better to do this in a separate task...

>Certainly agreed.

I opened https://jira.hyperledger.org/browse/FAB-18147 for this task.

jyellick (approved):

PR #1694, opened 2020-08-09
FAB-18122 BlockPuller goes into an endless loop on bad endpoint

tock-ibm:
If the `cluster.BlockPuller` receives a set of endpoint to which it
cannot connect (ever), and then `PullBlock()` is called, the calling
thread will go into an endless loop. The `tryFetchBlock` will try to
connect to some endpoint and will fails every time, sleep, and try
again without limit.

This prevents the follower.Chain (and possibly the etcdraft.Chain)
from implementing an exponential backoff and listening to the stop
signal. This means that this chain cannot be `Halt()` in order to
remove the follower.Chain when necessary, or stop it in tests.

It is better to return nil after a present number of retries rather
than get stuck in that loop for ever.

This commit includes a fix and a unit test that replicates the
problem and verifies it was fixed.

Change-Id: Ic012bb02af75b2296c8b02a39179e9faf8ec7997

#### Type of change

- Bug fix

#### Related issues

Bug: FAB-18122
Epic: FAB-17712

### DIFF
--- a/orderer/common/cluster/deliver.go
+++ b/orderer/common/cluster/deliver.go
@@ -114,11 +114,18 @@ func (p *BlockPuller) HeightsByEndpoints() (map[string]uint64, error) {
 
 func (p *BlockPuller) tryFetchBlock(seq uint64) *common.Block {
 	var reConnected bool
-	for p.isDisconnected() {
+
+	for retriesLeft := p.MaxPullBlockRetries; p.isDisconnected(); retriesLeft-- {
 		reConnected = true
 		p.connectToSomeEndpoint(seq)
 		if p.isDisconnected() {
+			p.Logger.Debugf("Failed to connect to some endpoint, going to try again in %v", p.RetryTimeout)
 			time.Sleep(p.RetryTimeout)
+		}
+		if retriesLeft == 0 && p.MaxPullBlockRetries > 0 {
+			p.Logger.Errorf("Failed to connect to some endpoint, attempts exhausted(%d), seq: %d, endpoints: %v",
+				p.MaxPullBlockRetries, seq, p.Endpoints)
+			return nil
 		}
 	}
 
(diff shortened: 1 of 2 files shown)


orderer/common/cluster/deliver.go
```diff
@@ -114,11 +114,18 @@ func (p *BlockPuller) HeightsByEndpoints() (map[string]uint64, error) {
 
 func (p *BlockPuller) tryFetchBlock(seq uint64) *common.Block {
 	var reConnected bool
-	for p.isDisconnected() {
+
+	for retriesLeft := p.MaxPullBlockRetries; p.isDisconnected(); {
```
yacovm: I don't think we need to attempt to pop if we exhausted the attempt count, because when we try to pull any block and fail doing so, we will anyway drop the entire buffer (when we return from **pullBlocks** with an error we close the puller and it resets the buffer), and whenever we close the puller we anyway reset the buffer. 

So, why not just return nil? I also argue that whenever we pull a sequence of blocks and keep them in the buffer, subsequent invocations of **tryFetchBlock** won't even try to connect anyway. It is only until the buffer is empty, when we would detect we are disconnected.

Also let's move the retriesLeft-- to the for loop and do something like:

```

	for retriesLeft := p.MaxPullBlockRetries; p.isDisconnected(); retriesLeft-- {
		reConnected = true
		p.connectToSomeEndpoint(seq)
		if p.isDisconnected() {
			time.Sleep(p.RetryTimeout)
		}

		if p.MaxPullBlockRetries > 0 && retriesLeft <= 0 {
			p.Logger.Errorf("Failed connecting to some endpoint, attempts exhausted(%d)", p.MaxPullBlockRetries)
			return nil
		}
	}
```

tock-ibm: done

yacovm (approved):

PR #1695, opened 2020-08-09
FAB-18123 Ch.Part.API: Update BlockPuller endpoints

tock-ibm:
FAB-18123 Ch.Part.API: Update BlockPuller endpoints
    
As the follower pulls blocks beyond the join-block, it needs to continuously apply the channel config from each config block to the BlockPuller. One aspect of the channel config are the endpoints.
    
It is possible to simply recreate the block puller on every config block. However this is inefficient because it drops the internal     buffer that is on average half-full of fetched and validated blocks. In order to avoid this we add a method that updates the endpoints, disconnects from the current connection (for safety), but does not drop the internal buffer.
    
The next invocations of PullBlock will first re-connect, and then pop blocks from the buffer until it is empty.

#### Type of change

- New feature

#### Related issues

Task: FAB-18123
Epic: FAB-17712

### DIFF
--- a/orderer/common/cluster/deliver.go
+++ b/orderer/common/cluster/deliver.go
@@ -67,6 +67,11 @@ func (p *BlockPuller) Clone() *BlockPuller {
 // Close makes the BlockPuller close the connection and stream
 // with the remote endpoint, and wipe the internal block buffer.
 func (p *BlockPuller) Close() {
+	p.disconnect()
+	p.blockBuff = nil
+}
+
+func (p *BlockPuller) disconnect() {
 	if p.cancelStream != nil {
 		p.cancelStream()
 	}
@@ -78,7 +83,6 @@ func (p *BlockPuller) Close() {
 	p.conn = nil
 	p.endpoint = ""
 	p.latestSeq = 0
-	p.blockBuff = nil
 }
 
 // PullBlock blocks until a block with the given sequence is fetched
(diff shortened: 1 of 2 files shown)


orderer/common/cluster/deliver.go
```diff
@@ -114,11 +114,18 @@ func (p *BlockPuller) HeightsByEndpoints() (map[string]uint64, error) {
 
 func (p *BlockPuller) tryFetchBlock(seq uint64) *common.Block {
 	var reConnected bool
-	for p.isDisconnected() {
+
+	for retriesLeft := p.MaxPullBlockRetries; p.isDisconnected(); {
```
yacovm, line 130: I don't think we need to attempt to pop if we exhausted the attempt count, because when we try to pull any block and fail doing so, we will anyway drop the entire buffer (when we return from **pullBlocks** with an error we close the puller and it resets the buffer), and whenever we close the puller we anyway reset the buffer. 

So, why not just return nil? I also argue that whenever we pull a sequence of blocks and keep them in the buffer, subsequent invocations of **tryFetchBlock** won't even try to connect anyway. It is only until the buffer is full, when we would detect we are disconnected.

Also let's move the retriesLeft-- to the for loop and do something like:

```

	for retriesLeft := p.MaxPullBlockRetries; p.isDisconnected(); retriesLeft-- {
		reConnected = true
		p.connectToSomeEndpoint(seq)
		if p.isDisconnected() {
			time.Sleep(p.RetryTimeout)
		}

		if p.MaxPullBlockRetries > 0 && retriesLeft <= 0 {
			p.Logger.Errorf("Failed connecting to some endpoint, attempts exhausted(%d)", p.MaxPullBlockRetries)
			return nil
		}
	}
```

yacovm (dismissed):

yacovm (approved):

PR #1697, opened 2020-08-10
Cache PKCS11 ObjectHandles

mastersingh24:
#### Type of change

- Improvement (performance)

#### Description

The pkcs11 implementation currently invokes
a find key operation for each sign and verify
operation.  This can be quite expensive especially
when communicating with network HSMs.

This change caches the ObjectHandle for private and
public keys as they should be valid across the current
PKCS11 session.

As ObjectHandles are only valid for the current session
context, if a stale session is found, the cache is cleared
as well.

### DIFF
--- a/bccsp/pkcs11/cache.go
+++ b/bccsp/pkcs11/cache.go
@@ -1,0 +1,47 @@
+/*
+Copyright IBM Corp. All Rights Reserved.
+
+SPDX-License-Identifier: Apache-2.0
+*/
+
+package pkcs11
+
+import (
+	"crypto/ecdsa"
+	"encoding/hex"
+	"sync"
+
+	"github.com/miekg/pkcs11"
+)
+
+type p11Key struct {
+	privH  pkcs11.ObjectHandle
+	pubH   pkcs11.ObjectHandle
+	pubKey *ecdsa.PublicKey
+}
+
+type keyCache struct {
+	mu      sync.RWMutex
+	p11Keys map[string]p11Key
+}
+
+func (k *keyCache) get(ski []byte) (p11Key, bool) {
+	k.mu.RLock()
+	defer k.mu.RUnlock()
+
+	h, ok := k.p11Keys[hex.EncodeToString(ski)]
+	return h, ok
+}
+
+func (k *keyCache) set(ski []byte, key p11Key) {
+	k.mu.Lock()
+	defer k.mu.Unlock()
+	k.p11Keys[hex.EncodeToString(ski)] = key
+
+}
+
+func (k *keyCache) clear() {
+	k.mu.Lock()
+	defer k.mu.Unlock()
+	k.p11Keys = make(map[string]p11Key)
+}
--- a/bccsp/pkcs11/pkcs11.go
+++ b/bccsp/pkcs11/pkcs11.go
@@ -76,6 +76,12 @@ func (csp *impl) getSession() (session pkcs11.SessionHandle, err error) {
 			logger.Warningf("Get session info failed [%s], closing existing session and getting a new session\n", err)
 			csp.ctx.CloseSession(session)
 
+			// since handles are only valid for the current session context, clear the p11Key cache
+			csp.p11Keys.clear()
+		} else {
+			logger.Debugf("Reusing existing pkcs11 session %+v on slot %d\n", session, csp.slot)
+		}
+
 		default:
 			// cache is empty (or completely in use), create a new session
 			return createSession(csp.ctx, csp.slot, csp.pin)
@@ -121,6 +127,14 @@ func (csp *impl) returnSession(session pkcs11.SessionHandle) {
 
 // Look for an EC key by SKI, stored in CKA_ID
 func (csp *impl) getECKey(ski []byte) (pubKey *ecdsa.PublicKey, isPriv bool, err error) {
+	// look for key in cache
+	if key, ok := csp.p11Keys.get(ski); ok {
+		if key.privH > 0 {
+			isPriv = true
+		}
+		return key.pubKey, isPriv, nil
+	}
+
 	p11lib := csp.ctx
 	session, err := csp.getSession()
 	if err != nil {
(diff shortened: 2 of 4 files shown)


mastersingh24:
Need to add some additional unit tests for `impl`, but wanted to see what people think so far

bccsp/pkcs11/pkcs11.go
```diff
@@ -398,7 +451,7 @@ const (
 	privateKeyType
 )
 
-func findKeyPairFromSKI(mod *pkcs11.Ctx, session pkcs11.SessionHandle, ski []byte, keyType keyType) (*pkcs11.ObjectHandle, error) {
+func findKeyPairFromSKI(mod *pkcs11.Ctx, session pkcs11.SessionHandle, ski []byte, keyType keyType) (pkcs11.ObjectHandle, error) {
```
sykesm, line 452: Perhaps break the cache into public and private key handle caches and maintain some of it here instead of at the call site?

bccsp/pkcs11/pkcs11.go
```diff
@@ -316,6 +337,16 @@ func (csp *impl) generateECKey(curve asn1.ObjectIdentifier, ephemeral bool) (ski
 		listAttrs(p11lib, session, pub)
 	}
 
+	// add key handles to cache
+	csp.p11Keys.set(
+		hex.EncodeToString(ski),
```
sykesm: nit: we can embed in the implementation instead of at the call site since the key at this level is the byte slice. This keeps the machinery for keys a details of the cache.

If we extend this to public/private handles, the cache can then generate keys appropriately based on context.

bccsp/pkcs11/cache.go
```diff
@@ -0,0 +1,46 @@
...
+	pubH   pkcs11.ObjectHandle
+	pubKey *ecdsa.PublicKey
+}
+
+type keyCache struct {
+	mu      sync.RWMutex
+	p11Keys map[string]p11Key
+}
+
+func (k *keyCache) get(ski string) (p11Key, bool) {
```
sykesm: The map needs a comparable key but the function doesn't. Maybe key conversion moves here?

bccsp/pkcs11/pkcs11.go
```diff
@@ -72,6 +72,9 @@ func (csp *impl) getSession() (session pkcs11.SessionHandle, err error) {
 			logger.Warningf("Get session info failed [%s], closing existing session and getting a new session\n", err)
 			csp.ctx.CloseSession(session)
 			session, err = createSession(csp.ctx, csp.slot, csp.pin)
+
+			// since handles are only valid for the current session context, clear the p11Key cache
+			csp.p11Keys.clear()
```
sykesm: Pretty much everything else is a nit except for this.

I'm not sure that this is quite sufficient. The select is not atomic with respect to the processing in each of the blocks. The way this is structured, if we get a session and think it's stale, we go off and close, recreate, and clear. If  this function can be called concurrently (and I don't see why it can't be) it can also do the same thing.

Now the cache clear and update are atomic but, depending on ordering, I think there are windows where the calls are interleaved such that a handle may appear to be valid but point to the wrong object. Don't know for sure, though.

sykesm (commented):
Mostly nits. My biggest worry is that the cache clear and session management are not atomic with respect to concurrent calls to get sessions. I think that can lead to situations where a session is created, gets a cached key with handles out of the cache, the cache is cleared, and then some other concurrent process causes the cache to get updated where the handle numbers no longer match. (They're just indexes into a vector, afaict.) Then the session can may mistakenly be using a handle to another object.

bccsp/pkcs11/cache.go
mastersingh24: I did think about doing the conversion in the get/set methods.  I'll go back and look, but I think I moved it out because there was a need for the conversation outside the cache.

bccsp/pkcs11/pkcs11.go
mastersingh24: I'll look into that .... I did think about that option .... need to look and see if there's a specific reason I did not go down that route

bccsp/pkcs11/pkcs11.go
mastersingh24: @sykesm  This is a tricky one and was one reason I was considering basing cache invalidation on invalid object handle errors rather than wiping out the entire cache for an invalid session.

For all sessions created in the same login context, the object handle for any given key will be the same (as far as I can tell from the spec) and any call to find key will return the same object handle for any given key.

An assumption I've made here is that if a session is invalid and when createSession is called a new login context in created, then all other sessions created under the prior login context will be invalid as well.  And in this case, you should not (cannot) end up using an object handle created under one context under a different context.

An alternative here would be to have a per session cache.

sykesm: I think it's probably simpler than that. I need a small break from hitting my head on the table so I'll take a quick look.

mastersingh24: I suppose we could have the same mutex on impl and the cache?  Then we could have an interleaved lock across both objects?

bccsp/pkcs11/pkcs11.go
```diff
@@ -362,9 +401,17 @@ func (csp *impl) verifyP11ECDSA(ski []byte, msg []byte, R, S *big.Int, byteSize
...
+	// check  cache for key
+	encodedSKI := hex.EncodeToString(ski)
+
+	if key, ok := csp.p11Keys.get(encodedSKI); ok {
+		pubHandle = key.pubH
+	} else {
+		pubHandle, err = findKeyPairFromSKI(p11lib, session, ski, publicKeyType)
+		if err != nil {
+			return false, fmt.Errorf("Public key not found [%s]", err)
+		}
```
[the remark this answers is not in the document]
mastersingh24: I did not break into separate caches, but did fix this to properly create/update an existing item in the cache

bccsp/pkcs11/pkcs11.go
mastersingh24: Done

bccsp/pkcs11/pkcs11.go
mastersingh24: Decided to stick with a single cache for now

bccsp/pkcs11/cache.go
mastersingh24: done

PR #1699, opened 2020-08-10
construct unreconciled missing pvtdata

cendhu:
#### Type of change

- New feature

#### Description

This PR adds a utility function to construct unreconciled missing pvtdata. 

#### Additional details

This utility function would be used by the reconciler to inform ledger about the unreconciled missing pvtdata. This information would be used by the pvtdata store to deprioritize these missing pvtdata and retried after N iterations of reconciliation. 

TxId field present in some of the structs defined in the reconciler/puller is not used. As some involve proto messages, it is not straightforward to remove it (due to rolling upgrade and change in the message format).

### DIFF
--- a/gossip/privdata/reconcile.go
+++ b/gossip/privdata/reconcile.go
@@ -316,3 +316,28 @@ func (r *Reconciler) groupRwsetByBlock(elements []*protosgossip.PvtDataElement) 
 	}
 	return rwSetByBlockByKeys
 }
+
+func constructUnreconciledMissingData(requestedMissingData privdatacommon.Dig2CollectionConfig, fetchedData []*protosgossip.PvtDataElement) ledger.MissingPvtDataInfo {
+	fetchedDataKeys := make(map[privdatacommon.DigKey]struct{})
+	for _, pvtData := range fetchedData {
+		key := privdatacommon.DigKey{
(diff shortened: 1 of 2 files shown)


gossip/privdata/reconcile.go
```diff
@@ -316,3 +316,29 @@ func (r *Reconciler) groupRwsetByBlock(elements []*protosgossip.PvtDataElement)
 	}
 	return rwSetByBlockByKeys
 }
+
+func constructUnreconciledMissingData(requestedMissingData privdatacommon.Dig2CollectionConfig, fetchedData []*protosgossip.PvtDataElement) ledger.MissingPvtDataInfo {
+	remainingRequestedData := requestedMissingData
```
manish-sethi: The name of the function does not indicate that the original input will be modified. If that's the intention, I would suggest to call it this function as `removeFetchedData` and do not return anything - instead let the caller prepared the `MissingPvtDataInfo` object.

cendhu: Yes. It does not sound good to manipulate the requested data. I have avoided removing entries from the requested data while constructing the unreconciled missing data.

manish-sethi (approved):

cendhu:

PR #1702, opened 2020-08-10
Use correct gossip SecretEnvelope reference

yacovm:
In case a membership request is received, it contains an inner
alive message envelope (which has a secret envelope).

The function handleMsgFromComm extracts an internal endpoint from
that secret envelope, however it was referencing the wrong envelope.

Instead of referencing the inner envelope of the alive message, it referenced
the outer envelope of the membership request, however, only alive messages
carry internal endpoints inside their secret envelopes.

As a result, all internal endpoints were always empty.

While this has no semantic effect, because the consumer of the internal endpoint doesn't
make use of the field, the field might still be logged, and we should use the correct
envelope which contains the internal endpoint and not the incorrect envelope.

Change-Id: I2452269e627e3444485425b0df5aced527d5149c

### DIFF
--- a/gossip/discovery/discovery_impl.go
+++ b/gossip/discovery/discovery_impl.go
@@ -339,8 +339,8 @@ func (d *gossipDiscoveryImpl) handleMsgFromComm(msg protoext.ReceivedMessage) {
 		}
 
 		var internalEndpoint string
-		if m.Envelope.SecretEnvelope != nil {
-			internalEndpoint = protoext.InternalEndpoint(m.Envelope.SecretEnvelope)
+		if memReq.SelfInformation.SecretEnvelope != nil {
+			internalEndpoint = protoext.InternalEndpoint(memReq.SelfInformation.SecretEnvelope)
 		}
 
 		// Sending a membership response to a peer may block this routine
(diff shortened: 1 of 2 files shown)


gossip/discovery/discovery_impl.go
```diff
@@ -339,8 +339,8 @@ func (d *gossipDiscoveryImpl) handleMsgFromComm(msg protoext.ReceivedMessage) {
 		}
 
 		var internalEndpoint string
-		if m.Envelope.SecretEnvelope != nil {
-			internalEndpoint = protoext.InternalEndpoint(m.Envelope.SecretEnvelope)
+		if memReq.SelfInformation.SecretEnvelope != nil {
+			internalEndpoint = protoext.InternalEndpoint(memReq.SelfInformation.SecretEnvelope)
```
sykesm, line 343: Tests?

> While this has no semantic effect, because the consumer of the internal endpoint doesn't make use of the field, the field might still be logged, and we should use the correct envelope which contains the internal endpoint and not the incorrect envelope.

That paragraph (and the fact that this PR is backported two releases) implies this is something that should be validated via automated test and assertions.

sykesm (dismissed):

gossip/discovery/discovery_impl.go
yacovm: Tests for what? we don't use this field in production code. At most it can be logged but nothing more.

yacovm: Added an assertion

sykesm (approved):
Thanks for adding the assertion.

PR #1706, opened 2020-08-10
Add Raft metrics to Raft config doc

joealewine:
Adding two metrics to Raft configuration doc to encourage
users to monitor their active nodes as compared to the
total number of nodes in the cluster

Change-Id: Ica35ee1a0a2feea6511cfd56fcee1bdf429b1334

#### Type of change

- Documentation update

#### Description

Adding two metrics to Raft configuration doc to encourage
users to monitor their active nodes as compared to the
total number of nodes in the cluster

### DIFF
--- a/docs/source/raft_configuration.md
+++ b/docs/source/raft_configuration.md
@@ -317,6 +317,12 @@ monitor:
    they can be shared with the consenter set. If this value begins to climb, this
    node may not be able to participate in consensus (which could lead to a
    service interruption for this node and possibly the network).
+* `consensus_etcdraft_cluster_size` and `consensus_etcdraft_active_nodes`: these
+   channel metrics help track the "active" nodes (which, as it sounds, are the nodes that
+   are currently contributing to the cluster, as compared to the total number of
+   nodes in the cluster). If the number of active nodes falls below a majority of
+   the nodes in the cluster, quorum will be lost and the ordering service will
+   stop processing blocks on the channel.
 
 ## Troubleshooting
 


joealewine:
@wlahti 

Took a shot at explaining this. Let me know if I got close enough to the mark.

docs/source/raft_configuration.md
```diff
@@ -317,6 +317,12 @@ monitor:
    they can be shared with the consenter set. If this value begins to climb, this
    node may not be able to participate in consensus (which could lead to a
    service interruption for this node and possibly the network).
+* `consensus_etcdraft_cluster_size` and `consensus_etcdraft_active_nodes`: these
```
wlahti, line 320: We should also mention that these metrics are at the channel level.

docs/source/raft_configuration.md
```diff
@@ -317,6 +317,12 @@ monitor:
...
+   metrics help track the "active" nodes (which, as it sounds, are the nodes that
```
wlahti: Missing the closing ).

wlahti (commented):
Thanks, Joe! Looks pretty good overall, just needs some updating to mention that we're talking about the channel level here.

joealewine:
> Thanks, Joe! Looks pretty good overall, just needs some updating to mention that we're talking about the channel level here.

Thanks for the quick review, Will. I've addressed those comments and pushed an update.

docs/source/raft_configuration.md
```diff
@@ -317,6 +317,12 @@ monitor:
    they can be shared with the consenter set. If this value begins to climb, this
    node may not be able to participate in consensus (which could lead to a
    service interruption for this node and possibly the network).
+* `consensus_etcdraft_cluster_size` and `consensus_etcdraft_active_nodes`: these
+   channel metrics help track the "active" nodes (which, as it sounds, are the nodes that
+   are currently contributing to the cluster, as compared to the total number of
+   nodes in the cluster). If the number of active nodes falls below a majority of
```
wlahti, line 323: Not sure if it would be redundant or useful to clarify that "the ordering service will stop processing blocks for the channel". Just quickly reading this makes it sound like the orderer will stop processing _all_ blocks but I'll defer to you based on what would be consistent with our other docs for the orderer/raft/consensus.

joealewine: My assumption was that people would understand that "stop processing blocks" referred to just the channel in question (since it's a channel metric), though unless channels have different sets of consenters it likely would be true for at least several channels. But I'm happy to add a slight clarification here to say "stop processing blocks on the channel".

wlahti (approved):
Looks good to me now. Thanks, Joe!

pamandrejko (approved):
Thanks for doing this Joe.

pamandrejko:

pamandrejko:

PR #1708, opened 2020-08-11
[FAB-18145] Implement CLI commands to submit/cancel/list snapshot requests

wenjianqiao:

#### Type of change
- New feature

#### Description
- peer snapshot submitrequest
- peer snapshot cancelrequest
- peer snapshot listpending

#### Additional Information
Currently, these commands are using blockNumber but server side is using channel height. Will change block height to block number through out the code in a separate PR. For now, blockNumber is treated as channel height.

#### Related issues
story: https://jira.hyperledger.org/browse/FAB-17690
subtask: https://jira.hyperledger.org/browse/FAB-18145


internal/peer/snapshot/cancelrequest.go
```diff
@@ -0,0 +1,84 @@
...
+		return errors.WithMessage(err, "failed to cancel the request")
+	}
+
+	fmt.Println("Snapshot request cancelled successfully")
+	return nil
+}
+
+func validateCancelRequest() error {
+	if channelID == "" {
+		return errors.New("the required parameter 'channelID' is empty. Rerun the command with -C flag")
```
lindluni, line 78: While it's commonly used throughout fabric, the `errors` package is deprecated and should no longer be used in places where `fmt.Errorf` can be used instead. We've been replacing it where we can as we push new changes.

caod123: Note `errors.New` is still appropriate but via the golang std library `errors` package and not through `github.com/pkg/errors`. `fmt.Errorf` is the appropriate replacement now for `errors.Errorf`. I think there are still many places in the code where we use `github.com/pkg/errors` for wrapped errors. That being said it's easy enough to create a wrapped error with `fmt.Errorf`: https://golang.org/pkg/errors/#pkg-overview

manish-sethi: Though this error here is much straight forward at the highest level, but in general, the most useful feature we have leveraging from `github.com/pkg/errors` is the inclusion of stack trace. In the `errors` package `Unwrap` is there but the primary intention is to check for a specific error type in the chain and handle (which we hardly do in Fabric). We can certainly leverage `Unwrap` and built-in mechanism of generating stack trace but that requires us to build almost what we have in `github.com/pkg/errors`. I believe that the Go2 seems to have some [proposal](https://go.googlesource.com/proposal/+/master/design/go2draft-error-values-overview.md) for printing line numbers of the source of the wrapped chain at the least. Until then, I would not like to loose on error stack traces which have been quite helpful. BTW, I am not sure whether `github.com/pkg/errors` is deprecated? https://github.com/pkg/errors#roadmap simply states that further feature addition is stopped with the upcoming Go2 error proposals which seems reasonable. So, I would be inclined towards keep using this unless you propose to build something similar within fabric (I hope not) until we get stack printing in some form built-in in the `errors`. Thoughts?

internal/peer/snapshot/submitrequest.go
```diff
@@ -0,0 +1,81 @@
...
+	cmd.SilenceUsage = true
+
+	// create client connection if not provided
+	if cc == nil {
+		var err error
+		cc, err = NewClientConnection(cmd.Name(), cryptoProvider)
+		if err != nil {
+			return err
+		}
+	}
```
manish-sethi: Just to take a note, here we have to compromise on the readability of the production code for testing with the mocks, primarily because of the flags. Looking at this code, one may wonder, who else would be creating the client (especially, the parameters of this function both are present, the `*ClientConnection` and the `cryptoProvider`)... not sure what is a cleaner way but that's how it seems to be in the existing code for the other commands as well.

internal/peer/snapshot/submitrequest.go
```diff
@@ -0,0 +1,81 @@
...
+	"github.com/hyperledger/fabric/bccsp"
+	"github.com/pkg/errors"
+	"github.com/spf13/cobra"
+)
+
+// submitRequestCmd returns the cobra command for snapshot submitrequest command
+func submitRequestCmd(clientConnection *ClientConnection, cryptoProvider bccsp.BCCSP) *cobra.Command {
+	snapshotSubmitRequestCmd := &cobra.Command{
+		Use:   "submitrequest",
+		Short: "Submit a snapshot request for the given block number.",
```
manish-sethi: Not sure if this conveys the message... especially, `snapshot for the` part. How about "Submit a request for a snapshot at the specified block"?

manish-sethi (commented):
I haven't looked at all the commands but I guess that the comments on `submitrequest` may be applicable to other commands as well.

internal/peer/snapshot/cancelrequest.go
yacovm: @btl5037 why is the `github.com/pkg/errors` deprecated?

caod123: It's not deprecated per se moreso just entering maintenance mode as mentioned in the link above in favor of Go2 errors once they're out. I'm fine with either approach, but it's something worth being aware of once Go2 is available. Given the usefulness of stack trace, continued usage of `github.com/pkg/errors` is likely fine for now.

wenjianqiao:
@manish-sethi Thank you. I have addressed your comments and pushed a new commit.

manish-sethi (approved):

PR #1709, opened 2020-08-11
Release 2.2   Multiple machine deployment,orderer warning

qiangqiang666:
QQ图片20200811112914

I used three servers, one to deploy Orderer and the other two to deploy different ORG, and after successful deployment, the chain code was also called. However, I was surprised to see these log warnings in Orderer, may I ask what might be the reason?
Ps: The IP address is one of my org addresses


sykesm:
The message you highlighted indicates the the client did not properly terminate the gRPC session.

For general questions and answers, please use our rocket chat, the mailing list, or stack overflow. A pull request is not a good way to ask a question. Everything is linked from the top level README: https://github.com/hyperledger/fabric#community

PR #1715, opened 2020-08-12
Refactor PrivateDataStore

cendhu:
#### Type of change

#### Description

1. Refactored `CommitPvtDataOfOldBlocks()`.
2. Moved the above and related functions to a separate file.

#### Additional details

1. The refactoring is needed before adding a PR for _deprioritize to be retried missing pvtdata_
2. The test will be refactored separately.


core/ledger/pvtdatastorage/reconcile_missing_pvtdata.go
```diff
@@ -0,0 +1,220 @@
...
+	batch := s.db.NewUpdateBatch()
+	if err := entries.addToUpdateBatch(batch); err != nil {
+		return err
+	}
+
+	logger.Debug("Committing the update batch to pvtdatastore")
+	return s.db.WriteBatch(batch, true)
+}
+
+func (s *Store) constructEntries(blocksPvtData map[uint64][]*ledger.TxPvtData) (*entriesForPvtDataOfOldBlocks, error) {
```
manish-sethi: While you are taking a stab at this, I guess that it will be further cleaner if we make it easy to differentiate the functions related to the path that processes old blocks. One option is to have a dedicated type, say,
`OldBlockDataProcessor` (that wraps the actual store) and have these functions (the functions that seem generic in name but are used only in the path of old block processing) on that type? The other option could be to keep adding OldPvtdata in the function names but that makes them unnecessarily longer.

manish-sethi: Just to clarify, we can construct this `OldBlockDataProcessor` in the function `CommitPvtDataOfOldBlocks`

cendhu: Agree. The goal is to first complete the deprioritization of missing pvtdata. I will do larger refactoring later. I made a few changes so that I can structure the deprioritization well.

cendhu: Misunderstood your comment. What you suggested can be done in this PR, itself. Pushing an update.

manish-sethi (approved):

PR #1716, opened 2020-08-12
[WIP] Prep work for Go 1.15

lindluni:
There is no intention of migrating to Go 1.15 until a later release when it has stabilized, this is simply prep-work that can be done without affecting the current implementations.

This commit series is only prep for Go 1.15 which has several changes that effect the existing behavior of our unit tests, in particular:

https://golang.org/doc/go1.15#vet -- Integers cast to strings now return a vet type check error enabled by default using `go test`
https://golang.org/doc/go1.15#commonname -- The Common Name of an x509 cert is no longer recognized as a SAN when no SAN's exists

Go 1.15 changes the default behavior of the x509 package. The Common Name is no longer recognized as a host name when a Subject Alternative Name is not specified. This change adds the hostname to the SANS section of the certs.


sykesm:
No objections, but can we move to generating them at runtime into a temporary directory instead of maintaining them as fixtures?

lindluni:
> No objections, but can we move to generating them at runtime into a temporary directory instead of maintaining them as fixtures?

Absolutely, I have additional changes coming in to address the rune typecasting, I'll come back and change this as well on this commit to generate at runtime

yacovm:
> https://golang.org/doc/go1.15#commonname -- The Common Name of an x509 cert is no longer recognized as a SAN when no SAN's exists


If a certificate doesn't have a SAN, has a CN that is a valid hostname then [commonNameAsHostname()](https://go-review.googlesource.com/c/go/+/231379/5/src/crypto/x509/verify.go#626) will make verification fail or succeed based on the value of `ignoreCN`.

The `ignoreCN` field in `crypto/verify/x509.go` default logic is now [flipped](https://go-review.googlesource.com/c/go/+/231379/5/src/crypto/x509/verify.go) flipped, so we need to at least document this, no?

lindluni:
> > https://golang.org/doc/go1.15#commonname -- The Common Name of an x509 cert is no longer recognized as a SAN when no SAN's exists
> 
> If a certificate is an intermediate or a root, doesn't have a SAN, has a CN that is a valid hostname then [commonNameAsHostname()](https://go-review.googlesource.com/c/go/+/231379/5/src/crypto/x509/verify.go#626) will make verification fail or succeed based on the value of `ignoreCN`.
> 
> The `ignoreCN` field in `crypto/verify/x509.go` default logic is now [flipped](https://go-review.googlesource.com/c/go/+/231379/5/src/crypto/x509/verify.go) flipped, so we need to at least document this, no?

Correct, I did bring it up with Dave this morning. This PR will make our tests work with both Go 1.14 and Go 1.15. When we move to Go 1.15 we will need to document it. It's going to be a nightmare as many users already don't use the SAN section and can't figure out why their handshakes fail, now the users who at least had a valid FQDN for the Common Name (but no SAN) are also going to start failing.

yacovm:
> > > https://golang.org/doc/go1.15#commonname -- The Common Name of an x509 cert is no longer recognized as a SAN when no SAN's exists
> > 
> > 
> > If a certificate is an intermediate or a root, doesn't have a SAN, has a CN that is a valid hostname then [commonNameAsHostname()](https://go-review.googlesource.com/c/go/+/231379/5/src/crypto/x509/verify.go#626) will make verification fail or succeed based on the value of `ignoreCN`.
> > The `ignoreCN` field in `crypto/verify/x509.go` default logic is now [flipped](https://go-review.googlesource.com/c/go/+/231379/5/src/crypto/x509/verify.go) flipped, so we need to at least document this, no?
> 
> Correct, I did bring it up with Dave this morning. This PR will make our tests work with both Go 1.14 and Go 1.15. When we move to Go 1.15 we will need to document it. It's going to be a nightmare as many users already don't use the SAN section and can't figure out why their handshakes fail, now the users who at least had a valid FQDN for the Common Name (but no SAN) are also going to start failing.

Oh I wasn't talking about TLS, I am talking about forking the blockchain. Now you essentially have a way to create a certificate that passes validation in version **A** but doesn't pass in version **B**.

lindluni:
> > > > https://golang.org/doc/go1.15#commonname -- The Common Name of an x509 cert is no longer recognized as a SAN when no SAN's exists
> > > 
> > > 
> > > If a certificate is an intermediate or a root, doesn't have a SAN, has a CN that is a valid hostname then [commonNameAsHostname()](https://go-review.googlesource.com/c/go/+/231379/5/src/crypto/x509/verify.go#626) will make verification fail or succeed based on the value of `ignoreCN`.
> > > The `ignoreCN` field in `crypto/verify/x509.go` default logic is now [flipped](https://go-review.googlesource.com/c/go/+/231379/5/src/crypto/x509/verify.go) flipped, so we need to at least document this, no?
> > 
> > 
> > Correct, I did bring it up with Dave this morning. This PR will make our tests work with both Go 1.14 and Go 1.15. When we move to Go 1.15 we will need to document it. It's going to be a nightmare as many users already don't use the SAN section and can't figure out why their handshakes fail, now the users who at least had a valid FQDN for the Common Name (but no SAN) are also going to start failing.
> 
> Oh I wasn't talking about TLS, I am talking about forking the blockchain. Now you essentially have a way to create a certificate that passes validation in version **A** but doesn't pass in version **B**.

I can't pretend to to have deep knowledge of the x509 verification process, but wouldn't this only affect the TLS certs? Signing certs would still be valid wouldn't they, as they don't have to go through a hostname verification like you do in the TLS handshake, so if they didn't include the SANS section, what would be the implication of the Go 1.15 change on a signing cert?

mastersingh24:
> I can't pretend to to have deep knowledge of the x509 verification process, but wouldn't this only affect the TLS certs? Signing certs would still be valid wouldn't they, as they don't have to go through a hostname verification like you do in the TLS handshake, so if they didn't include the SANS section, what would be the implication of the Go 1.15 change on a signing cert?

I was just looking at this yesterday ... we don't use SANs with enrollment/signing certs ... and of course we've always separated TLS and enrollment certs as well.

I believe that all TLS certs we generate (Fabric CA, cryptogen) for external use all contain SANs.
We need to double check the internal CA used for chaincode, but I believe it uses SANs as well.

yacovm:
> > I can't pretend to to have deep knowledge of the x509 verification process, but wouldn't this only affect the TLS certs? Signing certs would still be valid wouldn't they, as they don't have to go through a hostname verification like you do in the TLS handshake, so if they didn't include the SANS section, what would be the implication of the Go 1.15 change on a signing cert?
> 
> I was just looking at this yesterday ... we don't use SANs with enrollment/signing certs ... and of course we've always separated TLS and enrollment certs as well.
> 
> I believe that all TLS certs we generate (Fabric CA, cryptogen) for external use all contain SANs.
> We need to double check the internal CA used for chaincode, but I believe it uses SANs as well.

What if the CA is not generated by Fabric tooling? What if the CA is a third party one? There exists a configuration that rejects pre go1.5 validation chains, but accepts them post go1.5 and we should document this in my opinion, regardless if our tooling doesn't produce such a configuration, because our tooling sometimes is not used to produce the certificate configuration.


> > > > > https://golang.org/doc/go1.15#commonname -- The Common Name of an x509 cert is no longer recognized as a SAN when no SAN's exists
> > > > 
> > > > 
> > > > If a certificate is an intermediate or a root, doesn't have a SAN, has a CN that is a valid hostname then [commonNameAsHostname()](https://go-review.googlesource.com/c/go/+/231379/5/src/crypto/x509/verify.go#626) will make verification fail or succeed based on the value of `ignoreCN`.
> > > > The `ignoreCN` field in `crypto/verify/x509.go` default logic is now [flipped](https://go-review.googlesource.com/c/go/+/231379/5/src/crypto/x509/verify.go) flipped, so we need to at least document this, no?
> > > 
> > > 
> > > Correct, I did bring it up with Dave this morning. This PR will make our tests work with both Go 1.14 and Go 1.15. When we move to Go 1.15 we will need to document it. It's going to be a nightmare as many users already don't use the SAN section and can't figure out why their handshakes fail, now the users who at least had a valid FQDN for the Common Name (but no SAN) are also going to start failing.
> > 
> > 
> > Oh I wasn't talking about TLS, I am talking about forking the blockchain. Now you essentially have a way to create a certificate that passes validation in version **A** but doesn't pass in version **B**.
> 
> I can't pretend to to have deep knowledge of the x509 verification process, but wouldn't this only affect the TLS certs? Signing certs would still be valid wouldn't they, as they don't have to go through a hostname verification like you do in the TLS handshake, so if they didn't include the SANS section, what would be the implication of the Go 1.15 change on a signing cert?






Observe the following test and you will see what I mean:
```
func TestNameConstraints(t *testing.T) {
	caCert := `-----BEGIN CERTIFICATE-----
MIICWDCCAf6gAwIBAgIQEzpxUmpKr3isOWzPHzu+ZTAKBggqhkjOPQQDAjBpMQsw
CQYDVQQGEwJVUzETMBEGA1UECBMKQ2FsaWZvcm5pYTEWMBQGA1UEBxMNU2FuIEZy
YW5jaXNjbzEUMBIGA1UEChMLZXhhbXBsZS5jb20xFzAVBgNVBAMTDmNhLmV4YW1w
bGUuY29tMB4XDTIwMDgxMzEwMTIwMFoXDTMwMDgxMTEwMTIwMFowaTELMAkGA1UE
BhMCVVMxEzARBgNVBAgTCkNhbGlmb3JuaWExFjAUBgNVBAcTDVNhbiBGcmFuY2lz
Y28xFDASBgNVBAoTC2V4YW1wbGUuY29tMRcwFQYDVQQDEw5jYS5leGFtcGxlLmNv
bTBZMBMGByqGSM49AgEGCCqGSM49AwEHA0IABIhTmbATWCzZH2h993cJkysUy+W2
vUILG+0uWbqFIHXE2sOSMNZDtjMyC85gJ4c2is44si30QDdjrPbmJ18IzhajgYcw
gYQwDgYDVR0PAQH/BAQDAgGmMB0GA1UdJQQWMBQGCCsGAQUFBwMCBggrBgEFBQcD
ATAPBgNVHRMBAf8EBTADAQH/MCkGA1UdDgQiBCAbBLs7n95MxDtfT5C6OInES7r8
m8GgGU9Wlv6NO1UqXjAXBgNVHR4EEDAOoAwwCoIIYWNtZS5vcmcwCgYIKoZIzj0E
AwIDSAAwRQIhALHx6TpdDZFBfNWKCaNNrZLlwMYhVbgpOBBSLPsMd/8NAiAAwCWG
+SHCof2uJLiojVI20+Chb5bpu4BwvkzMnlMskA==
-----END CERTIFICATE-----`

	leafCert := `-----BEGIN CERTIFICATE-----
MIICCzCCAbKgAwIBAgIQVIDfbXWOout38odgJbze9DAKBggqhkjOPQQDAjBpMQsw
CQYDVQQGEwJVUzETMBEGA1UECBMKQ2FsaWZvcm5pYTEWMBQGA1UEBxMNU2FuIEZy
YW5jaXNjbzEUMBIGA1UEChMLZXhhbXBsZS5jb20xFzAVBgNVBAMTDmNhLmV4YW1w
bGUuY29tMB4XDTIwMDgxMzEwMTIwMFoXDTMwMDgxMTEwMTIwMFowWDELMAkGA1UE
BhMCVVMxEzARBgNVBAgTCkNhbGlmb3JuaWExFjAUBgNVBAcTDVNhbiBGcmFuY2lz
Y28xHDAaBgNVBAMTE29yZGVyZXIuZXhhbXBsZS5jb20wWTATBgcqhkjOPQIBBggq
hkjOPQMBBwNCAASWDhN4vwcP24xHnxVtOxNKRvVqV5zspA7fXhP7r7ve9GFO+Nfs
aUPxqXibdpIAV7VPWIcL9Z9Ejq5kvzApdH6bo00wSzAOBgNVHQ8BAf8EBAMCB4Aw
DAYDVR0TAQH/BAIwADArBgNVHSMEJDAigCAbBLs7n95MxDtfT5C6OInES7r8m8Gg
GU9Wlv6NO1UqXjAKBggqhkjOPQQDAgNHADBEAiB+2v4c+TwvuC5AAnyyXs8dJIyB
xkYi1KPtl8IA2PZPZAIgP77Zkv3mzVq2Vu3b5tRXHwQ+zAtZBeJa1xamIgg+Vkk=
-----END CERTIFICATE-----`

	bl, _ := pem.Decode([]byte(caCert))
	ca, _ := x509.ParseCertificate(bl.Bytes)

	bl, _ = pem.Decode([]byte(leafCert))
	leaf, _ := x509.ParseCertificate(bl.Bytes)

	caPool := x509.NewCertPool()
	caPool.AddCert(ca)

	_, err := leaf.Verify(x509.VerifyOptions{
		Roots: caPool,
		CurrentTime:  ca.NotBefore.Add(time.Microsecond),
	})

	fmt.Println(err) // it errors with "x509: issuer has name constraints but leaf doesn't have a SAN extension"
	// however if you run it with 'GODEBUG=x509ignoreCN=1' it doesn't error.
}
```

mastersingh24:
The real question is what stream in which to introduce this type of possibly breaking change.
We can document the `GODEBUG` option, but that's not the easiest thing for everyone to use especially if they have existing deployments.  In any case, we'd have to document that there is a breaking change here.

lindluni:
> The real question is what stream in which to introduce this type of possibly breaking change.
> We can document the `GODEBUG` option, but that's not the easiest thing for everyone to use especially if they have existing deployments. In any case, we'd have to document that there is a breaking change here.

fortunately we have 6 months to work out the details until 1.14 goes end of life

yacovm:
> The real question is what stream in which to introduce this type of possibly breaking change.
> We can document the `GODEBUG` option, but that's not the easiest thing for everyone to use especially if they have existing deployments. In any case, we'd have to document that there is a breaking change here.

The **only way** to be effected from this `ignoreCN` toggle change in the validation path, is if you have a CA with the name constraints extension and the leaf certificate doesn't have a SAN. In such a case, what will happen is that all transactions endorsed by such endorsers or config updates signed by such admins, will fail validation post ordering with pre 1.5 behavior.

The problem I see here, is that if something like this happens then the user will realize it and will issue certificates with SANs or will just update the CA certificate to not have the extension. Then, it will probably be forgotten with time, and when they upgrade to Go 1.5 and join new peers, they'll get a fork when processing these transactions by new peers because now these transactions are suddenly valid (ignoreCN is turned on by default).

Instead of setting an environment variable, can we not simply make the MSP enforce the legacy behavior by checking the condition manually? It's not computationally expensive since it's only to check:

- Do we have the name constraints extension on the issuer certificate?
- Does the leaf certificate lack SANs?
- Is the common name of the subject a valid hostname?

If yes then we fail the validation, and this preserves the pre 1.5 behavior.

Then again, I don't know if this extension is commonly used in industry / governments in the context of using blockchains, so it is possible that no one will really be effected by this and we really just need to document this and tell to activate the `x509ignoreCN` flag.

yacovm:
> The real question is what stream in which to introduce this type of possibly breaking change.
> We can document the `GODEBUG` option, but that's not the easiest thing for everyone to use especially if they have existing deployments. In any case, we'd have to document that there is a breaking change here.

The golang team [commented](https://github.com/golang/go/issues/39568#issuecomment-671424481) that the `x509ignoreCN` will be removed entirely in Go 1.16 so this is not a long term solution.

FiloSottile:
Sorry for the disruption folks. I just wanted to mention that crypto/x509 targets the WebPKI and evolves with it, so there are changes to the verification logic in almost all releases. This one just happens to be a larger one than usual. If you made X.509 verification logic consensus critical, you might want to fork crypto/x509 to have full control.

yacovm:
There are 2 kinds of changes that we care about:

1. Something that was **invalid** in the past logic, and is now valid.
2. Something that was **valid** in the past logic and now is invalid.

The first type is less of a problem because we can always add an explicit check in the logic that uses the x509 library, and I believe this case is of the first type.

The second type, however, is a problem because even if we catch an error from `Verify` and identify that it wasn't a problem in the past, we don't know if `Verify` would've failed further down in the execution, so we cannot return `nil` instead.

> I just wanted to mention that crypto/x509 targets the WebPKI and evolves with it, so there are changes to the verification logic in almost all releases. 

@FiloSottile do you have an example at hand for a change of the second type in the last 3-4 years that had a significant impact?

FiloSottile:
I don't have examples off the top of my head, you should review the commit history of the crypto/x509 package and keep in mind that we don't provide a backwards compatibility promise that's strict enough for consensus critical applications.

yacovm:
I found another issue with upgrading to Go 1.15: If you append trailing bytes to the signature on the CRL, Go 1.15 validation fails because it now uses [ecdsa.VerifyASN1](https://github.com/golang/go/commit/975c01342a25899962969833d8b2873dc8856a4f#) however older versions succeed because they use the lenient [asn1.Unmarshal](https://golang.org/pkg/encoding/asn1/#Unmarshal).

**To reproduce:** Replace the contents of this [CRL](https://github.com/hyperledger/fabric/blob/master/msp/testdata/revocation/crls/crl.pem) in the [testdata/revocation](https://github.com/hyperledger/fabric/tree/master/msp/testdata/revocation) folder with this:
```
-----BEGIN X509 CRL-----
MIIBODCB3AIBATAKBggqhkjOPQQDAjBXMQswCQYDVQQGEwJVUzETMBEGA1UECBMK
Q2FsaWZvcm5pYTEWMBQGA1UEBxMNU2FuIEZyYW5jaXNjbzEOMAwGA1UEChMFcm9v
dDIxCzAJBgNVBAMTAmNhFw0xNzEyMjExNzExNTRaFw0yNzEyMTkxNzExNTRaMCMw
IQIQaiOerd7fYdLv3WOe3G7maRcNMTcxMjIxMTcxMTU0WqAvMC0wKwYDVR0jBCQw
IoAgXQu9qpyvtmox2hOtTuxMDMUyYKoJu3r8ftlgn2TH8TowCgYIKoZIzj0EAwID
SwAwSAIhAOZOMOJps28IhWbg4HBibE2ft5S4pChxVT9AXFOVE0WCAiB/NFueHuNi
h+VjDOTEo7xzWdCWfGLDcDBfiofROn+1AQIBZA==
-----END X509 CRL-----
```

and then run **TestRevocation** and see it [fails](https://github.com/hyperledger/fabric/blob/master/msp/revocation_test.go#L40) with Go 1.15 but passes with Go 1.14.6 

Fortunately we can easily fix this by sanitizing the ECDSA signature [when we setup the CRLs](https://github.com/hyperledger/fabric/blob/master/msp/mspimplsetup.go#L197-L214) similarly to what we do with certificates. 

**Fun fact:** I _don't think_ we have the equivalent problem with identities because [we always sanitize the signature of the CA](https://github.com/hyperledger/fabric/blob/release-1.4/bccsp/utils/ecdsa.go#L69) (in case it's ECDSA) [regardless of which side of half the curve order the 's' value is found at](https://github.com/hyperledger/fabric/commit/506691955bd177996864c8ac2c40039b50c1f537),  this has the fortunate consequence that it ensures the ECDSA signatures of CAs are properly formed when they reach the x509 verification.

lindluni:
Matt patched the Runes and the RegExp in a separate PR. Will reopen with the SANS work and generating the  certs at runtime in a separate PR.

PR #1717, opened 2020-08-12
Exercise a full end-to-end flow with PKCS11

sykesm:
Execute the basic query-invoke-query pattern against a solo network with BCCSP configured to use PKCS11.

### DIFF
--- a/integration/pkcs11/pkcs11_test.go
+++ b/integration/pkcs11/pkcs11_test.go
@@ -3,6 +3,7 @@ Copyright IBM Corp. All Rights Reserved.
 
 SPDX-License-Identifier: Apache-2.0
 */
+
 package pkcs11
 
 import (
@@ -23,7 +24,6 @@ import (
 	"syscall"
 	"time"
 
-	docker "github.com/fsouza/go-dockerclient"
 	bpkcs11 "github.com/hyperledger/fabric/bccsp/pkcs11"
 	"github.com/hyperledger/fabric/integration/nwo"
 	"github.com/hyperledger/fabric/integration/nwo/fabricconfig"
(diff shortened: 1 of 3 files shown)


caod123 (approved):

integration/pkcs11/pkcs11_test.go
```diff
@@ -33,89 +33,106 @@ import (
 	"github.com/tedsuo/ifrit"
 )
 
-var _ = Describe("Configures PKCS#11 for peer and orderer", func() {
+var _ = Describe("PKCS11 enabled network", func() {
```
ale-linux, line 36: Very minor but the _real_ name of the standard is PKCS #11 (see here https://en.wikipedia.org/wiki/PKCS_11). So while the previous string was incorrect due to the missing space, your's isn't correct either. If you are okay with changing I'd change it here and everywhere else.

integration/pkcs11/pkcs11_test.go
```diff
@@ -125,15 +142,18 @@ func configurePeerPKCS11(ctx *pkcs11.Ctx, sess pkcs11.SessionHandle, network *nw
...
 		newOrdererPemCert := buildCert(caBytes, orgCAPath, peerCSR, peerSerial, peerPubKey)
 		updateMSPFolder(network.PeerLocalMSPDir(peer), fmt.Sprintf("peer.%s-cert.pem", domain), newOrdererPemCert)
 
 		By("Updating the peer admin user signcerts")
 		newAdminPemCert := buildCert(caBytes, orgCAPath, adminCSR, adminSerial, adminPubKey)
-
 		orgAdminMSPPath := network.PeerUserMSPDir(peer, "Admin")
 		updateMSPFolder(orgAdminMSPPath, fmt.Sprintf("Admin@%s-cert.pem", domain), newAdminPemCert)
+
+		By("Updating the peer user1 signcerts")
```
ale-linux, line 153: Let me see if I understand why you added this step: by updating the _user_ cert (that is, the one the CLI will use to sign proposal and transaction), are you ensuring that PKCS #11 is also used by the CLI? Assuming I got this right, would you agree that this would never happen in production since proposal/tx signing will be performed by the SDK which we don't exercise here?

ale-linux (commented):
Just a very minor nit and a question (and I now realise that in the meantime the PR already got merged but maybe you could still answer if you have a sec).

PR #1720, opened 2020-08-13
Populate Pvtdata purgemgr from snapshot files

manish-sethi:

#### Type of change
- New feature

#### Description

This PR
- Enables passing on the private data hashes to other consumers while loading the private data hashes from snapshot files into the statedb
- Registers Pvtdata PurgeMgr as a conusmer and populates PurgeMgr with the applicable expiry entries

#### Additional details
- The above mechanism is intended to be used to populate the pvtdata storage with the missing data entries
- In a future PR, we will explore if the data loading in the PurgeMgr can be made more performant via some sort of batching of writes or making separate entries for each write

#### Related issues
FAB-18154


core/ledger/kvledger/txmgmt/pvtstatepurgemgmt/snapshot_purge_mgr_builder.go
```diff
@@ -0,0 +1,69 @@
...
+	if err := b.expKeeper.update([]*expiryInfo{expInfo}, nil); err != nil {
+		return err
+	}
+	return nil
+}
+
+// Done implements the function in the interface privacuenabledstate.SnapshotPvtdataHashesConsumer.
+func (b *PurgeMgrBuilder) Done() error {
+	return nil
+}
```
cendhu: I understand that it will be useful when we do batch updates. Similar to purgeMgr, we may not be able to apply batch updates in the pvtdatastore also. Hence, IMO, we can remove it and add it if needed.

core/ledger/kvledger/txmgmt/pvtstatepurgemgmt/snapshot_purge_mgr_builder_test.go
```diff
@@ -0,0 +1,182 @@
...
+func TestPurgeMgrBuilder(t *testing.T) {
+	bookkeepingEnv := bookkeeping.NewTestEnv(t)
+	defer bookkeepingEnv.Cleanup()
+
+	ledgerID := "test-ledger"
+	bookkeepingProvider := bookkeepingEnv.TestProvider
+	btlPolicy := btltestutil.SampleBTLPolicy(
+		map[[2]string]uint64{
+			{"ns1", "coll1"}: 1,
+			{"ns2", "coll2"}: 2,
```
cendhu, line 30: one entry with 0 BTL to cover never expiring entries.

core/ledger/kvledger/txmgmt/pvtstatepurgemgmt/snapshot_purge_mgr_builder_test.go
```diff
@@ -0,0 +1,182 @@
...
+	require.NoError(t,
+		purgeMgrBuilder.ConsumeSnapshotData("ns1", "coll1", []byte("key-hash-2"), version.NewHeight(5, 1).ToBytes()),
+	)
+	// add data that expires at block-8
+	require.NoError(t,
+		purgeMgrBuilder.ConsumeSnapshotData("ns2", "coll2", []byte("key-hash-2"), version.NewHeight(5, 1).ToBytes()),
+	)
+	// add data that never expires
+	require.NoError(t,
+		purgeMgrBuilder.ConsumeSnapshotData("ns1", "no-btl-collection", []byte("key-hash-3"), version.NewHeight(5, 1).ToBytes()),
```
cendhu: If `GetExpiringBlock()` could not find the collection, wouldn't it return `NoSuchCollectionError`? Why isn't it returning error here?

core/ledger/kvledger/txmgmt/privacyenabledstate/snapshot_test.go
```diff
@@ -643,3 +645,98 @@ func TestSnapshotImportErrorPropagation(t *testing.T) {
...
+}
+
+func TestSnapshotImportPvtdataHashesConsumer(t *testing.T) {
+	var dbEnv *LevelDBTestEnv
+	var snapshotDir string
+	var cleanup func()
+	var err error
+
+	init := func() {
+		dbEnv = &LevelDBTestEnv{}
```
cendhu, line 660: may not cover  `versionFromSnapshotValue = statecouchdb.VersionFromSnapshotValue`?

core/ledger/kvledger/txmgmt/privacyenabledstate/snapshot_test.go
```diff
@@ -643,3 +645,98 @@ func TestSnapshotImportErrorPropagation(t *testing.T) {
...
+		dbEnv.Init(t)
+		db := dbEnv.GetDBHandle(generateLedgerID(t))
+		updateBatch := NewUpdateBatch()
+		updateBatch.PubUpdates.Put("ns-1", "key-1", []byte("value-1"), version.NewHeight(1, 1))
+		updateBatch.HashUpdates.Put("ns-1", "coll-1", []byte("key-hash-1"), []byte("value-hash-1"), version.NewHeight(1, 1))
+		require.NoError(t, db.ApplyPrivacyAwareUpdates(updateBatch, version.NewHeight(1, 1)))
+		snapshotDir, err = ioutil.TempDir("", "testsnapshot")
+		require.NoError(t, err)
+		_, err := db.ExportPubStateAndPvtStateHashes(snapshotDir, testNewHashFunc)
+		require.NoError(t, err)
```
cendhu, line 678: better to check for `err` explicitly and do cleanup before calling `t.Fatalf`

core/ledger/kvledger/snapshot.go
```diff
@@ -269,8 +275,15 @@ func (p *Provider) CreateFromSnapshot(snapshotDir string) (ledger.PeerLedger, er
 				errors.WithMessage(err, "error while importing data into config history Mgr"),
 			)
 	}
+	btlPolicy := pvtdatapolicy.ConstructBTLPolicy(
+		&mostRecentCollectionConfigFetcher{
+			DeployedChaincodeInfoProvider: p.initializer.DeployedChaincodeInfoProvider,
```
cendhu, line 280: Not related to this PR. We need to somehow get rid of this dependency. Had to spend some time to understand the difference between this `ConstructBTLPolicy()` and the following one https://github.com/hyperledger/fabric/blob/bbea1f582f360f2557314dd2aa3abb5c8e33a4b8/core/ledger/kvledger/kv_ledger.go#L101

core/ledger/kvledger/snapshot.go
```diff
@@ -371,3 +384,28 @@ func verifyFileHash(dir, file string, expectedHashInHex string, hashProvider led
...
+func (c *mostRecentCollectionConfigFetcher) CollectionInfo(chaincodeName, collectionName string) (*peer.StaticCollectionConfig, error) {
+	isImplicitCollection, mspID := implicitcollection.MspIDIfImplicitCollection(collectionName)
+	if isImplicitCollection {
+		return c.GenerateImplicitCollectionForOrg(mspID), nil
+	}
+
+	explicitCollections, err := c.MostRecentCollectionConfigBelow(math.MaxUint64, chaincodeName)
+	if err != nil || explicitCollections == nil || explicitCollections.CollectionConfig == nil {
+		return nil, errors.WithMessage(err, "error while fetching most recent collection config")
+	}
```
cendhu, line 402: Maybe we can return `nil` if the collectionConfig is `nil`? Anyway, `GetBTL()` does not expect an error for a `nil` collectionConfig https://github.com/hyperledger/fabric/blob/bbea1f582f360f2557314dd2aa3abb5c8e33a4b8/core/ledger/pvtdatapolicy/btlpolicy.go#L58-L64

core/ledger/kvledger/snapshot.go
```diff
@@ -371,3 +384,28 @@ func verifyFileHash(dir, file string, expectedHashInHex string, hashProvider led
 	}
 	return nil
 }
+
+type mostRecentCollectionConfigFetcher struct {
```
cendhu, line 388: maybe a ledger level integration test is needed as there are many components involved. Can be done in a separate PR as well.

cendhu (commented):
Thanks, Manish.

core/ledger/kvledger/snapshot.go
manish-sethi: That's exactly what this code is doing

core/ledger/kvledger/snapshot.go
manish-sethi: Not sure that I understand this comment...
There is a unit test (`TestMostRecentCollectionConfigFetcher`) for this code in this package. One high level test (`TestSnapshotGenerationAndNewLedgerCreation`) that exercises this in a e2e flow, also exists in this package. There will be final integration tests related to sanpshotting.
What else is the suggestion?

core/ledger/kvledger/snapshot.go
manish-sethi: Which dependency are you referring to? `ConstructBTLPolicy` or `DeployedChaincodeInfoProvider`?

core/ledger/kvledger/txmgmt/privacyenabledstate/snapshot.go
```diff
@@ -419,3 +473,8 @@ func (c *cursor) canMove() bool {
 func (c *cursor) currentNamespace() string {
 	return c.metadata[c.currentRow].namespace
 }
+
+type SnapshotPvtdataHashesConsumer interface {
+	ConsumeSnapshotData(namespace, coll string, keyHash []byte, version []byte) error
```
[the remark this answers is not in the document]
cendhu: The pvtdatastore needs both the block and transaction number. I am fine addressing this in a separate PR.

core/ledger/kvledger/txmgmt/pvtstatepurgemgmt/snapshot_purge_mgr_builder.go
cendhu: I see. Agree.

core/ledger/kvledger/txmgmt/privacyenabledstate/snapshot_test.go
cendhu: What I meant was different. If we remove `versionFromSnapshotValue = statecouchdb.VersionFromSnapshotValue` from db.go by mistake in the future, all tests in this package will pass. I know that we would not simply remove it but keeping that under check is good IMO.

core/ledger/kvledger/snapshot.go
cendhu: `DeployedChaincodeInfoProvider`

core/ledger/kvledger/snapshot.go
cendhu: Ah. Yeah, I was mistaken. The second/third time I misread this pattern
```
	if err != nil || explicitCollections == nil || explicitCollections.CollectionConfig == nil {
		return nil, errors.WithMessage(err, "error while fetching most recent collection config")
	} 
```
where the err can be nil here.

core/ledger/kvledger/snapshot.go
cendhu: The prugeMgr is not tested in the high-level test. What I suggested was that we need a test at `kvledger/tests` to ensure that `pvtdatapolicy`, `privacyenabledstate`, `purgemgr`, and other dependencies play well together while constructing the purgemgr. For example, the BTLPolicy used in the prugeMgr test is not the real one and it has a different behavior.

cendhu (approved):
Thanks, Manish. A few comments can be addressed in a separate PR.

core/ledger/kvledger/snapshot.go
manish-sethi: OK, got puzzled why you singled out this particular piece of code. There are many others aspects and scenarios that are unit tested and will get tested in the integration tests (at kvledger/tests and real integration tests). You saw one bug that was discovered with the real channel config. Yes, we are getting closer.

core/ledger/kvledger/snapshot.go
manish-sethi: Not sure what you meant by getting rid of this dependency. In general, it's not desired to get rid of the dependency - `DeployedChaincodeInfoProvider`, as the primary purpose of this is to contain the chaincode datamodel details in the chaincode package. However, what we should be doing is that the logic of implicit collection should be separated out from this in a separate dependency. I had started created a separate package for this - https://github.com/hyperledger/fabric/pull/1692

core/ledger/kvledger/txmgmt/privacyenabledstate/snapshot_test.go
manish-sethi: Ah, sure. yes, makes sense covering this scenario

core/ledger/kvledger/txmgmt/privacyenabledstate/snapshot.go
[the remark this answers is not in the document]
manish-sethi: Yes, I missed bitset part, will return version object from this API

PR #1724, opened 2020-08-14
FAB-18163 TLS Timeshift w/o Separate Cluster Port

jyellick:
#### Type of change

#### Description

The existing orderer code allows for a TLS timeshift to be specified
when starting the orderer with a separate cluster port.  This is useful
for situations where the consenter TLS certificates have accidentally
expired, preventing quorum from forming, and preventing channel config
updates to replace these certificates.

This change extends the concept to the general orderer port, which can
be optionally re-used, rather than binding to a new port.  If the
cluster port is not re-used, then the TLS timeshift parameter specified
in the general TLS section takes precdence over the parameter specified
in the cluster section.  Likewise, if there is a separate cluster port,
the parameter in the general TLS section has no impact on the cluster
listener.

### DIFF
--- a/integration/nwo/fabricconfig/orderer.go
+++ b/integration/nwo/fabricconfig/orderer.go
@@ -20,7 +20,7 @@ type Orderer struct {
 
 type General struct {
 	ListenAddress   string                 `yaml:"ListenAddress,omitempty"`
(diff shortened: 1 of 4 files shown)


yacovm (approved):

jyellick:

jyellick:

PR #1731, opened 2020-08-14
Pass snapshot metadata to init pvtdatastorage store properly

wenjianqiao:

#### Type of change
- Bug fix

#### Description
Add an initializer to pass snapshot metadata to open pvtdatastorage store so that it can init state properly.

#### Related issues
https://jira.hyperledger.org/browse/FAB-17689


wenjianqiao:
Close the PR as the real fix will be handled when pvtdata store is populated from snapshot.

PR #1735, opened 2020-08-15
Chan.Part.API IT: Join as follower and transition to member

wlahti:
#### Type of change

- Test update

#### Description

Integration test for join as follower and transition to member. 

#### Related issues

[FAB-18031](https://jira.hyperledger.org/browse/FAB-18031)

### DIFF
--- a/integration/raft/channel_participation_test.go
+++ b/integration/raft/channel_participation_test.go
@@ -108,13 +108,16 @@ var _ = Describe("ChannelParticipation", func() {
 			orderer2 := network.Orderer("orderer2")
 			orderer3 := network.Orderer("orderer3")
 			orderers := []*nwo.Orderer{orderer1, orderer2, orderer3}
+			members := []*nwo.Orderer{orderer1, orderer2}
 			peer := network.Peer("Org1", "peer0")
+
+			By("starting all three orderers")
 			for _, o := range orderers {
 				startOrderer(o)
 				channelparticipation.List(network, o, nil)
 			}
 
-			genesisBlock := applicationChannelGenesisBlock(network, orderers, peer, "participation-trophy")
+			genesisBlock := applicationChannelGenesisBlock(network, members, peer, "participation-trophy")
 			expectedChannelInfoPT := channelparticipation.ChannelInfo{
 				Name:            "participation-trophy",
 				URL:             "/participation/v1/channels/participation-trophy",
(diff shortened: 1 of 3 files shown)


wlahti:
This test is flakey without: https://github.com/hyperledger/fabric/pull/1734

wlahti:
And yep, hit the expected error. :)

integration/raft/channel_participation_test.go
```diff
@@ -123,59 +126,101 @@ var _ = Describe("ChannelParticipation", func() {
...
+				Name:            "participation-trophy",
+				URL:             "/participation/v1/channels/participation-trophy",
+				Status:          "onboarding",
+				ClusterRelation: "follower",
+				Height:          0,
+			}
+			channelparticipation.Join(network, orderer3, "participation-trophy", configBlockPT, expectedChannelInfoPTFollower)
+			channelInfo := channelparticipation.ListOne(network, orderer3, "participation-trophy")
+			Expect(channelInfo).To(Equal(expectedChannelInfoPTFollower))
+
```
tock-ibm: I think this may generate a race on the height and status. If the REST query is faster than the follower pulling the first block, no problem. However if the REST query is slow and the follower is fast, we'll get a mismatch, as the height will be >0. By the same logic, if the follower is really fast, it will finish onboarding....

wlahti: Agreed. Removed this check as it's unnecessary and not guaranteed to be equal.

integration/raft/channel_participation_test.go
```diff
@@ -123,59 +126,99 @@ var _ = Describe("ChannelParticipation", func() {
...
+			newLeader := findLeader([]*ginkgomon.Runner{ordererRunners[2]})
+			Expect(newLeader).To(Equal(leader))
+
+			By("submitting transaction to orderer3 to ensure it is active")
+			env = CreateBroadcastEnvelope(network, peer, "participation-trophy", []byte("hello-again"))
+			expectedChannelInfoPT.Height++
+			resp, err := ordererclient.Broadcast(network, orderer3, env)
+			Expect(err).NotTo(HaveOccurred())
+			Expect(resp.Status).To(Equal(common.Status_SUCCESS))
+			expectedBlockNumPerChannel := map[string]int{"participation-trophy": int(expectedChannelInfoPT.Height - 1)}
```
jyellick, line 199: I _think_ this number should be entirely predictable.  If so, I'd much rather see us using literals than storing it in a value which is mutated throughout the test.  If we change behavior, such that the block number is now wrong, then we want our test to catch the the behavior changed.  If the behavior changes and we understand and accept that, it's easy enough to change a literal when the test fails.

jyellick (approved):
Merging, but I think these tests might be worth going through and doing a quick audit to ensure we're using literals where we can.

PR #1771, opened 2020-08-20
[FAB-18171] Disregard certificate validity period in intra-orderer communication

yacovm:
This change makes the orderer cluster authentication infrastructure
disregard validity periods when comparing certificates, and only regard public keys.

With this change, one can replace the TLS certificate of Raft,
by a certificate that has the same public key without issuing channel config updates.

This is to ensure that if a certificate of a consenter expired,
one can quickly renew it and start the orderer without performing
a config update per channel.

Change-Id: Id1b29e220d37aa33617b2143b702075bf5b01f6f

### DIFF
--- a/common/crypto/expiration_test.go
+++ b/common/crypto/expiration_test.go
@@ -7,6 +7,9 @@ SPDX-License-Identifier: Apache-2.0
 package crypto
 
 import (
+	"bytes"
+	"encoding/pem"
+	"errors"
 	"fmt"
 	"io/ioutil"
 	"path/filepath"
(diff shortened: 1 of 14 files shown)


common/crypto/expiration_test.go
```diff
@@ -17,6 +18,7 @@ import (
 	"github.com/hyperledger/fabric-protos-go/msp"
 	"github.com/hyperledger/fabric/common/crypto/tlsgen"
 	"github.com/hyperledger/fabric/protoutil"
+	"github.com/stretchr/testify/assert"
```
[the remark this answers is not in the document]
yacovm: Well, looks like the push wasn't too successful:
```
yacovm@~/gopath/src/github.com/hyperledger/fabric (FAB-18171-m) $ grep -re "assert\." * | grep -v "^vendor/"| wc -l
298
```

I'll see if people give me comments on the code itself and if I'll need to push a new version I'll change the `assert` to `require`.

orderer/common/cluster/util.go
```diff
@@ -73,7 +74,7 @@ func (mp MemberMapping) ByID(ID uint64) *Stub {
 // LookupByClientCert retrieves a Stub with the given client certificate
 func (mp MemberMapping) LookupByClientCert(cert []byte) *Stub {
 	for _, stub := range mp {
-		if bytes.Equal(stub.ClientTLSCert, cert) {
+		if crypto.CertificatesWithSamePublicKey(stub.ClientTLSCert, cert) == nil {
```
jyellick: Not an expert in this area of the code, but I'm assuming this ends up cached to prevent constantly iterating over and parsing the certs, looking for a match?

jyellick (commented):
Overall structure and approach looks good to me.  Had a few comments, but mostly more questions than suggested changes.

orderer/consensus/etcdraft/consenter_test.go
```diff
@@ -41,6 +42,17 @@ import (
 	"go.uber.org/zap/zapcore"
 )
 
+var (
+	sampleCert = "LS0tLS1CRUdJTiBDRVJUSUZJQ0FURS0tLS0tCk1JSUJxakNDQVZHZ0F3SU" +
```
jyellick: In general, if we're going to have complex binary literals, at least a comment mentioning how they were generated would be appreciated.

orderer/consensus/etcdraft/util_test.go
```diff
@@ -38,6 +40,16 @@ func TestIsConsenterOfChannel(t *testing.T) {
 		"ZJQ0FURS0tLS0tCg==")
 	require.NoError(t, err)
 
+	someRandomCert, err := base64.StdEncoding.DecodeString("LS0tLS1CRUdJTiBDRVJUSUZJQ0FURS0tLS0tCk1JSUJxakNDQVZHZ0F3SU" +
```
jyellick: Same comment here for how to generate this cert.  It seems like we really mostly don't care about the contents, as implied by the name, but it very likely will not be clear to the next reader.

orderer/consensus/etcdraft/util.go
```diff
@@ -297,10 +299,30 @@ func (conCert ConsenterCertificate) IsConsenterOfChannel(configBlock *common.Blo
...
+		candidateBlock, _ = pem.Decode(consenter.ClientTlsCert)
+		if candidateBlock == nil {
+			return errors.Errorf("candidate client certificate %s is not a valid PEM", string(consenter.ClientTlsCert))
+		}
+		sameClientCertErr := crypto.CertificatesWithSamePublicKey(myCertDER, candidateBlock.Bytes)
+
+		if sameServerCertErr == nil || sameClientCertErr == nil {
 			return nil
 		}
+		conCert.Logger.Debugf("I am not %s:%d because %s, %s", consenter.Host, consenter.Port, sameServerCertErr, sameClientCertErr)
```
jyellick, line 326: If I read this correctly, because you've made the optimization around deferring the error text computation, when in this log message, these errors will no longer have the public key bytes rendered?

Perhaps here is not the right place... but I expect it would be helpful in debugging to ensure that we are printing the hex or base64 encoded bytes of our public key, and those of each certificate in the channel config somewhere.  Perhaps it's good enough that we are printing full PEMs in other places, which can always be used to extract the public key, but, just a thought.

orderer/common/cluster/util_test.go
```diff
@@ -1187,3 +1189,41 @@ qz8bY+s6AiBvO0aOfE8M4ibjmRE4vSXo0+gkOIJKqZcmiRdnJSr8Xw==
...
+		F: func(a, b []byte) bool {
+			invocations++
+			return bytes.Equal(a, b)
+		},
+	}
+
+	// Warm-up cache
+	for i := 0; i < 5; i++ {
+		notSame := m.Compare([]byte{byte(i)}, []byte{1, 2, 3})
+		assert.False(t, notSame)
```
jyellick: Mostly a nit, but most all of these asserts would probably be better as requires.

jyellick (commented):
Thanks for the updates.  I hate the additional complexity of including the caching layer, but, because of the potential for accidental self-DOS, I think it's worthwhile.  Also seems like a nice solution to defer computing the log text until the places where it's needed, tied to the error type itself.

A few comments, mostly nits around the testing.  But looks more or less ready to go to me.

orderer/consensus/etcdraft/util_test.go
yacovm: I removed it.

orderer/consensus/etcdraft/consenter_test.go
yacovm: Removed it

orderer/common/cluster/util_test.go
yacovm: I don't see why? A require is needed in case the test might fail further along in the flow, which is like when you check something is not nil and then a require would result in a nil pointer panic. An assert is perfectly fine here, especially since the test is small and serial and easy to debug. I changed to assert nonetheless, just to adhere to our (false) coding bible.

orderer/consensus/etcdraft/util.go
yacovm: > If I read this correctly, because you've made the optimization around deferring the error text computation, when in this log message, these errors will no longer have the public key bytes rendered?

You are not reading this correctly, this is unrelated. 
I put caching in a single place - when you read the certificate from the gRPC stream as a server, you cache the invocations of the function. All other invocations are not cached:
- Client side invocations (you as a client, authenticate a server) do not require caching because you connect once and persist the TLS connection until you disconnect. 
- All replication related invocations are either due to onboarding which is slow anyway, or are when Raft catches up, which means you are behind the rest, hence not contributing to any throughput and not in the critical path.

> Perhaps here is not the right place... but I expect it would be helpful in debugging to ensure that we are printing the hex or base64 encoded bytes of our public key, and those of each certificate in the channel config somewhere. 

I added a print in case a match is not found, which is most likely when you would want to troubleshoot.

orderer/common/cluster/util_test.go
jyellick: My assertion would be that any time a failure higher in a test case would cause future failures to be expected, you are better to require than to assert.

For example, testing how a sqrt function handles positive, negative, and zero cases would make sense as asserts

```
assert.Equal(t, mySqrt(9.0), 3.0)
assert.True(t, math.IsNan(mySqrt(-9.0)))
assert.Equal(t, mySqrt(0), 0)
```

Here, any single case failing has no impact on the future cases, and, seeing all three failures would be useful when trying to address any bugs.

On the other hand, the test you have here, if the cache inappropriately fails to handle insertion of the initial certificates, then the future failures are meaningless and expected.  Why continue executing and producing additional failure output when it provides negative value?

yacovm: > On the other hand, the test you have here, if the cache inappropriately fails to handle insertion of the initial certificates, then the future failures are meaningless and expected. 

I see what you mean in general, but we don't know that from the test. The only thing we know is that it returned an unexpected outcome. It may cache the value correctly despite returning the wrong result.

> Why continue executing and producing additional failure output when it provides negative value?

You're right in theory but I'm not sure it matters that much in practice. We've been using assert all over the project and now we moved almost all of them to require and I haven't noticed any difference, but maybe that's just me.

jyellick (approved):

PR #1784, opened 2020-08-20
Cleanup bccsp/pkcs11 tests

sykesm:


mastersingh24 (approved):
much cleaner.  we might want to look at doing something similar for SW at some point

PR #1791, opened 2020-08-21
[FAB-18117] Add policy checking to snapshot grpc service

wenjianqiao:
Signed-off-by: Wenjian Qiao wenjianq@gmail.com

#### Type of change
- New feature

#### Description
- Add ACLProvider to snapshot service to check ACL
- Add resources for snapshot submitrequest, cancelrequest and listpending
- Add CheckPolicyNoChannel to ACLProvider interface and implement CheckPolicyNoChannelBySignedData func in policy.go

#### Additional details
This PR is stacked on top of https://github.com/hyperledger/fabric/pull/1790

#### Related issues
https://jira.hyperledger.org/browse/FAB-18117

### DIFF
--- a/core/aclmgmt/resourceprovider.go
+++ b/core/aclmgmt/resourceprovider.go
@@ -181,3 +181,12 @@ func (rp *resourceProvider) CheckACL(resName string, channelID string, idinfo in
 
 	return rp.defaultProvider.CheckACL(resName, channelID, idinfo)
 }
+
+// CheckACLNoChannel implements the ACLProvider interface function
+func (rp *resourceProvider) CheckACLNoChannel(resName string, idinfo interface{}) error {
+	if !rp.enforceDefaultBehavior(resName, "", idinfo) {
+		return fmt.Errorf("cannot override peer type policy for channeless ACL check")
+	}
+
+	return rp.defaultProvider.CheckACLNoChannel(resName, idinfo)
+}
--- a/core/ledger/snapshotgrpc/snapshot_service.go
+++ b/core/ledger/snapshotgrpc/snapshot_service.go
@@ -8,17 +8,23 @@ package snapshotgrpc
 
 import (
 	"context"
+	"time"
 
 	"github.com/golang/protobuf/proto"
 	"github.com/golang/protobuf/ptypes/empty"
+	cb "github.com/hyperledger/fabric-protos-go/common"
 	pb "github.com/hyperledger/fabric-protos-go/peer"
+	"github.com/hyperledger/fabric/common/crypto"
+	"github.com/hyperledger/fabric/core/aclmgmt/resources"
 	"github.com/hyperledger/fabric/core/ledger"
+	"github.com/hyperledger/fabric/protoutil"
 	"github.com/pkg/errors"
 )
 
 // Snapshot Service implements SnapshotServer grpc interface
 type SnapshotService struct {
 	LedgerGetter LedgerGetter
+	ACLProvider  ACLProvider
 }
 
 // LedgerGetter gets the PeerLedger associated with a channel.
@@ -26,11 +32,20 @@ type LedgerGetter interface {
 	GetLedger(cid string) ledger.PeerLedger
 }
 
+// ACLProvider checks ACL for a channelless resource
+type ACLProvider interface {
+	CheckACLNoChannel(resName string, idinfo interface{}) error
+}
+
 // Generate generates a snapshot request.
 func (s *SnapshotService) Generate(ctx context.Context, signedRequest *pb.SignedSnapshotRequest) (*empty.Empty, error) {
 	request := &pb.SnapshotRequest{}
 	if err := proto.Unmarshal(signedRequest.Request, request); err != nil {
 		return nil, errors.Wrap(err, "failed to unmarshal snapshot request")
+	}
+
+	if err := s.checkACL(resources.Snapshot_submitrequest, request.SignatureHeader, signedRequest); err != nil {
+		return nil, err
 	}
 
 	lgr, err := s.getLedger(request.ChannelId)
@@ -52,6 +67,10 @@ func (s *SnapshotService) Cancel(ctx context.Context, signedRequest *pb.SignedSn
 		return nil, errors.Wrap(err, "failed to unmarshal snapshot request")
 	}
 
+	if err := s.checkACL(resources.Snapshot_cancelrequest, request.SignatureHeader, signedRequest); err != nil {
+		return nil, err
+	}
+
 	lgr, err := s.getLedger(request.ChannelId)
 	if err != nil {
 		return nil, err
@@ -71,6 +90,10 @@ func (s *SnapshotService) QueryPendings(ctx context.Context, signedRequest *pb.S
 		return nil, errors.Wrap(err, "failed to unmarshal snapshot request")
 	}
 
+	if err := s.checkACL(resources.Snapshot_listpending, query.SignatureHeader, signedRequest); err != nil {
+		return nil, err
+	}
+
 	lgr, err := s.getLedger(query.ChannelId)
 	if err != nil {
 		return nil, err
@@ -82,6 +105,30 @@ func (s *SnapshotService) QueryPendings(ctx context.Context, signedRequest *pb.S
 	}
 
 	return &pb.QueryPendingSnapshotsResponse{BlockNumbers: result}, nil
+}
+
+func (s *SnapshotService) checkACL(resName string, signatureHdr *cb.SignatureHeader, signedRequest *pb.SignedSnapshotRequest) error {
+	if signatureHdr == nil {
+		return errors.New("missing signature header")
+	}
+
+	expirationTime := crypto.ExpiresAt(signatureHdr.Creator)
+	if !expirationTime.IsZero() && time.Now().After(expirationTime) {
+		return errors.New("client identity expired")
+	}
+
+	if err := s.ACLProvider.CheckACLNoChannel(
+		resName,
+		[]*protoutil.SignedData{{
+			Identity:  signatureHdr.Creator,
+			Data:      signedRequest.Request,
+			Signature: signedRequest.Signature,
+		}},
+	); err != nil {
+		return err
+	}
+
+	return nil
 }
 
 func (s *SnapshotService) getLedger(channelID string) (ledger.PeerLedger, error) {
--- a/core/ledger/snapshotgrpc/snapshot_service_test.go
+++ b/core/ledger/snapshotgrpc/snapshot_service_test.go
@@ -8,9 +8,11 @@ package snapshotgrpc
 
 import (
 	"context"
+	"fmt"
 	"io/ioutil"
 	"testing"
 
+	"github.com/hyperledger/fabric-protos-go/common"
 	pb "github.com/hyperledger/fabric-protos-go/peer"
 	"github.com/hyperledger/fabric/common/configtx/test"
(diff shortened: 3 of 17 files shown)


core/policy/policy.go
```diff
@@ -172,3 +176,42 @@ func (p *policyChecker) CheckPolicyBySignedData(channelID, policyName string, sd
 
 	return nil
 }
+
+// CheckPolicyNoChannelBySignedData checks that the passed signed data are valid with the respect to
+// passed policy on the local MSP.
+func (p *policyChecker) CheckPolicyNoChannelBySignedData(policyName string, signedData []*protoutil.SignedData) error {
```
ale-linux, line 182: Does this mean that all the elements in the `signedData` slice must comply with the policy?

core/ledger/snapshotgrpc/snapshot_service.go
```diff
@@ -84,6 +103,33 @@ func (s *SnapshotService) QueryPendings(ctx context.Context, signedRequest *pb.S
 	return &pb.QueryPendingSnapshotsResponse{BlockNumbers: result}, nil
 }
 
+func (s *SnapshotService) checkACL(channelID string, resName string, signatureHdr *cb.SignatureHeader, signedRequest *pb.SignedSnapshotRequest) error {
+	if signatureHdr == nil {
+		return errors.New("missing signature header")
+	}
+
+	expirationTime := crypto.ExpiresAt(signatureHdr.Creator)
```
ale-linux, line 115: Just to double check: this ACL evaluation is not perfomed on the commit path right? Otherwise a time check like this would get us to fork.

core/policy/policy.go
```diff
@@ -144,7 +148,7 @@ func (p *policyChecker) CheckPolicyNoChannel(policyName string, signedProp *pb.S
 // passed policy on the passed channel.
 func (p *policyChecker) CheckPolicyBySignedData(channelID, policyName string, sd []*protoutil.SignedData) error {
 	if channelID == "" {
-		return errors.New("Invalid channel ID name during check policy on signed data. Name must be different from nil.")
+		return p.CheckPolicyNoChannelBySignedData(policyName, sd)
```
ale-linux: This function ends up being called in a ton of places, and this PR changes its failure model (prior to it, we would fail if the supplied channel ID was empty, now we don't). Have you verified that this change does not affect the verifier/committer path? Adding a new function to the 
```
type ACLProvider interface {
	//CheckACL checks the ACL for the resource for the channel using the
	//idinfo. idinfo is an object such as SignedProposal from which an
	//id can be extracted for testing against a policy
	CheckACL(resName string, channelID string, idinfo interface{}) error
}
```
interface (like a `CheckACLNoChannel` function) seems more desirable than changing the contract of the existing function, I'd go with that.

ale-linux (changes requested):

core/ledger/snapshotgrpc/snapshot_service.go
```diff
@@ -84,6 +103,33 @@ func (s *SnapshotService) QueryPendings(ctx context.Context, signedRequest *pb.S
...
+	if !expirationTime.IsZero() && time.Now().After(expirationTime) {
```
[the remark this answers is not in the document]
wenjianqiao: @ale-linux Thanks for your comment. I confirm that the events for which the ACL is checked for snapshot service are never on orderer events or on commit path.

core/policy/policy.go
wenjianqiao: Yes, that's the intention.

core/ledger/snapshotgrpc/snapshot_service.go
wenjianqiao: Same as above.

core/ledger/snapshotgrpc/snapshot_service.go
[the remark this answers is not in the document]
manish-sethi: I did not notice this expiration check on other APIs that allow admin operations (e.g., join channel). Better not to have a different behavior for snapshot related APIs.

core/policy/policy.go
wenjianqiao: When checking existing callers before making the change. I didn't find any code calling CheckACL with []*protoutil.SignedData. The existing code calls CheckACL with either SignedPropoal or Envelope (no []*protoutil.SignedData). However, you have a good point. It is better to add a CheckACLNoChannel and make the explicit call instead of overloading/changing existing behavior.

core/ledger/snapshotgrpc/snapshot_service.go
```diff
@@ -84,6 +103,33 @@ func (s *SnapshotService) QueryPendings(ctx context.Context, signedRequest *pb.S
...
+		return errors.New("client identity expired")
+	}
+
+	signedData := []*protoutil.SignedData{{
+		Identity:  signatureHdr.Creator,
+		Data:      signedRequest.Request,
+		Signature: signedRequest.Signature,
+	}}
```
manish-sethi: Nit: It's used immediately below, can be used directly in-line

core/ledger/snapshotgrpc/snapshot_service_test.go
```diff
@@ -87,37 +96,49 @@ func TestSnapshot(t *testing.T) {
...
 		},
 		{
-			name:      "cannot find ledger",
-			channelID: ledgerID,
-			errMsg:    "cannot find ledger for channel " + ledgerID,
+			name:          "checkACL error propagation",
+			channelID:     ledgerID,
+			signedRequest: createSignedQuery(ledgerID),
+			checkACLErr:   fmt.Errorf("fake-check-acl-error"),
+			errMsg:        "fake-check-acl-error",
```
manish-sethi: I found this aclprovider error test a bit of a force fit into the existing test case - definition. I feel that having this as independent one makes a better sense.

core/policy/policy.go
manish-sethi: Even the existing behavior of the [policy checker](https://github.com/hyperledger/fabric/blob/545754c4bab3311ca56f6098f08dd43cec7c0cf5/core/policy/policy.go#L38) is inconsistent. Two similar functions [CheckPolicy](https://github.com/hyperledger/fabric/blob/545754c4bab3311ca56f6098f08dd43cec7c0cf5/core/policy/policy.go#L51) and [CheckPolicyBySignedData](https://github.com/hyperledger/fabric/blob/545754c4bab3311ca56f6098f08dd43cec7c0cf5/core/policy/policy.go#L145) behave differently, if no channelID is specified. 

@ale-linux  - I agree that adding the exposing the function `CheckACLNoChannel` at the `ACLProvider` level makes it clearer but I would add a note (I assume that you meant this implicitly as well) that we should modify the existing callers go via this function instead of first function above. Otherwise, it will simply add to the confusion looking at the code for - which would look like
- if you work with `SignedProposal` and have no channel call the function `CheckACL` with channelID as empty string
- if you work with `SignedData`, use the function `CheckACLNoChannel`.
This modification can be done in a separate PR and can open a Jira for this.

core/ledger/snapshotgrpc/snapshot_service.go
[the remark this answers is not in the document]
ale-linux: +1 @manish-sethi, good point

core/policy/policy.go
ale-linux: Yes @manish-sethi, I share your view and the approach you propose

wenjianqiao: @manish-sethi agree, I have thought the same and will open a Jira for it.

core/ledger/snapshotgrpc/snapshot_service.go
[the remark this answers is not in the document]
wenjianqiao: @manish-sethi The other APIs that allow admin ops are done via endorsor ProcessProposal that checks expiration via an auth filter [expiration.go](https://github.com/hyperledger/fabric/blob/master/core/handlers/auth/filter/expiration.go#L49) Therefore, individual cmd doesn't need separate call to check certification expiration. However, because snapshot service is a separate grpc service, we should add this check.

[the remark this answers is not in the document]
wenjianqiao: @manish-sethi The other APIs that allow admin ops are done via endorsor ProcessProposal that checks expiration via an auth filter expiration.go Therefore, individual cmd doesn't need separate call to check certificate expiration. However, because snapshot service is a separate grpc service, we should add this check.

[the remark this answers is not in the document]
manish-sethi: I see, I missed that part here. Thanks for the clarification @wenjianqiao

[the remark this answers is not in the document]
manish-sethi: Just to take a note, ideally, that should also have been performed at one place like ACLCheck but not related to this PR.

wenjianqiao:
@ale-linux @manish-sethi I have addressed your comments. Also opened https://jira.hyperledger.org/browse/FAB-18185 to update the existing code to call CheckACLNoChannel when there is no channel.

ale-linux (commented):
LGTM. I won't merge yet to give @manish-sethi a chance to review as well.

ale-linux (dismissed):

core/policy/policy.go
manish-sethi: missed rolling back this change?

manish-sethi: missed rolling back this change?

wenjianqiao: Good catch. Done.

core/aclmgmt/resourceprovider.go
```diff
@@ -138,6 +141,11 @@ func (rp *aclmgmtPolicyProviderImpl) CheckACL(polName string, idinfo interface{}
 	return nil
 }
 
+// This function should not be called because CheckACLNoChannel does not support overriding default provider
+func (rp *aclmgmtPolicyProviderImpl) CheckACLNoChannel(resName string, idinfo interface{}) error {
+	return fmt.Errorf("overriding default acl provider is not supported")
+}
+
```
manish-sethi: This really looks bad... an implementation knows the behavior of the consumer (which is usually the other way around). I am not well versed with this package but is it not possible for this impl to implement a separate interface which does not have this function - or rather does it have to implement this interface at all? I see this getting used only at one place in the production code where it does not need to implement this interface.

wenjianqiao: Removed because peer resource does not allow override and never gets here.

core/policy/policy_test.go
```diff
@@ -162,31 +162,72 @@ func TestPolicyChecker(t *testing.T) {
...
+		require.NoError(t, err)
+
+		sProp, _ = protoutil.MockSignedEndorserProposalOrPanic("A", &peer.ChaincodeSpec{}, []byte("Bob"), []byte("msg2"))
+		// Bob is not a member of the local MSP, policy check must fail
+		err = pc.CheckPolicyNoChannel(mgmt.Members, sProp)
+		require.Error(t, err)
+		require.Contains(t, err.Error(), "Failed deserializing proposal creator during channelless check policy with policy [Members]: [Invalid Identity]")
+	})
+
+	t.Run("CheckPolicyNoChannel", func(t *testing.T) {
```
manish-sethi, line 200: Nit: To differentiate, better to use different names, perhaps `CheckPolicyNoChannelSignedProposal` and `CheckPolicyNoChannelSignedData`. But this should not hold this PR from merging.

manish-sethi (approved):

PR #1792, opened 2020-08-21
Use blocknumber instead of height for snapshot related code

manish-sethi:

#### Type of change

#### Additional details
We had planned for using the height as a number to use for the snapshots. But during the discussions for the end user commands (CLI), we decided to expose the block number instead of height. It's easy to maintain the code and correlate the logs by the end users if we use the block number all the way through the code.


cendhu (commented):
Thanks, Manish. A few minor nits.

cendhu (approved):
Thanks, Manish. LGTM. Other comments are not related to this PR.

PR #1800, opened 2020-08-25
[FAB-18171] Disregard certificate validity period in intra-orderer communication

yacovm:
This change set makes the orderer cluster authentication infrastructure
disregard validity periods when comparing certificates, and only regard public keys.

With this change, one can replace the TLS certificate of Raft,
by a certificate that has the same public key without issuing channel config updates.

This is to ensure that if a certificate of a consenter expired,
one can quickly renew it and start the orderer without performing
a config update per channel.

Change-Id: Id1b29e220d37aa33617b2143b702075bf5b01f6f

### DIFF
--- a/common/crypto/expiration.go
+++ b/common/crypto/expiration.go
@@ -7,12 +7,14 @@ SPDX-License-Identifier: Apache-2.0
 package crypto
 
 import (
+	"bytes"
 	"crypto/x509"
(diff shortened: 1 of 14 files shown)


denyeart (approved):
Deep review done on master...

denyeart:

PR #1804, opened 2020-08-25
[FAB-18188] Log orderer and peer cert expiration date upon startup

denyeart:
#### Type of change

- Improvement (improvement to logging)

#### Description

FAB-17000 added a warning if expiration is within a week.

This change additionally info logs the future enrollment, tls server, and tls client
cert expiration dates upon startup, regardless of whether they will expire within a week or not.

### DIFF
--- a/common/crypto/expiration.go
+++ b/common/crypto/expiration.go
@@ -41,35 +41,35 @@ func certExpirationTime(pemBytes []byte) time.Time {
 	return cert.NotAfter
 }
 
-// WarnFunc notifies a warning happened with the given format, and can be replaced with Warnf of a logger.
-type WarnFunc func(format string, args ...interface{})
+// MessageFunc notifies a message happened with the given format, and can be replaced with Warnf or Infof of a logger.
+type MessageFunc func(format string, args ...interface{})
 
 // Scheduler invokes f after d time, and can be replaced with time.AfterFunc.
 type Scheduler func(d time.Duration, f func()) *time.Timer
 
 // TrackExpiration warns a week before one of the certificates expires
-func TrackExpiration(tls bool, serverCert []byte, clientCertChain [][]byte, sIDBytes []byte, warn WarnFunc, now time.Time, s Scheduler) {
+func TrackExpiration(tls bool, serverCert []byte, clientCertChain [][]byte, sIDBytes []byte, info MessageFunc, warn MessageFunc, now time.Time, s Scheduler) {
 	sID := &msp.SerializedIdentity{}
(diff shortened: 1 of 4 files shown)


yacovm (changes requested):
I don't understand why modify the current code and make it even more complex.

Why not just export [certExpirationTime](https://github.com/hyperledger/fabric/blob/f54d3cda30c731b55981d7029b819e22fce4bf06/common/crypto/expiration.go#L31) and log it [outside](https://github.com/hyperledger/fabric/blob/01d28e583c2588dac1961d38750a4ec491725e6a/orderer/common/server/main.go#L199)?

denyeart:
@yacovm Because each of the three certs have checks against them in TrackExpiration, as does certExpirationTime(). Why duplicate all those checks and add complexity to the peer and orderer mains, when it can be done together with the existing logging in the dedicated expiration tracking function where people would expect it. At the end of the day, I don't think it matters much either way, but mostly I'd prefer to keep the mains nice and clean.

yacovm:
> @yacovm Because each of the three certs have checks against them in TrackExpiration, as does certExpirationTime(). Why duplicate all those checks and add complexity to the peer and orderer mains, when it can be done together with the existing logging in the dedicated expiration tracking function where people would expect it. At the end of the day, I don't think it matters much either way, but mostly I'd prefer to keep the mains nice and clean.

I think that if you make a single function that accepts all 3 as inputs (similar to track expiration) then it won't pollute the main, and will be much easier to understand the code.

denyeart:
> > @yacovm Because each of the three certs have checks against them in TrackExpiration, as does certExpirationTime(). Why duplicate all those checks and add complexity to the peer and orderer mains, when it can be done together with the existing logging in the dedicated expiration tracking function where people would expect it. At the end of the day, I don't think it matters much either way, but mostly I'd prefer to keep the mains nice and clean.
> 
> I think that if you make a single function that accepts all 3 as inputs (similar to track expiration) then it won't pollute the main, and will be much easier to understand the code.

@yacovm  But all three cert expiration dates are not available for logging, it depends on which are configured on the node. By the time we duplicate that logic, might as well just log it alongside the expiration Warning where I have it. I'm not seeing the complexity of keeping the expiration warning and expiration logging co-located in trackCertExpiration(), on the contrary it seems that a trackCertExpiration() function is the most straightforward place to both log the date and provide a warning if the date is coming soon.

yacovm (approved):

denyeart:

denyeart:

denyeart:

PR #1805, opened 2020-08-26
Use proto message for statedb snapshot export

manish-sethi:

#### Type of change

#### Description
In the current code, the internal serialized binary data for the statedbs is exported to snapshot files. Though, this brings performance efficiency, it compromises the code readability. This also causes the binary data in snapshot files to differs for leveldb and couchdb significantly. Moreover, while importing the snapshot data during creation of a channel from snapshot, some of the data needs to be interpreted outside the statedb packages. This commit uses a common proto message for exporting the statedb data for both leveldb and couchdb.


core/ledger/kvledger/txmgmt/privacyenabledstate/snapshot.go
```diff
@@ -483,5 +502,5 @@ func (c *cursor) currentNamespace() string {
 }
 
 type SnapshotPvtdataHashesConsumer interface {
-	ConsumeSnapshotData(namespace, coll string, keyHash []byte, version *version.Height) error
+	ConsumeSnapshotData(namespace, coll string, keyHash []byte, valueHash []byte, version *version.Height) error
```
cendhu, line 505: nit. 

Rather than passing 5 parameters, we could define a `PvtStateHash` struct in the `ledger_interface.g`o and use it here, `purgeMgr`, and `pvtdatastore` packages.

core/ledger/kvledger/txmgmt/privacyenabledstate/snapshot_record.proto
```diff
@@ -0,0 +1,18 @@
...
+
+option go_package = "github.com/hyperledger/fabric/core/ledger/kvledger/txmgmt/statedb/privacyenabledstate";
+
+package privacyenabledstate;
+
+message SnapshotRecord {
+    bytes key = 1;    
+    bytes value = 2;
+    bytes metadata = 3;
+    bytes version = 4;
```
cendhu, line 17: nit. 

Keeping all bytes look nice but we could use `version` as-is. Currently, we convert version to bytes during the export and then convert it back to version during the import. By using version as-is, we could reduce 8 lines of code (I don't think performance will improve) but the code would look a bit leaner. 

If we decide to use `EncodeBytes()` instead of proto, we can keep this as bytes.

core/ledger/kvledger/txmgmt/privacyenabledstate/snapshot_test.go
```diff
@@ -463,7 +462,7 @@ func TestSnapshotImportErrorPropagation(t *testing.T) {
 
 	// errors related to data files
 	for _, f := range []string{pubStateDataFileName, pvtStateHashesFileName} {
-		t.Run("error while checking the presence of "+f, func(t *testing.T) {
+		t.Run("error_while_checking_the_presence_of_"+f, func(t *testing.T) {
```
cendhu, line 465: nit. 

`_` is not needed as the `t.Run` adds it by default.

cendhu (commented):
Thanks, Manish. Looks nice and neat. 

Just a high-level comment/doubt (before merging this). 

As we don't use unknown fields and maps in the proto (also there may not be a reason to extend the proto schema), it is safe to assume the determinism of the snapshot for now. 

1. https://gist.github.com/kchristidis/39c8b310fd9da43d515c4394c3cd9510
2. https://developers.google.com/protocol-buffers/docs/encoding 

Given that our proto schema has bytes field only, we can also have a simple encoding/decoding scheme (we can, in fact, use the one present in the snapshot file writer, i.e., `EncodeBytes()`). 

Wdyt?

core/ledger/kvledger/txmgmt/privacyenabledstate/snapshot.go
manish-sethi: I too felt that and considered alternatives before choosing this one. Specifically, about your suggestion, we have kept `ledger_interface.go` to define types that are exposed to other components not for the cross components within ledger. For components within ledger, it's better to let them have the explicit dependency if they need. I do not find this worth either breaking the pattern for `ledger_interface.go` or introducing a dependency.

core/ledger/kvledger/txmgmt/privacyenabledstate/snapshot_record.proto
cendhu: Yeah, I missed the first part about being a proto message :-) 

Okay. Agree.

core/ledger/kvledger/txmgmt/privacyenabledstate/snapshot.go
cendhu: I remember having this discussion 2 years ago. We decided to have a place within ledger to share common types (within ledger). Finally, we ended up using the `ledger_interface.go` (i think, something related to pvtdata) deferring the former approach.

core/ledger/kvledger/txmgmt/privacyenabledstate/snapshot_test.go
cendhu: True :-)

cendhu (approved):

PR #1422, opened 2020-06-19
[FAB-17539] Always remember anchor peers in membership

wenjianqiao:
Signed-off-by: Wenjian Qiao <wenjianq@gmail.com>

#### Type of change
- Bug fix

#### Description
Gossip service removes a peer from its membership if the peer's alive message is expired.
However, it should not remove the anchor peers or bootstrap peers in order for
the peer to reconnect. Gossip already remembers bootstrap peers. This PR adds code to track
all anchor peers' endpoints and updates the expiration callback function to not delete
anchor peers.

#### Related issues
https://jira.hyperledger.org/browse/FAB-17539

PR #1815, opened 2020-08-26
[FAB-17539] Always remember anchor peers in membership (bp #1422)

wenjianqiao:

This is a backport of pull request #1422 via cherry pick and manual updates due to merge conflicts and code differences between v1.4 and v2.0.

#### Type of change
- Bug fix

#### Description
Gossip service removes a peer from its membership if the peer's alive message is expired.
However, it should not remove the anchor peers or bootstrap peers in order for
the peer to reconnect. Gossip already remembers bootstrap peers. This PR adds code to track
all anchor peers' endpoints and updates the expiration callback function to not delete
anchor peers.

#### Related issues
https://jira.hyperledger.org/browse/FAB-17539

### DIFF
--- a/gossip/discovery/discovery.go
+++ b/gossip/discovery/discovery.go
@@ -71,6 +71,11 @@ type CommService interface {
 	IdentitySwitch() <-chan common.PKIidType
 }
 
+// AnchorPeerTracker is an interface that is passed to discovery to check if an endpoint is an anchor peer
+type AnchorPeerTracker interface {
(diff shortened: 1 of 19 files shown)


integration/ports.go
```diff
@@ -0,0 +1,54 @@
+/*
```
yacovm: can't we backport it without dragging this file along?

wenjianqiao: Thanks Yacov. The file is removed.

yacovm (approved):

PR #1816, opened 2020-08-26
FAB-18187 back port go mod v14

petermetz:
Signed-off-by: Matthew Sykes <sykesmat@us.ibm.com>
(cherry picked from commit a6c1ff7b1d4f7ee834775041b924f0f18c69caf0)
Signed-off-by: Peter Somogyvari <peter.metz@unarin.com>

#### Type of change

#### Description

Makes it possible to author chain code targeted at 1.4.x ledgers while using go mods.

#### Additional details

I just took the commit that was on the 2.x branch and applied it on the 1.4 release branch.
Then I tested it locally by pulling in fabric via go mods while writing a chain code as per the 1.4x examples, it compiled and also was able to launch so I figured this was all that's needed.

I'm not a go pro, so happy to be corrected here: The assumption is that this is backwards compatible because adding go mod support does not prevent others from still depending on the code the old way via GOPATH and the like. Of course if this turns out to be not true I'm happy to update the PR to make sure that the change is indeed backward compatible.

#### Related issues

Relevant JIRA ticket: https://jira.hyperledger.org/browse/FAB-18187


lindluni:
This commit would only enable using Go Mod to version the shim and protos in your Chaincode, which would still need to be vendored anyway, as we don't provide support for module based chaincode in 1.4. Essentially you picked the wrong commit to backport, you wanted one from sometime last year which has a much smaller surface area and enhances the chaincode launch process to use Go Modules, not Fabric itself.

That being said, this isn't something we would target for backporting. We don't backport features, only bug fixes and security fixes, especially into an LTS branch where we guarantee backwards compatibily and stability. There is too much danger in backporting unnecessary changes that would be applied against running production networks. This isn't something we would consider for backport even with the right commits chosen.

While we understand the desire for it, it's just out of place for backport, especially with approximately 6 months until 1.4 is end of life.

Thank you for the contribution though.

lindluni:
I will also add, you can use modules in 1.4, you just have to end the process by vendoring, so it looks like this

```
export GO111MODULE=on
go get github.com/hyperledger/fabric@f208ffeb4dfda3f520ae5e481ebe0b58230addab #This will pin on the latest 1.4 commit
go mod tidy
go mod vendor
```

This will use modules to version and manage your dependencies, but provide them in the traditional vendor folder. The only difference between this and 2.0 Fabric is in 2.0 you don't have to explicitly have the vendor directory, as we will use modules to pull that in. So you get all the benefits of modules, except for the disk overhead of the vendor directory on your local machine.

PR #1822, opened 2020-08-27
Enable Custom System Chaincode via go build constraints

g2flyer:
#### Type of change

- New (build) feature

#### Description

As part of part of the effort to remove go plugins ( [FAB-15338](https://jira.hyperledger.org/browse/FAB-15338)), the support for custom system chaincodes to be loaded via go plugins was remvoed in [FAB-17564](https://jira.hyperledger.org/browse/FAB-17564) . The implication of this change is that currently any project which require custom system chaincode, e.g., [Fabric Private Chaincode](https://github.com/hyperledger-labs/fabric-private-chaincode/), has to patch existing fabric files.

This PR provides a a small extension, dubbed `opt_scc`, which allows to add such custom system chaincode by just adding new files, without touching existing ones, and then allows the selective enablement of said custom chaincodes in builds via build constraints/build tags.

#### Additional details

There are two commits in the PR:
- The core framework and unit tests
- A mock custom SCC to illustrate use of the opt_scc extension point

From a design perspective, the parameters currently passed to the factory functions are the common parameters across the current SCCs started in `internal/peer/node/start.go` and what we needed for FPC.
There seemed to be much less commonality among the remaining parameters, so i left it at the current two (but additional parameters could obviously be easily added.)

BTW: the framework would also easily allow using `opt_scc`  to launch `qscc`.  This is probably not desired but just in case i'm wrong, i would have the corresponding (small) patch at hand ...

### DIFF
--- a/core/scc/opt_scc/mock.go
+++ b/core/scc/opt_scc/mock.go
@@ -1,0 +1,28 @@
+//+build mock_scc
+
+/*
+Copyright IBM Corp. All Rights Reserved.
+Copyright 2020 Intel Corporation
+
+SPDX-License-Identifier: Apache-2.0
+*/
+
+package opt_scc
+
+import (
+	"github.com/hyperledger/fabric/common/flogging"
+	"github.com/hyperledger/fabric/core/aclmgmt"
+	"github.com/hyperledger/fabric/core/peer"
+	"github.com/hyperledger/fabric/core/scc"
+	"github.com/hyperledger/fabric/core/scc/opt_scc/mock"
+)
+
+var mscclogger = flogging.MustGetLogger("mscc")
+
+func init() {
+	mscclogger.Debug("Registring mock as system chaincode")
+	AddFactoryFunc(func(aclProvider aclmgmt.ACLProvider, p *peer.Peer) scc.SelfDescribingSysCC {
+		mscclogger.Debug("Enabling mock as system chaincode")
+		return mock.New(aclProvider, p)
+	})
+}
--- a/core/scc/opt_scc/mock/echo.go
+++ b/core/scc/opt_scc/mock/echo.go
@@ -1,0 +1,46 @@
+/*
+Copyright IBM Corp. All Rights Reserved.
+
+SPDX-License-Identifier: Apache-2.0
+*/
+
+package mock
+
+import (
+	"fmt"
+
+	"github.com/hyperledger/fabric-chaincode-go/shim"
+	pb "github.com/hyperledger/fabric-protos-go/peer"
+	"github.com/hyperledger/fabric/common/flogging"
+	"github.com/hyperledger/fabric/core/aclmgmt"
+	"github.com/hyperledger/fabric/core/peer"
+)
+
+func New(aclProvider aclmgmt.ACLProvider, p *peer.Peer) *EchoSCC {
+	return &EchoSCC{}
+}
+
+// EchoSCC echos any (string) request back as response
+type EchoSCC struct {
+}
+
+func (e *EchoSCC) Name() string              { return "mscc" }
+func (e *EchoSCC) Chaincode() shim.Chaincode { return e }
+
+var mscclogger = flogging.MustGetLogger("mscc")
+
+func (e *EchoSCC) Init(stub shim.ChaincodeStubInterface) pb.Response {
+	mscclogger.Info("Init MSCC")
+
+	return shim.Success(nil)
+}
+
+// Invoke is called with args[0] contains the query function name, args[1]
+func (e *EchoSCC) Invoke(stub shim.ChaincodeStubInterface) pb.Response {
+	args := stub.GetStringArgs()
+
+	mscclogger.Infof("Invoke MSCC(%v)", args)
+
+	bytes := []byte(fmt.Sprintf("%v", args))
+	return shim.Success(bytes)
+}
--- a/core/scc/opt_scc/registry.go
+++ b/core/scc/opt_scc/registry.go
@@ -1,0 +1,34 @@
+/*
+ Copyright 2020 Intel Corporation
+
+SPDX-License-Identifier: Apache-2.0
+*/
+
+package opt_scc
+
+import (
+	"github.com/hyperledger/fabric/common/flogging"
+	"github.com/hyperledger/fabric/core/aclmgmt"
+	"github.com/hyperledger/fabric/core/peer"
+	"github.com/hyperledger/fabric/core/scc"
+)
+
+var logger = flogging.MustGetLogger("opt_scc")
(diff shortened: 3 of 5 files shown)


sykesm (commented):
Thanks for the PR. I'd like to start with a couple of comments and then we can dig into the details as needed.

- I don't see where the `mock_scc` build tag is used in any of the tests or build process so I don't understand its intent within the context of this build tree or pull request.
- The function/callback pattern for registration is too fragile. It provides no mechanism to add additional references to extensions other than the two that are needed today. This can be resolved by providing an object to the chaincode between registration and activation that can be used to pull additional details. Ideally this is a value struct that we can add new fields to as we need to evolve.
- As the system chaincode already needs to satisfy the `SelfDescribingSysCC` interface, instead of using something like the "factory function", we should define an optional interface to be implemented that can receive the previously mentioned structure. A type assertion and check can be used to determine of the chaincode implements the behavior.
- The concept of "optional" (?) system chaincode doesn't warrant a new package. The `scc` package can be quickly extended to maintain the slice and lock required to manage external registration.
- As our existing, statically constructed system chaincode instances adhere to the `SelfDescribingSysCC` interface, we can manually register those from `peer/node/start.go` and simply reference the slice of registered system chaincodes in the deployment iteration.
- The block comment over `scc.BuiltinSCCs` is a statement of direction indicating that we expect all chaincode to be initialized in `_lifecycle` yet this PR adds all "optional" (?) chaincode to this list. I'd like to understand why that is.

At the end of the day, I think what's done in this PR could be scoped to:
- A new structure to surface dependencies for system chaincode. This should not be an interface.
- A new, optional interface that indicates a system chaincode expects an initialization callback to provide dependencies prior to deployment. If system chaincode implements the optional interface, the initialization callback will be driven with the dependencies structure.
- Functions to register and retrieve system chaincode during peer initialization.
- Simple tests to ensure that the registration, initialization callback, and deployment function correctly.

mastersingh24:
> At the end of the day, I think what's done in this PR could be scoped to:
> 
> * A new structure to surface dependencies for system chaincode. This should not be an interface.
> * A new, optional interface that indicates a system chaincode expects an initialization callback to provide dependencies prior to deployment. If system chaincode implements the optional interface, the initialization callback will be driven with the dependencies structure.
> * Functions to register and retrieve system chaincode during peer initialization.
> * Simple tests to ensure that the registration, initialization callback, and deployment function correctly.

This makes the most sense to me as well.

g2flyer:
Matt, thanks for the valuable feedback and suggestions for improvement. I agree with Gari they make perfectly sense, e.g., passing the struct rather than an explicit list definitely helps in making it much easier to evolve without braking backwards compatibility.   

To summarize, in the next iteration i will change as follows:
- `register` will take directly a  `SelfDescribingSysCC` object (instead of the factory function) as a parameter and all, even existing SCCs will be registered
- as registered SCCs are deployed, an `initialize` function will be called iff the SCC implements a corresponding (new and optional) interface. This allows the SCC to (optionally) finalize the initialization based on a provided structure which contains (a) the fields for the peer context objects such as the aclManager and (b) a reference to the builtinSCC object (i.e., SCCs are not blindly added to builtinSCC but only if a particular SCC explicitly opts-in)
- all of that will be directly integrated into package `scc` (making the separate package `opt_scc` obsolete).
- the unit test programs will be correspondingly refactored. I also will drop mock tests: The original purpose of it was to service as an illustration how it would work with build tag and to use  (manual) integration tests, again with/with-out build-tags. As discussed offline, the build-tag aspect is obvious and doesn't need extra tests. The unit tests plus the existing integration tests should hopefully also should cover in the new version sufficiently the end-to-end aspects given the tighter integration into scc.

I will be away the next two weeks (trying to escape the wild fires and smoke ...) but hope to have a next iteration by early october.

kopaygorodsky:
@g2flyer hey, I would like to help with that if you are busy. 
really need this feature :)

kopaygorodsky:
@g2flyer hey, I would like to help with this OR if you are busy.
really need this feature :)

kopaygorodsky:
I encountered an issue using a system chaincode with 2.0 capabilities. You might want to look at this comment: https://jira.hyperledger.org/browse/FAB-17564?focusedCommentId=70597&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-70597

lindluni:
@g2flyer hope all is well.

I wanted to check in and see if you were going to come back to this soon. Trying to clean up what we can before everyone heads out for the end of the year so we have the cleanest slate possible. Is this something you think you might get to soon?

If not can we close the PR, you can always resubmit it later.

Thank you again!

g2flyer:
Hi Brett,

> @g2flyer hope all is well.

As far as these crazy times go, i can't complain. Alas, FPC and other things have kept me more busy than expected which delayed progress on this PR.

> I wanted to check in and see if you were going to come back to this soon. Trying to clean up what we can before everyone heads out for the end of the year so we have the cleanest slate possible. Is this something you think you might get to soon?

I still hope i can resume work on it before Xmas and provide an update on it.   Otherwise, ..

> If not can we close the PR, you can always resubmit it later.

... unfortunately i guess we have to go that route :-(

caod123:
Hello @g2flyer and Happy New Year! Are there any updates or plans for this PR? If not, it probably makes the most sense to close this out for the time being.

g2flyer:
Hi Danny

> Hello @g2flyer and Happy New Year! 

Thanks! Also a good '21 to you

> Are there any updates or plans for this PR? 

Unfortunately, my hope that i might do it before Xmass turned out to all for naught given higher priority work for Fabric Private Chaincode (FPC) i had to complete.  Given the change of plans for FPC (see [Fabric RFC PR #40](https://github.com/hyperledger/fabric-rfcs/pull/40)), this item is not on the critical path anymore (actually, as of now is not on the plan at all). So ...

> If not, it probably makes the most sense to close this out for the time being.

... unless there is strong interest from other parties for such a feature, i would just close it for the time being.  From a general Fabric perspective, such an extension point should probably also be considered in a broader view covering also validation and endorsement plugins (e.g., [FAB-15338](https://jira.hyperledger.org/browse/FAB-15338?jql=text%20~%20%22plugin%22))?

caod123:
Let's close this for now then and reopen as needed.

kopaygorodsky:
I would like to continue on this PR, I just need an answer from anyone who understands VCSS system. 
https://jira.hyperledger.org/browse/FAB-17564?focusedCommentId=70597&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-70597
Do I need to write my own vcss plugin for that?

PR #1824, opened 2020-08-28
[FAB-7482] Add CRL support for communication infrastructure

yacovm:
This commit adds an implementation of RevokedCertificates.

The RevokedCertificates is initialized with pairs of issuers and CRL issued by the corresponding issuer,
and then it can be queried whether a given x509 certificate has been revoked or not.

Should this commit be merged, in next commits we would be able to move the ComparisonMemoizer
from orderer/common/cluster to common/ and then use it to cache CRL queries (both positive and negative outcomes)
to speedup the CRL lookup at time of connection establishment.

Then, at later commits, we could actually plug this infrastructure to the orderer/peer themselves,
given there would be no objections.

Change-Id: Icccbd5fbd57b1cb7d04ee6702568b7a260ab7a1f


mastersingh24:
@yacovm  - Why exactly do we now care about CRLs?  Raft of course uses client TLS at this point, but we have other checks to ensure that a node is actually in the consenter set.

We don't really promote the use of mutual TLS in general (I wish I had never implemented it) and should be moving to remove it's use wherever possible IMHO.

yacovm:
> @yacovm - Why exactly do we now care about CRLs? Raft of course uses client TLS at this point, but we have other checks to ensure that a node is actually in the consenter set.
> 
> We don't really promote the use of mutual TLS in general (I wish I had never implemented it) and should be moving to remove it's use wherever possible IMHO.

You should ask the past Gari who opened the [JIRA](https://jira.hyperledger.org/browse/FAB-7482).

I was under the impression that there was a need among potential Fabric users to be able to block connections from specific clients using CRLs so I implemented it.

mastersingh24:
> > @yacovm - Why exactly do we now care about CRLs? Raft of course uses client TLS at this point, but we have other checks to ensure that a node is actually in the consenter set.
> > We don't really promote the use of mutual TLS in general (I wish I had never implemented it) and should be moving to remove it's use wherever possible IMHO.
> 
> You should ask the past Gari who opened the [JIRA](https://jira.hyperledger.org/browse/FAB-7482).
> 
> I was under the impression that there was a need among potential Fabric users to be able to block connections from specific clients using CRLs so I implemented it.

That guy knew less then than he does now.  :-)

mastersingh24:
I'm ok with continuing down this path ... so re-opening for now

mastersingh24:
would be good to understand the full path of how this will work.

yacovm:
> would be good to understand the full path of how this will work.

From the commit message:

> 
> Should this commit be merged, in next commits we would be able to move the ComparisonMemoizer
> from orderer/common/cluster to common/ and then use it to cache CRL queries (both positive and negative outcomes)
> to speedup the CRL lookup at time of connection establishment.
> 
> Then, at later commits, we could actually plug this infrastructure to the orderer/peer themselves,
> given there would be no objections.

yacovm:
> > > @yacovm - Why exactly do we now care about CRLs? Raft of course uses client TLS at this point, but we have other checks to ensure that a node is actually in the consenter set.
> > > We don't really promote the use of mutual TLS in general (I wish I had never implemented it) and should be moving to remove it's use wherever possible IMHO.
> > 
> > 
> > You should ask the past Gari who opened the [JIRA](https://jira.hyperledger.org/browse/FAB-7482).
> > I was under the impression that there was a need among potential Fabric users to be able to block connections from specific clients using CRLs so I implemented it.
> 
> That guy knew less then than he does now. :-)

I'm happy to see that you both see eye to eye after all !

mastersingh24:
My question is more around how the CRLs will be distributed (and updated) .... currently the trusted roots for the gRPC server are populated from the tls root CAs in the channel config (as well as from statically configured ones as well).  The current CRLs in the channel config are used solely to check revocation for signing identities.

Given that we don't need deterministic processing for TLS handshakes across all orgs, perhaps we should just add CRLs to the config rather than distributing via channel config?

yacovm:
> My question is more around how the CRLs will be distributed (and updated) .... currently the trusted roots for the gRPC server are populated from the tls root CAs in the channel config (as well as from statically configured ones as well). The current CRLs in the channel config are used solely to check revocation for signing identities.
> 
> Given that we don't need deterministic processing for TLS handshakes across all orgs, perhaps we should just add CRLs to the config rather than distributing via channel config?

I never intended to touch the channel config. 

To me, TLS CRLs is not something that should be uniformly enforced upon all organizations anyway. Furthermore, I see this feature as a niche and very optional one, and I don't foresee many deployments using it, so I don't want to mess with the channel config needlessly. 

What I had in mind, is to specify a folder name in the **orderer.yaml**  / **core.yaml** which will contain a bunch of files of two types: 
- CA certificates
- CRLs signed by the CAs whose certificates are in the folder

Upon startup, the node will populate a mapping from a certificate of a CA to all CRLs found in the folder, and if there is a CRL with a missing CA certificate it will refuse to start the process.

While this leaves much to be desired, such as dynamically updating the CRLs without restarting the node, I think that this is enough for a first stage, and we can watch and see adoption, and in the future if there would be... sufficient pressure from users, we would be able to leverage the admin REST API to also be able to list/add/remove CRLs.

yacovm:
OK seems like 4 weeks have passed without any review, I am closing this due to lack of interest from the community.

mastersingh24:
I was waiting to see commit(s) from actually wiring this into the gRPC servers.  My expectation is that we'd have this single PR with multiple stacked commits.

I'm also not quite sure why we require you to provide both the CA(s) and CRL(s)?

I should have selected request changes.

mastersingh24 (changes requested):
I'd like to see the stack of commits for end to end in this PR

yacovm:
> I'm also not quite sure why we require you to provide both the CA(s) and CRL(s)?

I am failing to parse this sentence, but, we need to provide the CRLs because they contain the information to identify the revoked certificates. 

As for the CAs - I want to include the CAs in the setup, in order to ensure the CRLs were indeed produced (signed!) by the CA and were not fabricated. It's true, that the security model in Fabric implies that if you have local access to the peer, you are its administrator, but I still want to enforce proper procedures, namely have the CA sign the CRLs as in the standard. 

Now, if you agree that I need to validate the CA's signature on the CRLs, then I also need the CA public key. I want to access it at boot time, so if something is mis-configured, we would know about it right away, and troubleshooting will be easier.

yacovm:
I'm not sure if this is needed at this time. 
I'm closing this, and if there will be a need, someone can pick it up from here or from scratch.

PR #1828, opened 2020-09-01
[FAB-18191] Remove contents of leveldb dir instead of the dir itself when dropping dbs

wenjianqiao:

#### Type of change
- Improvement

#### Description
When dropping a leveldb, remove all contents of the directory instead of the directory itself. 

#### Additional details
This is common code for upgrade-dbs, rebuild, reset, and rollback. It does not impact ledger rebuild function. The change is requested because some leveldb directories (e.g., StateLevelDB) may be mounted and therefore cannot be deleted.

#### Related issues
https://jira.hyperledger.org/browse/FAB-18191

### DIFF
--- a/core/ledger/kvledger/tests/env.go
+++ b/core/ledger/kvledger/tests/env.go
@@ -79,7 +79,7 @@ func (e *env) cleanup() {
 	os.RemoveAll(e.initializer.Config.RootFSPath)
 }
 
-func (e *env) closeAllLedgersAndDrop(flags rebuildable) {
+func (e *env) closeAllLedgersAndRemoveDirContents(flags rebuildable) {
 	if e.ledgerMgr != nil {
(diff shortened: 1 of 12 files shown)


core/ledger/kvledger/tests/env.go
```diff
@@ -159,6 +159,24 @@ func (e *env) verifyRebuilableDoesNotExist(flags rebuildable) {
 	}
 }
```
manish-sethi: are these two functions still being used somewhere?

wenjianqiao: Removed. It was used by env.closeAllLedgersAndDrop. Per our slack, changed the func to remove contents instead of deleting the directory itself.

manish-sethi (approved):

lindluni:

PR #1831, opened 2020-09-02
Creating new channel with the same channelID of system channel is not forbidden

Zerppen:
#### fix Creating new channel with the same channelID of system channel is not forbidden

- Bug fix

#### Description
[FAB-18200](https://jira.hyperledger.org/browse/FAB-18200)
case: if I created a system channel and channelID is testnam_e.Then someone wants to create his user channel and channelID is _testname, he can create the new channel of couse,but what can be happen if he wants to install his chaincode?

Solution: when Creating channel should check if the channelID exists.

### DIFF
--- a/core/ledger/ledgermgmt/ledger_mgmt.go
+++ b/core/ledger/ledgermgmt/ledger_mgmt.go
@@ -98,18 +98,26 @@ func NewLedgerMgr(initializer *Initializer) *LedgerMgr {
 func (m *LedgerMgr) CreateLedger(id string, genesisBlock *common.Block) (ledger.PeerLedger, error) {
 	m.lock.Lock()
 	defer m.lock.Unlock()
-	logger.Infof("Creating ledger [%s] with genesis block", id)
-	l, err := m.ledgerProvider.Create(genesisBlock)
-	if err != nil {
-		return nil, err
+	logger.Infof("Checking if ledger [%s] with genesis block had created", id)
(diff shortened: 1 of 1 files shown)


cendhu:
Thank you for your contribution. We already check for an existing `ledgerID`, i.e., channel name, while creating a new one and throw this error https://github.com/hyperledger/fabric/blob/26f0813ee0a9c63bca4da924a0e5dd89c064a4bb/core/ledger/kvledger/kv_ledger_provider.go#L613 

https://github.com/hyperledger/fabric/blob/26f0813ee0a9c63bca4da924a0e5dd89c064a4bb/core/ledger/kvledger/kv_ledger_provider_test.go#L498-L499

Can you elaborate on the issue that you face?

guoger:
i don't think you would be able to create such channel. when you create an application channel `foo`, you are actually broadcasting a config msg to orderer, targeting channel `foo` which does not exist yet. However, when you broadcast config msg to `system-channel`, it would be seen as config update, which would very likely throw authorization errors. At the end of the day, i don't think a **new** channel with system name is actually being created.

it would be nice if you could describe in detail what you did and what you expected. Another way to convey your idea (more straightforward but requires some familiarity with fabric code) is to write or modify an integration test, so we could easily reproduce the issue

cc @Zerppen

Zerppen:
Yeah,that was my mistake.

PR #1846, opened 2020-09-04
[FAB-18194] Fix service discovery for legacy installed chaincodes

denyeart:
Service discovery endorsers query fails with errors:
"failed constructing descriptor for chaincodes"
"required chaincodes are not installed on sufficient peers"

The chaincodes installed with legacy chaincode lifecycle were not getting recognized
after peer restart since the path to installed chaincodes was set after
the legacy chaincode cache initialized.
Setting the installed chaincode path earlier in peer start sequence resolves the issue.

### DIFF
--- a/integration/gossip/gossip_test.go
+++ b/integration/gossip/gossip_test.go
@@ -134,6 +134,14 @@ var _ = Describe("Gossip Test", func() {
 			basePeerForTransactions := peer0Org1
 			nwo.DeployChaincodeLegacy(network, channelName, orderer, chaincode, basePeerForTransactions)
 
+			By("verifying peer0Org1 discovers all the peers and the legacy chaincode before starting the tests")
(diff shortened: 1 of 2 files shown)


guoger (approved):

guoger:
Question: how long do we plan to continue backporting fixes into 2.1? @denyeart

denyeart:
> Question: how long do we plan to continue backporting fixes into 2.1? @denyeart

@guoger 
We are trying to get all users to v1.4 and v2.2 LTS releases so that we only need to backport to those two releases, and eventually just to v2.2.
In the meantime, we are also taking the important fixes back to v2.1 since we know that some deployments went to v2.1 prior to the v2.2 LTS announcement. I expect they will be able to move up to v2.2 within a few months.

PR #1847, opened 2020-09-04
FAB-17915 Ch.Part.API: switch member to follower

tock-ibm:
An orderer that is a cluster member of a channel, detects that it is no longer a member by observing the incoming channel config. When that happens, the etcdraft.Chain shuts down and switches to follower.Chain.

This only works when there is no system-channel.

Change-Id: I39f901b8960ea0b96af059e985b3dad1bdb94b62

#### Type of change
- New feature

#### Related issues

Task: FAB-17915
Story: FAB-15711
Epic: FAB-17715

### DIFF
--- a/orderer/common/multichannel/registrar.go
+++ b/orderer/common/multichannel/registrar.go
@@ -516,6 +516,45 @@ func (r *Registrar) SwitchFollowerToChain(chainName string) {
 	logger.Infof("Created and started channel %s", cs.ChannelID())
 }
 
+// SwitchChainToFollower creates a follower.Chain from the tip of the ledger and removes the consensus.Chain.
+// It is called when an etcdraft.Chain detects it was evicted from the cluster (i.e. removed from the consenters set)
+// and halts, transferring execution to the follower.Chain.
+func (r *Registrar) SwitchChainToFollower(channelName string) {
+	r.lock.Lock()
+	defer r.lock.Unlock()
+
+	if _, chainExists := r.chains[channelName]; !chainExists {
+		logger.Infof("Channel %s consenter was removed", channelName)
+		return
+	}
+
+	if _, followerExists := r.followers[channelName]; followerExists {
+		logger.Panicf("Programming error, both a follower.Chain and a consensus.Chain exist, channel: %s", channelName)
+	}
+
+	rl, err := r.ledgerFactory.GetOrCreate(channelName)
+	if err != nil {
+		logger.Panicf("Failed obtaining ledger factory for %s: %v", channelName, err)
+	}
+
+	configBlock := ConfigBlockOrPanic(rl)
+	ledgerRes, clusterConsenter, err := r.initLedgerResourcesClusterConsenter(configBlock)
+	if err != nil {
+		logger.Panicf("Error initializing ledgerResources & clusterConsenter: %s", err)
+	}
+
+	delete(r.chains, channelName)
+	logger.Debugf("Removed consensus.Chain for channel %s", channelName)
+
+	fChain, _, err := r.createFollower(ledgerRes, clusterConsenter, nil, channelName)
+	if err != nil {
+		logger.Panicf("Failed to create follower.Chain for channel '%s', error: %s", channelName, err)
+	}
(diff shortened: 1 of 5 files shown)


orderer/common/multichannel/registrar.go
```diff
@@ -516,6 +516,45 @@ func (r *Registrar) SwitchFollowerToChain(chainName string) {
...
 
+// SwitchChainToFollower creates a follower.Chain from the tip of the ledger and removes the consensus.Chain.
+// It is called when an etcdraft.Chain detects it was evicted from the cluster (i.e. removed from the consenters set)
+// and halts, transferring execution to the follower.Chain.
+func (r *Registrar) SwitchChainToFollower(channelName string) {
+	r.lock.Lock()
+	defer r.lock.Unlock()
+
+	if _, chainExists := r.chains[channelName]; !chainExists {
+		logger.Infof("Channel %s already removed", channelName)
```
jyellick: Perhaps it would be clearer to say "Channel %d consenter is not active (is a follower, or was removed)" or similar?

orderer/common/multichannel/registrar.go
```diff
@@ -516,6 +516,45 @@ func (r *Registrar) SwitchFollowerToChain(chainName string) {
...
+		return
+	}
+
+	if _, followerExists := r.followers[channelName]; followerExists {
+		logger.Panicf("Programming error, follower already exists: %s", channelName)
```
jyellick: Hopefully this is a non-issue as this will never print, but my initial reading of this was a little confusing.  It's not an error that the follow exists, it's that the follower _and_ the consenter exist.

orderer/common/multichannel/registrar.go
```diff
@@ -516,6 +516,45 @@ func (r *Registrar) SwitchFollowerToChain(chainName string) {
...
+
+	rl, err := r.ledgerFactory.GetOrCreate(channelName)
+	if err != nil {
+		logger.Panicf("Failed obtaining ledger factory for %s: %v", channelName, err)
+	}
+
+	configBlock := ConfigBlockOrPanic(rl)
+	ledgerRes, clusterConsenter, err := r.initLedgerResourcesClusterConsenter(configBlock)
+	if err != nil {
+		logger.Panicf("Error initializing ledgerResources & clusterConsenter: %s", err)
```
jyellick, line 543: The call to `initLedgerResourcesClusterConsenter` will fail if the consenter is not a cluster type, is panic-ing a little heavy handed here?

orderer/common/multichannel/registrar.go
```diff
@@ -516,6 +516,45 @@ func (r *Registrar) SwitchFollowerToChain(chainName string) {
...
+	}
+
+	delete(r.chains, channelName)
+	logger.Debugf("Removed chain for channel %s", channelName)
```
jyellick: I realize we call the parameter `chains`, but I think perhaps 'consenter instance' or something would be much more readable for users. (General comment, this spot alone aside).

jyellick (commented):
Generally looks good, I'm still a little skeptical of the injection of callbacks for interaction with the registrar, rather than a scheme where flow control does not loop, but instead returns and continues, but... you are aware of this.

orderer/common/multichannel/registrar.go
tock-ibm: Done

orderer/common/multichannel/registrar.go
tock-ibm: That is correct, fixed.

orderer/common/multichannel/registrar.go
tock-ibm: I think this is justified, as `SwitchChainToFollower` is called only from `etcdraft.Chain` hence it is only expected to operate when the consenter is cluster-type. So this condition represents a programming error.

orderer/common/multichannel/registrar.go
tock-ibm: Done

jyellick (approved):

PR #1849, opened 2020-09-04
Update test_network.md

robmurgai:
Signed-off-by Rob Murgai smurgai@gmail.com

Made a change to replace "master" branch with "latest" branch

#### Type of change

- Documentation update

#### Description

Made a change to replace "master" branch with "latest" branch in keeping with new Guidelines.

### DIFF
--- a/docs/source/test_network.md
+++ b/docs/source/test_network.md
@@ -16,7 +16,7 @@ configured to connect to other running fabric nodes.
 **Note:** These instructions have been verified to work against the
 latest stable Docker images and the pre-compiled setup utilities within the
 supplied tar file. If you run these commands with images or tools from the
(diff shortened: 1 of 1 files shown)


lindluni:
Unlike Read the Docs, GitHub's strategy is to replace `master` with `main` as noted here: https://github.blog/changelog/2020-08-26-set-the-default-branch-for-newly-created-repositories/

I think we need to decide at large if we want to adopt `main` or some other convention and then replace all the doc

robmurgai:
@btl5037 Noted.  I wanted to fix the DCO check.  But, otherwise, I can close this if its not needed for now.

Rob.

ale-linux (dismissed):

PR #1850, opened 2020-09-07
Format code in chaincode docs

jougene:
Signed-off-by: Eugene Sinitsyn <sinitsyn.e@n-t.io>

#### Type of change

- Documentation update

### DIFF
--- a/docs/source/chaincode4ade.rst
+++ b/docs/source/chaincode4ade.rst
@@ -72,7 +72,7 @@ To keep things simple, let's use the following command:
 .. code:: bash
 
   // atcc is shorthand for asset transfer chaincode
-     mkdir atcc && cd atcc
+  mkdir atcc && cd atcc
 
 Now, let's create the module and the source file that we'll fill in with code:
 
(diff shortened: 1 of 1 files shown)


sykesm (dismissed):
Thanks for the pull request. Please remove the test.go from the commit. We don't store examples in the main fabric tree and we avoid code in the root of the tree.

jougene:
Oh, its just test file, created to check code format.) Will remove it.

C0rWin (approved):

docs/source/chaincode4ade.rst
```diff
@@ -159,102 +159,102 @@ in the ``CreateAsset`` function below.
...
-          Color:          color,
-          Size:           size,
-          Owner:          owner,
-          AppraisedValue: appraisedValue,
-        }
-        assetJSON, err := json.Marshal(asset)
-        if err != nil {
-          return err
-        }
+  func (s *SmartContract) UpdateAsset(ctx contractapi.TransactionContextInterface, id string, color string, size int, owner string, appraisedValue int) error {
```
nikhil550, line 216: The color formatting did not return for this function for some reason. All other functions look good.

jougene: I dont really understand, what you mean?

caod123: There's no color formatting in the `UpdateAsset` function in the built docs: https://hyperledger-fabric--1850.org.readthedocs.build/en/1850/chaincode4ade.html#initializing-the-chaincode (same for `DeleteAsset` and a couple other places). It looks like on master this was the case as well so it's probably not a huge deal but odd why it's not formatting on read-the-docs.

PR #1852, opened 2020-09-08
Reset batch

cendhu:
#### Type of change

- Bug fix
- Improvement (performance, etc)

#### Description

While committing data to leveldb in iteration, it is necessary to reset the batch. Otherwise, we would be appending entries to the batch and rewriting already written entries in the previous iteration. Though it does not affect the correctness, it is better to reset the batch to justify the iteration. 

This fix is already there in master.


cendhu:
Merged as part of #1854

PR #1859, opened 2020-09-08
fix missing err check in the block commit path [1]

cendhu:
#### Type of change

- Bug fix

#### Description

This PR addresses some of the missing err check in the ledger commit path.

#### Additional details

A careful analysis was done to ensure that this PR is not creating a possibility of a fork.


ale-linux (commented):
Thanks @cendhu!
> A careful analysis was done to ensure that this PR is not creating a possibility of a fork.
Are the outcomes of the analysis easy to share so that we could have a look and double check?

cendhu:
Thanks, @ale-linux, for the double-check. As this is a cherry-pick from the master, you can find the analysis for each missing error in this old pull request https://github.com/hyperledger/fabric/pull/1543 (both @manish-sethi and I verified separately as well as together to ensure that we haven't missed anything) .

mastersingh24 (approved):

PR #1860, opened 2020-09-08
FAB-18210 etcdraft.Chain private halt() respect exported Halt()

tock-ibm:
#### Type of change
- Bug fix

#### Description
After etcdraft.Chain Halt() was called, e.g. from RemoveChannel, it might happen
that the internal halt() is still called by one of the chain's
go-routine. In this case, halt() should not call the haltCallback.

#### Related issues
Issue: FAB-18210


jyellick (commented):
Looks good, but no test?

tock-ibm:
>Looks good, but no test?

Voila

C0rWin (approved):

PR #1861, opened 2020-09-08
[FAB-18127] Chan.Part.API IT: Join as follower

stephyee:
#### Type of change

- Test update

#### Description

Integration test for join application channel as follower.

#### Related issues

[FAB-18127](https://jira.hyperledger.org/browse/FAB-18127)

### DIFF
--- a/integration/raft/channel_participation_test.go
+++ b/integration/raft/channel_participation_test.go
@@ -141,21 +141,21 @@ var _ = Describe("ChannelParticipation", func() {
 			memberRunners := []*ginkgomon.Runner{ordererRunners[0], ordererRunners[1]}
 			leader := findLeader(memberRunners)
 
-			By("ensuring the channel is usable by submitting a transaction to each member")
-			env := CreateBroadcastEnvelope(network, peer, "participation-trophy", []byte("hello"))
-			for _, o := range members {
-				By("submitting transaction to " + o.Name)
-				expectedChannelInfoPT.Height++
-				resp, err := ordererclient.Broadcast(network, o, env)
-				Expect(err).NotTo(HaveOccurred())
-				Expect(resp.Status).To(Equal(common.Status_SUCCESS))
-				expectedBlockNumPerChannel := map[string]int{"participation-trophy": int(expectedChannelInfoPT.Height - 1)}
-				assertBlockReception(expectedBlockNumPerChannel, members, peer, network)
+			submitTxn(orderer1, peer, network, members, 1, channelparticipation.ChannelInfo{
+				Name:            "participation-trophy",
+				URL:             "/participation/v1/channels/participation-trophy",
+				Status:          "active",
+				ClusterRelation: "member",
+				Height:          2,
+			})
 
-				By("checking the channel height")
-				channelInfo := channelparticipation.ListOne(network, o, "participation-trophy")
-				Expect(channelInfo).To(Equal(expectedChannelInfoPT))
-			}
+			submitTxn(orderer2, peer, network, members, 2, channelparticipation.ChannelInfo{
+				Name:            "participation-trophy",
+				URL:             "/participation/v1/channels/participation-trophy",
+				Status:          "active",
+				ClusterRelation: "member",
+				Height:          3,
+			})
 
 			By("joining orderer3 to the channel as a follower")
 			// make sure we can join using a config block from one of the other orderers
@@ -171,7 +171,7 @@ var _ = Describe("ChannelParticipation", func() {
 
 			By("ensuring orderer3 completes onboarding successfully")
 			expectedChannelInfoPTFollower.Status = "active"
-			expectedChannelInfoPTFollower.Height = expectedChannelInfoPT.Height
+			expectedChannelInfoPTFollower.Height = 3
 			Eventually(func() channelparticipation.ChannelInfo {
 				return channelparticipation.ListOne(network, orderer3, "participation-trophy")
 			}, network.EventuallyTimeout).Should(Equal(expectedChannelInfoPTFollower))
@@ -185,7 +185,7 @@ var _ = Describe("ChannelParticipation", func() {
 
 			By("ensuring orderer3 transitions from follower to member")
 			// config update above added a block
-			expectedChannelInfoPT.Height++
+			expectedChannelInfoPT.Height = 4
 			Eventually(func() channelparticipation.ChannelInfo {
 				return channelparticipation.ListOne(network, orderer3, "participation-trophy")
 			}, network.EventuallyTimeout).Should(Equal(expectedChannelInfoPT))
@@ -195,12 +195,11 @@ var _ = Describe("ChannelParticipation", func() {
 			Expect(newLeader).To(Equal(leader))
 
 			By("submitting transaction to orderer3 to ensure it is active")
-			env = CreateBroadcastEnvelope(network, peer, "participation-trophy", []byte("hello-again"))
-			expectedChannelInfoPT.Height++
+			env := CreateBroadcastEnvelope(network, peer, "participation-trophy", []byte("hello-again"))
 			resp, err := ordererclient.Broadcast(network, orderer3, env)
 			Expect(err).NotTo(HaveOccurred())
 			Expect(resp.Status).To(Equal(common.Status_SUCCESS))
-			expectedBlockNumPerChannel := map[string]int{"participation-trophy": int(expectedChannelInfoPT.Height - 1)}
+			expectedBlockNumPerChannel := map[string]int{"participation-trophy": 4}
 			assertBlockReception(expectedBlockNumPerChannel, orderers, peer, network)
 
 			By("joining orderer1 to another channel as a member")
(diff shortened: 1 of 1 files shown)


integration/raft/channel_participation_test.go
```diff
@@ -254,6 +254,119 @@ var _ = Describe("ChannelParticipation", func() {
...
+			Expect(newLeader).To(Equal(leader))
+
+			By("submitting transaction to orderer3 to ensure it is active")
+			env = CreateBroadcastEnvelope(network, peer, "participation-trophy", []byte("hello-again"))
+			expectedChannelInfoPT.Height++
+			resp, err := ordererclient.Broadcast(network, orderer3, env)
+			Expect(err).NotTo(HaveOccurred())
+			Expect(resp.Status).To(Equal(common.Status_SUCCESS))
+			expectedBlockNumPerChannel := map[string]int{"participation-trophy": 5}
+			assertBlockReception(expectedBlockNumPerChannel, orderers, peer, network)
```
tock-ibm: How about doing ListOne again?

tock-ibm:
One minor comment, otherwise looks good to me.

integration/raft/channel_participation_test.go
stephyee: 🧹

tock-ibm (dismissed):

tock-ibm (dismissed):

integration/raft/channel_participation_test.go
```diff
@@ -365,6 +369,124 @@ var _ = Describe("ChannelParticipation", func() {
...
+			By("waiting for the leader to be ready")
+			memberRunners := []*ginkgomon.Runner{ordererRunners[0], ordererRunners[1]}
+			leader := findLeader(memberRunners)
+
+			By("ensuring the channel is usable by submitting a transaction to each member")
+			env := CreateBroadcastEnvelope(network, peer, "participation-trophy", []byte("hello"))
+			for i, o := range members {
+				By("submitting transaction to " + o.Name)
+				expectedChannelInfoPT.Height++
+				i++
```
C0rWin: a nit again, but I found it's very weird you are using slice iteration index to assert for certain block number reception, instead of having clearly readable and explicit variable for that.

integration/raft/channel_participation_test.go
```diff
@@ -365,6 +369,124 @@ var _ = Describe("ChannelParticipation", func() {
...
+			for _, o := range members {
+				By("joining " + o.Name + " to channel as a member")
+				channelparticipation.Join(network, o, "participation-trophy", genesisBlock, expectedChannelInfoPT)
+				channelInfo := channelparticipation.ListOne(network, o, "participation-trophy")
+				Expect(channelInfo).To(Equal(expectedChannelInfoPT))
+			}
+
+			By("waiting for the leader to be ready")
+			memberRunners := []*ginkgomon.Runner{ordererRunners[0], ordererRunners[1]}
+			leader := findLeader(memberRunners)
```
C0rWin, line 414: I wounder how long does it take to converge with leadership selection having only two active replicas.

integration/raft/channel_participation_test.go
wlahti: Agreed. This does seem weird to use "i" here. 

Maybe we should change this and the similar loops elsewhere in this test file to 
1) create a test helper that does the logic currently in the loop to submit a transaction and check the block number and channel info return by List. It can take the expected channel info and expected block number as arguments.
2) use the helper function directly for each orderer we're trying to test instead of looping (aka 'testHelper(members[0]...)` that way we explicitly set the expected channel info for each and directly pass the expected block number.

I know I'm responsible for the loop logic that existed before this PR so let me know if you'd like me to clean things up. :)

integration/raft/channel_participation_test.go
```diff
@@ -365,6 +369,124 @@ var _ = Describe("ChannelParticipation", func() {
 			channelInfo := channelparticipation.ListOne(network, orderer3, "participation-trophy")
 			Expect(channelInfo).To(Equal(expectedChannelInfoPT))
 		})
+
+		It("join application channel with join-block as follower via channel participation api", func() {
+			orderer1 := network.Orderer("orderer1")
+			orderer2 := network.Orderer("orderer2")
+			orderer3 := network.Orderer("orderer3")
+			orderers := []*nwo.Orderer{orderer1, orderer2}
```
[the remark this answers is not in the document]
stephyee: 🧹

integration/raft/channel_participation_test.go
stephyee: I revised using something similar to the 2nd option @wlahti proposed for this test

integration/raft/channel_participation_test.go
stephyee: **blinks**

integration/raft/channel_participation_test.go
```diff
@@ -140,13 +140,14 @@ var _ = Describe("ChannelParticipation", func() {
...
-			for _, o := range members {
+			for i, o := range members {
 				By("submitting transaction to " + o.Name)
+				i++
 				expectedChannelInfoPT.Height++
 				resp, err := ordererclient.Broadcast(network, o, env)
 				Expect(err).NotTo(HaveOccurred())
 				Expect(resp.Status).To(Equal(common.Status_SUCCESS))
-				expectedBlockNumPerChannel := map[string]int{"participation-trophy": int(expectedChannelInfoPT.Height - 1)}
+				expectedBlockNumPerChannel := map[string]int{"participation-trophy": i}
```
wlahti: I know you didn't introduce this code (guilty as charged) but can we do the similar fix here too?

integration/raft/channel_participation_test.go
```diff
@@ -240,18 +241,21 @@ var _ = Describe("ChannelParticipation", func() {
...
+
+			i := 5
 			for _, o := range orderers {
 				By("submitting transaction to " + o.Name)
 				expectedChannelInfoPT.Height++
 				resp, err := ordererclient.Broadcast(network, o, env)
 				Expect(err).NotTo(HaveOccurred())
 				Expect(resp.Status).To(Equal(common.Status_SUCCESS))
-				expectedBlockNumPerChannel := map[string]int{"participation-trophy": int(expectedChannelInfoPT.Height - 1)}
+				expectedBlockNumPerChannel := map[string]int{"participation-trophy": i}
```
wlahti: Similar fix here too?

integration/raft/channel_participation_test.go
stephyee: 🧹

integration/raft/channel_participation_test.go
stephyee: 🧹

integration/raft/channel_participation_test.go
```diff
@@ -353,22 +350,152 @@ var _ = Describe("ChannelParticipation", func() {
...
+			resp, err := ordererclient.Broadcast(network, orderer3, env)
+			Expect(err).NotTo(HaveOccurred())
+			Expect(resp.Status).To(Equal(common.Status_SUCCESS))
+			expectedBlockNumPerChannel := map[string]int{"participation-trophy": 5}
+			assertBlockReception(expectedBlockNumPerChannel, orderers, peer, network)
+
+			By("checking the channel height")
+			expectedChannelInfoPT.Height = 6
+			channelInfo := channelparticipation.ListOne(network, orderer3, "participation-trophy")
+			Expect(channelInfo).To(Equal(expectedChannelInfoPT))
```
wlahti: Nit: can't we use the submitTxn() helper here?

integration/raft/channel_participation_test.go
```diff
@@ -353,22 +350,152 @@ var _ = Describe("ChannelParticipation", func() {
...
+		})
+
 	})
 
 })
 
+func submitTxn(o *nwo.Orderer, peer *nwo.Peer, network *nwo.Network, orderers []*nwo.Orderer,
+	expectedBlkNum int, expectedChannelInfo channelparticipation.ChannelInfo) {
+	By("ensuring the channel is usable by submitting a transaction to " + o.Name)
```
wlahti: Nit: now that this is in a helper function, I feel like we shouldn't mention the "ensuring the channel is usable" part.
```suggestion
	By("submitting a transaction to " + o.Name)
```

wlahti (commented):
A few final nits but looking good otherwise. Thanks, Tiffany!

integration/raft/channel_participation_test.go
stephyee: 🧹

integration/raft/channel_participation_test.go
stephyee: 🧹

integration/raft/channel_participation_test.go
```diff
@@ -353,22 +350,159 @@ var _ = Describe("ChannelParticipation", func() {
...
+			Expect(newLeader).To(Equal(leader))
+
+			submitTxn(orderer3, peer, network, orderers, 5, channelparticipation.ChannelInfo{
+				Name:            "participation-trophy",
+				URL:             "/participation/v1/channels/participation-trophy",
+				Status:          "active",
+				ClusterRelation: "member",
+				Height:          6,
+			})
+			// By("submitting transaction to orderer3 to ensure it is active")
```
wlahti: Oops, you left this here. :)

stephyee: 🧹

integration/raft/channel_participation_test.go
```diff
@@ -533,6 +640,21 @@ var _ = Describe("ChannelParticipation", func() {
...
 })
 
+func submitTxn(o *nwo.Orderer, peer *nwo.Peer, network *nwo.Network, orderers []*nwo.Orderer,
+	expectedBlkNum int, expectedChannelInfo channelparticipation.ChannelInfo) {
+	By("submitting a transaction to " + o.Name)
+	env := CreateBroadcastEnvelope(network, peer, expectedChannelInfo.Name, []byte("hello"))
+	resp, err := ordererclient.Broadcast(network, o, env)
+	Expect(err).NotTo(HaveOccurred())
+	Expect(resp.Status).To(Equal(common.Status_SUCCESS))
+	expectedBlockNumPerChannel := map[string]int{"participation-trophy": expectedBlkNum}
```
wlahti, line 650:
```suggestion
	expectedBlockNumPerChannel := map[string]int{expectedChannelInfo.Name: expectedBlkNum}
```
Nit but we can fix it the next time we add/update an integration test here.

jyellick (approved):

PR #1862, opened 2020-09-09
FAB-18204 Ch.Part.API: decouple multichannel and etcdraft

tock-ibm:
The etcdraft package imports the multichannel package.
This prevents the multichannel package from using the onboarding
and cluster packages, which depend on etcdraft.

Here etcdraft is decoupled from multichannel by declaring an
interface that capture the dependecy on the Registrar.

This also eliminates some of the opaque function pointers to the
Registrar that the etcdraft package was using.

Change-Id: If8112edd7a8fe08a151bb3ca19bf0eeeedc2fb31

#### Type of change

#### Description

Refactoring to remove dependency.

#### Related issues
Task: FAB-18204
Epic: FAB-17712


jyellick (approved):
Looks like an improvement to me.  I still question a little why the consensus plugins should be responsible for transitioning to follower state, or calling 'CreateChain', but, better than the opaque function pointers.

PR #1863, opened 2020-09-09
Peer CLI communicate with orderers with expired TLS certs

wlahti:
#### Type of change

- New feature

#### Description

Implement a TLS handshake timeshift for the "peer channel fetch" and "peer channel update" comands to allow fetching config blocks and updating the config for orderers with expired TLS certificates.

#### Related issues

[FAB-18205](https://jira.hyperledger.org/browse/FAB-18205)

### DIFF
--- a/integration/nwo/commands/peer.go
+++ b/integration/nwo/commands/peer.go
@@ -8,6 +8,7 @@ package commands
 
 import (
 	"strconv"
+	"time"
 )
 
 type NodeStart struct {
@@ -135,11 +136,12 @@ func (c ChannelJoin) Args() []string {
 }
 
 type ChannelFetch struct {
-	ChannelID  string
-	Block      string
-	Orderer    string
-	OutputFile string
-	ClientAuth bool
+	ChannelID             string
+	Block                 string
+	Orderer               string
+	OutputFile            string
+	ClientAuth            bool
+	TLSHandshakeTimeShift time.Duration
 }
 
 func (c ChannelFetch) SessionName() string {
@@ -149,15 +151,10 @@ func (c ChannelFetch) SessionName() string {
 func (c ChannelFetch) Args() []string {
 	args := []string{
 		"channel", "fetch", c.Block,
-	}
-	if c.ChannelID != "" {
-		args = append(args, "--channelID", c.ChannelID)
-	}
-	if c.Orderer != "" {
-		args = append(args, "--orderer", c.Orderer)
-	}
-	if c.OutputFile != "" {
-		args = append(args, c.OutputFile)
+		"--channelID", c.ChannelID,
+		"--orderer", c.Orderer,
+		"--tlsHandshakeTimeShift", c.TLSHandshakeTimeShift.String(),
+		c.OutputFile,
 	}
 	if c.ClientAuth {
 		args = append(args, "--clientauth")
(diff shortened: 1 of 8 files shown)


integration/nwo/commands/peer.go
```diff
@@ -149,15 +151,10 @@ func (c ChannelFetch) SessionName() string {
...
-	if c.ChannelID != "" {
-		args = append(args, "--channelID", c.ChannelID)
-	}
-	if c.Orderer != "" {
-		args = append(args, "--orderer", c.Orderer)
-	}
-	if c.OutputFile != "" {
-		args = append(args, c.OutputFile)
+		"--channelID", c.ChannelID,
+		"--orderer", c.Orderer,
```
jyellick, line 155: Do we never use this command to pull blocks from a peer? Certainly channelID should not be empty, but it seems that the others could be?

internal/peer/channel/channel.go
```diff
@@ -86,6 +89,7 @@ func resetFlags() {
 	flags.StringVarP(&outputBlock, "outputBlock", "", common.UndefinedParamValue, `The path to write the genesis block for the channel. (default ./<channelID>.block)`)
 	flags.DurationVarP(&timeout, "timeout", "t", 10*time.Second, "Channel creation timeout")
 	flags.BoolVarP(&bestEffort, "bestEffort", "", false, "Whether fetch requests should ignore errors and return blocks on a best effort basis")
+	flags.DurationVarP(&tlsHandshakeTimeShift, "tlsHandshakeTimeShift", "", 0, "The amount of time to shift backwards during TLS handshakes")
```
jyellick: Perhaps expand a bit in the descriptive text, something like "... of time to shift backwards for certificate expiration checks during ..."

internal/peer/common/ordererclient.go
```diff
@@ -22,11 +23,12 @@ type OrdererClient struct {
 
 // NewOrdererClientFromEnv creates an instance of an OrdererClient from the
 // global Viper instance
-func NewOrdererClientFromEnv() (*OrdererClient, error) {
+func NewOrdererClientFromEnv(tlsHandshakeTimeShift time.Duration) (*OrdererClient, error) {
```
jyellick: Not saying that the viper approach is a great one, but it's a little odd that all of the parameters for setting up the ordering connection are wired in via `configFromEnv`, but not the timeshift.  And, it's also pretty ugly that everyone calling for this function has to explicitly disable the timeshift by passing in a zero value.  I don't know that it gets any less ugly following the existing pattern, but thought I'd ask?

integration/nwo/commands/peer.go
wlahti: I made this change because the code here was unnecessarily complex and inconsistent with (at least the majority) of the command helpers. For these args, we can just set them regardless of whether they're empty or not. The CLI logic will fetch from the peer if the specified orderer arg is empty.

internal/peer/common/ordererclient.go
wlahti: Yeah, I definitely don't love this change. I did it this way because I hadn't realized the viper config that `configFromEnv` was using actually derives from flags that are then `viper.Set` in `peer/common/ordererenv.go`'s `SetOrdererEnv()`. I'll update things to be consistent with the others now that I realize that.

internal/peer/channel/channel.go
wlahti: Done

internal/peer/common/ordererclient.go
wlahti: Fixed

jyellick (approved):

PR #1864, opened 2020-09-09
Create statedb indexes when creating ledger from snapshot

manish-sethi:

#### Type of change
- New feature

#### Description
When we create a new ledger from a snapshot, we need to create indexes in the statedb (couchdb) for the chaincodes that are already defined (and approved by the org) in the imported state from the snapshot, and are already installed on the bootstrapping peer. This PR adds the code for dealing with the chaincodes that are defined and installed via the new lifecycle (namely the `_lifecycle`).

#### Additional details
There will be a separate PR for handling the index creation for the chaincodes that are defined and installed via the legacy method (namely, the `lscc`). Note that, in the regular path of indexes creation, the legacy chaincodes are dealt with separately and hence it will be good to use the same mechanism in this scenario. As a side note, eventually this legacy path is expected to be removed when we remove the support of legacy chaincodes in a future release

#### Related issues
FAB-18158

### DIFF
--- a/core/chaincode/lifecycle/cache.go
+++ b/core/chaincode/lifecycle/cache.go
@@ -588,8 +588,26 @@ func (c *Cache) update(initializing bool, channelID string, dirtyChaincodes map[
 }
 
 // RegisterListener registers an event listener for receiving an event when a chaincode becomes invokable
-func (c *Cache) RegisterListener(channelID string, listener ledger.ChaincodeLifecycleEventListener) {
-	c.eventBroker.RegisterListener(channelID, listener)
+func (c *Cache) RegisterListener(
+	channelID string,
+	listener ledger.ChaincodeLifecycleEventListener,
+	needsExistingChaincodesDefinitions bool,
+) error {
+	c.mutex.RLock()
+	defer c.mutex.RUnlock()
+
+	channelChaincodes, ok := c.definedChaincodes[channelID]
+	if !ok {
+		return errors.Errorf("unknown channel '%s'", channelID)
+	}
+
+	if needsExistingChaincodesDefinitions {
+		c.eventBroker.RegisterListener(channelID, listener, channelChaincodes.Chaincodes)
+	} else {
+		c.eventBroker.RegisterListener(channelID, listener, nil)
+	}
+
+	return nil
 }
 
 func (c *Cache) InitializeMetadata(channel string) {
--- a/core/ledger/kvledger/kv_ledger.go
+++ b/core/ledger/kvledger/kv_ledger.go
@@ -68,6 +68,7 @@ type kvLedger struct {
 
 type lgrInitializer struct {
 	ledgerID                 string
+	initializingFromSnapshot bool
 	bootSnapshotMetadata     *snapshotMetadata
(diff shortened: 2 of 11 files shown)


core/chaincode/lifecycle/cache.go
```diff
@@ -588,8 +588,26 @@ func (c *Cache) update(initializing bool, channelID string, dirtyChaincodes map[
...
+		return errors.Errorf("unknown channel '%s'", channelID)
+	}
+
+	if needsExistingChaincodesDefinitions {
+		c.eventBroker.RegisterListener(channelID, listener, channelChaincodes.Chaincodes)
+	} else {
+		c.eventBroker.RegisterListener(channelID, listener, nil)
+	}
+
+	return nil
```
cendhu, line 610: nit. `return c.eventBroker.RegisterListener(channelID, listener, nil)`

core/ledger/kvledger/kv_ledger.go
```diff
@@ -148,7 +149,11 @@ func newKVLedger(initializer *lgrInitializer) (*kvLedger, error) {
 	logger.Debugf("Register state db for chaincode lifecycle events: %t", ccEventListener != nil)
 	if ccEventListener != nil {
 		cceventmgmt.GetMgr().Register(ledgerID, ccEventListener)
-		initializer.ccLifecycleEventProvider.RegisterListener(ledgerID, &ccEventListenerAdaptor{ccEventListener})
+		initializer.ccLifecycleEventProvider.RegisterListener(
```
cendhu, line 152: Now, the register function is doing more than just registering. However, I could not think of any other simple way as the existing dependencies are complex already.

1. ccLifecycle register a listener with the ledger to receive state updates whereas the ledger register a listener with ccLifecycle to receive the CouchDB index
2. ccLifecycle cache needs to be initialized from the ledger and the CouchDB indexes are initialized from the ccLifecycle for the ledger that are bootstrapped from a snapshot.

As a larger refactoring is required to fix the above, I am okay with it as-is.

core/ledger/kvledger/snapshot_test.go
```diff
@@ -232,6 +236,110 @@ func TestSnapshotDBTypeCouchDB(t *testing.T) {
...
+		actualResults := []*queryresult.KV{}
+		for {
+			queryResults, err := iter.Next()
+			require.NoError(t, err)
+			if queryResults == nil {
+				break
+			}
+			actualResults = append(actualResults, queryResults.(*queryresult.KV))
+		}
+		require.Len(t, actualResults, 1)
```
cendhu, line 337: nit. can check the value.

cendhu (approved):

core/chaincode/lifecycle/cache.go
manish-sethi: try this :-)

cendhu: Ha ha. Got it. Two returns are needed.

kalid117658:
https://github.com/hyperledger/fabric/pull/1864#discussion_r486391490

PR #1879, opened 2020-09-15
add config option for deprioritized missing data access time

cendhu:
#### Type of change

- New feature


ale-linux (commented):
You probably already had extensive discussions before you decided to add this new option but let me just give you a chance to present them in this forum too: what would you say to someone arguing that we already have way too many tunables and we don't need the umpteenth one?

cendhu:
> You probably already had extensive discussions before you decided to add this new option but let me just give you a chance to present them in this forum too: what would you say to someone arguing that we already have way too many tunables and we don't need the umpteenth one?

You can read my initial opinion on this here — https://github.com/hyperledger/fabric/pull/1797#discussion_r478715814 

Later, I got convinced that in certain scenarios, these configs are useful in the BaaS platform (but not for the app developer or user).

ale-linux:
Thanks Senthil, understood!

denyeart:
Merged https://github.com/hyperledger/fabric/pull/1880 instead, which contains this commit and others. Closing...

PR #1881, opened 2020-09-15
Adding notes for the usage of script during samples install

rijulaggarwal16:
Signed-off-by: Rijul Aggarwal <rijul.aggarwal@ibm.com>

#### Type of change

- Documentation update

#### Description

The main motivation behind this is that it's hard to know the details about the script that we use for installing and setting up fabric samples. It becomes easier to handle erroneous situations if we know what the script is capable of, and what it's trying to do.

The script to be used provides some options that a user can provide to entertain different behaviour from the script/install step.

#### Additional details

You can visit https://bit.ly/2ysbOFE to see the contents of the script used for install

### DIFF
--- a/docs/source/install.rst
+++ b/docs/source/install.rst
@@ -64,6 +64,10 @@ the binaries and images.
 	  the un-shortened URL:
 	  https://raw.githubusercontent.com/hyperledger/fabric/{BRANCH}/scripts/bootstrap.sh
 
+.. note:: For additional use pattern you can use the -h flag to view the help and available commands for the 
+          Fabric-Samples bootstrap script. For example:
+          ``curl -sSL https://bit.ly/2ysbOFE | bash -s -- -h``
+
 The command above downloads and executes a bash script
 that will download and extract all of the platform-specific binaries you
 will need to set up your network and place them into the cloned repo you


docs/source/install.rst
```diff
@@ -45,14 +45,17 @@ the binaries and images.
...
 .. note:: If you want a specific release, pass a version identifier for Fabric and Fabric-CA docker images.
           The command below demonstrates how to download the latest production releases -
           **Fabric v2.2.0** and **Fabric CA v1.4.8**
 
 .. code:: bash
 
-  curl -sSL https://bit.ly/2ysbOFE | bash -s -- <fabric_version> <fabric-ca_version>
-  curl -sSL https://bit.ly/2ysbOFE | bash -s -- 2.2.0 1.4.8
+  curl -sSL https://bit.ly/2ysbOFE | bash -s -- <fabric_version> <fabric-ca_version> <script_option>
+  curl -sSL https://bit.ly/2ysbOFE | bash -s -- 2.2.0 1.4.8 -s
```
lindluni: Adding the `-s` flag here actually isn't appropriate, while it make sense now, because the current master of Fabric-Samples is what is used for the `2.2.0` images today, once we cut the `2.2.0` tag in fabric samples the command will no longer be appropriate. I think these two lines should be reverted in favor of my comments above.

Today the preceding commands only have you clone the repo, they don't have you checkout a specific tag. So adding the `-s` flag would prevent actually checking out the correct tag in the coming weeks.

lindluni (changes requested):
Thank you for the contribution. Just a couple minor comments.

lindluni (approved):

lindluni:

PR #1883, opened 2020-09-15
Add extra queryexecutor nil checks

jyellick:
Because system chaincode sometimes executes in a channel-less context,
there is a broken path throughout the chaincode package where despite
being passed as a parameter, the query executor is nil.

This path has been significantly cleaned up in the v2.0 code, though, so
long as system chaincode exists which attempts to execute without a
channel context, we will have this problem.

This small PR simply adds extra nil checks to prevent a crash in the
cases where a nil query executor makes it to this point in the code
path.


lindluni:
Well thats a new flake (at least it was caused by CI and not our codebase)

lindluni (approved):

denyeart (approved):
I tested e2e from CLI and indeed was able to reproduce the issue, and now I get the new error.
(I had to disable the channelid validation from CLI first, to be able to reproduce).

denyeart:
And I confirmed the error is handled well on v2.x, even when using old lifecycle.

PR #1884, opened 2020-09-16
FAB-17911 Ch.Part.API: join system channel

tock-ibm:
An orderer without channels joins the system channel.
It must be a member of the system channel cluster as well.

1. The join-block is saved to the join-block filerepo.
2. If the join-block.number=0, it is appended to the ledger.
3. A degenerate ChainSupport is created, holding an inactive.Chain
   within, and set into the chains map and the "systemChannel" field.
   This is done to prevent further invocations to the API from issuing
   conflicting commands, as they will find a system channel there.
4. Only "List" and "Remove" will be accepted. Transactions sent to the
   system channel will be rejected.
5. As soon as the Join REST call returns successfully the orderer should
   be restarted.
6. Upon restart, (boostrapMethod="none") 'Main' will look for a system
   channel join block, and if it finds one, will treat it as a bootstrap block.
7. Normal boot then resumes, and the orderer will on-board (replicate) the
   system channel and any channel referred by it, just as if we had 
   boostrapMethod="file" with the same block as bootstrap block.
8. After replication ends, the system channel join block is removed.
9. The orderer now operates in the "legacy" system channel mode.

Change-Id: Ief7511de61eccc876f83157315bca98f9e30397c

#### Type of change

- New feature

#### Related issues
Task: FAB-17911
Epic: FAB-17712

### DIFF
--- a/orderer/common/multichannel/chainsupport.go
+++ b/orderer/common/multichannel/chainsupport.go
@@ -12,9 +12,11 @@ import (
 	"github.com/hyperledger/fabric/common/ledger/blockledger"
 	"github.com/hyperledger/fabric/internal/pkg/identity"
 	"github.com/hyperledger/fabric/orderer/common/blockcutter"
+	"github.com/hyperledger/fabric/orderer/common/localconfig"
 	"github.com/hyperledger/fabric/orderer/common/msgprocessor"
 	"github.com/hyperledger/fabric/orderer/common/types"
 	"github.com/hyperledger/fabric/orderer/consensus"
+	"github.com/hyperledger/fabric/orderer/consensus/inactive"
 	"github.com/hyperledger/fabric/protoutil"
 	"github.com/pkg/errors"
 )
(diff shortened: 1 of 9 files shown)


orderer/common/multichannel/chainsupport.go
```diff
@@ -180,3 +182,18 @@ func (cs *ChainSupport) Sequence() uint64 {
...
 	return cs.ledgerResources.ReadWriter.Append(block)
 }
+
+func newOnBoardingChainSupport(
+	ledgerResources *ledgerResources,
+	config localconfig.TopLevel,
+	bccsp bccsp.BCCSP,
+) (*ChainSupport, error) {
+	cs := &ChainSupport{ledgerResources: ledgerResources}
+	cs.Processor = msgprocessor.NewStandardChannel(cs, msgprocessor.CreateStandardChannelFilters(cs, config), bccsp)
```
jyellick, line 192: I would have thought this field would be safe to leave uninitialized? What path is the processor for this channel called?  If it's in the broadcast/deliver path, perhaps we'd be better off explicitly returning an error of some sort?

jyellick (approved):
I'm a little unsure why we're actually populating the processor on the inactive chain, see comment below, but otherwise looks good to me.

orderer/common/multichannel/chainsupport.go
tock-ibm: It is in the broadcast path. The first is "ClasifyHeader" and then "Process*" which will ultimately get to the inactive.Chain which is primed with an informative error message. We'll test that in IT.

PR #1885, opened 2020-09-16
FAB-18233 Ch.Part.API: BlockPuller verify skip-genesis

tock-ibm:
The verify returns an error if the block slice contains only
one block (number 0), since it tries to further verify an
empty slice.

Change-Id: I38589499c06acbf81b7b171031efd0fd47093132

#### Type of change

- Bug fix

#### Related issues

Issue: FAB-18233


wlahti (approved):
Looks good to me! Manually verified that it resolves the problem I hit while preparing the demo of the channel participation API for the contributors meeting. :)

yacovm (approved):

PR #1886, opened 2020-09-16
pass unreconciled missing pvtdata to pvtdata store (backport to release-2.2)

denyeart:
#### Type of change

- Bug fix

#### Description

- Pass unreconciled missing pvtdata to pvtdata store so that it can be deprioritized from subsequent reconciliation attempts (commit3)
- Refactor tests (commit2)
- Deprioritized missing private data will be re-attempted every deprioritizedDataReconcilerInterval (commit1)


sampleconfig/core.yaml
```diff
@@ -678,6 +678,16 @@ ledger:
...
+    # The missing data entries are classified into two categories:
+    # (1) prioritized
+    # (2) deprioritized
+    # Initially, all missing data are in the prioritized list. When the
+    # reconciler is unable to fetch the missing data from other peers,
+    # the unreconciled missing data would be moved to the deprioritized list.
+    # The reconciler would retry deprioritized missing data after every
+    # deprioritizedDataReconcilerInterval (unit: minutes). Note that the
+    # interval needs to be greater than the reconcileSleepInterval
+    deprioritizedDataReconcilerInterval: 60m
```
manish-sethi, line 690: I would have preferred to keep this config close to `reconcileSleepInterval` for usability and we could have a check in the code and print a warning if this is relatively smaller than that. However, that would criss-cross in the code unless we make reconciler control this and pass explicitly the category in the call... Per the current code, I am not sure what else can we do, so we can keep this as is.

manish-sethi (commented):
Though, its a back-port but I have a few comment that are applicable to the merged PRs in the master as well. They can be addressed in the master and back ported as well.

denyeart:
Comments addressed in https://github.com/hyperledger/fabric/pull/1890 (master) and will be backported as well.

manish-sethi (approved):

PR #1889, opened 2020-09-17
[FAB-18164] Change CreateLedgerFromSnapshot to async

wenjianqiao:

#### Type of change
- New feature

#### Description
- Change CreateLedgerFromSnapshot to async
- Track the ledger when it is under creation from a snapshot
- Do not allow to create another ledger if a ledger
  is under creation from a snapshot

#### Related issues
https://jira.hyperledger.org/browse/FAB-18164

### DIFF
--- a/core/ledger/ledgermgmt/ledger_mgmt.go
+++ b/core/ledger/ledgermgmt/ledger_mgmt.go
@@ -18,6 +18,7 @@ import (
 	"github.com/hyperledger/fabric/core/ledger"
 	"github.com/hyperledger/fabric/core/ledger/cceventmgmt"
 	"github.com/hyperledger/fabric/core/ledger/kvledger"
+	"github.com/hyperledger/fabric/internal/fileutil"
 	"github.com/pkg/errors"
 )
 
@@ -31,9 +32,13 @@ var ErrLedgerMgmtNotInitialized = errors.New("ledger mgmt should be initialized 
 
 // LedgerMgr manages ledgers for all channels
 type LedgerMgr struct {
-	lock               sync.Mutex
-	openedLedgers      map[string]ledger.PeerLedger
-	ledgerProvider     ledger.PeerLedgerProvider
+	creationLock           sync.Mutex
+	createFromSnapshotInfo string
+
+	lock           sync.Mutex
+	openedLedgers  map[string]ledger.PeerLedger
+	ledgerProvider ledger.PeerLedgerProvider
+
 	ebMetadataProvider MetadataProvider
 }
 
@@ -93,9 +98,17 @@ func NewLedgerMgr(initializer *Initializer) *LedgerMgr {
 }
 
 // CreateLedger creates a new ledger with the given genesis block.
-// This function guarantees that the creation of ledger and committing the genesis block would an atomic action
-// The channel id retrieved from the genesis block is treated as a ledger id
+// This function guarantees that the creation of ledger and committing the genesis block would an atomic action.
+// The channel id retrieved from the genesis block is treated as a ledger id.
+// It returns an error if another ledger is being created from a snapshot.
 func (m *LedgerMgr) CreateLedger(id string, genesisBlock *common.Block) (ledger.PeerLedger, error) {
+	m.creationLock.Lock()
+	defer m.creationLock.Unlock()
+
+	if m.createFromSnapshotInfo != "" {
+		return nil, errors.Errorf("a ledger is being created from a snapshot at %s. Call ledger creation again after it is done.", m.createFromSnapshotInfo)
+	}
+
 	m.lock.Lock()
 	defer m.lock.Unlock()
 	logger.Infof("Creating ledger [%s] with genesis block", id)
@@ -112,10 +125,40 @@ func (m *LedgerMgr) CreateLedger(id string, genesisBlock *common.Block) (ledger.
 	}, nil
 }
 
-// CreateLedgerFromSnapshot creates a new ledger with the given snapshot and returns the ledger and channel id.
-// This function guarantees that the creation of ledger and all ledger dbs would be an atomic action.
-// The channel id retrieved from the snapshot metadata is treated as a ledger id
-func (m *LedgerMgr) CreateLedgerFromSnapshot(snapshotDir string) (ledger.PeerLedger, string, error) {
+// CreateLedgerFromSnapshot creates a new ledger with the given snapshot and executes the callback function
+// after the ledger is created. This function launches to goroutine to create the ledger and call the callback func.
+// All ledger dbs would be created in an atomic action. The channel id retrieved from the snapshot metadata
+// is treated as a ledger id. It returns an error if another ledger is being created from a snapshot.
+func (m *LedgerMgr) CreateLedgerFromSnapshot(snapshotDir string, channelCallback func(ledger.PeerLedger, string)) error {
+	// verify snapshotDir exists and is not empty
+	empty, err := fileutil.DirEmpty(snapshotDir)
+	if err != nil {
+		return err
+	}
+	if empty {
+		return errors.Errorf("snapshot dir %s is empty", snapshotDir)
+	}
+
+	if err := m.setCreateFromSnapshotInfo(snapshotDir); err != nil {
+		return err
+	}
+
+	go func() {
+		defer m.unsetCreateFromSnapshotInfo()
+
+		ledger, cid, err := m.createFromSnapshot(snapshotDir)
+		if err != nil {
+			logger.Errorw("Error creating ledger from snapshot", "snapshotDir", snapshotDir, "error", err)
+			return
+		}
+
+		channelCallback(ledger, cid)
+	}()
+
+	return nil
+}
+
+func (m *LedgerMgr) createFromSnapshot(snapshotDir string) (ledger.PeerLedger, string, error) {
 	m.lock.Lock()
 	defer m.lock.Unlock()
 	logger.Infof("Creating ledger from snapshot at %s", snapshotDir)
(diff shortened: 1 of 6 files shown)


core/ledger/ledgermgmt/ledger_mgmt.go
```diff
@@ -31,10 +31,12 @@ var ErrLedgerMgmtNotInitialized = errors.New("ledger mgmt should be initialized
...
 // LedgerMgr manages ledgers for all channels
 type LedgerMgr struct {
-	lock               sync.Mutex
-	openedLedgers      map[string]ledger.PeerLedger
-	ledgerProvider     ledger.PeerLedgerProvider
-	ebMetadataProvider MetadataProvider
+	lock                sync.Mutex
+	underCreationLock   sync.Mutex
+	openedLedgers       map[string]ledger.PeerLedger
+	underCreationLedger string
```
cendhu: Is it possible to use an atomic flag and `CompareAndSwap()` from [atomic pkg](https://golang.org/pkg/sync/atomic/) rather than the `underCreationLock` and `underCreationLedger`?

core/ledger/ledgermgmt/ledger_mgmt.go
```diff
@@ -132,6 +167,26 @@ func (m *LedgerMgr) CreateLedgerFromSnapshot(snapshotDir string) (ledger.PeerLed
 	}, cid, nil
 }
 
+func (m *LedgerMgr) setUnderCreationLedger(snapshotDir string) error {
+	m.underCreationLock.Lock()
+	defer m.underCreationLock.Unlock()
+	if m.underCreationLedger != "" {
+		return errors.New("a ledger is being created from a snapshot. Call 'peer channel joinbysnapshotstatus' to check the status and execute the command again after it is done.")
```
manish-sethi: I don't think that putting actual command `peer channel joinbysnapshotstatus` here is a good idea. If something changes in the command, no one will know that has to be fixed here.

core/ledger/ledgermgmt/ledger_mgmt.go
wenjianqiao: @cendhu The `underCreationLedger` is not simple compare and swap. For CreateLedger (by genesisblock), it does not swap and also holds the `underCreationLock` longer so that another CreateLedgerBySnapshot cannot acquire `lock` before `CreateLedger` is done.

core/ledger/ledgermgmt/ledger_mgmt.go
```diff
@@ -112,10 +122,35 @@ func (m *LedgerMgr) CreateLedger(id string, genesisBlock *common.Block) (ledger.
...
+
+		ledger, cid, err := m.createFromSnapshot(snapshotDir)
+		if err != nil {
+			logger.Errorw("Error creating ledger from snapshot", "snapshotDir", snapshotDir, "error", err)
+			return
+		}
+		if cid != m.underCreationLedger {
+			// this should not happen unless there is a code bug
+			logger.Errorw("Mismatched ledger ID", "createdLedger", cid, "underCreationLedger", m.underCreationLedger)
+		}
```
manish-sethi: I think that this confusion is uncalled for. I would prefer one of the following...
- change the function signature of `createFromSnapshot` to just return the ledger instance, as you have already extracted the ledgerID beforehand.
- Don't extract the ledgerID beforehand and simply use the snapshot dir path instead of the ledgerID in the error messages.

core/ledger/ledgermgmt/ledger_mgmt.go
```diff
@@ -132,6 +167,26 @@ func (m *LedgerMgr) CreateLedgerFromSnapshot(snapshotDir string) (ledger.PeerLed
...
+	m.underCreationLock.Lock()
+	defer m.underCreationLock.Unlock()
+	if m.underCreationLedger != "" {
+		return errors.New("a ledger is being created from a snapshot. Call 'peer channel joinbysnapshotstatus' to check the status and execute the command again after it is done.")
+	}
+	ledgerID, err := kvledger.GetChannelNameFromSnapshot(snapshotDir)
+	if err != nil {
+		return err
+	}
+	m.underCreationLedger = ledgerID
```
manish-sethi: I guess that it's better to add a check and return error if `ledgerID == ""`. On the same note, as far as I recall, there were other restrictions on the channelName... good to verify on the safer side.

core/ledger/ledgermgmt/ledger_mgmt.go
```diff
@@ -112,10 +122,35 @@ func (m *LedgerMgr) CreateLedger(id string, genesisBlock *common.Block) (ledger.
...
+
+		channelCallback(ledger, cid)
+	}()
+
+	return nil
+}
+
+func (m *LedgerMgr) createFromSnapshot(snapshotDir string) (ledger.PeerLedger, string, error) {
 	m.lock.Lock()
 	defer m.lock.Unlock()
```
manish-sethi, line 163: Now, is acquiring this lock here needed?.. rather this begs a question whether we really need two locks at all? - If you release this lock here after setting the underconstruction status.

core/ledger/ledgermgmt/ledger_mgmt.go
wenjianqiao: The usage of `underCreationLedger` is not a simple compare and swap case. As Manish pointed out, we don't need 2 locks. I am going to remove `underCreationLock`.

core/ledger/ledgermgmt/ledger_mgmt.go
wenjianqiao: agreed

core/ledger/ledgermgmt/ledger_mgmt.go
wenjianqiao: Good point. One lock should be sufficient.

core/ledger/ledgermgmt/ledger_mgmt.go
wenjianqiao: Will use the 2nd option so that we don't have to extract `ledgerID` twice.

core/ledger/ledgermgmt/ledger_mgmt.go
cendhu: 1. The `lock` mutex is used to synchronize between `Create()`, `Open()`, and `Close()`. The current semantics is that only one operation is allowed at a time. If we are planning to remove `underCreationLock`, instead use `lock` to set the status flag and release the `lock` immediately while the ledger opening process is still running in the background, we would be diverging from the current semantics. I am not saying that we should stick to the existing semantics but it is necessary to analyze the impact. 
2. Given that the `underCreationLock` is mainly used to set a flag (whether a channel creation is under-progress or not), I doubt why an atomic variable is not possible here while still retaining the existing semantics. With this, all exported methods on the ledgerMgr would check the status of the flag and inform the user about the long wait time (i.e.,  the time taken to bootstrap from the snapshot). For both checking the status and setting the status, I think, we can use atomic variable if we want to retain the existing semantics. Correct me if I have misunderstood anything.

wenjianqiao: Discussed with @manish-sethi and @cendhu and agreed to keep the semantic and use 2 locks. Using atomic var may miss a corner case where CreateLedger and CreateLedgerSnapshot calls both pass the atomic var checking and have to compete LedgerMgr.lock.

core/ledger/ledgermgmt/ledger_mgmt.go
wenjianqiao: Since this PR has been updated to not extract ledgerID, the channelID validation will be handled separately.

core/ledger/ledgermgmt/ledger_mgmt.go
wenjianqiao: After discussion with Manish and Senthil, we decided to keep the lock so that it does not change the current semantics where ledger mgr only allows one operation at a time.

core/ledger/ledgermgmt/ledger_mgmt.go
```diff
@@ -112,10 +123,40 @@ func (m *LedgerMgr) CreateLedger(id string, genesisBlock *common.Block) (ledger.
...
+	go func() {
+		defer m.unsetCreateFromSnapshotInfo()
+
+		ledger, cid, err := m.createFromSnapshot(snapshotDir)
+		if err != nil {
+			logger.Errorw("Error creating ledger from snapshot", "snapshotDir", snapshotDir, "error", err)
+			return
+		}
+
+		channelCallback(ledger, cid)
```
manish-sethi, line 155: I guess that it's important to cover this in the test - i.e., on error, callback function is not invoked. It should be easy to cause this error condition in a test by deleting one of the files in the snapshot folder.

core/ledger/ledgermgmt/ledger_mgmt_test.go
```diff
@@ -110,32 +109,173 @@ func TestCreateLedgerFromSnapshot(t *testing.T) {
 	}
 	require.Eventually(t, snapshotGenerated, 30*time.Second, 100*time.Millisecond)
```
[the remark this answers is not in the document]
wenjianqiao: It is still needed to make sure the snapshot generation is completed so that it is safe to run sub-tests.

core/ledger/ledgermgmt/ledger_mgmt.go
wenjianqiao: Added the test.

core/ledger/ledgermgmt/ledger_mgmt.go
```diff
@@ -161,6 +224,14 @@ func (m *LedgerMgr) GetLedgerIDs() ([]string, error) {
...
 }
 
+// GetCreateFromSnapshotInfo returns the snapshot dir if a ledger is being created from a snapshot.
+// Otherwise, it returns an empty string.
+func (m *LedgerMgr) GetCreateFromSnapshotInfo() string {
+	m.creationLock.Lock()
+	defer m.creationLock.Unlock()
+	return m.createFromSnapshotInfo
+}
+
```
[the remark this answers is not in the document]
wenjianqiao: This method will be used in next PR for `joinbysnapshotstatus`.

manish-sethi (approved):

PR #1900, opened 2020-09-18
Allow BCCSP config to be set using environment variables

wlahti:
#### Type of change

- Bug fix

#### Related issues

[FAB-17969](https://jira.hyperledger.org/browse/FAB-17969)

### DIFF
--- a/bccsp/factory/opts.go
+++ b/bccsp/factory/opts.go
@@ -1,20 +1,19 @@
 /*
-Copyright IBM Corp. 2016 All Rights Reserved.
+Copyright IBM Corp. All Rights Reserved.
 
-Licensed under the Apache License, Version 2.0 (the "License");
-you may not use this file except in compliance with the License.
-You may obtain a copy of the License at
-
-		 http://www.apache.org/licenses/LICENSE-2.0
-
-Unless required by applicable law or agreed to in writing, software
-distributed under the License is distributed on an "AS IS" BASIS,
-WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
-See the License for the specific language governing permissions and
-limitations under the License.
+SPDX-License-Identifier: Apache-2.0
 */
 
 package factory
+
+import "github.com/hyperledger/fabric/bccsp/pkcs11"
+
+// FactoryOpts holds configuration information used to initialize factory implementations
+type FactoryOpts struct {
+	Default string             `json:"default" yaml:"Default"`
+	SW      *SwOpts            `json:"SW,omitempty" yaml:"SW,omitempty"`
+	PKCS11  *pkcs11.PKCS11Opts `json:"PKCS11,omitempty" yaml:"PKCS11"`
+}
 
 // GetDefaultOpts offers a default implementation for Opts
 // returns a new instance every time
@@ -22,6 +21,10 @@ func GetDefaultOpts() *FactoryOpts {
 	return &FactoryOpts{
 		Default: "SW",
 		SW: &SwOpts{
+			Hash:     "SHA2",
+			Security: 256,
+		},
+		PKCS11: &pkcs11.PKCS11Opts{
 			Hash:     "SHA2",
 			Security: 256,
 		},
--- a/common/viperutil/config_util.go
+++ b/common/viperutil/config_util.go
@@ -404,7 +404,7 @@ func bccspHook(f reflect.Type, t reflect.Type, data interface{}) (interface{}, e
 
 	err := mapstructure.WeakDecode(data, config)
 	if err != nil {
-		return nil, errors.Wrap(err, "could not decode bcssp type")
+		return nil, errors.Wrap(err, "could not decode bccsp type")
 	}
 
 	return config, nil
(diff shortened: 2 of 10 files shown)


wlahti:
This isn't pretty but it at least works for the peer. I tried the orderer environment variables as well for the new test but those don't work currently either...

bccsp/factory/opts.go
```diff
@@ -1,21 +1,20 @@
...
+SPDX-License-Identifier: Apache-2.0
 */
 
 package factory
 
+import "github.com/hyperledger/fabric/bccsp/pkcs11"
+
+// FactoryOpts holds configuration information used to initialize factory implementations
+type FactoryOpts struct {
+	Default string             `json:"default" yaml:"Default"`
```
jyellick, line 13: I realize you're just moving this code about, but it's really bothersome that the yaml and json keys are differently cased.  I'm not sure who would be depending on this behavior, so perhaps it's safest not to change, but just have to point it out.

bccsp/factory/opts.go
```diff
@@ -25,6 +24,10 @@ func GetDefaultOpts() *FactoryOpts {
 			Hash:     "SHA2",
 			Security: 256,
 		},
+		PKCS11: &pkcs11.PKCS11Opts{
```
jyellick, line 27: I guess it's fine to consolidate these into a single struct regardless of build tag because we never try to interact with the pkcs11 lib unless we follow the path set out in the pkcs11.go file?

common/viperutil/config_util.go
```diff
@@ -202,6 +203,59 @@ func getKeysRecursively(base string, getenv envGetter, nodeKeys map[string]inter
 	return result
 }
 
+type viperGetter func(key string) interface{}
+
+func getViperKeysRecursively(base string, getKey viperGetter, nodeKeys map[string]interface{}, oType reflect.Type) map[string]interface{} {
```
jyellick: Was this the function that added so much runtime to our tests? Have you profiled the runtime before/after this change?

jyellick (commented):
Overall looks pretty good, not a lot of actual code changes and mostly test which is of course great.  Still, a few questions.

common/viperutil/config_util.go
wlahti: On first glance, it does seem like the overall time to complete UT increases significantly with this code... 27 minutes v. 12-13 minutes for the other PRs out there. Force-pushed to this PR to retrigger but I'll see if there's anything else we can do to fix these overrides without reviving this inefficient code.

wlahti: Attempting to incorporate the same caching mechanism that was added to improve the performance for the orderer/configtxgen. We'll see if that ends up with the same performance boost.

bccsp/factory/opts.go
wlahti: Yes, that was the thinking here.

common/viperutil/config_util.go
wlahti: Abandoned this horribly inefficient viper usage. :)

internal/peer/common/common.go
```diff
@@ -144,6 +150,54 @@ func InitCrypto(mspMgrConfigDir, localMSPID, localMSPType string) error {
...
+		bccspConfig.SW.Hash = swHash
+	}
+
+	if swSecurity, exist := os.LookupEnv("CORE_PEER_BCCSP_SW_SECURITY"); exist {
+		swSec, err := strconv.Atoi(swSecurity)
+		if err != nil {
+			return errors.Errorf("CORE_PEER_BCCSP_SW_SECURITY set to non-integer value: %s", swSecurity)
+		}
+		bccspConfig.SW.Security = swSec
+	}
```
wlahti: We could remove these SW override lookups since BCCSP hardcodes these anyways.

common/viperutil/config_util.go
wlahti: Actually, if you look at the CI analytics, it seems the jump in total UT time is unrelated: https://dev.azure.com/Hyperledger/Fabric/_pipeline/analytics/duration?definitionId=50&contextType=build

internal/peer/common/common.go
stephyee: agreed

jyellick (dismissed):
I don't love implementing this environment parsing manually, but given the downsides of the implementation through viperutil, it seems like the lesser of the two evils.  Did you want to address Will's comment now, or should we merge as is?

internal/peer/common/common.go
stephyee: 🧹

stephyee:
@jyellick removed the overrides for hardcoded SW values

jyellick (approved):

lindluni:
This change had some a pretty well hidden side affect. It pushed PKCS11 code outside of the `pkcs11` build tag gating, which now pulls the pkcs11 library in, even when you aren't using pkcs11 factories. This prevents cross-compilation of fabric from one platform to another now unless you have a C cross-platform compiler installed and pass the proper CGO options when building since the miekg package uses CGO

stephyee:
@btl5037 what do you suggest? reverting?

denyeart:
Resolved in https://github.com/hyperledger/fabric/pull/3697

PR #1903, opened 2020-09-18
Bugfix in collection config history mgr

manish-sethi:
Use proper stop key to prevent iterator returning results from next namespace

#### Type of change
- Bug fix

#### Additional details
The bug seems to have escaped because in the code we always call this API with the block number present in the version of the private data (or hash) item. Ran into this issue when loading all the configs for a namespace for populating private data store from a snapshot.


denyeart (approved):

denyeart:

PR #1907, opened 2020-09-21
[FAB-18196] Add Channel Participation CLI

stephyee:
#### Type of change

- New feature

#### Description

Adds CLI commands for the list, join, and remove channel participation APIs.

#### Related issues

[FAB-18196](https://jira.hyperledger.org/browse/FAB-18196)


C0rWin:
@stephyee there is no unit tests in this PR, could you add some to make sure the new code proceeds as expected?

cmd/participation/remove.go
```diff
@@ -0,0 +1,32 @@
...
+	}
+
+	resp, err := client.Do(req)
+	if err != nil {
+		panic(err)
+	}
+
+	if resp.StatusCode != http.StatusNoContent {
+		panic(fmt.Sprintf("Removal of channel %s failed with status %d", channelID, resp.StatusCode))
+	}
```
tock-ibm: In case of a failed request that returns an error status code, the response contains an informative error message. It is better to print it as well. This relates to all commands. See 
https://docs.google.com/document/d/1kYoicS_Z59gNvmuuB7VFZ6WsOL31C34dWMeOFQiIygA/edit?usp=sharing

cmd/participation/join.go
```diff
@@ -0,0 +1,68 @@
...
+	bodyBytes, err := ioutil.ReadAll(resp.Body)
+	if err != nil {
+		panic(err)
+	}
+	resp.Body.Close()
+
+	err = printResponseAsJSON(bodyBytes, &channelInfo{}, os.Stdout)
+	if err != nil {
+		panic(err)
+	}
```
tock-ibm: What about checking the status to detect a failed request, and printing the body if it is a failure?

cmd/participation/list.go
```diff
@@ -0,0 +1,49 @@
...
+	client := httpClient(tlsDir)
+	bodyBytes, err := getBodyBytes(client, url)
+	if err != nil {
+		panic(err)
+	}
+
+	err = printResponseAsJSON(bodyBytes, &channelList{}, os.Stdout)
+	if err != nil {
+		panic(err)
+	}
```
tock-ibm: What about checking the status to detect a failed request, and printing the body if it is a failure?

cmd/participation/list.go
```diff
@@ -0,0 +1,49 @@
...
+
+	if bodyBytes == nil {
+		fmt.Printf("channel '%s' not found\n", channelID)
+		return
+	}
+
+	err = printResponseAsJSON(bodyBytes, &channelInfo{}, os.Stdout)
+	if err != nil {
+		panic(err)
+	}
```
tock-ibm: What about checking the status to detect a failed request, and printing the body if it is a failure?

cmd/participation/main.go
```diff
@@ -0,0 +1,169 @@
...
+}
+
+func getBodyBytes(client *http.Client, url string) ([]byte, error) {
+	resp, err := client.Get(url)
+	if err != nil {
+		return nil, err
+	}
+	if resp.StatusCode != http.StatusOK {
+		return nil, nil
+	}
```
tock-ibm: In case of a failed request that returns an error status code, the response contains an informative error message. It is better to print it as well.

cmd/participation/list.go
```diff
@@ -0,0 +1,49 @@
...
+	client := httpClient(tlsDir)
+	bodyBytes, err := getBodyBytes(client, url)
+	if err != nil {
+		panic(err)
+	}
+
+	if bodyBytes == nil {
+		fmt.Printf("channel '%s' not found\n", channelID)
+		return
+	}
```
tock-ibm: There is a specific error code that captures this. I think it is better to follow these.

cmd/participation/join.go
```diff
@@ -0,0 +1,68 @@
...
+	bodyBytes, err := ioutil.ReadAll(resp.Body)
+	if err != nil {
+		panic(err)
+	}
+	resp.Body.Close()
+
+	err = printResponseAsJSON(bodyBytes, &channelInfo{}, os.Stdout)
```
jyellick: In general, it's a little odd to have the console printing pushed out into the leaf of the calls.  I'd generally expect for a response and error to be propagated back up to the top, where, uniform decisions can be made about how to report success and failure, the output format, etc.

cmd/participation/main.go
```diff
@@ -0,0 +1,169 @@
...
+	listCmd.StringVar(&channelID, "channelID", "", "Channel ID")
+
+	removeCmd := flag.NewFlagSet("remove", flag.ExitOnError)
+	removeCmd.StringVar(&osn, "orderer", "", "Ordering service endpoint")
+	removeCmd.StringVar(&channelID, "channelID", "", "Channel ID")
+	removeCmd.StringVar(&tlsDir, "tlsDir", "", "Path to directory containing TLS materials")
+
+	joinCmd := flag.NewFlagSet("join", flag.ExitOnError)
+	joinCmd.StringVar(&osn, "orderer", "", "Ordering service endpoint")
+	joinCmd.StringVar(&tlsDir, "tlsDir", "", "Path to directory containing TLS materials")
```
jyellick: The contents of this directory should be validated before we proceed with the invocations.  Going back to a previous comment, we should do all of our parsing and setup early and in a consolidated place.

cmd/participation/main.go
```diff
@@ -0,0 +1,169 @@
...
+	bodyBytes, err := ioutil.ReadAll(resp.Body)
+	if err != nil {
+		return nil, errors.Wrap(err, "reading http response body")
+	}
+	resp.Body.Close()
+
+	return bodyBytes, nil
+}
+
+func printResponseAsJSON(bodyBytes []byte, response interface{}, out io.Writer) error {
```
jyellick: It's not the end of the world, but we have a JSON string, and we're turning it back into a map structure, then turning it back into JSON.  You could instead use vanilla indent https://golang.org/pkg/encoding/json/#Indent

jyellick (commented):
Looks like the function is here, but I think it could use some refactoring, and definitely use some tests.  I've added some comments of my own in addition to the good set from Yoav.

cmd/participation/remove.go
wlahti: Refactored the code. We now print the status and error generated by the server for all commands.

cmd/participation/join.go
wlahti: Refactored. Done.

cmd/participation/list.go
wlahti: Refactored. Done.

cmd/participation/list.go
wlahti: Refactored. Done.

cmd/participation/main.go
wlahti: Refactored. Done.

cmd/participation/list.go
wlahti: Refactored. Done.

cmd/participation/join.go
```diff
@@ -0,0 +1,68 @@
...
+	"mime/multipart"
+	"net/http"
+	"os"
+)
+
+// Joins a OSN to a new or existing channel.
+func join(osn, tlsDir, channelID, configBlockPath string) {
+	blockBytes, err := ioutil.ReadFile(configBlockPath)
+	if err != nil {
+		panic(err)
```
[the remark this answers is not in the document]
wlahti: Refactored. Done.

cmd/participation/join.go
wlahti: Refactored. Done.

cmd/participation/main.go
wlahti: Refactored. Done.

wlahti:
I did a pretty extensive refactoring of the code here. Still need to address the comments around flags and directory validation as well as add unit tests. More to come shortly.

cmd/participation/main.go
```diff
@@ -0,0 +1,84 @@
...
+		fmt.Printf("Error: %s\n", err)
+	}
+
+	printResponse(resp, os.Stdout)
+}
+
+func printResponse(resp *http.Response, out io.Writer) {
+	bodyBytes, err := readBodyBytes(resp.Body)
+	if err != nil {
+		log.Fatalf("failed to read http response body: %s", err)
```
wlahti: Not sure this is the "correct" thing to do here but for now I figured if we can't read the response body something has gone very wrong.

wlahti:
I've updated this CLI to use kingpin and added some unit tests around flag parsing. 

I think more unit tests are probably need but I'm a little confused on how to actually implement them. I don't see an obvious example to base my tests on. Should I create an http endpoint with a fake channel participation API that sends back faked responses or is that beyond the scope of UT here?

mastersingh24 (dismissed):
I'd really like to understand why we are doing this?   I apologize if I missed this as part of the RFC, but we really do not need another standalone CLI.

The beauty of REST APIs is that there are tons of code generators for virtually any language which can generate a client from Swagger.

I don't think we should merge this.

wlahti:
Added a fairly comprehensive set of unit tests now, which also helped me notice that we weren't supporting the various TLS configurations. This CLI now supports TLS, mutual TLS, and TLS disabled. I also moved away from the TLS directory flag with expected file names (an artifact of the test code we used as a basis for the http client code) to a "--caFile" flag (when the server is using TLS) and "--clientCert" + "--clientKey" flags (when the server is using mutual TLS).

jyellick (changes requested):
Generally looking pretty good -- test coverage looks nice.

I do think with a (small) rework of the main, you could get rid of all the forking, and session reading from your test.  And, since the Config struct is not a parameter anywhere outside of the main, it doesn't really need to be exported anymore (or exist).

The biggest thing though that I caught on this pass, is that as written, this CLI is incompatible with intermediate CAs.  I see this as a must-fix, although it should be pretty easy to do.  Assuming we go the route of allowing the user to concatenate PEMs into a single file, the golang x509 CertPool will happily parse it with no additional work.  Then it's just a matter of wiring the cert pool through.

cmd/osnadmin/main.go
```diff
@@ -0,0 +1,95 @@
...
+	"io/ioutil"
+	"net/http"
+	"os"
+
+	"github.com/hyperledger/fabric/internal/osnadmin"
+	"gopkg.in/alecthomas/kingpin.v2"
+)
+
+// command line flags
+var (
```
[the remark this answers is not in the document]
wlahti: Excellent idea. Done.

cmd/osnadmin/main.go
```diff
@@ -0,0 +1,95 @@
...
+	app        = kingpin.New("osnadmin", "Orderer Service Node (OSN) administration")
+	orderer    = app.Flag("orderer", "Endpoint of the OSN").String()
+	caFile     = app.Flag("caFile", "Path to the PEM-encoded orderer root CA certificate").String()
```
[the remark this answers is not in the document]
wlahti: Still need to look into this...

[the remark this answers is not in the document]
wlahti: Still need to look into this...

internal/osnadmin/httpclient.go
```diff
@@ -0,0 +1,44 @@
...
+package osnadmin
+
+import (
+	"crypto/tls"
+	"crypto/x509"
+	"net/http"
+)
+
+func httpClient(tlsClientCert *tls.Certificate, tlsCACert *x509.Certificate) *http.Client {
+	clientCertPool := x509.NewCertPool()
```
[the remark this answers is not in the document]
wlahti: Still need to look into this...

cmd/osnadmin/main_test.go
```diff
@@ -0,0 +1,668 @@
...
+		})
+	})
+
+	Describe("Flags", func() {
+		Context("when the ca cert cannot be read", func() {
+			BeforeEach(func() {
+				ordererCACert = "not-the-ca-cert-youre-looking-for"
+			})
+
+			It("returns with exit code 1 and prints the error", func() {
```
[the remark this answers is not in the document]
wlahti: Still need to look into this...

cmd/osnadmin/main_test.go
```diff
@@ -0,0 +1,668 @@
...
+
+func runCommand(args []string, expectedExitCode int) *gexec.Session {
+	cmd := exec.Command(cliPath, args...)
+	sess, err := gexec.Start(cmd, nil, nil)
+	Expect(err).NotTo(HaveOccurred())
+	Eventually(sess, 5*time.Second).Should(gexec.Exit(expectedExitCode))
+	return sess
+}
+
+func checkOutput(sess *gexec.Session, expectedStatus int, expectedOutput interface{}) {
```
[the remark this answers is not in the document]
wlahti: Done

cmd/osnadmin/main_test.go
```diff
@@ -0,0 +1,668 @@
...
+		})
+
+		It("uses the channel participation API to list all application and and the system channel (when it exists)", func() {
+			args := []string{
+				"channel",
+				"list",
+				"--orderer", urlString,
+				"--caFile", ordererCACert,
+			}
+			sess := runCommand(args, 0)
```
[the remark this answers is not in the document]
wlahti: Done.

jyellick (approved):
Looks good to me!

PR #1908, opened 2020-09-22
Implement interface SnapshotPvtdataHashesConsumer for pvtdata store

manish-sethi:
#### Type of change
- New feature

#### Description
This PR prepares the ground for populating private data store from snapshot files.

#### Additional details
In the followup PR(s), 
- The hashed data snapshot importer (introduced in this PR) will be integrated with the path where the hashed data is loaded in to statedb
- The reconciler will be integrated to work with the hashed data populated into the store (instead of from the blockstore) for the data below the snapshot height


core/ledger/pvtdatastorage/store.go
```diff
@@ -577,6 +584,10 @@ func (s *Store) purgeExpiredData(minBlkNum, maxBlkNum uint64) error {
 			)
 		}
 
+		for _, bootKVHashesKey := range bootKVHashesKeys {
+			batch.Delete(encodeBootKVHashesKey(bootKVHashesKey))
+		}
```
cendhu, line 589: good to add a test for this.

core/ledger/pvtdatastorage/snapshot_data_importer.go
```diff
@@ -0,0 +1,327 @@
...
+			)
+	}
+
+	if dataBlockNum <= history[len(history)-1].configBlockNum {
+		return false,
+			errors.Errorf(
+				"unexpected error - no collection config found below block number [%d] for <namespace=%s, collection=%s>",
+				dataBlockNum, namespace, collection,
+			)
+	}
```
cendhu, line 205: Are we storing all eligibility history to tackle this error case?

core/ledger/pvtdatastorage/snapshot_data_importer_test.go
```diff
@@ -0,0 +1,1034 @@
...
+
+	t.Run("ineligible-then-eligible", func(t *testing.T) {
+		eligibilityBTLCache.eligibilityHistory = map[nsColl][]*eligibility{
+			{ns: "ns", coll: "coll"}: {
+				{configBlockNum: 10, isEligible: true},
+				{configBlockNum: 5, isEligible: false},
+			},
+		}
+
+		for i := 6; i < 15; i++ {
```
cendhu, line 633: even for 1 to 5, this test would pass, right?

cendhu (commented):
Thanks, Manish. A few minor nits and some clarification questions.

core/ledger/pvtdatastorage/snapshot_data_importer_test.go
manish-sethi: No, it should expect an error for that range - `unexpected error - no collection config found below block number ...`

cendhu (approved):
LGTM. Thanks, Manish.

PR #1910, opened 2020-09-22
[FAB-18234] Add joinbysnapshotstatus CLI command

wenjianqiao:

#### Type of change
- New feature

#### Description
Add joinbysnapshotstatus CLI command to check if a previously submitted joinbysnapshot operation is still running.

#### Related issues
https://jira.hyperledger.org/browse/FAB-18234

### DIFF
--- a/core/peer/peer.go
+++ b/core/peer/peer.go
@@ -478,6 +478,11 @@ func (p *Peer) GetPolicyManager(cid string) policies.Manager {
 	return nil
 }
 
+// JoinBySnaphotStatus queries ledger mgr to get the status of joinbysnapshot
+func (p *Peer) JoinBySnaphotStatus() *pb.JoinBySnapshotStatus {
+	return p.LedgerMgr.JoinBySnapshotStatus()
+}
+
 // initChannel takes care to initialize channel after peer joined, for example deploys system CCs
 func (p *Peer) initChannel(cid string) {
 	if p.channelInitializer != nil {
--- a/internal/peer/channel/joinbysnapshot.go
+++ b/internal/peer/channel/joinbysnapshot.go
@@ -55,5 +55,10 @@ func joinBySnapshot(cmd *cobra.Command, args []string, cf *ChannelCmdFactory) er
 		Input:       &pb.ChaincodeInput{Args: [][]byte{[]byte(cscc.JoinChainBySnapshot), []byte(snapshotPath)}},
 	}
 
-	return executeJoin(cf, spec)
+	if err = executeJoin(cf, spec); err != nil {
+		return err
+	}
+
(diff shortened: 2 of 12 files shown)


internal/peer/channel/joinbysnapshotstatus_test.go
```diff
@@ -0,0 +1,72 @@
...
+	mockResponse.Response.Payload = []byte("")
+	status, err := client.joinBySnapshotStatus()
+	require.NoError(t, err)
+	require.Equal(t, "", status)
+
+	mockResponse.Response.Payload = []byte("mock_snapshot_directory")
+	status, err = client.joinBySnapshotStatus()
+	require.NoError(t, err)
+	require.Equal(t, "mock_snapshot_directory", status)
+
```
manish-sethi: It would have been better to cover the full output of the command by passing the writer from outside instead of using fmt.Println in the command code. But, it seems that all the commands in this package follow this pattern. To take a note of this, I think that it would be better if in a future PR, the command factory can be enhanced to pass the io.writer for this purpose.

internal/peer/channel/joinbysnapshot.go
```diff
@@ -55,5 +55,10 @@ func joinBySnapshot(cmd *cobra.Command, args []string, cf *ChannelCmdFactory) er
 		Input:       &pb.ChaincodeInput{Args: [][]byte{[]byte(cscc.JoinChainBySnapshot), []byte(snapshotPath)}},
 	}
 
-	return executeJoin(cf, spec)
+	if err = executeJoin(cf, spec); err != nil {
+		return err
+	}
+
+	logger.Info("The joinbysnapshot operation is in progress. Call 'peer channel joinbysnapshotstatus' to check the status of the operation.")
```
manish-sethi: I don;t think that hard-coding the actual command here is a good idea.

internal/peer/channel/joinbysnapshotstatus_test.go
wenjianqiao: Agree. Will open a jira for this improvement.

core/peer/peer.go
```diff
@@ -478,6 +478,12 @@ func (p *Peer) GetPolicyManager(cid string) policies.Manager {
 	return nil
 }
 
+// GetCreateFromSnapshotInfo queries ledger mgr to get information for any in-progress ledger creation
+// from a snapshot.
+func (p *Peer) GetCreateFromSnapshotInfo() string {
+	return p.LedgerMgr.GetCreateFromSnapshotInfo()
+}
+
```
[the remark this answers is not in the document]
wenjianqiao: Done

wenjianqiao:
Hit CI flakde FAB-17653

manish-sethi (approved):

PR #1911, opened 2020-09-23
Private data tutorial edits

nikhil550:
Signed-off-by: NIKHIL E GUPTA <ngupta@symbridge.com>

#### Type of change

- Documentation update

#### Description

Use the deployCC script to deploy the private data smart contract. The tutorial is already quite long, so this change will make life easier for users. 

Also made some minor edits

### DIFF
--- a/docs/source/private_data_tutorial.rst
+++ b/docs/source/private_data_tutorial.rst
@@ -23,8 +23,8 @@ configuring and using private data with Fabric:
 #. :ref:`pd-register-identities`
 #. :ref:`pd-store-private-data`
 #. :ref:`pd-query-authorized`
-#. :ref:`pd-purge`
 #. :ref:`pd-transfer-asset`
+#. :ref:`pd-purge`
 #. :ref:`pd-indexes`
 #. :ref:`pd-ref-material`
 
@@ -37,31 +37,30 @@ You should have completed the task :doc:`install`.
 
 Asset transfer private data sample use case
 -------------------------------------------
+
 This sample demonstrates the use of three private data collections, ``assetCollection``, ``Org1MSPPrivateCollection`` & ``Org2MSPPrivateCollection`` to transfer an asset between Org1 and Org2, using following use case:
 
 A member of Org1 creates a new asset, henceforth referred as owner. The public details of the asset,
 including the owner, are stored in the private data collection named ``assetCollection``. The asset is also created with an appraised
 value supplied by the owner. The appraised value is used by each participant to agree to the transfer of the asset, and is only stored in owner organization's collection. In our case, the initial appraisal value agreed by the owner is stored in the ``Org1MSPPrivateCollection``.
 
-To purchase the asset, the buyer needs to agree to the same appraised value as the asset owner. In this step, the buyer (a member of Org2) creates an agreement to trade and agree to an appraisal value using smart contract function ``'AgreeToTransfer'``. This value is stored in ``Org2MSPPrivateCollection`` collection.
+To purchase the asset, the buyer needs to agree to the same appraised value as the asset owner. In this step, the buyer (a member of Org2) creates an agreement to trade and agree to an appraisal value using smart contract function ``'AgreeToTransfer'``. This value is stored in ``Org2MSPPrivateCollection`` collection. Now, the asset owner can transfer the asset to the buyer using smart contract function ``'TransferAsset'``, which checks for the conditions that must be met for transfer to succeed.
 
-Now, the asset owner can transfer the asset to the buyer using smart contract function ``'TransferAsset'``, which checks for the conditions that must be met for transfer to succeed.
+Before we go through the transfer scenario, we will discuss how organizations can use private data collections in Fabric.
 
 .. _pd-build-json:
 
 Build a collection definition JSON file
 ---------------------------------------
 
-The first step in privatizing data on a channel is to build a collection
-definition which defines access to the private data.
+Before a set of organizations can transact using private data, all organizations
+on channel need to build a collection file that defines the private
+data collections associated with each chaincode. Data that is stored in a private
+data collection is only distributed to the peers of certain organizations instead
+of all members of the channel. The collection definition file describes all of the
+private data collections that organizations can read and write to from a chaincode.
 
-The collection definition describes who can persist data, how many peers the
-data is distributed to, how many peers are required to disseminate the private
-data, and how long the private data is persisted in the private database. Later,
-we will demonstrate how chaincode APIs ``PutPrivateData`` and ``GetPrivateData``
-are used to map the collection to the private data being secured.
-
-A collection definition is composed of the following properties:
+Each collection is defined by the following properties:
 
 .. _blockToLive:
 
@@ -92,7 +91,19 @@ A collection definition is composed of the following properties:
   enforce that only clients belonging to one of the collection member organizations
   are allowed write access to private data.
 
-To illustrate usage of private data, the asset transfer private data example contains
+- ``endorsementPolicy``: defines the endorsement policy that needs to be met in
+  order to write to the private data collection. The collection level endorsement policy
+  overrides to chaincode level policy. For more information on building a policy
+  definition refer to the :doc:`endorsement-policies` topic.
+
+The same collection definition file needs to be deployed by all organizations that
+use the chaincode, even if the organization does not belong to any collections. In
+addition to the collections that are explicitly defined in a collection file,
+each organization has access to an implicit collection on their peers that can only
+be read by their organization. For an example that uses implicit data collections,
+see the :doc:`secured_asset_transfer/secured_private_asset_transfer_tutorial`.
+
+To illustrate the use of private data, the asset transfer private data example contains
 three private data collection definitions: ``assetCollection``, ``Org1MSPPrivateCollection``,
 and ``Org2MSPPrivateCollection``. The ``policy`` property in the
 ``assetCollection`` definition allows all members of the channel (Org1 and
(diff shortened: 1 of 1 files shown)


denali49:
While I agree that the scripted approach makes life easier, it may be defeating the purpose of a 'tutorial' even if it is a bit long.  Most maintainers work with Fabric every day and the mundane bits can be dispensed with, but for the new user exposing them to the 'how' is beneficial to learning Fabric.  At the very least, we should give the user the option of taking the easy way out with the script, or going through the long tutorial.  My two cents.

nikhil550:
> While I agree that the scripted approach makes life easier, it may be defeating the purpose of a 'tutorial' even if it is a bit long. Most maintainers work with Fabric every day and the mundane bits can be dispensed with, but for the new user exposing them to the 'how' is beneficial to learning Fabric. At the very least, we should give the user the option of taking the easy way out with the script, or going through the long tutorial. My two cents.

I am sympathetic to this argument. That is why I left the commands to deploy the collection configuration along with the script, and linked to the chaincode lifecycle topic. There is a separate tutorial that helps users use the cc lifecycle. Note that the other tutorials in this series use the deployment script as well.

My worry is that the long tutorial will discourage users, and that they will not get to the private data parts of the tutorial. I also think that keeping the documentation modular will help users. They can use the private data tutorial for private data, couchdb for couchDB, and the deployment tutorial for lifecycle. I would be interested in what others think on this.

docs/source/private_data_tutorial.rst
```diff
@@ -434,202 +418,45 @@ use indexes with private data.
...
-The command will return the same package identifier as the install command.
-You should see output similar to the following:
-
-.. code:: bash
+   ./network.sh deployCC -ccn private -ccep "OR('Org1MSP.peer','Org2MSP.peer')" -cccg ../asset-transfer-private-data/chaincode-go/collections_config.json
 
-    Installed chaincodes on peer:
-    Package ID: privatev1:c10ad5913bd3166c1bcc294bb378ecef5f815b516b717f98d8a68ff2f6f4eee9, Label: privatev1
+Note that we need to pass the path to the private data collection definition file
+to the command. As part of the deploying the chaincode to the channel, both organizations
```
denyeart:
```suggestion
to the command. As part of deploying the chaincode to the channel, both organizations
```

docs/source/private_data_tutorial.rst
```diff
@@ -434,202 +418,45 @@ use indexes with private data.
...
+on the channel need to agree deploy the private data collection as part
```
denyeart:
```suggestion
on the channel must pass identical private data collection definitions as part 
```

docs/source/private_data_tutorial.rst
```diff
@@ -1083,6 +881,79 @@ If you go back and look at the block height in your logs, you will see that the
...
 because when the asset transfer occurred, Org1 did not meet the ``read`` policy of the ``Org2MSPPrivateCollection``.
 
+
+.. _pd-purge:
+
+Purge Private Data
+------------------
+
+For use cases where private data only needs to be on the ledger until it can be
+replicated into an off-chain database, it is possible to "purge" the data after
```
denyeart: replication to an off-chain database is one use case, but there are other use cases. So perhaps say something more general like, e.g.
"In some cases private data only needs to be persisted for a short time, for example while a multi-party business process is in progress, or until the data is replicated into an off-chain database"...

docs/source/private_data_tutorial.rst
```diff
@@ -1083,6 +881,79 @@ If you go back and look at the block height in your logs, you will see that the
...
+a certain set number of blocks, leaving behind only hash of the data that serves
```
denyeart:
```suggestion
a certain set number of blocks, leaving behind only a hash of the data that serves
```

docs/source/private_data_tutorial.rst
```diff
@@ -1083,6 +881,79 @@ If you go back and look at the block height in your logs, you will see that the
...
+as immutable evidence of the transaction.
+
+There may be private data including personal or confidential
+information, such as the ``appraisedValue`` data in our example, that the transacting
+parties don't want disclosed to other organizations on the channel. Thus, it
```
denyeart: Private data in a collection is never disclosed to non-member peers, so I wouldn't say that is the reason to purge it.
Focus on the limited lifespan aspect.  The appraisedValue scenario is good, as you may want some time limit on the agreement (of course, number of blocks is a very rough way to specify time).

docs/source/private_data_tutorial.rst
```diff
@@ -1083,6 +881,79 @@ If you go back and look at the block height in your logs, you will see that the
...
+has a limited lifespan, and can be purged after existing unchanged on the
+blockchain for a designated number of blocks using the ``blockToLive`` property
+in the collection definition.
+
+The ``Org2MSPPrivateCollection`` definition has a ``blockToLive``
+property value of ``3``, meaning this data will live on the side database for
+three blocks and then after that it will get purged. If we create additional
+blocks on the channel, the appraisedValue agreed to by Org2 will eventually
+get purged. We can create 3 new blocks to demostrate:
```
denyeart:
```suggestion
get purged. We can create 3 new blocks to demonstrate:
```

denyeart (commented):
I like how you handle the lifecycle Nik... agreed that the detailed steps don't need to be repeated in every tutorial, and I like how you still discussed the important aspect around collection definition during approval and commit.

denyeart:
@nikhil550 @denali49 I think there are some aspects mentioned in the [fabric-samples readme scenario description ](https://github.com/hyperledger/fabric-samples/tree/master/asset-transfer-private-data/chaincode-go#private-data-asset-transfer-scenario)that are not yet covered in the tutorial. Could you verify that all the important information from the readme is included in the tutorial, and then we can replace the readme content with a link to the doc tutorial.

pamandrejko:
We discussed this in our Community call this morning. 

One argument for keeping the lifecycle steps in this tutorial is that you must provide the private data collection during that process. A newbie could overlook that part of the deployment process so it seems worthwhile including the steps in this tutorial so that it is very clear. This is the one case where that part of the lifecycle process is actually required.

docs/source/private_data_tutorial.rst
```diff
@@ -37,23 +37,24 @@ You should have completed the task :doc:`install`.
...
 
 .. _pd-build-json:
 
 Build a collection definition JSON file
 ---------------------------------------
 
-The first step in privatizing data on a channel is to build a collection
-definition which defines access to the private data.
+Before a set of organizations can transact using private data on a channel, all organizations
+on channel need to build a collection definition which defines access to the private data.
```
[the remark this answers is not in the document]
nikhil550: Rephrasing: Before a set of organizations can transact using private data, all organizations
on channel need to build a collection file that defines the private
data collections associated with each chaincode. Data that is stored in a private
data collection is only distributed to the peers of certain organizations instead
of all members of the channel. The collection definition file describes all of the
private data collections that organizations can read and write to from a chaincode.

further down:The same collection definition file needs to be deployed by all organizations that
use the chaincode, even if the organization does not belong to any collections. In
addition to the collections that are explicitly defined in a collection file,
each organization has access to an implicit collection on their peers that can only
be read by their organization. For an example that uses implicit data collections,
see the :doc:`secured_asset_transfer/secured_private_asset_transfer_tutorial`.

docs/source/private_data_tutorial.rst
nikhil550: done.

docs/source/private_data_tutorial.rst
nikhil550: done

docs/source/private_data_tutorial.rst
nikhil550: For use cases where private data only needs to be persisted for a short period of time,
it is possible to "purge" the data after a certain set number of blocks, leaving
behind only hash of the data that serves as immutable evidence of the transaction.
An organization could decide to purge private data if the data contained sensitive
information that was used by another transaction, but is not longer needed, or
if the data is being replicated into an off-chain database.

docs/source/private_data_tutorial.rst
nikhil550: The ``appraisedValue`` data in our example contains a private agreement that
the organization may want to expire after a certain period of time. Thus, it
has a limited lifespan, and can be purged after existing unchanged on the
blockchain for a designated number of blocks using the ``blockToLive`` property
in the collection definition.

docs/source/private_data_tutorial.rst
nikhil550: done

docs/source/private_data_tutorial.rst
nikhil550: done.

nikhil550:
I can do this in a follow on PR sometime this week.

> @nikhil550 @denali49 I think there are some aspects mentioned in the [fabric-samples readme scenario description ](https://github.com/hyperledger/fabric-samples/tree/master/asset-transfer-private-data/chaincode-go#private-data-asset-transfer-scenario)that are not yet covered in the tutorial. Could you verify that all the important information from the readme is included in the tutorial, and then we can replace the readme content with a link to the doc tutorial.

Most of the information in the readme is in the tutorial. However, some of the information about which members can read and write to each collection is not in the tutorial. Most importantly, the use of the `GetPrivateDataHash()` to verify private data is not in the tutorial and should be added.

denyeart (approved):

denyeart:
> We discussed this in our Community call this morning.
> 
> One argument for keeping the lifecycle steps in this tutorial is that you must provide the private data collection during that process. A newbie could overlook that part of the deployment process so it seems worthwhile including the steps in this tutorial so that it is very clear. This is the one case where that part of the lifecycle process is actually required.

@pamandrejko Nik's update includes the approve and commit commands that reference the collection definition file. This is the important aspect for people to understand.  I think highlighting only these commands is the best way to highlight that, rather than it getting buried in the full set of lifecycle commands.  The lifecycle tutorial shows the full set of lifecycle commands. This private data tutorial shows the delta difference for private data chaincodes.

PR #1912, opened 2020-09-23
[FAB-18237] always update stateInfo message upon chaincode update

yacovm:
With the change of FAB-18208, gossip saves CPU cycles by only updating
metadata information if peer membership is not empty.

However, this makes discovery incapable of operating in case
anchor/bootstrap peers are not defined, or, in case
the peer is the only peer in the network.

This change makes gossip always update its information, including
signing the stateInfo message, whenever it receives a chaincode update.

Change-Id: Iaf04441944a3fe5d889233d35bdcb1173c31e2bc

### DIFF
--- a/gossip/gossip/channel/channel.go
+++ b/gossip/gossip/channel/channel.go
@@ -414,6 +414,11 @@ func (gc *gossipChannel) publishStateInfo() {
 		return
 	}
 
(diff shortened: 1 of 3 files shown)


C0rWin (approved):
cherry pick from #1913

PR #1921, opened 2020-09-24
[FAB-18242] Remove GetSessionInfo Call

lindluni:
The GetSessionInfo call was added to the getSession function which retrieves a session from the pool. The call was used to validate the state of the session before using it to ensure it was still healthy. Unfortunately the call is exceedingly expensive with preliminary tests showing nearly a second to perform the single operation. With a transaction require requiring several signings it is possible for it to take several seconds for a transaction just to retrieve all of it signatures.

This change reverts the call and in future implementations we can look into other ways of verifying if a session has become stale, perhaps by analyzing failed transaction error messages and ejecting the session on transaction failures.

### DIFF
--- a/bccsp/pkcs11/pkcs11.go
+++ b/bccsp/pkcs11/pkcs11.go
@@ -303,14 +303,7 @@ func (csp *Provider) getSession() (session pkcs11.SessionHandle, err error) {
 	for {
 		select {
 		case session = <-csp.sessPool:
-			if _, err = csp.ctx.GetSessionInfo(session); err == nil {
-				logger.Debugf("Reusing existing pkcs11 session %d on slot %d\n", session, csp.slot)
-				return session, nil
-			}
-
-			logger.Warningf("Get session info failed [%s], closing existing session and getting a new session\n", err)
-			csp.closeSession(session)
-
+			return
 		default:
 			// cache is empty (or completely in use), create a new session
 			return csp.createSession()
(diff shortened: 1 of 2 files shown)


lindluni:
Need to remove an earlier test that pollutes the session pool with bad sessions

yacovm:
> This change reverts the call and in future implementations we can look into other ways of verifying if a session has become stale, perhaps by analyzing failed transaction error messages and ejecting the session on transaction failures.

Perhaps by looking at the [strerror](https://github.com/miekg/pkcs11/blob/287d9350987cc9334667882061e202e96cdfb4d0/error.go#L54-L61)?  Seems like there is a finite list of error codes and if you map them into 2 groups - session problems and the rest, then you can know if to create a new session or just go with the error?

lindluni:
> > This change reverts the call and in future implementations we can look into other ways of verifying if a session has become stale, perhaps by analyzing failed transaction error messages and ejecting the session on transaction failures.
> 
> Perhaps by looking at the [strerror](https://github.com/miekg/pkcs11/blob/287d9350987cc9334667882061e202e96cdfb4d0/error.go#L54-L61)? Seems like there is a finite list of error codes and if you map them into 2 groups - session problems and the rest, then you can know if to create a new session or just go with the error?

That's the plan, there are just lots of places these checks need to happen. So we decided to pull it out entirely for 2.2.1 and 1.4.9 (which was just a revert to the pre 2.0.0 and 1.4.7 implementation) and then address the invalid sessions afterwards since we release tomorrow.

mastersingh24 (approved):

PR #1923, opened 2020-09-24
[FAB-18242] Remove GetSessionInfo Call

lindluni:
The GetSessionInfo call was added to the getSession
function which retrieves a session from the pool.
The call was used to validate the state of the session
before using it to ensure it was still healthy.
Unfortunately the call is exceedingly expensive with
preliminary tests showing nearly a second to perform the
single operation. With a transaction require requiring
several signings it is possible for it to take several
seconds for a transaction just to retrieve all of it signatures.

This change reverts the call and in future implementations
we can look into other ways of verifying if a session has
become stale, perhaps by analyzing failed transaction error
messages and ejecting the session on transaction failures.

### DIFF
--- a/bccsp/pkcs11/pkcs11.go
+++ b/bccsp/pkcs11/pkcs11.go
@@ -67,14 +67,7 @@ func (csp *impl) getSession() (session pkcs11.SessionHandle, err error) {
 	for {
 		select {
 		case session = <-csp.sessPool:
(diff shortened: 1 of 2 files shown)


lindluni:
Need to remove an earlier test that pollutes the session pool with bad sessions

mastersingh24 (approved):

PR #1935, opened 2020-09-28
[FAB-18250] Check Error Before Returning Session to Pool

lindluni:
FAB-17722 introduced logic for evicting invalid sessions from
the pkcs11 session pool. The GetSessionInfo call was expensive,
taking nearly a full second per call. The change was reverted
in FAB-18242 due to the extreme performance regression.

This change introduces a new method for evicting sessions by
checking the error message returned when pkcs11 call fail
and closing sessions that resulted in SESSION errors.

### DIFF
--- a/bccsp/pkcs11/pkcs11.go
+++ b/bccsp/pkcs11/pkcs11.go
@@ -15,6 +15,7 @@ import (
 	"fmt"
 	"math/big"
 	"os"
+	"regexp"
 	"sync"
 	"time"
 
@@ -28,6 +29,7 @@ import (
 )
 
 var logger = flogging.MustGetLogger("bccsp_p11")
+var regex = regexp.MustCompile(".*0xB.:\\sCKR.+")
 
 type Provider struct {
 	bccsp.BCCSP
@@ -374,7 +376,7 @@ func (csp *Provider) getECKey(ski []byte) (pubKey *ecdsa.PublicKey, isPriv bool,
 	if err != nil {
 		return nil, false, err
 	}
-	defer csp.returnSession(session)
+	defer func() { csp.handleSessionReturn(err, session) }()
 
 	isPriv = true
 	_, err = csp.findKeyPairFromSKI(session, ski, privateKeyType)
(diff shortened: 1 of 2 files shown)


lindluni:
I think there are some layers here that can be peeled back in future PR to clean up the call tree. Will try to unwind some of that in a future PR.

bccsp/pkcs11/pkcs11.go
```diff
@@ -800,6 +817,19 @@ func (csp *Provider) ecPoint(session pkcs11.SessionHandle, key pkcs11.ObjectHand
 	return ecpt, oid, nil
 }
 
+func (csp *Provider) handleSessionReturn(err error, session pkcs11.SessionHandle) {
+	if err != nil {
+		for _, sessionErr := range sessionErrors {
+			if strings.Contains(err.Error(), sessionErr) {
```
yacovm: why not extract the error code from the string, e.g: "0xB1:" and then do a map lookup?

yacovm: Also, it seems that the only errors you care about, and all the errors you care about, [start with a B](https://github.com/miekg/pkcs11/blob/287d9350987cc9334667882061e202e96cdfb4d0/error.go#L54-L61), so why not just assert that it is the case?

bccsp/pkcs11/pkcs11.go
```diff
@@ -800,6 +801,18 @@ func (csp *Provider) ecPoint(session pkcs11.SessionHandle, key pkcs11.ObjectHand
 	return ecpt, oid, nil
 }
 
+func (csp *Provider) handleSessionReturn(err error, session pkcs11.SessionHandle) {
+	if err != nil {
+		re := regexp.MustCompile(".*0xB.:\\sCKR.+")
```
yacovm: Perhaps compile the regexp once in a global variable?

lindluni: Fixed

yacovm (approved):

bccsp/pkcs11/pkcs11.go
```diff
@@ -374,7 +376,7 @@ func (csp *Provider) getECKey(ski []byte) (pubKey *ecdsa.PublicKey, isPriv bool,
 	if err != nil {
 		return nil, false, err
 	}
-	defer csp.returnSession(session)
+	defer func() { csp.handleSessionReturn(err, session) }()
```
yacovm, line 379: Another option that might be cleaner, is that the **csp.getSession()** returns a cleanup callback according to the error status- if the error matches the regexp, the cleanup callback is a no-op. otherwise, it is the **closeSesson**.

mastersingh24 (commented):
I'm good with this .... given we already had the returnSession function in any case, this seems to be a good way to add the check more or less transparently.

lindluni:

PR #1939, opened 2020-09-28
Expose new Admin endpoint on orderer for the channel participation API

wlahti:
#### Type of change

- New feature

#### Description

First commit
- split out common system logic for http endpoints from existing operations system code

Second commit
- expose new Admin endpoint on the orderer and move the channel participation API to this endpoint

#### Related issues

[FAB-18246](https://jira.hyperledger.org/browse/FAB-18246)


sampleconfig/orderer.yaml
```diff
@@ -334,6 +334,35 @@ Metrics:
...
       Prefix:
 
+################################################################################
+#
+#   Admin Configuration
+#
+#   - This configures the admin server endpoint for the orderer
+#
+################################################################################
+Operations:
```
yacovm: I'm confused, why do you have another operations endpoint, [exactly like the existing one](https://github.com/hyperledger/fabric/blob/master/sampleconfig/orderer.yaml#L288-L309)?

Aren't both endpoints supposed to be used by users possessing local admin credentials? Why do you need a separate port, isn't it possible to use the existing one?

wlahti: My understanding is that the role using this admin endpoint to join/leave channels and the role using the operations endpoint to access metrics, change log levels. etc. are distinct and should be configurable separately.

wlahti:
Opened a new flake, [FAB-18251](https://jira.hyperledger.org/browse/FAB-18251), for the earlier flake (and pushed a likely fix: https://github.com/hyperledger/fabric/pull/1941). Current flake is [FAB-16233](https://jira.hyperledger.org/browse/FAB-16233). Grabbed the logs for that one and I'll take a look to see if anything jumps out.

common/system/system.go
```diff
@@ -0,0 +1,144 @@
...
+	Warnf(template string, args ...interface{})
+}
+
+type Options struct {
+	Logger        Logger
+	ListenAddress string
+	TLS           TLS
+}
+
+type System struct {
```
jyellick: This name stutters, and I'm generally a little confused by the choice of package name.  As a reader I would struggle to know what a `system.System` logically represents.  I'd invite feedback for a better name... the more natural one would be `http.Server` or similar, but there is the obvious mental conflict with the stdlib.  Maybe something like `fabhttp.Server` or `mtlshttp.Server` or... hopefully someone else has a better thought, but I really don't love `system.System`.

common/system/system.go
```diff
@@ -0,0 +1,144 @@
...
+	options    Options
+	httpServer *http.Server
+	mux        *http.ServeMux
+	addr       string
+}
+
+func NewSystem(o Options) *System {
+	logger := o.Logger
+	if logger == nil {
+		logger = flogging.MustGetLogger("system.runner")
```
jyellick: Why is this `system.runner` instead of simply the package name? What value does `runner` add?

orderer/common/admin/system.go
```diff
@@ -0,0 +1,143 @@
...
+}
+
+type Options struct {
+	Logger        Logger
+	ListenAddress string
+	TLS           TLS
+	Version       string
+}
+
+type System struct {
```
jyellick: Wasn't the whole point of factoring the operations.System out into system.System so that you could re-use it? This seems like entirely duplicated code?

sampleconfig/orderer.yaml
jyellick: Yes, the key here is that the identity performing operations is not a Fabric admin, it's a system admin.  We would not want the identity used to perform health check, pull metrics, or modify log level to be the same identity responsible for joining/leaving channels.  This is something that @sykesm brought up during the RFC review process and a new port is the simplest option going forward.  In the future, the plan is to bring some non-MTLS authentication mechanism in, but that it outside the scope of the channel participation effort.

common/system/system.go
wlahti: Agreed. Of those options, fabhttp.Server seems more obvious to me in terms of describing what it is.

orderer/common/admin/system.go
wlahti: Hmmm, I'm honestly not sure what happened with the code here. It looks like I messed things up when I split things into two commits. I'll get this back to what I intended and push an update...

wlahti: Oh right, this was my first pass making sure I understood the common code that was needed before splitting it out into a separate package. I'll remove this dead code.

common/system/system.go
wlahti: I was following what's there in the operations system code. I'm not opposed to renaming this.

common/system/system.go
wlahti: Renamed to fabhttp.Server for now.

orderer/common/admin/system.go
wlahti: Removed.

sampleconfig/orderer.yaml
yacovm: > Yes, the key here is that the identity performing operations is not a Fabric admin, it's a system admin. We would not want the identity used to perform health check, pull metrics, or modify log level to be the same identity responsible for joining/leaving channels.

You can still accomplish this with a single port. You can allow both types of connections and then classify the client by its certificate.

integration/nwo/orderer_template.go
```diff
@@ -121,6 +121,17 @@ Metrics:
     {{- end }}
     WriteInterval: 5s
     Prefix: {{ ReplaceAll (ToLower Orderer.ID) "." "_" }}
+Admin:
+  ListenAddress: 127.0.0.1:{{ .OrdererPort Orderer "Admin" }}
+  TLS:
+    Enabled: {{ .TLSEnabled }}
+    PrivateKey: {{ $w.OrdererLocalTLSDir Orderer }}/server.key
+    Certificate: {{ $w.OrdererLocalTLSDir Orderer }}/server.crt
```
tock-ibm, line 129: Does this (using `OrdererLocalTLSDir`) mean that in all the tests we share the crypto material of the Operations and Admin endpoints?
Is this what we want? 
Is there a testing benefit for separating `OrdererLocalTLSDir` from say, `OrdererAdminTLSDir`?

tock-ibm:
Update commit message to reflect package names after last commit.

wlahti:
> Update commit message to reflect package names after last commit.

Thanks for catching this. Noticed it last week but forgot to get back to updating it.

tock-ibm:
Looks good to me.

tock-ibm (dismissed):

sampleconfig/orderer.yaml
wlahti: Ah, thanks for clarifying, Yoav. Definitely overlooked this. Fixed.

common/system/system.go
wlahti: Renamed.

orderer/common/server/main.go
```diff
@@ -895,6 +900,20 @@ func newOperationsSystem(ops localconfig.Operations, metrics localconfig.Metrics
...
+func newAdminServer(ops localconfig.Admin) *fabhttp.Server {
+	return fabhttp.NewServer(fabhttp.Options{
+		Logger:        flogging.MustGetLogger("orderer.admin"),
+		ListenAddress: ops.ListenAddress,
+		TLS: fabhttp.TLS{
+			Enabled:            ops.TLS.Enabled,
+			CertFile:           ops.TLS.Certificate,
+			KeyFile:            ops.TLS.PrivateKey,
+			ClientCertRequired: ops.TLS.ClientAuthRequired,
+			ClientCACertFiles:  ops.TLS.ClientRootCAs,
```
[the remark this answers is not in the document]
wlahti: Fixed.

wlahti:
Just a rebase on master since this has been sitting around a while.

common/fabhttp/server.go
```diff
@@ -0,0 +1,144 @@
...
+		logger:  logger,
+		options: o,
+	}
+
+	server.initializeServer()
+
+	return server
+}
+
+func (s *Server) Run(signals <-chan os.Signal, ready chan<- struct{}) error {
```
C0rWin, line 59: I might be missed that, but where do you use this method? It's seems unused.

wlahti: This allows us to run as an ifrit.Runner. It's currently used in tests via ifrit.Invoke. I preserved this functionality as it was added by Matt when he implemented the operations service.

C0rWin: Ok, thanks for the clarification.

C0rWin (approved):

PR #1952, opened 2020-09-30
[FAB-18082] Make RemoveChannel crash fault tolerant

stephyee:
#### Type of change

- New feature

#### Description

Implement filestore to make remove operation crash fault tolerant

#### Related issues

[FAB-18082](https://jira.hyperledger.org/browse/FAB-18082)

### DIFF
--- a/common/ledger/blockledger/fileledger/factory.go
+++ b/common/ledger/blockledger/fileledger/factory.go
@@ -7,11 +7,14 @@ SPDX-License-Identifier: Apache-2.0
 package fileledger
 
 import (
+	"os"
+	"path/filepath"
 	"sync"
 
 	"github.com/hyperledger/fabric/common/ledger/blkstorage"
 	"github.com/hyperledger/fabric/common/ledger/blockledger"
 	"github.com/hyperledger/fabric/common/metrics"
+	"github.com/hyperledger/fabric/orderer/common/filerepo"
 )
 
 //go:generate counterfeiter -o mock/block_store_provider.go --fake-name BlockStoreProvider . blockStoreProvider
@@ -26,6 +29,7 @@ type fileLedgerFactory struct {
 	blkstorageProvider blockStoreProvider
 	ledgers            map[string]*FileLedger
 	mutex              sync.Mutex
+	removeFileRepo     *filerepo.Repo
 }
 
 // GetOrCreate gets an existing ledger (if it exists) or creates it
@@ -62,12 +66,20 @@ func (f *fileLedgerFactory) Remove(channelID string) error {
 		ledger.blockStore.Shutdown()
 	}
 
+	if err := f.removeFileRepo.Save(channelID, []byte{}); err != nil && err != os.ErrExist {
+		return err
+	}
+
 	err := f.blkstorageProvider.Drop(channelID)
 	if err != nil {
 		return err
 	}
 
 	delete(f.ledgers, channelID)
+
+	if err := f.removeFileRepo.Remove(channelID); err != nil {
+		return err
+	}
 
 	return nil
 }
@@ -97,8 +109,30 @@ func New(directory string, metricsProvider metrics.Provider) (blockledger.Factor
 	if err != nil {
 		return nil, err
 	}
-	return &fileLedgerFactory{
+
+	fileRepo, err := filerepo.New(filepath.Join(directory, "filerepo"), "remove")
+	if err != nil {
+		return nil, err
+	}
+
+	factory := &fileLedgerFactory{
 		blkstorageProvider: p,
 		ledgers:            map[string]*FileLedger{},
-	}, nil
+		removeFileRepo:     fileRepo,
+	}
+
+	files, err := factory.removeFileRepo.List()
+	if len(files) != 0 {
+		for _, fileName := range files {
+			channelID := factory.removeFileRepo.FileToBaseName(fileName)
+			err = factory.Remove(channelID)
+			if err != nil {
+				logger.Errorf("Failed to remove channel %s: %s", channelID, err.Error())
+				return nil, err
+			}
+			logger.Infof("Removed channel: %s", channelID)
+		}
+	}
+
+	return factory, nil
 }
--- a/common/ledger/blockledger/fileledger/factory_test.go
+++ b/common/ledger/blockledger/fileledger/factory_test.go
@@ -8,12 +8,15 @@ package fileledger
 
 import (
 	"errors"
+	"fmt"
 	"io/ioutil"
 	"os"
+	"path/filepath"
 	"testing"
 
 	"github.com/hyperledger/fabric/common/ledger/blockledger/fileledger/mock"
 	"github.com/hyperledger/fabric/common/metrics/disabled"
+	"github.com/hyperledger/fabric/orderer/common/filerepo"
 	"github.com/stretchr/testify/require"
 )
 
@@ -24,17 +27,19 @@ type fileLedgerBlockStore interface {
 }
 
 func TestBlockStoreProviderErrors(t *testing.T) {
-	setup := func() (*fileLedgerFactory, *mock.BlockStoreProvider) {
+	setup := func(fileRepo *filerepo.Repo) (*fileLedgerFactory, *mock.BlockStoreProvider) {
 		m := &mock.BlockStoreProvider{}
+
 		f := &fileLedgerFactory{
 			blkstorageProvider: m,
 			ledgers:            map[string]*FileLedger{},
+			removeFileRepo:     fileRepo,
 		}
 		return f, m
 	}
 
 	t.Run("list", func(t *testing.T) {
-		f, mockBlockStoreProvider := setup()
+		f, mockBlockStoreProvider := setup(nil)
 		mockBlockStoreProvider.ListReturns(nil, errors.New("boogie"))
 		require.PanicsWithValue(
 			t,
@@ -45,7 +50,7 @@ func TestBlockStoreProviderErrors(t *testing.T) {
 	})
 
 	t.Run("open", func(t *testing.T) {
-		f, mockBlockStoreProvider := setup()
+		f, mockBlockStoreProvider := setup(nil)
 		mockBlockStoreProvider.OpenReturns(nil, errors.New("woogie"))
 		_, err := f.GetOrCreate("foo")
 		require.EqualError(t, err, "woogie")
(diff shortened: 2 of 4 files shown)


wlahti (commented):
Thanks, Tiffany! Looking good overall. A few minor requests and one nit for your consideration.

common/ledger/blockledger/fileledger/factory.go
```diff
@@ -90,8 +101,28 @@ func New(directory string, metricsProvider metrics.Provider) (blockledger.Factor
...
+		return nil, err
+	}
+
+	files, err := fileRepo.List()
+	if len(files) != 0 {
+		for _, fileName := range files {
+			channelName := fileRepo.FileToBaseName(fileName)
+			err := fileRepo.Remove(channelName)
+			if err != nil {
+				logger.Errorf("error removing ledger for channel %s", channelName)
```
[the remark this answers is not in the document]
stephyee: 🧹

common/ledger/blockledger/fileledger/factory_test.go
```diff
@@ -77,6 +87,9 @@ func TestMultiReinitialization(t *testing.T) {
 	require.Equal(t, 3, len(f.ChannelIDs()), "Expected channel to be recovered")
 	f.Close()
 
+	_, err = os.Create(dir + "/remove/bar2.remove")
```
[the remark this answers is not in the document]
stephyee: 🧹

common/ledger/blockledger/fileledger/factory_test.go
```diff
@@ -87,3 +100,33 @@ func TestMultiReinitialization(t *testing.T) {
...
+
+	mockBlockStore.DropReturns(nil)
+	err = f.Remove("foo")
+	require.NoError(t, err, "Error removing channel")
+	require.Equal(t, 2, mockBlockStore.DropCallCount(), "Expected 2 Drop() calls")
+
+	os.RemoveAll(dir)
+	mockBlockStore.DropReturns(nil)
+	err = f.Remove("foo")
+	require.EqualError(t, err, fmt.Sprintf("error while creating file:%s/remove/foo.remove~: open %s/remove/foo.remove~: no such file or directory", dir, dir))
```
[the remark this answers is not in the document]
stephyee: 🧹

common/ledger/blockledger/fileledger/factory.go
```diff
@@ -90,8 +101,28 @@ func New(directory string, metricsProvider metrics.Provider) (blockledger.Factor
 	if err != nil {
 		return nil, err
 	}
+
+	fileRepo, err := filerepo.New(directory, "remove")
+	if err != nil {
+		return nil, err
+	}
```
tock-ibm: This deviates from the pattern we already have in here: 

https://github.com/hyperledger/fabric/blob/9adde2e7d30e1cb8778f8dbb564e08bc49afb70d/orderer/common/multichannel/registrar.go#L120

where we put all different file-repos under the ledger directory in a "filerepo" folder. The "remove" will be another subfolder, along side the "joinblock" file-repo. 
This way we'll have 
```
- ledgerDir
    - chains
        - mychain
        - yourchain
    - filerepo
        - joinblock
        - remove
```

common/ledger/blockledger/fileledger/factory.go
```diff
@@ -90,8 +101,28 @@ func New(directory string, metricsProvider metrics.Provider) (blockledger.Factor
...
+
+	files, err := fileRepo.List()
+	if len(files) != 0 {
+		for _, fileName := range files {
+			channelName := fileRepo.FileToBaseName(fileName)
+			err := fileRepo.Remove(channelName)
```
tock-ibm: You have to actually complete removing the channel here, with the equivalent of
```go
err := f.blkstorageProvider.Drop(channelID)
if err != nil {
   return err
}

delete(f.ledgers, channelID)
```
and only then remove the filerepo marker. The fact that we have a marker here means the `Remove` did not complete before the server crashed.

common/ledger/blockledger/fileledger/factory_test.go
```diff
@@ -77,6 +88,9 @@ func TestMultiReinitialization(t *testing.T) {
 	require.Equal(t, 3, len(f.ChannelIDs()), "Expected channel to be recovered")
 	f.Close()
 
+	_, err = os.Create(filepath.Join(dir, "remove", "bar2.remove"))
+	require.NoError(t, err, "Error creating temp file: %s", err)
+
```
tock-ibm: This is not enough. Create a `bar2` channel dir as well, and see that the failed `Drop()` actually is repeated after restart. The end result should be that both the channel dir `bar2` and the `bar2.remove` marker file are removed.

common/ledger/blockledger/fileledger/factory_test.go
```diff
@@ -87,3 +101,38 @@ func TestMultiReinitialization(t *testing.T) {
...
+		ledgers:            map[string]blockledger.ReadWriter{},
+		removeFileRepo:     fileRepo,
+	}
+
+	t.Run("success", func(t *testing.T) {
+		mockBlockStore.DropReturns(nil)
+		err = f.Remove("foo")
+		require.NoError(t, err, "Error removing channel")
+		require.Equal(t, 1, mockBlockStore.DropCallCount(), "Expected 1 Drop() calls")
+	})
```
tock-ibm: This does not check the existence of a `foo.remove` file. You could use `mockBlockStore.DropCalls()` instead, and check that the respective marker file was created prior to the call to `Drop`.
You  should then require that the  `foo.remove` file is gone, after `Remove` returns.

common/ledger/blockledger/fileledger/factory_test.go
```diff
@@ -87,3 +101,38 @@ func TestMultiReinitialization(t *testing.T) {
...
+
+	t.Run("drop fails", func(t *testing.T) {
+		mockBlockStore.DropReturns(errors.New("oogie"))
+		err = f.Remove("foo")
+		require.EqualError(t, err, "oogie")
+	})
```
tock-ibm: If drop fails, the `foo.remove` remains. you have to check for that.

common/ledger/blockledger/fileledger/factory.go
stephyee: 🧹

common/ledger/blockledger/fileledger/factory_test.go
stephyee: 🧹

common/ledger/blockledger/fileledger/factory_test.go
stephyee: 🧹

common/ledger/blockledger/fileledger/factory.go
```diff
@@ -97,8 +109,30 @@ func New(directory string, metricsProvider metrics.Provider) (blockledger.Factor
...
+
+	files, err := factory.removeFileRepo.List()
+	if len(files) != 0 {
+		for _, fileName := range files {
+			channelID := factory.removeFileRepo.FileToBaseName(fileName)
+			err = factory.Remove(channelID)
+			if err != nil {
+				logger.Errorf("error removing ledger for channel %s: %s", channelID, err.Error())
+				continue
+			}
```
tock-ibm: If we retry the "Remove" and fail, do we really want to continue? Depending on what the failure was, there is a chance the channel will be detected and initialized, but then since it went through a botched remove, it might fail with panic when the server starts to run it. Or start operating with an inconsistent state.

It seems that in this case the right thing to do is to return an error rather than suppress it. 

I think that the operator can then intervene, analyze the problem and fix it.
For example she can delete the channel's folder and the index manually and restart.

stephyee: I think you are right, as such I've updated this logic to return the error.

common/ledger/blockledger/fileledger/factory.go
```diff
@@ -97,8 +109,30 @@ func New(directory string, metricsProvider metrics.Provider) (blockledger.Factor
...
+	}
+
+	files, err := factory.removeFileRepo.List()
+	if len(files) != 0 {
+		for _, fileName := range files {
+			channelID := factory.removeFileRepo.FileToBaseName(fileName)
+			err = factory.Remove(channelID)
+			if err != nil {
+				logger.Errorf("error removing ledger for channel %s: %s", channelID, err.Error())
+				return nil, err
```
tock-ibm: I suggest
```go
logger.Errorf("Failed to remove channel %s: %s", channelID, err.Error())
```
because it is not only the ledger removal that may cause this error to be thrown.

In addition, there is no unit test coverage for when this happens. This can be triggered by changing the permission on one of the files to be deleted, for example.

tock-ibm: What if you create a filerepo/remove folder with mychannel.remove in it, then shut everything down, and then change the permission on filerepo/remove, and then restart? I think this flow down his error path, no?
In a separate test case? and cleanup afterwards?

stephyee: 🧹  The missing piece was including execute permissions on the parent directory

common/ledger/blockledger/fileledger/factory_test.go
stephyee: 🧹

common/ledger/blockledger/fileledger/factory.go
stephyee: 🧹

tock-ibm (approved):

tock-ibm:
Looks good

common/ledger/blockledger/fileledger/factory.go
```diff
@@ -97,8 +109,30 @@ func New(directory string, metricsProvider metrics.Provider) (blockledger.Factor
 	if err != nil {
 		return nil, err
 	}
-	return &fileLedgerFactory{
+
+	fileRepo, err := filerepo.New(filepath.Join(directory, "filerepo"), "remove")
```
jyellick, line 113: I'm guessing this is not being introduced here.... but "filerepo" seems almost uselessly vague.  Something like "pendingops" or something more descriptive could be nice.  (No need to change here, but, if agreed, perhaps open a JIRA?)

jyellick (approved):
One comment, but no need to address here.

common/ledger/blockledger/fileledger/factory.go
stephyee: Opened https://jira.hyperledger.org/browse/FAB-18280

PR #1957, opened 2020-10-01
Revert and re-apply prepare for next

denyeart:
The what's new doc link to v1.4.9 release notes was incorrect but was fixed in the subsequent prepare for next commit.
This PR includes two commits
1) revert the prepare for next (v1.4.10) updates so that we can get a clean commit with the link fix.
2) re-apply the prepare for next (v1.4.10).

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -44,8 +44,8 @@
 #   - docker-tag-stable - re-tags the images made by 'make docker' with the :stable tag
 #   - help-docs - generate the command reference docs
 
-BASE_VERSION = 1.4.10
-PREV_VERSION = 1.4.9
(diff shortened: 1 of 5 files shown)


lindluni (approved):

denyeart:
We may not merge this... we may keep the "release-1.4" doc approach instead of the "v1.4.9" doc approach so that we can continue to update docs in real time without requiring a certain tag.

denyeart:
We will go back to a release branch approach for docs (e.g. "release-1.4"), rather than a tag approach (e.g. "v1.4.9") so that doc updates on release branches go live right away, without having to wait for the next tagged version.

PR #1958, opened 2020-10-01
[FAB-17953] Add integration tests for snapshot feature (leveldb, no pvtdata)

wenjianqiao:

#### Type of change
- New feature

#### Description
Add basic end to end tests for the snapshot feature:
- generate snapshots and compare snapshots generated on different peers
- bootstrap peer from a snapshot for peers in an existing org and new org
- generate snapshot again on a peer bootstrapped from a snapshot
- verify bootstrapped peer: chaincode invocation and upgrade, channel config update,
  join new channel, qscc, history query, new chaincode deployment

It covers leveldb without private data. Tests for couchdb and private data will be covered in separate PRs.

#### Related issues
https://jira.hyperledger.org/browse/FAB-17953

### DIFF
--- a/integration/ledger/snapshot_test.go
+++ b/integration/ledger/snapshot_test.go
@@ -1,0 +1,716 @@
+/*
+Copyright IBM Corp All Rights Reserved.
+
+SPDX-License-Identifier: Apache-2.0
+*/
+
+package ledger
+
+import (
+	"encoding/base64"
+	"encoding/json"
+	"fmt"
+	"io/ioutil"
+	"os"
+	"path/filepath"
+	"strconv"
+	"strings"
+	"syscall"
+	"time"
+
+	docker "github.com/fsouza/go-dockerclient"
+	"github.com/hyperledger/fabric-protos-go/common"
+	"github.com/hyperledger/fabric/integration/chaincode/kvexecutor"
+	"github.com/hyperledger/fabric/integration/nwo"
+	"github.com/hyperledger/fabric/integration/nwo/commands"
+	"github.com/hyperledger/fabric/integration/runner"
+	"github.com/hyperledger/fabric/protoutil"
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+	"github.com/onsi/gomega/gbytes"
+	"github.com/onsi/gomega/gexec"
+	"github.com/tedsuo/ifrit"
+)
+
+const (
+	testchannelID = "testchannel"
+)
+
+var _ bool = Describe("Snapshot Generation and Bootstrap", func() {
+	var (
+		setup                 *setup
+		helper                *marblesTestHelper
+		couchProcess          []ifrit.Process
+		legacyChaincode       nwo.Chaincode
+		newlifecycleChaincode nwo.Chaincode
+	)
+
+	BeforeEach(func() {
+		By("initializing and starting the network")
+		setup = initAndStartMultiOrgsNetwork()
+
+		helper = &marblesTestHelper{
+			networkHelper: &networkHelper{
+				Network:   setup.network,
+				orderer:   setup.orderer,
+				peers:     setup.peers,
+				testDir:   setup.testDir,
+				channelID: setup.channelID,
+			},
+		}
+
+		legacyChaincode = nwo.Chaincode{
+			Name:        "marbles",
+			Version:     "0.0",
+			Path:        chaincodePathWithIndex,
+			Ctor:        `{"Args":[]}`,
+			Policy:      `OR ('Org1MSP.member','Org2MSP.member','Org3MSP.member','Org4MSP.member')`,
+			PackageFile: filepath.Join(setup.testDir, "marbles_legacy.tar.gz"),
+		}
+
+		newlifecycleChaincode = nwo.Chaincode{
+			Name:            "marbles",
+			Version:         "0.0",
+			Path:            components.Build(chaincodePathWithIndex),
+			Lang:            "binary",
+			CodeFiles:       filesWithIndex,
+			PackageFile:     filepath.Join(setup.testDir, "marbles.tar.gz"),
+			SignaturePolicy: `OR ('Org1MSP.member','Org2MSP.member','Org3MSP.member','Org4MSP.member')`,
+			Sequence:        "1",
+			Label:           "marbles",
+		}
+
+		By("deploying legacy chaincode on initial peers (stateleveldb)")
+		nwo.DeployChaincodeLegacy(setup.network, testchannelID, setup.orderer, legacyChaincode)
+	})
+
+	AfterEach(func() {
+		setup.cleanup()
+		for _, proc := range couchProcess {
+			proc.Signal(syscall.SIGTERM)
+			Eventually(proc.Wait(), setup.network.EventuallyTimeout).Should(Receive())
+		}
+		os.RemoveAll(setup.network.RootDir)
+	})
+
+	// Below test does the following when peers are using either leveldb or couchdb.
+	// Note that we do not support a channel with mixed DBs. However, for testing,
+	// it would be fine to use mixed DBs to test with both couchdb and leveldb
+	// - create snapshots on the 2 peers and verify they are same
+	// - bootstrap a peer (couchdb) in an existing org from the snapshot
+	// - bootstrap a peer (leveldb) in a new org from the snapshot
+	// - verify couchdb index exists
+	// - verify chaincode invocation, history, qscc, channel config update
+	// - upgrade to new lifecycle chaincode
+	// - create a new snapshot from a peer (couchdb) bootstrapped from a snapshot
+	// - bootstrap peers (couchdb) in existing org and new org from the new snapshot
+	// - verify couchdb index exists
+	// - verify chaincode invocation, history, qscc, channel config update
+	// - verify chaincode upgrade and new chaincode install on all peers
+	It("generates snapshot and bootstraps from snapshots with statecouchdb", func() {
+		org1peer0 := setup.network.Peer("Org1", "peer0")
+		org2peer0 := setup.network.Peer("Org2", "peer0")
+
+		By("invoking marbles chaincode")
+		testKey := "marble-0"
+		helper.invokeMarblesChaincode(legacyChaincode.Name, org1peer0, "initMarble", "marble-0", "blue", "35", "tom")
+		helper.invokeMarblesChaincode(legacyChaincode.Name, org1peer0, "initMarble", "marble-1", "red", "100", "tom")
+
+		By("getting an txid from a block before snapshot is generated")
+		txidBeforeSnapshot := getTxIDFromLastBlock(setup.network, org1peer0)
+
+		// verify snapshot commands with different parameters")
+		blockNum := nwo.GetLedgerHeight(setup.network, org1peer0, testchannelID) - 1
+		verifySnapshotRequestCmds(setup.network, org1peer0, testchannelID, blockNum)
+
+		// test 1: generate snapshots on 2 peers for the same blockNum and verify they are same
+		_, snapshotDir2 := generateAndCompareSnapshots(setup.network, org1peer0, org2peer0, blockNum)
+
+		// test 2: bootstrap a peer in an existing org from snapshot and verify
+		By("starting new peer org2peer1 in existing org2 (couchdb)")
+		org2peer1, couchProc := startPeer(setup, "Org2", "peer1", testchannelID, true)
+		couchProcess = append(couchProcess, couchProc)
+
+		By("installing legacy chaincode on new peer org2peer1")
+		nwo.InstallChaincodeLegacy(setup.network, legacyChaincode, org2peer1)
+
+		By("joining new peer org2peer1 to the channel")
+		joinBySnapshot(setup.network, setup.orderer, org2peer1, testchannelID, snapshotDir2, blockNum)
+
+		By("verifying index created on org2peer1")
+		verifySizeIndexExists(setup.network, testchannelID, setup.orderer, org2peer1, "marbles")
+
+		By("invoking marbles chaincode on bootstrapped peer org2peer1")
+		helper.invokeMarblesChaincode(legacyChaincode.Name, org2peer1, "transferMarble", testKey, "newowner2")
+
+		By("verifying history on peer org2peer1")
+		expectedHistory := []*marbleHistoryResult{
+			{IsDelete: "false", Value: newMarble(testKey, "blue", 35, "newowner2")},
+		}
+		helper.assertGetHistoryForMarble(legacyChaincode.Name, org2peer1, expectedHistory, testKey)
+
+		verifyQSCC(setup.network, org2peer1, testchannelID, blockNum, txidBeforeSnapshot)
+
+		// test 3: bootstrap a peer in a new org from snapshot and verify
+		By("starting a peer Org3.peer0 in new org3 (stateleveldb)")
+		org3peer0, _ := startPeer(setup, "Org3", "peer0", testchannelID, false)
+
+		By("installing legacy chaincode on new peer org3peer0")
+		nwo.InstallChaincodeLegacy(setup.network, legacyChaincode, org3peer0)
+
+		By("joining new peer org3peer0 to the channel")
+		joinBySnapshot(setup.network, setup.orderer, org3peer0, testchannelID, snapshotDir2, blockNum)
+
+		By("invoking marbles chaincode on bootstrapped peer org3peer0")
+		helper.invokeMarblesChaincode(legacyChaincode.Name, org3peer0, "transferMarble", testKey, "newowner3")
+
+		By("verifying history on peer org3peer0")
+		expectedHistory = []*marbleHistoryResult{
+			{IsDelete: "false", Value: newMarble(testKey, "blue", 35, "newowner3")},
+			{IsDelete: "false", Value: newMarble(testKey, "blue", 35, "newowner2")},
+		}
+		helper.assertGetHistoryForMarble(legacyChaincode.Name, org3peer0, expectedHistory, testKey)
+
+		verifyQSCC(setup.network, org3peer0, testchannelID, blockNum, txidBeforeSnapshot)
+
+		// test 4: upgrade legacy chaincode to new lifecycle
+		By("enabling V2_0 capabilities")
+		channelPeers := setup.network.PeersWithChannel(testchannelID)
+		nwo.EnableCapabilities(setup.network, testchannelID, "Application", "V2_0", setup.orderer, channelPeers...)
+
+		By("upgrading legacy chaincode to new lifecycle chaincode")
+		nwo.DeployChaincode(setup.network, testchannelID, setup.orderer, newlifecycleChaincode, channelPeers...)
+
+		By("invoking chaincode after upgraded to new lifecycle chaincode")
+		helper.invokeMarblesChaincode(legacyChaincode.Name, org1peer0, "initMarble", "marble-upgrade", "blue", "35", "tom")
+
+		// test 5: generate snapshot again on a peer bootstrapped from a snapshot and upgraded to new lifecycle chaincode
+		blockNumForNextSnapshot := nwo.GetLedgerHeight(setup.network, org2peer1, testchannelID)
+		By(fmt.Sprintf("generating a snapshot at blockNum %d on org2peer1 that was bootstrapped by a snapshot", blockNumForNextSnapshot))
+		submitSnapshotRequest(setup.network, testchannelID, blockNumForNextSnapshot, org2peer1, false, "Snapshot request submitted successfully")
+
+		// invoke chaincode to trigger snapshot generation
+		// 1st call should be committed before snapshot generation, 2nd call should be committed after snapshot generation
+		helper.invokeMarblesChaincode(legacyChaincode.Name, org1peer0, "transferMarble", testKey, "newowner_beforesnapshot")
+		helper.invokeMarblesChaincode(legacyChaincode.Name, org1peer0, "transferMarble", testKey, "newowner_aftersnapshot")
+
+		By("verifying snapshot completed on org2peer1")
+		verifyNoPendingSnapshotRequest(setup.network, org2peer1, testchannelID)
+		snapshotDir := filepath.Join(setup.network.PeerLedgerDir(org2peer1), "snapshots/completed", testchannelID, strconv.Itoa(blockNumForNextSnapshot))
+
+		// test 6: bootstrap a peer in a different org from the new snapshot
+		By("starting a peer (org1peer1) in existing org1 (couchdb)")
+		org1peer1, couchProc := startPeer(setup, "Org1", "peer1", testchannelID, true)
+		couchProcess = append(couchProcess, couchProc)
+
+		By("installing new lifecycle chaincode on peer org1peer1")
+		nwo.InstallChaincode(setup.network, newlifecycleChaincode, org1peer1)
+
+		By("joining new peer org1peer1 to the channel")
+		joinBySnapshot(setup.network, setup.orderer, org1peer1, testchannelID, snapshotDir, blockNumForNextSnapshot)
+
+		By("verifying index created on org1peer1")
+		verifySizeIndexExists(setup.network, testchannelID, setup.orderer, org1peer1, "marbles")
+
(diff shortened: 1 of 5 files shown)


integration/nwo/network.go
```diff
@@ -1115,6 +1115,31 @@ func (n *Network) JoinChannel(name string, o *Orderer, peers ...*Peer) {
 	}
 }
 
+func (n *Network) JoinChannelBySnapshot(snapshotDir string, o *Orderer, peers ...*Peer) {
```
manish-sethi: unused param `o *Orderer`

integration/nwo/network.go
```diff
@@ -1115,6 +1115,31 @@ func (n *Network) JoinChannel(name string, o *Orderer, peers ...*Peer) {
...
+		sess, err := n.PeerAdminSession(p, commands.ChannelJoinBySnapshot{
+			SnapshotPath: snapshotDir,
+			ClientAuth:   n.ClientAuthRequired,
+		})
+		Expect(err).NotTo(HaveOccurred())
+		Eventually(sess, n.EventuallyTimeout).Should(gexec.Exit(0))
+	}
+}
+
+func (n *Network) JoinBySnapshotStatus(o *Orderer, p *Peer) []byte {
```
manish-sethi: unused param `o *Orderer`

integration/ledger/snapshot_test.go
```diff
@@ -0,0 +1,1049 @@
...
+
+			By("deploying legacy chaincode on peers with stateleveldb")
+			nwo.DeployChaincodeLegacy(setup.network, testchannelID, setup.orderer, legacyChaincode)
+
+			height := helper.getLedgerHeight(org1peer0)
+			By(fmt.Sprintf("adding marbles at height %d", height))
+			testKey := "marble-0"
+			channelPeers := setup.network.PeersWithChannel(testchannelID)
+			helper.invokeMarblesChaincode(legacyChaincode.Name, org1peer0, "initMarble", "marble-0", "blue", "35", "tom")
+			nwo.WaitUntilEqualLedgerHeight(setup.network, testchannelID, height+1, channelPeers...)
```
manish-sethi: Here and many other places - doesn't the function invokeMarblesChaincode waits for commit on all the peers?

integration/ledger/snapshot_test.go
```diff
@@ -0,0 +1,1049 @@
...
+			helper.invokeMarblesChaincode(legacyChaincode.Name, org1peer0, "initMarble", "marble-1", "blue", "35", "tom")
+			nwo.WaitUntilEqualLedgerHeight(setup.network, testchannelID, height+2, channelPeers...)
+
+			By("getting txid from a block before snapshot is generated")
+			txid := getTxIDFromLastBlock(setup.network, org1peer0)
```
manish-sethi: This appears a round-about way to get the TxID, it will be preferred to create a signed proposal explicitly - (I believe that there is a function for this). Perhaps, you can modify the function "invokeMarblesChaincode" to returns the TxID or have another variation of this funciton.

integration/ledger/snapshot_test.go
```diff
@@ -0,0 +1,1049 @@
...
+			_, snapshotDir2 := generateAndCompareSnapshots(setup.network, setup.orderer, org1peer0, org2peer0, legacyChaincode, blockNum)
+
+			// test 2: bootstrap a peer in an existing org
+			By("bootstrapping a peer (org2peer1) in existing org from a snapshot")
+			org2peer1, _ := bootstrapPeerInExistingOrgFromSnapshot(setup, "Org2", "peer1", testchannelID, snapshotDir2, blockNum, false)
+
+			By("installing legacyChaincode on org2peer1")
+			nwo.InstallChaincodeLegacy(setup.network, legacyChaincode, org2peer1)
+
+			availableHistoryResult := verifyPeerAfterJoinBySnapshot(helper, org2peer1, testchannelID, &legacyChaincode, "", blockNum, txid, testKey, []*marbleHistoryResult{})
```
manish-sethi: - In order to reduce the number of parameters to this function, can you construct a struct (say expectedState) that encapsulates the test data related parameters.

- I guess that it will be better to more readable if you remove the return from this function and always supply the expected history explicitly rather than recycling it - see related comment in this function.

integration/ledger/snapshot_test.go
```diff
@@ -0,0 +1,1049 @@
...
+			couchProcess = append(couchProcess, couchProc)
+
+			By("installing legacyChaincode on org2peer1")
+			nwo.InstallChaincodeLegacy(setup.network, legacyChaincode, org2peer1)
+
+			availableHistoryResult := verifyPeerAfterJoinBySnapshot(helper, org2peer1, testchannelID, &legacyChaincode, "", blockNum, txid, testKey, []*marbleHistoryResult{})
+
+			// TODO:
+			// Index does not exist right after InstallChaincodeLegacy, is this expected?
+			// On the other hand, for a peer in new org, index exists right after InstallChaincodeLegacy
```
manish-sethi: No, this is not expected... In fact, I would rather suggest to install the chaincode first and then bootstrap in this case so both the sequences are tested.

integration/ledger/snapshot_test.go
```diff
@@ -0,0 +1,1049 @@
...
+	By("submitting snaphost request for a future blockNum, expecting success")
+	submitSnapshotRequest(n, channel, blockNum+10, peer, false, "Snapshot request submitted successfully")
+	By("submitting snaphost request at same blockNum again, expecting error")
+	submitSnapshotRequest(n, channel, blockNum+10, peer, true,
+		fmt.Sprintf("duplicate snapshot request for block number %d", blockNum+10))
+	By("submitting snaphost request for a previous blockNum, expecting error")
+	submitSnapshotRequest(n, channel, blockNum-1, peer, true,
+		fmt.Sprintf("requested snapshot for block number %d cannot be less than the last committed block number %d", blockNum-1, blockNum))
+	By("listing pending snaphost requests, expecting success")
+	listPendingSnapshotRequests(n, channel, peer, n.PeerAddress(peer, nwo.ListenPort), false)
```
manish-sethi:
```
pendingRequests := listPendingSnapshotRequests(...)
Expect(pendingRequests).To(Equal("....."))
```

integration/ledger/snapshot_test.go
```diff
@@ -0,0 +1,1049 @@
...
+func workDir(dir, pattern string) (string, error) {
+	trypath := filepath.Join(dir, pattern)
+	os.RemoveAll(trypath)
+	err := os.Mkdir(trypath, 0700)
+	return trypath, err
+}
+
+// verifySnapshotRequestCmds invokes snapshot commands and verify the expected rersults.
+// At the end, there will be no pending request.
+func verifySnapshotRequestCmds(n *nwo.Network, peer *nwo.Peer, channel string, blockNum int) {
```
manish-sethi, line 406: Do we need so many `By`s in this function... it compromises with the readability. The same comment about other similar functions as well

integration/ledger/snapshot_test.go
```diff
@@ -0,0 +1,1049 @@
...
+	By("canceling a pending snaphost request, expecting success")
+	cancelSnapshotRequest(n, channel, blockNum+10, peer, n.PeerAddress(peer, nwo.ListenPort), false, "Snapshot request cancelled successfully")
+	By("canceling the same snaphost request, expecting error")
+	cancelSnapshotRequest(n, channel, blockNum+10, peer, n.PeerAddress(peer, nwo.ListenPort), true,
+		fmt.Sprintf("no snapshot request exists for block number %d", blockNum+10))
+	By("listing pending snaphost requests, expecting success")
+	listPendingSnapshotRequests(n, channel, peer, n.PeerAddress(peer, nwo.ListenPort), false)
+}
+
+func generateAndCompareSnapshots(n *nwo.Network, orderer *nwo.Orderer, peer1, peer2 *nwo.Peer, chaincode nwo.Chaincode, blockNumForSnapshot int) (string, string) {
```
manish-sethi: unused parameters `orderer` and `chaincode`

integration/ledger/snapshot_test.go
```diff
@@ -0,0 +1,1049 @@
...
+	dscan := bufio.NewScanner(df)
+
+	for sscan.Scan() {
+		dscan.Scan()
+		sbytes := sscan.Bytes()
+		dbytes := dscan.Bytes()
+		if !bytes.Equal(sbytes, dbytes) {
+			return false
+		}
+	}
```
manish-sethi:
```
fileContent, err := ioutil.ReadFile(filePath)
```

integration/ledger/snapshot_test.go
```diff
@@ -0,0 +1,1049 @@
...
+
+// verify the following functions for the peer after joining a channel by snapshot
+// - invoke/query chaincode
+// - upgrade chaincode
+// - channel config update
+// - join new channel
+// - qscc
+// - query history
+func verifyPeerAfterJoinBySnapshot(helper *marblesTestHelper, peer *nwo.Peer, channelID string, chaincode *nwo.Chaincode, newCCSequence string,
+	lastBlockInSnapshot int, oldTxid string, testKey string, availableHistoryResult []*marbleHistoryResult) []*marbleHistoryResult {
```
manish-sethi: parameters names need to be clearer. It's hard to see what some of them are in the context of "verifyPeerAfterJoinBySnapshot". See related comment above about using a struct for test data related params

integration/ledger/snapshot_test.go
```diff
@@ -0,0 +1,1049 @@
...
+	peerID := peer.ID()
+
+	helper.invokeMarblesChaincode(chaincode.Name, peer, "transferMarble", testKey, "newowner_"+newCCSequence)
+
+	if newCCSequence != "" && newCCSequence != chaincode.Sequence {
+		chaincode.Version = newCCSequence + ".0"
+		chaincode.Sequence = newCCSequence
+		By("upgrading chaincode to version " + chaincode.Version + " on peer " + peerID)
+		nwo.DeployChaincode(n, channelID, orderer, *chaincode)
+	}
```
manish-sethi: This kind of side-effect in the helper function causes confusion in the main flow. I guess that it's cleaner to pass such details from the main flow.

integration/ledger/snapshot_test.go
```diff
@@ -0,0 +1,1049 @@
...
+	nwo.UpdateConfig(n, orderer, channelID, currentConfig, updatedConfig, true, peer)
+
+	By("joining a new channel testchannelID3 on peer " + peerID)
+	n.JoinChannel(testchannelID3, orderer, peer)
+
+	By("waiting for all peers to have the same ledger height")
+	channelPeers := n.PeersWithChannel(channelID)
+	maxLedgerHeight := nwo.GetMaxLedgerHeight(n, channelID, channelPeers...)
+	nwo.WaitUntilEqualLedgerHeight(n, channelID, maxLedgerHeight, channelPeers...)
+
```
manish-sethi: doesn't function "DeployChaincode" and "UpdateConfig" take care of waiting for the transaction to be committed on all the peers?

integration/ledger/snapshot_test.go
```diff
@@ -0,0 +1,1049 @@
...
+	helper.invokeMarblesChaincode(chaincode.Name, peer, "transferMarble", testKey, "newowner_update_"+newCCSequence)
+
```
manish-sethi: It will be cleaner if we keep the helper functions free from hard-coded data and pass the test data from the main flow - similar comment below as well

integration/ledger/snapshot_test.go
```diff
@@ -0,0 +1,1049 @@
...
+	By("verifying qscc GetBlockByNumber succeeds for a block number after snapshot on peer " + peerID)
+	callQSCC(n, peer, "qscc", "GetBlockByNumber", 0, "testchannel", fmt.Sprintf("%d", lastBlockInSnapshot+1))
+
+	By("verifying history on peer " + peerID)
+	By("verifying history query")
+	expectedHistoryResult := []*marbleHistoryResult{
+		{IsDelete: "false", Value: newMarble(testKey, "blue", 35, "newowner_update_"+newCCSequence)},
+		{IsDelete: "false", Value: newMarble(testKey, "blue", 35, "newowner_"+newCCSequence)},
+	}
```
manish-sethi: this is very hidden here in the helper function...  the expected history should be passed as an parameter to this function.

integration/ledger/snapshot_test.go
```diff
@@ -0,0 +1,1049 @@
...
+	expectedHistoryResult = append(expectedHistoryResult, availableHistoryResult...)
+	helper.assertGetHistoryForMarble(chaincode.Name, peer, expectedHistoryResult, testKey)
+
+	return expectedHistoryResult
```
manish-sethi: if you go with the above suggestion, this function won't return anything

manish-sethi (commented):
I feel that the we need to simplify the code in this file... A few suggestions
- The additional channels are causing more of a distraction. My suggestion would be to keep it simple and have a single channel. If we feel the need of multiple channels, we can extend it later.
- There is no need to repeat the tests separately for leveldb and couchdb. Perhaps, we can have the existing peers on leveldb and the bootstrapping peer can always use couchdb to test the additional function of index creation. Though, we do not support mixed stateDBs but for test that should be fine.

integration/ledger/snapshot_test.go
wenjianqiao: I split this function into multiple parts, some directly called from main flow and others in more specific helper functions.

wenjianqiao: Have split this functioninto multiple parts, some directly called from main flow and others in more specific helper functions.

integration/ledger/snapshot_test.go
wenjianqiao: Figured out why. If `--orderer` parameter is not present when running `peer chaincode invoke`, the chaincode CLI command will first make a `cscc` call to retrieve channel config block to get the orderer endpoint. This caused an error (e.g., `cannot serve block [2]. The ledger is bootstrapped from a snapshot. First available block = [6]`) because the channel config block was not available if a peer was bootstrapped from a snapshot and no channel config update was made. After updating couchdb index test to pass orderer parameter, the problem is resolved. 

Outside of this PR, will need to understand if SDK or discovery service has code to access the channel config block.

integration/ledger/snapshot_test.go
```diff
@@ -0,0 +1,816 @@
...
+		nwo.EnableCapabilities(setup.network, testchannelID, "Application", "V2_0", setup.orderer, channelPeers...)
+
+		By("upgrading legacy chaincode to new lifecycle chaincode")
+		nwo.DeployChaincode(setup.network, testchannelID, setup.orderer, newlifecycleChaincode, channelPeers...)
+
+		By("invoking chaincode after upgraded to new lifecycle chaincode")
+		helper.invokeMarblesChaincode(legacyChaincode.Name, org1peer0, "initMarble", "marble-upgrade", "blue", "35", "tom")
+
+		// test 5: generate snapshot again on a peer bootstrapped from a snapshot and upgraded to new lifecycle chaincode
+		blockNum2 := nwo.GetLedgerHeight(setup.network, org2peer1, testchannelID)
```
manish-sethi: Nit: Better to use more meaningful names - e.g., blockNumForNextSnapshot

integration/ledger/snapshot_test.go
```diff
@@ -0,0 +1,816 @@
...
+
+	return couchProc
+}
+
+// initAndStartMultiChannelNetwork creates a network with 2 orgs (Org1, Org2), 2 peers (peer0, peer1) per org,
+// and 3 channels (testchannel, testchannel2, testchannel3).
+// peer0 in both org: join testchannel and testchannel2
+// peer1 in org1: join no channel
+// peer1 in org2: join testchannel2
+func initAndStartMultiChannelNetwork() *setup {
```
manish-sethi: Stale comments and function name - only one channel now

integration/ledger/snapshot_test.go
```diff
@@ -0,0 +1,816 @@
...
+		peer.Channels = []*nwo.PeerChannel{
+			{Name: testchannelID, Anchor: true},
+		}
+	}
+
+	for _, org := range config.Organizations {
+		if org.Name == "Org1" || org.Name == "Org2" {
+			org.Users = 3
+		}
+	}
```
manish-sethi: Is this required now?

integration/ledger/snapshot_test.go
```diff
@@ -0,0 +1,816 @@
...
+	Eventually(sess, n.EventuallyTimeout).Should(gexec.Exit(0))
+	return sess.Buffer().Contents()
+}
+
+func compareSnapshotMetadata(snapshotDir1, snapshotDir2 string) {
+	for _, snapshotDir := range []string{snapshotDir1, snapshotDir2} {
+		By("verifying snapshot dir exists: " + snapshotDir)
+		stat, err := os.Stat(snapshotDir)
+		Expect(err).NotTo(HaveOccurred())
+		Expect(stat).NotTo(BeNil())
```
manish-sethi: ` Expect(snapshotDir).To(BeADirectory())`

integration/ledger/snapshot_test.go
```diff
@@ -0,0 +1,816 @@
...
+		By("comparing metadata file from snapshots on multiple peers: " + file)
+		fileContent1, err := ioutil.ReadFile(filepath.Join(snapshotDir1, file))
+		Expect(err).NotTo(HaveOccurred())
+		fileContent2, err := ioutil.ReadFile(filepath.Join(snapshotDir2, file))
+		Expect(err).NotTo(HaveOccurred())
+		Expect(fileContent1).To(Equal(fileContent2))
+	}
+}
+
+func startPeerInExistingOrgFromSnapshot(s *setup, orgName, peerName, channelID, snapshotDir string, lastBlockInSnapshot int, useCouchDB bool) (*nwo.Peer, ifrit.Process) {
```
manish-sethi: unused parameters `lastBlockInSnapshot`, `snapshotDir`

integration/ledger/snapshot_test.go
```diff
@@ -0,0 +1,816 @@
...
+		couchProc = configPeerWithCouchDB(s, peer)
+	}
+
+	By("starting the new peer " + peer.ID())
+	s.startPeer(peer)
+
+	return peer, couchProc
+}
+
+func startPeerInNewOrgFromSnapshot(s *setup, org *nwo.Organization, newPeerName, channelID, snapshotDir string, lastBlockInSnapshot int, useCouchDB bool) (*nwo.Peer, ifrit.Process) {
```
manish-sethi: unused parameters `lastBlockInSnapshot`, `snapshotDir`

integration/ledger/snapshot_test.go
```diff
@@ -0,0 +1,816 @@
...
+	Expect(err).NotTo(HaveOccurred())
+	Eventually(sess, n.EventuallyTimeout).Should(gexec.Exit(0))
+	orgGroup := &ordererext.DynamicOrdererOrgGroup{ConfigGroup: &common.ConfigGroup{}}
+	err = protolator.DeepUnmarshalJSON(bytes.NewBuffer(sess.Out.Contents()), orgGroup)
+	Expect(err).NotTo(HaveOccurred())
+	updatedConfig.ChannelGroup.Groups["Application"].Groups[org.Name] = orgGroup.ConfigGroup
+	nwo.UpdateConfig(n, orderer, channelID, currentConfig, updatedConfig, true, existingPeers[0], n.Peers...)
+}
+
+func joinBySnapshot(n *nwo.Network, orderer *nwo.Orderer, peer *nwo.Peer, channelID string, snapshotDir string, lastBlockInSnapshot int) {
```
manish-sethi, line 547: `lastBlockInSnapshot` can be avoided as inside the function the `channelHeight` is computed explicitly.

integration/ledger/snapshot_test.go
```diff
@@ -0,0 +1,816 @@
...
+	channelInfoStr := strings.TrimPrefix(string(sess.Buffer().Contents()[:]), "Blockchain info:")
+	var bcInfo = common.BlockchainInfo{}
+	err = json.Unmarshal([]byte(channelInfoStr), &bcInfo)
+	Expect(err).NotTo(HaveOccurred())
+	Expect(bcInfo.Height).To(Equal(uint64(channelHeight)))
+	Expect(bcInfo.BootstrappingSnapshotInfo.LastBlockInSnapshot).To(Equal(uint64(lastBlockInSnapshot)))
+
+	By("waiting for all peers to have the same ledger height")
+	channelPeers := n.PeersWithChannel(channelID)
+	maxLedgerHeight := nwo.GetMaxLedgerHeight(n, channelID, channelPeers...)
```
manish-sethi: Do you need this again as `channelHeight` already computed above?

integration/ledger/snapshot_test.go
```diff
@@ -0,0 +1,816 @@
...
+	nwo.WaitUntilEqualLedgerHeight(n, channelID, maxLedgerHeight, channelPeers...)
```
manish-sethi: why not pass only the newly bootstrapped peer?

integration/ledger/snapshot_test.go
```diff
@@ -0,0 +1,816 @@
...
+	var org *nwo.Organization
+	for _, o := range n.Organizations {
+		if o.Name == peer.Organization {
+			org = o
+			break
+		}
+	}
+	currentConfig := nwo.GetConfig(n, peer, orderer, channelID)
+	updatedConfig := proto.Clone(currentConfig).(*common.Config)
+	updatedConfig.ChannelGroup.Groups["Application"].Groups[org.Name].Policies["Writers"].Policy.Value = protoutil.MarshalOrPanic(policydsl.SignedByMspMember(org.MSPID))
```
manish-sethi: At the integration level tests, I would avoid some dummy paths. If someone adds a legitimate check later, this will start failing. The channel config update will anyway gets tested when you upgrade the capability to 2_0 - I guess that we can avoid this function entirely.

manish-sethi (commented):
Flow looks good overall, except a few comments.
In addition, other comments mostly related to cleanups - seems leftovers from previous patch at many places.

integration/chaincode/kvexecutor/chaincode.go
```diff
@@ -101,6 +113,45 @@ func (t *KVExcutor) readWriteKVs(stub shim.ChaincodeStubInterface, args []string
 	return shim.Success(resultBytes)
 }
 
+func (t *KVExcutor) getHistory(stub shim.ChaincodeStubInterface, args []string) pb.Response {
+	if len(args) != 1 {
+		return shim.Error("Incorrect number of arguments. Expecting 1 (key)")
+	}
+
+	key := args[0]
```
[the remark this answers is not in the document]
wenjianqiao: Purposely keep the change to kvexecutor for completeness. Agree it is not used in this PR. Will remove it and use a separate PR for it.

integration/ledger/snapshot_test.go
wenjianqiao: We need `lastBlockInSnapshot` to verify the bootstrapping snapshot info. `channelHeight` is for current channel height only.

integration/ledger/snapshot_test.go
wenjianqiao: Not required now, remove it from this PR. Still will need more users for pvtdata tests.

integration/ledger/snapshot_test.go
```diff
@@ -0,0 +1,816 @@
...
+			MSPID:         "Org3MSP",
+			Name:          "Org3",
+			Domain:        "org3.example.com",
+			EnableNodeOUs: true,
+			Users:         2,
+			CA: &nwo.CA{
+				Hostname: "ca",
+			},
+		}
+		org3peer0, _ := startPeerInNewOrgFromSnapshot(setup, org3, "peer0", testchannelID, snapshotDir2, blockNum, false)
```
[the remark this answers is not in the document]
wenjianqiao: I have updated the test code to add org3 and org4 during network setup to simplify the test flow related to snapshot. With the change, org3 is already added to the channel before snapshot is generated, which is the the expected scenario in the rfc.

manish-sethi (approved):
Minor nits... can be fixed later.

wlahti:
This test is failing on master in CI and locally. Opened [FAB-18288](https://jira.hyperledger.org/browse/FAB-18288).

wlahti:
Fix merged! No further action needed for the above.

PR #1963, opened 2020-10-02
Allow BCCSP config to be set using environment variables (release-1.4)

wlahti:
#### Type of change

- Bug fix
- Test update

#### Description

First three commits:
- backport PKCS11 test to release-1.4

Fourth commit:
- backport BCCSP changes to enable environment variable override of PKCS11 configuration

#### Related issues

[FAB-17969](https://jira.hyperledger.org/browse/FAB-17969)

### DIFF
--- a/bccsp/factory/nopkcs11.go
+++ b/bccsp/factory/nopkcs11.go
@@ -15,13 +15,6 @@ import (
 
 const pkcs11Enabled = false
 
-// FactoryOpts holds configuration information used to initialize factory implementations
-type FactoryOpts struct {
-	ProviderName string      `mapstructure:"default" json:"default" yaml:"Default"`
-	SwOpts       *SwOpts     `mapstructure:"SW,omitempty" json:"SW,omitempty" yaml:"SwOpts"`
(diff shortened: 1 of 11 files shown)


lindluni:
I genuinely can't decide if this should be merged. The parent of this cherry-pick was merged into master, but it pushed pkcs11 packages out of the pkcs11 build tag. I know this is a cherry-pick of a merged changed, but I don't think it should have been merged. I think the root of the problem with why you can't set environment variables is because you don't have access to the pkcs11 code without the pkcs11 build tag.

wlahti:
> I genuinely can't decide if this should be merged. The parent of this cherry-pick was merged into master, but it pushed pkcs11 packages out of the pkcs11 build tag. I know this is a cherry-pick of a merged changed, but I don't think it should have been merged. I think the root of the problem with why you can't set environment variables is because you don't have access to the pkcs11 code without the pkcs11 build tag.

I think we could actually fix things on release-1.4 by moving the PKCS11Opts out of the pkcs11 package to avoid pulling in the rest of the package. I tried this quickly on master but ran into issues due to some newly added private fields of the PKCS11Opts struct that don't exist in this release.

lindluni:
Closing this. It needs to be revisited in master as well, I am reverting the change in master. We need to approach it differently to prevent cross-compilation from breaking due to the PKCS11 code being pushed out of the build tag.

PR #1967, opened 2020-10-05
Couchdb indexes during bootstrap from a snapshot  for legacy chaincodes

manish-sethi:

#### Type of change
- New feature

#### Description
This PR enables the creation of the couchdb indexes for legacy chaincodes when creating a channel from a snapshot. This is performed for the legacy chaincodes that are defined (and not yet upgraded using the new chaincode lifecycle),
as per the data available in the snaphot files for statedb data and are installed on the peer.

### DIFF
--- a/core/ledger/cceventmgmt/mgr.go
+++ b/core/ledger/cceventmgmt/mgr.go
@@ -59,6 +59,28 @@ func (m *Mgr) Register(ledgerid string, l ChaincodeLifecycleEventListener) {
 	m.ccLifecycleListeners[ledgerid] = append(m.ccLifecycleListeners[ledgerid], l)
 }
 
+// RegisterAndInvokeFor registers the listener and in addition invokes the listener for each chaincode that is present in the supplied
+// list of legacyChaincodes and is installed on the peer
+func (m *Mgr) RegisterAndInvokeFor(legacyChaincodes []*ChaincodeDefinition, ledgerid string, l ChaincodeLifecycleEventListener) error {
+	m.rwlock.Lock()
+	defer m.rwlock.Unlock()
+	m.ccLifecycleListeners[ledgerid] = append(m.ccLifecycleListeners[ledgerid], l)
+	for _, chaincodeDefinition := range legacyChaincodes {
+		installed, dbArtifacts, err := m.infoProvider.RetrieveChaincodeArtifacts(chaincodeDefinition)
+		if err != nil {
+			return err
+		}
+		if !installed {
+			continue
+		}
+		if err := l.HandleChaincodeDeploy(chaincodeDefinition, dbArtifacts); err != nil {
+			return err
+		}
+		l.ChaincodeDeployDone(true)
+	}
+	return nil
+}
+
 // HandleChaincodeDeploy is expected to be invoked when a chaincode is deployed via a deploy transaction
 // The `chaincodeDefinitions` parameter contains all the chaincodes deployed in a block
 // We need to store the last received `chaincodeDefinitions` because this function is expected to be invoked
--- a/core/ledger/kvledger/kv_ledger.go
+++ b/core/ledger/kvledger/kv_ledger.go
@@ -142,18 +142,19 @@ func newKVLedger(initializer *lgrInitializer) (*kvLedger, error) {
 	}
 	l.isPvtstoreAheadOfBlkstore.Store(isAhead)
 
-	// TODO Move the function `GetChaincodeEventListener` to ledger interface and
-	// this functionality of registering for events to ledgermgmt package so that this
-	// is reused across other future ledger implementations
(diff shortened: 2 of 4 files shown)


core/ledger/cceventmgmt/mgr.go
```diff
@@ -59,6 +59,28 @@ func (m *Mgr) Register(ledgerid string, l ChaincodeLifecycleEventListener) {
 	m.ccLifecycleListeners[ledgerid] = append(m.ccLifecycleListeners[ledgerid], l)
 }
 
+// RegisterAndInvokeFor registers the listener and in addition invokes the listener for each chaincode that is present in the supplied
+// list of legacyChaincodes and is installed on the peer
+func (m *Mgr) RegisterAndInvokeFor(legacyChaincodes []*ChaincodeDefinition, ledgerid string, l ChaincodeLifecycleEventListener) error {
```
mastersingh24, line 64: A nit, but shouldn't this method just be called `RegisterAndInvoke` or `RegisterAndInvokeListener` ?

core/ledger/kvledger/snapshot_test.go
```diff
@@ -339,6 +320,132 @@ func TestSnapshotCouchDBIndexCreation(t *testing.T) {
...
+				Name:    "ns",
+				Version: "version",
+				Hash:    []byte("hash"),
+			},
+			installedChaincodeInfoProvider.RetrieveChaincodeArtifactsArgsForCall(0),
+		)
+		verifyIndexCreatedOnMarbleSize(lgr)
+	})
+
+	t.Run("errors-propagation", func(t *testing.T) {
```
mastersingh24, line 402: Another nit, but why nest the other tests under this?

core/ledger/kvledger/kv_ledger.go
```diff
@@ -174,6 +175,77 @@ func newKVLedger(initializer *lgrInitializer) (*kvLedger, error) {
...
+		legacyChaincodesLifecycleEventsProvider.Register(l.ledgerID, stateDBIndexCreator)
+		return nil
+	}
+
+	// opening of ledger after creating from a snapshot -
+	// it would have been better if we could explicitly retrieve the list of invocable chaincodes instead of
+	// passing the flag initializer.initializingFromSnapshot to the ccLifecycleEventProvider (which is essentially
+	// the _lifecycle cache) for directing ccLifecycleEventProvider to call us back. However, the lock that ensures
+	// the synchronization with the chaincode installer is maintained in the lifecycle cache and by design the lifecycle
+	// cache takes the responsibility of calling any listerner under the lock
```
mastersingh24:
```suggestion
	// cache takes the responsibility of calling any listener under the lock
```
typo

core/ledger/kvledger/kv_ledger.go
```diff
@@ -174,6 +175,77 @@ func newKVLedger(initializer *lgrInitializer) (*kvLedger, error) {
...
+	if err := chaincodesLifecycleEventsProvider.RegisterListener(
+		l.ledgerID, &ccEventListenerAdaptor{stateDBIndexCreator}, true); err != nil {
+		return errors.WithMessage(err, "error while creating statdb indexes after bootstrapping from snapshot")
```
mastersingh24, line 203: could probably pull `error while creating statdb indexes after bootstrapping from snapshot` into a constant

mastersingh24 (changes requested):

core/ledger/kvledger/snapshot_test.go
manish-sethi: My main purpose was to save the cost of invoking the `setup` function (which prepares a ledger, creates a snapshot as a starting point) each time for an error case tests. Otherwise, though I find that grouping of all error cases under an umbrella adds clarity but I am open to change this if you want me to over look the first part.

core/ledger/kvledger/kv_ledger.go
mastersingh24: Sure ... more detailed error is fine if you think it will provide better context than the underlying wrapped error.

core/ledger/kvledger/snapshot_test.go
mastersingh24: Not a big deal

core/ledger/cceventmgmt/mgr.go
mastersingh24: Was more curious than anything.  As long as we strive for consistency I’m ok

mastersingh24 (approved):

core/ledger/kvledger/kv_ledger.go
manish-sethi: For now I don't think that more context is needed until going forward we realize the need. Just that binding these via a common constant gives an impression that we want to keep these same - which is not the intention.

PR #1974, opened 2020-10-06
Improve Consistency When Generating Errors / Stacks on Channel Participation

jcastrence:
#### Type of change

#### Description

- Use errors.Wrap() for errors that were generated by an external (vendored) package (to ensure we generate a stack) and use errors.WithMessage() for errors that were generated by fabric (to ensure we don't generate another stack). Also, use errors.Errorf() for errors that are currently coded to use fmt.Errorf().

#### Related issues

[FAB-18229](https://jira.hyperledger.org/browse/FAB-18229)

### DIFF
--- a/orderer/common/multichannel/chainsupport.go
+++ b/orderer/common/multichannel/chainsupport.go
@@ -56,7 +56,7 @@ func newChainSupport(
 	// Assuming a block created with cb.NewBlock(), this should not
 	// error even if the orderer metadata is an empty byte slice
 	if err != nil {
-		return nil, errors.Wrapf(err, "error extracting orderer metadata for channel: %s", ledgerResources.ConfigtxValidator().ChannelID())
+		return nil, errors.WithMessagef(err, "error extracting orderer metadata for channel: %s", ledgerResources.ConfigtxValidator().ChannelID())
 	}
 
 	// Construct limited support needed as a parameter for additional support
@@ -86,7 +86,7 @@ func newChainSupport(
 
 	cs.Chain, err = consenter.HandleChain(cs, metadata)
 	if err != nil {
-		return nil, errors.Wrapf(err, "error creating consenter for channel: %s", cs.ChannelID())
+		return nil, errors.WithMessagef(err, "error creating consenter for channel: %s", cs.ChannelID())
 	}
 
 	cs.MetadataValidator, ok = cs.Chain.(consensus.MetadataValidator)
@@ -141,7 +141,7 @@ func (cs *ChainSupport) ProposeConfigUpdate(configtx *cb.Envelope) (*cb.ConfigEn
 	}
 
 	if err = checkResources(bundle); err != nil {
-		return nil, errors.Wrap(err, "config update is not compatible")
+		return nil, errors.WithMessage(err, "config update is not compatible")
 	}
 
 	if err = cs.ValidateNew(bundle); err != nil {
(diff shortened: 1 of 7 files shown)


orderer/common/multichannel/chainsupport.go
```diff
@@ -159,8 +159,8 @@ func (cs *ChainSupport) ProposeConfigUpdate(configtx *cb.Envelope) (*cb.ConfigEn
 		return nil, errors.New("new config is missing orderer group")
 	}
 
-	if err = cs.ValidateConsensusMetadata(oldOrdererConfig, newOrdererConfig, false); err != nil {
-		return nil, errors.Wrap(err, "consensus metadata update for channel config update is invalid")
+	if err = cs.ValidateConsensusMetadata(oldMetadata, newMetadata, false); err != nil {
```
btl5037, line 162: I think you forgot to update the variable declarations as well, these variable don't exist yet, this is why CI is failing

lindluni (dismissed):
Truncated

orderer/common/multichannel/ledger_resources.go
```diff
@@ -24,10 +24,10 @@ func checkResources(res channelconfig.Resources) error {
 		return errors.New("config does not contain orderer config")
 	}
 	if err := oc.Capabilities().Supported(); err != nil {
-		return errors.Wrapf(err, "config requires unsupported orderer capabilities: %s", err)
+		return errors.WithMessagef(err, "config requires unsupported orderer capabilities: %s", err)
```
btl5037, line 27: Do we need to both add a message to the existing error and then embed that same error?

orderer/common/multichannel/ledger_resources.go
```diff
@@ -24,10 +24,10 @@ func checkResources(res channelconfig.Resources) error {
...
 	}
 	if err := res.ChannelConfig().Capabilities().Supported(); err != nil {
-		return errors.Wrapf(err, "config requires unsupported channel capabilities: %s", err)
+		return errors.WithMessagef(err, "config requires unsupported channel capabilities: %s", err)
```
btl5037, line 30: Do we need to both add a message to the existing error and then embed that same error?

orderer/common/multichannel/registrar.go
```diff
@@ -436,27 +435,27 @@ func (r *Registrar) newLedgerResources(configTx *cb.Envelope) (*ledgerResources,
 
 	chdr, err := protoutil.UnmarshalChannelHeader(payload.Header.ChannelHeader)
 	if err != nil {
-		return nil, errors.Wrapf(err, "error unmarshaling channel header")
+		return nil, errors.WithMessagef(err, "error unmarshaling channel header")
```
wlahti: Super nit (aka it doesn't matter at all functionally): there isn't a formatting directive (e.g. %s) so this can just be WithMessage

```suggestion
		return nil, errors.WithMessage(err, "error unmarshaling channel header")
```

orderer/common/multichannel/registrar.go
```diff
@@ -370,7 +369,7 @@ func (r *Registrar) SystemChannel() *ChainSupport {
 func (r *Registrar) BroadcastChannelSupport(msg *cb.Envelope) (*cb.ChannelHeader, bool, *ChainSupport, error) {
 	chdr, err := protoutil.ChannelHeader(msg)
 	if err != nil {
-		return nil, false, nil, fmt.Errorf("could not determine channel ID: %s", err)
+		return nil, false, nil, errors.Errorf("could not determine channel ID: %s", err)
```
wlahti: Let's change this one to WithMessage instead. 
```suggestion
		return nil, false, nil, errors.WithMessage(err, "could not determine channel ID")
```

wlahti (changes requested):
Your first fabric PR! Thanks for this, Julian. Looking pretty good overall. I noted a few things in the code. Fix those and the general PR comments we discussed via chat (squashing the commits down into one, adding the JIRA number and your sign-off to the commit message) and this should be good to go.

caod123 (approved):

PR #1981, opened 2020-10-07
Allow tick interval override via orderer.yaml

jyellick:
#### Type of change

#### Description

If a raft network becomes unstable, sometimes, adjusting the tick
interval can be effective to restore it.  However, the tick interval is
stored in the channel config, so if the network is not operational,
modifying it is very challenging.  This commit adds a new option to the
orderer etcdraft consensus config, allowing the channel config parameter
to be overridden from the local configuration.

### DIFF
--- a/docs/source/raft_configuration.md
+++ b/docs/source/raft_configuration.md
@@ -133,8 +133,8 @@ used to further fine tune the cluster communication or replication mechanisms:
   * `SnapDir`: specifies the location at which snapshots for `etcd/raft` are stored.
   Each channel will have its own subdirectory named after the channel ID.
 
-There is also a hidden configuration parameter that can be set by adding it to
-the consensus section in the `orderer.yaml`:
+There are also two hidden configuration parameters that can each be set by adding
+them the consensus section in the `orderer.yaml`:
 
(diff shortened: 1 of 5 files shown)


docs/source/raft_configuration.md
```diff
@@ -143,6 +143,10 @@ the consensus section in the `orderer.yaml`:
   certificate), the node halts its operation for that channel. A node suspects
   its channel eviction when it doesn't know about any elected leader nor can be
   elected as leader in the channel. Defaults to 10 minutes.
+  * `TickIntervalOverride`: If set, the value here will be preferred over the
+  tick interval configured in all channels this orderer consents in.  This value
+  should be set only with great care, as a mismatch in tick interval across
+  orderers could result in a loss of quorum for one or more channels.
```
joealewine: Maybe better as:

If set, this value will be preferred over the tick interval configured in all channels where this ordering node is a consenter.

jyellick: Done

jyellick:
@guoger (or anyone else) This one should be pretty straight forward -- I have backports ready for release-2.2 and release-1.4

yacovm (approved):

PR #1991, opened 2020-10-09
[FAB-18265] Always Use DummyKeystore For PKCS11 BCCSP Provider

lindluni:
This change stops using a real keystore when the BCCSP provider
is of the PKCS11 type. PKCS11 doesn't require a keystore as the
private keys are stored in the HSM. For backwards compatibility
though we need some kind of keystore. So we should always use
the DummyKeystore implementation that exists and remove the path
for actually creating a local keystore on disk


mastersingh24 (approved):
LGTM

PR #1992, opened 2020-10-09
Shutdown blockstore (if open) before dropping ledger 

wlahti:
#### Type of change

- Bug fix

#### Description

With kubernetes and NFS, the channel participation code is unable to remove a ledger successfully. This is a first attempt to see if shutting down the underlying blockstore for the ledger allows us to remove it.

#### Related issues

[FAB-18261](https://jira.hyperledger.org/browse/FAB-18261)


common/ledger/blockledger/fileledger/factory_test.go
```diff
@@ -36,16 +41,29 @@ func TestBlockStoreProviderErrors(t *testing.T) {
...
-		err := f.Remove("foo")
-		require.EqualError(t, err, "oogie")
+		t.Run("ledger doesn't exist", func(t *testing.T) {
+			err := f.Remove("foo")
+			require.NoError(t, err)
+			require.Equal(t, 1, mockBlockStoreProvider.DropCallCount())
+		})
+
+		t.Run("dropping the blockstore fails", func(t *testing.T) {
+			mockBlockStore := &mock.BlockStore{}
```
wenjianqiao: Better not to use BlockStore interface directly in this package. This package has already defined `FileLedgerBlockStore` local interface. Can you add Shutdown() to that interface and create mock for that interface instead?

common/ledger/blockledger/fileledger/factory.go
```diff
@@ -55,6 +55,13 @@ func (f *fileLedgerFactory) Remove(channelID string) error {
 	f.mutex.Lock()
 	defer f.mutex.Unlock()
 
+	// check cache for open blockstore and, if one exists,
+	// shut it down in order to avoid resource contention
+	ledger, ok := f.ledgers[channelID]
+	if ok {
+		ledger.blockStore.Shutdown()
+	}
```
wenjianqiao, line 63: I assume that Remove can be called without first calling GetOrCreate. In that case, since blockstore is not opened yet, it should not be a problem.

wlahti: Right, that was my thinking here. I was hitting an error in the unit tests with existing ledgers on the filesystem since GetOrCreate hadn't been called and thus they weren't in the fileLedgerFactory's ledgers cache.

common/ledger/blockledger/fileledger/factory_test.go
wlahti: Sounds good. I'll try that.

wlahti: Ah, I see. I think it's just the naming of the mock that was confusing. I've updated the mock name to be consistent with the interface being mocked.

wenjianqiao (approved):

lindluni (approved):

common/ledger/blockledger/fileledger/factory_test.go
```diff
@@ -36,16 +41,29 @@ func TestBlockStoreProviderErrors(t *testing.T) {
...
 	})
 
 	t.Run("remove", func(t *testing.T) {
-		mockBlockStore.DropReturns(errors.New("oogie"))
-		err := f.Remove("foo")
-		require.EqualError(t, err, "oogie")
+		t.Run("ledger doesn't exist", func(t *testing.T) {
+			err := f.Remove("foo")
+			require.NoError(t, err)
+			require.Equal(t, 1, mockBlockStoreProvider.DropCallCount())
```
manish-sethi, line 54: You may want to check that the parameter passed to the mock is correct `foo`?

common/ledger/blockledger/fileledger/factory_test.go
```diff
@@ -36,16 +41,29 @@ func TestBlockStoreProviderErrors(t *testing.T) {
...
+
+		t.Run("dropping the blockstore fails", func(t *testing.T) {
+			mockBlockStore := &mock.FileLedgerBlockStore{}
+			f.ledgers["foo"] = &FileLedger{blockStore: mockBlockStore}
+			mockBlockStoreProvider.DropReturns(errors.New("oogie"))
+
+			err := f.Remove("foo")
+			require.EqualError(t, err, "oogie")
+			require.Equal(t, 1, mockBlockStore.ShutdownCallCount())
+			require.Equal(t, 2, mockBlockStoreProvider.DropCallCount())
```
manish-sethi, line 65: It's cleaner to write sub-test that makes them independent of each other. If you run this sub-test independently from outside `go test -run TestBlockStoreProviderErrors/remove/dropping_the_blockstore_fails`, Or if someone changes the order of subtests,  it will fail.

My suggestion would be to have a clean setup for each subtest... Here are a couple patterns that we have been following for this - [1](https://github.com/hyperledger/fabric/blob/f79abaa110fc6d6d6089b183ce62333afb5a6368/common/ledger/util/leveldbhelper/leveldb_provider_test.go#L387) or [2](https://github.com/hyperledger/fabric/blob/f79abaa110fc6d6d6089b183ce62333afb5a6368/core/ledger/pvtdatastorage/snapshot_data_importer_test.go#L57)

manish-sethi (approved):
A couple minor comments in the test file. You can address in a separate PR if you like.

PR #1997, opened 2020-10-09
Integration test - update MSP and consenters set in single config update

wlahti:
#### Type of change

- Test update

#### Description

Add an integration test that updates both the MSP and the consenters set with a newly added cert to ensure we're using the simulated config to validate consenter certs. 

#### Related issues

[FAB-18192](https://jira.hyperledger.org/browse/FAB-18192)

### DIFF
--- a/integration/raft/config_test.go
+++ b/integration/raft/config_test.go
@@ -374,19 +374,20 @@ var _ = Describe("EndToEnd reconfiguration and onboarding", func() {
 			By("Launching orderer3")
 			launch(orderer3)
 
-			By("Expanding the TLS root CA certificates")
-			nwo.UpdateOrdererMSP(network, peer, orderer, "systemchannel", "OrdererOrg", func(config msp.FabricMSPConfig) msp.FabricMSPConfig {
+			By("Expanding the TLS root CA certificates and adding orderer3 to the channel")
+			updateOrdererMSPAndConsensusMetadata(network, peer, orderer, "systemchannel", "OrdererOrg", func(config msp.FabricMSPConfig) msp.FabricMSPConfig {
 				config.TlsRootCerts = append(config.TlsRootCerts, caCert)
 				return config
-			})
-
-			By("Adding orderer3 to the channel")
-			addConsenter(network, peer, orderer, "systemchannel", etcdraft.Consenter{
-				ServerTlsCert: thirdOrdererCertificate,
-				ClientTlsCert: thirdOrdererCertificate,
-				Host:          "127.0.0.1",
-				Port:          uint32(network.OrdererPort(orderer3, nwo.ClusterPort)),
-			})
+			},
+				func(metadata *etcdraft.ConfigMetadata) {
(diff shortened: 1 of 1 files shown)


integration/raft/config_test.go
```diff
@@ -374,19 +374,20 @@ var _ = Describe("EndToEnd reconfiguration and onboarding", func() {
...
-			By("Adding orderer3 to the channel")
-			addConsenter(network, peer, orderer, "systemchannel", etcdraft.Consenter{
-				ServerTlsCert: thirdOrdererCertificate,
-				ClientTlsCert: thirdOrdererCertificate,
-				Host:          "127.0.0.1",
-				Port:          uint32(network.OrdererPort(orderer3, nwo.ClusterPort)),
-			})
+			},
+				func(metadata *etcdraft.ConfigMetadata) {
+					metadata.Consenters = append(metadata.Consenters, &etcdraft.Consenter{
```
ale-linux, line 383: why are you not using addConsenter as we were doing previously?

integration/raft/config_test.go
```diff
@@ -1806,3 +1807,51 @@ func updateEtcdRaftMetadata(network *nwo.Network, peer *nwo.Peer, orderer *nwo.O
...
+	Expect(err).NotTo(HaveOccurred())
+
+	f(metadata)
+
+	newMetadata, err := proto.Marshal(metadata)
+	Expect(err).NotTo(HaveOccurred())
+	return newMetadata
+}
+
+func updateOrdererMSPAndConsensusMetadata(network *nwo.Network, peer *nwo.Peer, orderer *nwo.Orderer, channel, orgID string, mutateMSP nwo.MSPMutator, f func(md *etcdraft.ConfigMetadata)) {
```
ale-linux, line 1823: I don't really fully understand what's going on: we update the orderer config (which we didn't do before) but we don't seem to assert that anything changed in the test - what am I missing?

wlahti: We're doing the same config updates as previously, it's just that these two are combined into a single update, which would have failed before the fix for FAB-18192 since the new consenter cert was being validated against the existing channel config instead of the in-flight (simulated) config. 

That being said, I really dislike how the existing update helpers that I awkwardly incorporated here are written since it makes following things extremely confusing to read/review as well as enhance to update multiple things at once. I wrote this originally with release-2.2 in mind (where this test has already been merged), so I avoided using the configtx library for the updates. I should probably go ahead and rewrite things here in master to utilize that. I think it'll make everything much clearer and easier to maintain.

integration/raft/config_test.go
wlahti: Because addConsenter not only adds the consenter but also submits the config update.

integration/raft/config_test.go
ale-linux: Understood, thanks!

ale-linux (approved):

PR #2002, opened 2020-10-10
Fix Node OU error message (cherry pick to release-2.2)

denyeart:
#### Type of change

- Fix error message

#### Description

If signing identity didn't match any expected Node OU roles,
the error message indicated that it matched multiple roles. This confused users.
Now there will be a separate error message for no OU roles versus multiple OU roles.


denyeart:
Opened UT flake: https://jira.hyperledger.org/browse/FAB-18267

ale-linux (approved):

PR #2010, opened 2020-10-14
Persist complete collection in the private blockstore

manish-sethi:

#### Type of change
- Bug fix

#### Description
- In the current code, while reconciling the private data corresponding to the hashes present in the snapshot world-state, we trim the collection to match the key-values present in the booting snapshot. As per the current design assumption, this suffices for transferring the private data to another peer that may bootstrap from the same or a newer snapshot at a later point in time. However, a possibility of booting a peer from an older snapshot (or from genesis block), in an environment where other peers are bootstrapped relatively newer snapshot cannot be ruled out and is good to be able to support. This commit allows for the above by not removing the unusedkey-values (if any) from the collection data for a transaction.

- A bug fix by skipping a check; wherein we check for the transaction validation flag before applying the missing private data. Because, in the snapshot state, the data appears only from the valid transactions

### DIFF
--- a/core/ledger/kvledger/kv_ledger.go
+++ b/core/ledger/kvledger/kv_ledger.go
@@ -853,7 +853,12 @@ func (l *kvLedger) CommitPvtDataOfOldBlocks(reconciledPvtdata []*ledger.Reconcil
 
 func (l *kvLedger) applyValidTxPvtDataOfOldBlocks(hashVerifiedPvtData map[uint64][]*ledger.TxPvtData) error {
 	logger.Debugf("[%s:] Filtering pvtData of invalidation transactions", l.ledgerID)
-	committedPvtData, err := filterPvtDataOfInvalidTx(hashVerifiedPvtData, l.blockStore)
+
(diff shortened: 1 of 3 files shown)


core/ledger/kvledger/kv_ledger.go
```diff
@@ -984,10 +989,17 @@ func (a *ccEventListenerAdaptor) ChaincodeDeployDone(succeeded bool) {
...
+	blockStore *blkstorage.BlockStore,
+	lastBlockInBootstrapSnapshot uint64,
+) (map[uint64][]*ledger.TxPvtData, error) {
 	committedPvtData := make(map[uint64][]*ledger.TxPvtData)
 	for blkNum, txsPvtData := range hashVerifiedPvtData {
-
+		if blkNum <= lastBlockInBootstrapSnapshot {
+			committedPvtData[blkNum] = txsPvtData
+			continue
+		}
```
cendhu, line 1002: ledger level integration test can be added in a separate PR along with other remaining tests.

cendhu (approved):

PR #2011, opened 2020-10-14
[FAB-18270] Disable debug of CouchDB response body

denyeart:
#### Type of change

- Improvement to debug logging

#### Description

The CouchDB query response body should not be added to debug log.
This change reduces the debug to the HTTP headers only, so that
the peer debug log can still be associated with the CouchDB logs, without
logging the full response content.


lindluni (approved):

denyeart:

PR #2014, opened 2020-10-14
[FAB-18130] Chan.Part.API: IT creating the system channel with a genesis block

stephyee:
#### Type of change

- Test update

#### Description

Creating the system channel with a genesis block, so no onboarding is needed.

#### Related issues

[FAB-18130](https://jira.hyperledger.org/browse/FAB-18130)

### DIFF
--- a/integration/raft/channel_participation_test.go
+++ b/integration/raft/channel_participation_test.go
@@ -69,6 +69,16 @@ var _ = Describe("ChannelParticipation", func() {
 		os.RemoveAll(testDir)
 	})
 
+	restartOrderer := func(o *nwo.Orderer, index int) {
+		ordererProcesses[index].Signal(syscall.SIGKILL)
+		Eventually(ordererProcesses[index].Wait(), network.EventuallyTimeout).Should(Receive(MatchError("exit status 137")))
+		ordererRunner := network.OrdererRunner(o)
+		ordererProcess := ifrit.Invoke(ordererRunner)
+		Eventually(ordererProcess.Ready(), network.EventuallyTimeout).Should(BeClosed())
+		ordererProcesses[index] = ordererProcess
+		ordererRunners[index] = ordererRunner
+	}
+
 	Describe("three node etcdraft network without a system channel", func() {
 		startOrderer := func(o *nwo.Orderer) {
 			ordererRunner := network.OrdererRunner(o)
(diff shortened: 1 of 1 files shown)


integration/raft/channel_participation_test.go
```diff
@@ -638,6 +638,84 @@ var _ = Describe("ChannelParticipation", func() {
...
+
+			systemChannelBlockBytes, err := ioutil.ReadFile(network.OutputBlockPath(network.SystemChannel.Name))
+			Expect(err).NotTo(HaveOccurred())
+			systemChannelBlock := &common.Block{}
+			err = proto.Unmarshal(systemChannelBlockBytes, systemChannelBlock)
+			Expect(err).NotTo(HaveOccurred())
+
+			expectedChannelInfoPT := channelparticipation.ChannelInfo{
+				Name:            network.SystemChannel.Name,
+				URL:             fmt.Sprintf("/participation/v1/channels/%s", network.SystemChannel.Name),
```
wlahti: In these expectedChannelInfo calls, I think we should directly specify the system channel name instead of using network.SystemChannel.Name. We can also avoid fmt.Sprintf.

wlahti: Nit: PT at the end stood for the "participation-trophy" channel in other tests. I think we should just remove that part and go with a plain expectedChannelInfo here.

integration/raft/channel_participation_test.go
```diff
@@ -638,6 +638,84 @@ var _ = Describe("ChannelParticipation", func() {
...
+			channelparticipation.Join(network, orderer3, network.SystemChannel.Name, systemChannelBlock, expectedChannelInfoPT)
+
+			for i, o := range orderers {
+				restartOrderer(o, i)
+			}
+
+			By("listing the channels")
+			expectedChannelInfo := channelparticipation.ChannelInfo{
+				Name:            network.SystemChannel.Name,
+				URL:             fmt.Sprintf("/participation/v1/channels/%s", network.SystemChannel.Name),
```
wlahti: Same as above here.

integration/raft/channel_participation_test.go
```diff
@@ -638,6 +638,84 @@ var _ = Describe("ChannelParticipation", func() {
...
+
+		BeforeEach(func() {
+			network = nwo.New(nwo.MultiNodeEtcdRaft(), testDir, client, StartPort(), components)
+			network.Consensus.ChannelParticipationEnabled = true
+			network.Consensus.BootstrapMethod = "none"
+			network.GenerateConfigTree()
+			network.Bootstrap()
+		})
+
+		It("Creating the system channel with a genesis block, so no onboarding is needed", func() {
```
wlahti:
```suggestion
		It("creates the system channel with a genesis block", func() {
```

integration/raft/channel_participation_test.go
```diff
@@ -638,6 +638,84 @@ var _ = Describe("ChannelParticipation", func() {
 			}
 		})
 	})
+
+	Describe("create system channel on empty network", func() {
```
wlahti:
```suggestion
	Describe("system channel creation using the channel participation API", func() {
```

wlahti: Nit with the naming. And actually, this feels weird to have as a separate Describe. Any reason this shouldn't be part of the one above for "three node etcdraft network without a system channel".

wlahti (changes requested):
It isn't explicitly mentioned in the JIRA (sorry about that) but I think we should also test that we can create a channel using "peer channel create" after we restart the orderers.

integration/raft/channel_participation_test.go
```diff
@@ -638,6 +638,84 @@ var _ = Describe("ChannelParticipation", func() {
...
+			By("joining orderers to systemchannel")
+			channelparticipation.Join(network, orderer1, network.SystemChannel.Name, systemChannelBlock, expectedChannelInfoPT)
+			channelparticipation.Join(network, orderer2, network.SystemChannel.Name, systemChannelBlock, expectedChannelInfoPT)
+			channelparticipation.Join(network, orderer3, network.SystemChannel.Name, systemChannelBlock, expectedChannelInfoPT)
+
+			for i, o := range orderers {
+				restartOrderer(o, i)
+			}
+
+			By("listing the channels")
```
wlahti: We should also call regular List to get the higher level view of the channels so we can ensure this new channel "systemchannel" is actually configured as _the_ system channel.

integration/raft/channel_participation_test.go
stephyee: 🧹

integration/raft/channel_participation_test.go
stephyee: 🧹

integration/raft/channel_participation_test.go
stephyee: 🧹

integration/raft/channel_participation_test.go
stephyee: As you noted, I moved the test to the existing describe. 🧹

integration/raft/channel_participation_test.go
stephyee: Agreed 🧹

tock-ibm:
Looks good to me.

tock-ibm (approved):

lindluni (approved):

PR #2018, opened 2020-10-16
Copy Keystore Password Into Correct Slice

lindluni:
A previous PR refactored the clone command but erroneously overwrote the cloned value.

### DIFF
--- a/bccsp/sw/fileks.go
+++ b/bccsp/sw/fileks.go
@@ -78,7 +78,7 @@ func (ks *fileBasedKeyStore) Init(pwd []byte, path string, readOnly bool) error 
 	ks.path = path
 
 	clone := make([]byte, len(pwd))
-	copy(ks.pwd, pwd)
+	copy(clone, pwd)
 	ks.pwd = clone
 	ks.readOnly = readOnly
 


denyeart:
Can the commit message net out the user impact, so that we can understand if/how to release note the fix in the back releases.

lindluni:
> Can the commit message net out the user impact, so that we can understand if/how to release note the fix in the back releases.

Done

lindluni:
I'm still looking at the code paths to determine what the impact was to this being an empty value

lindluni:
Actually, I'm not sure what the intent of `ks.pwd` was ever meant to be (looks like it was never implemented), as it looks like pwd is always passed into `Init` as nil, and `Init` is the only place `ks.pwd` is ever assigned (at least best as I can tell), though I'd be willing to be corrected.

I can't find a place where we allow the user to provide a password for an encrypted keystore.

yacovm:
> Actually, I'm not sure what the intent of `ks.pwd` was ever meant to be (looks like it was never implemented), as it looks like pwd is always passed into `Init` as nil, and `Init` is the only place `ks.pwd` is ever assigned (at least best as I can tell), though I'd be willing to be corrected.
> 
> I can't find a place where we allow the user to provide a password for an encrypted keystore.

The intent of the password was probably to encrypt the keys of the keystore with the password.
It is useful only in cases where the password is either given interactively by a user, or in case the password is saved outside of the files of the node, and injected as an environment variable.

I guess this is yet another "fallen between the cracks" case where we implemented support for something but never incorporated it into the product. We have numerous cases such as these, for example - Gossip supports dynamic (no restart) TLS certificate replacement, but the peer doesn't.

mastersingh24 (approved):
This is really a no-op change but it does correct the flawed logic

PR #2025, opened 2020-10-19
Chan.Part.API: channel participation relation and status metrics

wlahti:
#### Type of change

- New feature

#### Description

Add metrics to track the channel participation relation and status per channel for the orderer.

#### Related issues

[FAB-18087](https://jira.hyperledger.org/browse/FAB-18087)


tock-ibm:
Generally, I am a bit confused by the design of the metrics here.
 
A channel can have cluster-relation:  `{member, follower, config-tracker, none}`
so why only the boolean `is_member` `is_follower` ?

In addition, the status can be `{onboarding, active, inactive}` (and in the future also `error` or `failed`).

Labeling with `channel` is fine, but labeling with `status` will create |status| x  |clusterRelation| = 12 boolean gauges.
Is this what we want? 

Can't we have a gauge that directly spans the enum values of status? for example:
status.%channel = 0 (=onboarding) / 1 (=active) / 2 (=inactive)  / 3 (=error)

wlahti:
Adjusted the metrics and updated the places I was already setting the metrics in this PR. Need to add reporting of the metrics to at least the config-tracker codepath (and consider whether we care to report cluster relation "none" for non-raft orderers).

orderer/common/multichannel/metrics.go
```diff
@@ -0,0 +1,87 @@
...
+		r = 0
+	case types.ClusterRelationConfigTracker:
+		r = 1
+	case types.ClusterRelationFollower:
+		r = 2
+	case types.ClusterRelationMember:
+		r = 3
+	}
+	m.ClusterRelation.With("channel", channel).Set(float64(r))
+}
```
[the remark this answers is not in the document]
tock-ibm: or maybe a more robust approach would be to add to the `types` package a couple of methods that convert the `types.ClusterRelation*` and `type.Status*` into integers?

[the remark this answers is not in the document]
wlahti: Do you think it's likely we end up with additional cluster relations/statuses? I do think we should have a default, but maybe panic instead stating it's a programming error and a new metric conversion is needed in the switch statement?

I ordered these in ascending order to go along with the highest number being the most active in terms of participating in the cluster/channel, but if we anticipate more, maybe it'd be best to have 0 remain none/inactive, 1 = member/active, and then the rest. That way we keep it stable and there isn't a presumed progression of the values. Thoughts? 

So maybe for cluster relation: 
0 = none
1 = member
2 = follower
3 = config-tracker


And status:
0 = inactive
1 = active
2 = onboarding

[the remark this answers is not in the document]
tock-ibm: The ordering looks good. I like the fact that <1,1> is the "normal". I agree that panic() is a good solution for default. This way we would hopefully catch inconsistencies in IT.

I think one additional status that is already in the cards is `types.StatusFailure` - see [FAB-18106]( https://jira.hyperledger.org/browse/FAB-18106), and also my comment [here](https://github.com/hyperledger/fabric/pull/2027#discussion_r516476013). 

Maybe adding it (`types.StatusFailure`) in the context of this task is not a bad idea?

docs/source/metrics_reference.rst
```diff
@@ -61,6 +61,12 @@ The following orderer metrics are exported for consumption by Prometheus.
 |                                              |           |                                                            +-----------+--------------------------------------------------------------------+
 |                                              |           |                                                            | channel   |                                                                    |
 +----------------------------------------------+-----------+------------------------------------------------------------+-----------+--------------------------------------------------------------------+
+| cluster_relation                             | gauge     | The cluster relation of the node: 0 if none, 1 if member,  | channel   |                                                                    |
+|                                              |           | 2 if follower, 3 if config-tracker.                        |           |                                                                    |
```
tock-ibm: "member" -> "consenter"

docs/source/metrics_reference.rst
```diff
@@ -232,6 +238,12 @@ associated with the metric.
 +---------------------------------------------------------------------------+-----------+------------------------------------------------------------+
 | cluster.comm.msg_send_time.%{host}.%{channel}                             | histogram | The time it takes to send a message in seconds.            |
 +---------------------------------------------------------------------------+-----------+------------------------------------------------------------+
+| cluster.relation.%{channel}                                               | gauge     | The cluster relation of the node: 0 if none, 1 if member,  |
+|                                                                           |           | 2 if follower, 3 if config-tracker.                        |
```
tock-ibm: member -> consenter

orderer/common/multichannel/metrics.go
```diff
@@ -0,0 +1,92 @@
...
+		Help:         "The status of the cluster node: 0 if inactive, 1 if active, 2 if onboarding.",
+		LabelNames:   []string{"channel"},
+		StatsdFormat: "%{#fqname}.%{channel}",
+	}
+
+	clusterRelationOpts = metrics.GaugeOpts{
+		Namespace:    "cluster",
+		Subsystem:    "",
+		Name:         "relation",
+		Help:         "The cluster relation of the node: 0 if none, 1 if member, 2 if follower, 3 if config-tracker.",
```
tock-ibm: member -> consenter

orderer/common/multichannel/metrics.go
```diff
@@ -0,0 +1,92 @@
...
+	}
+	m.ClusterStatus.With("channel", channel).Set(float64(s))
+}
+
+func (m *Metrics) reportClusterRelation(channel string, relation types.ClusterRelation) {
+	var r int
+	switch relation {
+	case types.ClusterRelationNone:
+		r = 0
+	case types.ClusterRelationMember:
```
tock-ibm: wasn't this supposed to be `types.ClusterRelationConsenter`? probably needs a rebase as well

tock-ibm:
Looks good, except that the transition of cluster-relation member->consenter needs to be applied here as well.

wlahti:
> Looks good, except that the transition of cluster-relation member->consenter needs to be applied here as well.

Thanks for catching that! Forgot that this PR was also affected by the rename somehow. :)

orderer/common/multichannel/metrics.go
wlahti: Done

orderer/common/multichannel/metrics.go
wlahti: Done

orderer/common/follower/follower_chain.go
```diff
@@ -374,12 +386,15 @@ func (c *Chain) pullUpToJoin() error {
...
-// It checks whether the chain was stopped between blocks.
+// pullAfterJoin pulls blocks continuously, inspecting the fetched config
+// blocks for membership. On every config block, it renews the BlockPuller,
+// to take in the new configuration. It will exit with 'nil' if it detects
+// a config block that indicates the orderer is a member of the cluster. It
+// checks whether the chain was stopped between blocks.
 func (c *Chain) pullAfterJoin() error {
 	c.setStatus(types.StatusActive)
+	clusterRelation, status := c.StatusReport()
+	c.clusterMetricsReporter.ReportClusterRelationAndStatusMetrics(c.ledgerResources.ChannelID(), clusterRelation, status)
```
jyellick: Why not embed this into `setStatus`?  You need to call it any time the status is updated, and, you won't have to re-acquire the lock that's done in StatusReport().  Also, if someone ever sets status elsewhere, they won't have to remember to update the metric.

orderer/common/multichannel/metrics.go
```diff
@@ -0,0 +1,92 @@
...
+	"github.com/hyperledger/fabric/common/metrics"
+	"github.com/hyperledger/fabric/orderer/common/types"
+)
+
+var (
+	clusterStatusOpts = metrics.GaugeOpts{
+		Namespace:    "cluster",
+		Subsystem:    "",
+		Name:         "status",
+		Help:         "The status of the cluster node: 0 if inactive, 1 if active, 2 if onboarding.",
```
jyellick: I'm not totally convinced here, but wouldn't it be more natural to go 0 inactive, 1 onboarding, 2 active? (ie, typically you transition from 0 to 1 to 2?)

orderer/common/multichannel/metrics.go
```diff
@@ -0,0 +1,92 @@
...
+package multichannel
+
+import (
+	"github.com/hyperledger/fabric/common/metrics"
+	"github.com/hyperledger/fabric/orderer/common/types"
+)
+
+var (
+	clusterStatusOpts = metrics.GaugeOpts{
+		Namespace:    "cluster",
```
jyellick: So, we already have a "cluster" namespace, used by `orderer/common/cluster`, but it does specify a subsystem of `comm`.  Perhaps we should add a subsystem of `participation` here? Or pick a different namespace?

orderer/common/multichannel/metrics.go
```diff
@@ -0,0 +1,92 @@
...
+		Subsystem:    "",
+		Name:         "status",
```
jyellick, line 18: The name `status` is pretty hopelessly vague, unless we add more context with a subsystem or different namespace.  A cluster status of 'active' for instance does not imply that the other nodes are reachable, etc.  It is the `participation_status` or similar.

orderer/common/multichannel/metrics.go
```diff
@@ -0,0 +1,92 @@
...
+		Help:         "The status of the cluster node: 0 if inactive, 1 if active, 2 if onboarding.",
+		LabelNames:   []string{"channel"},
+		StatsdFormat: "%{#fqname}.%{channel}",
+	}
+
+	clusterRelationOpts = metrics.GaugeOpts{
+		Namespace:    "cluster",
+		Subsystem:    "",
+		Name:         "relation",
```
jyellick: I also wonder if `relation` is the best we can do here?  I understand that's what it's referred to in the code, but it's not exactly obvious to someone looking at a metrics dashboard.

orderer/consensus/etcdraft/consenter.go
```diff
@@ -160,10 +162,12 @@ func (c *Consenter) HandleChain(support consensus.ConsenterSupport, metadata *co
 	id, err := c.detectSelfID(consenters)
 	if err != nil {
 		if c.InactiveChainRegistry != nil {
-			// There is a system channel, use the InactiveChainRegistry to track the future config updates of application channel.
+			// There is a system channel, use the InactiveChainRegistry to track the
+			// future config updates of application channel.
 			c.InactiveChainRegistry.TrackChain(support.ChannelID(), support.Block(0), func() {
 				c.ChainManager.CreateChain(support.ChannelID())
 			})
+			c.ChainManager.ReportClusterRelationAndStatusMetrics(support.ChannelID(), types.ClusterRelationConfigTracker, types.StatusInactive)
```
jyellick: I _really_ do not like this.  And this is one of the strong reasons why I asked that we please ditch the notion of the raft consenter switching the chain out, and delegate this back up to the multichannel registrar.  Consensus plugins have no business setting cluster relation status.

wlahti: Yes, I did not like doing this but it's the only way I found with the current code flow.

orderer/common/multichannel/metrics.go
wlahti: I definitely like the idea of adding "participation" as the subsystem. And I wasn't in love with using "cluster" as the namespace. Maybe it should really be "channel"?

orderer/common/multichannel/metrics.go
wlahti: That's how I originally implemented this but decided against it due to the fact that we may end up adding different options in the future. For the sake of keeping a stable base set of statuses and allowing new ones in the future, I thought it might be best to move away from the concept of a progression in the enum.

orderer/common/multichannel/metrics.go
wlahti: Went with "cluster" and "participation" for now (and updated the help text) since the channel name is included as a label.

orderer/common/follower/follower_chain.go
wlahti: Good idea. Done (also added to setClusterRelation).

orderer/common/multichannel/metrics.go
jyellick: The more I think on this, the more convinced I am that 'cluster' isn't right, and I'm a little confused why we call it cluster status at all.  It has nothing to do with the cluster? Being a follower does not make you a member of the cluster... why not a vanilla channel_participation?

orderer/consensus/etcdraft/consenter.go
jyellick: Do we have an outstanding JIRA to fix this flow?

wlahti: No, I do not see a JIRA for this. I can try to create one but I'm not sure if I have enough context in this area to get it completely right.

orderer/common/multichannel/metrics.go
wlahti: Are you suggesting we change the clusterRelation in general reported by the channel participation API in addition to the metric? We could rename that to Relation if it's not cluster-related and then the metric could be namepsace "channel" subsystem "participation" and name "relation" for channel_participation_relation.

wlahti: Updated to what I mentioned in my previous comment.

orderer/consensus/etcdraft/consenter.go
tock-ibm: I believe there is a Jira for it, or a variation on this theme: https://jira.hyperledger.org/browse/FAB-18152

However, the big landmine we have to dismantle is the pattern with the `InactiveChainRegistry` etc., and we decided back then to finish the channel participation functionality first, and only then start refactoring. Moreover, the dependencies around the `InactiveChainRegistry` and `BlockPuller` are pretty complex, so this should be done very carefully... (if at all - "if it ain't broke don't fix it" comes to mind). Hopefully when we eventually deprecate the system channel this task will be a lot easier.

orderer/common/multichannel/metrics.go
wlahti: Updated to "participation" for the namespace and no subsystem.

orderer/common/multichannel/metrics.go
wlahti: Done. It's participation_status now with the namespace change.

orderer/common/multichannel/metrics.go
wlahti: Created follow-up task https://jira.hyperledger.org/browse/FAB-18327 for renaming this (and the underlying types.ClusterRelation) to consensus relation.

jyellick (approved):

PR #2027, opened 2020-10-20
[FAB-18132] Ch.Part.API: Remove ledger resources asynchronously

stephyee:
#### Type of change

- New feature

#### Description

Removes ledger resources in asynchronously when channel participation APIs remove is called.  

#### Related issues

[FAB-18132](https://jira.hyperledger.org/browse/FAB-18132)

### DIFF
--- a/common/ledger/blockledger/fileledger/factory.go
+++ b/common/ledger/blockledger/fileledger/factory.go
@@ -59,15 +59,15 @@ func (f *fileLedgerFactory) Remove(channelID string) error {
 	f.mutex.Lock()
 	defer f.mutex.Unlock()
 
+	if err := f.removeFileRepo.Save(channelID, []byte{}); err != nil && err != os.ErrExist {
+		return err
+	}
+
 	// check cache for open blockstore and, if one exists,
 	// shut it down in order to avoid resource contention
 	ledger, ok := f.ledgers[channelID]
 	if ok {
 		ledger.blockStore.Shutdown()
-	}
-
-	if err := f.removeFileRepo.Save(channelID, []byte{}); err != nil && err != os.ErrExist {
-		return err
 	}
 
 	err := f.blkstorageProvider.Drop(channelID)
--- a/integration/raft/channel_participation_test.go
+++ b/integration/raft/channel_participation_test.go
@@ -102,7 +102,8 @@ var _ = Describe("ChannelParticipation", func() {
 			orderer1 := network.Orderer("orderer1")
 			startOrderer(orderer1)
 
-			channelparticipation.List(network, orderer1, nil)
+			cl := channelparticipation.List(network, orderer1)
+			Expect(cl).To(Equal(channelparticipation.ChannelList{}))
 
 			By("attempting to create a channel without a system channel defined")
 			sess, err := network.PeerAdminSession(network.Peer("Org1", "peer0"), commands.ChannelCreate{
@@ -128,7 +129,8 @@ var _ = Describe("ChannelParticipation", func() {
 			By("starting all three orderers")
 			for _, o := range orderers {
 				startOrderer(o)
-				channelparticipation.List(network, o, nil)
+				cl := channelparticipation.List(network, o)
+				Expect(cl).To(Equal(channelparticipation.ChannelList{}))
 			}
 
 			genesisBlock := applicationChannelGenesisBlock(network, consenters, peer, "participation-trophy")
@@ -219,7 +221,8 @@ var _ = Describe("ChannelParticipation", func() {
 			Expect(channelInfo).To(Equal(expectedChannelInfoAPT))
 
 			By("listing all channels for orderer1")
-			channelparticipation.List(network, orderer1, []string{"participation-trophy", "another-participation-trophy"})
+			cl := channelparticipation.List(network, orderer1)
+			channelparticipation.ChannelListMatcher(cl, []string{"participation-trophy", "another-participation-trophy"})
 
 			By("removing orderer1 from the consenter set")
 			channelConfig = nwo.GetConfig(network, peer, orderer2, "participation-trophy")
@@ -261,9 +264,27 @@ var _ = Describe("ChannelParticipation", func() {
 
 			By("removing orderer1 from a channel")
 			channelparticipation.Remove(network, orderer1, "participation-trophy")
+			Eventually(func() channelparticipation.ChannelList {
+				return channelparticipation.List(network, orderer1)
+			}, network.EventuallyTimeout).Should(Equal(channelparticipation.ChannelList{
+				SystemChannel: nil,
+				Channels: []channelparticipation.ChannelInfoShort{
+					{
+						Name: "another-participation-trophy",
+						URL:  "/participation/v1/channels/another-participation-trophy",
+					},
+				},
+			}))
+
+			By("submitting transaction to orderer1")
+			env := CreateBroadcastEnvelope(network, peer, "participation-trophy", []byte("hello"))
+			resp, err := ordererclient.Broadcast(network, orderer1, env)
+			Expect(err).NotTo(HaveOccurred())
+			Expect(resp.Status).To(Equal(common.Status_BAD_REQUEST))
 
 			By("listing all channels for orderer1")
-			channelparticipation.List(network, orderer1, []string{"another-participation-trophy"})
+			cl = channelparticipation.List(network, orderer1)
+			channelparticipation.ChannelListMatcher(cl, []string{"another-participation-trophy"})
 
 			By("ensuring the channel is still usable by submitting a transaction to each remaining consenter for the channel")
 			submitPeerTxn(orderer2, peer, network, consenters, 7, channelparticipation.ChannelInfo{
@@ -307,7 +328,8 @@ var _ = Describe("ChannelParticipation", func() {
 			By("starting two orderers")
 			for _, o := range orderers {
 				startOrderer(o)
-				channelparticipation.List(network, o, nil)
+				cl := channelparticipation.List(network, o)
+				Expect(cl).To(Equal(channelparticipation.ChannelList{}))
 			}
 
 			genesisBlock := applicationChannelGenesisBlock(network, orderers, peer, "participation-trophy")
@@ -354,7 +376,8 @@ var _ = Describe("ChannelParticipation", func() {
 
 			By("starting third orderer")
 			startOrderer(orderer3)
-			channelparticipation.List(network, orderer3, nil)
+			cl := channelparticipation.List(network, orderer3)
+			Expect(cl).To(Equal(channelparticipation.ChannelList{}))
 
 			By("adding orderer3 to the consenters set")
 			channelConfig = nwo.GetConfig(network, peer, orderer2, "participation-trophy")
@@ -401,7 +424,8 @@ var _ = Describe("ChannelParticipation", func() {
 			By("starting two orderers")
 			for _, o := range orderers {
 				startOrderer(o)
-				channelparticipation.List(network, o, nil)
+				cl := channelparticipation.List(network, o)
+				Expect(cl).To(Equal(channelparticipation.ChannelList{}))
 			}
 
 			genesisBlock := applicationChannelGenesisBlock(network, orderers, peer, "participation-trophy")
@@ -448,7 +472,8 @@ var _ = Describe("ChannelParticipation", func() {
 
 			By("starting third orderer")
 			startOrderer(orderer3)
-			channelparticipation.List(network, orderer3, nil)
+			cl := channelparticipation.List(network, orderer3)
+			Expect(cl).To(Equal(channelparticipation.ChannelList{}))
 
 			By("joining orderer3 to the channel as a follower")
 			// make sure we can join using a config block from one of the other orderers
@@ -555,7 +580,8 @@ var _ = Describe("ChannelParticipation", func() {
 
 			for _, o := range orderers {
 				By("listing the channels for " + o.Name)
-				channelparticipation.List(network, o, []string{"testchannel"}, "systemchannel")
+				cl := channelparticipation.List(network, o)
+				channelparticipation.ChannelListMatcher(cl, []string{"testchannel"}, []string{"systemchannel"}...)
 			}
 
 			expectedChannelInfo = channelparticipation.ChannelInfo{
@@ -644,7 +670,8 @@ var _ = Describe("ChannelParticipation", func() {
 					return channelparticipation.ListOne(network, o, "testchannel")
 				}, network.EventuallyTimeout).Should(Equal(expectedChannelInfo))
 				By("listing all channels")
-				channelparticipation.List(network, o, []string{"testchannel"}, "systemchannel")
+				cl := channelparticipation.List(network, o)
+				channelparticipation.ChannelListMatcher(cl, []string{"testchannel"}, []string{"systemchannel"}...)
 			}
 
 			By("submitting a transaction to ensure the system channel is active after restart")
@@ -756,9 +783,9 @@ var _ = Describe("ChannelParticipation", func() {
 
 			By("listing the channels after removing the system channel")
 			for _, o := range orderers1and2 {
-				channelparticipation.List(network, o, []string{"testchannel"})
+				cl := channelparticipation.List(network, o)
+				channelparticipation.ChannelListMatcher(cl, []string{"testchannel"})
 			}
-			channelparticipation.List(network, orderer3, []string{})
 
 			By("submitting a transaction to each active orderer after restart")
 			submitPeerTxn(orderer1, peer, network, orderers1and2, 8, channelparticipation.ChannelInfo{
--- a/orderer/common/channelparticipation/restapi.go
+++ b/orderer/common/channelparticipation/restapi.go
@@ -257,6 +257,11 @@ func (h *HTTPHandler) sendJoinError(err error, resp http.ResponseWriter) {
 	case types.ErrAppChannelsAlreadyExists:
 		// The client is trying to join the system-channel that does not exist, but app channels exist.
 		h.sendResponseJsonError(resp, http.StatusForbidden, errors.WithMessage(err, "cannot join"))
+	case types.ErrChannelPendingRemoval:
+		// The client is trying to join a channel that is currently being removed.
+		h.sendResponseJsonError(resp, http.StatusConflict, errors.WithMessage(err, "cannot join"))
+	case types.ErrChannelRemovalFailure:
+		h.sendResponseJsonError(resp, http.StatusInternalServerError, errors.WithMessage(err, "cannot join"))
 	default:
 		h.sendResponseJsonError(resp, http.StatusBadRequest, errors.WithMessage(err, "cannot join"))
 	}
@@ -289,6 +294,10 @@ func (h *HTTPHandler) serveRemove(resp http.ResponseWriter, req *http.Request) {
 		h.sendResponseNotAllowed(resp, errors.WithMessage(err, "cannot remove"), http.MethodGet)
 	case types.ErrChannelNotExist:
 		h.sendResponseJsonError(resp, http.StatusNotFound, errors.WithMessage(err, "cannot remove"))
+	case types.ErrChannelPendingRemoval:
+		h.sendResponseJsonError(resp, http.StatusConflict, errors.WithMessage(err, "cannot remove"))
+	case types.ErrChannelRemovalFailure:
+		h.sendResponseJsonError(resp, http.StatusInternalServerError, errors.WithMessage(err, "cannot remove"))
 	default:
 		h.sendResponseJsonError(resp, http.StatusBadRequest, errors.WithMessage(err, "cannot remove"))
 	}
--- a/orderer/common/multichannel/registrar.go
+++ b/orderer/common/multichannel/registrar.go
@@ -10,7 +10,9 @@ SPDX-License-Identifier: Apache-2.0
 package multichannel
 
 import (
+	"fmt"
 	"path/filepath"
+	"strings"
 	"sync"
 
 	"github.com/golang/protobuf/proto"
@@ -45,9 +47,11 @@ var logger = flogging.MustGetLogger("orderer.commmon.multichannel")
 type Registrar struct {
 	config localconfig.TopLevel
 
-	lock            sync.RWMutex
-	chains          map[string]*ChainSupport
-	followers       map[string]*follower.Chain
+	lock      sync.RWMutex
+	chains    map[string]*ChainSupport
+	followers map[string]*follower.Chain
+	// true indicates a ledger removal is pending; false indicates a ledger removal failed
+	pendingRemoval  map[string]bool
 	systemChannelID string
 	systemChannel   *ChainSupport
 
@@ -96,6 +100,7 @@ func NewRegistrar(
 		config:             config,
 		chains:             make(map[string]*ChainSupport),
 		followers:          make(map[string]*follower.Chain),
+		pendingRemoval:     make(map[string]bool),
 		ledgerFactory:      ledgerFactory,
 		signer:             signer,
 		blockcutterMetrics: blockcutter.NewMetrics(metricsProvider),
@@ -664,6 +669,13 @@ func (r *Registrar) ChannelList() types.ChannelList {
 		list.Channels = append(list.Channels, types.ChannelInfoShort{Name: name})
 	}
 
+	for c := range r.pendingRemoval {
+		list.Channels = append(list.Channels, types.ChannelInfoShort{
+			Name: c,
+		})
+
+	}
+
 	return list
 }
 
@@ -687,6 +699,15 @@ func (r *Registrar) ChannelInfo(channelID string) (types.ChannelInfo, error) {
 		return info, nil
 	}
 
+	_, ok := r.pendingRemoval[channelID]
+	if ok {
+		return types.ChannelInfo{
+			Name:            channelID,
+			ClusterRelation: types.ClusterRelationFollower,
+			Status:          types.StatusInactive,
+		}, nil
+	}
+
 	return types.ChannelInfo{}, types.ErrChannelNotExist
 }
 
@@ -695,6 +716,14 @@ func (r *Registrar) ChannelInfo(channelID string) (types.ChannelInfo, error) {
 func (r *Registrar) JoinChannel(channelID string, configBlock *cb.Block, isAppChannel bool) (info types.ChannelInfo, err error) {
 	r.lock.Lock()
 	defer r.lock.Unlock()
+
+	if removalStatus, ok := r.pendingRemoval[channelID]; ok {
+		if removalStatus {
+			return types.ChannelInfo{}, types.ErrChannelPendingRemoval
+		} else {
+			return types.ChannelInfo{}, types.ErrChannelRemovalFailure
+		}
+	}
 
 	if r.systemChannelID != "" {
 		return types.ChannelInfo{}, types.ErrSystemChannelExists
@@ -886,6 +915,15 @@ func (r *Registrar) joinSystemChannel(
 func (r *Registrar) RemoveChannel(channelID string) error {
 	r.lock.Lock()
 	defer r.lock.Unlock()
+
+	if removalStatus, ok := r.pendingRemoval[channelID]; ok {
+		if removalStatus {
+			return types.ErrChannelPendingRemoval
+		} else {
+			return types.ErrChannelRemovalFailure
+		}
+	}
+
 	if r.systemChannelID != "" {
 		if channelID != r.systemChannelID {
 			return types.ErrSystemChannelExists
@@ -896,7 +934,8 @@ func (r *Registrar) RemoveChannel(channelID string) error {
 	cs, ok := r.chains[channelID]
 	if ok {
 		cs.Halt()
-		return r.removeMember(channelID, cs)
+		r.removeMember(channelID, cs)
+		return nil
 	}
 
 	fChain, ok := r.followers[channelID]
@@ -908,17 +947,12 @@ func (r *Registrar) RemoveChannel(channelID string) error {
 	return types.ErrChannelNotExist
 }
 
-func (r *Registrar) removeMember(channelID string, cs *ChainSupport) error {
-	err := r.ledgerFactory.Remove(channelID)
-	if err != nil {
-		return errors.Errorf("error removing ledger for channel %s", channelID)
-	}
+func (r *Registrar) removeMember(channelID string, cs *ChainSupport) {
+	r.removeLedgerAsync(channelID)
 
 	delete(r.chains, channelID)
 
 	logger.Infof("Removed channel: %s", channelID)
-
-	return nil
 }
 
 func (r *Registrar) removeFollower(channelID string, follower *follower.Chain) error {
@@ -931,10 +965,7 @@ func (r *Registrar) removeFollower(channelID string, follower *follower.Chain) e
 		return err
 	}
 
-	err := r.ledgerFactory.Remove(channelID)
-	if err != nil {
-		return errors.Errorf("error removing ledger for channel %s", channelID)
-	}
+	r.removeLedgerAsync(channelID)
 
 	delete(r.followers, channelID)
 
@@ -1000,6 +1031,19 @@ func (r *Registrar) removeSystemChannel() error {
 	r.systemChannel.Halt()
 	delete(r.chains, systemChannelID)
 
+	// remove system channel resources
+	err := r.ledgerFactory.Remove(systemChannelID)
+	if err != nil {
+		return errors.WithMessagef(err, "failed removing ledger for system channel %s", r.systemChannelID)
+	}
+
+	// remove system channel references
+	r.systemChannel = nil
+	r.systemChannelID = ""
+	logger.Infof("removed system channel: %s", systemChannelID)
+
+	failedRemovals := []string{}
+
 	// halt all application channels
 	for channel, cs := range r.chains {
 		cs.Halt()
@@ -1014,23 +1058,21 @@ func (r *Registrar) removeSystemChannel() error {
 			return errors.WithMessagef(err, "failed to determine channel membership for channel: %s", channel)
 		}
 		if !isChannelMember {
-			logger.Debugf("Not a member of channel %s, removing it", channel)
-			err := r.removeMember(channel, cs)
+			logger.Debugf("not a member of channel %s, removing it", channel)
+			err := r.ledgerFactory.Remove(channel)
 			if err != nil {
-				return errors.WithMessagef(err, "failed to remove channel: %s", channel)
+				logger.Errorf("failed removing ledger for channel %s, error: %v", channel, err)
+				failedRemovals = append(failedRemovals, channel)
+				continue
 			}
+
+			delete(r.chains, channel)
 		}
 	}
 
-	// remove system channel resources
-	err := r.ledgerFactory.Remove(systemChannelID)
-	if err != nil {
-		return errors.WithMessagef(err, "error removing ledger for system channel %s", r.systemChannelID)
+	if len(failedRemovals) > 0 {
+		return fmt.Errorf("failed removing ledger for channel(s): %s", strings.Join(failedRemovals, ", "))
 	}
-
-	// remove system channel references
-	r.systemChannel = nil
-	r.systemChannelID = ""
 
 	// reintialize the registrar to recreate every channel
 	r.init(r.consenters)
@@ -1038,7 +1080,22 @@ func (r *Registrar) removeSystemChannel() error {
 	// restart every channel
 	r.startChannels()
 
-	logger.Infof("Removed system channel: %s", systemChannelID)
-
 	return nil
+}
+
+func (r *Registrar) removeLedgerAsync(channelID string) {
+	r.pendingRemoval[channelID] = true
+
+	go func() {
+		if err := r.ledgerFactory.Remove(channelID); err != nil {
+			r.lock.Lock()
+			r.pendingRemoval[channelID] = false
+			r.lock.Unlock()
+			logger.Errorf("ledger factory failed to remove empty ledger '%s', error: %s", channelID, err)
+			return
+		}
+		r.lock.Lock()
+		delete(r.pendingRemoval, channelID)
+		r.lock.Unlock()
+	}()
 }
--- a/orderer/common/multichannel/registrar_test.go
+++ b/orderer/common/multichannel/registrar_test.go
@@ -938,6 +938,39 @@ func TestRegistrar_JoinChannel(t *testing.T) {
 		os.RemoveAll(tmpdir)
 	}
 
+	t.Run("Reject join when removal is occurring", func(t *testing.T) {
+		setup(t)
+		defer cleanup()
+
+		newLedger(ledgerFactory, "sys-raft-channel", genesisBlockSysRaft)
+
+		registrar := NewRegistrar(localconfig.TopLevel{}, ledgerFactory, mockCrypto(), &disabled.Provider{}, cryptoProvider, nil)
+		registrar.Initialize(mockConsenters)
+
+		registrar.pendingRemoval["some-app-channel"] = true
+		info, err := registrar.JoinChannel("some-app-channel", &cb.Block{}, true)
+		require.Equal(t, err, types.ErrChannelPendingRemoval)
+		require.Equal(t, types.ChannelInfo{}, info)
+		joinBlockPath := filepath.Join(tmpdir, "pendingops", "joinblock", "some-app-channel.joinblock")
+		_, err = os.Stat(joinBlockPath)
+		require.True(t, os.IsNotExist(err))
+	})
+
+	t.Run("Reject join when removal previously failed", func(t *testing.T) {
+		setup(t)
+		defer cleanup()
+
+		newLedger(ledgerFactory, "sys-raft-channel", genesisBlockSysRaft)
+
+		registrar := NewRegistrar(localconfig.TopLevel{}, ledgerFactory, mockCrypto(), &disabled.Provider{}, cryptoProvider, nil)
+		registrar.Initialize(mockConsenters)
+
+		registrar.pendingRemoval["some-app-channel"] = false
+		info, err := registrar.JoinChannel("some-app-channel", &cb.Block{}, true)
+		require.Equal(t, types.ErrChannelRemovalFailure, err)
+		require.Equal(t, types.ChannelInfo{}, info)
+	})
+
 	t.Run("Reject join when system channel exists", func(t *testing.T) {
 		setup(t)
 		defer cleanup()
@@ -1026,6 +1059,7 @@ func TestRegistrar_JoinChannel(t *testing.T) {
 			require.EqualError(t, err, "failed to determine cluster membership from join-block: apple")
 
 			// After join failure, check that everything has been cleaned up
+			require.Eventually(t, func() bool { return len(ledgerFactory.ChannelIDs()) == 0 }, time.Minute, time.Second)
 			require.Empty(t, ledgerFactory.ChannelIDs())
 			joinBlockPath := filepath.Join(tmpdir, "pendingops", "joinblock", "my-raft-channel.joinblock")
 			_, err = os.Stat(joinBlockPath)
@@ -1049,6 +1083,7 @@ func TestRegistrar_JoinChannel(t *testing.T) {
 			require.EqualError(t, err, "failed to create chain support: error creating consenter for channel: my-raft-channel: banana")
 
 			// After join failure, check that everything has been cleaned up
+			require.Eventually(t, func() bool { return len(ledgerFactory.ChannelIDs()) == 0 }, time.Minute, time.Second)
 			require.Empty(t, ledgerFactory.ChannelIDs())
 			joinBlockPath := filepath.Join(tmpdir, "pendingops", "joinblock", "my-raft-channel.joinblock")
 			_, err = os.Stat(joinBlockPath)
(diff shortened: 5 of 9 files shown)


common/ledger/blockledger/fileledger/factory.go
```diff
@@ -50,22 +51,33 @@ func (f *fileLedgerFactory) GetOrCreate(channelID string) (blockledger.ReadWrite
...
 
-	err := f.blkstorageProvider.Drop(channelID)
-	if err != nil {
-		return err
-	}
+	go func() {
+		err := f.blkstorageProvider.Drop(channelID)
+		if err != nil {
+			logger.Info(err.Error())
+			f.toBeRemoved[channelID] = false
```
yacovm: the mutex doesn't apply in this goroutine

common/ledger/blockledger/fileledger/factory.go
```diff
@@ -50,22 +51,33 @@ func (f *fileLedgerFactory) GetOrCreate(channelID string) (blockledger.ReadWrite
...
 	ledger, ok := f.ledgers[channelID]
 	if ok {
 		ledger.blockStore.Shutdown()
 	}
 
-	err := f.blkstorageProvider.Drop(channelID)
-	if err != nil {
-		return err
-	}
+	go func() {
```
yacovm: what happens if you crash during this procedure? does it make sense to somehow "mark" the channel for deletion synchronously, and then just asynchronously remove it?

stephyee:
I noticed this change introduced a data races along with a few other bad things. Working on fixing this up.

orderer/common/multichannel/registrar.go
```diff
@@ -102,41 +103,53 @@ func NewRegistrar(
...
 
 	joinBlockFileRepo, err := filerepo.New(fileRepoDir, "joinblock")
 	if err != nil {
-		return nil, err
+		return nil, nil, err
+	}
+
+	removeFileRepo, err := filerepo.New(fileRepoDir, "remove")
+	if err != nil {
+		return nil, nil, err
```
tock-ibm: I think the `removeFileRepo` should remain encapsulated within file ledger.

orderer/common/multichannel/registrar.go
```diff
@@ -188,19 +204,22 @@ func (r *Registrar) discoverLedgers(channelsWithJoinBlock map[string]*cb.Block)
...
 			}
 		}
 	}
-
 	return r.ledgerFactory.ChannelIDs()
 }
 
 // initSystemChannel scan for and initialize the system channel, if it exists.
-func (r *Registrar) initSystemChannel(existingChannels []string) {
+func (r *Registrar) initSystemChannel(existingChannels []string, systemChannelID string) {
```
tock-ibm: The job of this method is to search for a system channel, so there really is no point in passing in that string

orderer/common/multichannel/registrar.go
```diff
@@ -188,19 +204,22 @@ func (r *Registrar) discoverLedgers(channelsWithJoinBlock map[string]*cb.Block)
...
 	for _, channelID := range existingChannels {
+		if channelID == systemChannelID {
+			continue
+		}
+
```
tock-ibm: again this does not serve any purpose

orderer/common/multichannel/registrar.go
```diff
@@ -259,7 +278,7 @@ func (r *Registrar) initSystemChannel(existingChannels []string) {
 
 // initAppChannels initializes application channels, assuming that the system channel does NOT exist.
 // This implies that the orderer is using the channel participation API for joins (channel creation).
-func (r *Registrar) initAppChannels(existingChannels []string, channelsWithJoinBlock map[string]*cb.Block) {
+func (r *Registrar) initAppChannels(existingChannels []string, channelsWithJoinBlock map[string]*cb.Block, sysChannel string) {
```
tock-ibm: by this time r.systemChannelID signals whether we have a system channel or not.... and, read the comment:
"assuming that the system channel does NOT exist"

orderer/common/multichannel/registrar.go
```diff
@@ -714,8 +741,10 @@ func (r *Registrar) JoinChannel(channelID string, configBlock *cb.Block, isAppCh
 
 	defer func() {
 		if err != nil {
-			if err2 := r.ledgerFactory.Remove(channelID); err2 != nil {
+			r.pendingRemoval[channelID] = true
+			if err2 := r.ledgerFactory.Remove(channelID, r.removeChannelFromPendingRemovals); err2 != nil {
```
tock-ibm: this could be done synchronously, as it is a small ledger, only just constructed

orderer/common/multichannel/registrar.go
```diff
@@ -909,8 +943,10 @@ func (r *Registrar) RemoveChannel(channelID string) error {
 }
 
 func (r *Registrar) removeMember(channelID string, cs *ChainSupport) error {
-	err := r.ledgerFactory.Remove(channelID)
+	r.pendingRemoval[channelID] = true
+	err := r.ledgerFactory.Remove(channelID, r.removeChannelFromPendingRemovals)
 	if err != nil {
+		r.pendingRemoval[channelID] = false
```
tock-ibm: I think a go routine could be launched here, calling `r.ledgerFactory.Remove`, and when it is finished, it could    remove the channel ID from pendingRemoval, no?

orderer/common/multichannel/registrar.go
```diff
@@ -1025,8 +1063,10 @@ func (r *Registrar) removeSystemChannel() error {
 	}
 
 	// remove system channel resources
-	err := r.ledgerFactory.Remove(systemChannelID)
+	r.pendingRemoval[systemChannelID] = true
+	err := r.ledgerFactory.Remove(systemChannelID, r.removeChannelFromPendingRemovals)
 	if err != nil {
+		r.pendingRemoval[systemChannelID] = false
```
tock-ibm: Remove the system channel could be done synchronously, that would save us a ton of trouble.

common/ledger/blockledger/fileledger/factory.go
```diff
@@ -54,8 +54,8 @@ func (f *fileLedgerFactory) GetOrCreate(channelID string) (blockledger.ReadWrite
 }
 
 // Remove removes an existing ledger and its indexes. This operation
-// is blocking.
-func (f *fileLedgerFactory) Remove(channelID string) error {
+// performs an async deletion.
+func (f *fileLedgerFactory) Remove(channelID string, finishRemove func(string, bool)) error {
```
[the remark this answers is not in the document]
stephyee: As you suggested, I moved the async logic out to the registrar.

common/ledger/blockledger/fileledger/factory.go
```diff
@@ -70,10 +70,15 @@ func (f *fileLedgerFactory) Remove(channelID string) error {
...
-	}
+	go func() {
+		err := f.blkstorageProvider.Drop(channelID)
+		if err != nil {
+			logger.Info(err.Error())
+			finishRemove(channelID, false)
+		} else {
+			finishRemove(channelID, true)
+		}
+	}()
```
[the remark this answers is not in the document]
stephyee: See previous comment.

orderer/common/multichannel/registrar.go
```diff
@@ -60,7 +60,8 @@ type Registrar struct {
 	bccsp              bccsp.BCCSP
 	clusterDialer      *cluster.PredicateDialer
 
-	joinBlockFileRepo *filerepo.Repo
+	joinBlockFileRepo, removeFileRepo *filerepo.Repo
```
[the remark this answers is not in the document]
stephyee: agreed, it'll remain in the factory.

orderer/common/multichannel/registrar.go
```diff
@@ -102,41 +103,53 @@ func NewRegistrar(
...
-	r.init(consenters)
+	r.init(consenters, "")
 
 	r.lock.Lock()
 	defer r.lock.Unlock()
 	r.startChannels()
 }
 
-func (r *Registrar) init(consenters map[string]consensus.Consenter) {
+func (r *Registrar) init(consenters map[string]consensus.Consenter, systemChannelID string) {
```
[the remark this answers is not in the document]
stephyee: removed.

common/ledger/blockledger/fileledger/factory.go
```diff
@@ -70,10 +70,15 @@ func (f *fileLedgerFactory) Remove(channelID string) error {
 		return err
 	}
```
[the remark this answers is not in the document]
stephyee: Reordered.

orderer/common/multichannel/registrar.go
```diff
@@ -188,19 +204,22 @@ func (r *Registrar) discoverLedgers(channelsWithJoinBlock map[string]*cb.Block)
 		if rl.Height() == 0 {
 			if _, ok := channelsWithJoinBlock[channelID]; !ok {
 				logger.Warnf("Channel '%s' has an empty ledger without a join-block, removing it", channelID)
-				if err := r.ledgerFactory.Remove(channelID); err != nil {
+				if err := r.ledgerFactory.Remove(channelID, r.removeChannelFromPendingRemovals); err != nil {
```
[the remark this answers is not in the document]
stephyee: done.

common/ledger/blockledger/fileledger/factory.go
```diff
@@ -115,24 +120,10 @@ func New(directory string, metricsProvider metrics.Provider) (blockledger.Factor
...
-		for _, fileName := range files {
-			channelID := factory.removeFileRepo.FileToBaseName(fileName)
-			err = factory.Remove(channelID)
-			if err != nil {
-				logger.Errorf("Failed to remove channel %s: %s", channelID, err.Error())
-				return nil, err
-			}
-			logger.Infof("Removed channel: %s", channelID)
-		}
-	}
```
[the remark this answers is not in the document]
stephyee: It was removed to the registrar but in the most recent commit it's returned to it's original location.

orderer/common/multichannel/registrar.go
stephyee: Yes most recent commits take this approach now.

common/ledger/blockledger/fileledger/factory.go
stephyee: Moved the goroutine out to the registrar.

common/ledger/blockledger/fileledger/factory.go
stephyee: See previous comment.

orderer/common/multichannel/registrar_test.go
```diff
@@ -938,6 +938,24 @@ func TestRegistrar_JoinChannel(t *testing.T) {
...
+		newLedger(ledgerFactory, "sys-raft-channel", genesisBlockSysRaft)
+
+		registrar := NewRegistrar(localconfig.TopLevel{}, ledgerFactory, mockCrypto(), &disabled.Provider{}, cryptoProvider, nil)
+		registrar.Initialize(mockConsenters)
+
+		registrar.pendingRemoval["some-app-channel"] = true
+		info, err := registrar.JoinChannel("some-app-channel", &cb.Block{}, true)
+		require.EqualError(t, err, "channel does not exist")
+		require.Equal(t, types.ChannelInfo{}, info)
+		joinBlockPath := filepath.Join(tmpdir, "filerepo", "joinblock", "some-app-channel.joinblock")
```
wlahti:
```suggestion
		joinBlockPath := filepath.Join(tmpdir, "pendingops", "joinblock", "some-app-channel.joinblock")
```

wlahti (changes requested):
I'll do a more thorough review shortly, but for now, I noticed one spot in the tests that needs to be updated now that https://github.com/hyperledger/fabric/pull/2047 has been merged.

orderer/common/multichannel/registrar_test.go
```diff
@@ -1543,6 +1565,59 @@ func TestRegistrar_RemoveChannel(t *testing.T) {
...
+
+		require.Nil(t, registrar.GetChain("my-raft-channel"))
+		require.NotContains(t, ledgerFactory.ChannelIDs(), "my-raft-channel")
+		info, err := registrar.JoinChannel("my-raft-channel", genesisBlockAppRaft, true)
+		require.NoError(t, err)
+		require.Equal(t, types.ChannelInfo{Name: "my-raft-channel", URL: "", ClusterRelation: "member", Status: "active", Height: 1}, info)
+		require.NotNil(t, registrar.GetChain("my-raft-channel"))
+		require.Contains(t, ledgerFactory.ChannelIDs(), "my-raft-channel")
+
+		os.RemoveAll(filepath.Join(tmpdir, "filerepo", "remove"))
```
wlahti: And here too...
```suggestion
		os.RemoveAll(filepath.Join(tmpdir, "pendingops, "remove"))
```

orderer/common/multichannel/registrar.go
```diff
@@ -1017,10 +1017,7 @@ func (r *Registrar) removeSystemChannel() error {
 		}
 		if !isChannelMember {
 			logger.Debugf("Not a member of channel %s, removing it", channel)
-			err := r.removeMember(channel, cs)
-			if err != nil {
-				return errors.WithMessagef(err, "failed to remove channel: %s", channel)
-			}
+			r.removeMember(channel, cs)
```
tock-ibm: This will remove the channels asynch. That process may take longer then what it takes to remove the system channel, below. One must not call `r.init(r.consenters)` below before all these channels are deleted. 
Moreover, currently, the lower level `r.ledgerFactory` will not allow concurrent `Remove()` anyway. So all theses goroutines will be executed serially anyway.

I would change the flow of the system channel removal:
1. loop to stop all channels, and collect in a slice the app channels the need to be removed.
2. remove the system channel first, synchronously
3. loop to remove app channels, synchronously. If in the future the ledger will permit concurrent drops, we'll improve on that and do this in parallel.
4. wait for all removals to complete, and only then init all remaining channels, etc.

This way if we fail during system channel removal, a restart will complete the removal of the system channel. If  we fail during the removal of subsequent app channels, we may miss removal of some of the app channels, but that only means they will end up as followers after restart. Not a big deal because the admin can then remove them using the API.

orderer/common/multichannel/registrar_test.go
```diff
@@ -1543,6 +1565,60 @@ func TestRegistrar_RemoveChannel(t *testing.T) {
...
+		require.Contains(t, ledgerFactory.ChannelIDs(), "my-raft-channel")
+
+		registrar.pendingRemoval["my-raft-channel"] = true
+		err = registrar.RemoveChannel("my-raft-channel")
+		require.NoError(t, err)
+
+		// After removing the channel, it no longer exists in the registrar or the ledger
+		require.Nil(t, registrar.GetChain("my-raft-channel"))
+		require.Equal(t, len(ledgerFactory.ChannelIDs()), 1)
+		require.Contains(t, ledgerFactory.ChannelIDs(), "my-raft-channel")
```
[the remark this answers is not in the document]
stephyee: doh 🛠️

orderer/common/multichannel/registrar.go
```diff
@@ -908,17 +916,12 @@ func (r *Registrar) RemoveChannel(channelID string) error {
 	return types.ErrChannelNotExist
 }
```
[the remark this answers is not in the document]
stephyee: 🛠️

orderer/common/multichannel/registrar_test.go
```diff
@@ -1543,6 +1565,60 @@ func TestRegistrar_RemoveChannel(t *testing.T) {
...
+		require.NotContains(t, ledgerFactory.ChannelIDs(), "my-raft-channel")
+		info, err := registrar.JoinChannel("my-raft-channel", genesisBlockAppRaft, true)
+		require.NoError(t, err)
+		require.Equal(t, types.ChannelInfo{Name: "my-raft-channel", URL: "", ClusterRelation: "member", Status: "active", Height: 1}, info)
+		require.NotNil(t, registrar.GetChain("my-raft-channel"))
+		require.Contains(t, ledgerFactory.ChannelIDs(), "my-raft-channel")
+
+		registrar.pendingRemoval["my-raft-channel"] = true
+		err = registrar.RemoveChannel("my-raft-channel")
+		require.NoError(t, err)
```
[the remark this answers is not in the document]
stephyee: 🛠️

orderer/common/multichannel/registrar.go
```diff
@@ -1044,3 +1041,20 @@ func (r *Registrar) removeSystemChannel() error {
 
 	return nil
 }
+
+func (r *Registrar) removeLedgerAsync(channelID string) {
+	removalStatus, ok := r.pendingRemoval[channelID]
+	if removalStatus && ok {
+		return
+	}
```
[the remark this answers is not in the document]
stephyee: correct

orderer/common/multichannel/registrar.go
```diff
@@ -696,6 +698,11 @@ func (r *Registrar) JoinChannel(channelID string, configBlock *cb.Block, isAppCh
 	r.lock.RLock()
 	defer r.lock.RUnlock()
 
+	removalStatus, ok := r.pendingRemoval[channelID]
+	if removalStatus && ok {
+		return types.ChannelInfo{}, types.ErrChannelNotExist
+	}
+
```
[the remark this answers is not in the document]
stephyee: I added the new error type

orderer/common/multichannel/registrar.go
```diff
@@ -102,6 +103,7 @@ func NewRegistrar(
 		callbacks:          callbacks,
 		bccsp:              bccsp,
 		clusterDialer:      clusterDialer,
+		pendingRemoval:     make(map[string]bool),
```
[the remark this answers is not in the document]
stephyee: 🧹

orderer/common/multichannel/registrar.go
```diff
@@ -61,6 +61,7 @@ type Registrar struct {
 	clusterDialer      *cluster.PredicateDialer
 
 	joinBlockFileRepo *filerepo.Repo
+	pendingRemoval    map[string]bool
```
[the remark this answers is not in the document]
stephyee: 🧹

orderer/common/multichannel/registrar.go
```diff
@@ -1044,3 +1041,20 @@ func (r *Registrar) removeSystemChannel() error {
...
+	}
+
+	r.pendingRemoval[channelID] = true
+	go func() {
+		if err := r.ledgerFactory.Remove(channelID); err != nil {
+			r.pendingRemoval[channelID] = false
+			logger.Errorf("Ledger factory failed to remove empty ledger '%s', error: %s", channelID, err)
+			return
+		}
+		delete(r.pendingRemoval, channelID)
```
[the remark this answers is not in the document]
stephyee: Also need the lock to update the value too. Are we opposed to use of the sync.Map instead of map[string]bool?

orderer/common/multichannel/registrar.go
stephyee: 🧹

orderer/common/multichannel/registrar.go
```diff
@@ -1044,3 +1057,15 @@ func (r *Registrar) removeSystemChannel() error {
 
 	return nil
 }
+
+func (r *Registrar) removeLedgerAsync(channelID string) {
+	go func() {
+		r.pendingRemoval.Store(channelID, true)
```
tock-ibm: insert into the map[string]bool before launching the goroutine:
```
pendingRemoval[channelID] = true
go func() { ... //remove from ledger }
```

pendingRemoval is protected by the `r.lock` here.

orderer/common/multichannel/registrar.go
```diff
@@ -1044,3 +1057,15 @@ func (r *Registrar) removeSystemChannel() error {
...
+		if err := r.ledgerFactory.Remove(channelID); err != nil {
+			r.pendingRemoval.Store(channelID, false)
+			logger.Errorf("Ledger factory failed to remove empty ledger '%s', error: %s", channelID, err)
+			return
+		}
+		r.pendingRemoval.Delete(channelID)
```
tock-ibm: just take the lock, remove, and unlock:
```
r.lock.Lock()
delete(r.pendingRemoval,channelID)
r.lock,Unlock()
```

or you can define this 3 liner as a defer in the entrance to the function you give to the goroutine.

orderer/common/multichannel/registrar.go
```diff
@@ -48,6 +48,7 @@ type Registrar struct {
 	lock            sync.RWMutex
 	chains          map[string]*ChainSupport
 	followers       map[string]*follower.Chain
+	pendingRemoval  sync.Map
```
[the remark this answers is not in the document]
stephyee: Thanks, I had mixed feelings about the sync.map vs my previous impl.

orderer/common/multichannel/registrar.go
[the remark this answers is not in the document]
stephyee: As you noted in another comment, sticking with the map[string]bool.

integration/raft/channel_participation_test.go
```diff
@@ -261,9 +261,24 @@ var _ = Describe("ChannelParticipation", func() {
...
+
+			By("submitting transaction to orderer1")
+			env := CreateBroadcastEnvelope(network, peer, "participation-trophy", []byte("hello"))
+			resp, err := ordererclient.Broadcast(network, orderer1, env)
+			Expect(err).NotTo(HaveOccurred())
+			Expect(resp.Status).To(Equal(common.Status_BAD_REQUEST))
 
 			By("listing all channels for orderer1")
-			channelparticipation.List(network, orderer1, []string{"another-participation-trophy"})
+			channelparticipation.List(network, orderer1, []string{"participation-trophy", "another-participation-trophy"})
```
tock-ibm: Is this because the channel is still pending removal? If so this may introduce a race and a flakey test, because if the test thread is slower then the orderer1 process, you won't get to see this transitory state.

integration/raft/channel_participation_test.go
```diff
@@ -261,9 +261,24 @@ var _ = Describe("ChannelParticipation", func() {
...
 			channelparticipation.Remove(network, orderer1, "participation-trophy")
+			Eventually(func() channelparticipation.ChannelInfo {
+				return channelparticipation.ListOne(network, orderer1, "participation-trophy")
+			}, network.EventuallyTimeout).Should(Equal(channelparticipation.ChannelInfo{
+				Name:            "participation-trophy",
+				URL:             "/participation/v1/channels/participation-trophy",
+				Status:          "inactive",
+				ClusterRelation: "none",
+				Height:          0,
+			}))
```
tock-ibm: Is this because the channel is still pending removal? If so this may introduce a race and a flakey test, because if the test thread is slower then the orderer1 process, you won't get to see this transitory state.

orderer/common/multichannel/registrar.go
```diff
@@ -687,6 +696,15 @@ func (r *Registrar) ChannelInfo(channelID string) (types.ChannelInfo, error) {
 		return info, nil
 	}
 
+	_, ok := r.pendingRemoval[channelID]
+	if ok {
+		return types.ChannelInfo{
+			Name:            channelID,
+			ClusterRelation: types.ClusterRelationNone,
```
tock-ibm: I am sorry... I know I said `types.ClusterRelationNone`... but now that I look at it, `types.ClusterRelationFollower` makes more sense. After all, if these were not stopped and made inactive, they would have been followers, right?

orderer/common/multichannel/registrar.go
```diff
@@ -886,6 +908,11 @@ func (r *Registrar) joinSystemChannel(
 func (r *Registrar) RemoveChannel(channelID string) error {
 	r.lock.Lock()
 	defer r.lock.Unlock()
+
+	if removalStatus := r.pendingRemoval[channelID]; removalStatus {
+		return types.ErrChannelPendingRemoval
+	}
```
tock-ibm: please add to the unit tests of channelparticipation/restapi.go a test case in which the mock of the Registrar adds this error code and verify it is treated sensibly. e.g. restapi_test.go L:286, L:403, etc

orderer/common/multichannel/registrar.go
```diff
@@ -1044,3 +1066,18 @@ func (r *Registrar) removeSystemChannel() error {
...
+	r.pendingRemoval[channelID] = true
+
+	go func() {
+		if err := r.ledgerFactory.Remove(channelID); err != nil {
+			r.pendingRemoval[channelID] = false
+			logger.Errorf("Ledger factory failed to remove empty ledger '%s', error: %s", channelID, err)
+			return
+		}
+		r.lock.Lock()
+		r.pendingRemoval[channelID] = true
```
tock-ibm: needs to be 
`r.pendingRemoval[channelID] = false`

or better `delete(r.pendingRemoval, channelID)` to avoid keeping all those removed channels in the map

orderer/common/multichannel/registrar.go
```diff
@@ -1044,3 +1066,18 @@ func (r *Registrar) removeSystemChannel() error {
 
 	return nil
 }
+
+func (r *Registrar) removeLedgerAsync(channelID string) {
+	r.pendingRemoval[channelID] = true
+
+	go func() {
+		if err := r.ledgerFactory.Remove(channelID); err != nil {
+			r.pendingRemoval[channelID] = false
```
tock-ibm, line 1092: this needs to enveloped in the lock unlock as well; 
delete() is better, `delete(r.pendingRemoval, channelID)` to avoid keeping all those removed channels in the map.

integration/raft/channel_participation_test.go
stephyee: Yes since the pending removals still exist in the list API now. Need a better way to wait for the removal to complete in the test.

integration/raft/channel_participation_test.go
stephyee: Yes agreed, see previous comment

orderer/common/multichannel/registrar.go
stephyee: I used false in consideration adding some ongoing retry logic but I'll delete them for now.

stephyee: actually I meant to delete this but copied the wrong line.

orderer/common/multichannel/registrar_test.go
```diff
@@ -1543,6 +1571,119 @@ func TestRegistrar_RemoveChannel(t *testing.T) {
...
+		require.NotNil(t, registrar.GetChain("raft-sys-channel"))
+		require.Contains(t, ledgerFactory.ChannelIDs(), "raft-sys-channel")
+
+		os.RemoveAll(filepath.Join(tmpdir, "pendingops", "remove"))
+
+		err := registrar.RemoveChannel("raft-sys-channel")
+		require.Error(t, err)
+		require.Contains(t, err.Error(), "error removing ledger for system channel raft-sys-channel")
+
+		// After removing the system channel, it no longer exists in the registrar or the ledger
```
tock-ibm: It seems this is the wrong comment for what's going on here

orderer/common/multichannel/registrar_test.go
```diff
@@ -1543,6 +1571,119 @@ func TestRegistrar_RemoveChannel(t *testing.T) {
...
+		require.Error(t, err)
+		require.Contains(t, err.Error(), "failed removing ledger for channel channel-im-not-a-member-of")
+
+		// After removing the system channel, it no longer exists in the registrar or the ledger
+		require.Empty(t, registrar.SystemChannelID())
+		require.Nil(t, registrar.SystemChannel())
+		require.Nil(t, registrar.GetChain("raft-sys-channel"))
+		require.NotContains(t, ledgerFactory.ChannelIDs(), "raft-sys-channel")
+
+		// channels the orderer wasn't a member no longer exist
```
tock-ibm: again, probably cut & paste error, please check all the comments here...

orderer/common/multichannel/registrar_test.go
```diff
@@ -1543,6 +1571,119 @@ func TestRegistrar_RemoveChannel(t *testing.T) {
...
+
+		require.Nil(t, registrar.GetChain("my-raft-channel"))
+		require.NotContains(t, ledgerFactory.ChannelIDs(), "my-raft-channel")
+		info, err := registrar.JoinChannel("my-raft-channel", genesisBlockAppRaft, true)
+		require.NoError(t, err)
+		require.Equal(t, types.ChannelInfo{Name: "my-raft-channel", URL: "", ClusterRelation: "member", Status: "active", Height: 1}, info)
+		require.NotNil(t, registrar.GetChain("my-raft-channel"))
+		require.Contains(t, ledgerFactory.ChannelIDs(), "my-raft-channel")
+
+		// This will force the failure of the underlying factory.Remove() because the pendingops/remove will not exist
```
tock-ibm, line 1711: This looks exactly as if the operation was a success. I understand that the failure is asynch, in another thread, so that the call starts ok. However, if the asynch removal failed, the channel is not "gone" from the ledger... and the .remove pendingops file was not deleted, I assume. So maybe we need to keep a channel that its removal failed in the "pendingRemoval" map, so that an attempt to re-join or re-remove would fail, and that it would show up as pending in the List and ListOne?

orderer/common/multichannel/registrar_test.go
```diff
@@ -1543,6 +1571,119 @@ func TestRegistrar_RemoveChannel(t *testing.T) {
...
+		os.RemoveAll(filepath.Join(tmpdir, "pendingops", "remove"))
+
+		err = registrar.RemoveChannel("my-raft-channel")
+		require.NoError(t, err)
+
+		// After removing the channel, it no longer exists in the registrar or the ledger
```
tock-ibm: but it does exist in the ledger...

orderer/common/multichannel/registrar.go
```diff
@@ -1042,3 +1064,20 @@ func (r *Registrar) removeSystemChannel() error {
...
+
+func (r *Registrar) removeLedgerAsync(channelID string) {
+	r.pendingRemoval[channelID] = true
+
+	go func() {
+		if err := r.ledgerFactory.Remove(channelID); err != nil {
+			r.lock.Lock()
+			delete(r.pendingRemoval, channelID)
+			r.lock.Unlock()
+			logger.Errorf("Ledger factory failed to remove empty ledger '%s', error: %s", channelID, err)
```
tock-ibm: I am thinking (see also comment below in the unit test), that in case of remove failure we need to keep the entry in the map, such that future join or remove to the same channel would fail? this would also alert the admin to the fact that he needs to intervene, by keeping the channel in a status "inactive".

orderer/common/multichannel/registrar.go
```diff
@@ -1000,6 +1020,16 @@ func (r *Registrar) removeSystemChannel() error {
...
+	// remove system channel resources
+	err := r.ledgerFactory.Remove(systemChannelID)
+	if err != nil {
+		return errors.WithMessagef(err, "error removing ledger for system channel %s", r.systemChannelID)
+	}
+
+	// remove system channel references
+	r.systemChannel = nil
+	r.systemChannelID = ""
+
```
tock-ibm: I suggest to move this 
`logger.Infof("Removed system channel: %s", systemChannelID)`
from the bottom to here.

orderer/common/multichannel/registrar.go
```diff
@@ -1015,23 +1045,15 @@ func (r *Registrar) removeSystemChannel() error {
 		}
 		if !isChannelMember {
 			logger.Debugf("Not a member of channel %s, removing it", channel)
-			err := r.removeMember(channel, cs)
+			err := r.ledgerFactory.Remove(channel)
 			if err != nil {
-				return errors.WithMessagef(err, "failed to remove channel: %s", channel)
+				return errors.WithMessagef(err, "failed removing ledger for channel %s", channel)
```
tock-ibm: Can we try all of them and not stop on the first? We can return an error at the end if one or more failed, but how about trying to remove them all? this would at least stop all the ones that needs stopping, and won't keep the server with channels running when they are still expecting the system channel to be there.

orderer/common/multichannel/registrar.go
```diff
@@ -696,6 +698,11 @@ func (r *Registrar) JoinChannel(channelID string, configBlock *cb.Block, isAppCh
 	r.lock.RLock()
 	defer r.lock.RUnlock()
 
+	removalStatus, ok := r.pendingRemoval.Load(channelID)
```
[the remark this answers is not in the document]
stephyee: using a map[string]bool again

orderer/common/multichannel/registrar.go
```diff
@@ -886,6 +893,12 @@ func (r *Registrar) joinSystemChannel(
 func (r *Registrar) RemoveChannel(channelID string) error {
 	r.lock.Lock()
 	defer r.lock.Unlock()
+
+	removalStatus, ok := r.pendingRemoval.Load(channelID)
+	if ok && removalStatus.(bool) {
```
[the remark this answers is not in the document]
stephyee: using a map[string]bool again

orderer/common/multichannel/registrar.go
```diff
@@ -1002,10 +1008,26 @@ func (r *Registrar) removeSystemChannel() error {
 	r.systemChannel.Halt()
 	delete(r.chains, systemChannelID)
 
+	pendingRemovals := []string{}
+
 	// halt all application channels
 	for channel, cs := range r.chains {
 		cs.Halt()
+		pendingRemovals = append(pendingRemovals, channel)
+	}
```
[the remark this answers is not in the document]
stephyee: Simplified as you suggested

orderer/common/multichannel/registrar.go
stephyee: 🧹

orderer/common/multichannel/registrar.go
stephyee: wrapped in the lock

orderer/common/multichannel/registrar.go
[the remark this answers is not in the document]
stephyee: Added pending removals to both API results

integration/raft/channel_participation_test.go
stephyee: Updated the channelparticipation list helper to return all the channels to do a exact comparison

orderer/common/multichannel/registrar.go
stephyee: No worries, I updated it to use Follower.

orderer/common/multichannel/registrar.go
stephyee: done. @tock-ibm do you agree on the statuses returned for removals and joins?

orderer/common/multichannel/registrar.go
stephyee: As you noted later, it's better to keep these failed removals in the map to indicate failure on further attempts to join or remove the same channel.

orderer/common/multichannel/registrar_test.go
stephyee: 🧹  removed.

orderer/common/multichannel/registrar_test.go
stephyee: 🧹  removed.

orderer/common/multichannel/registrar_test.go
stephyee: @tock-ibm I decided to use the bool value to determine a "pending" vs "failed" removal. Thoughts?

orderer/common/multichannel/registrar_test.go
stephyee: cleaned up bad comment.

orderer/common/multichannel/registrar.go
stephyee: 🧹

orderer/common/multichannel/registrar.go
stephyee: Yes now it attempts all the removals before erring

orderer/common/multichannel/registrar.go
stephyee: See updated logic

orderer/common/channelparticipation/restapi.go
```diff
@@ -289,6 +290,8 @@ func (h *HTTPHandler) serveRemove(resp http.ResponseWriter, req *http.Request) {
 		h.sendResponseNotAllowed(resp, errors.WithMessage(err, "cannot remove"), http.MethodGet)
 	case types.ErrChannelNotExist:
 		h.sendResponseJsonError(resp, http.StatusNotFound, errors.WithMessage(err, "cannot remove"))
+	case types.ErrChannelPendingRemoval, types.ErrChannelRemoval:
+		h.sendResponseJsonError(resp, http.StatusConflict, errors.WithMessage(err, "cannot remove"))
```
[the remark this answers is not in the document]
stephyee: Yes I think that's better. 🧹

orderer/common/multichannel/registrar.go
```diff
@@ -1014,31 +1057,44 @@ func (r *Registrar) removeSystemChannel() error {
...
+			delete(r.chains, channel)
 		}
 	}
 
-	// remove system channel resources
-	err := r.ledgerFactory.Remove(systemChannelID)
-	if err != nil {
-		return errors.WithMessagef(err, "error removing ledger for system channel %s", r.systemChannelID)
+	if len(failedRemovals) > 0 {
+		return fmt.Errorf("failed removing %s ledger for channel(s)", strings.Join(failedRemovals, ", "))
```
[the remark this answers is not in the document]
stephyee: Much better, thanks. 🧹

orderer/common/multichannel/registrar.go
```diff
@@ -1014,31 +1057,44 @@ func (r *Registrar) removeSystemChannel() error {
...
 	return nil
 }
+
+func (r *Registrar) removeLedgerAsync(channelID string) {
+	r.pendingRemoval[channelID] = true
+
+	go func() {
+		if err := r.ledgerFactory.Remove(channelID); err != nil {
+			r.lock.Lock()
+			r.pendingRemoval[channelID] = false
```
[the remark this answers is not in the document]
stephyee: 💬  added.

orderer/common/channelparticipation/restapi.go
```diff
@@ -254,8 +254,9 @@ func (h *HTTPHandler) sendJoinError(err error, resp http.ResponseWriter) {
 		// The client is trying to join an app-channel that exists, but the system channel does not;
 		// The client is trying to join the system-channel, and it exists. GET & DELETE are allowed on the channel.
 		h.sendResponseNotAllowed(resp, errors.WithMessage(err, "cannot join"), http.MethodGet, http.MethodDelete)
-	case types.ErrAppChannelsAlreadyExists:
+	case types.ErrAppChannelsAlreadyExists, types.ErrChannelPendingRemoval:
```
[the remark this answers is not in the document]
stephyee: IIUC it would not necessarily always be the server's fault. It might be something wrong with the network config. I did update both to return Conflict for pending removals and bad request for removal failures.

orderer/common/types/errors.go
```diff
@@ -21,3 +21,9 @@ var ErrAppChannelsAlreadyExists = errors.New("application channels already exist
 
 // ErrChannelNotExist is returned when trying to remove or list a channel that does not exist
 var ErrChannelNotExist = errors.New("channel does not exist")
+
+// ErrChannelPendingRemoval is returned when trying to remove or list a channel that is being removed.
+var ErrChannelPendingRemoval = errors.New("channel pending removal")
+
+// ErrChannelRemoval is returned when a removal attempt failure has been recorded.
+var ErrChannelRemoval = errors.New("channel removal failure")
```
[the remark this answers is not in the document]
stephyee: Okay with me.

orderer/common/multichannel/registrar_test.go
stephyee: Opened https://jira.hyperledger.org/browse/FAB-18315 as a follow-up

tock-ibm:
Looks good. I think you can upgrade and submit for review to a maintainer.

stephyee:
@tock-ibm will do, first I need to clean up some flakiness with the int test.

orderer/common/channelparticipation/restapi.go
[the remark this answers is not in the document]
tock-ibm: The error response codes look good now, thanks.

Let me just clarify what I meant (at the risk of stating the obvious). If the "Remove" fails because of any reason that is the fault of the server, then I think it is a 5xx. In contrast, if the request caused it, or the user could have avoided it by altering the request, it is 4xx. So even if the "Remove" failed because of a networking problem in the "backend" of the server, say an NFS problem, or a cluster communication problem, it translates to 5xx. The client doesn't care that the "server" is really complex distributed system.

integration/raft/channel_participation_test.go
```diff
@@ -261,9 +264,27 @@ var _ = Describe("ChannelParticipation", func() {
...
+						URL:  "/participation/v1/channels/another-participation-trophy",
+					},
+				},
+			}))
+
+			By("submitting transaction to orderer1")
+			env := CreateBroadcastEnvelope(network, peer, "participation-trophy", []byte("hello"))
+			resp, err := ordererclient.Broadcast(network, orderer1, env)
+			Expect(err).NotTo(HaveOccurred())
+			Expect(resp.Status).To(Equal(common.Status_BAD_REQUEST))
```
tock-ibm, line 283: This looks good. 

I would add a test case for removing a channel and joining it again. But let's do this in separate PR, rather then preventing this one from merging. I created this one: https://jira.hyperledger.org/browse/FAB-18319

tock-ibm (dismissed):

tock-ibm:
Looks good to me. 
I added one minor comment for adding a test case for remove, but lets tackle it in another PR.

orderer/common/multichannel/registrar_test.go
```diff
@@ -1543,6 +1586,146 @@ func TestRegistrar_RemoveChannel(t *testing.T) {
...
+
+		consenter.IsChannelMemberReturns(true, nil)
+		registrar := NewRegistrar(config, ledgerFactory, mockCrypto(), &disabled.Provider{}, cryptoProvider, dialer)
+		registrar.Initialize(mockConsenters)
+
+		require.Nil(t, registrar.GetChain("my-raft-channel"))
+		require.NotContains(t, ledgerFactory.ChannelIDs(), "my-raft-channel")
+		info, err := registrar.JoinChannel("my-raft-channel", genesisBlockAppRaft, true)
+		require.NoError(t, err)
+		require.Equal(t, types.ChannelInfo{Name: "my-raft-channel", URL: "", ClusterRelation: "member", Status: "active", Height: 1}, info)
```
wlahti: FYI: ClusterRelation "member" was just renamed to "consenter" in master. You'll need to rebase and update this test and any other tests/code that I may have missed that reference "member" (or types.ClusterRelationMember but I don't think that applies to this PR).

wlahti (changes requested):
This PR now needs a rebase/rename for the "member" -> "consenter" change I noted. Sorry! That one was merged surprisingly quickly...

orderer/common/multichannel/registrar_test.go
stephyee: 🧹

wlahti (approved):
Looks good to me. Thanks, Tiffany!

tock-ibm:
Looks good to me.

orderer/common/multichannel/registrar.go
```diff
@@ -96,6 +100,7 @@ func NewRegistrar(
 		config:             config,
 		chains:             make(map[string]*ChainSupport),
 		followers:          make(map[string]*follower.Chain),
+		pendingRemoval:     make(map[string]bool),
```
yacovm, line 103: why not map[string]struct ?

stephyee: @yacovm that has been suggested by Yoav in a follow-up https://github.com/hyperledger/fabric/pull/2027#discussion_r516476013

orderer/common/multichannel/registrar.go
```diff
@@ -1014,31 +1058,44 @@ func (r *Registrar) removeSystemChannel() error {
...
-	logger.Infof("Removed system channel: %s", systemChannelID)
-
 	return nil
 }
+
+func (r *Registrar) removeLedgerAsync(channelID string) {
+	r.pendingRemoval[channelID] = true
+
+	go func() {
+		if err := r.ledgerFactory.Remove(channelID); err != nil {
```
yacovm, line 1090: can't we make this simpler? 

```
err := r.ledgerFactory.Remove(channelID)
r.lock.Lock()
defer r.lock.Unlock()
if err != nil { 
.... 
}
```

orderer/common/multichannel/registrar.go
```diff
@@ -886,6 +915,15 @@ func (r *Registrar) joinSystemChannel(
 func (r *Registrar) RemoveChannel(channelID string) error {
 	r.lock.Lock()
 	defer r.lock.Unlock()
+
+	if removalStatus, ok := r.pendingRemoval[channelID]; ok {
+		if removalStatus {
+			return types.ErrChannelPendingRemoval
+		} else {
+			return types.ErrChannelRemovalFailure
```
yacovm, line 923: Why can't we "try again"?

yacovm (approved):
Let's address in another PR

orderer/common/multichannel/registrar.go
stephyee: very much so yes - I'll handle in a follow-up.

orderer/common/multichannel/registrar.go
stephyee: Maybe https://github.com/hyperledger/fabric/pull/2027#discussion_r515592791 this should only apply to the Joining a channel that experienced a removal failure.

PR #2029, opened 2020-10-20
Deduplicate orderer server TLS root CAs

yacovm:
When the orderer TLS root CAs are updated, an aggregation of all root TLS CA certificates over all channels is injected into the PredicateDialer.
Then, upon client TLS handshake, a fresh TLS config object is built (for orthogonal purposes), however the operation entails parsing of all
root CAs all over again.

In case the orderer is part of too many channels, this induces a high and unnecessary processing overhead.

This commit simply performs a deduplication of the bespoken TLS root CA certificates prior to updating the root CAs.

Change-Id: I21b2ed483afc9595c2ccd7fbe9ec0cf475cc5f62

### DIFF
--- a/orderer/common/server/main.go
+++ b/orderer/common/server/main.go
@@ -1007,15 +1007,26 @@ func (mgr *caManager) updateClusterDialer(
 
 	// Iterate over all orderer root CAs for all chains and add them
 	// to the root CAs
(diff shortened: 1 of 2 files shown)


jyellick (approved):
Looks good to me

yacovm:

PR #2031, opened 2020-10-20
Log TLS hanshake duration

yacovm:
This commit adds logging to the duration of the TLS handshakes, both client and server.

Change-Id: I5f630a62362e2131882542c78d3f321aed7fc2d2

### DIFF
--- a/core/comm/creds.go
+++ b/core/comm/creds.go
@@ -13,12 +13,16 @@ import (
 	"errors"
 	"net"
 	"sync"
+	"time"
 
 	"github.com/hyperledger/fabric/common/flogging"
 	"google.golang.org/grpc/credentials"
 )
 
 var (
+	// Logger for TLS client connections
+	tlsClientLogger = flogging.MustGetLogger("comm.tls")
+
 	ClientHandshakeNotImplError = errors.New("core/comm: Client handshakes" +
 		"are not implemented with serverCreds")
 	OverrrideHostnameNotSupportedError = errors.New(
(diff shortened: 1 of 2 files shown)


ale-linux (dismissed):
Why only in 1.4?

core/comm/creds.go
```diff
@@ -148,7 +155,16 @@ func (dtc *DynamicClientCredentials) latestConfig() *tls.Config {
 }
 
 func (dtc *DynamicClientCredentials) ClientHandshake(ctx context.Context, authority string, rawConn net.Conn) (net.Conn, credentials.AuthInfo, error) {
-	return credentials.NewTLS(dtc.latestConfig()).ClientHandshake(ctx, authority, rawConn)
+	l := tlsClientLogger.With("remote address", rawConn.RemoteAddr().String())
+	creds := credentials.NewTLS(dtc.latestConfig())
+	start := time.Now()
+	conn, auth, err := creds.ClientHandshake(ctx, authority, rawConn)
+	if err != nil {
+		l.Debugf("Client TLS handshake failed after %s with error: %s", time.Since(start), err)
```
ale-linux: Actually, isn't this l.Errorf?

yacovm: whoops, that's what you get when you code so late

yacovm:
> Why only in 1.4?

I will push other branches too

ale-linux (dismissed):

core/comm/producer.go
```diff
@@ -14,10 +14,13 @@ import (
 	"time"
 
 	"github.com/hyperledger/fabric/common/flogging"
+
 	"google.golang.org/grpc"
 )
 
-var logger = flogging.MustGetLogger("ConnProducer")
+var (
+	connProdLogger = flogging.MustGetLogger("ConnProducer")
```
denyeart: We use lowercase loggers everywhere else.

yacovm: Well but renaming the logger is not the purpose of this pull request, I am just changing the variable name

yacovm: I left it as is, I just renamed the variable name.

I removed the file from the change set. 

Feel free to change the logger name in a separate pull request if you feel this is needed

denyeart (approved):

PR #2036, opened 2020-10-21
Deduplicate orderer server TLS root CAs

yacovm:
When the orderer TLS root CAs are updated, an aggregation of all root TLS CA certificates over all channels is injected into the PredicateDialer.
Then, upon client TLS handshake, a fresh TLS config object is built (for orthogonal purposes), however the operation entails parsing of all
root CAs all over again.

In case the orderer is part of too many channels, this induces a high and unnecessary processing overhead.

This commit simply performs a deduplication of the bespoken TLS root CA certificates prior to updating the root CAs.

Change-Id: I21b2ed483afc9595c2ccd7fbe9ec0cf475cc5f62

### DIFF
--- a/orderer/common/server/main_test.go
+++ b/orderer/common/server/main_test.go
@@ -5,7 +5,6 @@ package server
 
 import (
 	"fmt"
-	"github.com/hyperledger/fabric/orderer/common/onboarding"
 	"io/ioutil"
 	"net"
 	"net/http"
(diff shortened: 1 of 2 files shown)


orderer/common/server/main_test.go
```diff
@@ -643,6 +643,32 @@ func TestUpdateTrustedRoots(t *testing.T) {
 	grpcServer.Listener().Close()
 }
 
+func TestRootServerCertAggregation(t *testing.T) {
```
ale-linux, line 646: I was too lazy to inspect the test artefacts here - how do you ensure that dedup takes place?

yacovm: We have 2 channels: **foo** and **bar**. They both contain the same root TLS CA of **ca1** which is picked uniformly at random.
Additionally we have two "local config" TLS root CAs which are the same root CA **ca2**, also picked uniformly at random.

I update the dialer and then probe its config and ensure it contains only 2 entries - one of each root TLS CA.

ale-linux: ack - thanks Yacov!

ale-linux:
backport?

ale-linux (approved):

yacovm:
> backport?

You just merged the last of [them](https://github.com/hyperledger/fabric/pulls?q=Deduplicate+orderer+server+TLS+root+CAs).

PR #2044, opened 2020-10-26
Return errors in rollback/rebuild commands when using snapshots

manish-sethi:

#### Type of change
- Bug fix

#### Description
- Returns errors in the rollback and rebuild-dbs commands if any channel is bootstrapped from snapshot.
- Checks for upgrade format before dropping the dbs

#### Additional details
see [FAB-17800](https://jira.hyperledger.org/browse/FAB-17800

### DIFF
--- a/internal/peer/node/rebuild_dbs_test.go
+++ b/internal/peer/node/rebuild_dbs_test.go
@@ -7,15 +7,9 @@ SPDX-License-Identifier: Apache-2.0
 package node
 
 import (
-	"io/ioutil"
 	"os"
-	"path"
-	"path/filepath"
 	"testing"
 
-	"github.com/hyperledger/fabric/core/config"
-	"github.com/hyperledger/fabric/core/ledger/kvledger"
-	"github.com/hyperledger/fabric/internal/fileutil"
 	"github.com/spf13/viper"
 	"github.com/stretchr/testify/require"
 )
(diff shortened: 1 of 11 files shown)


internal/peer/node/rebuild_dbs_test.go
```diff
@@ -26,33 +20,9 @@ func TestRebuildDBsCmd(t *testing.T) {
...
-		require.NoError(t, ioutil.WriteFile(path.Join(dbPath, "dummyfile.txt"), []byte("this is a dummy file for test"), 0644))
-	}
-
-	// check dbs exist before upgrade
-	for _, dbPath := range dbPaths {
-		_, err := os.Stat(dbPath)
-		require.False(t, os.IsNotExist(err))
-	}
-
+	// this should return an error as no ledger has been set up
```
denyeart, line 23: The test name TestRebuildDBsCmd() implies a positive test, but it is actually a negative test to verify an error, should we update the test name to indicate the scenario?

internal/peer/node/rollback_test.go
```diff
@@ -26,7 +26,7 @@ func TestRollbackCmd(t *testing.T) {
 		args := []string{"-c", "ch1", "-b", "10"}
 		cmd.SetArgs(args)
 		err := cmd.Execute()
-		expectedErr := "ledgerID [ch1] does not exist"
-		require.Equal(t, expectedErr, err.Error())
+		// this should return an error as no ledger has been set up
+		require.Contains(t, err.Error(), "error while checking if any ledger has been bootstrapped from snapshot")
```
denyeart, line 30: The test name TestRollbackCmd() implies a positive test, but it is actually a negative test to verify an error, should we update the test name to indicate the scenario?

internal/peer/node/upgrade_dbs_test.go
```diff
@@ -21,5 +21,5 @@ func TestUpgradeDBsCmd(t *testing.T) {
 	defer os.RemoveAll(testPath)
 
 	cmd := upgradeDBsCmd()
-	require.NoError(t, cmd.Execute())
+	require.EqualError(t, cmd.Execute(), "the data format is already up to date. No upgrade is required")
```
denyeart, line 24: The test name TestUpgradeDBsCmd() implies a positive test, but it is actually a negative test to verify an error, should we update the test name to indicate the scenario?

internal/peer/node/rebuild_dbs_test.go
manish-sethi: The unit test at this level is all about whether the code in this package invokes the actual code in the kvledger package. So, with that intention it's still a positive test. Checking for error case rather helps minimizing the dependency on the kvledger package as it does not need to know internal details of kvledger package. Same for the below comments as well.

denyeart: Ok, given the good code comments, I agree.

denyeart (approved):

PR #2047, opened 2020-10-26
Rename channel particicpation CFT filerepo directory to pendingops

jcastrence:
#### Type of change

#### Description

- Renamed the filerepo directory associated with storing pending operations for channel participation CFT to pendingops for better context

#### Related issues

[FAB-18280](https://jira.hyperledger.org/browse/FAB-18280)

### DIFF
--- a/common/ledger/blockledger/fileledger/factory_test.go
+++ b/common/ledger/blockledger/fileledger/factory_test.go
@@ -61,7 +61,7 @@ func TestBlockStoreProviderErrors(t *testing.T) {
 		dir, err := ioutil.TempDir("", "fileledger")
 		require.NoError(t, err, "Error creating temp dir: %s", err)
 		defer os.RemoveAll(dir)
-		fileRepo, err := filerepo.New(filepath.Join(dir, "filerepo"), "remove")
+		fileRepo, err := filerepo.New(filepath.Join(dir, "pendingops"), "remove")
 		require.NoError(t, err, "Error creating temp file repo: %s", err)
 
 		t.Run("ledger doesn't exist", func(t *testing.T) {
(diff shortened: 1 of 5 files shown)


common/ledger/blockledger/fileledger/factory_test.go
```diff
@@ -222,7 +222,7 @@ func TestRemove(t *testing.T) {
 		err = f.Remove("foo")
 		require.EqualError(t, err, "oogie")
 
-		dest := filepath.Join(dir, "filerepo", "remove", "foo.remove")
+		dest := filepath.Join(dir, "pendingops", "remove", "foo.remove")
 		_, err = os.Stat(dest)
 		require.NoError(t, err, "Expected foo.remove to exist")
 	})
```
wlahti, line 228: Missed the rename in the test case right below this one.

wlahti (commented):
Almost got all of them on your first pass. :) Looks like you missed the one I noted in one of the files and then also [here](https://github.com/hyperledger/fabric/blob/master/orderer/common/server/etcdraft_test.go#L163).

common/ledger/blockledger/fileledger/factory_test.go
jcastrence: Thanks for catching that

jcastrence: Fixed

lindluni (approved):

common/ledger/blockledger/fileledger/factory_test.go
```diff
@@ -222,15 +222,15 @@ func TestRemove(t *testing.T) {
...
 
-	t.Run("saving to file repo fails", func(t *testing.T) {
+	t.Run("saving to pending ops fails", func(t *testing.T) {
```
wlahti, line 230: Super nit: this didn't need to change and confuses things slightly but it doesn't matter. :) We're still using the filerepo as the underlying implementation (and that's what's failing to save for this test), it's just the public naming of the folder that changed.

wlahti (approved):
One extremely minor nit in the naming of a test case but otherwise looks good. Thanks, Julian!

PR #2050, opened 2020-10-27
FAB-18291 Ch.Part.API: BlockPuller that can be stopped, improve follower Halt() responsiveness

tock-ibm:
When a follower is started with a block that has bad endpoints,
it will cause the internal BlockPuller to retry a lot of times.
Calling Halt() on said follower will block until the BlockPuller
gives up, which can take a very long time. This causes an attempt
to Remove that channel via the participation API to block for too long.

This commit improves the BlockPuller by adding a channel to signal
it to stop retrying and exit. It also integrates with the
follower.Chain to make it more responsive to Halt().

Change-Id: I49354d94a28402ce69a469689c12ea595b3e0a7e

#### Type of change
- Bug fix

#### Related issues
FAB-18291

### DIFF
--- a/orderer/common/cluster/deliver.go
+++ b/orderer/common/cluster/deliver.go
@@ -40,6 +40,12 @@ type BlockPuller struct {
 	Dialer              Dialer
 	VerifyBlockSequence BlockSequenceVerifier
 	Endpoints           []EndpointCriteria
+
+	// A 'stopper' goroutine may signal the go-routine servicing PullBlock & HeightsByEndpoints to stop by closing this
+	// channel. Note: all methods of the BlockPuller must be serviced by a single goroutine, it is not thread safe.
+	// It is the responsibility of the 'stopper' not to close the channel more then once.
+	StopChannel chan struct{}
+
 	// Internal state
 	stream       *ImpatientStream
 	blockBuff    []*common.Block
@@ -100,7 +106,11 @@ func (p *BlockPuller) PullBlock(seq uint64) *common.Block {
 			p.Logger.Errorf("Failed pulling block [%d]: retry count exhausted(%d)", seq, p.MaxPullBlockRetries)
 			return nil
 		}
-		time.Sleep(p.RetryTimeout)
+
+		if waitOnStop(p.RetryTimeout, p.StopChannel) {
+			p.Logger.Info("Received a stop signal")
+			return nil
+		}
 	}
 }
 
@@ -124,6 +134,16 @@ func (p *BlockPuller) UpdateEndpoints(endpoints []EndpointCriteria) {
 	p.disconnect()
 }
 
+// waitOnStop waits duration, but returns immediately with true if the stop channel fires first.
+func waitOnStop(duration time.Duration, stop <-chan struct{}) bool {
+	select {
+	case <-stop:
+		return true
+	case <-time.After(duration):
+		return false
+	}
+}
+
 func (p *BlockPuller) tryFetchBlock(seq uint64) *common.Block {
 
 	block := p.popBlock(seq)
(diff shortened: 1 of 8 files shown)


orderer/common/follower/follower_chain.go
```diff
@@ -172,7 +172,7 @@ func NewChain(
 
 		// Check the block puller creation function once before we start the follower. This ensures we can extract
 		// the endpoints from the join-block.
-		puller, err := blockPullerFactory.BlockPuller(joinBlock)
+		puller, err := blockPullerFactory.BlockPuller(joinBlock, make(chan struct{}))
```
wlahti: Nit: why not just pass nil here if we don't expect to use the channel?

wlahti (dismissed):
Works for me. Manually verified this resolves the issue that led me to open the JIRA item.

orderer/common/cluster/deliver.go
```diff
@@ -40,6 +40,12 @@ type BlockPuller struct {
 	Dialer              Dialer
 	VerifyBlockSequence BlockSequenceVerifier
 	Endpoints           []EndpointCriteria
+
+	// A 'stopper' goroutine may signal the go-routine servicing PullBlock & HeightsByEndpoints to stop by closing this
+	// channel. Note: all methods of the BlockPuller must be serviced by a single goroutine, it is not thread safe.
+	// It is the responsibility of the 'stopper' not to close the channel more then once.
+	StopChannel chan struct{}
```
yacovm, line 47: The way of externally closing an internal channel is somewhat crude. I think that we should maintain the channel ourselves from within this struct, that is, to close it in a "Close" instance method.

orderer/common/cluster/deliver.go
```diff
@@ -100,7 +106,13 @@ func (p *BlockPuller) PullBlock(seq uint64) *common.Block {
 			p.Logger.Errorf("Failed pulling block [%d]: retry count exhausted(%d)", seq, p.MaxPullBlockRetries)
 			return nil
 		}
-		time.Sleep(p.RetryTimeout)
+
+		select {
```
yacovm: This is a recurring pattern (both here and below) so can we not simply extract it to a method, and then add in the for loop a condition of isClosed() that simply probes if the channel is closed? Much more appealing to the eye

orderer/common/cluster/deliver.go
tock-ibm: It is not necessarily an internal channel. It is possible to give the BlockPuller an external channel, in the case of the follower, the channel that signals the follower itself to stop. This way that channel is closed when someone calls follower.Halt() and it signals both the follower and the block puller it holds to stop retrying and exit. This is fine because in the follower we have only a single goroutine that goes in and out of the block puller.

orderer/common/cluster/deliver.go
tock-ibm: Done

orderer/common/cluster/deliver.go
```diff
@@ -100,7 +106,11 @@ func (p *BlockPuller) PullBlock(seq uint64) *common.Block {
 			p.Logger.Errorf("Failed pulling block [%d]: retry count exhausted(%d)", seq, p.MaxPullBlockRetries)
 			return nil
 		}
-		time.Sleep(p.RetryTimeout)
+
+		if waitOnStop(p.RetryTimeout, p.StopChannel) {
+			p.Logger.Warn("Received a stop signal")
```
yacovm: why is it warning? it's fine to stop it. i think it should be info

tock-ibm: done

yacovm (approved):

PR #2056, opened 2020-10-28
Add function GetChannelConfig in cscc and use in peer client

manish-sethi:

#### Type of change
- New feature

#### Description
CSCC has a function GetConfigBlock that is used by the peer client to extract the orderer endpoints. A peer bootstrapped from a snapshot may not have the config block until next channel config update is committed. This commit introduces a function a cscc to return channel config and makes use of this in the peer client.

#### Additional details
Though, in general we do not encourage to add functions in these built-in chanincodes, but here we cannot modify the existing function GetConfigBlock as some external clients may be using this and it will break the backward compatibility. However, the intention is to encourage the users to use this new API to pave the way for the removal of the existing API.

#### Related issues
[FAB-18300](https://jira.hyperledger.org/browse/FAB-18300)

### DIFF
--- a/core/aclmgmt/resources/resources.go
+++ b/core/aclmgmt/resources/resources.go
@@ -51,6 +51,7 @@ const (
 	Cscc_JoinChainBySnapshot  = "cscc/JoinChainBySnapshot"
 	Cscc_JoinBySnapshotStatus = "cscc/JoinBySnapshotStatus"
 	Cscc_GetConfigBlock       = "cscc/GetConfigBlock"
+	Cscc_GetChannelConfig     = "cscc/GetChannelConfig"
 	Cscc_GetChannels          = "cscc/GetChannels"
 
 	//Peer resources
--- a/internal/peer/chaincode/common_test.go
+++ b/internal/peer/chaincode/common_test.go
@@ -20,16 +20,12 @@ import (
 	"github.com/golang/protobuf/proto"
 	cb "github.com/hyperledger/fabric-protos-go/common"
 	pb "github.com/hyperledger/fabric-protos-go/peer"
-	"github.com/hyperledger/fabric/bccsp/factory"
 	"github.com/hyperledger/fabric/bccsp/sw"
 	"github.com/hyperledger/fabric/common/policydsl"
 	"github.com/hyperledger/fabric/core/config/configtest"
-	"github.com/hyperledger/fabric/internal/configtxgen/encoder"
-	"github.com/hyperledger/fabric/internal/configtxgen/genesisconfig"
 	"github.com/hyperledger/fabric/internal/peer/chaincode/mock"
 	"github.com/hyperledger/fabric/internal/peer/common"
 	"github.com/hyperledger/fabric/internal/pkg/identity"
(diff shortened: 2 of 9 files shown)


internal/peer/chaincode/common_test.go
```diff
@@ -159,12 +159,13 @@ func TestGetOrdererEndpointFromConfigTx(t *testing.T) {
...
-	pgen := encoder.New(config)
-	genesisBlock := pgen.GenesisBlockForChannel(mockchain)
+	profile := genesisconfig.Load(genesisconfig.SampleInsecureSoloProfile, configtest.GetDevConfigDir())
+	channelGroup, err := encoder.NewChannelGroup(profile)
+	require.NoError(t, err)
+	channelConfig := &cb.Config{ChannelGroup: channelGroup}
 
 	mockResponse := &pb.ProposalResponse{
-		Response:    &pb.Response{Status: 200, Payload: protoutil.MarshalOrPanic(genesisBlock)},
+		Response:    &pb.Response{Status: 200, Payload: protoutil.MarshalOrPanic(channelConfig)},
```
denyeart: The final line of the test is:
`require.Equal(t, len(ordererEndpoints), 0)`

Is there a test that actually returns one or more orderer endpoints from the returned config, to be sure the client can still retrieve them?

core/aclmgmt/resources/resources.go
```diff
@@ -51,6 +51,7 @@ const (
 	Cscc_JoinChainBySnapshot  = "cscc/JoinChainBySnapshot"
 	Cscc_JoinBySnapshotStatus = "cscc/JoinBySnapshotStatus"
 	Cscc_GetConfigBlock       = "cscc/GetConfigBlock"
+	Cscc_GetChannelConfig     = "cscc/GetChannelConfig"
```
denyeart, line 54: I grepped "GetConfigBlock" to ensure we make all the associated updates.
It looks like you got the code references, but these other test, doc, and sample files should also be updated when introducing a new acl resource:
integration/e2e/acl_test.go
docs/source/access_control.md
sampleconfig/configtx.yaml

internal/peer/chaincode/common_test.go
manish-sethi: Not sure why but this test is in a different package from the source and yes, does not seem to do much. I have moved this to the right package and enhanced the test (and during this, I hit a bug related to suppression of error in the existing code).

core/aclmgmt/resources/resources.go
manish-sethi: 1) Added a call in acl_test.go (BTW, I noticed only GetConfigBlock from cscc. I guess that this file needs a revisit)
2) This grep was just a reference as an example. Do you know if we are documenting these APIs somewhere or only end CLI is documented?
3) I am not sure if we are consistently updating this (I see many missing here as well), but added an entry for this new function

denyeart: 2. I believe only the end CLI is documented.

denyeart (approved):

PR #2067, opened 2020-11-02
Optimize importing snapshot data into pvtdatastore

manish-sethi:

#### Type of change
- New feature

#### Description
The pvtdata hashes in the snapshot files are maintained in the sort order of key-hashes. However, the data in pvtdata store is primarily maintained as group by block number. This causes inefficiencies during importing the data in the pvtdata store primarily because without any optimization, we end up performing write for each data item independently.

This would not have been as bad as such, if we write to leveldb with async I/O, however, due to a bug previously observed in leveldb, we always perform every write to leveldb as a sync write. This makes the overall data loading into pvtdata store very poor performing because of significantly high numbers of fsyncs, as observed in the system performance tests.

In this commit, we adopt to a different algorithm where in we sort the snapshot data by block number in a temporary space and group by desired attributes for different entry types in memory and finally write the entries in pvtdata store in batches.

### DIFF
--- a/common/ledger/blkstorage/blockindex.go
+++ b/common/ledger/blkstorage/blockindex.go
@@ -38,7 +38,7 @@ var (
 	indexSavePointKey              = []byte(indexSavePointKeyStr)
 	errIndexSavePointKeyNotPresent = errors.New("NoBlockIndexed")
 	errNilValue                    = errors.New("")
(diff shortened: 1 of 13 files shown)


denyeart (approved):
Looks good to me.
I also did my own tests against the PR, joining a peer from snapshot with eligible private data, ineligible private data, and expired private data. Everything reconciled as expected.

PR #2069, opened 2020-11-02
[FAB-18308] Add RSA support back

filipeandre:
#### Make it possible to upgrade from a ledger with rsa certificates

- Bug fix

#### Description

It's not possible to upgrade from fabric 1.4 using an existing ledger with rsa certificates.

#### Related issues

[FAB-18308 ](https://jira.hyperledger.org/projects/FAB/issues/FAB-18308)

### DIFF
--- a/bccsp/bccsp_test.go
+++ b/bccsp/bccsp_test.go
@@ -21,6 +21,7 @@ import (
 	"strings"
 	"testing"
 
+	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 )
 
@@ -44,6 +45,23 @@ func TestAESOpts(t *testing.T) {
 	require.True(t, opts.Ephemeral())
 	opts.Temporary = false
 	require.False(t, opts.Ephemeral())
+}
+
+func TestRSAOpts(t *testing.T) {
+	test := func(ephemeral bool) {
+		for _, opts := range []KeyGenOpts{
+			&RSA1024KeyGenOpts{ephemeral},
+			&RSA2048KeyGenOpts{ephemeral},
+			&RSA3072KeyGenOpts{ephemeral},
+			&RSA4096KeyGenOpts{ephemeral},
+		} {
+			expectedAlgorithm := reflect.TypeOf(opts).String()[7:14]
+			assert.Equal(t, expectedAlgorithm, opts.Algorithm())
+			assert.Equal(t, ephemeral, opts.Ephemeral())
+		}
+	}
+	test(true)
+	test(false)
 }
 
 func TestECDSAOpts(t *testing.T) {
@@ -116,13 +134,17 @@ func TestHMAC(t *testing.T) {
 
 func TestKeyGenOpts(t *testing.T) {
 	expectedAlgorithms := map[reflect.Type]string{
-		reflect.TypeOf(&HMACImportKeyOpts{}):       "HMAC",
-		reflect.TypeOf(&X509PublicKeyImportOpts{}): "X509Certificate",
-		reflect.TypeOf(&AES256ImportKeyOpts{}):     "AES",
+		reflect.TypeOf(&HMACImportKeyOpts{}):        "HMAC",
+		reflect.TypeOf(&RSAKeyGenOpts{}):            "RSA",
+		reflect.TypeOf(&RSAGoPublicKeyImportOpts{}): "RSA",
+		reflect.TypeOf(&X509PublicKeyImportOpts{}):  "X509Certificate",
+		reflect.TypeOf(&AES256ImportKeyOpts{}):      "AES",
 	}
 	test := func(ephemeral bool) {
 		for _, opts := range []KeyGenOpts{
 			&HMACImportKeyOpts{ephemeral},
+			&RSAKeyGenOpts{ephemeral},
+			&RSAGoPublicKeyImportOpts{ephemeral},
 			&X509PublicKeyImportOpts{ephemeral},
 			&AES256ImportKeyOpts{ephemeral},
 		} {
--- a/bccsp/opts.go
+++ b/bccsp/opts.go
@@ -22,6 +22,19 @@ const (
 	// ECDSAReRand ECDSA key re-randomization
 	ECDSAReRand = "ECDSA_RERAND"
 
+	// RSA at the default security level.
+	// Each BCCSP may or may not support default security level. If not supported than
+	// an error will be returned.
+	RSA = "RSA"
+	// RSA at 1024 bit security level.
+	RSA1024 = "RSA1024"
+	// RSA at 2048 bit security level.
+	RSA2048 = "RSA2048"
+	// RSA at 3072 bit security level.
+	RSA3072 = "RSA3072"
+	// RSA at 4096 bit security level.
+	RSA4096 = "RSA4096"
+
 	// AES Advanced Encryption Standard at the default security level.
 	// Each BCCSP may or may not support default security level. If not supported than
 	// an error will be returned.
@@ -262,5 +275,37 @@ func (opts *X509PublicKeyImportOpts) Algorithm() string {
 // Ephemeral returns true if the key to generate has to be ephemeral,
 // false otherwise.
 func (opts *X509PublicKeyImportOpts) Ephemeral() bool {
+	return opts.Temporary
+}
+
+// RSAKeyGenOpts contains options for RSA key generation.
+type RSAKeyGenOpts struct {
+	Temporary bool
+}
+
+// Algorithm returns the key generation algorithm identifier (to be used).
+func (opts *RSAKeyGenOpts) Algorithm() string {
+	return RSA
+}
+
+// Ephemeral returns true if the key to generate has to be ephemeral,
+// false otherwise.
+func (opts *RSAKeyGenOpts) Ephemeral() bool {
+	return opts.Temporary
+}
+
+// ECDSAGoPublicKeyImportOpts contains options for RSA key importation from rsa.PublicKey
+type RSAGoPublicKeyImportOpts struct {
+	Temporary bool
+}
+
+// Algorithm returns the key importation algorithm identifier (to be used).
+func (opts *RSAGoPublicKeyImportOpts) Algorithm() string {
+	return RSA
+}
+
+// Ephemeral returns true if the key to generate has to be ephemeral,
+// false otherwise.
+func (opts *RSAGoPublicKeyImportOpts) Ephemeral() bool {
 	return opts.Temporary
 }
--- a/bccsp/rsaopts.go
+++ b/bccsp/rsaopts.go
@@ -1,0 +1,81 @@
+/*
+Copyright IBM Corp. 2016 All Rights Reserved.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+		 http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package bccsp
+
+// RSA1024KeyGenOpts contains options for RSA key generation at 1024 security.
+type RSA1024KeyGenOpts struct {
+	Temporary bool
+}
+
+// Algorithm returns the key generation algorithm identifier (to be used).
+func (opts *RSA1024KeyGenOpts) Algorithm() string {
+	return RSA1024
+}
+
+// Ephemeral returns true if the key to generate has to be ephemeral,
+// false otherwise.
+func (opts *RSA1024KeyGenOpts) Ephemeral() bool {
+	return opts.Temporary
+}
+
+// RSA2048KeyGenOpts contains options for RSA key generation at 2048 security.
+type RSA2048KeyGenOpts struct {
+	Temporary bool
+}
+
+// Algorithm returns the key generation algorithm identifier (to be used).
+func (opts *RSA2048KeyGenOpts) Algorithm() string {
+	return RSA2048
+}
+
+// Ephemeral returns true if the key to generate has to be ephemeral,
+// false otherwise.
+func (opts *RSA2048KeyGenOpts) Ephemeral() bool {
+	return opts.Temporary
+}
+
+// RSA3072KeyGenOpts contains options for RSA key generation at 3072 security.
+type RSA3072KeyGenOpts struct {
+	Temporary bool
+}
+
+// Algorithm returns the key generation algorithm identifier (to be used).
+func (opts *RSA3072KeyGenOpts) Algorithm() string {
+	return RSA3072
+}
+
+// Ephemeral returns true if the key to generate has to be ephemeral,
+// false otherwise.
+func (opts *RSA3072KeyGenOpts) Ephemeral() bool {
+	return opts.Temporary
+}
+
+// RSA4096KeyGenOpts contains options for RSA key generation at 4096 security.
+type RSA4096KeyGenOpts struct {
+	Temporary bool
+}
+
+// Algorithm returns the key generation algorithm identifier (to be used).
+func (opts *RSA4096KeyGenOpts) Algorithm() string {
+	return RSA4096
+}
+
+// Ephemeral returns true if the key to generate has to be ephemeral,
+// false otherwise.
+func (opts *RSA4096KeyGenOpts) Ephemeral() bool {
+	return opts.Temporary
+}
--- a/bccsp/sw/conf.go
+++ b/bccsp/sw/conf.go
@@ -20,6 +20,7 @@ type config struct {
 	ellipticCurve elliptic.Curve
 	hashFunction  func() hash.Hash
 	aesBitLength  int
+	rsaBitLength  int
 }
 
 func (conf *config) setSecurityLevel(securityLevel int, hashFamily string) (err error) {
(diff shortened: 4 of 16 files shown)


denyeart:
It looks like RSA was removed in commit https://github.com/hyperledger/fabric/commit/80a20e3aaf5fcfa70e2e3cd8a3e10eb31b1d5901 and the commit says "The RSA operations are not used within Fabric." @sykesm can you provide some more of the background?

ale-linux:
I guess if any of the endorsers/creators used RSA, then it's possible that after an upgrade a peer could no longer verify signatures generated by those.. @filipeandre could you comment?

sykesm:
> It looks like RSA was removed in commit [80a20e3](https://github.com/hyperledger/fabric/commit/80a20e3aaf5fcfa70e2e3cd8a3e10eb31b1d5901) and the commit says "The RSA operations are not used within Fabric." @sykesm can you provide some more of the background?

From my perspective, we are here because:

1. No code within the Fabric repo or integration test suites related to Fabric were exercising the RSA paths.
2. There was a JIRA item opened to deprecate or remove the code
3. We had an explicit statement about RSA in our documentation for Fabric 1.2 that [reads][1] "We emphasize that in Hyperledger Fabric there is no support for certificates including RSA keys."

[1]: https://hyperledger-fabric.readthedocs.io/en/release-1.2/msp.html#how-to-generate-msp-certificates-and-their-signing-keys

yacovm:
The problem is that in a **blockchain**, if you **remove code paths that are used in transaction processing**, you either end up in a **fork**, or you are **stuck forever**, because the code in the present still needs to be able to process data from the past in the same way it did.

yacovm:
The problem is that in a **blockchain**, if you **remove code paths that are used in transaction processing**, you either end up in a **fork**, or you are **stuck forever**, because the code in the present still needs to be able to process data from the past in the same way it did.

So, what is the decision for this PR? are we going to revert the RSA removal? Are the affected users going to need to fork their own Fabric version because we don't revert it?

denyeart:
Given there was a decision not to support RSA, and that the decision was in the official documentation since v1.0, I'd suggest the correct path would be to open an RFC at https://github.com/hyperledger/fabric-rfcs/ for adding RSA support. It should mention the rationale, but more importantly what is the proposed test and support approach, as I believe those were the considerations behind the prior decision.

sykesm:
> The problem is that in a **blockchain**, if you **remove code paths that are used in transaction processing**, you either end up in a **fork**, or you are **stuck forever**, because the code in the present still needs to be able to process data from the past in the same way it did.

Yes, we all realize this.

> So, what is the decision for this PR? are we going to revert the RSA removal? 

This PR is not mergeable as-is so let's focus on the bigger issue.

Given the explicit statement in the documentation that there is no support for RSA (which exists in all versions of the MSP article [since v1.0.0][1]), this comes down to whether or not we want to implement and maintain formal support for RSA, do just enough to enable a migration, or do nothing. It's clear from the documentation and lack of test that this was not something that we intended to support yet we did not add the guard rails to prevent it from being used.

If we wish to support RSA, the code needs to be restored, integration tests need to be written, HSM support needs to be validated (since there was code there as well), and the documentation needs to be updated. This, to me, is more than a simple bug fix.

> Are the affected users going to need to fork their own Fabric version because we don't revert it?

Clearly not ideal.

[1]: https://github.com/hyperledger/fabric/commit/4441f533a18f4b2baa94070b639a4153e118f8d1

yacovm:
> > So, what is the decision for this PR? are we going to revert the RSA removal?
> 
> This PR is not mergeable as-is so let's focus on the bigger issue.
> 

I think that an another issue is that currently, there might be more networks of Fabric 1.4.x out there that once they try upgrade to 2.x they are in for an unpleasant surprise. 

We should probably circulate a warning through the mailing list.

filipeandre:
> I guess if any of the endorsers/creators used RSA, then it's possible that after an upgrade a peer could no longer verify signatures generated by those.. @filipeandre could you comment?

Yes, that's it.

filipeandre:
> If we wish to support RSA, the code needs to be restored, integration tests need to be written, HSM support needs to be validated (since there was code there as well), and the documentation needs to be updated. This, to me, is more than a simple bug fix.



> tement

filipeandre:
> > The problem is that in a **blockchain**, if you **remove code paths that are used in transaction processing**, you either end up in a **fork**, or you are **stuck forever**, because the code in the present still needs to be able to process data from the past in the same way it did.
> 
> Yes, we all realize this.
> 
> > So, what is the decision for this PR? are we going to revert the RSA removal?
> 
> This PR is not mergeable as-is so let's focus on the bigger issue.
> 
> Given the explicit statement in the documentation that there is no support for RSA (which exists in all versions of the MSP article [since v1.0.0](https://github.com/hyperledger/fabric/commit/4441f533a18f4b2baa94070b639a4153e118f8d1)), this comes down to whether or not we want to implement and maintain formal support for RSA, do just enough to enable a migration, or do nothing. It's clear from the documentation and lack of test that this was not something that we intended to support yet we did not add the guard rails to prevent it from being used.
> 
> If we wish to support RSA, the code needs to be restored, integration tests need to be written, HSM support needs to be validated (since there was code there as well), and the documentation needs to be updated. This, to me, is more than a simple bug fix.
> 
> > Are the affected users going to need to fork their own Fabric version because we don't revert it?
> 
> Clearly not ideal.

Since the beginning of fabric we were using it with RSA.
Our CA generates rsa keys by default:
`fabric-ca-server start -b ${CA_EMAIL}:${CA_PASSWORD}  --csr.keyrequest.algo rsa --csr.keyrequest.size 4096
`
I believe at some point of time there was a documentation statement saying that fabric supports ecda as well as rsa ?

I suggest to restore the RSA code as it was proven to be working.
Meanwhile the full tests are not written, we could state in the documentation about this, instead of removing it from the code.

pmcostaioi:
> Given there was a decision not to support RSA, and that the decision was in the official documentation since v1.0, I

Sorry I must have missed it. But I have not seen that mentioned on the documentation until 1.2. Can you point me out in the direction of where it was mentioned on 1.0? To ensure I don't miss such important announcement of a diference between what the code supports and of how it should be used?

viktor-iownit:
> I think that an another issue is that currently, there might be more networks of Fabric 1.4.x out there that once they try upgrade to 2.x they are in for an unpleasant surprise.
> 
> We should probably circulate a warning through the mailing list.

@yacovm This is exactly where we (with @filipeandre, @jorgecaeiros and @pmcostaioi) are at the moment - trying to find out how to upgrade our live production Fabric from 1.4.6 to 2.2.x and realized that there is no way to convert ca, orderers and peers from using RSA to the default type keys. We are happy to participate in every possible way in the development of the migration procedure or bring back the RSA support.

jorgecaeiros:
Our production system has over 20 different organizations, all running on 1.4.6 and using MSPs dynamically generated by HLF CAs using RSA algorithm.
Before upgrading to the latest 2.x version of fabric we attempted a simpler test setup:

- Single organization
- 5 orderers (orderer 0, orderer 1, …. orderer 4) working using Raft consensus mechanism
- 1 CA using RSA encryption
- 1 system channel which the orderers are using
- 1 application channel used, well, by our application
- 2 peers that joined the application channel
- All fabric components running on fabric 1.4.6
- Ledger type = file on both orderer and peers

This is the procedure we executed to move away from RSA to ECSDA (before upgrading to 2.2):
- Had the 5 orderers and 2 peers running
- Stopped the CA
- Regenerated the CA’s MSP based on ECSDA
- Started the CA
- updated both system channel and application channel to add this new public certificate of the CA to the org using the MSP of the org’s admin (signed with the previous CA’s MSP). This was done by generating a channel config update transaction based on a delta computed and converted to and from json/protbuf using configtxlator
- Regenerated the org’s admin MSP to be based on ECSDA and added his public certificate to the org (signed with the previous CA’s MSP). This was done by generating a channel config update transaction based on a delta computed and converted to and from json/protbuf using configtxlator
- Regenerated orderer 0's MSP based on ECSDA
- Updated the raft consenter section of the config block (of both system and application channels) to update orderer 0's certificate (orderer 0 is still running). Did that via the config update transaction
- Once the new block was distributed, backed up orderer 0, stopped orderer 0, and restarted it using the new MSP
- Orderer 0 started with the new MSP and used the latest config block to join the Raft cluster
- Repeated these steps for orderers 1-4
- After this we had all orderers communicating over raft using ECSDA MSPs
- Then we regenerated both peers MSPs based on ECSDA
- Then we backed up one of the peers, stopped it and restarted it using the new MSP
- Repeated the process for the other peer
- Then, using the new admin’s MSP, we generated a config update transaction to remove the previous CA and admin certificates from both the system and application channels
- Then executed some test transactions using the new MSPs and verified that it worked fine
- Extracted the latest config block (from both system and application channeles), converted it to JSON and ensured that there were no RSA certificates present there

Then we attempted to upgrade to fabric 2.2:
- we backed up orderer 0
- stopped orderer 0
- upgraded the base image to 2.2
- started it

And it failed when it was validating the existing ledger:

2020-10-28 12:22:25.282 UTC [orderer.common.server] Main -> PANIC Failed validating bootstrap block: initializing channelconfig failed: could not create channel Orderer sub-group config: setting up the MSP manager failed: Failed importing key with opts [&{true}]: Certificate’s public key type not recognized. Supported keys: [ECDSA]
panic: Failed validating bootstrap block: initializing channelconfig failed: could not create channel Orderer sub-group config: setting up the MSP manager failed: Failed importing key with opts [&{true}]: Certificate’s public key type not recognized. Supported keys: [ECDSA]

yacovm:
> 2020-10-28 12:22:25.282 UTC [orderer.common.server] Main -> PANIC Failed validating bootstrap block: initializing channelconfig failed: could not create channel Orderer sub-group config: setting up the MSP manager failed: Failed importing key with opts [&{true}]: Certificate’s public key type not recognized. Supported keys: [ECDSA]

Well you need to replace the bootstrap block with the latest configuration block of the system channel, one that doesn't have RSA certificates in it. But even if you do it, you will never be able to add new nodes to the network.

jorgecaeiros:
@yacovm We tried that approach as well once we had everything running on ECSDA
- stopped orderer 0 (without any backup)
- set up the genesis block to be the current config block
- started it (with the ECSDA MSP and an empty production folder)
- It used to block correctly and connected to the raft consensus
- It started receiving the ledger from both channels

It crashed with the same error while it was validating the ledger.

As far as we can determine, there is no way to migrate an existing ledger from fabric 1.4.6 to 2.x if there are RSA certificates in a block of a channel

yacovm:
> > I think that an another issue is that currently, there might be more networks of Fabric 1.4.x out there that once they try upgrade to 2.x they are in for an unpleasant surprise.
> > We should probably circulate a warning through the mailing list.
> 
> @yacovm This is exactly where we (with @filipeandre, @jorgecaeiros and @pmcostaioi) are at the moment - trying to find out how to upgrade our live production Fabric from 1.4.6 to 2.2.x and realized that there is no way to convert ca, orderers and peers from using RSA to the default type keys. We are happy to participate in every possible way in the development of the migration procedure or bring back the RSA support.

For the short term I suggest that you fork Fabric and add RSA support yourself.
For the longer term, as @denyeart  suggested, you can open an [RFC](https://github.com/hyperledger/fabric-rfcs/pulls) about making Fabric have RSA again.

yacovm:
> started it (with the ECSDA MSP and an empty production folder)

Why would you start it with an empty production folder? Replicating the blocks will cause it to see the RSA certificates again.  You need to restart it with a bootstrap block of the latest config block that doesn't contain RSA, but while retaining the ledger, so that it will not try to replicate blocks with RSA certificates in them.

jorgecaeiros:
> > started it (with the ECSDA MSP and an empty production folder)
> 
> Why would you start it with an empty production folder? Replicating the blocks will cause it to see the RSA certificates again.

to force it to get the data from the other nodes so we could pinpoint in which stage the error was occurring.
With the genesis block as the latest config and with backup present, it still failed while validating the application channel

viktor-iownit:
> For the short term I suggest that you fork Fabric and add RSA support yourself.
> For the longer term, as @denyeart suggested, you can open an [RFC](https://github.com/hyperledger/fabric-rfcs/pulls) about making Fabric have RSA again.

Thank you @yacovm! we will create RFC asap.

denyeart:
I believe @sykesm is working on a proposal for RSA enablement (albeit still unsupported) so that existing RSA users could switch over to ECDSA.

sykesm:
> I believe @sykesm is working on a proposal for RSA enablement (albeit still unsupported) so that existing RSA users could switch over to ECDSA.

The general proposal was to bring back RSA behind a build tag. Builds and images published by the Fabric team would not include this code as it's untested, unsupported, and explicitly documented as so. This would enable a migration to the 2.x stream and ECDSA, and if things go to plan, eventually a path to checkpoint and removal of the old MSP definitions.

That said, before doing anything with RSA, I wanted to try running 1.4 with RSA. The goal was to test-drive the RSA paths by building a configuration that works with 1.4 and then moving it to a v2 network. What I found was that the 1.4 peer panics at bootstrap while attempting to sign with an RSA key. This is because the required RSA signing opts are not provided on any call to sign outside of bccsp unit tests. This is not surprising as, again, RSA is not supported and not recommended for Fabric.

With a mixed configuration where Fabric only verifies signatures, it's possible things hold together so that's the path I'm on right now.

While I've read your topology description, can you please describe the key type of the CA and issued certificates for each MSP? I ask as I don't see how the signing certs can be RSA.

Thanks.

viktor-iownit:
> While I've read your topology description, can you please describe the key type of the CA and issued certificates for each MSP? I ask as I don't see how the signing certs can be RSA.

We have our network started and working with RSA keys by doing this (short version):
* start hlf CA with command: fabric-ca-server start -b ${CA_EMAIL}:${CA_PASSWORD}  --csr.keyrequest.algo rsa --csr.keyrequest.size 4096
* register 5 orderers on the CA
* perform enrollment with each orderer to get their MSP
* use the public certificate (and the tls one) and set it up on the genesis block on the raft consenter section
* start the orderers
This will create a system channel with RSA certificates, using RSA based MSPs. The rest seems standard. Happy to discuss this process in details.

sykesm:
> We have our network started and working with RSA keys by doing this (short version)
> ...
> The rest seems standard.

I'd rather not assume so I'm trying to explicitly confirm what key types were generated when constructing the CSR for enrollment. My expectation is that they are not RSA. Can you please confirm?

I understand that the CA certificates in the MSP use RSA; I'm asking explicitly about the signing certificates in the local MSPs.

EDIT: I'd also like to know what key types are being used to sign the transaction proposals.

sykesm:
Within release-1.4, the Verify operation for RSA keys require the caller to provide signer options

https://github.com/hyperledger/fabric/blob/b5a12de28553481140374865850e1632cbb2b016/bccsp/sw/rsa.go#L58-L72

and the Sign operation for RSA keys requires the caller to provide signer options

https://github.com/hyperledger/fabric/blob/b5a12de28553481140374865850e1632cbb2b016/bccsp/sw/rsa.go#L30-L36

However, the MSP implementation specifies no options when verifying signatures

https://github.com/hyperledger/fabric/blob/b5a12de28553481140374865850e1632cbb2b016/msp/identities.go#L156-L183

and no options when signing

https://github.com/hyperledger/fabric/blob/b5a12de28553481140374865850e1632cbb2b016/msp/identities.go#L232-L256

All of this is further evidence that RSA was never supported by the MSP.

The only thing I have observed that worked in 1.4 and no longer works in 2.x are CA certificates associated with an RSA key. If that's the only aspect of this that has been dropped, all that's needed is a trivial change to the key import path to generate an MSP identity from an RSA public key.

IMO, this is _far_ from implying all of the RSA code that was pulled from bccsp needs to be restored.

Please build and test https://github.com/sykesm/fabric/commit/f47c396103def0c87057cd4ee822a9a2a270d8ae using your configuration and let us know what you see.

filipeandre:
> Within release-1.4, the Verify operation for RSA keys require the caller to provide signer options
> 
> https://github.com/hyperledger/fabric/blob/b5a12de28553481140374865850e1632cbb2b016/bccsp/sw/rsa.go#L58-L72
> 
> and the Sign operation for RSA keys requires the caller to provide signer options
> 
> https://github.com/hyperledger/fabric/blob/b5a12de28553481140374865850e1632cbb2b016/bccsp/sw/rsa.go#L30-L36
> 
> However, the MSP implementation specifies no options when verifying signatures
> 
> https://github.com/hyperledger/fabric/blob/b5a12de28553481140374865850e1632cbb2b016/msp/identities.go#L156-L183
> 
> and no options when signing
> 
> https://github.com/hyperledger/fabric/blob/b5a12de28553481140374865850e1632cbb2b016/msp/identities.go#L232-L256
> 
> All of this is further evidence that RSA was never supported by the MSP.
> 
> The only thing I have observed that worked in 1.4 and no longer works in 2.x are CA certificates associated with an RSA key. If that's the only aspect of this that has been dropped, all that's needed is a trivial change to the key import path to generate an MSP identity from an RSA public key.
> 
> IMO, this is _far_ from implying all of the RSA code that was pulled from bccsp needs to be restored.
> 
> Please build and test [sykesm@f47c396](https://github.com/sykesm/fabric/commit/f47c396103def0c87057cd4ee822a9a2a270d8ae) using your configuration and let us know what you see.

Thank you very much sykesm!
I compiled and tested it. It works with our setup which means that the rest of code is not needed 👍

sykesm:
RSA key import:

- master: https://github.com/hyperledger/fabric/pull/2121
- release-2.2: https://github.com/hyperledger/fabric/pull/2122

denyeart:
@filipeandre thanks for your verification.

I'm happy to report that the PRs are merged:

- master (upcoming v2.3 release): #2121
- release-2.2 (future v2.2.2 release): #2122

I believe this issue is now resolved, I'll go ahead and close this PR.

PR #2075, opened 2020-11-05
Add function TxIDExists and use in validation path

manish-sethi:

#### Type of change
- New feature

#### Description
This commit adds a function in ledger TxIDExists and uses this in validation path for checking duplicate txids.

#### Additional details
The current validation code, in order to checks for a duplicate txid, uses the API `GetTransactionByID` on ledger and relies on the error types returned by this API. For a ledger that is bootstrapped from a snapshot, this API returns another untyped error (stating that the TxID details not available) if the TxID belongs to a block that is committed prior to the snapshot height. In the current code, in validation path, this error will be interpreted as general ledger error that results in a peer panic. 
One option is to return a typed error (e.g., `TxIDBeforeSnapshot`) and modify the validation to check for this error type as well. However, instead of exposing one more error type for validation code, it is cleaner to add an explicit function for checking the existence of a TxID. This makes the existing code relatively more readable as well.

#### Related issues
[FAB-18318](https://jira.hyperledger.org/browse/FAB-18318)


common/ledger/blkstorage/blockindex.go
```diff
@@ -223,6 +223,24 @@ func (index *blockIndex) getTxValidationCodeByTxID(txID string) (peer.TxValidati
...
+		return false, ErrAttrNotIndexed
+	}
+	rangeScan := constructTxIDRangeScan(txID)
+	itr, err := index.db.GetIterator(rangeScan.startKey, rangeScan.stopKey)
+	if err != nil {
+		return false, errors.WithMessagef(err, "error while trying to check the presence of TXID [%s]", txID)
+	}
+	defer itr.Release()
+
+	present := itr.Next()
```
ale-linux, line 237: so, just to understand how we're changing this: did we use to return `ledger.NotFoundInIndexErr` before if present was false?

core/ledger/ledger_interface.go
```diff
@@ -157,6 +157,8 @@ type PeerLedgerProvider interface {
 // that tells apart valid transactions from invalid ones
 type PeerLedger interface {
 	commonledger.Ledger
+	// TxIDExists returns true if the specified txID is already present in one of the already committed blocks
+	TxIDExists(txID string) (bool, error)
```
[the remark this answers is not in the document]
manish-sethi: done

common/ledger/blkstorage/blockindex.go
manish-sethi: We are not changing the existing function but yes in that function it returns an error at this point. The existing function `GetTransactionByID(txID string) (*peer.ProcessedTransaction, error)` is more comprehensive that was originally meant for a general purpose client that wants to retrieve a `TxEnvelope`. However when it was used in the validation path, it became necessary to distinguish the error types (because the fist return would always be nil in that case). 
For your amusement, you has fixed this in the validation path - https://jira.hyperledger.org/browse/FAB-10201 :-)

denyeart (approved):

PR #2078, opened 2020-11-05
[FAB-18298] Default cluster cert and key 

stephyee:
#### Type of change

- New feature

#### Description

When cluster listener is reused, default cluster client certificate and private key (if not set) to the TLS server certificate and key.

#### Related issues

[FAB-18298](https://jira.hyperledger.org/browse/FAB-18298)

PR #2080, opened 2020-11-06
Deploy production ordering service doc

joealewine:
Change-Id: Ib7631cf4974da7005ae4a6fa7c0939a5ebe6588c

#### Type of change

- Documentation update

#### Description

Like the peer and CA deployment guides, this series of docs talks people through the process to plan and deploy an ordering service.


docs/source/deployorderer/ordererchecklist.md
```diff
@@ -0,0 +1,307 @@
...
+```
+
+**`LocalMSPID`**: (default value should be overridden) The MSP ID must match the orderer organization MSP ID that exists in the configuration of the system channel. This means the MSP ID must have been listen in the `configtx.yaml` used to create the genesis block of the system channel (or have been added later to the list of system channel administrators).
+
+## General.BCCSP.*
+
+BCCSP.SW can’t be configured at the moment. Leave those parameters as they are.
+Should this section be similar to the same content in the  CA and Peer sections?
+
+**`PKCS11`**:
```
pamandrejko: PKCS11 resides inside the BCCSP section no alongside it. I think we can make these recommendations match the peer recommendations for these parameters instead?

docs/source/deployorderer/ordererchecklist.md
```diff
@@ -0,0 +1,307 @@
...
+- **`BCCSP.PKCS11.*:`** Provide this set of parameters according to your HSM configuration. Refer to this [example]((../hsm.html) of an HSM configuration for more information.
+
+## FileLedger.Location
+
+```
+# Location: The directory to store the blocks in.
+Location: /var/hyperledger/production/orderer
+```
+
+**`Location`**: (default value should be overridden, especially if more than node will be created by this organization) Every channel on which the node is a consenter will have its own subdirectory at this location.
```
pamandrejko: The user running the orderer needs to own and have write access to this directory.

pamandrejko (commented):
A few comments on the parameters.

docs/source/deployorderer/ordererchecklist.md
```diff
@@ -0,0 +1,307 @@
...
+# sample configuration provided has an MSP ID of "SampleOrg".
+LocalMSPID: SampleOrg
+```
+
+**`LocalMSPID`**: (default value should be overridden) The MSP ID must match the orderer organization MSP ID that exists in the configuration of the system channel. This means the MSP ID must have been listen in the `configtx.yaml` used to create the genesis block of the system channel (or have been added later to the list of system channel administrators).
+
+## General.BCCSP.*
+
+BCCSP.SW can’t be configured at the moment. Leave those parameters as they are.
+Should this section be similar to the same content in the  CA and Peer sections?
```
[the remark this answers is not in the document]
joealewine: I did put that at the end and intended to delete this line.

joealewine (commented):
Thanks for these comments, Pam!

docs/source/deployorderer/ordererdeploy.md
```diff
@@ -0,0 +1,272 @@
...
+export FABRIC_CFG_PATH=<PATH_TO_ORDERER_YAML>
+```
+
+### When you start the orderer, it fails with the following error:
+```
+PANI 003 Failed to setup local msp with config: administrators must be declared when no admin ou classification is set
+```
+
+**Solution:**   
+Your local MSP definition is missing the `config.yaml` file. Create the file and copy it into the local MSP `/msp` folder of orderer. See the [Fabric CA](https://hyperledger-fabric-ca.readthedocs.io/en/release-1.4/deployguide/use_CA.html#nodeous) documentation for more instructions.
```
wlahti, line 245: I think I noted this in the channel participation tutorial as well but which `config.yaml` are we referring to here?

docs/source/deployorderer/ordererdeploy.md
```diff
@@ -0,0 +1,272 @@
...
+
+
+### When you start the orderer, it fails with the following error:
+```
+PANI 005 Failed validating bootstrap block: initializing channelconfig failed: could not create channel Orderer sub-group config: setting up the MSP manager failed: administrators must be declared when no admin ou classification is set
+```
+
+**Solution:**   
+The system channel configuration is missing `config.yaml` file. If you are creating a new ordering service, the `MSPDir` referenced in `configtx.yaml` file is missing the `config.yaml` file. Follow instructions in the [Fabric CA](https://hyperledger-fabric-ca.readthedocs.io/en/release-1.4/deployguide/use_CA.html#nodeous) documentation to generate this file and then rerun `configtxgen` to regenerate the genesis block for the system channel.
```
wlahti, line 255: Same `config.yaml` comment here.

docs/source/deployorderer/ordererplan.md
```diff
@@ -0,0 +1,90 @@
+# Planning for an ordering service
+
+Audience: Architects, network operators, users setting up a production Fabric network who are familiar with Transport Layer Security (TLS), Public Key Infrastructure (PKI) and Membership Service Providers (MSPs).
+
+Check out the conceptual topic on [The Ordering Service](./orderer/ordering_service.html) for an overview on ordering service concepts, implementations, and the role an ordering service plays in a transaction.
+
+In Hyperledger Fabric network, a node or collection of nodes together form what's called an "ordering service", which literally orders transactions into blocks and sends them to peers to be committed to their ledgers. This separates Fabric from other distributed blockchains, such as Ethereum and Bitcoin, in which this ordering is done by any and all nodes, requiring the establishment of probabilistic consensus algorithms which eventually guarantee ledger consistency to a high degree of probability, but which are still vulnerable to divergent ledgers (also known as a ledger “fork”), where different participants in the network have a different view of the accepted order of transactions.
```
wlahti:
```suggestion
In a Hyperledger Fabric network, a node or collection of nodes together form what's called an "ordering service", which literally orders transactions into blocks and sends them to peers to be committed to their ledgers. This separates Fabric from other distributed blockchains, such as Ethereum and Bitcoin, in which this ordering is done by any and all nodes, requiring the establishment of probabilistic consensus algorithms which eventually guarantee ledger consistency to a high degree of probability, but which are still vulnerable to divergent ledgers (also known as a ledger “fork”), where different participants in the network have a different view of the accepted order of transactions.
```

wlahti: `which literally orders transactions into blocks`

Silly side note: wouldn't be an accurate reflection of the 2020 lexicon without the usage (and thankfully accurate usage) of `literally` in this doc. ;)

docs/source/deployorderer/ordererplan.md
```diff
@@ -0,0 +1,90 @@
...
+
+## Generate ordering node identities and Membership Service Providers (MSPs)
+
+Before proceeding with this topic, you should have reviewed the process for a Deploying a Certificate Authority (CA) for your organization in order to generate the identities and MSPs for the admins and ordering nodes in your organization. To learn how to use a CA to create these identities, check out [Registering and enrolling identities with a CA](https://hyperledger-fabric-ca.readthedocs.io/en/release-1.4/deployguide/use_CA.html). Note that the best practice is to register and enroll a separate node identity for each ordering node and to use distinct TLS certificates for each node.
+
+Note that the “cryptogen” tool should never be used to generate any identities in a production scenario.
```
wlahti: For consistency with the other references to tools
```suggestion
Note that the `cryptogen` tool should never be used to generate any identities in a production scenario.
```

docs/source/deployorderer/ordererplan.md
```diff
@@ -0,0 +1,90 @@
...
+
+This “system channel” is a special channel run by the ordering service and contains, among other things, the list of peer organizations that are allowed to create application channels (this list is known as the “consortium”). Although this system channel cannot be joined by peers or peer organizations (and thus, no transactions other than configuration transactions can be made on it), it does contain many of the same configuration parameters that application channels contain. Because application channels inherit these configuration values by default unless they are changed during the channel creation process, take care when creating your genesis block to keep the use case of your network in mind.
+
+If you’re creating an ordering service, you must create this genesis block by specifying the necessary parameters in `configtx.yaml` and using the `configtxgen` tool to create the block.
+
+If you are adding a node to the system channel, the best practice is to bootstrapped using the latest configuration block of the system channel. Similarly, an ordering node added to the consenter of an application channel will be boostrapped using the latest configuration block of that channel.
```
wlahti:
```suggestion
If you are adding a node to the system channel, the best practice is to bootstrap using the latest configuration block of the system channel. Similarly, an ordering node added to the consenter of an application channel will be boostrapped using the latest configuration block of that channel.
```

docs/source/deployorderer/ordererplan.md
```diff
@@ -0,0 +1,90 @@
...
+
+Note that the `configtx.yaml` that is shipped with Fabric binaries is identical to the [sample `configtx.yaml` found here](https://github.com/hyperledger/fabric/blob/master/sampleconfig/configtx.yaml), and contains the same channel "profiles" that are used to specify particular desired policies and parameters (for example, it can be used to which ordering nodes that are consenters in the system channel will be used in an application channel). When creating a channel (whether for an orderer system channel or an application channel), you specify a particular profile by name in your channel creation command, and that profile, along with the other parameters specified in `configtx.yaml`, are used to build the configuration block.
```
wlahti:
```suggestion
Note that the `configtx.yaml` that is shipped with Fabric binaries is identical to the [sample `configtx.yaml` found here](https://github.com/hyperledger/fabric/blob/master/sampleconfig/configtx.yaml), and contains the same channel "profiles" that are used to specify particular desired policies and parameters (for example, it can be used to specify(?) which ordering nodes that are consenters in the system channel will be used in an application channel). When creating a channel (whether for an orderer system channel or an application channel), you specify a particular profile by name in your channel creation command, and that profile, along with the other parameters specified in `configtx.yaml`, are used to build the configuration block.
```

docs/source/deployorderer/ordererplan.md
```diff
@@ -0,0 +1,90 @@
...
+
+Check out the [tutorial on creating a channel](./create_channel/create_channel.html#the-orderer-system-channel) for more information on how to create a genesis block.
+
+### Creating profiles for application channels
+
+Both the system and all application channels are built using a `configtx.yaml` file. Therefore, when editing your `configtx.yaml` to create the genesis block for your system channel, you can also add profiles for any application channels that will be created on this network. However, note that while you can define any set of consenters for each channel, **every consenter added to an application channel must first be a part of the system channel**. You cannot specify a consenter that is not a part of the system channel. Also, it is not possible to control the leader of the consenter set. Leaders are chosen by the `etcd` protocol used by the ordering nodes.
```
wlahti:
```suggestion
Both the system and all application channels are built using a `configtx.yaml` file. Therefore, when editing your `configtx.yaml` to create the genesis block for your system channel, you can also add profiles for any application channels that will be created on this network. However, note that while you can define any set of consenters for each channel, **every consenter added to an application channel must first be a part of the system channel**. You cannot specify a consenter that is not a part of the system channel. Also, it is not possible to control the leader of the consenter set. Leaders are chosen by the `etcdraft` protocol used by the ordering nodes.
```
Or similar like `raft` or `etcd/raft`, plain `etcd` seems wrong to me on initial glance. I'm sure @jyellick will know the accurate way to reference it here.

docs/source/deployorderer/ordererplan.md
```diff
@@ -0,0 +1,90 @@
...
+
+## Sizing your ordering node resources
+
+Because ordering nodes do not host a state database or chaincode, an ordering node will typically only have a single container associated with it. Like the “peer container” associated with the peer, this container encapsulates the ordering processes that order and commit transactions for all channels on which the ordering node is a consenter (ordering nodes also validating actions in particular cases). The ordering node storage includes the blockchain for all of channels on which the node is a consenter.
```
wlahti:
```suggestion
Because ordering nodes do not host a state database or chaincode, an ordering node will typically only have a single container associated with it. Like the “peer container” associated with the peer, this container encapsulates the ordering processes that order and commit transactions for all channels on which the ordering node is a consenter (ordering nodes also validate actions in particular cases). The ordering node storage includes the blockchain for all of channels on which the node is a consenter.
```

docs/source/deployorderer/ordererplan.md
```diff
@@ -0,0 +1,90 @@
...
+
+For these reasons, Raft clusters of more than a few dozen nodes begin to see noticeable performance degradation. Once clusters reach about 100 nodes, they begin having trouble maintaining quorum. The stage at which a deployment experiences issues is dependent on factors such as networking speeds and other resources available, and there are parameters such as the tick interval which can be used to mitigate the larger communications overhead.
+
+The optimum number of ordering nodes for your ordering service ultimately depends on your use case, your resources, and your topology. However, clusters of three, five, seven, or nine nodes, are the most popular, with no more than about 50 channels per orderer.
```
wlahti:
```suggestion
The optimal number of ordering nodes for your ordering service ultimately depends on your use case, your resources, and your topology. However, clusters of three, five, seven, or nine nodes, are the most popular, with no more than about 50 channels per orderer.
```
Nit: looks like this is an acceptable usage of the word but it seems uncommon.

docs/source/deployorderer/ordererplan.md
```diff
@@ -0,0 +1,90 @@
...
+
+## Storage considerations and monitoring
+
+The storage that should be allocated to an ordering node depends on factors such as the estimate transaction throughput, the size of blocks, and number of channels the node will be joined to. Your needs will depend on your use case. However, the best practice is to monitor the storage available to your nodes closely. You may also decide to enable an autoscaler, which will allocate more resources to your node, if your infrastructure allows it.
```
wlahti:
```suggestion
The storage that should be allocated to an ordering node depends on factors such as the estimated transaction throughput, the size of blocks, and number of channels the node will be joined to. Your needs will depend on your use case. However, the best practice is to monitor the storage available to your nodes closely. You may also decide to enable an autoscaler, which will allocate more resources to your node, if your infrastructure allows it.
```

wlahti (dismissed):

docs/source/deployorderer/ordererchecklist.md
```diff
@@ -0,0 +1,302 @@
...
+**`ServerCertificate`**
+**`ServerPrivateKey`**
+
+Bote: the `ServerCertificate` and `ServerPrivateKey` must be set to the same values as the `General.TLS.PrivateKey` and `General.TLS.Certificate`.
+
+## General.BoostrapMethod
+
+```
+# Bootstrap method: The method by which to obtain the bootstrap block
+# system channel is specified. The option can be one of:
```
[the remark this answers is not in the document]
joealewine: That's a quote from the orderer.yaml.

docs/source/deployorderer/ordererchecklist.md
```diff
@@ -0,0 +1,302 @@
...
+**`Certificate`**: (required when using the operations service) Can be the same file as the `orderer.tls.cert.file`.
+**`PrivateKey`**: (required when using the operations service) Can be the same file as the `orderer.tls.key.file`.
+**`ClientAuthRequired`**: (required when using the operations service) Must be set to `true` to enable mutual TLS between the client and the server.
+**`ClientRootCAs`**: (required when using the operations service) Similar to the client root CA cert file in TLS, it contains a list of client root CA certificates that can be used to verify client certificates. If the client enrolled with the peer organization CA, then this value is the orderer organization root CA cert.
+
+## Metrics.Provider.*
+
+By default this is disabled, but if you want to monitor the metrics for the orderer, you need to choose either `statsd` or `Prometheus` as your metric provider. `Statsd` uses a "push" model, pushing metrics from the orering node to a `statsd` endpoint. Because of this, it does not require configuration of the operations service itself. See the list of [Available metrics for the orderer](../metrics_reference.html#orderer-metrics).
```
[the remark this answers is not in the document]
joealewine: The inconsistency here is just an error, but in general we try to adhere to the common rules for capitalization used by the company developing a product or what is just sort of used by the internet in general. That's why we capitalize "Certificate Authority" and not "peer". 

The common usage I can find is "StatsD", so I'm going to change all of these to that (aside from the quotes from orderer.yaml, which unfortunately use the lowercase form).

docs/source/deployorderer/ordererdeploy.md
joealewine: `config.yaml` is a file that enables Node OUs when it's included in an MSP.

docs/source/deployorderer/ordererplan.md
joealewine: Haha. It felt weird to not throw that in there because you'd just be saying than an orderer orders something.

docs/source/deployorderer/ordererplan.md
joealewine: I'll change this one to `etcdraft` for now, as that seems like the right answer to me.

docs/source/deployorderer/ordererplan.md
joealewine: This one led me down a bit of a rabbit hole lol. They both seem fine but I don't mind changing this to optimal if it's the more common usage.

docs/source/deployorderer/ordererplan.md
joealewine: I think "expected" is actually better here, but thank you for flagging that "estimate" was definitely a typo.

docs/source/deployorderer/ordererplan.md
```diff
@@ -0,0 +1,90 @@
...
+
+You will likely have to modify one of these profiles in order to create your system channel and to create your application channels (if nothing else, you are likely to have to modify the sample organization names). Note that to create a Raft ordering service, you will have to specify an `OrdererType` of `etcdraft`.
+
+Check out the [tutorial on creating a channel](../create_channel/create_channel.html#the-orderer-system-channel) for more information on how to create a genesis block.
```
denyeart:
```suggestion
Check out the [tutorial on creating a channel](../create_channel/create_channel.html#the-orderer-system-channel) for more information on how to create a system channel and application channels.
```

docs/source/deployorderer/ordererplan.md
```diff
@@ -0,0 +1,90 @@
...
+
+## Certificates from a non-Fabric CA
+
+While it is possible to use a non-Fabric CA to generate identities, this process requires that you manually construct the MSP folders the ordering service and its organization need. That process will not be covered here and will instead focus on using a Fabric CA to generate the identities and MSP folders for you.
+
+## Transport Layer Security (TLS) enablement
+
+To prevent “man in the middle” attacks and otherwise secure communications, the use of TLS is a requirement for any production network. Therefore, in addition to registering your ordering nodes identities with your organization CA, you will also need to create certificates for your ordering nodes with the TLS CA of your organization. These TLS certificates will be used by the ordering nodes when communicating with the network.
+
+## Creating the genesis block
```
denyeart:
```suggestion
## Creating the system channel genesis block
```

docs/source/deployorderer/ordererplan.md
```diff
@@ -0,0 +1,90 @@
...
+
+## Sizing your ordering node resources
+
+Because ordering nodes do not host a state database or chaincode, an ordering node will typically only have a single container associated with it. Like the “peer container” associated with the peer, this container encapsulates the ordering processes that order and commit transactions for all channels on which the ordering node is a consenter (ordering nodes also validate actions in particular cases). The ordering node storage includes the blockchain for all of channels on which the node is a consenter.
```
denyeart: I'd reserve the word "commit" for peers.

```suggestion
Because ordering nodes do not host a state database or chaincode, an ordering node will typically only have a single container associated with it. Like the “peer container” associated with the peer, this container encapsulates the ordering process that orders transactions into blocks for all channels on which the ordering node is a consenter (ordering nodes also validate actions in particular cases). The ordering node storage includes the blockchain for all of channels on which the node is a consenter.
```

docs/source/deployorderer/ordererplan.md
```diff
@@ -0,0 +1,90 @@
...
+
+## Storage considerations and monitoring
+
+The storage that should be allocated to an ordering node depends on factors such as the expected transaction throughput, the size of blocks, and number of channels the node will be joined to. Your needs will depend on your use case. However, the best practice is to monitor the storage available to your nodes closely. You may also decide to enable an autoscaler, which will allocate more resources to your node, if your infrastructure allows it.
+
+If the storage for an ordering node is exhausted you also have the option to deploy a new node with a larger storage allocation and allow it to sync with the relevant ledgers. If you have several ordering nodes available to use, ensure that each node is a consenter on approximately the same number of channels.
+
+In a production environment you should also monitor the CPU and memory allocated to an ordering node using widely available tooling. If you see an ordering node struggling to keep up (for example, it might be calling for leader elections when none is needed), it is a sign that you might need to increase its resource allocation.
+
+## Gossip
```
denyeart: Remove this whole section - orderers do not participate in gossip at all.

While peers pull blocks from orderers, and we sometimes put this under the banner of "gossip", it isn't actually gossip, it's just a simple grpc pull, and then peers "gossip" to other peers. Orderers do not participate in gossip whatsoever.

docs/source/deployorderer/ordererdeploy.md
```diff
@@ -0,0 +1,275 @@
...
+- If you require mutual TLS (`General.TLS.ClientAuthRequired` set to true), you need to indicate to the orderer which TLS CA root certificates to use to authorize clients. Copy the organization's TLS CA root certificate `ca-cert.pem` to `organizations/ordererOrganizations/ordererOrg1.example.com/orderers/orderer0.ordererOrg1.example.com/tls/tls-cert.pem` so that the organization's clients will be authorized.  If `General.TLS.ClientAuthRequired` is set to false, you can skip this step.
+
+**Note:** Don't forget to create the `config.yaml` file and add it to the organization MSP and local MSP folder for each ordering node. This file enables Node OU support for the channel, an important feature that allows the channel to recognize certificates that contain the "admin" OU as an admin identity of the channel. Learn more in the [Fabric CA](https://hyperledger-fabric-ca.readthedocs.io/en/release-1.4/deployguide/use_CA.html#nodeous) documentation.
```
denyeart: This paragraph wasn't quite right... I believe it was also copy/pasted into the channel participation tutorial, and perhaps elsewhere.

```suggestion
**Note:** Don't forget to create the `config.yaml` file and add it to the organization MSP and local MSP folder for each ordering node. This file enables Node OU support for the MSP, an important feature that allows the MSP's admin to be identified based on an "admin" OU in an identity's certificate. Learn more in the [Fabric CA](https://hyperledger-fabric-ca.readthedocs.io/en/release-1.4/deployguide/use_CA.html#nodeous) documentation.
```

docs/source/deployorderer/ordererdeploy.md
```diff
@@ -0,0 +1,275 @@
...
+
+If you are using a containerized solution for running your network (which for obvious reasons is a popular choice), **it is a best practice to mount these folders (volumes) external to the container where the node itself is running. This will allow the certificates to be used to create a new node should the node container go down, become corrupted, or is restarted.**
```
denyeart:
```suggestion
If you are using a containerized solution for running your network (which for obvious reasons is a popular choice), **it is a best practice to mount volumes for the certificate directories external to the container where the node itself is running. This will allow the certificates to be used by an ordering node container, regardless whether the ordering node container goes down, becomes corrupted, or is restarted.**
```

docs/source/deployorderer/ordererdeploy.md
```diff
@@ -0,0 +1,275 @@
...
+
+Before using `configtxgen`, confirm you have set the `FABRIC_CFG_PATH` environment variable to the path of the directory that contains your local copy of the `configtx.yaml` file. You can verify that are able to use the tool by printing the `configtxgen` help text:
+
+```
+configtxgen --help
+```
+
+#### The `configtx.yaml` file
+
+The `configtx.yaml` file is used to specify the **channel configuration** of new channels. The information that is required to build the channel configuration is specified in a readable and editable form in the `configtx.yaml` file. The `configtxgen` tool uses the channel profiles that are defined in `configtx.yaml` to create the channel configuration block in the [protobuf format](https://developers.google.com/protocol-buffers).
```
denyeart:
```suggestion
The `configtx.yaml` file is used to specify the **channel configuration** of the system channel and application channels. The information that is required to build the channel configuration is specified in a readable and editable form in the `configtx.yaml` file. The `configtxgen` tool uses the channel profiles that are defined in `configtx.yaml` to create the channel configuration block in the [protobuf format](https://developers.google.com/protocol-buffers).
```

docs/source/deployorderer/ordererdeploy.md
```diff
@@ -0,0 +1,275 @@
...
+
+The `configtxgen` tool uses `configtx.yaml` file to create the genesis block for the channel. A detailed version of the `configtx.yaml` file is available in the [Fabric sample configuration](https://github.com/hyperledger/fabric/blob/{BRANCH}/sampleconfig/configtx.yaml). Refer to the [Using configtx.yaml to build a channel configuration](../create_channel/create_channel_config.html) tutorial to learn more about the  settings in this file.
+
+#### Generate the genesis block
+
+After you have completed editing the `configtx.yaml`, you can start the process of creating a system channel for the ordering service. Every channel configuration starts with a genesis block. We have already set the environment variables that are required to use the `configtxgen` tool. Run the following command to build the genesis block for `channel1` using the `SampleAppGenesisEtcdRaft` profile:
+
+```
+configtxgen -profile SampleAppGenesisEtcdRaft -outputCreateChannelTx ./channel-artifacts/channel1.tx -channelID channel1
```
denyeart: Hang on, the first sentence of this section said we were going to create the genesis block of the system channel. We've jumped straight to an application channel creation here.

We need to do the steps in this order:
1) generate the system channel genesis block
2) use the system channel genesis block to bootstrap ordering nodes
3) generate application channel create transaction
4) create the application channel on the ordering service

The 'create a channel' tutorial already does these steps (at least steps 1,3,4). Do we want to document the steps twice, or just point to the create a channel tutorial?

docs/source/deployorderer/ordererchecklist.md
```diff
@@ -0,0 +1,307 @@
...
+* [Consensus.*](#consensus)
+
+## General.ListenAddress
+
+```
+# Listen address: The IP on which to bind to listen.
+ListenAddress: 127.0.0.1
+```
+
+**`ListenAddress`**: (default value should be overridden) This is the location where the orderer will listen, for example, `0.0.0.0:9051`.
```
denyeart: Unlike the peer, the orderer.yaml separates the address and the port.

```suggestion
**`ListenAddress`**: (default value should be overridden) This is the location where the orderer will listen, for example, `0.0.0.0`.
```

docs/source/deployorderer/ordererchecklist.md
```diff
@@ -0,0 +1,307 @@
...
+
+## General.Cluster
+
+```
+# SendBufferSize is the maximum number of messages in the egress buffer.
+# Consensus messages are dropped if the buffer is full, and transaction
+# messages are waiting for space to be freed.
+SendBufferSize: 10
+# ClientCertificate governs the file location of the client TLS certificate
+# used to establish mutual TLS connections with other ordering service nodes.
```
denyeart: This comment is being added in PR #2123 :
`# If not set, the server General.TLS.Certificate is re-used.`

docs/source/deployorderer/ordererchecklist.md
```diff
@@ -0,0 +1,307 @@
...
+ClientCertificate:
+# ClientPrivateKey governs the file location of the private key of the client TLS certificate.
```
denyeart: This comment is being added in PR #2123 :
`# If not set, the server General.TLS.PrivateKey is re-used.`

docs/source/deployorderer/ordererchecklist.md
```diff
@@ -0,0 +1,307 @@
...
+# ListenAddress defines the IP on which to listen to intra-cluster communication.
+ListenAddress:
+# ServerCertificate defines the file location of the server TLS certificate used for intra-cluster
+# communication.
+ServerCertificate:
+# ServerPrivateKey defines the file location of the private key of the TLS certificate.
+ServerPrivateKey:
+```
+
+When TLS is enabled you must configure the `General.Cluster.ClientCertificate:` and the `General.Cluster.ClientPrivateKey:` to be the server-side signed certificate for the node, the same as the value in `General.TLS.Certificate`.
```
denyeart: No, it doesn't have to be the same as General.TLS.Certificate.
It can be a different Certificate and PrivateKey pair. However, to keep the config simple, it is suggested to keep them the same.
As of v2.3.0, if unset, the ClientCertificate and ClientPrivateKey default to the server General.TLS.Certificate and General.TLS.PrivateKey.

docs/source/deployorderer/ordererdeploy.md
pamandrejko: Good catch.
I grabbed the wrong profile. 
I've updated the instructions to be more specific to the system channel genesis block. 
and used the `TwoOrgsOrdererGenesis:` profile  from the Test network instead.

For steps 3 and 4 I link to those instructions in the create a channel tutorial.

docs/source/deployorderer/ordererplan.md
```diff
@@ -0,0 +1,84 @@
+# Planning for an ordering service
+
+Audience: Architects, network operators, users setting up a production Fabric network who are familiar with Transport Layer Security (TLS), Public Key Infrastructure (PKI) and Membership Service Providers (MSPs).
+
+Check out the conceptual topic on [The Ordering Service](../orderer/ordering_service.html) for an overview on ordering service concepts, implementations, and the role an ordering service plays in a transaction.
+
+In a Hyperledger Fabric network, a node or collection of nodes together form what's called an "ordering service", which literally orders transactions into blocks that are processed peers and committed to their ledgers. This separates Fabric from other distributed blockchains, such as Ethereum and Bitcoin, in which this ordering is done by any and all nodes, requiring the establishment of probabilistic consensus algorithms which eventually guarantee ledger consistency to a high degree of probability, but which are still vulnerable to divergent ledgers (also known as a ledger “fork”), where different participants in the network have a different view of the accepted order of transactions.
```
denyeart: Beyond the suggestion I'm making in the first sentence, the last sentence is still too long and complex to understand. I'm not understanding the point you are trying to make... whether peer nodes participate in ordering or not has nothing to do with probabilistic consensus and potential forks. I think the entire last sentence could be removed.

```suggestion
In a Hyperledger Fabric network, a node or collection of nodes together form what's called an "ordering service", which literally orders transactions into blocks, which peers will then validate and commit to their ledgers. This separates Fabric from other distributed blockchains, such as Ethereum and Bitcoin, in which this ordering is done by any and all nodes, requiring the establishment of probabilistic consensus algorithms which eventually guarantee ledger consistency to a high degree of probability, but which are still vulnerable to divergent ledgers (also known as a ledger “fork”), where different participants in the network have a different view of the accepted order of transactions.
```

docs/source/deployorderer/ordererplan.md
```diff
@@ -0,0 +1,84 @@
...
+
+## Transport Layer Security (TLS) enablement
+
+To prevent “man in the middle” attacks and otherwise secure communications, the use of TLS is a requirement for any production network. Therefore, in addition to registering your ordering nodes identities with your organization CA, you will also need to create certificates for your ordering nodes with the TLS CA of your organization. These TLS certificates will be used by the ordering nodes when communicating with the network.
+
+## Creating the system channel genesis block
+
+Note: “consenters” refers to the nodes servicing a particular channel at a particular time. For each channel, the “consenters” may be a subset of the ordering nodes available in the system channel.
+
+Every ordering node must be bootstrapped with a configuration block, known as the "genesis block", that contains configuration of a channel. While application channels also contain genesis blocks, the genesis block an ordering node is bootstrapped with is the genesis block of what’s called the “orderer system channel”.
```
denyeart: Doesn't have to be genesis block, can also be a later configuration block (note, this is why the orderer.yaml config option name changed from 'GenesisFile' to 'BootstrapFile' starting in v2.0.

```suggestion
Every ordering node must be bootstrapped with a configuration block from the system channel, either the system channel "genesis block" or a later configuration block. This guide will assume you are creating a new ordering service and will therefore bootstrap ordering nodes from a system channel genesis block.
```

docs/source/deployorderer/ordererchecklist.md
```diff
@@ -0,0 +1,307 @@
...
+# communication.
+ServerCertificate:
+# ServerPrivateKey defines the file location of the private key of the TLS certificate.
+ServerPrivateKey:
+```
+
+As of Fabric v2.3.0, if unset, the `ClientCertificate` and `ClientPrivateKey` default to the server `General.TLS.Certificate` and `General.TLS.PrivateKey`.
+
+**`ClientCertificate`**:  Provide the path to and filename of the public certificate (also known as a signed certificate) generated by your TLS CA for this node.
+**`ClientPrivateKey`**: Provide the path to and filename of the private key generated by your TLS CA for this node.
```
denyeart: This entry and others in this file are not on separate lines in the generated doc.

docs/source/deployorderer/ordererdeploy.md
```diff
@@ -0,0 +1,299 @@
...
+  - **Orderer organization MSP**
+  - **Orderer TLS CA certificates**
+  - **Orderer local MSP (enrollment certificate and private key of the orderer)**
+
+You will either need to use the Fabric CA client to generate the certificates directly into the recommended folder structure or you will need to copy the generated certificates to their recommended folders after they are generated. Whichever method you choose, most users are ultimately likely to script this process so it can be repeated as needed. A list of the certificates and their locations is provided here for your convenience.
+
+If you are using a containerized solution for running your network (which for obvious reasons is a popular choice), **it is a best practice to mount these folders (volumes) external to the container where the node itself is running. This will allow the certificates to be used to create a new node should the node container go down, become corrupted, or is restarted.**
```
denyeart: My prior comment on this paragraph was not addressed, even though Pam said "Done". Makes me wonder if a whole set of updates was missed in the latest commit?

https://github.com/hyperledger/fabric/pull/2080#discussion_r523108405

docs/source/deployorderer/ordererdeploy.md
```diff
@@ -0,0 +1,299 @@
...
+
+Note that the local MSP contains the signed certificate (public key) and the private key for the orderer. The private key is used by the node to sign transactions, and is therefore not shared and must be secured. For maximum security, a Hardware Security Module (HSM) can be configured to generate and store this private key.
+
+### Create the ordering service genesis block
+
+The first channel that is created in a Fabric network is the "system" channel. The system channel defines the set of ordering nodes that form the ordering service and the set of organizations that serve as ordering service administrators. Peers transact on private "application" channels that are derived from the ordering service system channel, which also defines the "consortium" (the peer organizations known to the ordering service). Therefore, before you can deploy an ordering service, you need to generate the system channel configuration by creating the "genesis block". We use the system channel genesis block to bootstrap each ordering node. Fabric provides the `configtxgen` tool for this purpose.
```
denyeart:
```suggestion
The first channel that is created in a Fabric network is the "system" channel. The system channel defines the set of ordering nodes that form the ordering service and the set of organizations that serve as ordering service administrators. Peers transact on private "application" channels that are derived from the ordering service system channel, which also defines the "consortium" (the peer organizations known to the ordering service). Therefore, before you can deploy an ordering service, you need to generate the system channel configuration by creating the "genesis block". Fabric provides the `configtxgen` tool for this purpose. We'll then use the generated system channel genesis block to bootstrap each ordering node. 
```

docs/source/deployorderer/ordererdeploy.md
```diff
@@ -0,0 +1,299 @@
...
+- **Orderer:** Which ordering nodes will form the Raft consenter set of the channel.
+- **Policies** Different sections of the file work together to define the channel policies that will govern how organizations interact with the channel and which organizations need to approve channel updates. For the purposes of this tutorial, we will use the defaults that are used by Fabric. For more information about policies, check out [Policies](../policies/policies.html).
+- **Profiles** Each channel profile references information from other sections of the `configtx.yaml` file to build a channel configuration. The profiles are used to create the genesis block of the channel.
+
+The `configtxgen` tool uses `configtx.yaml` file to create the genesis block for the channel. A detailed version of the `configtx.yaml` file is available in the [Fabric sample configuration](https://github.com/hyperledger/fabric/blob/{BRANCH}/sampleconfig/configtx.yaml). Refer to the [Using configtx.yaml to build a channel configuration](../create_channel/create_channel_config.html) tutorial to learn more about the  settings in this file.
+
+#### Generate the genesis block
```
denyeart:
```suggestion
#### Generate the system channel genesis block
```

docs/source/deployorderer/ordererdeploy.md
```diff
@@ -0,0 +1,299 @@
...
+
+After you have completed editing the `configtx.yaml` to reflect the orderer and peer organizations that will participate in your network, run the following command to create the genesis block of the system channel:
+```
+configtxgen -profile TwoOrgsOrdererGenesis -channelID system-channel -outputBlock ./system-genesis-block/genesis.block
+```
+
+Where:
+- `-profile` refers to the `TwoOrgsOrdererGenesis` profile in `configtx.yaml`.
+- `-channelID` must be `system-channel`.
```
denyeart: Doesn't have to be "system-channel", can be any name.

docs/source/deployorderer/ordererdeploy.md
```diff
@@ -0,0 +1,299 @@
...
+- `-outputCreateChannelTx` refers to the location of the generated genesis block.
```
denyeart:
```suggestion
- `outputBlock` refers to the location of the generated genesis block.
```

docs/source/deployorderer/ordererdeploy.md
```diff
@@ -0,0 +1,299 @@
...
+
+When the command is successful, you will see logs of `configtxgen` loading the `configtx.yaml` file and printing a channel creation transaction:
+```
+INFO 001 Loading configuration
+INFO 002 Loaded configuration: /Usrs/fabric-samples/test-network/configtx/configtx.yaml
+INFO 003 Generating new channel configtx
+INFO 004 Writing new channel tx
```
denyeart:
```suggestion
INFO 004 Generating genesis block
INFO 005 Creating system channel genesis block
INFO 006 Writing genesis block
```

docs/source/deployorderer/ordererdeploy.md
```diff
@@ -0,0 +1,299 @@
...
+```
+
+Make note of the generated `.pb` file. This is your genesis block and will be referenced in the `orderer.yaml` file below.
```
denyeart:
```suggestion
Make note of the generated output block filename. This is your genesis block and will be referenced in the `orderer.yaml` file below.
```

docs/source/deployorderer/ordererdeploy.md
```diff
@@ -0,0 +1,299 @@
...
+- `General.ListenAddress` - Hostname that the ordering node listens on.
+- `General.ListenPort` - Port that the ordering node listens on.
+- `General.TLS.Enabled: true` - Server-side TLS should be enabled in all production networks.
+- `General.TLS.PrivateKey ` - Ordering node private key from TLS CA.
+- `General.TLS.Certificate ` - Ordering node signed certificate (public key) from the TLS CA.
+- `General.TLS.RootCAS` - This value should be unset.
+- `General.Cluster.ClientCertificate` - Ordering node signed certificate (public key) from the TLS CA.
+- `General.Cluster.ClientPrivateKey` - Ordering node private key from TLS CA.
+- `General.BoostrapMethod:file` - Start ordering service with a system channel.
+- `General.BootstrapFile` - Path to and name of the genesis block `.pb` file for the ordering service system channel.
```
denyeart:
```suggestion
- `General.BootstrapFile` - Path to and name of the genesis block file for the ordering service system channel.
```

docs/source/deployorderer/ordererdeploy.md
```diff
@@ -0,0 +1,299 @@
...
+
+At a minimum, you need to configure the following parameters:
+- `General.ListenAddress` - Hostname that the ordering node listens on.
+- `General.ListenPort` - Port that the ordering node listens on.
+- `General.TLS.Enabled: true` - Server-side TLS should be enabled in all production networks.
+- `General.TLS.PrivateKey ` - Ordering node private key from TLS CA.
+- `General.TLS.Certificate ` - Ordering node signed certificate (public key) from the TLS CA.
+- `General.TLS.RootCAS` - This value should be unset.
+- `General.Cluster.ClientCertificate` - Ordering node signed certificate (public key) from the TLS CA.
+- `General.Cluster.ClientPrivateKey` - Ordering node private key from TLS CA.
```
denyeart: These two are optional as of v2.3 since they now default based on General.TLS.Certificate and General.TLS.PrivateKey

docs/source/deployorderer/ordererdeploy.md
```diff
@@ -0,0 +1,299 @@
...
+INFO 01b Applied config change to add node 1, current nodes in channel: [1] channel=syschannel node=1
+INFO 01c Applied config change to add node 2, current nodes in channel: [1 2] channel=syschannel node=1
+INFO 01d Applied config change to add node 3, current nodes in channel: [1 2 3] channel=syschannel node=1
+INFO 01e raft.node: 1 elected leader 2 at term 11 channel=syschannel node=1
+INFO 01f Raft leader changed: 0 -> 2 channel=syschannel node=1
+```
+
+## Next steps
+
+Your ordering service is started and ready to order transactions into blocks. Working together to form the ordering service, the orderers receive transactions from application clients and package the transactions into a well-defined sequence of blocks. These blocks form the blockchain and are ultimately distributed to all peers on the channel to be committed to their ledger.
```
denyeart: Actually, at this point, we are not ready to receive transactions from application clients, since we've only created the system channel, we haven't crated application channels yet.

docs/source/deployorderer/ordererplan.md
joealewine: I think it's important for new users to understand that one of the key differences between Fabric and other blockchain networks is that the ordering process is performed by a dedicated node. This language is largely copied from the ordering service key concept doc. I think it's probably fine to stop at "any and all nodes", though.

joealewine (commented):
Thanks for these comments, Dave!

docs/source/deployorderer/ordererdeploy.md
pamandrejko: Since this will be ported back to v2.2 I think we wanted to make another pass with updates like this for v2.3 after this gets merged for 2.2

pamandrejko: Although I did ask for this fix to be ported back to v2.2. Let me check if it made it in.

pamandrejko: It was backported [here](https://github.com/hyperledger/fabric/pull/2078#issuecomment-726768203) so we can add it

docs/source/deployorderer/ordererchecklist.md
```diff
@@ -0,0 +1,307 @@
...
+## General.LocalMSPDir
+
+```
+# LocalMSPDir is where to find the private crypto material needed by the
+# orderer. It is set relative here as a default for dev environments but
+# should be changed to the real location in production.
+LocalMSPDir: msp
+```
+
+**`LocalMSPDir`**: (default value should not be overridden) This is the path to the ordering node's local MSP, which must be created before it can be deployed. The path can be absolute or relative to `FABRIC_CFG_PATH` (by default, it is `/etc/hyperledger/fabric` in the orderer image). Unless an absolute path is specified to a folder named something other than "msp", the ordering node defaults to looking for a folder called “msp” at the path (in other words, `FABRIC_CFG_PATH/msp`) and when using the orderer image: `/etc/hyperledger/fabric/msp`. If you are using the recommended folder structure described in the [Registering and enrolling identities with a CA](https://hyperledger-fabric-ca.readthedocs.io/en/release-1.4/deployguide/use_CA.html) topic, it would be relative to the `FABRIC_CFG_PATH` as follows:
```
denyeart: This will often be overridden.
This paragraph seems to have 2x the number of words as needed :-)

docs/source/deployorderer/ordererchecklist.md
```diff
@@ -0,0 +1,307 @@
...
+            # Location of key store. If this is unset, a location will be
+            # chosen using: 'LocalMSPDir'/keystore
+            FileKeyStore:
+                KeyStore:
+```
+
+(Optional) This section is used to configure the Blockchain crypto provider.
+
+* **`BCCSP.Default:`** If you plan to use a Hardware Security Module (HSM), then this must be set to `PKCS11`.
+* **`BCCSP.PKCS11.*:`** Provide this set of parameters according to your HSM configuration. Refer to this [example]((../hsm.html) of an HSM configuration for more information.
```
denyeart: The PKCS11 section should be listed in the sample config above.

docs/source/deployorderer/ordererchecklist.md
```diff
@@ -0,0 +1,307 @@
...
+        # at the TLS layer to access all resources.
+        ClientAuthRequired: false
+
+        # Paths to PEM encoded ca certificates to trust for client authentication
+        ClientRootCAs: []
+```
+
+* **`ListenAddress`**: (required when using the operations service) Specify the address and port of the operations server.
+* **`Enabled`**: (required when using the operations service) Must be `true` if the operations service is being used.
+* **`Certificate`**: (required when using the operations service) Can be the same file as the `orderer.tls.cert.file`.
```
denyeart:
```suggestion
* **`Certificate`**: (required when using the operations service) Can be the same file as the `General.TLS.Certificate`.
```

docs/source/deployorderer/ordererchecklist.md
```diff
@@ -0,0 +1,307 @@
...
+* **`PrivateKey`**: (required when using the operations service) Can be the same file as the `orderer.tls.key.file`.
```
denyeart:
```suggestion
* **`PrivateKey`**: (required when using the operations service) Can be the same file as the `General.TLS.PrivateKey`.
```

docs/source/deployorderer/ordererchecklist.md
```diff
@@ -0,0 +1,307 @@
...
+
+      # The interval at which locally cached counters and gauges are pushed
+      # to statsd; timings are pushed immediately
+      WriteInterval: 30s
+
+      # The prefix is prepended to all emitted statsd metrics
+      Prefix:
+```
+
+* **`Statsd.Provider`**: Must be `enabled` to use `StatsD` or `Prometheus` metrics for the ordering node.
```
denyeart: I know that lowercase values work. Not sure about the mixed case values unless somebody has mentioned this.

docs/source/deployorderer/ordererchecklist.md
```diff
@@ -0,0 +1,307 @@
...
+* **`Statsd.Address`**: (required to use `StatsD` or `Prometheus` metrics for the ordering node) When `StatsD` is enabled, you will need to configure the `hostname` and `port` of the `StatsD` server so that the ordering node can push metric updates.
```
denyeart:
```suggestion
* **`Statsd.Address`**: (required to use `StatsD` metrics for the ordering node) When `StatsD` is enabled, you will need to configure the `hostname` and `port` of the `StatsD` server so that the ordering node can push metric updates.
```

denyeart (commented):
This review focuses on the checklist.

docs/source/deployorderer/ordererdeploy.md
```diff
@@ -0,0 +1,299 @@
...
+
+If you are using a containerized solution for running your network (which for obvious reasons is a popular choice), **it is a best practice to mount volumes for the certificate directories external to the container where the node itself is running. This will allow the certificates to be used by an ordering node container, regardless whether the ordering node container goes down, becomes corrupted, or is restarted.**
+
+#### TLS certificates
+
+For the ordering node to launch successfully, the locations of the TLS certificates you specified in the [Checklist for a production ordering node](./ordererchecklist.html) must point to the correct certificates. To do this:
+
+- Copy the **TLS CA Root certificate** that contains the public key that is associated with the signing (private) key certificate, which by default is called `ca-cert.pem`, to the orderer organization MSP definition  `organizations/ordererOrganizations/ordererOrg1.example.com/msp/tlscacerts/tls-cert.pem`.
```
denyeart: Was this a copy/paste from somewhere? The description is confusing... I was expecting it to describe what a TLS CA root certificate is, not what a certificate is in general.

docs/source/deployorderer/ordererdeploy.md
```diff
@@ -0,0 +1,299 @@
...
+- Copy the **CA Root certificate** that contains the public key that is associated with the signing (private) key certificate, which by default is called `ca-cert.pem`, to the orderer organization MSP definition  `organizations/ordererOrganizations/ordererOrg1.example.com/msp/cacerts/ca-cert.pem`.
```
denyeart: Was this a copy/paste from somewhere? The description is confusing... I was expecting it to describe what a CA root certificate is, not what a certificate is in general.

docs/source/deployorderer/ordererdeploy.md
```diff
@@ -0,0 +1,299 @@
...
+- When you enroll the orderer identity with the TLS CA, the public key is generated in the `signcerts` folder, and the private key is located in the `keystore` directory. Rename the private key in the `keystore` folder to `orderer0-tls-key.pem` so that it can be easily recognized later as the TLS private key for this node.
+- Copy the orderer TLS certificate and private key files to `organizations/ordererOrganizations/ordererOrg1.example.com/orderers/orderer0.ordererOrg1.example.com/tls`. The path and name of the certificate and private key files correspond to the values of the `General.TLS.Certificate` and `General.TLS.PrivateKey` parameters in the `orderer.yaml`.
+- If you require mutual TLS (`General.TLS.ClientAuthRequired` set to true), you need to indicate to the orderer which TLS CA root certificates to use to authorize clients. Copy the organization's TLS CA root certificate `ca-cert.pem` to `organizations/ordererOrganizations/ordererOrg1.example.com/orderers/orderer0.ordererOrg1.example.com/tls/tls-cert.pem` so that the organization's clients will be authorized.  If `General.TLS.ClientAuthRequired` is set to false, you can skip this step.
```
denyeart: The TLS CA root cert from EACH org that is going to connect to the orderer should be copied, that is, application clients and peers that will connect to this orderer.

docs/source/deployorderer/ordererdeploy.md
```diff
@@ -0,0 +1,299 @@
...
+
+Note that the local MSP contains the signed certificate (public key) and the private key for the orderer. The private key is used by the node to sign transactions, and is therefore not shared and must be secured. For maximum security, a Hardware Security Module (HSM) can be configured to generate and store this private key.
+
+### Create the ordering service genesis block
+
+The first channel that is created in a Fabric network is the "system" channel. The system channel defines the set of ordering nodes that form the ordering service and the set of organizations that serve as ordering service administrators. Peers transact on private "application" channels that are derived from the ordering service system channel, which also defines the "consortium" (the peer organizations known to the ordering service). Therefore, before you can deploy an ordering service, you need to generate the system channel configuration by creating the "genesis block". abric provides the `configtxgen` tool for this purpose. We'll then use the generated system channel genesis block to bootstrap each ordering node.
```
denyeart:
```suggestion
The first channel that is created in a Fabric network is the "system" channel. The system channel defines the set of ordering nodes that form the ordering service and the set of organizations that serve as ordering service administrators. Peers transact on private "application" channels that are derived from the ordering service system channel, which also defines the "consortium" (the peer organizations known to the ordering service). Therefore, before you can deploy an ordering service, you need to generate the system channel configuration by creating the system channel "genesis block". Fabric provides the `configtxgen` tool for this purpose. We'll then use the generated system channel genesis block to bootstrap each ordering node.
```

docs/source/deployorderer/ordererdeploy.md
```diff
@@ -0,0 +1,299 @@
...
+
+#### Set up the `configtxgen` tool
+
+All channels are created by building a channel creation transaction and submitting the transaction to the ordering service. The channel creation transaction specifies the initial configuration of the channel and is used by the ordering service to write the channel genesis block. While it is possible to build the channel creation transaction file manually, it is easier to use the [configtxgen](../commands/configtxgen.html) tool, which works by reading a `configtx.yaml` file that defines the configuration of your channel and then writing the relevant information into a configuration block known as the "genesis block".
```
denyeart: This paragraph seems to talk about application channels, while the prior paragraph and next paragraphs talk about system channel genesis block creation. I'd highlight that configtxgen is used for both.

docs/source/deployorderer/ordererchecklist.md
```diff
@@ -0,0 +1,307 @@
...
+RootCAs:
+  - tls/ca.crt
+ClientAuthRequired: false
+ClientRootCAs:
+```
+
+* **`Enabled`**: (default value should be overridden) In a production network, you should be using TLS-secured communications. This value should be `true`.
+* **`PrivateKey`**: (default value should be overridden). Provide the path to and file name of the private key generated by your TLS CA for this node.
+* **`Certificate`**: (default value should be overridden) Provide the path to and filename of the public certificate (also known as the sign certificate) generated by your TLS CA for this node.
+* **`RootCAs`**: (should be commented out) This parameter needs to be unset for normal use. It can be used to augment the orderer TLS certificate in the channel configuration orderer `MSPDir/tlscacerts` folder.
```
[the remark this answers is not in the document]
joealewine: As I recall, the language that was here was suggested by Jason in the google doc this doc was created from. Also I'm not really sure how this one field in orderer.yaml can augment a set of TLS certs from multiple MSPs?

docs/source/deployorderer/ordererchecklist.md
```diff
@@ -0,0 +1,307 @@
...
+# Certificate governs the file location of the server TLS certificate.
+Certificate: tls/server.crt
+RootCAs:
+  - tls/ca.crt
+ClientAuthRequired: false
+ClientRootCAs:
+```
+
+* **`Enabled`**: (default value should be overridden) In a production network, you should be using TLS-secured communications. This value should be `true`.
+* **`PrivateKey`**: (default value should be overridden). Provide the path to and file name of the private key generated by your TLS CA for this node.
```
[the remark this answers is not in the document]
joealewine: I think it's just missing commas: "Provide the path to, and filename of,"

docs/source/deployorderer/ordererchecklist.md
```diff
@@ -0,0 +1,307 @@
...
+
+```
+# Bootstrap method: The method by which to obtain the bootstrap block
+# system channel is specified. The option can be one of:
+#   "file" - path to a file containing the genesis block or config block of system channel
+#   "none" - allows an orderer to start without a system channel configuration
+BootstrapMethod: file
+```
+
+* **`BootstrapMethod`**: (default value should not be overridden) Unless you plan to use a file type other than “file”, this value should be left as is.
```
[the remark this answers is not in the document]
joealewine: This first cut at the deployment guide describes the system channel process, so let's leave this as is and add the note about the participation API for later. This edit would only confuse people as it would be the only reference to the new channel creation process in the whole guide.

docs/source/deployorderer/ordererchecklist.md
joealewine: Given that cert stuff like this is the thing people seem to most often screw up, I don't mind oversharing.

docs/source/deployorderer/ordererchecklist.md
joealewine: I used "StatsD" to refer to StatsD generally (as this is the proper capitalization from what I could find online, but used the capitalization from the YAML when explicitly referring to the YAML.

docs/source/deployorderer/ordererdeploy.md
joealewine: It's basically identical to what we have in the peer deployment guide.

docs/source/deployorderer/ordererdeploy.md
joealewine: It's basically identical to what we have in the peer deployment guide.

docs/source/deployorderer/ordererdeploy.md
joealewine: In this tutorial, there's only one org. But I'll add a clarification note here anyway.

docs/source/deployorderer/ordererdeploy.md
joealewine: I don't get an "application channel" vibe from this. Not sure exactly what you're referring to. I changed this first sentence to: 

"Both the system channel and all application channels are created by building a channel creation transaction and submitting the transaction to the ordering service."

Does that help?

docs/source/deployorderer/ordererdeploy.md
joealewine: Which too? Was this added and I missed it? @pamandrejko

docs/source/deployorderer/ordererchecklist.md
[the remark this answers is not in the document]
joealewine: As I recall, the language that was here was suggested by Jason in the google doc this doc was created from. Also I'm not really sure how this one field in orderer.yaml can augment a set of TLS certs from multiple MSPs?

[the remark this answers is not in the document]
denyeart: Let's take test network as an example. There are three orgs in the channel config (orderer org, org1, org2). Each of these orgs has a tls ca cert in their msp/tlscacerts directory that gets encoded into the channel config. This set of three tls ca certs are used to verify connections.  Now, at a later time, suppose one of those certs gets expired and the associated org is not able to immediately update their tls ca cert in the channel config. That org could provide a new tls ca cert, and it could be included in orderer.yaml for this orderer to augment the list of three from the channel config. So now it has a list of four (3 from channel config, one from orderer.yaml). This is consistent with what Jason said in the google doc - his main point was that any cert here would augment instead of override the list from the channel config.

This is a rare scenario, which is why we say it should be commented out, even in the sample orderer.yaml.

(Furthermore, I don't think this list is even used anymore, but that's a different story).

docs/source/deployorderer/ordererchecklist.md
```diff
@@ -0,0 +1,307 @@
...
+ClientAuthRequired: false
+ClientRootCAs:
+```
+
+* **`Enabled`**: (default value should be overridden) In a production network, you should be using TLS-secured communications. This value should be `true`.
+* **`PrivateKey`**: (default value should be overridden). Provide the path to and file name of the private key generated by your TLS CA for this node.
+* **`Certificate`**: (default value should be overridden) Provide the path to and filename of the public certificate (also known as the sign certificate) generated by your TLS CA for this node.
+* **`RootCAs`**: (should be commented out) This parameter needs to be unset for normal use. It can be used to augment the orderer TLS certificate in the channel configuration orderer `MSPDir/tlscacerts` folder.
+* **`ClientAuthRequired`**: (Mutual TLS only) Setting this value to “true” will enable mutual TLS on your network, and must be done for the entire network, not just one node.
+* **`ClientRootCAs`**: (Mutual TLS only)  Can be left blank if mutual TLS is disabled. If mutual TLS is enabled, this is the path to the root certificate of the TLS CA.
```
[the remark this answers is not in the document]
denyeart: Jason made the same point about this one in the google doc actually, which I missed during my initial review... the set of org tls ca certs from the channel config will be used to verify client connections. You can add additional tls ca certs in ClientRootCAs to augment that list. Again, it would be rare.

docs/source/deployorderer/ordererchecklist.md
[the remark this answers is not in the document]
denyeart: ok

docs/source/deployorderer/ordererchecklist.md
denyeart: It doesn't matter what case is used online or in the YAML keys. The valid values for Metrics.Provider are:
"disabled"
"statsd"
"prometheus"

docs/source/deployorderer/ordererdeploy.md
denyeart: Ok, that's why I asked if it was copy/paste, so that we can fix in both places.  Lets just leave it as "TLS CA Root certificate", I feel talking about public and private signing keys are only going to confuse people, as it gives the impression these certs are going to be used for signing transactions or communications. They're not, they're only used for verifying issued certs.

docs/source/deployorderer/ordererdeploy.md
denyeart: Ok, that's why I asked if it was copy/paste, so that we can fix in both places.  Lets just leave it as "CA Root certificate", I feel talking about public and private signing keys are only going to confuse people, as it gives the impression these certs are going to be used for signing transactions or communications. They're not, they're only used for verifying issued certs.

docs/source/deployorderer/ordererdeploy.md
denyeart: No. 
"All channels are created by building a channel creation transaction and submitting the transaction to the ordering service."
That is true for application channels.
It is not true for system channels. First of all, if you are creating the system channel there is not yet an ordering service to submit to!  Second of all, for the system channel you generate a system channel genesis block (not a channel create transaction).

docs/source/deployorderer/ordererchecklist.md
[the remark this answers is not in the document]
jyellick: +1 I think I may have used a '/' notation meaning "or", I did not mean an explicit path.  Dave's wording is correct.

docs/source/deployorderer/ordererchecklist.md
```diff
@@ -0,0 +1,320 @@
...
+# ListenAddress defines the IP on which to listen to intra-cluster communication.
+ListenAddress:
+# ServerCertificate defines the file location of the server TLS certificate used for intra-cluster
+# communication.
+ServerCertificate:
+# ServerPrivateKey defines the file location of the private key of the TLS certificate.
+ServerPrivateKey:
+```
+
+As of Fabric v2.3.0, if unset, the `ClientCertificate` and `ClientPrivateKey` default to the server `General.TLS.Certificate` and `General.TLS.PrivateKey`.
```
jyellick: ... when the orderer is not configured to use a separate cluster port.

docs/source/deployorderer/ordererchecklist.md
```diff
@@ -0,0 +1,320 @@
...
+
+* **`ClientCertificate`**:  Provide the path to, and filename of, the public certificate (also known as a signed certificate) generated by your TLS CA for this node.
+* **`ClientPrivateKey`**: Provide the path to, and filename of, the private key generated by your TLS CA for this node.
+
+In general, these four parameters would only need to be configured if you want to configure a separate listener and TLS certificates for intra-cluster communication (with other Raft orderers), as opposed to using the listener that peer clients and application clients utilize. This is an advanced deployment option. Either set these four parameters together or leave them unset.
```
jyellick: Perhaps we should be saying "all six parameters" here and above?  It's a little awkward because you may only set the first two, but if you set the last four, the first two must be set as well.

docs/source/deployorderer/ordererchecklist.md
```diff
@@ -0,0 +1,320 @@
...
+* **`BCCSP.PKCS11.*:`** Provide this set of parameters according to your HSM configuration. Refer to this [example]((../hsm.html) of an HSM configuration for more information.
+
+## FileLedger.Location
+
+```
+# Location: The directory to store the blocks in.
+Location: /var/hyperledger/production/orderer
+```
+
+* **`Location`**: (default value should be overridden, especially if more than one node will be created by this organization) Every channel on which the node is a consenter will have its own subdirectory at this location. The user running the orderer needs to own and have write access to this directory. **The best practice is to store this data in persistent storage**. This prevents the ledger from being lost if your orderer containers are destroyed for some reason.
```
jyellick: I'm a little confused why having more than one node means this location should be overridden? I guess you could put the host name into the path in the container, but you could also just name your volume properly.

My guess is, you are referring to running two non-containerized binaries on the same host.  But it's odd here that we assume this, while below for the operations endpoint we talk about "in the unlikely event where two ordering nodes are running on the same node"

docs/source/deployorderer/ordererchecklist.md
```diff
@@ -0,0 +1,320 @@
...
+* **`Certificate`**: (required when using the operations service) Can be the same file as the `General.TLS.Certificate`.
+* **`PrivateKey`**: (required when using the operations service) Can be the same file as the `General.TLS.PrivateKey`.
+* **`ClientAuthRequired`**: (required when using the operations service) Must be set to `true` to enable mutual TLS between the client and the server.
+* **`ClientRootCAs`**: (required when using the operations service) Similar to the client root CA cert file in TLS, it contains a list of client root CA certificates that can be used to verify client certificates. If the client enrolled with the orderer organization CA, then this value is the orderer organization root CA cert.
+
+## Metrics.*
+
+By default this is disabled, but if you want to monitor the metrics for the orderer, you need to use `StatsD` as your metric provider. `StatsD` uses a "push" model, pushing metrics from the orering node to a `StatsD` endpoint. Because of this, it does not require configuration of the operations service itself. See the list of [Available metrics for the orderer](../metrics_reference.html#orderer-metrics).
```
jyellick: These seems like it was copied maybe? But it's very misleading to me.

> if you want to monitor the metrics for the orderer, you need to use `StatsD` as your metric provider

You can monitor metrics using prometheus too?  I would think it should read: "if you want to monitor the metrics for the orderer, you must set this value to either `statsd` or `prometheus`.  Then have a paragraph about each.

docs/source/deployorderer/ordererchecklist.md
```diff
@@ -0,0 +1,320 @@
...
+      # The prefix is prepended to all emitted statsd metrics
+      Prefix:
+```
+
+* **`Provider`**: Must be `enabled` to use `StatsD` or `Prometheus` metrics for the ordering node.
+* **`Statsd.Address`**: (required to use `StatsD` metrics for the ordering node) When `StatsD` is enabled, you will need to configure the `hostname` and `port` of the `StatsD` server so that the ordering node can push metric updates.
+
+## Consensus.*
+
+These values could vary depending on consensus plugin. For instance, a BFT plugin would very likely have a different set of options here.
```
jyellick: I think this is my text verbatim, but I think perhaps it might make more sense to say something like:

"The values of this section vary by consensus plugin.  If you are using the recommended etcdraft consensus plugin, see the following documentation.  If you are using a different consensus plugin, refer to its documentation for allowed keys and recommended values.

docs/source/deployorderer/ordererchecklist.md
```diff
@@ -0,0 +1,320 @@
...
+# WALDir specifies the location at which Write Ahead Logs for etcd/raft are
+# stored. Each channel will have its own subdir named after channel ID.
+WALDir: /var/hyperledger/production/orderer/etcdraft/wal
+
+# SnapDir specifies the location at which snapshots for etcd/raft are
+# stored. Each channel will have its own subdir named after channel ID.
+SnapDir: /var/hyperledger/production/orderer/etcdraft/snapshot
+```
+
+* **`WALDir`**: (default value should be overridden.) This is the path to the write ahead logs on the local filesystem of the ordering node. It can be an absolute path or relative to `FABRIC_CFG_PATH`. It defaults to `/var/hyperledger/production/orderer/etcdraft/wal`. Each channel will have its own subdirectory named after the channel ID. The user running the ordering node needs to own and have write access to this directory.
```
jyellick: it's a little odd that for WALDir and SnapDir, we don't mention that these needs to be on volumes, but we do for the ledger.  It's critical that these not be destroyed (and we have seen a lot of users make the mistake of not preserving them).

docs/source/deployorderer/ordererchecklist.md
joealewine: Ok changed this to "(default value should be overridden in the unlikely event where two ordering nodes are running on the same node)"

docs/source/deployorderer/ordererchecklist.md
joealewine: Seems sufficient to me to just link to the metrics reference doc, but if we want to add a paragraph of information here for each metrics type, I'm open to it. Not being an expert on metrics, I'm not really sure what we should say, though.

docs/source/deployorderer/ordererdeploy.md
joealewine: Ok done.

docs/source/deployorderer/ordererdeploy.md
joealewine: Ok done.

docs/source/deployorderer/ordererdeploy.md
joealewine: The simplest solution here seemed to be to just cut everything before the sentence beginning with "While it is possible", as that's really what's relevant to setting up configtxgen.

docs/source/deployorderer/ordererdeploy.md
```diff
@@ -0,0 +1,299 @@
...
+- Copy the orderer TLS certificate and private key files to `organizations/ordererOrganizations/ordererOrg1.example.com/orderers/orderer0.ordererOrg1.example.com/tls`. The path and name of the certificate and private key files correspond to the values of the `General.TLS.Certificate` and `General.TLS.PrivateKey` parameters in the `orderer.yaml`.
+- If you require mutual TLS (`General.TLS.ClientAuthRequired` set to true), you need to indicate to the orderer which TLS CA root certificates to use to authorize clients. Copy the TLS CA root certificate `ca-cert.pem` to `organizations/ordererOrganizations/ordererOrg1.example.com/orderers/orderer0.ordererOrg1.example.com/tls/tls-cert.pem` so that the organization's clients will be authorized. If more than one organization is going to connect to this organization, you would need to add their TLS certificates also. If `General.TLS.ClientAuthRequired` is set to false, you can skip this step.
```
denyeart: I believe this entire bullet (line 80) relates to General.TLS.ClientRootCAs, which is not common to be set, as explained in the property checklist. Therefore I'd recommend to remove this entire bullet, as it is more likely to confuse than help. In the rare cases where people need to set it, they can see the reference explanation in the checklist.

docs/source/deployorderer/ordererchecklist.md
```diff
@@ -0,0 +1,324 @@
...
+# ListenAddress defines the IP on which to listen to intra-cluster communication.
+ListenAddress:
+# ServerCertificate defines the file location of the server TLS certificate used for intra-cluster
+# communication.
+ServerCertificate:
+# ServerPrivateKey defines the file location of the private key of the TLS certificate.
+ServerPrivateKey:
+```
+
+As of Fabric v2.3.0, if unset, the `ClientCertificate` and `ClientPrivateKey` default to the server `General.TLS.Certificate` and `General.TLS.PrivateKey` when the orderer is not configured to use a separate cluster port.
```
pamandrejko: This was ported back to v2.2 as well.  Do we need to even call out the release then?

docs/source/deployorderer/ordererdeploy.md
pamandrejko: We have covered this in the checklist where we say:
As of Fabric v2.3.0, if unset, the `ClientCertificate` and `ClientPrivateKey` default to the server `General.TLS.Certificate` and `General.TLS.PrivateKey` when the orderer is not configured to use a separate cluster port.

Therefore I think we can remove the following two parameters from here since it would not be a typical use case to specify them.
- `General.Cluster.ClientCertificate` - Ordering node signed certificate (public key) from the TLS CA.
- `General.Cluster.ClientPrivateKey` - Ordering node private key from TLS CA.

denyeart (approved):

joealewine:

joealewine:

PR #2077, opened 2020-11-05
osnadmin CLI command reference

wlahti:
#### Type of change

- Documentation update

#### Description

Command reference documentation for the osnadmin CLI.

#### Related issues

[FAB-18309](https://jira.hyperledger.org/browse/FAB-18309)

PR #2081, opened 2020-11-06
[FAB-18179] Channel Participation API tutorial

pamandrejko:

Introduce the Channel Participation API via a new tutorial that contains the steps for using the osnadmin CLI with your existing ordering service or by creating a new one.

#### Type of change

- Documentation update

#### Description

Content coexists with the osnadmin CLI command documentation in PR #2077 

#### Related issues
FAb017712

#### Release Note
Fabric v2.3 introduces the ability to create channels between network members without requiring an ordering service system channel. This feature greatly simplifies the channel creation process and provides a mechanism for Raft consenters to join a channel with zero downtime. See the new tutorial to learn more and get started with this new feature.


docs/source/create_channel/create_channel_participation.md
```diff
@@ -0,0 +1,500 @@
...
+
+**Benefits of the new process:**
+
+ * **Increased Privacy** - A single orderer does not know about all the channels on the ordering service.
+ * **Scalability** - When there is a large number of ordering service nodes (orderers) on the system channel, starting the orderer can take a long time as it has to wait for each orderer to replicate the channel ledger.
+ * **Operational benefits** - Updating an application channel no longer requires write permissions from a separate ordering organization. Also, you now have a way to list all the channels an orderer is a member of. Lastly, you have the ability to easily remove a channel and its respective resources from an orderer.
+
+Whereas in the past an orderer could only be part of a single ordering service, now orderers can now join (or leave) any number of channels as needed, similar to how peers can participate in multiple channels.
```
wlahti: "now orderers can now"

docs/source/create_channel/create_channel_participation.md
```diff
@@ -0,0 +1,500 @@
...
+
+- [Use existing ordering service](#use-existing-ordering-service)
+- [Deploy a new ordering service](#deploy-a-new-ordering-service)
+
+### Use existing ordering service
+
+Before you can take advantage of this feature on a deployed ordering service, you need to remove the system channel from each ordering node in the application channel consenter set. A "mixed mode" of orderers on a channel, where some nodes are part of a system channel and others are not, is not supported. The `osadmin` CLI includes both a `channel list` and a `system-channel remove` command to facilitate the process of removing the system channel. If you prefer to deploy a new ordering service instead, skip ahead to [Deploy a new ordering service](#deploy-a-new-ordering-service).
```
wlahti: There is not a separate "system-channel remove" command, it is simply using the "channel remove" command and specifying the --channel-id of the system channel.

docs/source/create_channel/create_channel_participation.md
```diff
@@ -0,0 +1,500 @@
...
+      },
+      "channels":[
+          {
+            "name": "channel1",
+            "url": "/participation/v1/channels/channel1"
+          }
+      ]
+    }
+    ```
+    Run `osnadmin system-channel remove` to remove the system channel from the node configuration:
```
wlahti: As noted above, this is not a separate command but simply "osnadmin channel remove" and specifying the --channel-id of the system channel.

docs/source/create_channel/create_channel_participation.md
```diff
@@ -0,0 +1,500 @@
...
+    Examine the output of the command to verify that the system channel is no longer listed:
+    ```
+    {
+      "channels":[
+          {
+            "name": "channel1",
+            "url": "/participation/v1/channels/channel1"
+          }
+      ]
+    }
```
wlahti: This output isn't accurate. There should be a "systemChannel" entry that shows

"systemChannel": null 

after removing it.

wlahti (changes requested):
Only made it about halfway through for now. I'll get back to reviewing this later but wanted to get my current comments out there.

docs/source/create_channel/create_channel_participation.md
pamandrejko: Thought I got those! There were four other occurrences so I think I got them all now.

docs/source/create_channel/create_channel_participation.md
pamandrejko: Updated to:
```
{
      "systemChannel": null,
      "channels":[
          {
            "name": "channel1",
            "url": "/participation/v1/channels/channel1"
          }
      ]
    }
```

docs/source/create_channel/create_channel_participation.md
```diff
@@ -0,0 +1,527 @@
...
+- `-channelID` - Specify the name of the channel you want to create. Channel names must be all lower case, less than 250 characters long and match the regular expression ``[a-z][a-z0-9.-]*``. The command uses the `-profile` flag to reference the `SampleAppGenesisEtcdRaft:` profile from `configtx.yaml`.
+
+
+When the command successful, you will see logs of `configtxgen` loading the `configtx.yaml` file and printing a channel creation transaction:
+```
+2020-03-11 16:37:12.695 EDT [common.tools.configtxgen] main -> INFO 001 Loading configuration
+2020-03-11 16:37:12.738 EDT [common.tools.configtxgen.localconfig] Load -> INFO 002 Loaded configuration: /Usrs/fabric-samples/test-network/configtx/configtx.yaml
+2020-03-11 16:37:12.740 EDT [common.tools.configtxgen] doOutputChannelCreateTx -> INFO 003 Generating new channel configtx
+2020-03-11 16:37:12.789 EDT [common.tools.configtxgen] doOutputChannelCreateTx -> INFO 004 Writing new channel tx
+```
```
wlahti, line 464: This looks like it was generated by the -createChannelTx command and not -outputBlock. :)

docs/source/create_channel/create_channel_participation.md
```diff
@@ -0,0 +1,527 @@
...
+- `TLS_CA_ROOT_CERT` is the fully qualified path and file name of the orderer organization TLS CA root certificate and intermediate certificate if using an intermediate TLS CA.
+- `OSN_TLS_SIGN_CERT` is the orderer signed certificate from the TLS CA.
+- `OSN_TLS_PRIVATE_KEY` is the orderer private key from the TLS CA.
+
+For example:
+```
+osnadmin channel join –-channel-id channel1 --config-block genesis_block.pb -o OSN1.example.com:7050 --ca-file $TLS_CA_ROOT_CERT --client-cert $OSN_TLS_SIGN_CERT --client-key $OSN_TLS_PRIVATE_KEY
+```
+
+<!--**Note:** Because this tutorial assumes you have enabled _server-side_ TLS communications on your network, which is recommended for a Production environment, you need to provide the TLS CA root certificate. If _mutual TLS_ is enabled on your network, you also need to pass the `--client-cert`, and `--clientKey` files on the commands which are omitted from the sample commands throughout this tutorial. -->
```
wlahti, line 495: This needs to be updated since mutual TLS is currently required no matter what the network is using for TLS.

docs/source/create_channel/create_channel_participation.md
```diff
@@ -0,0 +1,527 @@
...
+{
+  "name": "channel1",
+  "url": "participation/v1/channels/channel1",
+  "consensusRelation": "consenter",
+  "status": "active",
+  "height": 1
+}
+```
+
+When successful, the channel is created with a ledger height `1`. Because this orderer is joining from the genesis block, the `status` is "active". And because the orderer is part of the channel consenter, its `consensusRelation` is "consenter". We delve more into the `consensusRelation` value in the next section.
```
wlahti:
```suggestion
When successful, the channel is created with a ledger height `1`. Because this orderer is joining from the genesis block, the `status` is "active". And because the orderer is part of the channel's consenter set, its `consensusRelation` is "consenter". We delve more into the `consensusRelation` value in the next section.
```

docs/source/create_channel/create_channel_participation.md
```diff
@@ -0,0 +1,527 @@
...
+
+You can repeat the same command on the other two orderers. Because these two orderers join the channel from the genesis block and are part of the consenter set, their status transitions almost immediately from  "onboarding" to "active" and the "consensusRelation" status is "consenter". At this point, the channel becomes operational because the default policy of a MAJORITY of consenters (as defined in the channel configuration) are "active".
```
wlahti: This isn't policy related, right? It's just that raft requires a majority to form quorum.

docs/source/create_channel/create_channel_participation.md
```diff
@@ -0,0 +1,527 @@
...
+
+Ensure that the orderer organization has write privileges in the channel configuration so the orderer can become a consenter. When the channel update transaction is complete, you can use the `osnadmin channel list` command again to confirm that the `consensusRelation` status for the orderer automatically changes to **consenter**.
+
+```
+{
+  "name": "channel1",
+  "url": "participation/v1/channels/channel1",
+  "status": "active",
+  "consensusRelation": "consenter",
+  "height": 1151
```
wlahti, line 616: For correctness here, I'd bump the height to 1152 as the config update to add the consenter to the consenters set results in a block on the ledger.

wlahti (changes requested):

docs/source/create_channel/create_channel_participation.md
```diff
@@ -4,19 +4,21 @@ Fabric v2.3 introduces the ability to create channels between network members wi
...
+ * **Consortium no longer required**: You no longer need to define the set of organizations, known as the "consortium", who are permitted to create channels on a particular ordering service. With this new process, **all channels are application channels**, so the concept of a list of organizations who can create channels no longer applies. Any set of organizations can get together and create a channel using a defined set of ordering nodes (which become the ordering service of that channel). Additionally, if the MSP of the peer organization, as listed in the system channel, needs to be changed, the peer organization to coordinate with an orderer organization to change the MSP.
```
wlahti, line 9: `Additionally, if the MSP of the peer organization, as listed in the system channel, needs to be changed, the peer organization to coordinate with an orderer organization to change the MSP.` This reads as halfway in between what I think you were going for. Do you want to highlight that you no longer need to coordinate with an orderer organization here?

docs/source/create_channel/create_channel_participation.md
```diff
@@ -4,19 +4,21 @@ Fabric v2.3 introduces the ability to create channels between network members wi
...
 
 **Benefits of the new process:**
 
- * **Increased Privacy** - A single orderer does not know about all the channels on the ordering service.
- * **Scalability** - When there is a large number of ordering service nodes (orderers) on the system channel, starting the orderer can take a long time as it has to wait for each orderer to replicate the channel ledger.
- * **Operational benefits** - Updating an application channel no longer requires write permissions from a separate ordering organization. Also, you now have a way to list all the channels an orderer is a member of. Lastly, you have the ability to easily remove a channel and its respective resources from an orderer.
+ * **Increased privacy**: Because all ordering nodes used to be joined to the system channel, every ordering node in a network knew about the existence of every channel on that ordering service, even if the node itself wasn't a consenter on that channel and therefore wasn't ordering or store its blocks. Now, an ordering node only knows about the channels it is joined to.
```
wlahti, line 14: `therefore wasn't ordering or storing its blocks` (not sure why the suggestion option isn't appearing here)

docs/source/create_channel/create_channel_participation.md
```diff
@@ -4,19 +4,21 @@ Fabric v2.3 introduces the ability to create channels between network members wi
...
+ * **Scalability**: When there is a large number of ordering nodes on the system channel, starting the node can take a long time as it has to wait for each orderer to replicate the system channel ledger showing it has joined. Additionally, large numbers of nodes on the system channel increases the number of "alive" messages being sent between nodes. In large system channels with large numbers of nodes, this can cause processing issues, even if no application channels has a large number of ordering nodes on it.
+ * **Flexibility**: Whereas ordering nodes could only be part of a single ordering service, orderers can now join or leave channels as needed, similar to how peers can join any channel its organization is a member of.
+ * **Operational benefits**:
+  * It is now easy to list the channels the ordering node is a consenter on.
+  * Is now easy to delete a channel and all of the ledger associated with that channel.
```
wlahti, line 19: Consistency with the bullet above: `It is now easy`

docs/source/create_channel/create_channel_participation.md
```diff
@@ -0,0 +1,527 @@
...
+- **Orderer local MSP:**`organizations/ordererOrganizations/ordererOrg1.example.com/ordering-service-nodes/osn1.ordererOrg1.example.com/msp` folder, also known as the orderer local MSP, contains the enrollment certificate and private key for the ordering service `osn1` node. This folder is automatically generated when you enroll the orderer identity with the organization CA.
+- **TLS certificates:** `organizations/ordererOrganizations/ordererOrg1.example.com/ordering-service-nodes/osn1.ordererOrg1.example.com/tls` folder contains the TLS certificate and private key for the orderer `osn1` node as well as the TLS CA root cert `tls-ca-cert.pem`.  
+
+The `config.yaml` file in the orderer organization MSP and orderer local MSP enables Node OU support for the channel, an important feature that allows the channel to recognize certificates that contain the "admin" OU as an admin identity of the channel. Learn more in the [Fabric CA](https://hyperledger-fabric-ca.readthedocs.io/en/release-1.4/deployguide/use_CA.html#nodeous) documentation.
```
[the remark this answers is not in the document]
pamandrejko: It's config.yaml. I changed this to:

**Note:** You need to create the `config.yaml` file in the orderer organization MSP and orderer local MSP. This file enables Node OU support for the channel, allowing the channel to recognize certificates that contain the "admin" OU as representing organization admins (who are admins of the ordering nodes owned by the organization by default). Learn more in the [Fabric CA](https://hyperledger-fabric-ca.readthedocs.io/en/release-1.4/deployguide/use_CA.html#nodeous) documentation.

docs/source/create_channel/create_channel_participation.md
```diff
@@ -0,0 +1,527 @@
...
+
+#### Configure the orderer.yaml file for each orderer
+
+Follow the instructions in the [ordering service deployment guide](LINK) to build an ordering service with three ordering nodes. Because the system channel is no longer required when you start an ordering service, you can skip the entire process of [generating the genesis block](LINK) in those instructions.  However, when you configure the `orderer.yaml` file for each orderer, there are a few additional modifications you need to make to leverage this feature. You can refer to the orderer [sampleconfig](https://github.com/hyperledger/fabric/blob/{BRANCH}/sampleconfig/orderer.yaml) for more information about these new parameters.
```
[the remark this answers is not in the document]
pamandrejko: Joe kindly addressed this for us.

docs/source/create_channel/create_channel_participation.md
pamandrejko: Done

docs/source/create_channel/create_channel_participation.md
```diff
@@ -0,0 +1,527 @@
...
+
+## Step one: Generate the genesis block of the channel
+
+Channels are created by building a channel creation transaction, also referred to as the "genesis block", and submitting the transaction to the ordering service. The channel creation transaction specifies the initial configuration of the channel and is used by the ordering service to write the channel genesis block. While only one member needs to create the genesis block, it can be shared out of band with the other members on the channel who can inspect it to ensure they agree to the channel configuration and then used by each orderer.
```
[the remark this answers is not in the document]
pamandrejko: Done.

docs/source/create_channel/create_channel_participation.md
pamandrejko: Done

docs/source/create_channel/create_channel_participation.md
pamandrejko: Done.

docs/source/create_channel/create_channel_participation.md
pamandrejko: Done

docs/source/create_channel/create_channel_participation.md
```diff
@@ -0,0 +1,527 @@
...
+
+The  `configtxgen` tool is located in the `bin` folder of downloaded Fabric binaries.
+
+Before using `configtxgen`, confirm you have to set the `FABRIC_CFG_PATH` environment variable to the path of the directory that contains your local copy of the `configtx.yaml` file, for example:
+
+```
+export FABRIC_CFG_PATH=../config
+```
+
+ You can check that you can are able to use the tool by printing the `configtxgen` help text:
```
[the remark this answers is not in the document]
pamandrejko: Done.

docs/source/create_channel/create_channel_participation.md
```diff
@@ -0,0 +1,527 @@
...
+- **Policies** Different sections of the file work together to define the channel policies that will govern how organizations interact with the channel and which organizations need to approve channel updates. For the purposes of this tutorial, we will use the default that are used by Fabric.
+- **Profiles** Each channel profile references information from other sections of the `configtx.yaml` file to build a channel configuration. The profiles are used to create the genesis block of the channel.  
+
+The `configtxgen` tool uses `configtx.yaml` file to create the genesis block for the channel.
+
+You should refer to the [Using configtx.yaml to build a channel configuration](create_channel_config.html) tutorial to learn more about this file. However, the following three sections require specific configuration in order to create a channel without a system channel.
+
+**Note:** Peer organizations are not required when you initially create the channel, but if you know them it is recommended to add them now to the Profiles `Organizations:` and `Applications` section to avoid having to update the channel configuration later.
```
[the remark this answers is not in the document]
pamandrejko: Done.

docs/source/create_channel/create_channel_participation.md
pamandrejko: Changed it to:
INFO 001 Loading configuration
INFO 002 orderer type: etcdraft
INFO 003 Loaded configuration: /fabric/config/configtx.yaml
INFO 004 Generating genesis block
INFO 005 Writing genesis block

docs/source/create_channel/create_channel_participation.md
```diff
@@ -0,0 +1,527 @@
...
+2. In your terminal window set `FABRIC_CFG_PATH` to point to the location of the `orderer.yaml` file, relative to where you are running the fabric commands from. For example, if you download the binaries, and run the commands from the `/bin` directory and the `orderer.yaml` is under `/config`, the path would be:
+    ```
+    export FABRIC_CFG_PATH=../config
+    ```
+3. You can now start the ordering service by running the following command on each ordering node:
+    ```
+    orderer start
+    ```
+
+When the ordering service starts successfully and a Raft leader is elected, you should see something similar to the following output:
```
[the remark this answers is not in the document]
pamandrejko: changed it to:
INFO 01d Registrar initializing without a system channel, number of application channels: 1, with 1 consensus.Chain(s) and 0 follower.Chain(s)
INFO 01e This node is picked to start campaign channel=channel1 node=2
INFO 01f Starting orderer:

docs/source/create_channel/create_channel_participation.md
pamandrejko: I think some words were accidentally dropped in there. I changed it to:
`Additionally, if the MSP of the peer organization, as listed in the system channel, needs to be changed, the peer organization no longer needs to to coordinate with an orderer organization to change the MSP.`

docs/source/create_channel/create_channel_participation.md
pamandrejko: Done

docs/source/create_channel/create_channel_participation.md
```diff
@@ -0,0 +1,527 @@
...
+```
+
+### The configtx.yaml file
+
+The `configtx.yaml` file specifies the **channel configuration** of new channels. The information that is required to build the channel configuration is specified in a readable and editable form in the `configtx.yaml` file. The `configtxgen` tool uses the channel profiles that are defined in the `configtx.yaml` file to create the channel configuration and write it to the [protobuf format](https://developers.google.com/protocol-buffers) that can be read by Fabric.
+
+The `configtx.yaml` file is located in the `config` folder of the binary alongside images that you downloaded. The file contains the following configuration sections that we need to create our new channel:
+
+- **Organizations:** The organizations that can become members of your channel. Each organization has a reference to the cryptographic material that is used to build the [channel MSP](../membership/membership.html). Exactly which members are part of a channel configuration is defined in the **Profiles** section below.
+- **Orderer:** Contains the list of ordering nodes will form the Raft consenter set of the channel.
```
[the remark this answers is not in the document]
pamandrejko: Done.

docs/source/create_channel/create_channel_participation.md
```diff
@@ -0,0 +1,527 @@
...
+- **Policies** Different sections of the file work together to define the channel policies that will govern how organizations interact with the channel and which organizations need to approve channel updates. For the purposes of this tutorial, we will use the default that are used by Fabric.
```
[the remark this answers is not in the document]
pamandrejko: Done.

docs/source/create_channel/create_channel_participation.md
```diff
@@ -0,0 +1,527 @@
...
+          Port: 7092
+          ClientTLSCert: ../config/organizations/ordererOrganizations/ordererOrg1.example.com/ordering-service-nodes/osn3.ordererOrg1.example.com/tls/signcerts/cert.pem
+          ServerTLSCert: ../config/organizations/ordererOrganizations/ordererOrg1.example.com/ordering-service-nodes/osn3.ordererOrg1.example.com/tls/signcerts/cert.pem
+```
+
+When the channel configuration block is created, the configtxgen tool reads the paths to the TLS certificates, and replaces the paths with the corresponding bytes of the certificates.
+
+#### Profiles:
+
+If you are familiar with the legacy process for creating channels, you will recall that the `Profiles:` section of the `configtx.yaml` contained a consortium section under the `Orderer:` group. The consortium definition is no longer required. Rather, under the `Orderer:` section you simply include MSP ID of the ordering organization or organizations in the case of a multi-organization Raft ordering service. And in the `Application:` section, you include the MSP ID of the peer organizations that will be recognized as members of the channel. At least one orderer organization and one peer organization must be provided.
```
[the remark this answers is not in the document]
pamandrejko: Changing to `One orderer organization must be specified and it is recommended to provide at least one peer organization.`

docs/source/create_channel/create_channel_participation.md
```diff
@@ -0,0 +1,527 @@
...
+
+The following snippet is an example of a channel profile that contains an orderer configuration based on the default channel, orderer, organization, and policy configurations. The Application section, where the peer organizations are listed, includes the default Application settings as well as at least one peer organization (SampleOrg) and the corresponding policies for the channel.
```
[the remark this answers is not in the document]
pamandrejko: Done.

docs/source/create_channel/create_channel_participation.md
pamandrejko: Done.

docs/source/create_channel/create_channel_participation.md
```diff
@@ -0,0 +1,544 @@
...
+ * **Consortium no longer required**: You no longer need to define the set of organizations, known as the "consortium", who are permitted to create channels on a particular ordering service. With this new process, **all channels are application channels**, so the concept of a list of organizations who can create channels no longer applies. Any set of organizations can get together and create a channel using a defined set of ordering nodes (which become the ordering service of that channel). Additionally, if the MSP of the peer organization, as listed in the system channel, needs to be changed, the peer organization no longer needs to coordinate with an orderer organization to change the MSP.
```
wlahti: The `all channels are application channels` bit isn't technically accurate, but for the sake of this document it's probably what we want to include/recommend. You can still create a system channel genesis block and use the API to create the system channel (or onboard a new orderer to an existing system channel).

docs/source/create_channel/create_channel_participation.md
```diff
@@ -0,0 +1,544 @@
...
+ * **Flexibility**: Whereas ordering nodes could only be part of a single ordering service, orderers can now join or leave channels as needed, similar to how peers can join any channel its organization is a member of.
+ * **Operational benefits**:
+    * Easy to list the channels that the ordering node is a consenter on.
+    * Simple process to delete a channel and the ledger associated with that channel.
+    * Ordering nodes now detect that they have been added to a channel more quickly. Previously, this process could take several minutes, causing ordering node admins to restart their nodes so that the node could detect that it was joined more quickly.
+
+Whereas in the past an orderer could only be part of a single ordering service, now orderers can join (or leave) any number of channels as needed, similar to how peers can participate in multiple channels.
+
+**Note:** The osnadmin `join channel` and `remove channel` commands only work with Raft ordering nodes, they cannot be used with a Solo or Kafka ordering service.
```
wlahti: Do we want to reference these in the ordering that they're specified via the CLI? `channel join` and `channel remove` if so.

docs/source/create_channel/create_channel_participation.md
```diff
@@ -0,0 +1,544 @@
...
+
+Ensure that the orderer organization has write privileges in the channel configuration so the orderer can become a consenter. When the channel update transaction is complete, you can use the `osnadmin channel list` command again to confirm that the `consensusRelation` status for the orderer automatically changes to **consenter**.
+
+```
+{
+  "name": "channel1",
+  "url": "participation/v1/channels/channel1",
+  "status": "active",
+  "consensusRelation": "consenter",
+  "height": 1152
```
wlahti: For accuracy, this would be 1153 now after the config update to add the orderer to the consenters set. :)

docs/source/create_channel/create_channel_participation.md
```diff
@@ -0,0 +1,544 @@
...
+    ```
+    export TLS_CA_ROOT_CERT=../config/organizations/ordererOrganizations/ordererOrg1.example.com/ordering-service-nodes/osn1.ordererOrg1.example.com/tls/cacerts/tls-ca-cert.pem
+    export OSN_TLS_SIGN_CERT=../config/organizations/ordererOrganizations/ordererOrg1.example.com/ordering-service-nodes/osn1.ordererOrg1.example.com/tls/signcerts/cert.pem
+    export OSN_TLS_PRIVATE_KEY=../config/organizations/ordererOrganizations/ordererOrg1.example.com/ordering-service-nodes/osn1.ordererOrg1.example.com/tls/keystore/key.pem
+
+    osnadmin channel list -o [ORDERER_ADMIN_LISTENADDRESS] --ca-file $TLS_CA_ROOT_CERT --client-cert $OSN_TLS_SIGN_CERT --client-key $OSN_TLS_PRIVATE_KEY
+    ```
```
tock-ibm, line 166: I am confused by the naming of these variables. 
This `--ca-file $TLS_CA_ROOT_CERT` is the **server's** ROOT CA, correct? 
However, these two are the certificate and private key of the **admin client**, right?
`--client-cert $OSN_TLS_SIGN_CERT --client-key $OSN_TLS_PRIVATE_KEY`

So I think the naming should be revised - 
TLS_CA_ROOT_CERT -> OSN_TLS_CA_ROOT_CERT
OSN_TLS_SIGN_CERT -> ADMIN_TLS_SIGN_CERT
OSN_TLS_PRIVATE_KEY -> ADMIN_TLS_PRIVATE_KEY 

In addition, the folder structure above may be revised to clearly reflect the division between "server" crypto material and "client" crypto material. 

Is the tutorial using shared material for simplicity?

Note that the same client can be used to command all three OSNs!

docs/source/create_channel/create_channel_participation.md
```diff
@@ -0,0 +1,544 @@
...
+- `Admin.ListenAddress` - The orderer admin server address (host and port) that can be used by the `osnadmin` command to configure channels on the ordering service. This value should be a unique `host:port` combination to avoid conflicts.
+- `Admin.TLS.Enabled:` - This value must be set to `true`.
+- `Admin.TLS.PrivateKey:` - The path to and file name of the orderer private key issued by the TLS CA.
+- `Admin.TLS.Certificate:` - The path to and file name of the orderer signed certificate issued by the TLS CA.
+- `Admin.TLS.RootCAs:`  - The path to and file name of the TLS CA Root cert `tls-ca-cert.pem`.
+- `Admin.TLS.ClientAuthRequired:` This value must be set to `true`. Note that while Mutual TLS is required for all operations on the orderer Admin endpoint, the entire network is not required to use Mutual TLS.
+- `Admin.TLS.ClientRootCAs:` - The path to and file name of the TLS CA Root cert `tls-ca-cert.pem`.
```
tock-ibm: We are using a same ROOT CA for server and clients here, which is not necessarily the case in practice.
I think using `client-tls-ca-cert.pem` will be clearer.

docs/source/create_channel/create_channel_participation.md
```diff
@@ -0,0 +1,544 @@
...
+osnadmin channel join –-channel-id [CHANNEL_NAME]  --config-block [CHANNEL_CONFIG_BLOCK] -o [ORDERER_ADMIN_LISTENADDRESS] --ca-file $TLS_CA_ROOT_CERT --client-cert $OSN_TLS_SIGN_CERT --client-key $OSN_TLS_PRIVATE_KEY
+```
+
+Replace:
+- `CHANNEL_NAME` with the name you want to call this channel.
+- `CHANNEL_CONFIG_BLOCK` with the path and file name of the genesis block you created in Step one if you are creating the channel. Each subsequent ordering node can join the configuration starting with the genesis block, or they can join by providing the latest config block instead.
+- `ORDERER_ADMIN_LISTENADDRESS` corresponds to the `Orderer.Admin.ListenAddress` defined in the `orderer.yaml` for this orderer.
+- `TLS_CA_ROOT_CERT` with the path and file name of the orderer organization TLS CA root certificate and intermediate certificate if using an intermediate TLS CA.
+- `OSN_TLS_SIGN_CERT` with the path and file name of the orderer signed certificate from the TLS CA.
+- `OSN_TLS_PRIVATE_KEY` with the path and file name of the orderer private key from the TLS CA.
```
tock-ibm: This should be the admin client's certificate and private key, not the server's.

docs/source/create_channel/create_channel_participation.md
```diff
@@ -0,0 +1,544 @@
...
+
+For example:
+```
+osnadmin channel join –-channel-id channel1 --config-block genesis_block.pb -o OSN1.example.com:7050 --ca-file $TLS_CA_ROOT_CERT --client-cert $OSN_TLS_SIGN_CERT --client-key $OSN_TLS_PRIVATE_KEY
+```
+
+**Note:** The connection between the `osnadmin` CLI and the orderer requires mutual TLS, although it is not required for the network itself. This means you need to pass the `--client-cert` and `--client-key` parameters on each `osadmin` command. The `--client-cert` parameter points to the orderer certificate issued by the TLS CA, and `--client-key` refers to the orderer private key issued by the TLS CA.
```
tock-ibm: `client-cert` parameter points to the client's certificate issued by the CLIENT TLS CA, and `--client-key` refers to the client's private key.`

docs/source/create_channel/create_channel_participation.md
```diff
@@ -0,0 +1,544 @@
...
+  "url": "participation/v1/channels/channel1",
+  "consensusRelation": "consenter",
+  "status": "active",
+  "height": 1
+}
+```
+
+When successful, the channel is started with a ledger height `1`. Because this orderer is joining from the genesis block, the `status` is "active". And because the orderer is part of the channel's consenter set, its `consensusRelation` is "consenter". We delve more into the `consensusRelation` value in the next section.
+
+You can repeat the same command on the other two orderers. Remember to update the `OSN_TLS_SIGN_CERT` and `OSN_TLS_PRIVATE_KEY` environment variables to point to the certificates for the corresponding orderer. Because these two orderers join the channel from the genesis block and are part of the consenter set, their status transitions almost immediately from  "onboarding" to "active" and the "consensusRelation" status is "consenter". At this point, the channel becomes operational because a majority of consenters (as defined in the channel configuration) are "active". You should see something similar to the following in your orderer logs:
```
tock-ibm: Again, these (the `OSN_TLS_SIGN_CERT` and `OSN_TLS_PRIVATE_KEY`) are actually the client's certificate and key. They can stay the same.

If all 3 orderers come from the same org and have the same ROOT CA, then the only thing that needs to change it the orderer endpoint: `-o OSN1.example.com:7050` -> `-o OSN2.example.com:7050`

docs/source/create_channel/create_channel_participation.md
```diff
@@ -0,0 +1,544 @@
...
+
+```
+INFO 087 raft.node: 1 elected leader 1 at term 2 channel=channel1 node=1
+INFO 088 Raft leader changed: 0 -> 1 channel=channel1 node=1
+INFO 089 Start accepting requests as Raft leader at block [0] channel=channel1 node=1
+```
+
+After the channel is started by the first orderer, subsequent nodes can join from either the genesis block or from the latest config block. When an orderer joins from a config block, its status is always "onboarding" while its ledger catches up to the config block that was specified in the join command, after which the status is automatically updated to "active".
+
+Use the `osnadmin channel list` command with the `--channel-id` flag to view the detailed `status` and `consensusRelation` of any ordering node:
```
tock-ibm, line 520: `... of any **channel** on any ordering node:`

docs/source/create_channel/create_channel_participation.md
```diff
@@ -0,0 +1,544 @@
...
+Otherwise, if joining from the latest config block, the `status` is "onboarding". In that case, after the channel ledger has caught up to the specified config block, you can issue the same `osnadmin channel list` command to confirm the status changes to **active** and the ledger height has caught up. For example, if the total number of blocks on the channel ledger at the specified config block was `1151`, after the ledger catches up, the output would look similar to:
+
+```
+{
+  "name": "channel1",
+  "url": "participation/v1/channels/channel1",
+  "status": "active",
+  "consensusRelation": "follower",
+  "height": 1152
+}
```
tock-ibm: 1151, actually; as the follower pulls the config-block that corresponds to the join-block given to it, it changes to "active".

However, it is important to note that the transition from "onboarding" to "active" is not the only criteria used to decide whether it is the right time to add the new OSN to the channel's consenters set. What it more important is the `height` that the new OSN reached, compared to the height of the consenters. For example, the last config block may be at height 1151, but there might be a huge number of normal block after it.... After OSN4 becomes "active" (caught up to the last config), it is better to compare the height of OSN1 with that of OSN4, and if they are "close enough" - only then add OSN4 to the consenters set of the channel.

docs/source/create_channel/create_channel_participation.md
pamandrejko: This is an important point - thank you for mentioning it because I had not really considered it. My thinking though is that since this is a tutorial about creating a channel without a system channel, it doesn't really fit in this topic per se and could be confusing. I think we should def include this as appropriate when we update all the other topics that talk about the system channel though.

docs/source/create_channel/create_channel_participation.md
pamandrejko: We absolutely do. Good catch. Fixed.

docs/source/create_channel/create_channel_participation.md
pamandrejko: Had not even considered this. Yes. good catch. Done.

tock-ibm: For completeness: If the config block used for join was at 1151, and there were, say, 1000 normal blocks after it, then the config block which added OSN4 would be at 2152,

docs/source/create_channel/create_channel_participation.md
```diff
@@ -0,0 +1,544 @@
...
+
+## Next steps
+
+### Join peers to the channel
+
+After the channel has been created, you can follow the normal process to join peers to the channel and configure anchor peers.  If the peer organization was not originally included in the channel configuration, you need to submit a [channel configuration update transaction](../config_update.html) to add the organization. If it did include the peer organization, then all peer organizations that are members of the channel can fetch the channel genesis block from the ordering service using the [peer channel fetch](../commands/peerchannel.html#peer-channel-fetch) command. The organization can then use the genesis block to join the peer to the channel using the [peer channel join](../commands/peerchannel.html#peer-channel-join) command. Once the peer is joined to the channel, the peer will build the blockchain ledger by retrieving the other blocks on the channel from the ordering service.
+
+### Add or remove orderers from existing channels
+
+You can continue to use the `osnadmin channel join` and `osnadmin channel remove` commands to add and remove orderers on each channel according to your business needs.
```
tock-ibm: I think it is important to mention that it is recommended to configure channel `A` and remove `OSNx` from the consenters set, prior to deleting channel `A` from orderer `OSNx`.

docs/source/create_channel/create_channel_participation.md
```diff
@@ -0,0 +1,550 @@
...
+
+Certificate names used in this example are for illustration purposes and may not reflect the actual names of the certificates generated by the CA. When you generate the certificates you can rename them accordingly to make it easier to differentiate them.
+
+**Important:** You need to create the `config.yaml` file in the orderer organization MSP and orderer local MSP. This file enables "Node OU" support for the channel, a feature that allows the channel to recognize certificates that contain the "admin" OU as representing organization admins (who are admins of the ordering nodes owned by the organization by default). Learn more in the [Fabric CA](https://hyperledger-fabric-ca.readthedocs.io/en/release-1.4/deployguide/use_CA.html#nodeous) documentation.
```
denyeart: "This file enables "Node OU" support for the channel"

I'm forgetting right now... is this Node OU support for both the organization msp on the channel, and for the local msp admin?
If so, what is the difference between what a local msp admin can do on an OSN (as authorized by the "admin" node OU), versus what a osnadmin client can do on an OSN (as authorized by the channel participation API mutual TLS)?

docs/source/create_channel/create_channel_participation.md
```diff
@@ -0,0 +1,550 @@
...
+
+If you are using a containerized solution for running your network (which for obvious reasons is a popular choice), **it is a best practice to mount these folders (also known as "volumes") external to the container where the node itself is running. This will allow the certificates to be used to create a new node should the node container go down, become corrupted, or be restarted.**
+
+## Prerequisites
+
+Because `osnadmin` commands are performed against an ordering service node (which executes the actions), the ordering service needs to exist before you can create a channel. You can attempt this tutorial with your existing ordering service or deploy a new ordering service. If you decide to use `osnadmin` commands with an existing ordering service, the system channel must first be removed from each ordering node before you can create any new channels. Choose whether you want to use your existing ordering service or deploy a new one:
```
denyeart: "the ordering service needs to exist before you can create a channel"

I see what you mean, but the phrasing here seems misleading. I think of an ordering service as something that is already consenting on channels (either system channel or application channels). While the focus of this tutorial is spinning up OSNs without any channels and then creating/joining an application channel. It seems like once the tutorial is _done_ and you've got three OSNs consenting on a channel, you'll have an ordering service at the end. If you agree with this comment, subsequent references to "ordering service" would also need to be updated.

docs/source/create_channel/create_channel_participation.md
```diff
@@ -0,0 +1,550 @@
...
+- [Use existing ordering service](#use-existing-ordering-service)
+- [Deploy a new ordering service](#deploy-a-new-ordering-service)
+
+### Use existing ordering service
+
+Before you can take advantage of this feature on a deployed ordering service, you need to remove the system channel from each ordering node that is a consenter in your application channels. A "mixed mode" of orderers on a channel, where some nodes are part of a system channel and others are not, is not supported. The `osadmin` CLI includes both a `channel list` and a `channel remove` command to facilitate the process of removing the system channel. If you prefer to deploy a new ordering service instead, skip ahead to [Deploy a new ordering service](#deploy-a-new-ordering-service).
+
+#### Remove the system channel
+
+- Before attempting these steps ensure that you have [upgraded](../upgrade.html) your Fabric network to v2.3 or higher.
```
denyeart, line 135:
```suggestion
- Before attempting these steps ensure that you have [upgraded](../upgrade.html) your Fabric ordering service nodes to v2.3 or higher.
```

docs/source/create_channel/create_channel_participation.md
```diff
@@ -0,0 +1,550 @@
...
+2. In your terminal window set the `FABRIC_CFG_PATH` to point to the location of the `orderer.yaml` file, relative to where you are running the Fabric commands from. For example, if you download the binaries, and run the commands from the `/bin` directory and the `orderer.yaml` is under `/config`, the path would be:
+    ```
+    export FABRIC_CFG_PATH=../config
+    ```
+3. You can now start the ordering service by running the following command on each ordering node:
+    ```
+    orderer start
+    ```
+
+When the ordering service starts successfully, you should see something similar to the following output:
```
denyeart: Again, I wouldn't think of this as an "ordering service" starting yet. At this point it is just three independent OSNs starting and each displaying the same output.

docs/source/create_channel/create_channel_participation.md
```diff
@@ -0,0 +1,550 @@
...
+INFO 01f Starting orderer:
+```
+This action starts the ordering nodes without any channels. Repeat these steps for each orderer. At this point, the three nodes are not communicating with each other until we create a channel in the subsequent steps.
+
+### Define your peer organizations
+
+Because the channel you are creating is meant to be used by two or more peer organizations to transact privately on the network, you need to have at least one peer organization defined to act as the channel administrator who can add other organizations. Technically, the peer nodes themselves do not yet have to be deployed, but you do need to create one or more peer organization MSP definitions and at least one peer organization needs to be provided in the `configtx.yaml` in the next step. Before proceeding to the next section, follow the steps in the [Fabric CA](https://hyperledger-fabric-ca.readthedocs.io/en/release-1.4/deployguide/use_CA.html#create-the-org-msp-needed-to-add-an-org-to-a-channel) documentation to build your peer organization MSP definition. If the peers have been deployed, you should also include their address in the `AnchorPeers:` section.
```
denyeart, line 314: "If the peers have been deployed, you should also include their address in the `AnchorPeers:` section."

Is this still true? I thought since configtxgen flag --outputAnchorPeersUpdate is deprecated, that we wouldn't recommend this anymore. May need @jyellick opinion.

docs/source/create_channel/create_channel_participation.md
```diff
@@ -0,0 +1,550 @@
...
+
+```
+configtxgen --help
+```
+
+### The `configtx.yaml` file
+
+The `configtx.yaml` file specifies the **channel configuration** of new channels in a format that is both readable and editable and that can be used by `configtxgen`. The `configtxgen` tool uses the channel profiles that are defined in `configtx.yaml` to create the channel configuration and write it to the [protobuf format](https://developers.google.com/protocol-buffers) that can be consumed by Fabric.
+
+The `configtx.yaml` file is located in the `config` folder of the binary alongside images that you downloaded. The file contains the following configuration sections that we need to create our new channel:
```
denyeart:
```suggestion
A sample `configtx.yaml` file is located in the `config` folder of the binary alongside images that you downloaded. The file contains the following configuration sections that we need to create our new channel:
```

docs/source/create_channel/create_channel_participation.md
```diff
@@ -0,0 +1,550 @@
...
+        - Host: Host2
+          Port: 7091
+          ClientTLSCert: ../config/organizations/ordererOrganizations/ordererOrg1.example.com/ordering-service-nodes/osn2.ordererOrg1.example.com/tls/signcerts/tls-cert.pem
+          ServerTLSCert: ../config/organizations/ordererOrganizations/ordererOrg1.example.com/ordering-service-nodes/osn2.ordererOrg1.example.com/tls/signcerts/tls-cert.pem
+        - Host: Host3
+          Port: 7092
+          ClientTLSCert: ../config/organizations/ordererOrganizations/ordererOrg1.example.com/ordering-service-nodes/osn3.ordererOrg1.example.com/tls/signcerts/tls-cert.pem
+          ServerTLSCert: ../config/organizations/ordererOrganizations/ordererOrg1.example.com/ordering-service-nodes/osn3.ordererOrg1.example.com/tls/signcerts/tls-cert.pem
+```
+
```
denyeart, line 394: Mention that for simplicity in this example, that each orderer uses the same TLS cert for both server and client, although this need not be the case.

docs/source/create_channel/create_channel_participation.md
```diff
@@ -0,0 +1,550 @@
...
+                - <<: *SampleOrg
+                  Policies:
+                      <<: *SampleOrgPolicies
+                      Admins:
+                          Type: Signature
+                          Rule: "OR('SampleOrg.member')"
+
+```
+
+**Note:** For simplicity, this snippet assumes that the peers and orderers belong to the same organization `SampleOrg`. You can refer to the [sample config](https://github.com/hyperledger/fabric/blob/{BRANCH}/sampleconfig/configtx.yaml) for the full example. For a production deployment however, it is recommended that the peer and ordering nodes belong to separate organizations.
```
denyeart: This tutorial switches back and forth between ordererOrg1.example.com and SampleOrg. This will likely confuse readers.

docs/source/create_channel/create_channel_participation.md
```diff
@@ -0,0 +1,550 @@
...
+```
+INFO 001 Loading configuration
+INFO 002 orderer type: etcdraft
+INFO 003 Loaded configuration: /fabric/config/configtx.yaml
+INFO 004 Generating genesis block
+INFO 005 Writing genesis block
+
+```
+
+## Step two: Use the `osnadmin` CLI to start the channel
```
denyeart: Do we really want to use the verb "start"? Maybe that decision was already made...

I think of it as "adding the first OSN to a channel" rather than "starting" the channel. Although, it is described better in the next paragraph.

denyeart (commented):
Nice job folks!

Mostly nits...

docs/source/create_channel/create_channel_participation.md
tock-ibm: You don't really need to do this, it is recommended. If the channel has fault tolerance, removing a channel from an orderer is equivalent to a single failure.

docs/source/create_channel/create_channel_participation.md
joealewine: Orderer, ordering node, ordering service, ordering service node are all terms that have been used more or less interchangeably. I am partial "ordering node" (which I think was a Gari suggestion) and don't care much for "ordering service node". Fwiw, there's also "consenter", but that seems to be a term we've strictly used just for nodes that are consenting on an active channel.

I'm actually sort of fine with the grab bag of terms. We've been using all of them for long enough that our users should be used to it. But if we want to tighten up our references and more strictly define these terms going forward, we can. We will just need broader agreement, which will necessitate a discussion.

docs/source/create_channel/create_channel_participation.md
```diff
@@ -0,0 +1,550 @@
...
+ * **Consortium no longer required**: You no longer need to define the set of organizations, known as the "consortium", who are permitted to create channels on a particular ordering service. With this new process, **all channels are application channels**, so the concept of a list of organizations who can create channels no longer applies. Any set of organizations can get together and create a channel using a defined set of ordering nodes (which become the ordering service of that channel).
+ * **Simplified ordering node creation process**: Because a genesis block no longer needs to exist before an ordering node is created, admins can now focus on the process of setting up their infrastructure and making sure their node is working correctly before joining a particular application channel. Additionally, because an ordering node no longer needs to scan the system channel ledger to determine which channels it is joined to and then replicate those ledgers, the node starts up more rapidly.
```
jyellick: > Additionally, because an ordering node no longer needs to scan the system channel ledger to determine which channels it is joined to and then replicate those ledgers, the node starts up more rapidly.

This isn't really true.  The only time the orderer scans the system channel for channels is when bootstrapping from a late join config block.  Otherwise, the existing ledgers are simply loaded.  Where did this sentence/claim originate from?

docs/source/create_channel/create_channel_participation.md
```diff
@@ -0,0 +1,550 @@
...
+
+**Benefits of the new process:**
+
+ * **Increased privacy**: Because all ordering nodes used to be joined to the system channel, every ordering node in a network knew about the existence of every channel on that ordering service, even if the node itself wasn't a consenter on that channel and therefore wasn't ordering or storing its blocks. Now, an ordering node only knows about the channels it is joined to.
+ * **Scalability**: When there is a large number of ordering nodes on the system channel, starting the node can take a long time as it has to wait for each orderer to replicate the system channel ledger showing it has joined. Additionally, large numbers of nodes on the system channel increases the number of "alive" messages being sent between nodes. In large system channels with a large number of nodes, this can cause processing issues, even if no application channels have a large number of ordering nodes on it.
```
jyellick: Why `"alive"` instead of `"heartbeat"`?  I don't particularly have a problem with it, just seems odd to invent new terminology.

docs/source/create_channel/create_channel_participation.md
```diff
@@ -0,0 +1,550 @@
...
+ * **Flexibility**: Whereas ordering nodes could only be part of a single ordering service, orderers can now join or leave channels as needed, similar to how peers can join any channel its organization is a member of.
+ * **Operational benefits**:
+    * Easy to list the channels that the ordering node is a consenter on.
+    * Simple process to delete a channel and the ledger associated with that channel.
+    * Ordering nodes now detect that they have been added to a channel more quickly. Previously, this process could take several minutes, causing ordering node admins to restart their nodes so that the node could detect that it was joined more quickly.
```
jyellick: > Ordering nodes now detect that they have been added to a channel more quickly. 

I guess this is in the section about benefits of the new process, but, my first reading implied that this was an overall improvement.  Maybe it would be better to say something like: "Orderer nodes can now track channels as a follower before being added to a consenter, allowing them to detect this addition more quickly."

docs/source/create_channel/create_channel_participation.md
```diff
@@ -0,0 +1,550 @@
...
+    * If the MSP of a peer organization, as listed in the system channel, needs to be changed, the peer organization no longer needs to coordinate with an orderer organization to change the MSP.
+
+**Note: the new mode of creating channels is incompatible with creating a channel using a system channel. If the system channel exists, the `Join` and `Remove` operations will not be supported. Similarly, they cannot be used with a Solo or Kafka ordering service.**
```
jyellick: Maybe it's too detailed here, but `Remove` works, but only for the system channel.

docs/source/create_channel/create_channel_participation.md
```diff
@@ -0,0 +1,550 @@
...
+
+Note: “Mixed mode” management, where the system channel is to create channels on some ordering nodes and the new process is used to create channels on other ordering nodes is not supported and highly discouraged. You must either transition to the new process or continue to use the system channel process. For information about removing the system channel from an existing ordering service as part of transitioning to the new process, check out [Remove the system channel](#remove-the-system-channel).
+
+In the process of creating the channel, this tutorial will take you through the following steps and concepts:
+
+- [Prerequisites](#prerequisites)
+- [Step one: Generate the genesis block of the channel](#step-one-generate-the-genesis-block-of-the-channel)
+- [Step two: Use the osnadmin CLI to create the channel](#step-two-use-the-osnadmin-cli-to-create-the-channel)
```
jyellick: I suppose this could devolve into a philosophical debate, but I would argue the channel is created when the genesis block is created.  The channel isn't very useful though until the orderer joins and begins to participate in consensus for that channel.

I'm also a bit confused why joining the first node is somehow a special step, while joining additional nodes is a separate step.  The channel will not be usable until enough nodes to form quorum have been joined.

docs/source/create_channel/create_channel_participation.md
```diff
@@ -0,0 +1,550 @@
...
+
+#### Remove the system channel
+
+- Before attempting these steps ensure that you have [upgraded](../upgrade.html) your Fabric network to v2.3 or higher.
+- Modify the `orderer.yaml` for each ordering node to support this feature and restart the node. See the orderer [sampleconfig](https://github.com/hyperledger/fabric/blob/{BRANCH}/sampleconfig/orderer.yaml) for more information about these required parameters.
+    - `General.BootstrapMethod` - Set this value to `none`. Because the system channel is no longer required, the `orderer.yaml` file on each orderer needs to be configured with `BootstrapMethod: none` which means that no bootstrap block is required or used to start up the ordering service.
+    - `Admin.ListenAddress` - The orderer admin server address (host and port) that can be used by the `osnadmin` command to configure channels on the ordering service. This value should be a unique `host:port` combination to avoid conflicts.
+    - `Admin.TLS.Enabled:` - This value must be set to `true`.
```
jyellick: I'm not sure whether we want to make this less clear, but Will did add support for this being false.  However, it's a very bad idea to set it to false in all but toy networks.

docs/source/create_channel/create_channel_participation.md
```diff
@@ -0,0 +1,550 @@
...
+    - `Admin.TLS.PrivateKey:` - The path to and file name of the orderer private key issued by the TLS CA.
+    - `Admin.TLS.Certificate:` - The path to and file name of the orderer signed certificate issued by the TLS CA.
+    - `Admin.TLS.RootCAs:`  - The path to and file name of the orderer TLS CA Root certificate.
```
jyellick: This variable is not useful, I would not document it (and we can/should remove it from the config sample).

docs/source/create_channel/create_channel_participation.md
```diff
@@ -0,0 +1,550 @@
...
+    - `ChannelParticipation.Enabled:` - Set this value to `true` to enable this feature on the orderer.
+- Restart each ordering node.
+- Put the [system channel into maintenance mode](../kafka_raft_migration.html#entry-to-maintenance-mode) using the same process for the Kafka to Raft migration. This action stops new channel creation transactions from coming in.
+- Remove the system channel from the set of orderers, one by one. During the system channel removal from an orderer, the orderer stops servicing all other application channels it is a member of until removal is complete. Therefore, care must be taken to ensure that the remaining orderers can still function and reach consensus. This operation can be staggered, such that depending on the fault tolerance setup of the respective channels, no channel down time is experienced. If an application channel can tolerate one server offline, you should still be able to submit transactions to the channel, via the other orderers that are not undergoing the removal at that time.  Use the `osnadmin channel list` command to view the channels that this orderer is a member of:
```
jyellick: > During the system channel removal from an orderer, the orderer stops servicing all other application channels it is a member of until removal is complete.

Is this true? @stephyee @tock-ibm

docs/source/create_channel/create_channel_participation.md
```diff
@@ -0,0 +1,550 @@
...
+
+Because this tutorial demonstrates the process for creating a channel with **three orderers** deployed for a single organization, you need to generate enrollment and TLS certificates for each node. Why three orderers? This configuration allows for a policy of `MAJORITY` quorum on the Raft cluster. Namely, when there are three orderers, one at a time can go down for maintenance, while a `MAJORITY` (two of three) is maintained. For more information about the number of nodes you should deploy in production, check out [The Ordering Service](../orderer/ordering_service.html#raft). For simplicity and learning purposes, you have the ability to deploy a single node ordering service, even though such an ordering service will not be highly available and cannot therefore be considered a "production" deployment.
```
jyellick, line 251: > Why three orderers? This configuration allows for a policy of `MAJORITY` quorum on the Raft cluster.

This seems to conflate channel config policies and the Raft thesis.  Raft quorum is always majority, regardless of any policy or configuration, this is simply how quorums work.

docs/source/create_channel/create_channel_participation.md
```diff
@@ -0,0 +1,550 @@
...
+
+- `General.BootstrapMethod` - Set this value to `none`. Because the system channel is no longer required, the `orderer.yaml` file on each orderer needs to be configured with `BootstrapMethod: none` which means that no bootstrap block is required or used to start up the ordering service.
+- `Admin.ListenAddress` - The orderer admin server address (host and port) that can be used by the `osnadmin` command to configure channels on the ordering service. This value should be a unique `host:port` combination to avoid conflicts.
+- `Admin.TLS.Enabled:` - This value must be set to `true`.
+- `Admin.TLS.PrivateKey:` - The path to and file name of the orderer private key issued by the TLS CA.
+- `Admin.TLS.Certificate:` - The path to and file name of the orderer signed certificate issued by the TLS CA.
+- `Admin.TLS.RootCAs:`  - The path to and file name of the orderer TLS CA Root certificate.
```
jyellick: Once more, this value doesn't really make sense for the admin section.

docs/source/create_channel/create_channel_participation.md
jyellick: I also made this comment at the anchor, but at the top of the document, this is referred to as 'creating' the channel, which is less correct than 'start' to me.  But I agree with Dave... this is simply joining the first node.

docs/source/create_channel/create_channel_participation.md
```diff
@@ -0,0 +1,550 @@
...
+INFO 002 orderer type: etcdraft
+INFO 003 Loaded configuration: /fabric/config/configtx.yaml
+INFO 004 Generating genesis block
+INFO 005 Writing genesis block
+
+```
+
+## Step two: Use the `osnadmin` CLI to start the channel
+
+Now that the genesis block has been created, the first ordering node that receives the `osnadmin channel join` command command effectively "activates" the channel, though the channel is not fully operational until a quorum of consenters is established (if your profile listed three consenters, at least one more node, for a total of two, would have to join using the `osnadmin channel join` command).
```
jyellick: I'm really not sure what "activates" is supposed to mean.  If you subsequently remove the channel from the first orderer, is it still activated? From a peer perspective, how does is an "activated" channel different from one that is not?

I guess the question really is "If a channel genesis block is created, but nobody joins it, does the channel exist?"  I would argue that it does.

docs/source/create_channel/create_channel_participation.md
```diff
@@ -0,0 +1,550 @@
...
+  "height": 1
+}
+```
+
+The following diagram summarizes the steps you have completed:
+
+![create-channel.osnadmin1-3](./osnadmin1-3.png)
+_You used the configtxgen command to create the channel genesis block and provided that file when you ran the osnadmin channel join command for each orderer by targeting the admin endpoint on each node._
+
+Assuming you have successfully run the `osnadmin channel join` on all three ordering nodes, you now have an active channel and the ordering service is ready to order blocks. Peers can join the channel and begin to transact.
```
jyellick: > ready to order blocks

I would say "is ready to order transactions into blocks"

> Peers can join the channel and begin to transact

I would say "Peers can join the channel and clients can begin to transact"

docs/source/create_channel/create_channel_participation.md
```diff
@@ -0,0 +1,550 @@
...
+
+Anytime a new orderer joins the cluster, you need to ensure that you do not lose quorum while the ledger on the new orderer catches up. Adding a fourth orderer to an existing three node cluster changes the majority from two to three. Therefore, during the catch-up process and until the new orderer starts executing consensus, the cluster is in a state of reduced fault tolerance. To account for this situation and avoid any downtime, a command that can ascertain the `consensusRelation` status of an orderer, which can be either "consenter" or "follower", has been introduced. When a node joins as a follower, the channel ledger is replicated from other orderers, but as it does not take part in consensus, channel operations are not impacted. Because replicating a long chain of blocks could take a long time, joining as a follower is useful for channels with high block heights. To join an orderer to a channel as a follower, **do not include the node in the channel configuration consenter set**.
```
jyellick: > Because replicating a long chain of blocks could take a long time, joining as a follower is useful for channels with high block heights.

I would say "is most useful".  But it's generically useful because it allows you to confirm that an OSN is able to replicate blocks before having it count towards quorum.  (You can imagine for instance, a network firewall rule which prevents the orderer from contacting the others being detectable in this way).

jyellick (commented):
Thanks for all the hard work on this, it's looking good.  A few comments though.

docs/source/create_channel/create_channel_participation.md
jyellick: `--outputAnchorPeersUpdate` is deprecated because it's not possible to reliably compute a safe and correct update without knowing the original configuration.

There is nothing strictly wrong with specifying anchor peers in a genesis block, because there is no delta, and no lost safety/correctness.

That being said, it's always struck me as odd/wrong from a workflow perspective to have anchor peers in the genesis block.  This is because inherently, the peer cannot be joined to the channel before the block is created, and, the anchor peer is incorrect until the peer has been joined.  That being said, for the purposes of configuring a working network (with gossip) in as few steps as possible, I think it's probably fine.

docs/source/create_channel/create_channel_participation.md
pamandrejko: Done.

docs/source/create_channel/create_channel_participation.md
joealewine: I added this --- I was trying to avoid the philosophical question about when a channel is really "created". Is it when the genesis block is created or when nodes are joined to it or is it when enough nodes are joined to it. This is basically saying that, yes, a genesis block has been created, but the channel isn't really "active" until nodes are joined to it. Sort of splitting the difference.

Anyway, I think if we come to broader agreement about what terms to use in these situations that will be clear up what terminology should be used here and elsewhere.

docs/source/create_channel/create_channel_participation.md
pamandrejko: Good point. I changed it to
Because replicating a long chain of blocks could take a long time, joining as a follower is most useful for channels with high block heights because it allows you to confirm that an orderer is able to replicate blocks before having it count towards quorum.

docs/source/create_channel/create_channel_participation.md
pamandrejko: We decided to include 2 steps here because 
Step two is about  *consenters* and step three is about *followers*.
It felt like a logical way to break up all the content. 

I do see your point though. I changed it to 
Step two: Use the `osnadmin` CLI to add the first orderer to the channel
although in that step we join all 3 orderers so I'm thinking of simplifying it to 
Step two: Use the `osnadmin` CLI to join the orderers to the channel
But then 
Step three: Join additional ordering nodes
is confusing. I agree. I need to think on this a little more. I was trying to avoid using the terms consenter/ follower since is has not been introduced yet in the topic.

docs/source/create_channel/create_channel_participation.md
joealewine: The idea that orderers do not have to scan the ordering service for the channels anymore (and that this is an improvement over the old way) it is in the RFC under "scalability problems". It was my inference that this would help nodes start faster. 

Should this whole sentence be deleted?

docs/source/create_channel/create_channel_participation.md
pamandrejko: Yes. The Node OU support is for both the organization msp on the channel, and for the local msp admin.
My understanding is that an ordering node admin (local MSP) really has no special privileges at all -- unlike a peer admin which is required to install a smart contract on a peer.

But we know that  osnadmin client can join an orderer to a channel and remove it from the channel and list the channels that the orderer is a member of.

@denyeart Not sure if I've addressed your comment here? Perhaps there is something more you need me to clarify in the doc?

docs/source/create_channel/create_channel_participation.md
pamandrejko: I went through and carefully assessed each occurrence of ordering service and edited it to orderer if it was in the context of bringing up OSNs i.e. before all three are up.

docs/source/create_channel/create_channel_participation.md
pamandrejko: Given he said it not really true. I removed it.

docs/source/create_channel/create_channel_participation.md
pamandrejko: OK. I'll leave it in for now.

docs/source/create_channel/create_channel_participation.md
pamandrejko: Welcome better suggestions there.

pamandrejko (commented):
I also added a section about the osnadmin CLI command in the Using configtx.yaml to build a channel configuration topic.

docs/source/create_channel/create_channel_participation.md
denyeart: @pamandrejko I added clarifications to this paragraph in the orderer deployment guide:
https://github.com/hyperledger/fabric/pull/2080#discussion_r523103698

The discussion about osnadmin authorization versus orderer local msp "admin" was just an observations between us, that we kind of have two different representations of an orderer node admin now. I expect this will get resolved when we shift from TLS to a new authentication framework for APIs like these.

docs/source/create_channel/create_channel_participation.md
denyeart: Good, I'm fine using "orderer" and "ordering node" and "ordering service node" somewhat interchangeably, but I think you've addressed my concern that we really don't have an "ordering service" until there is a quorum of orderers consenting together.

docs/source/create_channel/create_channel_participation.md
denyeart: @pamandrejko I think it is fine with your clarification now. If we invest further, I'd recommend getting closer to test network configtx.yaml (with the separate orderer org) rather than getting closer to the sampleconfig configtx.yaml (with SampleOrg).

docs/source/create_channel/create_channel_participation.md
stephyee: Here's the comment, https://github.com/hyperledger/fabric/pull/2081#discussion_r522216983. IIUC this is true in the context of the API, I think we have a int test for this.

docs/source/create_channel/create_channel_participation.md
pamandrejko: Great. I saw that and dropped it into this tutorial as well:

**Important:** You need to create the `config.yaml` file and add it to the organization MSP and local MSP folder for each ordering node. This file enables Node OU support for the MSP, an important feature that allows the MSP's admin to be identified based on an "admin" OU in an identity's certificate. Learn more in the [Fabric CA](https://hyperledger-fabric-ca.readthedocs.io/en/release-1.4/deployguide/use_CA.html#nodeous) documentation.

denyeart:
Nice doc, let's go ahead and merge, we can always tweak further in a subsequent PR.

denyeart (approved):

denyeart:
This PR was merged and then reverted due to a squash issue.
Ultimately the doc content was merged using subsequent PR https://github.com/hyperledger/fabric/pull/2127.

PR #2082, opened 2020-11-06
Chan.Part.API: join system channel using config block IT

wlahti:
#### Type of change

- Test update

#### Description

First commit:
-  Use the configtx library to create the system channel genesis block, as it provides an easy way in tests for explicit control over the orderer endpoints and consenters for a channel.

Second commit: 
- join system channel using config block IT

Third commit:
- submitTxn IT helper cleanup to remove functions parameters that are no longer used

#### Related issues

[FAB-18273](https://jira.hyperledger.org/browse/FAB-18273)

### DIFF
--- a/integration/raft/channel_participation_test.go
+++ b/integration/raft/channel_participation_test.go
@@ -149,7 +149,7 @@ var _ = Describe("ChannelParticipation", func() {
 				Expect(channelInfo).To(Equal(expectedChannelInfoPT))
 			}
 
-			submitPeerTxn(orderer1, peer, network, consenters, 1, channelparticipation.ChannelInfo{
+			submitPeerTxn(orderer1, peer, network, channelparticipation.ChannelInfo{
 				Name:              "participation-trophy",
 				URL:               "/participation/v1/channels/participation-trophy",
 				Status:            "active",
@@ -157,7 +157,7 @@ var _ = Describe("ChannelParticipation", func() {
 				Height:            2,
 			})
 
-			submitPeerTxn(orderer2, peer, network, consenters, 2, channelparticipation.ChannelInfo{
+			submitPeerTxn(orderer2, peer, network, channelparticipation.ChannelInfo{
 				Name:              "participation-trophy",
 				URL:               "/participation/v1/channels/participation-trophy",
 				Status:            "active",
@@ -199,7 +199,7 @@ var _ = Describe("ChannelParticipation", func() {
 			}, network.EventuallyTimeout).Should(Equal(expectedChannelInfoPT))
 
 			By("submitting transaction to orderer3 to ensure it is active")
-			submitPeerTxn(orderer3, peer, network, orderers, 4, channelparticipation.ChannelInfo{
+			submitPeerTxn(orderer3, peer, network, channelparticipation.ChannelInfo{
 				Name:              "participation-trophy",
 				URL:               "/participation/v1/channels/participation-trophy",
 				Status:            "active",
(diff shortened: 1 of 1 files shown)


integration/raft/channel_participation_test.go
```diff
@@ -845,26 +974,25 @@ var _ = Describe("ChannelParticipation", func() {
...
+func submitOrdererTxn(o *nwo.Orderer, peer *nwo.Peer, n *nwo.Network, expectedChannelInfo channelparticipation.ChannelInfo) {
 	env := CreateBroadcastEnvelope(n, o, expectedChannelInfo.Name, []byte("hello"))
-	submitTxn(o, env, peer, n, orderers, expectedBlkNum, expectedChannelInfo)
+	submitTxn(o, env, peer, n, expectedChannelInfo)
 }
 
 // submit the envelope to the orderer and ensure it is committed
 // to the ledger
-func submitTxn(o *nwo.Orderer, env *common.Envelope, peer *nwo.Peer, n *nwo.Network, orderers []*nwo.Orderer, expectedBlkNum int, expectedChannelInfo channelparticipation.ChannelInfo) {
+func submitTxn(o *nwo.Orderer, env *common.Envelope, peer *nwo.Peer, n *nwo.Network, expectedChannelInfo channelparticipation.ChannelInfo) {
```
jyellick: peer is a parameter but is never used...

integration/raft/channel_participation_test.go
```diff
@@ -845,26 +974,25 @@ var _ = Describe("ChannelParticipation", func() {
...
+func submitPeerTxn(o *nwo.Orderer, peer *nwo.Peer, n *nwo.Network, expectedChannelInfo channelparticipation.ChannelInfo) {
 	env := CreateBroadcastEnvelope(n, peer, expectedChannelInfo.Name, []byte("hello"))
-	submitTxn(o, env, peer, n, orderers, expectedBlkNum, expectedChannelInfo)
+	submitTxn(o, env, peer, n, expectedChannelInfo)
 }
 
 // submit a transaction signed by the orderer and ensure it is
 // committed to the ledger
-func submitOrdererTxn(o *nwo.Orderer, peer *nwo.Peer, n *nwo.Network, orderers []*nwo.Orderer, expectedBlkNum int, expectedChannelInfo channelparticipation.ChannelInfo) {
+func submitOrdererTxn(o *nwo.Orderer, peer *nwo.Peer, n *nwo.Network, expectedChannelInfo channelparticipation.ChannelInfo) {
```
jyellick: Why is a peer involved in this at all?

integration/raft/channel_participation_test.go
```diff
@@ -968,6 +1096,80 @@ func applicationChannelGenesisBlock(n *nwo.Network, orderers []*nwo.Orderer, p *
 	return genesisBlock
 }
 
+func systemChannelGenesisBlock(n *nwo.Network, orderers []*nwo.Orderer, peers []*nwo.Peer, channel string) *common.Block {
```
jyellick, line 1099: This seems to be duplication of the existing logic for bootstrapping a network.  Did you look at simply re-using what's already present via nwo?

integration/raft/channel_participation_test.go
wlahti: Oops, missed that. It's because this function originated as a helper to submit a transaction signed by a peer instead of an orderer.

integration/raft/channel_participation_test.go
wlahti: Oops, missed that. It's because this function originated as a helper to submit a transaction signed by a peer instead of an orderer.

integration/raft/channel_participation_test.go
wlahti: The key difference is the flexibility of including different orderers in the consenters set without having a different configtx.yaml (or template). By default, nwo includes all orderers in the consenters set and the endpoints.

integration/raft/channel_participation_test.go
wlahti: Done

integration/raft/channel_participation_test.go
wlahti: Done

jyellick (approved):

PR #2083, opened 2020-11-06
Ch.Part.API: Allow retry of failed removals

stephyee:
#### Type of change

- Improvement 

#### Description

Allow retrying of failed channel removals per [comment](https://github.com/hyperledger/fabric/pull/2027/files#r518875811)

### DIFF
--- a/orderer/common/multichannel/registrar.go
+++ b/orderer/common/multichannel/registrar.go
@@ -916,12 +916,8 @@ func (r *Registrar) RemoveChannel(channelID string) error {
 	r.lock.Lock()
 	defer r.lock.Unlock()
 
-	if removalStatus, ok := r.pendingRemoval[channelID]; ok {
-		if removalStatus {
-			return types.ErrChannelPendingRemoval
-		} else {
-			return types.ErrChannelRemovalFailure
-		}
+	if removalStatus := r.pendingRemoval[channelID]; removalStatus {
+		return types.ErrChannelPendingRemoval
 	}
 
 	if r.systemChannelID != "" {
@@ -1087,15 +1083,14 @@ func (r *Registrar) removeLedgerAsync(channelID string) {
 	r.pendingRemoval[channelID] = true
 
 	go func() {
-		if err := r.ledgerFactory.Remove(channelID); err != nil {
-			r.lock.Lock()
+		err := r.ledgerFactory.Remove(channelID)
+		r.lock.Lock()
+		defer r.lock.Unlock()
+		if err != nil {
 			r.pendingRemoval[channelID] = false
-			r.lock.Unlock()
 			logger.Errorf("ledger factory failed to remove empty ledger '%s', error: %s", channelID, err)
 			return
 		}
-		r.lock.Lock()
 		delete(r.pendingRemoval, channelID)
-		r.lock.Unlock()
 	}()
 }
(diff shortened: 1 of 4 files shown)


orderer/common/multichannel/registrar.go
```diff
@@ -916,12 +916,8 @@ func (r *Registrar) RemoveChannel(channelID string) error {
 	r.lock.Lock()
 	defer r.lock.Unlock()
 
-	if removalStatus, ok := r.pendingRemoval[channelID]; ok {
-		if removalStatus {
-			return types.ErrChannelPendingRemoval
-		} else {
-			return types.ErrChannelRemovalFailure
-		}
+	if removalStatus := r.pendingRemoval[channelID]; removalStatus {
```
yacovm, line 919: Can't we simply make the `pendingRemoval` map, a map from `string` to `struct{}`  ? 

I see it this way:
1. If we are in the process of removing, then we have the element in the map, so we return a pending error.
2. If we succeeded the removal, then great, we return that it doesn't exist so we can't remove it.
3. Else, we failed the removal, so we just re-try and it's equivalent to the first removal attempt, no?

Why do we need to keep a boolean map value if we have 3 possible states, and 2 maps (existing channels and pending channels) which can encode 4 states?

wlahti: Agreed. Definitely makes sense.

tock-ibm: I disagree. The remove is asynch, so first the request thread stops the chain/follower, moves the chain/follower to the pendingRemove map, spawns a go-routine to remove it, and returns. You need some way of signaling to a potential second remove request thread that the remove is on-going in the background (with the go-routine spawned by the first). This is exactly the boolean in the third map - `true` signals an ongoing remove, `false` a failed remove, and the key-value is deleted when the remove succeeds. If we had a struct{} in there, there was no way of detecting that background removal is in progress or when it ended, for good or bad. Note that the chains/followers/pendingRemove maps are mutually exclusive. A `map[string]boolean` can encodes these 3 states, whereas a `map[string]struct{}` cannot.

In addition, see this comment here: https://github.com/hyperledger/fabric/pull/2027#discussion_r516476013 which advocated for a 
```
struct {
  cRel     types.ClusterRelation
  status  types.Status
}
```
in the `pendingRemove` map instead of a boolean, and a new additional status `types.StatusFailure` to make diagnosing failures even easier.

tock-ibm:
Looks good to me!
Good suggestion by @yacovm .

orderer/common/multichannel/registrar.go
wlahti: Hmm, I'll admit to not being as familiar with the flow here and will defer to Yoav. :)

yacovm: >  You need some way of signaling to a potential second remove request thread that the remove is on-going in the background (with the go-routine spawned by the first). 

The fact that the channel is in the pending map conveys that. As long as the mapping exists, there is a goroutine active, and whenever the mapping doesn't exist, there is no such goroutine. This implies that the goroutine deletes the mapping once it is done - whether it succeeded or failed.

tock-ibm: That is not a good idea. 
The go-routine should set the mapping channelID->false when it fails and terminates. This way:
1. a `List` (GET request) of the channel can reveal that the removal had failed
2. subsequent Join requests are rejected (the ledger may be in an inconsistent state)
3. but remove can be retried safely.
If we don't keep it in the map, the Registrar has no record that the removal had failed, or that the channel exists, since it is not in either channels/followers/pendingRemove.

yacovm: Under the assumption that we want to reject joins, then let's go ahead with what you propose

yacovm (approved):

PR #2085, opened 2020-11-09
[FAB-18319] Ch.Part.API: IT Remove and rejoin channel from orderer

stephyee:
#### Type of change

- Test update

#### Description

Test removal of a channel from an orderer and join it back again to the same orderer.

#### Related issues

[FAB-18319](https://jira.hyperledger.org/browse/FAB-18319)

### DIFF
--- a/integration/raft/channel_participation_test.go
+++ b/integration/raft/channel_participation_test.go
@@ -286,13 +286,58 @@ var _ = Describe("ChannelParticipation", func() {
 			cl = channelparticipation.List(network, orderer1)
 			channelparticipation.ChannelListMatcher(cl, []string{"another-participation-trophy"})
 
+			By("joining orderer1 to channel it was previously removed from as consenter")
+			configBlockPT = nwo.GetConfigBlock(network, peer, orderer2, "participation-trophy")
+			expectedChannelInfoPTFollower = channelparticipation.ChannelInfo{
+				Name:              "participation-trophy",
+				URL:               "/participation/v1/channels/participation-trophy",
+				Status:            "onboarding",
+				ConsensusRelation: "follower",
+				Height:            0,
+			}
+			channelparticipation.Join(network, orderer1, "participation-trophy", configBlockPT, expectedChannelInfoPTFollower)
+
+			By("ensuring orderer1 completes onboarding successfully")
+			expectedChannelInfoPTFollower.Status = "active"
+			expectedChannelInfoPTFollower.Height = 7
+			Eventually(func() channelparticipation.ChannelInfo {
+				return channelparticipation.ListOne(network, orderer1, "participation-trophy")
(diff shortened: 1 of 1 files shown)


wlahti:
A general comment... why isn't this just added on to the existing test you used as the starting point for this? Seems like this PR would be about 30 lines instead of 155 and would be easier to review that way.

integration/raft/channel_participation_test.go
```diff
@@ -731,6 +731,161 @@ var _ = Describe("ChannelParticipation", func() {
...
+			}, network.EventuallyTimeout).Should(Equal(expectedChannelInfo))
+
+			submitPeerTxn(orderer3, peer, network, channelparticipation.ChannelInfo{
+				Name:              "participation-trophy",
+				URL:               "/participation/v1/channels/participation-trophy",
+				Status:            "active",
+				ConsensusRelation: "consenter",
+				Height:            9,
+			})
+		})
```
tock-ibm: Testing that onboarding completes should be performed against orderer1, which just joined, right?
Then test orderer1 is a follower, picking up the latest block, turn orderer1 into a consenter (member), and submit a tx to it as well?

tock-ibm: BTW, is there a test case where we delete a channel from an orderer while it is a consenter, not a follower? i.e first delete the channel from ordererX, and only then remove ordererX from the consenters set of that channel?

stephyee: As @wlahti noted, this test is based off another and I forget to switch to orderer1 instead as he suggested I'm going to add to the existing test. Also no, we don't have a test where we delete while the orderer is a consenter.

integration/raft/channel_participation_test.go
```diff
@@ -731,6 +731,161 @@ var _ = Describe("ChannelParticipation", func() {
 				Height:            7,
 			})
 		})
+
+		It("joins application channels from genesis block and removes a channel using the channel participation API", func() {
```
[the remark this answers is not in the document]
stephyee: I missed that. Now, I've added the rejoin logic to an existing test.

integration/raft/channel_participation_test.go
```diff
@@ -731,6 +776,7 @@ var _ = Describe("ChannelParticipation", func() {
 				Height:            7,
 			})
 		})
+
```
wlahti: Super nit that doesn't need to be addressed: I guess your editor settings like to format in this extra line?

wlahti (dismissed):
Looks good to me!

tock-ibm (dismissed):

integration/raft/channel_participation_test.go
stephyee: or me not sure 😃

tock-ibm (approved):

Qdigital (approved):

C0rWin (approved):

PR #2089, opened 2020-11-09
[FAB-18329] Fix data race in cluster/comm_test#TestRenewCertificates

yacovm:
This commit fixes a data race in orderer/common/cluster/comm_test.go#TestRenewCertificates.

The race occurred because the TLS configuration was updated from the test goroutine,
while a gRPC connection was being established from another goroutine, and as a result,
the TLS configuration was loaded without memory synchronization with the first goroutine.

I made the test close the gRPC connections before reconfiguring the communication layer.

Change-Id: I82b9cc685e9160e480ce77ec7e0a233b106eb0e5

### DIFF
--- a/orderer/common/cluster/comm_test.go
+++ b/orderer/common/cluster/comm_test.go
@@ -197,6 +197,11 @@ func (cn *clusterNode) stop() {
 }
 
 func (cn *clusterNode) renewCertificates() {
+	// Stop the gRPC service
+	cn.srv.Stop()
+	// and restart it afterwards
+	defer cn.resurrect()
+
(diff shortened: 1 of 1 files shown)


caod123 (dismissed):
Verified with `go test -run TestRenewCertificate --race -count=100 -failfast`

caod123 (dismissed):

caod123 (approved):

yacovm:
> Verified with `go test -run TestRenewCertificate --race -count=100 -failfast`

Yet it failed the first time it ever ran in CI. 
These VMs are jinxed. No other explanation.

PR #2093, opened 2020-11-09
Revert "Allow BCCSP config to be set using env var (#1900)"

lindluni:
This reverts commit 0a9f7663a3a548385f3e4fa55e147b6dbce055d4 which pushed the pkcs11 code outside of the pkcs11 build tag. This prevents cross-compilation of Fabric due to the fact the miekg/pkcs11 package contains CGO which requires a cross-compiler for each platform you are cross compiling for.

We need to find a better approach to solving this problem that doesn't pull the code out of the build tag.

### DIFF
--- a/bccsp/factory/nopkcs11.go
+++ b/bccsp/factory/nopkcs11.go
@@ -15,6 +15,12 @@ import (
 
 const pkcs11Enabled = false
 
+// FactoryOpts holds configuration information used to initialize factory implementations
+type FactoryOpts struct {
+	Default string  `json:"default" yaml:"Default"`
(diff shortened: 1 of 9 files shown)


denyeart:
@btl5037 @wlahti  I assume the release-2.2 PR also needs to be reverted?

lindluni:
> @btl5037 @wlahti I assume the release-2.2 PR also needs to be reverted?

I didn't realize this was backported, I'll revert it there as well.

wlahti:
We never really wanted to do this one but you guys kept pushing us. ;) Makes sense to me.

caod123 (approved):

PR #2105, opened 2020-11-10
Revert "Allow BCCSP config to be set using env var (#1900)"

caod123:
This reverts commit f2e9e6e7b307a25dc4738e97a1ffb08daea4f6dc.
which pushed the pkcs11 code outside of the pkcs11 build tag.
This prevents cross-compilation of Fabric due to the fact the
miekg/pkcs11 package contains CGO which requires a cross-compiler
for each platform you are cross compiling for.

We need to find a better approach to solving this problem
that doesn't pull the code out of the build tag.

### DIFF
--- a/bccsp/factory/nopkcs11.go
+++ b/bccsp/factory/nopkcs11.go
@@ -15,6 +15,12 @@ import (
 
 const pkcs11Enabled = false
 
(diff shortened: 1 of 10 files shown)


caod123:
Backport revert of #2093

lindluni (approved):
Thanks Danny

PR #2108, opened 2020-11-10
[FAB-18315] Ch.Part.API: Add remove failure metric for chain status

stephyee:
#### Type of change

- New feature

#### Description
Adds StatusFailure metric for channel removal failures. Ensure GET List/:channelID returns correct ChannelInfo for pending channel removals. 

#### Related issues
[FAB-18315](https://jira.hyperledger.org/browse/FAB-18315)

### DIFF
--- a/orderer/common/multichannel/registrar.go
+++ b/orderer/common/multichannel/registrar.go
@@ -50,8 +50,10 @@ type Registrar struct {
 	lock      sync.RWMutex
 	chains    map[string]*ChainSupport
 	followers map[string]*follower.Chain
-	// true indicates a ledger removal is pending; false indicates a ledger removal failed
-	pendingRemoval  map[string]bool
+	// existence indicates removal is in-progress or failed
+	// when failed, the status will indicate failed all other states
+	// denote an in-progress removal
+	pendingRemoval  map[string]consensus.StaticStatusReporter
 	systemChannelID string
 	systemChannel   *ChainSupport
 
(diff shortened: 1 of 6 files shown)


orderer/common/multichannel/registrar.go
```diff
@@ -719,11 +721,11 @@ func (r *Registrar) JoinChannel(channelID string, configBlock *cb.Block, isAppCh
...
 
-	if removalStatus, ok := r.pendingRemoval[channelID]; ok {
-		if removalStatus {
-			return types.ChannelInfo{}, types.ErrChannelPendingRemoval
-		} else {
+	if status, ok := r.pendingRemoval[channelID]; ok {
+		if status.Status == types.StatusFailure {
 			return types.ChannelInfo{}, types.ErrChannelRemovalFailure
+		} else {
+			return types.ChannelInfo{}, types.ErrChannelPendingRemoval
```
wlahti: Style nit: no need for the else here.
```suggestion
		if status.Status == types.StatusFailure {
			return types.ChannelInfo{}, types.ErrChannelRemovalFailure
		} 
	       return types.ChannelInfo{}, types.ErrChannelPendingRemoval
```

orderer/common/types/channelinfo.go
```diff
@@ -59,6 +59,7 @@ const (
 	StatusOnBoarding Status = "onboarding"
 	// The orderer is not storing any blocks for this channel.
 	StatusInactive Status = "inactive"
+	StatusFailure  Status = "failed"
```
wlahti: Add a comment here to explain the status? 

And I think we should have the constant name consistent with the value, so either StatusFailure = "failure" or StatusFailed = "failed". Not sure how I feel about "failed"... maybe "error" instead? Any thoughts @tock-ibm @jyellick ?

orderer/common/multichannel/registrar.go
stephyee: 🧹

orderer/common/types/channelinfo.go
stephyee: For now, I went with StatusFailed. The comment you may hate.

yacovm (approved):

PR #2109, opened 2020-11-11
Add snapshots to peer deploy checklist

joealewine:
Change-Id: If4f87761b9af1ea2941cc23dd0266f7f52512315

#### Type of change

- Documentation update

#### Description

Add snapshot parameter to peer deployment guide. Specifically, the "checklist" documenting parameters in core.yaml.

### DIFF
--- a/docs/source/deploypeer/peerchecklist.md
+++ b/docs/source/deploypeer/peerchecklist.md
@@ -330,6 +330,16 @@ This section is used to select your ledger database type, either `goleveldb` `Co
 
 - **`ledger.state.couchDBConfig.password:`** (Required when using CouchDB.) Specify the password for the CouchDB user with read and write authority to the database.
 
+The `ledger` section also contains your default snapshot directory where snapshots are stored. For more information about snapshots, check out [Taking ledger snapshots and using them to join channels](../peer_ledger_snapshot.html).
+
+```
+snapshots:
+  # Path on the file system where peer will store ledger snapshots
+  rootDir: /var/hyperledger/production/snapshots
+```
+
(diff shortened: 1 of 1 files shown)


docs/source/deploypeer/peerchecklist.md
```diff
@@ -330,6 +330,16 @@ This section is used to select your ledger database type, either `goleveldb` `Co
...
 
+The `ledger` section also contains your default snapshot directory where snapshots are stored. For more information about snapshots, check out [Taking ledger snapshots and using them to join channels](../peer_ledger_snapshot.html).
+
+```
+snapshots:
+  # Path on the file system where peer will store ledger snapshots
+  rootDir: /var/hyperledger/production/snapshots
+```
+
+- **`ledger.snapshots.rootDir:`** (Default value should be overridden.) This is the path to where snapshots are stored on the local filesystem of the peer. It can be an absolute path or relative to `FABRIC_CFG_PATH` and defaults to `/var/hyperledger/production/snapshots`, further organized by the status, channel name, and block number of the snapshot. For example, `{ledger.snapshots.rootDir}/completed/{channelName}/{lastBlockNumberInSnapshot}`. The user running the peer needs to own and have write access to this directory.
```
manish-sethi: > further organized by the status, channel name, and block number of the snapshot. For example, `{ledger.snapshots.rootDir}/completed/{channelName}/{lastBlockNumberInSnapshot}.

I don't think that this portion makes sense at this level. However, if you want to keep this, then it's better to clarify that user does not have to configure anything for this. Also, there is no other status as such. So, I would rewrite this as - Peer node generates the snapshot for a channel at a specific block number under in a directory `{ledger.snapshots.rootDir}/completed/{channelName}/{lastBlockNumberInSnapshot}`

joealewine: Done.

manish-sethi (approved):

PR #2114, opened 2020-11-12
Remove custom error types from txmgr package

manish-sethi:

#### Type of change

### DIFF
--- a/core/ledger/kvledger/txmgmt/txmgr/query_executor.go
+++ b/core/ledger/kvledger/txmgmt/txmgr/query_executor.go
@@ -7,8 +7,6 @@ SPDX-License-Identifier: Apache-2.0
 package txmgr
 
 import (
-	"fmt"
-
 	"github.com/hyperledger/fabric-protos-go/ledger/queryresult"
 	"github.com/hyperledger/fabric-protos-go/ledger/rwset/kvrwset"
 	commonledger "github.com/hyperledger/fabric/common/ledger"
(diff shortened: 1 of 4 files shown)


core/ledger/kvledger/txmgmt/txmgr/query_executor.go
```diff
@@ -214,9 +212,10 @@ func (q *queryExecutor) GetPrivateData(ns, coll, key string) ([]byte, error) {
 		return nil, err
 	}
 	if !version.AreSame(hashVersion, ver) {
-		return nil, &ErrPvtdataNotAvailable{Msg: fmt.Sprintf(
+		return nil, errors.Errorf(
```
cendhu, line 215: Does it return a stack trace to the chaincode invoker or just the error message? Same question for other log messages too. I think just the error message is sufficient for the end-user.

manish-sethi: We shouldn't be worrying at this deep level about this. There are many layers above this as well and keeping track at each level is not a good way. The last layer should propagate just the string and wherever in between we log, the stack is printed. I checked in the chaincode package [here](https://github.com/hyperledger/fabric/blob/24868e956881c604d7dc1b561ce3f18ce2fc62b5/core/chaincode/handler.go#L250) and at that level it returns the string only `err.Error()`

denyeart (approved):

core/ledger/kvledger/txmgmt/txmgr/query_executor.go
cendhu: I see. Make sense.

PR #2121, opened 2020-11-13
Restore ability to use RSA CA with Fabric MSP

sykesm:
While RSA has never been supported by Fabric for transaction signing and verification, prior to version 2.x, MSPs could include CA certificates that included RSA public keys. This was broken with the removal of the unsupported RSA code.

This change modifies the key import function used by the MSP to convert an RSA public key to a bccsp.Key that can then be used as part of an MSP identity.

This code does *not* add support for RSA on transaction paths.

### DIFF
--- a/bccsp/sw/keyimport.go
+++ b/bccsp/sw/keyimport.go
@@ -8,6 +8,7 @@ package sw
 
 import (
 	"crypto/ecdsa"
+	"crypto/rsa"
 	"crypto/x509"
 	"errors"
 	"fmt"
@@ -123,12 +124,16 @@ func (ki *x509PublicKeyImportOptsKeyImporter) KeyImport(raw interface{}, opts bc
 
 	pk := x509Cert.PublicKey
 
-	switch pk.(type) {
+	switch pk := pk.(type) {
 	case *ecdsa.PublicKey:
 		return ki.bccsp.KeyImporters[reflect.TypeOf(&bccsp.ECDSAGoPublicKeyImportOpts{})].KeyImport(
 			pk,
 			&bccsp.ECDSAGoPublicKeyImportOpts{Temporary: opts.Ephemeral()})
+	case *rsa.PublicKey:
+		// This path only exists to support environments that use RSA certificate
+		// authorities to issue ECDSA certificates.
+		return &rsaPublicKey{pubKey: pk}, nil
 	default:
-		return nil, errors.New("Certificate's public key type not recognized. Supported keys: [ECDSA]")
+		return nil, errors.New("Certificate's public key type not recognized. Supported keys: [ECDSA, RSA]")
 	}
 }
--- a/integration/msp/rsaca_test.go
+++ b/integration/msp/rsaca_test.go
@@ -1,0 +1,467 @@
+/*
+Copyright IBM Corp. All Rights Reserved.
+
+SPDX-License-Identifier: Apache-2.0
+*/
+
+package msp
+
+import (
+	"crypto"
+	"crypto/ecdsa"
+	"crypto/elliptic"
+	"crypto/rand"
+	"crypto/rsa"
+	"crypto/sha256"
+	"crypto/x509"
+	"crypto/x509/pkix"
+	"encoding/pem"
+	"fmt"
+	"io/ioutil"
+	"math/big"
+	"net"
+	"os"
+	"path/filepath"
+	"syscall"
+	"time"
+
+	docker "github.com/fsouza/go-dockerclient"
+	"github.com/hyperledger/fabric/integration/nwo"
+	"github.com/hyperledger/fabric/integration/nwo/commands"
+	fabricmsp "github.com/hyperledger/fabric/msp"
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+	"github.com/onsi/gomega/gexec"
+	"github.com/tedsuo/ifrit"
+	"gopkg.in/yaml.v2"
+)
+
+var _ = Describe("MSPs with RSA Certificate Authorities", func() {
+	var (
+		client  *docker.Client
+		testDir string
+		network *nwo.Network
+		process ifrit.Process
+	)
+
+	BeforeEach(func() {
+		var err error
+		testDir, err = ioutil.TempDir("", "msp")
+		Expect(err).NotTo(HaveOccurred())
+
+		client, err = docker.NewClientFromEnv()
+		Expect(err).NotTo(HaveOccurred())
+
+		network = nwo.New(nwo.BasicEtcdRaft(), testDir, client, StartPort(), components)
+		network.GenerateConfigTree()
+
+		By("manually bootstrapping MSPs with RSA CAs")
+		generateRSACACrypto(network)
+		network.CreateDockerNetwork()
+		sess, err := network.ConfigTxGen(commands.OutputBlock{
+			ChannelID:   network.SystemChannel.Name,
+			Profile:     network.SystemChannel.Profile,
+			ConfigPath:  network.RootDir,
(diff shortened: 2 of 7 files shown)


integration/msp/rsaca_test.go
```diff
@@ -0,0 +1,467 @@
...
+		nwo.EnableCapabilities(
+			network,
+			"testchannel",
+			"Application", "V2_0",
+			orderer,
+			network.Peer("Org1", "peer0"),
+			network.Peer("Org2", "peer0"),
+		)
+		nwo.DeployChaincode(network, "testchannel", orderer, chaincode)
+		RunQueryInvokeQuery(network, orderer, org1Peer0, 100)
```
yacovm, line 127: How about fetching the channel config and ensuring that the CAs are indeed RSA?

sykesm: Go for it. Since they were explicitly generated by the code below the test and the test failed before changing the production code (ie test-driven), I personally don't see the value.

bccsp/sw/keyimport.go
```diff
@@ -123,12 +124,16 @@ func (ki *x509PublicKeyImportOptsKeyImporter) KeyImport(raw interface{}, opts bc
...
 		return ki.bccsp.KeyImporters[reflect.TypeOf(&bccsp.ECDSAGoPublicKeyImportOpts{})].KeyImport(
 			pk,
 			&bccsp.ECDSAGoPublicKeyImportOpts{Temporary: opts.Ephemeral()})
+	case *rsa.PublicKey:
+		// This path only exists to support environments that use RSA certificate
+		// authorities to issue ECDSA certificates.
+		return &rsaPublicKey{pubKey: pk}, nil
 	default:
-		return nil, errors.New("Certificate's public key type not recognized. Supported keys: [ECDSA]")
+		return nil, errors.New("Certificate's public key type not recognized. Supported keys: [ECDSA, RSA]")
```
yacovm, line 137: This makes me wonder - let's say I have a v2.2 Fabric environment prior to this commit and I add an RSA CA to my config block. What happens? Does the peer crash when it initializes an MSP manager? Or does the peer simply invalidate transactions signed by identities issued by the RSA CA? If it's the former then we're safe to add this, but if it's the latter it is some potential for non-determinism across versions.

ale-linux (commented):
Thanks Matt! I'm just wondering whether it'd be actually better to exercise the existing e2e integration tests (e.g. those in integration/e2e/e2e_test.go) with the only difference that one of the CAs uses RSA. Also, shouldn't we test whether such a channel is upgradable?

bccsp/sw/keyimport.go
sykesm: I had a discussion with @jyellick about this. My takeaway was that the failure mode presents as crashing peers and orderers. 

In the case of the peer, if an active configuration is using an RSA CA and the peer is missing the RSA code (2.x), starting a peer that does not support it results in a crash. If a config block arrives with an RSA CA, validation fails and the peer panics.

There were subtleties related to validation performed by the orderer running in a mixed version cluster that I didn't completely grok so he may need to comment on those scenarios.

At the end of the conversation, we did come away concluding that this would not result in non-determinism.

integration/msp/rsaca_test.go
```diff
@@ -0,0 +1,467 @@
...
+	})
+})
+
+// What follows is a bunch of code to build a hand-crafted set of MSPs
+// where the CAs use RSA keys and everything else is ECDSA. It's not
+// pretty but it gets the job done for testing.
+
+func generateRSACACrypto(n *nwo.Network) {
```
yacovm, line 135: So I don't understand why we are essentially replicating what cryptogen is doing only with RSA CAs. 
Isn't it simpler to:

1. Run cryptogen
2. Locate all CAs, replace their public key and re-sign them.
3. Locate all certificate, replace their signature with a new signature
?

bccsp/sw/keyimport.go
jyellick: To add a little more context to what @sykesm has said above, in this case, because we are dealing only with CA certs and not with signing certs a lot of the complications disappear.  The only transactions which depend on CA cert evaluations are channel config updates.

Channel config updates are always preprocessed by the orderer.  In general, the update is first checked by the orderer to which it was Broadcast.  Then there may be a further check by the leader in the Raft case.  If either of these checks fails, then the config update is simply dropped, and it does not appear in the chain.  If both of these checks pass, then the config update is applied as a new config block in the chain.  The important thing to note is that config correctness checks always occur _before_ consensus.  This was a critical change that was made between v1.0 and v1.1 which practically eliminates the possibility of forking the blockchain at the orderer based on non-deterministic config update processing.

Once a config change achieves order via consensus, it is simply applied, with no validation step.  If the apply fails, then the node simply crashes.

At the peer side, the behavior is similar, though slightly different.  The peer does not trust that the ordering service appropriately validated the config update, so the peer redoes the signature validation etc. before finally applying the new config.  If the signature checking fails, or applying the new config fails, the peer, like the orderer, simply crashes.

So, for any config update, which is non-deterministically evaluated, we still have the guarantee that it either applies everywhere or not at all.  In a mixed mode network, with v1.4.x and v2.1.x nodes, the v2.1.x orderers would reject the config update, the v1.4.x ones would accept it.  So, if you submitted to a v1.4.x node, and the leader for the channel was a v1.4.x node, all the v2.1.x orderers would crash, but, there would be no fork.  If either the node receiving the broadcast or the leader were v2.1.x then the config update would simply be rejected or dropped.

TLDR; Matt is absolutely right here, there is no potential to fork.  Although, it's possible for a malicious config update to get through in a mixed mode network with binaries that do and don't support RSA TLS CAs, this would simply result in a crash, which could be fixed by moving to a newer version of Fabric.

integration/msp/rsaca_test.go
```diff
@@ -0,0 +1,467 @@
...
+	signcertKeyFilename := filepath.Join(mspDir, "keystore", fmt.Sprintf("%x_sk", signcert.SubjectKeyId))
+	writeKey(signcertKeyFilename, priv)
+
+	// populate tls
+	tlsDir := filepath.Join(baseDir, "tls")
+	err = os.MkdirAll(tlsDir, 0755)
+	Expect(err).NotTo(HaveOccurred())
+	writeCertificate(filepath.Join(tlsDir, "ca.crt"), tlsCA.certBytes)
+
+	tlsKey := generateRSAKey()
```
yacovm, line 298: I understand you're testing this, but why is this needed? BCCSP never processes any TLS certificates so the production code change is not affecting this part of the test.

bccsp/sw/keyimport.go
yacovm: Got it, thanks Matt and Jason.

integration/msp/rsaca_test.go
sykesm: Nothing in the integration tests RSA TLS certificates. Since this test is explicitly targeted to RSA, it seemed reasonable to have something here to help prevent future regressions.

integration/msp/rsaca_test.go
sykesm: This started out with the assumption that RSA was used throughout the MSP so this code generated an entire MSP tree with RSA. When it became clear that signing certificates could not use RSA, I changed that path to use ECDSA. As you can see, that resulted in the introduction of `generateECKey` and a call to it from one site.

As the comment at the top of the code says, "It's not pretty but it gets the job done for testing." I stand by that. Changing this to walk through all of the crypto directories to re-sign certificates is nearly as complicated as what's here. What's here now works, is self-contained, and isolated to the single test that needs it. I do not believe rewriting it to parse and re-sign certs is worth the effort.

denyeart (approved):

PR #2133, opened 2020-11-17
Add check for invalid key before hitting couchdb

manish-sethi:

#### Type of change
- Bug fix

#### Related issues
[FAB-13897](https://jira.hyperledger.org/browse/FAB-13897)

### DIFF
--- a/core/ledger/kvledger/txmgmt/statedb/statecouchdb/statecouchdb.go
+++ b/core/ledger/kvledger/txmgmt/statedb/statecouchdb/statecouchdb.go
@@ -526,6 +526,9 @@ func (vdb *VersionedDB) readFromDB(namespace, key string) (*keyValue, error) {
 	if err != nil {
 		return nil, err
 	}
+	if err := validateKey(key); err != nil {
+		return nil, err
+	}
 	couchDoc, _, err := db.readDoc(key)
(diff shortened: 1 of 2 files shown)


core/ledger/kvledger/txmgmt/statedb/statecouchdb/statecouchdb.go
```diff
@@ -526,6 +526,9 @@ func (vdb *VersionedDB) readFromDB(namespace, key string) (*keyValue, error) {
 	if err != nil {
 		return nil, err
 	}
+	if err := validateKey(key); err != nil {
+		return nil, err
+	}
```
cendhu, line 531: If the goal is to avoid sending invalid keys to CouchDB but not to the cache, this PR LGTM. Is that the goal? If not, we need to perform this check before accessing the cache also.

manish-sethi: Cache would eventually be populated from DB content only. So, checking in cache does not add any value. Rather, in the current way, it amortizes the cost of this check.

denyeart (approved):

cendhu (approved):

denyeart:

PR #2141, opened 2020-11-17
Update core.yaml, orderer.yaml and docs for TLS

denyeart:
#### Type of change

- Documentation update

#### Description

Update TLS instructions in yaml files and docs.
Specifically, indicate that TLS CA certs and client TLS CA certs do not
need to be configured on every peer and orderer, since peers and orderers
automatically use the TLS CA certs that exist in the channel configuration.
However, additional TLS CA certs can be configured for peers and orderers to
augment the list of TLS CA certs from the channel configuration.


docs/source/deploypeer/peerchecklist.md
```diff
@@ -200,26 +200,31 @@ Peers leverage the Gossip data dissemination protocol to broadcast ledger and ch
...
+        file: tls/server.key
+    # rootcert.file represents the trusted root certificate chain used for verifying certificates
+    # of other nodes during outbound connections.
+    # It is not required to be set, but can be used to augment the set of TLS CA certificates
+    # available from the MSPs of each channel’s configuration.
+    rootcert:
+        file: tls/ca.crt
+    # If mutual TLS is enabled, clientRootCAs.files contains a list of additional root certificates
+    # used for verifying certificates of client connections.
+    # It is not required to be set, but can be used to augment the set of TLS CA certificates
```
jyellick, line 223: I don't think this is quite right... if you require client auth, and you do not set this, how can you join a peer to a channel? It's only 'not required' once your peer has channels.

yacovm (approved):

docs/source/deploypeer/peerchecklist.md
yacovm: I think he meant it is not required to be set since the mutual TLS is also not required

PR #2146, opened 2020-11-18
[FAB-18339] CLI: update GetCerificate to only load client certificate as it is supposed to be

wenjianqiao:

#### Type of change
- Bug fix

#### Description
Both "peer lifecycle chaincode" and "peer chaincode" commands need to get
the client certificate (when peer.tls.clientAuthRequired=true) for deliver service.
GetCertificate() returns a client certificate to pass to DeliverService. Currently, it
creates a PeerClient that unnecessarily loads server tls root file based on env vars,
which caused a problem loading tls files from env even if --tlsRootCertFile
already overwrites the env var. 

The fix is to update GetCertificate to only load client certificate and rename it 
to GetClientCertificate to avoid confusion.

#### Related issues
https://jira.hyperledger.org/browse/FAB-18339

### DIFF
--- a/internal/peer/common/common.go
+++ b/internal/peer/common/common.go
@@ -73,8 +73,8 @@ var (
 	GetOrdererEndpointOfChainFnc func(chainID string, signer Signer,
 		endorserClient pb.EndorserClient, cryptoProvider bccsp.BCCSP) ([]string, error)
 
-	// GetCertificateFnc is a function that returns the client TLS certificate
-	GetCertificateFnc func() (tls.Certificate, error)
+	// GetClientCertificateFnc is a function that returns the client TLS certificate
+	GetClientCertificateFnc func() (tls.Certificate, error)
 )
 
 type CommonClient struct {
@@ -90,7 +90,7 @@ func init() {
 	GetOrdererEndpointOfChainFnc = GetOrdererEndpointOfChain
 	GetDeliverClientFnc = GetDeliverClient
 	GetPeerDeliverClientFnc = GetPeerDeliverClient
-	GetCertificateFnc = GetCertificate
+	GetClientCertificateFnc = GetClientCertificate
 }
 
 // InitConfig initializes viper config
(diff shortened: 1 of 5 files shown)


internal/peer/common/peerclient.go
```diff
@@ -24,6 +24,14 @@ type PeerClient struct {
 	CommonClient
 }
 
+func NewPeerClient(address, tlsRootCertFile string) (*PeerClient, error) {
+	if address != "" {
+		return NewPeerClientForAddress(address, tlsRootCertFile)
+	} else {
+		return NewPeerClientFromEnv()
+	}
```
wlahti: Nit: now that this is in its own function no need for the else.
```suggestion
	} 
	return NewPeerClientFromEnv()
```

wenjianqiao: Done

ale-linux:
Thanks @wenjianqiao, one Q: if we don't load server certs, how can the client be sure that it contacts the intended OS node?

wenjianqiao:
@ale-linux Good question. GetCertificate should be called GetClientCertificate because it only `returns the client's TLS certificate` (the client cert is passed to deliver service when mutual TLS is enabled). Therefore it is unnecessary to load server certs in this function. The CLI commands already load the server certs in GetEndorserClient, and GetPeerDeliverClient, etc. I will rename GetCertificate to GetClientCertificate to avoid confusion.

ale-linux:
> @ale-linux Good question. GetCertificate should be called GetClientCertificate because it only `returns the client's TLS certificate` (the client cert is passed to deliver service when mutual TLS is enabled). Therefore it is unnecessary to load server certs in this function. The CLI commands already load the server certs in GetEndorserClient, and GetPeerDeliverClient, etc. I will rename GetCertificate to GetClientCertificate to avoid confusion.

Thanks @wenjianqiao, so the server certs are already available, and yet `GetCertificate` was loading them anyway

internal/peer/common/peerclient.go
```diff
@@ -148,25 +137,30 @@ func GetEndorserClient(address, tlsRootCertFile string) (pb.EndorserClient, erro
...
+
+	key, certificate, err := getClientAuthInfoFromEnv("peer")
 	if err != nil {
 		return tls.Certificate{}, err
 	}
-	return peerClient.Certificate(), nil
+	if key == nil || certificate == nil {
+		return tls.Certificate{}, errors.New("both Key and Certificate are required when using mutual TLS")
+	}
```
wlahti: I was going to request that we update the error message here to mention the exact names of the config parameters in the error message, but looking closer, it seems like we'll never reach this code. `getClientAuthInfoFromEnv()` only returns nils when the error is not nil. If we're trying to check for empty files here, `ioutil.ReadFile` seems to return a zero-length byte array in that case.

wenjianqiao: @wlahti good point. I have removed the checking because `tls.X509KeyPair` will return an error for empty key/cert. Also added unit tests for GetClientCertificate.

internal/peer/common/common.go
```diff
@@ -255,25 +255,29 @@ func configFromEnv(prefix string) (address, override string, clientConfig comm.C
...
 	}
 	clientConfig.SecOpts = secOpts
 	return
 }
 
+// getClientAuthInfoFromEnv reads client tls key file and cert file and returns the bytes for the files
+func getClientAuthInfoFromEnv(prefix string) ([]byte, []byte, error) {
+	keyPEM, res := ioutil.ReadFile(config.GetPath(prefix + ".tls.clientKey.file"))
+	if res != nil {
+		return nil, nil, errors.WithMessage(res, fmt.Sprintf("unable to load %s.tls.clientKey.file", prefix))
```
wlahti: Nit (`WithMessagef` didn't exist when a lot of this code was written):
```suggestion
		return nil, nil, errors.WithMessagef(res, "unable to load %s.tls.clientKey.file", prefix)
```

internal/peer/common/common.go
```diff
@@ -255,25 +255,29 @@ func configFromEnv(prefix string) (address, override string, clientConfig comm.C
...
 		}
-		secOpts.Certificate = certPEM
 	}
 	clientConfig.SecOpts = secOpts
 	return
 }
 
+// getClientAuthInfoFromEnv reads client tls key file and cert file and returns the bytes for the files
+func getClientAuthInfoFromEnv(prefix string) ([]byte, []byte, error) {
+	keyPEM, res := ioutil.ReadFile(config.GetPath(prefix + ".tls.clientKey.file"))
```
wlahti: Nit: I know you didn't introduce this, but it seems odd to call the returned error `res` instead of `err`. Feels like we should go ahead and update this for consistency with the rest of the codebase.

wlahti (commented):
A few small nits that should be quick/easy to address. 

One thing that you didn't introduce to these tests that I think can be split out into a follow-up JIRA is to stop using cert fixtures and instead use `tlsgen` ([here](https://github.com/hyperledger/fabric/blob/master/common/crypto/tlsgen/ca.go)) to generate the certs/keys and write them to a temp folder.

wlahti (approved):
Thanks, Wenjian! Looks good to me (I also manually verified the behavior before/after the fix just to be sure). I'll open a follow-up task to move away from the cert fixtures in these tests.

PR #2174, opened 2020-12-01
Remove string conversion and add error judgment

mask-pp:
Remove string conversion and remove err return

### DIFF
--- a/msp/cache/cache.go
+++ b/msp/cache/cache.go
@@ -83,14 +83,16 @@ func (c *cachedMSP) DeserializeIdentity(serializedIdentity []byte) (msp.Identity
 }
 
 func (c *cachedMSP) Setup(config *pmsp.MSPConfig) error {
-	c.cleanCache()
+	if err := c.cleanCache(); err != nil {
+		return err
+	}
 
 	return c.MSP.Setup(config)
 }
 
 func (c *cachedMSP) Validate(id msp.Identity) error {
 	identifier := id.GetIdentifier()
-	key := string(identifier.Mspid + ":" + identifier.Id)
+	key := identifier.Mspid + ":" + identifier.Id
 
 	_, ok := c.validateIdentityCache.get(key)
 	if ok {
@@ -108,7 +110,7 @@ func (c *cachedMSP) Validate(id msp.Identity) error {
 
 func (c *cachedMSP) SatisfiesPrincipal(id msp.Identity, principal *pmsp.MSPPrincipal) error {
 	identifier := id.GetIdentifier()
-	identityKey := string(identifier.Mspid + ":" + identifier.Id)
+	identityKey := identifier.Mspid + ":" + identifier.Id
 	principalKey := string(principal.PrincipalClassification) + string(principal.Principal)
 	key := identityKey + principalKey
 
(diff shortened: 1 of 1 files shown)


wlahti:
Hi, thanks for contributing! 

Without an associated JIRA item, it's unclear what the motivation behind this change is. I notice you named your branch `bug fix` so I assume you noticed something unexpected related to this code? In general, it's best to open a JIRA item (https://jira.hyperledger.org/) and describe the issue you're seeing to help others understand the motivation behind your code changes. We also like to have unit tests included for most code changes, so try your best to figure out a test case that can be added to verify the new behavior. See the [contribution guide](https://hyperledger-fabric.readthedocs.io/en/release-2.2/CONTRIBUTING.html?highlight=contribution#contribution-guide) for more details on the general process for Hyperledger Fabric PRs.

msp/cache/cache.go
```diff
@@ -83,14 +83,16 @@ func (c *cachedMSP) DeserializeIdentity(serializedIdentity []byte) (msp.Identity
 }
 
 func (c *cachedMSP) Setup(config *pmsp.MSPConfig) error {
-	c.cleanCache()
+	if err := c.cleanCache(); err != nil {
```
yacovm: instead of doing this maybe simply make the cleanCache method not return an error? I don't see how it can fail.

mask-pp: Yes this is redundant

guoger:
@wlahti i don't think this PR is addressing any bug - it's more of style change or code cleaning. As I'm looking into our contribution guideline, it doesn't really describe how-to for such changes. I agree that the original branch is somehow misleading but i think we could still merge this one as is.

@mask-pp Fabric does expect JIRA to be open and used to track issues

guoger (approved):

lindluni:
> @wlahti i don't think this PR is addressing any bug - it's more of style change or code cleaning. As I'm looking into our contribution guideline, it doesn't really describe how-to for such changes. I agree that the original branch is somehow misleading but i think we could still merge this one as is.
> 
> @mask-pp Fabric does expect JIRA to be open and used to track issues

@guoger When we moved to GitHub we amended the contribution guide to allow for non-Jira contributions (it is in there, I just can't pull a link to it on my phone). Authors are asked to either open a Jira to link against, or to provide details in the commit message about what the commit is trying to achieve. We also ask that PRs be squashed into the smallest set of logical commits.

For a change as small as this, a descriptive commit title or commit message would suffice (no need for a Jira), but the two commits should be squashed into one before merge.

Given that the two commits modify the same lines of code in different ways, please squash the commits and update the single commit message to provide a descriptive reason for the change and we will get this merged.

guoger:
> Given that the two commits modify the same lines of code in different ways, please squash the commits and update the single commit message to provide a descriptive reason for the change and we will get this merged.

i failed to notice this :) @mask-pp please do a squash as instructed by Brett

guoger (changes requested):
Looks good, commits need to be squashed though, thx

mask-pp (commented):
Remove unnecessary type conversion and error return.

guoger (approved):

PR #2182, opened 2020-12-02
Use Deliver to ensure a leader has been elected after removing the system channel

wlahti:
#### Type of change

- Test update

#### Description

Avoid flakiness in this channel participation integration test by polling on deliver for each active orderer for the application channel after removing the system channel to ensure a leader has been elected before proceeding to submit transactions.

#### Related issues

[FAB-18362](https://jira.hyperledger.org/browse/FAB-18362)

### DIFF
--- a/integration/raft/channel_participation_test.go
+++ b/integration/raft/channel_participation_test.go
@@ -1215,6 +1215,30 @@ var _ = Describe("ChannelParticipation", func() {
 				cl := channelparticipation.List(network, o)
 				channelparticipation.ChannelListMatcher(cl, []string{"testchannel"})
 			}
+			cl := channelparticipation.List(network, orderer3)
+			channelparticipation.ChannelListMatcher(cl, nil)
+
+			By("fetching a block from each orderer to ensure a leader has been elected for the existing application channel")
+			for _, o := range orderers1and2 {
+				FetchBlock(network, o, 0, "testchannel")
+			}
+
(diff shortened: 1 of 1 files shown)


integration/raft/channel_participation_test.go
```diff
@@ -1302,6 +1309,14 @@ func submitTxn(o *nwo.Orderer, env *common.Envelope, n *nwo.Network, expectedCha
...
 }
 
+func deliverBlock(o *nwo.Orderer, n *nwo.Network, channel string) {
+	env := CreateDeliverEnvelope(n, o, 0, channel)
+	Expect(env).NotTo(BeNil())
+
+	By("fetching block from " + o.Name)
+	Eventually(deliverFunc(n, o, env), n.EventuallyTimeout, time.Second).ShouldNot(BeNil())
+}
+
```
guoger: maybe we could reuse `FetchBlock` from the same package? https://github.com/hyperledger/fabric/blob/25457a3ce5b2b9fa5e6908236cefbbb7cbb66670/integration/raft/client.go#L19

guoger (changes requested):

integration/raft/channel_participation_test.go
wlahti: Ah, thanks! Somehow kept overlooking that that function existed. Done.

guoger (approved):
LGTM

PR #2183, opened 2020-12-02
[FAB-17039] Skip retrieving pvtdata from transient store if txid is empty

wenjianqiao:

#### Type of change
- Bug fix

#### Description
Skip calling fromTransientStore when txid is empty. This could happen for a reconciliation request.

#### Related issues
https://jira.hyperledger.org/browse/FAB-17039

### DIFF
--- a/gossip/privdata/dataretriever.go
+++ b/gossip/privdata/dataretriever.go
@@ -55,14 +55,27 @@ func (dr *dataRetriever) CollectionRWSet(digests []*protosgossip.PvtDataDigest, 
 		// if there is an error getting info from the ledger, we need to try to read from transient store
 		return nil, false, errors.Wrap(err, "wasn't able to read ledger height")
 	}
-	if height <= blockNum {
+
+	// The condition may be true for either commit or reconciliation case when another peer sends a request to retrieve private data.
+	// For the commit case, get the private data from the transient store because the block has not been committed.
+	// For the reconciliation case, this peer is further behind the ledger height than the peer that requested for the private data.
+	// In this case, the ledger does not have the requested private data. Also, the data cannot be queried in the transient store,
+	// as the txID in the digest will be missing.
+	if height <= blockNum { // Check whenever current ledger height is equal or below block sequence num.
 		dr.logger.Debug("Current ledger height ", height, "is below requested block sequence number",
 			blockNum, "retrieving private data from transient store")
-	}
 
-	if height <= blockNum { // Check whenever current ledger height is equal or below block sequence num.
 		results := make(Dig2PvtRWSetWithConfig)
 		for _, dig := range digests {
+			// skip retrieving from transient store if txid is not available
+			if dig.TxId == "" {
+				dr.logger.Infof("Skip querying transient store for chaincode %s, collection name %s, block number %d, sequence in block %d, "+
+					"as the txid is missing, perhaps because it is a reconciliation request",
+					dig.Namespace, dig.Collection, blockNum, dig.SeqInBlock)
+
+				continue
+			}
+
 			filter := map[string]ledger.PvtCollFilter{
 				dig.Namespace: map[string]bool{
 					dig.Collection: true,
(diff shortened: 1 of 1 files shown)


gossip/privdata/dataretriever.go
```diff
@@ -55,14 +55,25 @@ func (dr *dataRetriever) CollectionRWSet(digests []*protosgossip.PvtDataDigest,
...
+	if height <= blockNum { // Check whenever current ledger height is equal or below block sequence num.
 		dr.logger.Debug("Current ledger height ", height, "is below requested block sequence number",
 			blockNum, "retrieving private data from transient store")
-	}
 
-	if height <= blockNum { // Check whenever current ledger height is equal or below block sequence num.
 		results := make(Dig2PvtRWSetWithConfig)
 		for _, dig := range digests {
+			// It is the reconciliation case since TxId is empty, skip retrieving private data from transient store.
+			if dig.TxId == "" {
```
ale-linux, line 71: Can we formulate the condition differently? These hidden meanings attached to special values of fields (0, -1 for integers, zero-length for strings etc.) tend to be quickly forgotten, and then it's hard to find out what goes wrong in the future. It would be better if we could find a way of making it explicit that it's reconciliation without resorting to zero-length string check.

wenjianqiao: Thank you Ale. Agree it would be nice to have a way to make it explicit. Ideally, the puller request message should contain a field to explicitly indicate the request type (e.g., reconciliation vs. commit). However, the protobuf message does not have such a field. Adding a new field will introduce backward compatibility issue where some peers are in new versions and others may be in old version. Therefore, this PR adds a new func isReconciliationRequest to help clarify the use case for reconciliation requests.

If you still think we need a field to explicitly indicate the request type, we can open a new jira since it is out of the scope of this PR for bug fix.

ale-linux: Thanks @wenjianqiao , I think this is already better as it makes the fact at least explicit (with tests attached). LGTM

gossip/privdata/dataretriever.go
```diff
@@ -55,14 +55,25 @@ func (dr *dataRetriever) CollectionRWSet(digests []*protosgossip.PvtDataDigest,
...
+	if height <= blockNum { // Check whenever current ledger height is equal or below block sequence num.
 		dr.logger.Debug("Current ledger height ", height, "is below requested block sequence number",
 			blockNum, "retrieving private data from transient store")
-	}
 
-	if height <= blockNum { // Check whenever current ledger height is equal or below block sequence num.
 		results := make(Dig2PvtRWSetWithConfig)
 		for _, dig := range digests {
+			// skip retrieving from transient store for reconciliation requests
+			if isReconciliationRequest(dig) {
```
manish-sethi: Though this is correct that txID is not included in the case of reconciliation at the caller side, but drawing this conclusion at this function which the the sender side, breaks the principal of isolation. I find previous patch more suitable here that if caller does not supply txID (for whatever reason), skip looking in the transient store. We can certainly write a comment when this can happen but the caller code is free to change in general. For instance, caller is free not to send txID even in the case of commit path. In fact, commit path/ reconciliation terms are meaningful only at the caller side. This place is just getting asked for data either by txID, or by blockNum:TxNum (or both).

Also, having this if condition in a loop, may give someone a wrong impression that the requests for reconciliation and commit path gets combined on the caller side.

@ale-linux - second thought?

gossip/privdata/dataretriever.go
```diff
@@ -55,14 +55,25 @@ func (dr *dataRetriever) CollectionRWSet(digests []*protosgossip.PvtDataDigest,
...
+				dr.logger.Warningf("No private data was found for chaincode %s, collection name %s, sequence in block %d "+
```
manish-sethi: This should be an `info` as there is nothing wrong happening here.

gossip/privdata/dataretriever.go
ale-linux: I trust your judgement since you certainly understand the pvtdata code better than I do. I just don't like these less-than-transparent conditions where the behaviour of a function is determined by special values of a field that otherwise plays a completely different role in the system (case in point, the txid's principal meaning is to avoid replay attacks and provide a unique handle to a transaction, and it is here used to determine whether data should be retrieved from storage).

These non-obvious relationships (here between txis(=uniquness) and behaviour of pvt data retrieval) tend to be quickly forgotten and make the code hard to maintain.

That said, if you believe this to be the right approach, I won't stand in the way.

manish-sethi (approved):

lindluni:

btl5037:

btl5037:

lindluni:

PR #2185, opened 2020-12-02
[WIP][FAB-18363] Update Protobuf to 1.4.3

lindluni:
Need to sort through the `orderer/consensus/etcdraft` issues yet and anything that comes out of integration tests as I haven't run them locally yet (doing it now).

Also: Yes, I will squash once I'm done


btl5037:
Need to sort through the `orderer/consensus/etcdraft` issues yet and anything that comes out of integration tests as I haven't run them locally yet (doing it now).

lindluni:
Likely blocked due to this: https://github.com/etcd-io/etcd/pull/12398#discussion_r505322880

More info here: https://github.com/etcd-io/etcd/issues/12124

lindluni:
It seems there are two issues at play here:

1. `etcd` doesn't support grpc > 1.29
2. `etcd` still uses `gogo/protobuf` which doesn't support the 1.4 protobuf standard
3. `etcd` needs to release the `3.5-alpha.0` build to deal with module issues they have (likely won't happen before the end of the year)

This blocks the upgrade of many of our dependencies unless we do some hacky things in our `go.mod` to pin our deps to a version that `etcd` doesn't use, and then do a replace on their versions to something they support.

PR #2194, opened 2020-12-03
More doc edits for system channel removal

joealewine:
Change-Id: I1c7f546ec4d2ba2cbd1470f1886c7c554f38c001

#### Type of change

- Documentation update

#### Description

More edits to docs to either remove references to the system channel or to create links back to previous doc versions, where necessary.


docs/source/upgrade_to_newest_version.md
```diff
@@ -4,7 +4,16 @@ In this topic we'll cover recommendations for upgrading to the newest release fr
...
+The v2.1 and v2.2 releases of Fabric are stabilization releases, featuring bug fixes and other forms of code hardening. As such there are no particular considerations needed for upgrade, and no new capability levels requiring particular image versions or channel configuration updates.
+
+### Upgrading from 2.2 to 2.3
+
+The v2.3 release of Fabric includes two main new features:
+
+1. The ability to take snapshots of the ledgers on peers (and to bootstrap a new peer from a snapshot). For more information, check out [Taking ledger snapshots and using them to join channels](./peer_ledger_snapshot.html).
+2. Channels can now be created without first creating a system channel. For more information, check out [Creating a channel without a system channel](./create_channel/create_channel_participation.html).
+
+Neither of these features require channel updates to capability in order to function. However, you will need to upgrade to v2.3 to take advantage of both features and, in the case of the new channel creation process, need to migrate away from the system channel.
```
pamandrejko: > "...in the case of the new channel creation process, need to migrate away from the system channel."...

When I read this I ask myself, where can I find how to do that? I know it is covered in Creating a channel without a system channel, so maybe we should tell them that.

"However, you will need to upgrade to v2.3 to take advantage of both features and, in the case of the new channel creation process, need to migrate away from the system channel which is covered in the [Creating a channel without a system channel](./create_channel/create_channel_participation.html) tutorial."

What do you think?

pamandrejko (commented):
Good progress. A few nits.

docs/source/upgrade_to_newest_version.md
joealewine: I try to avoid linking to the same doc back to back like that and instead create a structure where the link only has to appear once, but I can make an exception here, as there's no good way to avoid it that I can find.

wlahti (approved):
Looks good to me!

pamandrejko (approved):

pamandrejko:

PR #2196, opened 2020-12-06
FAB-16233 fail test with names instead of brutal panic

guoger:
It's better to call t.FailNow() or t.Fatal(), instead of panic()
in tests, so we could have more info, i.e. which exact test has failed.

This is to aid debugging flaky UT described in FAB-16233, and should be
a good practice overall.

### DIFF
--- a/orderer/common/cluster/deliver_test.go
+++ b/orderer/common/cluster/deliver_test.go
@@ -176,9 +176,7 @@ func (ds *deliverServer) Deliver(stream orderer.AtomicBroadcast_DeliverServer) e
 		return err
 	}
 	seekInfo, channel, err := readSeekEnvelope(stream)
-	if err != nil {
-		panic(err)
-	}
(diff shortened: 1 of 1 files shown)


orderer/common/cluster/deliver_test.go
```diff
@@ -187,7 +185,7 @@ func (ds *deliverServer) Deliver(stream orderer.AtomicBroadcast_DeliverServer) e
 
 	select {
 	case <-timer.C:
-		panic("timed out waiting for seek assertions to receive a value")
+		ds.t.Fatalf("timed out waiting for seek assertions to receive a value\n")
```
ale-linux, line 188: why the newline here and not on the one at line 205?

guoger: good catch, done

ale-linux (approved):

PR #2197, opened 2020-12-07
Remove Short Names and Replace With Full Path in Fabric

jcastrence:
#### Type of change

- Documentation update

#### Description

- Short names are a crutch that can prevent someone from learning how to deploy their own chaincode. Short names have been replaced with their full path name.

### DIFF
--- a/docs/source/channel_update_tutorial.rst
+++ b/docs/source/channel_update_tutorial.rst
@@ -571,7 +571,7 @@ use the ``test-network`` script to deploy the Basic chaincode:
 .. code:: bash
 
   cd fabric-samples/test-network
-  ./network.sh deployCC -ccn basic -ccl go
+  ./network.sh deployCC -ccn basic -ccp ../asset-transfer-basic/chaincode-go/ -ccl go
 
 The script will install the Basic chaincode on the Org1 and Org2 peers, approve
 the chaincode definition for Org1 and Org2, and then commit the chaincode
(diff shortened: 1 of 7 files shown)


lindluni (commented):
TRUNCATED

lindluni (dismissed):
I know you didn't introduce the original shortnames, and that we've already merged it in Samples, but, I really don't like the use of `ccp` here. In Fabric `ccp` is our abbreviation for the extremely ubiquitous `Common Connection Profile`. It almost seems like it's the output or input location of the connection profile. Saving a single character to create these abbreviations doesn't really seem to help. `-name`, `-path`, `-lang`, `-policy` (for endorsement policy) seem like better choices, especially since they align with the actual Peer CLI. I don't see a reason to shorten them, but if we were going to, I'd be in favor of the same shortnames as the Peer CLI has, `-n`, `-p`, `-l`, `-P`.

For users just getting started, conflating one our most common abbreviations with something else doesn't seem helpful.

Not sure how other's feel, but putting this out there.

denyeart:
@btl5037 I generally agree, but there are some considerations, e.g.
`--sequence` does not have a one letter abbreviation, if it did, you'd expect it to be `-s`. However `-s` is already used in network.sh for the state database choice. I suspect the creator of this script used `-cc*` to namepsace chaincode flags so they wouldn't conflict with other flags. We can continue the dialog, but I'd suggest it is outside the scope of this change so we shouldn't let it block this PR.

denyeart (approved):

lindluni:

lindluni:

PR #2209, opened 2020-12-11
[FAB-18378] Log warning when peer is lagging behind and cannot catch up

yacovm:
Fabric 2.2 made gossip state transfer be disabled by default.

When a peer is a gossip follower, it does not pull blocks from the ordering service,
and relies on the leader peer of its organization to forward blocks directly or indirectly.

However, if a peer misses a block and the rest of the peers in its organization advance too much,
it will never catch up because its state transfer is disabled.

This commit makes the peer log a warning if it is lagging behind, and its configuration
might prevent it from obtaining blocks via state transfer and the orderer.

Change-Id: Ie7bc67cb13cb2d8e7ffab638a49e0365535865c2

### DIFF
--- a/gossip/state/state.go
+++ b/gossip/state/state.go
@@ -41,9 +41,9 @@ type GossipStateProvider interface {
 }
 
 const (
-	defAntiEntropyBatchSize = 10
-
-	defMaxBlockDistance = 20
+	stragglerWarningThreshold = 100
+	defAntiEntropyBatchSize   = 10
+	defMaxBlockDistance       = 20
 
 	blocking    = true
 	nonBlocking = false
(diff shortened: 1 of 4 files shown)


gossip/state/state.go
```diff
@@ -761,6 +761,10 @@ func (s *GossipStateProviderImpl) addPayload(payload *proto.Payload, blockingMod
 	}
 
 	if !blockingMode && payload.SeqNum-height >= uint64(s.config.StateBlockBufferSize) {
+		if s.straggler(height, payload) {
+			s.logger.Warningf("Detected a distant block sequence, we are lagging behind and state transfer is disabled in configuration " +
```
mastersingh24: It is possible to log the channel as well as how far behind the peer actually is?

gossip/state/state.go
```diff
@@ -761,6 +761,11 @@ func (s *GossipStateProviderImpl) addPayload(payload *proto.Payload, blockingMod
 	}
 
 	if !blockingMode && payload.SeqNum-height >= uint64(s.config.StateBlockBufferSize) {
+		if s.straggler(height, payload) {
+			s.logger.Warningf("[%s] Detected a distant block sequence (%d), we are lagging behind (%d) and state transfer is disabled in configuration "+
```
denyeart: Getting better. But while you know what a 'block sequence' is in this context, an admin or SRE looking at the log won't. Could you put it into terms that the admin or SRE will understand? E.g. current peer block height.

yacovm: Can you perhaps spell it out for me? What exactly do you want me to write?

denyeart: Something like this:

`[mychannel] Current block height [100] is lagging too far behind other peers at height [500] to be able to get blocks via peer-to-peer gossip. Consider setting the peer to be a leader (peer.gossip.orgLeader = true) so that it receives blocks from the ordering service. Alternatively enable state transfer (peer.gossip.state.enabled = true) so that the peer pulls older blocks from other peers.`

I may not have gotten it exactly correct since I don't know the logic in this part of the code...

denyeart (approved):
Good message :-)

denyeart:

denyeart:

PR #2226, opened 2020-12-17
[FAB-18382] Fix staticcheck issues in gossip/privdata

Taction:
ignore staticcheck ST1019 because it is used intentionally.
break statement is ineffective and is not mean to break out of the outer loop, just delete it

FAB-18382 #done

Signed-off-by: Taction <zchao9100@gmail.com>

### DIFF
--- a/gossip/privdata/reconcile.go
+++ b/gossip/privdata/reconcile.go
@@ -120,7 +120,6 @@ func (r *Reconciler) run() {
 			r.logger.Debug("Start reconcile missing private info")
 			if err := r.reconcile(); err != nil {
 				r.logger.Error("Failed to reconcile missing private info, error: ", err.Error())
-				break
 			}
 		}
 	}


gossip/privdata/reconcile.go
```diff
@@ -120,7 +120,6 @@ func (r *Reconciler) run() {
 			r.logger.Debug("Start reconcile missing private info")
 			if err := r.reconcile(); err != nil {
 				r.logger.Error("Failed to reconcile missing private info, error: ", err.Error())
-				break
```
ale-linux, line 123: If the (misguided in this case) intent was to break out of the for loop, shouldn't we add the code to actually do so rather than just removing this ineffectual break? @yacovm

yacovm: I [don't think](https://github.com/hyperledger/fabric/commit/fefb5a772e5607363df07516827a64e8e966563a) the intent was to break out of the for.

Taction: I should have provided more information to make it easier to review.The for loop is [only start once](https://github.com/hyperledger/fabric/blob/master/gossip/privdata/reconcile.go#L108) when [InitializeChannel](https://github.com/hyperledger/fabric/blob/master/gossip/service/gossip_service.go#L376),so i think an error should not stop the entire function (for the next try or get other private data) . What's more the commit yacovm provide show more clearly that the purpose is to prevent the execution of the following statement.But now it is the last statement.

C0rWin: in this case intent was not to break out of the for loop, but rather from select, cause reconciliation is a background process triggered periodically. As @Taction pointed out break Is the last statement here hence could be removed as doesn't add anything to the code flow.

C0rWin (approved):

PR #2238, opened 2021-01-04
[FAB-18357] Generate certs for peer CLI tests

stephyee:
#### Type of change

- Test update

#### Description
Generate certificates for peer CLI tests instead of using cert fixtures

#### Related issues

[FAB-18357
](https://jira.hyperledger.org/browse/FAB-18357
)


wlahti (approved):
Thanks, Tiffany!!

PR #2258, opened 2021-01-11
Add note to AnchorPeers in sampleconfig/configtx.yaml

wlahti:
#### Type of change

- Documentation update

#### Description

This section was only used by the deprecated `configtxgen --outputAnchorsPeerUpdate` command. As we move to the channel participation flow without a system channel, we should stop recommending this value be set in configtx.yaml as it doesn't make sense to include anchor peers in the genesis block for a channel. The anchor peers should be set per org using the channel config update process.

#### Related issues

[FAB-18381](https://jira.hyperledger.org/browse/FAB-18381)

### DIFF
--- a/sampleconfig/configtx.yaml
+++ b/sampleconfig/configtx.yaml
@@ -66,8 +66,12 @@ Organizations:
             - "127.0.0.1:7050"
 
         # AnchorPeers defines the location of peers which can be used for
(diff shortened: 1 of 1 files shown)


ale-linux (dismissed):

wlahti:
Oops, there's one test that needs to be updated. I'll push an update shortly.

ale-linux (approved):

PR #2265, opened 2021-01-12
Remove GetSigningIdentity from MSP interface

sykesm:
There were no calls to the method in production code and the two production implementations in fabric (bccspmsp and idemixmsp) returned errors indicating the function was not implemented.

Messages in test assertions were also updated to reflect the name of the function that was actually being used in most cases (`GetDefaultSigningIdentity`).


jyellick (approved):
Always happy to see some dead code removed!

PR #2287, opened 2021-01-14
Orderer deployment tutorial update

pamandrejko:
Explicitly call out the orderer.yaml settings that are required for the `osnadmin` command to work.

#### Type of change

- Documentation update

#### Description

Because this tutorial assumes that a system channel genesis block will not be used when bootstrapping the orderer, I  added the set of Admin.* parameters in the orderer.yaml that need to be configured in order to use the osnadmin command to create an application channel.


pamandrejko:
fix commits

PR #2291, opened 2021-01-16
Update Brett Logan's Information

lindluni:


sykesm (commented):
You have an entry under release manager as well.

lindluni:
> You have an entry under release manager as well.

Good catch, updated

sykesm (approved):

PR #2314, opened 2021-01-27
[FABGW-1] Add gateway configuration

jt-nti:
Just includes a feature flag for now

### DIFF
--- a/internal/pkg/gateway/config.go
+++ b/internal/pkg/gateway/config.go
@@ -1,0 +1,20 @@
+/*
+Copyright IBM Corp. All Rights Reserved.
+
+SPDX-License-Identifier: Apache-2.0
+*/
+
+package gateway
+
+// GatewayOptions is used to configure the gateway settings
+type Options struct {
+	// GatewayEnabled is used to enable the gateway service
+	Enabled bool
+}
+
+// DefaultOptions gets the default Gateway configuration Options
+func DefaultOptions() Options {
+	return Options{
+		Enabled: false,
+	}
+}
--- a/core/peer/config.go
+++ b/core/peer/config.go
@@ -30,6 +30,7 @@ import (
 
 	"github.com/hyperledger/fabric/core/config"
 	"github.com/hyperledger/fabric/internal/pkg/comm"
+	"github.com/hyperledger/fabric/internal/pkg/gateway"
 	"github.com/pkg/errors"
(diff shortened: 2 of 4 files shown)


andrew-coleman (dismissed):

internal/pkg/gateway/config.go
```diff
@@ -0,0 +1,20 @@
+/*
+Copyright IBM Corp. All Rights Reserved.
+
+SPDX-License-Identifier: Apache-2.0
+*/
+
+package gateway
+
+// Configuration defaults
+var (
```
sykesm: Please replace the package level var with a function that returns an instance containing the default options.

sykesm (changes requested):
We have too much package level state in Fabric and we want to avoid adding more. Please change the mechanism used to acquire a default configuration to an exported function that returns an instance instead of using an exported var.

Thanks.

internal/pkg/gateway/config.go
```diff
@@ -0,0 +1,21 @@
...
+
+// GatewayOptions is used to configure the gateway settings
+type GatewayOptions struct {
+	// GatewayEnabled is used to enable the gateway service
+	GatewayEnabled bool
+}
+
+// GetDefaultConfig gets a GatewayOptions with the default
+// Gateway configuration
+func GetDefaultConfig() GatewayOptions {
```
jt-nti: Is something like this ok @sykesm ?

sykesm: Yes. It addresses my concern about package state and is fine.

If you want my really picky feedback,
- `gateway.GatewayOptions` is a stuttering name and could simply be `gateway.Options`
- accessors in go, in general, forego the `Get` prefix so this would be `func DefaultOptions() Options { ... }`. In context, callers would use `gateway.DefaultOptions()` to acquire the defaults.
- Prefixing all of the options with `Gateway` seems unnecessary. `gateway.GatewayOptions.GatewayEnabled` seems relatively repetitive.

sykesm (dismissed):
I'm okay with merging as-is but I'll wait to see if you want to do anything about the naming.

internal/pkg/gateway/config.go
jt-nti: Feedback is good :)
That naming looks much better, thanks

sykesm (approved):

PR #2318, opened 2021-01-27
Minor updates related to documentation for commands and more osnadmin updates

sykesm:
- Prevent osnadmin tests from exiting the test execution process when a bad flag is provided
- Update the help doc generation script to filter trailing spaces from command help
- Update the reference to the old `channel-id` that was included in the osnadmin command help

### DIFF
--- a/cmd/osnadmin/main.go
+++ b/cmd/osnadmin/main.go
@@ -52,7 +52,7 @@ func executeForArgs(args []string) (output string, exit int, err error) {
 	joinChannelID := join.Flag("channelID", "Channel ID").Short('c').Required().String()
 	configBlockPath := join.Flag("config-block", "Path to the file containing an up-to-date config block for the channel").Short('b').Required().String()
 
-	list := channel.Command("list", "List channel information for an Ordering Service Node (OSN). If the channel-id flag is set, more detailed information will be provided for that channel.")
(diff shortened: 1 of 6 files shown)


docs/source/commands/peerlifecycle.md
```diff
@@ -595,7 +595,7 @@ also outputs which organizations have approved the chaincode definition. If an
 organization has approved the chaincode definition specified in the command, the
 command will return a value of true. You can use this command to learn whether enough
 channel members have approved a chaincode definition to meet the
-`/Channel/Application/Endorsement` policy (a majority by default) before the
```
yacovm, line 598: how is this helpful or even related to the change set?

yacovm: Did you remove this with a batch search-replace?

sykesm: It's related because it's the output of the automated doc generation for command help. It's clear that commands have been getting updated without regenerating the docs. The regen of the docs was explicitly called out in the commit message.

~~In this case, commit 49ed882a28 changed the postscript without updating docs.~~
In this case, 5df63f1425 changed the generated doc instead of the postscript. That change was incorrect.

yacovm (approved):

pamandrejko:
@sykesm OK to mergifyio this back to the 2.3 branch?

sykesm:
> @sykesm OK to mergifyio this back to the 2.3 branch?

I think it's better to do it manually. I can do that if you want.

PR #2323, opened 2021-01-28
FABGW-1: Allow gateway direct access to discovery service

bestbeforetoday:
#### Type of change

- New feature

#### Description

Allow the embedded Gateway service to access capabilities of the peer's discovery service directly without requiring a client identity. This allows the embedded Gateway service to:
1. Identify active channel peers and their ledger heights
2. Obtain endorsement plans

### DIFF
--- a/internal/peer/node/start.go
+++ b/internal/peer/node/start.go
@@ -93,6 +93,7 @@ import (
 	peergossip "github.com/hyperledger/fabric/internal/peer/gossip"
 	"github.com/hyperledger/fabric/internal/peer/version"
 	"github.com/hyperledger/fabric/internal/pkg/comm"
+	gateway "github.com/hyperledger/fabric/internal/pkg/gateway/server"
 	"github.com/hyperledger/fabric/msp"
 	"github.com/hyperledger/fabric/msp/mgmt"
 	"github.com/hyperledger/fabric/protoutil"
@@ -802,8 +803,9 @@ func serve(args []string) error {
 		coreConfig.ValidatorPoolSize,
 	)
 
+	var discoveryService gateway.DiscoveryService
 	if coreConfig.DiscoveryEnabled {
-		registerDiscoveryService(
+		discoveryService = registerDiscoveryService(
 			coreConfig,
 			peerInstance,
 			peerServer,
@@ -818,7 +820,7 @@ func serve(args []string) error {
 	}
 
 	if coreConfig.GatewayOptions.Enabled {
-		logger.Info("Starting peer with Gateway enabled")
+		registerGateway(peerServer, discoveryService)
 	}
 
 	logger.Infof("Starting peer with ID=[%s], network ID=[%s], address=[%s]", coreConfig.PeerID, coreConfig.NetworkID, coreConfig.PeerAddress)
@@ -948,7 +950,7 @@ func registerDiscoveryService(
 	polMgr policies.ChannelPolicyManagerGetter,
 	metadataProvider *lifecycle.MetadataProvider,
 	gossipService *gossipservice.GossipService,
-) {
+) gateway.DiscoveryService {
 	mspID := coreConfig.LocalMSPID
 	localAccessPolicy := localPolicy(policydsl.SignedByAnyAdmin([]string{mspID}))
 	if coreConfig.DiscoveryOrgMembersAllowed {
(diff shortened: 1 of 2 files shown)


yacovm (changes requested):
It's better to just use the support directly https://github.com/hyperledger/fabric/blob/c583c29fef905875a0ab1122ddc8db46f2382170/discovery/support/support.go#L89 and not change discovery just to bypass access control.

discovery/service.go
```diff
@@ -36,12 +36,18 @@ type certHashExtractor func(ctx context.Context) []byte
 // dispatcher defines a function that dispatches a query
 type dispatcher func(q *discovery.Query) *discovery.QueryResult
 
+// Service describes service discovery methods directly accessible to other components
```
yacovm: Why do you need to define this? You can just export the method itself and define the interface where you need it. 
I mean this in general, in Go we define an interface where we need it and not where it is implemented.

bestbeforetoday: Noted and removed

yacovm (approved):

internal/peer/node/start.go
```diff
@@ -948,7 +950,7 @@ func registerDiscoveryService(
 	polMgr policies.ChannelPolicyManagerGetter,
 	metadataProvider *lifecycle.MetadataProvider,
 	gossipService *gossipservice.GossipService,
-) {
+) gateway.DiscoveryService {
```
sykesm, line 953: We should not be using a foreign interface type as a return value from what is logically a constructor.

sykesm (changes requested):
I'll fill in some comments and rationale in a moment.

internal/peer/node/start.go
```diff
@@ -980,6 +982,18 @@ func registerDiscoveryService(
 	}, support)
 	logger.Info("Discovery service activated")
 	discprotos.RegisterDiscoveryServer(peerServer.Server(), svc)
+	return svc
+}
+
+func registerGateway(peerServer *comm.GRPCServer, discoveryService gateway.DiscoveryService) {
+	if nil == discoveryService {
```
sykesm, line 989: We don't use yoda tests in idiomatic go.

internal/peer/node/start.go
sykesm: We have a common idiom in go: "accept interfaces, return structs". In this case, we should be returning the concrete type from the discovery package. That said, the discovery package is not idiomatic go and has exported a constructor that returns an unexported struct. This means that you, as the consumer, can't assign it to anything other than an interface you define.

The code in the discovery package should really be exporting `service` as `Service`.  The unexported fields remain inaccessible but this code can now use the appropriate concrete type.

Later, when you go to register the implementation, your gateway interface would be used to ensure that what the caller provides satisfies your contract.

With that...
1. we should have a change before this that exports `service` as `Service` from the discovery layer
2. we should avoid explicitly binding to gateway.DiscoveryService interface within this package if possible

Finally, a consistent problem within fabric is that we define our interfaces before the implementation. This is backwards for go. Interfaces should be defined when we need to enforce a contract on a type we accept instead of defining the contract we want to implement.

sykesm (changes requested):

internal/peer/node/start.go
```diff
@@ -802,8 +803,9 @@ func serve(args []string) error {
 		coreConfig.ValidatorPoolSize,
 	)
 
+	var discoveryService gateway.DiscoveryService
 	if coreConfig.DiscoveryEnabled {
-		registerDiscoveryService(
+		discoveryService = registerDiscoveryService(
```
sykesm, line 808: I'm also not a fan keeping the name `registerXXX` when we're relying on a side-effect. This should be broken into a function that creates the service and then one that registers it.

internal/peer/node/start.go
bestbeforetoday: Do you have any rationale or pointers to go style guides in this area? Testing the other way round reads more nicely but increases the risk of accidentally assigning nil to the variable

sykesm: This isn't C or Java. The condition`if foo = nil` will not even compile in Go. https://play.golang.org/p/B2qhmJMw6Hm

https://github.com/hyperledger/fabric/blob/master/docs/source/style-guides/go-style.rst points to a number of resources that describe what we consider good practices. If you haven't walked through those resources, it's probably worth your time.

internal/peer/node/start.go
sykesm: Just to follow-up on the comment about interfaces, please see https://github.com/golang/go/wiki/CodeReviewComments#interfaces (referenced in our style guide) for details.

internal/peer/node/start.go
sykesm: It may also be worthwhile to look through the list of default `staticcheck` rules.
https://staticcheck.io/docs/checks#ST1017 is the check for yoda conditions.

bestbeforetoday:
@sykesm Feedback noted. It doesn't really leave anything meaningful to be delivered in this PR since the only change would be to rename `discovery.service` to `discovery.Service`. I'll close this PR and leave Andy to do that in a change he's working on that will deliver some actual Gateway code.

internal/peer/node/start.go
bestbeforetoday: Perhaps that should be added to the list of repository style guides since this idiom doesn't appear to be mentioned in any of the others. Or better still, include staticcheck as part of the build if its rules need to be followed.

sykesm: > include staticcheck as part of the build if its rules need to be followed.

I'd love to see that. Just need to evaluate the ~2000 issues that are currently flagged in master.

PR #2324, opened 2021-01-28
Small refactors to setup for #2323

sykesm:
Addressing my own comments from #2323, export `discovery.Service` and separate the creation and registration of the service.


jyellick (approved):
LGTM

PR #2325, opened 2021-01-29
Add staticcheck static analysis and style checks in CI

sykesm:
Based on some interactions in another PR, these commits fix issues flagged by staticcheck and enable these checks in CI.

### DIFF
--- a/bccsp/pkcs11/pkcs11.go
+++ b/bccsp/pkcs11/pkcs11.go
@@ -29,7 +29,7 @@ import (
 )
 
 var logger = flogging.MustGetLogger("bccsp_p11")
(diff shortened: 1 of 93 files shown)


jyellick (approved):
Big but rote -- LGTM once CI starts behaving

lindluni:
I need to fix CI, Dave and I reassigned some of the service connections yesterday and they were associated by their SHA, not by their name, so AZP can't find the association anymore

denyeart:
BTW CI passed after Matt's re-trigger:
https://dev.azure.com/Hyperledger/Fabric/_build/results?buildId=29234&view=results

It must just be having trouble posting back.

PR #2326, opened 2021-01-29
Test network does not work with latest Docker

pamandrejko:

We've noticed problems with the test network when the latest Docker version 3.1.0  is installed. The workaround for now is to uninstall and reinstall a lower version of Docker. I've successfully used it with Docker v2.5.0.1.
- Add a mention of this in the existing prereqs topic
- Mentioned this at the very top of the Test Network tutorial
- Added a troubleshooting topic to the test network tutorial in case users experience this problem.

#### Type of change

- Documentation update

### DIFF
--- a/docs/source/prereqs.rst
+++ b/docs/source/prereqs.rst
@@ -34,6 +34,8 @@ operating, or developing on (or for), Hyperledger Fabric:
     Toolbox <https://docs.docker.com/toolbox/toolbox_install_windows/>`__ -
     again, Docker version Docker 17.06.2-ce or greater is required.
 
+The Fabric sample test network has been successfully 
+verified with Docker 2.5.0.1. Higher versions may not work at this time.
 You can check the version of Docker you have installed with the following
 command from a terminal prompt:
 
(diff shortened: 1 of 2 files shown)


denali49 (approved):
LGTM, thanks Pam!

denyeart (approved):
We should still work to identify root cause of the Docker issue, but I agree we should get this into docs in the meanwhile.

denyeart:
Sounds like Pam will make one more update, so let's hold off on merge until the next push.

pamandrejko:
Added similar update to the test network sample tutorial.

nikhil550:
I have encountered the 3.1.0 error myself. If I deploy a smart contract, it seems like the busybox container hangs when I try to bring down the network: 
```
CONTAINER ID   IMAGE     COMMAND                  CREATED         STATUS         PORTS     NAMES
d35a71258440   busybox   "sh -c 'cd /data && …"   8 seconds ago   Up 7 seconds             gifted_curran
```

pamandrejko:

PR #2327, opened 2021-01-29
FABGW-1 Move gateway server into the peer

andrew-coleman:
Part of Gateway feature work, embedding into peer
https://hyperledger.github.io/fabric-rfcs/text/0000-fabric-gateway.html

This is WIP, moving the current development snapshot of the
gateway server from a standalone process to an embedded
component of the peer.
This commit adds the API and minimal implementation.
It can be switch on using the peer.gateway.enabled config option which defaults to false.


internal/pkg/gateway/registry/registry.go
```diff
@@ -0,0 +1,97 @@
...
+	"google.golang.org/grpc/credentials"
+)
+
+// Config represents the startup configuration of the gateway
+type Config interface {
+}
+
+type registry struct {
+	localEndorser peer.EndorserClient
+	discovery     *discovery.Service
```
yacovm: you should define an interface of what you need from discovery and just mock it.

internal/pkg/gateway/registry/registry.go
```diff
@@ -0,0 +1,97 @@
...
+}
+
+var logger = flogging.MustGetLogger("gateway.registry")
```
yacovm: It's better to use an instance level logger

internal/pkg/gateway/server/mocks/mock_registry.go
```diff
@@ -0,0 +1,167 @@
+/*
+Copyright 2020 IBM All Rights Reserved.
```
yacovm: It's better to use auto-generated mocks. Search for "counterfeiter" or "mockery" (my favorite is the latter but most people in the project wrongly prefer the former) in the code base

internal/pkg/gateway/registry/registry.go
```diff
@@ -0,0 +1,97 @@
...
+
+func NewRegistry(discovery *discovery.Service) (*registry, error) {
```
sykesm: I don't get it. :) I commented just yesterday on how an exported constructor returning an unexported type makes little to no sense.

internal/pkg/gateway/registry/registry.go
```diff
@@ -0,0 +1,97 @@
...
+	"github.com/hyperledger/fabric-protos-go/peer"
+	"github.com/hyperledger/fabric/common/flogging"
+	pp "github.com/hyperledger/fabric/core/peer"
+	"github.com/hyperledger/fabric/discovery"
+	"google.golang.org/grpc"
+	"google.golang.org/grpc/credentials"
+)
+
+// Config represents the startup configuration of the gateway
+type Config interface {
```
sykesm: What is this providing? Why would it be an interface?

go.mod
```diff
@@ -39,7 +39,7 @@ require (
 	github.com/hyperledger/fabric-chaincode-go v0.0.0-20200128192331-2d899240a7ed
 	github.com/hyperledger/fabric-config v0.0.7
 	github.com/hyperledger/fabric-lib-go v1.0.0
-	github.com/hyperledger/fabric-protos-go v0.0.0-20201028172056-a3136dde2354
+	github.com/hyperledger/fabric-protos-go v0.0.0-20210127161553-4f432a78f286
```
sykesm, line 33: nit: can we bump deps in a separate commit?

sykesm (changes requested):
There's a lot going on here and there's a lot that's incomplete. It may make sense to break your work into separate commits and, perhaps, provide a task list in the PR that describes where things are.

Our master branch is what we ship from so we like to keep it working and relatively clean even during development so merging large bits of commented out code doesn't really make a lot of sense.

When it comes to style, tools, and patterns, it's also easiest to work on them early on instead of going back and fixing them (because it rarely gets done).

PTAL at the comments and let's see how we can proceed.

internal/pkg/gateway/registry/registry.go
andrew-coleman: Thanks.  I've removed reference to the discovery service for this PR, and will add it back in a later commit once we've worked out the interface

internal/pkg/gateway/registry/registry.go
andrew-coleman: I've removed this for now

go.mod
andrew-coleman: okay, do you want me to split this out, or are you happy to leave this in? (I'll know in future)

internal/pkg/gateway/server/api_test.go
```diff
@@ -0,0 +1,177 @@
...
+		t.Fatalf("Failed to prepare the transaction: %s", err)
+	}
+
+	if !bytes.Equal(result.Response.Value, []byte("MyResult")) {
+		t.Fatalf("Incorrect value: %s", result.Response.Value)
+	}
+}
+
+func TestSubmit(t *testing.T) {
+	//proposal, err := createProposal()
```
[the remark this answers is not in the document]
andrew-coleman: removed this

lindluni (commented):
I see you added generated mocks, all of our testing stubs and fixtures are for the most part generated programmatically. A `go:generate` directive should be included so the mocks can be regenerated via the simple `go generate ./...` command if/when the underlying interface changes.

Here is an example: https://github.com/hyperledger/fabric/blob/master/discovery/support/support.go#L19

lindluni:
You can use /ci-run, this is for people who don't have enough permissions

internal/pkg/gateway/server/test/utils.go
```diff
@@ -0,0 +1,106 @@
+/*
```
sykesm: Having a package to contain constructors for mocks is overkill. You're also building a package that I promise will eventually lead to import cycles.

sykesm (commented):
Before adding more to this PR, please look at reusing the protoutil package that already exists in Fabric. If you can use (or augment) those,  you'll get the concrete types you want, the error messages you need, and we can avoid yet another copy of protobuf marshaling utilties.

On the testify front, please review the available method and take a look at other bits of Fabric code to see how they're used. You'll see there are things like `require.Error`, `require.NoError`, `require.EqualError`, etc that establish additional constraints and make the code more readable. The assertion failures also result in messages that are generally sufficient to debug so there's no need to include another copy of arguments in messages.

The introduction of a `test` package is going to turn into a problem. You may not see it yet but you will. It's a utility package that will aggregate imports from all over the code base and that will turn into issues with import cycles. 

I see you chose to go with "mockery" mocks instead of "counterfeiter" fakes. Please note that mockery is unable to generate fakes for unexported interfaces and that the implementations need to be setup. Most of us have found counterfeiter fakes simpler and more flexible. I'd strongly recommend you evaluate the pros and cons before doubling down on mockery.

bestbeforetoday:
> I see you chose to go with "mockery" mocks instead of "counterfeiter" fakes. Please note that mockery is unable to generate fakes for unexported interfaces and that the implementations need to be setup. Most of us have found counterfeiter fakes simpler and more flexible. I'd strongly recommend you evaluate the pros and cons before doubling down on mockery.

@sykesm To help people contribute more effectively, is it worth removing Mockery as a suggested mocking framework from the style guide, or at least highlighting Counterfeiter as a strong recommendation because of the reasons you've stated?

internal/peer/node/start.go
```diff
@@ -91,6 +92,7 @@ import (
 	peergossip "github.com/hyperledger/fabric/internal/peer/gossip"
 	"github.com/hyperledger/fabric/internal/peer/version"
 	"github.com/hyperledger/fabric/internal/pkg/comm"
+	gatewayservice "github.com/hyperledger/fabric/internal/pkg/gateway/server"
```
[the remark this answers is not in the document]
andrew-coleman: changed to gatewayserver

internal/peer/node/start.go
```diff
@@ -818,7 +821,18 @@ func serve(args []string) error {
...
+			logger.Info("Starting peer with Gateway enabled")
+			gs, err := gatewayservice.CreateGatewayServer(serverEndorser)
+			if err != nil {
+				logger.Panicf("Failed to start Gateway server: %s", err)
+			}
+			gatewayprotos.RegisterGatewayServer(peerServer.Server(), gs)
+			logger.Info("Gateway server activated")
+
+		} else {
+			logger.Error("Discovery service must be enabled for embedded gateway")
```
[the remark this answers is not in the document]
andrew-coleman: The gateway will depend on discovery in order to work, so this will be an error in as far as we won't be able to start the gateway server without the discovery server.  However, in this initial commit, we are not using discovery so I have removed the if/else condition entirely for now.

internal/pkg/gateway/server/api.go
```diff
@@ -0,0 +1,89 @@
...
+	channel, chaincodeID, err := getChannelAndChaincodeFromSignedProposal(proposedTransaction.Proposal)
+	if err != nil {
+		return nil, fmt.Errorf("failed to unpack channel header: %w", err)
+	}
+	logger.Infof("Endorsing transaction, channel=%s, chaincodeID=%s", channel, chaincodeID)
+	endorsers := gs.registry.Endorsers(channel, chaincodeID)
+
+	var responses []*peer.ProposalResponse
+	// send to all the endorsers
+	for i := 0; i < len(endorsers); i++ {
```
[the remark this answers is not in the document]
andrew-coleman: No reason.  I've now changed it.

internal/pkg/gateway/server/api.go
```diff
@@ -0,0 +1,89 @@
...
+	signedProposal := proposedTransaction.Proposal
+	var proposal peer.Proposal
+	if err := proto.Unmarshal(signedProposal.ProposalBytes, &proposal); err != nil {
+		return nil, fmt.Errorf("failed to unmarshal signed proposal: %w", err)
+	}
+	channel, chaincodeID, err := getChannelAndChaincodeFromSignedProposal(proposedTransaction.Proposal)
+	if err != nil {
+		return nil, fmt.Errorf("failed to unpack channel header: %w", err)
+	}
+	logger.Infof("Endorsing transaction, channel=%s, chaincodeID=%s", channel, chaincodeID)
```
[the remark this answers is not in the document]
andrew-coleman: Sure.  For now, I've removed this one as suggested.

internal/pkg/gateway/server/api_test.go
```diff
@@ -0,0 +1,60 @@
...
+*/
+
+package server
+
+import (
+	"context"
+	"testing"
+
+	pb "github.com/hyperledger/fabric-protos-go/gateway"
+	"github.com/hyperledger/fabric/internal/pkg/gateway/server/test"
```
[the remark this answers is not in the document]
andrew-coleman: I've removed the test package

internal/pkg/gateway/server/protoutils.go
```diff
@@ -0,0 +1,228 @@
+/*
```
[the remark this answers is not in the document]
andrew-coleman: Thanks.  I've removed most of the code in favour of invoking these.
I need the ability to create an unsigned transaction envelope and I was hoping to add it to protoutil and refactor `protoutil.CreateSignedTx` to use it.  However, CreateSignedTx checks that the signer matches the signer in the signature header, so this was not so straightforward.  I see that there is a comment in this code suggesting this check might be worth removing (https://github.com/hyperledger/fabric/blob/master/protoutil/txutils.go#L156).  If removed, then this refactor would be easy.  For now, I've done a rather messy workaround by passing in a dummy signer.

internal/pkg/gateway/server/protoutils.go
```diff
@@ -0,0 +1,228 @@
...
+	"fmt"
+
+	"github.com/golang/protobuf/proto"
+	"github.com/hyperledger/fabric-protos-go/common"
+	pb "github.com/hyperledger/fabric-protos-go/gateway"
+	"github.com/hyperledger/fabric-protos-go/peer"
+	"github.com/hyperledger/fabric/internal/pkg/identity"
+)
+
+func (gs *Server) signProposal(proposal *peer.Proposal, sign identity.Signer) (*peer.SignedProposal, error) {
```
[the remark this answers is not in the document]
andrew-coleman: removed

internal/pkg/gateway/server/test/utils.go
andrew-coleman: I've removed this package. Thanks :)

andrew-coleman:
> I see you chose to go with "mockery" mocks instead of "counterfeiter" fakes. Please note that mockery is unable to generate fakes for unexported interfaces and that the implementations need to be setup. Most of us have found counterfeiter fakes simpler and more flexible. I'd strongly recommend you evaluate the pros and cons before doubling down on mockery.

I've changed the mock generation to counterfeiter.  Thanks.

internal/pkg/gateway/registry/registry.go
```diff
@@ -0,0 +1,46 @@
...
+)
+
+type Registry struct {
+	localEndorser gateway.Endorser
+	logger        *flogging.FabricLogger
+}
+
+// Construct a New registry representing the gateway's view of the network
+func New(localEndorser gateway.Endorser) (*Registry, error) {
+	logger := flogging.MustGetLogger("gateway.registry")
```
sykesm, line 23: nit: would be nice for the caller to be able to provide a logger.

internal/pkg/gateway/registry/registry.go
```diff
@@ -0,0 +1,46 @@
...
+	"github.com/hyperledger/fabric-protos-go/orderer"
+	"github.com/hyperledger/fabric/common/flogging"
+)
+
+type Registry struct {
+	localEndorser gateway.Endorser
+	logger        *flogging.FabricLogger
+}
+
+// Construct a New registry representing the gateway's view of the network
```
sykesm: nit: we have the doc string linters disabled because our code is awful about doc. That said, in general, godoc should be complete sentences ending with a period.

internal/pkg/gateway/registry/registry_test.go
```diff
@@ -0,0 +1,24 @@
...
+	"github.com/stretchr/testify/require"
+)
+
+func TestRegistry_Endorsers(t *testing.T) {
+	endorser := &mocks.Endorser{}
+	registry, err := New(endorser)
+	require.NoError(t, err, "Failed to create registry")
+
+	endorsers := registry.Endorsers("my_channel", "my_chaincode")
+	require.Equal(t, 1, len(endorsers), "There should be one endorser returned")
```
sykesm: There's a matcher for length: https://pkg.go.dev/github.com/stretchr/testify/require#Len

internal/pkg/gateway/server/api_test.go
```diff
@@ -0,0 +1,126 @@
...
+			ChaincodeId: &peer.ChaincodeID{Name: "my_chaincode"},
+			Input:       &peer.ChaincodeInput{Args: nil},
+		},
+	}
+
+	proposal, _, err := protoutil.CreateChaincodeProposal(
+		cp.HeaderType_ENDORSER_TRANSACTION,
+		"mychannel",
+		invocationSpec,
+		[]byte(""),
```
sykesm: nit: just as easy to do `[]byte{}`

internal/pkg/gateway/server/api_test.go
```diff
@@ -0,0 +1,126 @@
...
+		Payload:     marshal(payload, t),
+		Response:    response,
+		Endorsement: endorsement,
+	}
+}
+
+func marshal(msg proto.Message, t *testing.T) []byte {
+	buf, err := proto.Marshal(msg)
+	if err != nil {
+		t.Fatalf("Failed to marshal message: %s", err)
```
sykesm: seems this should be using testify?

internal/pkg/gateway/server/api_test.go
```diff
@@ -0,0 +1,126 @@
...
+	}
+
+	proposal, _, err := protoutil.CreateChaincodeProposal(
+		cp.HeaderType_ENDORSER_TRANSACTION,
+		"mychannel",
+		invocationSpec,
+		[]byte(""),
+	)
+
+	require.NoError(t, err, "Failed to create the proposal")
```
sykesm, line 94: may want to consider using `t.Helper()` for helper functions like this.

internal/pkg/gateway/server/apiutils_test.go
```diff
@@ -0,0 +1,63 @@
...
+	require.Equal(t, []byte("MyResult"), result.Value, "Incorrect value")
+}
+
+func TestCreateUnsignedTx(t *testing.T) {
+	proposal := &peer.Proposal{
+		Header:  []byte{},
+		Payload: []byte{},
+	}
+	response := createProposalResponse("MyResult", t)
+	_, err := createUnsignedTx(proposal, response)
```
sykesm: Why aren't we verifying the result? The next line asserts no error implying we expect a good value. I'd assume the value we're ignoring is the _raison d'étre_ of this function.

Put differently, asserting on no error is not the same as asserting we got back what we expect. There used to be lots of tests in Fabric that did this and we've slowly gone back to fix most of them.

edit: Looks like this applies to most of what's below?

internal/pkg/gateway/server/server.go
```diff
@@ -0,0 +1,35 @@
...
+// Server represents the GRPC server for the Gateway
+type Server struct {
+	registry gateway.Registry
+}
+
+// CreateGatewayServer creates an embedded instance of the Gateway
+func CreateGatewayServer(localEndorser gateway.Endorser) (*Server, error) {
+	registry, err := registry.New(localEndorser)
+	if err != nil {
+		logger.Errorf("failed to create gateway registry: %s", err)
```
sykesm: We're already logging and panicking higher up the stack.
```
		gs, err := gatewayserver.CreateGatewayServer(serverEndorser)
		if err != nil {
			logger.Panicf("Failed to start Gateway server: %s", err)
		}
```

Copying that message also highlights to me that Create != Start.

sykesm (commented):
We are making very good progress. To me, this is quite a bit cleaner than it was.

The large outstanding comments are around the tests. Asserting no error is not the same as making assertions on the result of a method or function. We need assertions to validate what we got back is what we expect.

internal/peer/node/start.go
```diff
@@ -800,8 +802,9 @@ func serve(args []string) error {
 		coreConfig.ValidatorPoolSize,
 	)
 
+	var ds *discovery.Service
```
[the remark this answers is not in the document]
andrew-coleman: true, this is left over from when I was passing it into the gateway.  I'll revert it for now.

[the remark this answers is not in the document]
sykesm: fwiw, when I prefix a comment with 'nit' I'm trying to indicate that the comment alone would not prevent a merge. I'm happy to see them addressed if there are other comments to resolve but they are not the things that stop a merge.

internal/pkg/gateway/server/apiutils_test.go
andrew-coleman: Actually, now that I've rewritten my implementation of createUnsignedTx to call down into protoutils, most of these tests are redundant (they duplicate test cases in txutils_test.go). So I'll cut this back to a single test.

internal/pkg/gateway/server/apiutils_test.go
```diff
@@ -0,0 +1,40 @@
...
+	"github.com/hyperledger/fabric/protoutil"
+	"github.com/stretchr/testify/require"
+)
+
+func TestGetValueFromResponse(t *testing.T) {
+	response := createProposalResponse("MyResult", t)
+
+	result, err := getValueFromResponse(response)
+	require.NoError(t, err, "Failed to extract value from response")
+	require.Equal(t, []byte("MyResult"), result.Value, "Incorrect value")
```
sykesm, line 21: this is better. At some point you're going to need to check the result more completely. When that happens, you're going to want to use `proto.Equal` to compare.

internal/pkg/gateway/server/apiutils_test.go
```diff
@@ -0,0 +1,40 @@
...
+func TestCreateUnsignedTx(t *testing.T) {
+	proposal := createProposal(t)
+	response1 := createProposalResponse("MyResult", t)
+	response2 := createProposalResponse("MyResult", t)
+	tx, err := createUnsignedTx(proposal, response1, response2)
+
+	require.NoError(t, err, "Failed to create unsigned tx")
+	payload, err := protoutil.UnmarshalPayload(tx.Payload)
+	require.NoError(t, err, "Failed to unmarshal the tx payload")
+	transaction, err := protoutil.UnmarshalTransaction(payload.Data)
```
sykesm, line 33: This is an example of where comparing against the full struct is generally easier.

sykesm (approved):

mastersingh24 (approved):
looks ok to me

PR #2332, opened 2021-02-01
Replace `gofmt` with the more opinionated `gofumpt`

sykesm:
`gofumpt` is a Go source code formatter that is backwards compatible with `gofmt` but more strict. This PR contains commits to update the source with `gofumpt` and to use it in CI for linting.


jyellick (approved):
Looks like this fixes a number of annoying things not caught by normal gofmt.  I see you even updated the doc, looks good to me.

lindluni:
NICE!

PR #2333, opened 2021-02-01
Replace hostname with invalid IP address

jcastrence:
#### Type of change

- Bug fix

#### Description

- Test failed locally because 'junk' was a resolvable name on the network. By
switching to an invalid IP address, the connection will always fail.

### DIFF
--- a/core/ledger/kvledger/txmgmt/statedb/statecouchdb/couchdb_test.go
+++ b/core/ledger/kvledger/txmgmt/statedb/statecouchdb/couchdb_test.go
@@ -271,12 +271,12 @@ func TestIsEmpty(t *testing.T) {
 	require.True(t, isEmpty)
 
 	configCopy := *config
-	configCopy.Address = "junk"
+	configCopy.Address = "address-and-port.invalid:0"
 	configCopy.MaxRetries = 0
 	couchInstance.conf = &configCopy
 	_, err = couchInstance.isEmpty(ignore)
 	require.Error(t, err)
-	require.Regexp(t, `unable to connect to CouchDB, check the hostname and port: http error calling couchdb: Get "?http://junk/_all_dbs"?`, err.Error())
+	require.Regexp(t, `unable to connect to CouchDB, check the hostname and port: http error calling couchdb: Get "?http://address-and-port.invalid:0/_all_dbs"?`, err.Error())
 }
 
 func TestDBBadDatabaseName(t *testing.T) {


core/ledger/kvledger/txmgmt/statedb/statecouchdb/couchdb_test.go
```diff
@@ -271,12 +271,12 @@ func TestIsEmpty(t *testing.T) {
 	require.True(t, isEmpty)
 
 	configCopy := *config
-	configCopy.Address = "junk"
+	configCopy.Address = "0.0.0.1"
```
lindluni: While `0.0.0.1` is a reserved address and unlikely to be assigned, it's not entirely impossible that it might exist. So, while `junk` existing was a corner case, so is `0.0.0.1`, perhaps using an address made up of words but far more likely to not exist like `thisisanonexistentaddress`

sykesm: The problem is that you can't ensure a host name is bad unless it's in a domain you control or it does not adhere to the RFC. A bad IP address, however, is always bad.

The 0/8 addresses are explicitly reserved by a number of RFCs just as 127/8 is reserved. One reference is https://tools.ietf.org/html/rfc6890#section-2.2.2 but there are others.

sykesm: So 0.0.0.0/8 is reserved but it looks like Linux now allows them to be used. And we all know if linux lets you use them, they'll get used.

I still don't think relying on a hostname is a good thing...

lindluni: Agreed, choosing a truly invalid IP is a better choice in general

sykesm: Last update and then I'll be quiet. If we have to use DNS, we can use the `.invalid` domain

https://tools.ietf.org/html/rfc2606

yacovm: You can always generate a long random string....

sykesm:
```suggestion
	configCopy.Address = "address-and-port.invalid:0"
```
Putting it all together, we know we can't use 0 as a destination port and we know that the `.invalid` domain "is intended for use in online construction of domain names that are sure to be invalid and which it is obvious at a glance are invalid"

sykesm (changes requested):
I think the change is probably fine but the suggestion should ensure that.

lindluni (approved):

sykesm (approved):

ale-linux (approved):

PR #2338, opened 2021-02-02
deps: bump go-dockerclient

sykesm:


lindluni:
Hit the channel-participation flake, re-ran failed check

lindluni (approved):
Pre-approving for when you take the draft off

PR #2366, opened 2021-02-04
Enable RepoLinter Tooling

lindluni:
I trimmed this down to exclude unnecessary checks and introduce it into PR checks.

Honestly, I'm not sure how I feel about it. Introducing node.js tooling into the repo simply to do this seems unnecessary. The repolinter is also making use of many out-of-date packages that are expected to break in upcoming node.js releases.

I would like to hear others thoughts on doing this, as I don't really like the idea of it, but thats just my opinion.

### DIFF
--- a/SECURITY.md
+++ b/SECURITY.md
@@ -8,5 +8,5 @@ There are two ways to report a security bug. The easiest is to email a descripti
 
 The other way is to file a confidential security bug in our [JIRA bug tracking system](https://jira.hyperledger.org). Be sure to set the “Security Level” to “Security issue”.
 
-The process by which the Hyperledger Security Team handles security bugs is documented further in our [Defect Response page](https://wiki.hyperledger.org/display/HYP/Defect+Response) on our [wiki](https://wiki.hyperledger.org).
+The process by which the Hyperledger Security Team handles security bugs is documented further in our [Defect Response page](https://wiki.hyperledger.org/display/SEC/Defect+Response) on our [wiki](https://wiki.hyperledger.org).
 
--- a/ci/azure-pipelines.yml
+++ b/ci/azure-pipelines.yml
@@ -30,6 +30,17 @@ stages:
           - script: ./ci/scripts/evaluate_commits.sh
             name: SetJobTriggers
 
+      - job: LintRepo
+        pool:
+          vmImage: ubuntu-18.04
+        container: hyperledger-tools.jfrog.io/repolinter:0.10.0
+        steps:
+          - checkout: self
+            path: 'fabric'
+            displayName: Checkout Fabric Code
+          - script: bundle exec /app/bin/repolinter.js
+            displayName: Lint
+
   - stage: UnitTests
     dependsOn: VerifyBuild
     jobs:
--- a/repolint.json
+++ b/repolint.json
@@ -1,0 +1,122 @@
+{
+  "axioms": {
+    "linguist": "language",
+    "licensee": "license",
+    "packagers": "packager"
+  },
+  "rules": {
+    "all": {
+      "apache-license-file:file-contents": [
+        "error",
+        {
+          "files": [
+            "LICENSE*"
+          ],
+          "content": "Apache License.*Version 2.0",
+          "fail-on-non-existent": true
+        }
+      ],
+      "code-of-conduct-file:file-contents": [
+        "error",
+        {
+          "files": [
+            "CODE_OF_CONDUCT*"
+          ],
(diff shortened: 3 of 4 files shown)


lindluni:
This doesn't work out of the box anyway on our linux runners due to the way they commented parameters in their function signatures...

sykesm:
The first change seems to be unrelated and unnecessary. The package name was chosen to avoid collisions when used by other packages with fakes. This is idiomatic. 

For the rest,  I’m not really big on anything that does node in this repo and if we had to, it should be within a container and not installed and massaged every time the task runs.

lindluni:
> The first change seems to be unrelated and unnecessary. The package name was chosen to avoid collisions when used by other packages with fakes. This is idiomatic.

I disagree with that statement (if anything I used 'idiomatic' too loosely), where in fabric do we do that anywhere else in the repo. A quick look shows tons of collisions for mock and fake packages, we handle them by renaming the import in every other instance but this one. Hasn't our position been, if you need a mock in a test instead of importing it from a secondary package, you should generate it in the test directories mock directory? I can see examples in history where we moved shared mocks and replicated the generation so we weren't importing across packages?. So really I should have re-generated the mock where it was needed. (It's also a stuttering package name)

But honestly it's neither here nor there and I don't have a strong opinion on it, I only changed it so I could standardize the globs I was using so I didn't have to deviate the pattern to handle an edge case.

Originally I tried to use a container, but that also doesn't install out of the box in the official `node:10` container. I thought maybe doing it in CI on the ephemeral runner might work, but that also seems like a lost cause. I'm willing to say that, compounded with introducing node into the repo is reason enough not to do this.

lindluni:
Closing till someone can make this run out of the box

lindluni:
I just noticed why I couldn't find a prebuilt container image, they are hosting it on GHCR, not DockerHub, so let me redo it with that image.

Edit: And nevermind again, they just started publishing their image, but the only release failed to publish, so there is no available docker image: https://github.com/todogroup/repolinter/runs/1626549660?check_suite_focus=true

denyeart:
The idiomatic debate aside, why have the metricsfakes commit in this PR?  The Fabric PR may be a precedent for the repolinter, so would be nice to keep the PR small and clean.

lindluni:
> The idiomatic debate aside, why have the metricsfakes commit in this PR? The Fabric PR may be a precedent for the repolinter, so would be nice to keep the PR small and clean.

I wanted to keep the glob patterns in `repolint.json` as simple as possible and as much like each other as possible. Otherwise the `mocks` and `fakes` patterns would have been different to account for the one off `metricsfakes` directory. Seemed better to me to standardize the directory than to deviate on the glob pattern, since it's literally the only one that follows this pattern.

lindluni:
I'm really not trying to start a debate on the directories, my purpose was just to get the repolinter enabled and behind us. I should take my previous comments back and just back that commit out of this PR (which is what I'm going to do)

sykesm:
>  I wanted to keep the glob patterns in `repolint.json` as simple as possible and as much like each other as possible.

Globs aren’t necessarily the right approach. A lot of development tools identify generated code from a regex in the header and can be configured to ignore them.

lindluni:
There we go, running in CI in a container. Backed out all the other changes

denyeart:
@lindluni Thanks for the work on this. We discussed in Hyperledger TSC meeting today and it sounds like there is growing consensus on a slightly different approach across all the projects:
- Point to a common Hyperledger repolint.json in a Hyperledger repo instead of copying into each project repo (assuming that it can be cleaned up for common basic project structure, remove language specific things, etc)
- Run a github action instead of an Azure Pipeline (that's the strategic direction regardless, one of the other projects prototyped this with ease).

So, less maintenance for each project on both points. I expect TSC will provide final guidance in the coming week or two.

@lehors FYI.

lehors (changes requested):
Thanks for looking into this. While I support the idea of reformatting repolint.json to make it more readable I think it's unfortunate to do that at the same time as changing it. It makes spotting what has actually changed difficult.

As Dave said, the TSC discussed this (enabling repolinter) and the general sentiment is that we should try to have a single common config file used across all projects if possible. So, I suggest holding off on merging this PR which adds a copy, just like mine did, and instead suggest changes against the reference config file at: https://github.com/hyperledger-labs/hyperledger-community-management-tools/blob/main/repo_structure/repolint.json

mastersingh24:
Honestly, this is just a waste of time.  But thanks @lindluni for doing this.

PR #2375, opened 2021-02-08
Sync release notes from release-2.3 into main

sykesm:
This commit simply copies the current release_notes folder from 2.3 into the development branch.

### DIFF
--- a/release_notes/v2.3.0.md
+++ b/release_notes/v2.3.0.md
@@ -1,0 +1,84 @@
+v2.3.0 Release Notes - November 18, 2020
+========================================
+
+What's New in Hyperledger Fabric v2.3
+-------------------------------------
+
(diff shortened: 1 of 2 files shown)


sykesm:
Cool. I hit _both_ of the prominent flakes on a PR that changed no code.

- Timeout running common/cluster (`FAIL github.com/hyperledger/fabric/orderer/common/cluster 1200.197s`)
- Active vs. Inactive from the channel participation test (`fabric/integration/raft/channel_participation_test.go:1133`)

denyeart:
We've decided to just keep the release notes in the associated release branch, rather than copy into master every release.

If your intent is to move the release notes forward as a template for the next release from master, rename the file to v2.4.0 and remove the v2.3.0 content.

sykesm:
> We've decided to just keep the release notes in the associated release branch, rather than copy into master every release.

Then maybe remove the old stuff to make that clear?

PR #2379, opened 2021-02-08
CRL signature validation change in preparation for Go 1.15

sykesm:
The Go team has tightened up ECDSA signature validation in the standard library. If a CRL contains a signature that used to validate but no longer does, identities revoked by the CRL would no longer be considered invalid.

This commit adds a test that asserts on the go 1.14 behavior. To preserve that behavior, it modifies the MSP initialization to round-trip ECDSA signatures on CRLs to drop extra data from the encoded signature that would be treated as an error by [`ecdsa.VerifyASN1`](https://golang.org/pkg/crypto/ecdsa/#VerifyASN1) but would be accepted by the simpler `asn1.Unmarshal` used in go 1.14.

FAB-18419

### DIFF
--- a/msp/configbuilder.go
+++ b/msp/configbuilder.go
@@ -354,11 +354,12 @@ func getMspConfig(dir string, ID string, sigid *msp.SigningIdentityInfo) (*msp.M
 		FabricNodeOus:                 nodeOUs,
 	}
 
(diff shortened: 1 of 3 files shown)


yacovm (approved):

yacovm:
@adecaro  @ale-linux  maybe it can be good if you two look at it as well?

PR #2384, opened 2021-02-10
Reproduce name constraint verification from standard library in the MSP

sykesm:
Implemented @yacovm's suggestion from  https://github.com/hyperledger/fabric/pull/1716#issuecomment-673583502.

### DIFF
--- a/msp/mspimpl.go
+++ b/msp/mspimpl.go
@@ -10,8 +10,10 @@ import (
 	"bytes"
 	"crypto/x509"
 	"crypto/x509/pkix"
+	"encoding/asn1"
 	"encoding/hex"
 	"encoding/pem"
+	"strings"
 
 	"github.com/golang/protobuf/proto"
 	m "github.com/hyperledger/fabric-protos-go/msp"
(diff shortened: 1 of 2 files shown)


yacovm:
Let's get some feedback from @adecaro  and @ale-linux

msp/mspimpl.go
```diff
@@ -728,9 +730,107 @@ func (msp *bccspmsp) getUniqueValidationChain(cert *x509.Certificate, opts x509.
...
 
+var (
+	oidExtensionSubjectAltName  = asn1.ObjectIdentifier{2, 5, 29, 17}
+	oidExtensionNameConstraints = asn1.ObjectIdentifier{2, 5, 29, 30}
+)
+
+// verifyLegacyNameConstraints exercises the name constraint validation rules
+// that were part of the certificate verification process in Go 1.14.
+//
+// If a signing certificate contains a name constratint, the leaf certificate
```
yacovm, line 750: s/constratint/constraint/

yacovm:
This looks OK to me and is aligned with the logic in **x509/verify.go**  but would like other people to take a look too

msp/mspimpl.go
```diff
@@ -728,9 +730,107 @@ func (msp *bccspmsp) getUniqueValidationChain(cert *x509.Certificate, opts x509.
...
+		}
+	}
+	return false
+}
+
+// validHostname reports whether host is a valid hostname that can be matched or
+// matched against according to RFC 6125 2.2, with some leniency to accommodate
+// legacy values.
+//
+// This implementation is sourced from the standaard library.
```
jyellick, line 790: Doesn't matter, but 'standard'

jyellick (commented):
Count me in as 'one other person' who has looked and seems to think it looks good.  But, would like to see @adecaro or @ale-linux weigh in still.

ale-linux:
I had a look at the code and the go sources and with this patch we seem to indeed be replicating the pre-1.15 behaviour. That said: I'm wondering whether that's the right path forward, or whether we should just follow the idea that I think @denyeart once proposed of using our own fork of `crypto/x509`. Otherwise the already-complex msp code would become messier and messier.

denyeart:
I'm actually in favor of Matt's approach for now. Our own fork of crypto/x509 would require regular review and maintenance. If the MSP logic gets messier and messier over time, we could re-assess.

> I had a look at the code and the go sources and with this patch we seem to indeed be replicating the pre-1.15 behaviour. That said: I'm wondering whether that's the right path forward, or whether we should just follow the idea that I think @denyeart once proposed of using our own fork of `crypto/x509`. Otherwise the already-complex msp code would become messier and messier.

ale-linux:
I'm no longer sure about the regular review/maintenance. We face the very same choice every time we upgrade the go toolchain. Forking would only mean that we decouple the two problems i) updating the go toolchain and ii) updating our x509 dependency. Note that now those two coincide.

Neither option (forking x509 vs. patching msp) sounds very exciting, mind you. If other maintainers are convinced that further patching the msp is the right way forward, I'm happy to go with that.

yacovm:
I think it's better to go down the current path, and hope that there won't be changes that affect us (too much, or at least too soon).

If we do encounter a dead-end (like reducing the space of valid certificates instead of increasing it like 1.14->1.15) then we can just fork the x509 library in that future.

jyellick (approved):
There seems to be consensus that this patch is correct, and that we can always re-examine the approach later, going to go ahead and merge.

yacovm:
> I think it's better to go down the current path, and hope that there won't be changes that affect us (too much, or at least too soon).
> 
> If we do encounter a dead-end (like reducing the space of valid certificates instead of increasing it like 1.14->1.15) then we can just fork the x509 library in that future.

It's possible we have just [encountered](https://tip.golang.org/doc/go1.19#library) a dead-end:

> The path builder of [Certificate.Verify](https://tip.golang.org/pkg/crypto/x509/#Certificate.Verify) was overhauled and should now produce better chains and/or be more efficient in complicated scenarios. Name constraints are now also enforced on non-leaf certificates.

PR #2385, opened 2021-02-10
Use TLS certificates with a SAN in the peer package tests

sykesm:
The peer tests were using TLS certificate fixtures with hostnames in the subject CommonName and without SAN extensions and Go 1.15, by default, ignores the common name. Instead of simply regenerating the fixtures, the tests were updated to generate and use certificates at runtime. This addresses the SAN requirement and reduces the fixture footprint.


sykesm:
This is waiting on #2384.

jyellick (commented):
Hooray or removing fixtures in favor of generation.  Looks good, but not approving because of pending dep as noted.

jyellick (dismissed):

jyellick:
@sykesm I know you prefer for your commits not to be squashed, but this PR has commits from a previously merged PR in it, so cannot be automatically rebased.  I can either squash, or please push a manual rebasing of your own.

jyellick (approved):

PR #2386, opened 2021-02-10
Update build and CI to use Go 1.15

sykesm:


sykesm:
This is waiting on #2385.

jyellick (commented):
Looks good to me, not approving because of the noted prereq.

jyellick (approved):

PR #2394, opened 2021-02-12
FABGW-8: Add a commit notifier for embedded Gateway

bestbeforetoday:
Add a notifier that enables the embedded Gateway to get the validation code for a specific transaction to be submitted to the orderer, once it is committed to the local peer's ledger. The notifier provides an abstraction of the way transaction commits are obtained to enable an alternative implementation (with a shared interface) to be developed for a standalone Gateway.

Commit notification is not yet wired into the Gateway runtime. This will occur in a follow-on commit that enables the end-to-end submit transaction flow for the embedded Gateway.


manish-sethi (commented):
Thanks @bestbeforetoday. I have not yet reviewed the tests as some of those may change if you consider the comments on the main code.

internal/pkg/gateway/commit/blockparser.go
```diff
@@ -0,0 +1,59 @@
+/*
+Copyright 2021 IBM All Rights Reserved.
+
+SPDX-License-Identifier: Apache-2.0
+*/
+
+package commit
```
manish-sethi, line 7: Nit: The package name commit seems a bit over. Can we name it commitnotifications?

internal/pkg/gateway/commit/notifier.go
```diff
@@ -0,0 +1,91 @@
...
+
+// ChannelFactory is a single-method interface view of peer.Peer to allow access to channels.
+type ChannelFactory interface {
+	Channel(channelID string) (Channel, error)
+}
+
+// Channel interface view of peer.Channel Provides only the methods needed for commit notification.
+type Channel interface {
+	Reader() blockledger.Reader
+}
```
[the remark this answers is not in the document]
bestbeforetoday: - Go seems to discourage `Get...()` method names: https://golang.org/doc/effective_go.html#Getters
- The reason for the two layers of interface is that is how the capability in existing Fabric structs I'm using is exposed, but I agree fewer interfaces would be nicer here so I'll try to rework to a single interface, as you suggest 👍

internal/pkg/gateway/commit/notifier.go
```diff
@@ -0,0 +1,91 @@
...
+	}
+
+	blockReader := channel.Reader()
+	go readCommit(commit, blockReader, transactionID)
+
+	return nil
+}
+
+func readCommit(commit chan<- peerProto.TxValidationCode, blockReader blockledger.Reader, transactionID string) {
+	blockIter, _ := blockReader.Iterator(seekNextCommit())
```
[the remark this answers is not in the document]
bestbeforetoday: - The code is not behaving as you describe. The iterator is obtained once and used to obtain blocks until the required transaction is located
- I have no intention of exposing this function directly via gRPC. It is intended to be used internally by the Gateway to identify when transactions submitted to the orderer have committed on the local peer
- Some kind of timeout will be required once the end-to-end behaviour is implemented, and perhaps this code might want a `cancel` channel so it can be signalled to stop processing when that timeout is hit. Agile development though. Small steps at a time. I don't want to boil the ocean in one PR

internal/pkg/gateway/commit/notifier.go
```diff
@@ -0,0 +1,91 @@
...
+}
+
+func transactionStatus(block *commonProto.Block, transactionID string) (peerProto.TxValidationCode, bool) {
+	parser := blockParser{
+		Block: block,
+	}
+	validationCodes, err := parser.TransactionValidationCodes()
+	if err != nil {
+		// TODO: log error
+		return peerProto.TxValidationCode(0), false
```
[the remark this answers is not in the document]
bestbeforetoday: If the channel is closed then reads will return 0 anyway, and I can't control that behaviour. It seemed more consistent to return the same value for all cases where a valid value can't be obtained, but it doesn't really matter to me either way. The calling code needs to check the status, not just the value. I can change to whatever error return value will make you happy

internal/pkg/gateway/commit/blockparser.go
```diff
@@ -0,0 +1,59 @@
...
+
+import (
+	"github.com/golang/protobuf/proto"
+	commonProto "github.com/hyperledger/fabric-protos-go/common"
+	peerProto "github.com/hyperledger/fabric-protos-go/peer"
+	"github.com/pkg/errors"
+)
+
+type blockParser struct {
```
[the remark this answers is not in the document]
bestbeforetoday: You're right, there is more than one way this could be implemented, and plain functions is one. That would require a couple of functions and their calls to have an additional `Block` parameter. I find it nice to avoid that for this self-contained set of inter-related functions. Functions with fewer parameters are generally considered more readable and maintainable

internal/pkg/gateway/commit/blockparser.go
```diff
@@ -0,0 +1,59 @@
...
+	Block *commonProto.Block
+}
+
+func (parser *blockParser) TransactionValidationCodes() (map[string]peerProto.TxValidationCode, error) {
+	result := make(map[string]peerProto.TxValidationCode)
+
+	envelopes := parser.Block.Data.Data
+	for i, envelopeBytes := range envelopes {
+		channelHeader, err := unmarshallChannelHeader(envelopeBytes)
```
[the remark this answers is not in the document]
bestbeforetoday: I agree it is better to ignore malformed committed transactions rather than stop with an error. I'll change that. Checking for nil is also a good call so I'll add some unit tests to drive that implementation.

Is it expected that committed blocks can contain malformed transaction protobufs?

internal/pkg/gateway/commit/blockparser.go
```diff
@@ -0,0 +1,59 @@
...
+		if err != nil {
+			return nil, errors.Wrapf(err,
+				"failed to unmarshall channel header from envelope at index %v in block number %v", i, parser.Block.Header.Number)
```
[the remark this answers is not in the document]
bestbeforetoday: `%d` and `%v` produce identical result. Why does it matter, other than `%v` being less likely to fail at runtime if the code is changed in the future?

I'm not sure I understand the rule on errors clearly. The error being wrapped here is one returned from a third-party library (the calls to `proto.Unmarshal()`). Is there a good resource for error-handling best-practices beyond https://blog.golang.org/go1.13-errors ?

internal/pkg/gateway/commit/blockparser.go
```diff
@@ -0,0 +1,59 @@
...
+		}
+
+		result[channelHeader.TxId] = parser.validationCode(i)
+	}
```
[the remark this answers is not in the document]
bestbeforetoday: This has blown my mind! So transaction IDs are not unique, even with a single committed block? And a committed block may contain multiple versions of the same transaction, where some have committed successfully and some have failed? If a block contains two versions of the same transaction, should I use the status of the first one or if any of them have a successful status can I consider the commit successful?

If the channel header does not contain the status of an endorser transaction, how can I tell whether that transaction was successfully committed and updated the ledger?

internal/pkg/gateway/commit/blockparser.go
bestbeforetoday: That seems to contradict the general advice that Go package names are short (typically not multi-word) and avoid stuttering in struct names contained within that package (e.g. `commitnotifications.Notifier`): https://golang.org/doc/effective_go.html#package-names

internal/pkg/gateway/commit/notifier.go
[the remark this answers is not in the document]
bestbeforetoday: Yes, absolutely that would be a great thing to do in the future

internal/pkg/gateway/commit/blockparser.go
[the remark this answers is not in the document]
bestbeforetoday: How compatible with Go's built-in error wrapping is the GitHub errors package?

internal/pkg/gateway/commit/blockparser.go
```diff
@@ -0,0 +1,72 @@
...
+)
+
+type blockParser struct {
+	Block *commonProto.Block
+}
+
+func (parser *blockParser) TransactionValidationCodes() (map[string]peerProto.TxValidationCode, error) {
+	results := make(map[string]peerProto.TxValidationCode)
+
+	envelopes := parser.Block.Data.Data
```
jyellick, line 25: This is fine so long as your only consumer is inside the peer process (there should be no way to get back a malformed block internally), but if you want to make this operable outside the peer, probably worth a nil check on `block.Data`

internal/pkg/gateway/commit/blockparser.go
```diff
@@ -0,0 +1,72 @@
...
+	for i, envelopeBytes := range envelopes {
+		channelHeader, err := unmarshalChannelHeader(envelopeBytes)
+		if err != nil {
+			return nil, ghErrors.WithMessagef(err, "failed to unmarshal channel header from envelope at index %v in block number %v",
```
jyellick, line 29: In general, we use github.com/pkg/errors for our errors in Fabric. If you really want to use vanilla `errors` I suppose that's not the end of the world, but let's not use both.

internal/pkg/gateway/commit/notifier.go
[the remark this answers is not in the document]
manish-sethi: As this impacts the peer's performance indirectly, I am not sure what motivates to push this to the future and not getting it right in the first cut.

internal/pkg/gateway/commit/blockparser.go
```diff
@@ -0,0 +1,72 @@
...
+				i, parser.Block.Header.Number)
+		}
+
+		validationCode := parser.validationCode(i)
+		if validationCode == peerProto.TxValidationCode_BAD_PROPOSAL_TXID {
```
jyellick, line 34: I understand this code since we discussed it, but perhaps a small comment explaining that why we single out this particular status type.  

```
// Ignore txids which could have been forged by other parties,
// otherwise accept the first failure since it originated from the gateway client.
```

internal/pkg/gateway/commit/blockparser.go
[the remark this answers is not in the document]
jyellick: +1, I won't argue that this be changed, but it's odd stylistically to me to create a structure with no state, just to implicitly pass a parameter around.

internal/pkg/gateway/commit/notifier.go
```diff
@@ -0,0 +1,80 @@
...
+// Notify the supplied channel when the specified transaction is committed to a channel's ledger.
+func (notifier *Notifier) Notify(channelID string, transactionID string) (<-chan peerProto.TxValidationCode, error) {
+	// TODO: add a context parameter to enable the notifier to be cancelled
+	// TODO: pooling of ledger iterators over multiple invocations
+	blockIterator, err := notifier.reader.Iterator(channelID, seekNextCommit())
+	if err != nil {
+		return nil, err
+	}
+
+	commitChannel := make(chan peerProto.TxValidationCode, 1)
```
jyellick, line 37: Just a note.  I think that returning the channel for the result here is adequate, because in this case, we always expect for `readCommit` to read indefinitely, and never expect an error to occur other than (once wired) the context cancellation.  If `readCommit` were to have an expected error conditions which would cause the channel to be simply closed without populating a result, then I think we would want to use a slightly more sophisticated data structure which could pass an error or return status.  But, I think this is fine for this case.

internal/pkg/gateway/commit/notifier.go
```diff
@@ -0,0 +1,80 @@
...
+}
+
+func readCommit(commit chan<- peerProto.TxValidationCode, blockIterator blockledger.Iterator, transactionID string) {
+	defer blockIterator.Close()
+	defer close(commit)
+
+	for {
+		block, status := blockIterator.Next()
+		if status != commonProto.Status_SUCCESS {
+			return
```
jyellick, line 50: Since you noted it below, I would likewise log a warning here, if the iterator fails, this is an unexpected error condition.

internal/pkg/gateway/commit/notifier_test.go
```diff
@@ -0,0 +1,299 @@
...
+	"github.com/stretchr/testify/assert"
+)
+
+func TestNotifier(t *testing.T) {
+	type Transaction struct {
+		ID     string
+		Status peerProto.TxValidationCode
+	}
+
+	NewPayload := func(transactionID string) *commonProto.Payload {
```
jyellick, line 28: It's a little to have a locally scoped variable start with a capital.  This is repeated throughout the test.

internal/pkg/gateway/commit/notifier_test.go
```diff
@@ -0,0 +1,299 @@
...
+		channelHeader := &commonProto.ChannelHeader{
+			Type: int32(commonProto.HeaderType_ENDORSER_TRANSACTION),
+			TxId: transactionID,
+		}
+		channelHeaderBytes, err := proto.Marshal(channelHeader)
+		assert.NoError(t, err)
```
jyellick, line 34: Here, and in the rest of these setup function, we'd be better off with `require.NoError`, since if this step fails, almost certainly we will encounter future red herrings in assertions.

internal/pkg/gateway/commit/notifier_test.go
```diff
@@ -0,0 +1,299 @@
...
+
+		return blockReader
+	}
+
+	NewTestNotifier := func(blocks ...*commonProto.Block) *commit.Notifier {
+		blockReader := NewMockBlockReader(blocks...)
+		return commit.NewNotifier(blockReader)
+	}
+
+	t.Run("Notify returns error if block iterator cannot be obtained", func(t *testing.T) {
```
jyellick, line 100: There's nothing strictly wrong with the pattern you're following here, embedding all of the setup functions using closures which capture the *testing.T in order to not have to pass it around.  That being said, it would also have been perfectly acceptable (and perhaps more typical) to simply make the closures vanilla functions which take a `*testing.T` as their first parameter since presumably, they will be useful in other tests.  But, no need to change it now.

Additionally, it seems possible that  each of these test cases could be consolidated into a table test like: https://github.com/hyperledger/fabric/blob/5bad6aec88e576a377a65192a4eaf69e8307f5f7/core/chaincode/chaincode_support_test.go#L1218-L1290

I'm not sure how perfectly these would transition into a table test, but there seems to be quite a bit of shared setup, followed by a set of assertions over the same results.  Not a requirement, but you might want to see if it makes sense.

internal/pkg/gateway/commit/notifier_test.go
```diff
@@ -0,0 +1,299 @@
...
+		expected := errors.New("MY_ERROR_MESSAGE")
+		blockReader := &commitMocks.BlockReader{}
+		blockReader.IteratorReturns(nil, expected)
+		notifier := commit.NewNotifier(blockReader)
+
+		_, err := notifier.Notify("channel", "transactionID")
+		assert.ErrorContains(t, err, expected.Error())
```
jyellick, line 107: For tests, we generally prefer assertions to be made over literals where possible, rather than deriving them.  (It also gets rid of the need to store the intermediate error, so saves you a line in your setup).

jyellick (commented):
Generally looks good, a lot of comments, some should be fixed, but many of them are either trivial or only require action if you'd like to.

internal/pkg/gateway/commit/blockparser.go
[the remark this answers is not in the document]
manish-sethi: I think that should work, see [this](https://github.com/pkg/errors/blob/5dd12d0cfe7f152f80558d591504ce685299311e/errors.go#L163). I haven't tried this though.

internal/pkg/gateway/commit/notifier.go
```diff
@@ -0,0 +1,80 @@
...
+
+SPDX-License-Identifier: Apache-2.0
+*/
+
+package commit
+
+import (
+	commonProto "github.com/hyperledger/fabric-protos-go/common"
+	ordererProto "github.com/hyperledger/fabric-protos-go/orderer"
+	peerProto "github.com/hyperledger/fabric-protos-go/peer"
```
sykesm, line 12: I just want to come back around an highlight that these are not typical import names for Go. We avoid camel case for the package names.

Can se please revisit these imports and use consistent, lower case import names when needed?

jyellick: +1 not sure how I missed this

internal/pkg/gateway/commit/blockparser.go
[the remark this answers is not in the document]
bestbeforetoday: It's upsetting enough people, I'll change it

internal/pkg/gateway/commit/blockparser.go
bestbeforetoday: This code is very specific to an implementation embedded within a peer, but if it's allowable for `block.Data` (or any other elements) to be nil once committed to the ledger, I will add unit tests and code to handle that

internal/pkg/gateway/commit/blockparser.go
bestbeforetoday: I'll change to never use Go's own error APIs. Can we add the need to use GitHub's error package to the project coding guidelines?

internal/pkg/gateway/commit/notifier.go
bestbeforetoday: There are currently error paths where the channel can be closed without populating a result. How should I best handle this?

internal/pkg/gateway/commit/notifier_test.go
bestbeforetoday: 👍

internal/pkg/gateway/commit/notifier_test.go
bestbeforetoday: 👍

internal/pkg/gateway/commit/notifier_test.go
bestbeforetoday: Good spot. The direct use of `t` in the closure is a bug in my test code and would fail at runtime if any of the assertions in the helper functions tripped up. I'll fix this

bestbeforetoday: I think there is too much variation in the set up and assertions across these unit tests for table testing to be useful here, but I'll look again

internal/pkg/gateway/commit/notifier_test.go
bestbeforetoday: 👍

internal/pkg/gateway/commit/blockparser.go
```diff
@@ -0,0 +1,59 @@
...
+
+SPDX-License-Identifier: Apache-2.0
+*/
+
+package commit
+
+import (
+	"github.com/golang/protobuf/proto"
+	commonProto "github.com/hyperledger/fabric-protos-go/common"
+	peerProto "github.com/hyperledger/fabric-protos-go/peer"
```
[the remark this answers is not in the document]
bestbeforetoday: Thank you for the link. I haven't read that blog before. I was personally finding my reading of this code easier when I could see at a glance where I was working with protobufs rather than one of the other Fabric packages with the same name but I appreciate this is not idiomatic so I'll correct it

internal/pkg/gateway/commit/notifier.go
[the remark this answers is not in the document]
bestbeforetoday: Before making performance optimizations I would normally want to have benchmarking data to:

1. Confirm there is a performance issue worth investing effort, and adding complexity and risk to fix
2. Ensure the optimizations actually deliver results

Since this code is not yet in Fabric, and this PR doesn't deliver any code that is actually executed at runtime, I would appreciate your guidance on how to obtain this data.

It sounds like the the performance hot spot is getting and using blockledger iterators, so we should avoid doing that more than necessary. Is it worth exploring the possibility of implementing optimizations within the iterator itself so all consuming code can benefit? Again, I would appreciate your collaboration to implement optimizations based on measured performance.

bestbeforetoday:
To try to make progress more smoothly, I will reduce the scope of this PR to just reading from a `blockledger.Iterator` and sending appropriate status to a channel. I will open a new PR with that reduced scope submission once I can put something together.

PR #2395, opened 2021-02-12
Initial gateway integration tests

jt-nti:

### DIFF
--- a/integration/gateway/gateway_suite_test.go
+++ b/integration/gateway/gateway_suite_test.go
@@ -1,0 +1,103 @@
+/*
+Copyright IBM Corp All Rights Reserved.
+
+SPDX-License-Identifier: Apache-2.0
+*/
+
+package gateway
+
+import (
+	"encoding/json"
+	"testing"
+
+	"github.com/hyperledger/fabric-protos-go/common"
+	"github.com/hyperledger/fabric-protos-go/gateway"
+	"github.com/hyperledger/fabric-protos-go/peer"
+	"github.com/hyperledger/fabric/common/flogging"
+	"github.com/hyperledger/fabric/integration"
+	"github.com/hyperledger/fabric/integration/nwo"
+	"github.com/hyperledger/fabric/protoutil"
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+)
+
+func TestGateway(t *testing.T) {
+	RegisterFailHandler(Fail)
+	RunSpecs(t, "Gateway Service Suite")
+}
+
+var (
+	buildServer *nwo.BuildServer
+	components  *nwo.Components
+)
+
+var _ = SynchronizedBeforeSuite(func() []byte {
+	buildServer = nwo.NewBuildServer()
+	buildServer.Serve()
+
+	components = buildServer.Components()
+	payload, err := json.Marshal(components)
+	Expect(err).NotTo(HaveOccurred())
+
+	return payload
+}, func(payload []byte) {
+	err := json.Unmarshal(payload, &components)
+	Expect(err).NotTo(HaveOccurred())
+
+	flogging.SetWriter(GinkgoWriter)
+})
+
+var _ = SynchronizedAfterSuite(func() {
+}, func() {
+	buildServer.Shutdown()
+})
+
+func StartPort() int {
+	return integration.GatewayBasePort.StartPortForNode()
+}
+
+func NewProposedTransaction(signingIdentity *nwo.SigningIdentity, channelName, chaincodeName, transactionName string, args ...[]byte) *gateway.ProposedTransaction {
+	proposal, txnID := newProposalProto(signingIdentity, channelName, chaincodeName, transactionName, args...)
+	signedProposal, err := protoutil.GetSignedProposal(proposal, signingIdentity)
+	Expect(err).NotTo(HaveOccurred())
+
+	txn := &gateway.ProposedTransaction{
+		Proposal:  signedProposal,
+		TxId:      txnID,
+		ChannelId: channelName,
+	}
+
+	return txn
+}
+
+func newProposalProto(signingIdentity *nwo.SigningIdentity, channelName, chaincodeName, transactionName string, args ...[]byte) (*peer.Proposal, string) {
+	creator, err := signingIdentity.Serialize()
+	Expect(err).NotTo(HaveOccurred())
+
+	invocationSpec := &peer.ChaincodeInvocationSpec{
+		ChaincodeSpec: &peer.ChaincodeSpec{
+			Type:        peer.ChaincodeSpec_NODE,
+			ChaincodeId: &peer.ChaincodeID{Name: chaincodeName},
+			Input:       &peer.ChaincodeInput{Args: chaincodeArgs(transactionName, args...)},
+		},
+	}
+
+	result, transactionID, err := protoutil.CreateChaincodeProposal(
+		common.HeaderType_ENDORSER_TRANSACTION,
+		channelName,
+		invocationSpec,
+		creator,
+	)
+	Expect(err).NotTo(HaveOccurred())
+
+	return result, transactionID
+}
+
+func chaincodeArgs(transactionName string, args ...[]byte) [][]byte {
+	result := make([][]byte, len(args)+1)
+
+	result[0] = []byte(transactionName)
+	copy(result[1:], args)
+
+	return result
+}
--- a/integration/gateway/gateway_test.go
+++ b/integration/gateway/gateway_test.go
@@ -1,0 +1,143 @@
+/*
+Copyright IBM Corp All Rights Reserved.
+
+SPDX-License-Identifier: Apache-2.0
+*/
+
+package gateway
+
+import (
+	"context"
+	"io/ioutil"
+	"os"
+	"path/filepath"
+	"syscall"
+
+	docker "github.com/fsouza/go-dockerclient"
+	"github.com/golang/protobuf/proto"
+	"github.com/hyperledger/fabric-protos-go/gateway"
+	"github.com/hyperledger/fabric/integration/nwo"
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+	"github.com/tedsuo/ifrit"
+)
+
+var _ = Describe("GatewayService", func() {
+	var (
+		testDir   string
+		network   *nwo.Network
+		org1Peer0 *nwo.Peer
+		process   ifrit.Process
+	)
+
+	BeforeEach(func() {
+		var err error
+		testDir, err = ioutil.TempDir("", "gateway")
+		Expect(err).NotTo(HaveOccurred())
+
+		client, err := docker.NewClientFromEnv()
+		Expect(err).NotTo(HaveOccurred())
+
+		config := nwo.BasicEtcdRaft()
+		network = nwo.New(config, testDir, client, StartPort(), components)
+
+		network.GatewayEnabled = true
+
+		network.GenerateConfigTree()
+		network.Bootstrap()
+
+		networkRunner := network.NetworkGroupRunner()
+		process = ifrit.Invoke(networkRunner)
+		Eventually(process.Ready(), network.EventuallyTimeout).Should(BeClosed())
+
+		orderer := network.Orderer("orderer")
+		network.CreateAndJoinChannel(orderer, "testchannel")
+		network.UpdateChannelAnchors(orderer, "testchannel")
+		network.VerifyMembership(
+			network.PeersWithChannel("testchannel"),
+			"testchannel",
+		)
+		nwo.EnableCapabilities(
+			network,
+			"testchannel",
+			"Application", "V2_0",
+			orderer,
+			network.PeersWithChannel("testchannel")...,
+		)
+
+		chaincode := nwo.Chaincode{
+			Name:            "gatewaycc",
+			Version:         "0.0",
+			Path:            components.Build("github.com/hyperledger/fabric/integration/chaincode/simple/cmd"),
+			Lang:            "binary",
+			PackageFile:     filepath.Join(testDir, "gatewaycc.tar.gz"),
+			Ctor:            `{"Args":[]}`,
+			SignaturePolicy: `AND ('Org1MSP.peer')`,
+			Sequence:        "1",
+			InitRequired:    false,
+			Label:           "gatewaycc_label",
+		}
+
+		nwo.DeployChaincode(network, "testchannel", orderer, chaincode)
+
+		org1Peer0 = network.Peer("Org1", "peer0")
+	})
+
+	AfterEach(func() {
+		if process != nil {
+			process.Signal(syscall.SIGTERM)
+			Eventually(process.Wait(), network.EventuallyTimeout).Should(Receive())
+		}
+		if network != nil {
+			network.Cleanup()
+		}
+		os.RemoveAll(testDir)
+	})
+
+	Describe("calling Evaluate", func() {
+		It("should respond with the expected result", func() {
+			conn := network.PeerClientConn(org1Peer0)
+			defer conn.Close()
+			gatewayClient := gateway.NewGatewayClient(conn)
+			ctx, cancel := context.WithTimeout(context.Background(), network.EventuallyTimeout)
+			defer cancel()
+
+			signingIdentity := network.PeerUserSigner(org1Peer0, "User1")
+			txn := NewProposedTransaction(signingIdentity, "testchannel", "gatewaycc", "respond", []byte("200"), []byte("conga message"), []byte("conga payload"))
+
+			result, err := gatewayClient.Evaluate(ctx, txn)
+			Expect(err).NotTo(HaveOccurred())
+			expectedResult := &gateway.Result{
+				Value: []byte("conga payload"),
+			}
+			Expect(result.Value).To(Equal(expectedResult.Value))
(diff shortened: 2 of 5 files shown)


jt-nti:
I'm not about next week but it would be good to get some feedback so I can finish it off when I get back
Thanks

integration/gateway/gateway_suite_test.go
```diff
@@ -0,0 +1,169 @@
...
+	result := make([][]byte, len(args)+1)
+
+	result[0] = []byte(transactionName)
+	copy(result[1:], args)
+
+	return result
+}
+
+// TODO not the only suite doing this - move this to nwo?
+func GetSigningIdentity(network *nwo.Network, org, user, mspID, mspType string) msp.SigningIdentity {
```
sykesm: This doesn't make sense to me. If we need key material for a user or admin, the network provides functions to get the keys and certs. If we need the key for an orderer or a peer, I'd ask why, but we should not need an MSP to sign a transaction.

```
// OrdererUserKey returns the path to the private key for the specified user in
// the orderer organization.
func (n *Network) OrdererUserKey(o *Orderer, user string) string { ... }
```

```
// PeerUserKey returns the path to the private key for the specified user in
// the peer organization.
func (n *Network) PeerUserKey(p *Peer, user string) string { ... }
```

integration/gateway/gateway_suite_test.go
```diff
@@ -0,0 +1,169 @@
...
+	mspInstance := loadLocalMSPAt(mspConfigPath, mspID, mspType)
+
+	signingIdentity, err := mspInstance.GetDefaultSigningIdentity()
+	Expect(err).NotTo(HaveOccurred())
+	return signingIdentity
+}
+
+// TODO not the only suite doing this - move this to nwo?
+func loadLocalMSPAt(dir, id, mspType string) msp.MSP {
+	Expect(mspType).To(Equal("bccsp"))
```
sykesm: No, instead I'd ask why we're using an MSP and BCCSP in these tests. I expect there are some good use cases but, for this one, the cert is the serialized identity and the key is the signer.

Put differently, we need this:
```
// Signer is the interface needed to sign a transaction
type Signer interface {
	Sign(msg []byte) ([]byte, error)
	Serialize() ([]byte, error)
}
```
and this is building
```

// Identity interface defining operations associated to a "certificate".
// That is, the public part of the identity could be thought to be a certificate,
// and offers solely signature verification capabilities. This is to be used
// at the peer side when verifying certificates that transactions are signed
// with, and verifying signatures that correspond to these certificates.///
type Identity interface {

	// ExpiresAt returns the time at which the Identity expires.
	// If the returned time is the zero value, it implies
	// the Identity does not expire, or that its expiration
	// time is unknown
	ExpiresAt() time.Time

	// GetIdentifier returns the identifier of that identity
	GetIdentifier() *IdentityIdentifier

	// GetMSPIdentifier returns the MSP Id for this instance
	GetMSPIdentifier() string

	// Validate uses the rules that govern this identity to validate it.
	// E.g., if it is a fabric TCert implemented as identity, validate
	// will check the TCert signature against the assumed root certificate
	// authority.
	Validate() error

	// GetOrganizationalUnits returns zero or more organization units or
	// divisions this identity is related to as long as this is public
	// information. Certain MSP implementations may use attributes
	// that are publicly associated to this identity, or the identifier of
	// the root certificate authority that has provided signatures on this
	// certificate.
	// Examples:
	//  - if the identity is an x.509 certificate, this function returns one
	//    or more string which is encoded in the Subject's Distinguished Name
	//    of the type OU
	// TODO: For X.509 based identities, check if we need a dedicated type
	//       for OU where the Certificate OU is properly namespaced by the
	//       signer's identity
	GetOrganizationalUnits() []*OUIdentifier

	// Anonymous returns true if this is an anonymous identity, false otherwise
	Anonymous() bool

	// Verify a signature over some message using this identity as reference
	Verify(msg []byte, sig []byte) error

	// Serialize converts an identity to bytes
	Serialize() ([]byte, error)

	// SatisfiesPrincipal checks whether this instance matches
	// the description supplied in MSPPrincipal. The check may
	// involve a byte-by-byte comparison (if the principal is
	// a serialized identity) or may require MSP validation
	SatisfiesPrincipal(principal *msp.MSPPrincipal) error
}

// SigningIdentity is an extension of Identity to cover signing capabilities.
// E.g., signing identity should be requested in the case of a client who wishes
// to sign transactions, or fabric endorser who wishes to sign proposal
// processing outcomes.
type SigningIdentity interface {

	// Extends Identity
	Identity

	// Sign the message
	Sign(msg []byte) ([]byte, error)

	// GetPublicVersion returns the public parts of this identity
	GetPublicVersion() Identity
}
```

integration/chaincode/gateway/go.mod
```diff
@@ -0,0 +1,19 @@
+module github.com/hyperledger/fabric/integration/chaincode/gateway
+
+go 1.15
```
sykesm: We're not using go 1.15 in Fabric yet.

integration/chaincode/gateway/main.go
```diff
@@ -0,0 +1,79 @@
...
+*/
+
+package main
+
+import (
+	"errors"
+	"fmt"
+	"log"
+
+	"github.com/hyperledger/fabric-contract-api-go/contractapi"
```
sykesm: Do we need another chaincode or can we get away with extending one that exists?
Do we really need to pull in the contract API when all we need in `fabric-chaincode-go`?

integration/ports.go
```diff
@@ -38,6 +38,7 @@ const (
 	PrivateDataBasePort
 	RaftBasePort
 	SBEBasePort
+	GatewayBasePort
```
sykesm: These are in alphabetical order.

integration/gateway/gateway_test.go
```diff
@@ -0,0 +1,120 @@
...
+				ctx, cancel := context.WithTimeout(context.Background(), 100*time.Second)
+				defer cancel()
+
+				signingIdentity := GetSigningIdentity(network, "Org1", "User1", "Org1MSP", "bccsp")
+				txn := NewProposedTransaction(signingIdentity, "testchannel", "gatewaycc", "Echo", []byte("conga"))
+
+				result, err := gatewayClient.Evaluate(ctx, txn)
+				Expect(err).NotTo(HaveOccurred())
+				Expect(result).NotTo(BeNil())
+				Expect(result.Value).To(Equal([]byte("conga")))
```
sykesm: using `proto.Equal` with an expected struct would be preferred.

integration/gateway/gateway_suite_test.go
```diff
@@ -0,0 +1,169 @@
...
+
+import (
+	"encoding/json"
+	"io/ioutil"
+	"path/filepath"
+	"testing"
+	"time"
+
+	"github.com/hyperledger/fabric-protos-go/common"
+	gateway "github.com/hyperledger/fabric-protos-go/gateway"
```
sykesm: Why do we need a named import when the package is already named 'gateway'?

integration/gateway/gateway_suite_test.go
```diff
@@ -0,0 +1,169 @@
...
+	"github.com/hyperledger/fabric-protos-go/peer"
+	"github.com/hyperledger/fabric/bccsp/sw"
+	"github.com/hyperledger/fabric/common/flogging"
+	"github.com/hyperledger/fabric/integration"
+	"github.com/hyperledger/fabric/integration/nwo"
+	"github.com/hyperledger/fabric/internal/pkg/comm"
```
sykesm: Do we need to use an explicitly internal package in the outside-in integration tests instead of the grpc package?

sykesm (commented):
Mostly comments and questions with the general implication that these tests should not be using bccsp, msp, and internal packages. I also don't want to see our deps pull in the contract api unless necessary. Nothing in fabric references it today outside of the doc tree.

integration/gateway/gateway_suite_test.go
andrew-coleman: @jt-nti PR https://github.com/hyperledger/fabric/pull/2408 will help address this, I think

jt-nti: Aha, #2408 seems like what I was looking for, I'll give it a go. Thanks @sykesm @andrew-coleman

integration/gateway/gateway_suite_test.go
jt-nti: No idea what happened there- probably brain freeze!

integration/gateway/gateway_suite_test.go
jt-nti: Now using PeerClientConn so this will go

integration/gateway/gateway_test.go
jt-nti: Why is that the preferred option for such a simple struct? It seems to provide less useful output when the test fails:

```
Expected
  <bool>: false
to be true
```

sykesm: When fields are added to a struct, we'd like the assertions to include them. When we assert on individual fields, we're not asserting the compatible zero values of new fields or the non-zero fields in responses that should get tested.

As for proto.Equal, if you go through proto marshaling, depending the version of the proto package, a number of fields are populated that impact the basic equality tests.

If you need a matcher, we can add one. They're trivial:
```go
// Copyright IBM Corp. All Rights Reserved.
// SPDX-License-Identifier: Apache-2.0

package matcher

import (
	"fmt"

	"github.com/onsi/gomega/types"
	"google.golang.org/protobuf/proto"
)

// ProtoEqual is a gomega matcher for protobuf messages.
func ProtoEqual(expected proto.Message) types.GomegaMatcher {
	return &protoEqualMatcher{
		expected: expected,
	}
}

type protoEqualMatcher struct {
	expected proto.Message
}

func (e *protoEqualMatcher) Match(actual interface{}) (bool, error) {
	actualMessage, ok := actual.(proto.Message)
	if !ok {
		return false, fmt.Errorf("ProtoEqual expects a proto.Message")
	}

	return proto.Equal(e.expected, actualMessage), nil
}

func (e *protoEqualMatcher) FailureMessage(actual interface{}) string {
	return fmt.Sprintf("Expected\n\t%#v\nto proto.Equal\n\t%#v", actual, e.expected)
}

func (e *protoEqualMatcher) NegatedFailureMessage(actual interface{}) string {
	return fmt.Sprintf("Expected\n\t%#v\nnot to proto.Equal\n\t%#v", actual, e.expected)
}
```

integration/nwo/network.go
```diff
@@ -189,16 +190,17 @@ func New(c *Config, rootDir string, dockerClient *docker.Client, startPort int,
...
+		Consensus:      c.Consensus,
+		Orderers:       c.Orderers,
+		Peers:          c.Peers,
+		SystemChannel:  c.SystemChannel,
+		Channels:       c.Channels,
+		Profiles:       c.Profiles,
+		Consortiums:    c.Consortiums,
+		Templates:      c.Templates,
+		TLSEnabled:     true,  // Set TLS enabled as true for default
+		GatewayEnabled: false, // Set Gateway enabled as false for default
```
sykesm, line 205: nit: no need to set the zero value

integration/chaincode/gateway/go.mod
jt-nti: It looks like #2386 has updated Fabric to use go 1.15

integration/gateway/gateway_test.go
jt-nti: A matcher sounds like it would be nice to have in the future https://jira.hyperledger.org/browse/FAB-10301

integration/chaincode/gateway/main.go
sykesm: @jt-nti I'd like to come back to this. I see no reason why we need to introduce dependencies on the contract API for this integration test. I'm also not sure why one of the existing chaincode implementations that exist are insufficient to exercise these paths.

integration/gateway/gateway_test.go
```diff
@@ -0,0 +1,124 @@
...
+				Version:         "0.0",
+				Path:            chaincodePath,
+				Lang:            "golang",
+				PackageFile:     filepath.Join(testDir, "gatewaycc.tar.gz"),
+				Ctor:            `{"Args":[]}`,
+				SignaturePolicy: `AND ('Org1MSP.peer')`,
+				Sequence:        "1",
+				InitRequired:    false,
+				Label:           "gatewaycc",
+			}
```
sykesm: Can we use the binary builder and avoid spinning up the extra docker containers?

integration/gateway/gateway_test.go
sykesm: If it's nice to have, it would be even nicer for someone that needs it to pick up the story and implement it. That would be really nice...

As for the "trivial" nature, I disagree. When behavior changes at this level (the outside most API), it should most certainly require some explicit action. I've tried to explain the rationale for why evaluating the entire message is preferred over a subset.

integration/nwo/network.go
sykesm: Please revisit this when you come back on the other comments.

integration/chaincode/gateway/go.mod
sykesm: Yes, but most modules still have 1.13 or 1.14 compatibility levels declared in them because we're not making use of go 1.15. This significance of this directive is growing now that modules are the "standard" way of dealing with dependencies.

https://golang.org/ref/mod#go-mod-file-go

That said, I've already indicated in other comments that I don't think this chaincode (and associated module) is needed to test what you're trying to test and we should use or adapt what we have.

integration/gateway/gateway_test.go
```diff
@@ -0,0 +1,124 @@
...
+			org2Peer0 = network.Peer("Org2", "peer0")
+
+			network.CreateAndJoinChannel(orderer, "testchannel")
+			network.UpdateChannelAnchors(orderer, "testchannel")
+			network.VerifyMembership(network.PeersWithChannel("testchannel"), "testchannel")
+
+			nwo.EnableCapabilities(network, "testchannel", "Application", "V2_0", orderer, org1Peer0, org2Peer0)
+
+			chaincodePath, err := filepath.Abs("../chaincode/gateway")
+			Expect(err).NotTo(HaveOccurred())
```
sykesm: Can we please flatten this structure? We have a Describe/Describe/Context/It nesting for a single test.
There's also no symmetry between the before / after each implementations at the various levels.
There's also no apparent structure to the collection of the vars at the top level relative to where they are located.

Given we're dealing with one network, creating one client, to drive one scenario, I'd expect a single BeforeEach, a single AfterEach, and a single It (that has an implied context) .

integration/gateway/gateway_test.go
```diff
@@ -0,0 +1,124 @@
...
+			gatewayClient = gateway.NewGatewayClient(conn)
+		})
+
+		AfterEach(func() {
+			conn.Close()
+		})
+
+		Context("when I evaluate an echo transaction with the arguments [\"conga\"]", func() {
+			It("should respond with \"conga\"", func() {
+				ctx, cancel := context.WithTimeout(context.Background(), 100*time.Second)
```
sykesm: If we're going to use a timeout, please use the EventuallyTimeout on the network.

sykesm (commented):
The big discussion point is around  a new chaincode and the associated contract API. That seems gratuitous relative what we're trying to do in the test.

sykesm:
@jt-nti Please take a look at #2450. That includes a commit on top of what you've done here that applies the changes I've requested. You'll notice that it reduces the test execution time from 66s to 33s and does not require a new chaincode implementation, module dependencies, or a docker build process.

jt-nti:
Looks similar to the changes I've been working on- the binary chaincode is definitely an improvement. I just used the existing "respond" transaction though, which has highlighted some changes we need in the gateway protos.

jt-nti:
@sykesm updated this PR based on your PR, plus a gateway submit test since that PR went in. Alternatively, happy to do a follow on PR to #2450 if you prefer.

integration/gateway/gateway_test.go
```diff
@@ -0,0 +1,142 @@
...
+			InitRequired:    false,
+			Label:           "gatewaycc_label",
+		}
+
+		nwo.DeployChaincode(network, "testchannel", orderer, chaincode)
+
+		org1Peer0 = network.Peer("Org1", "peer0")
+		conn = network.PeerClientConn(org1Peer0)
+		gatewayClient = gateway.NewGatewayClient(conn)
+		ctx, cancel = context.WithTimeout(context.Background(), network.EventuallyTimeout)
```
sykesm: As was done in the commit I pointed you to, please move these to the `It` blocks (tests) and use a `defer` to handle the cancel and close instead of creating a block level `var` and relying on `AfterEach` to clean this bit up.

After that change, I'm happy with what's here.

jt-nti: No problem- moved. Out of interest why wasn't it a good idea to remove the duplication and rely on `AfterEach`?

sykesm (approved):

integration/gateway/gateway_test.go
sykesm: I started writing a reply but it looks like I lost it so I'll provide the highlights.

The first is that it's a common idiom in Go to acquire a connection and have the `defer conn.Close()` next to it after the error check. When it's missing, the reader then goes searching for where it's done. There's a similar rationale (and generally more important) around calling the cancel function in a defer.

The second is philosophical. There's a general preference for DRY production code and DAMP (and/or WET) test code. This makes it easy for people to jump to a failing test and quickly get context without having to walk through all of the callbacks used to setup for the test.

jt-nti:
Thanks @sykesm @lindluni

PR #2401, opened 2021-02-15
Isolate Kafka tests to their own suite

sykesm:
Move Kafka e2e, health checks, and Kafka to raft migration tests to their own suite and remove Kafka and ZooKeeper from the required docker image list.


C0rWin (dismissed):

C0rWin (commented):
I've missed it was marked as Draft, will once it will be ready...

C0rWin (approved):

PR #2408, opened 2021-02-16
Expose a SigningIdentity type on the integration test network model

sykesm:
Test suites that exercise gRPC services instead of the command line tools were instantiating an MSP to obtain a signing identity. In order to enable reuse of code and reduce the coupling of tests to the MSP and BCCSP, this introduces a Signer that satisfies the `Sign` and `Serialize` operations used when assembling protocol buffer messages.

With the introduction of `nwo.SigningIdentity`, additional commits update the existing suites to use the new object.

### DIFF
--- a/integration/nwo/signingid.go
+++ b/integration/nwo/signingid.go
@@ -1,0 +1,71 @@
+/*
+Copyright IBM Corp. All Rights Reserved.
+
+SPDX-License-Identifier: Apache-2.0
+*/
+
+package nwo
+
+import (
+	"crypto/ecdsa"
+	"crypto/rand"
+	"crypto/sha256"
+	"crypto/x509"
+	"encoding/pem"
+	"fmt"
+	"io/ioutil"
+
+	"github.com/golang/protobuf/proto"
+	"github.com/hyperledger/fabric-protos-go/msp"
+	"github.com/hyperledger/fabric/bccsp/utils"
+)
+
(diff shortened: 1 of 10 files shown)


C0rWin (dismissed):

integration/nwo/signingid.go
```diff
@@ -0,0 +1,71 @@
...
+	"fmt"
+	"io/ioutil"
+
+	"github.com/golang/protobuf/proto"
+	"github.com/hyperledger/fabric-protos-go/msp"
+	"github.com/hyperledger/fabric/bccsp/utils"
+)
+
+// A SigningIdentity represents an MSP signing identity.
+type SigningIdentity struct {
```
yacovm, line 24: We already have this [here](https://github.com/hyperledger/fabric/blob/master/cmd/common/signer/signer.go).

sykesm: Yes, in a production code package that is used within the context of the CLI.

It was used in one of the three tests that this PR is updating to stick within the nwo abstraction. In other words, I’m aware. 

I’m doing this because the stuff in these integration tests should be avoiding internals as much as possible. Originally we only used the CLI tools we built to drive these scenarios but now we’re using the gRPC services as well. We need to draw the line somewhere and this seems reasonable.

C0rWin (dismissed):

jyellick (approved):

PR #2409, opened 2021-02-17
FABGW-6 Implement Submit()

andrew-coleman:
Use discovery to get peer endorsers and orderers in order to submit a transaction to the ledger

A number of TODO comments are in the code marking where additional logic will be added as part of this user story (or future story where specified). This includes adding retry logic according to endorsement layouts, configurable timeouts, updating stored channel membership in response to config updates.


internal/pkg/gateway/registry/registry.go
```diff
@@ -7,40 +7,307 @@ SPDX-License-Identifier: Apache-2.0
...
+func (reg *Registry) newTLSConnection(url, host, mspid string) (*grpc.ClientConn, error) {
+	tlscert, ok := reg.msps[mspid]
+	if !ok {
+		return nil, fmt.Errorf("unknown msp: %s", mspid)
+	}
+	certPool := x509.NewCertPool()
+	if !certPool.AppendCertsFromPEM(tlscert) {
+		return nil, fmt.Errorf("failed to append certificate to client credentials")
+	}
+	creds := credentials.NewClientTLSFromCert(certPool, host)
```
yacovm: 1. It's better to use the [already existing](https://github.com/hyperledger/fabric/blob/master/internal/pkg/comm/client.go) client to create connections.
2. Perhaps consider decoupling the communication from the registry? All you need is some kind of "factory" for NewAtomicBroadcastClient right?

andrew-coleman: Thanks for the suggestions, Yacov.  I've made both these changes.

internal/pkg/gateway/gateway.go
```diff
@@ -7,25 +7,13 @@ SPDX-License-Identifier: Apache-2.0
...
-	ProcessProposal(ctx context.Context, in *peer.SignedProposal) (*peer.ProposalResponse, error)
-}
-
 //go:generate counterfeiter -o mocks/registry.go --fake-name Registry . Registry
 
 // Registry represents the current network topology
 type Registry interface {
-	Endorsers(channel string, chaincode string) []Endorser
-	Orderers(channel string) []orderer.AtomicBroadcast_BroadcastClient
+	Endorsers(channel string, chaincode string) ([]endpoint.Endorser, error)
```
sykesm: Now we're binding two packages by using a foreign interface (that I'm still not convinced we need).

internal/pkg/gateway/server/server_test.go
```diff
@@ -9,13 +9,15 @@ package server
...
 
 func TestCreateGatewayServer(t *testing.T) {
-	endorser := &mocks.Endorser{}
+	endorser := &endpointMocks.Endorser{}
+	discovery := &discoveryMocks.Support{}
 
-	_, err := CreateGatewayServer(endorser)
+	_, err := CreateGatewayServer(endorser, discovery, "localhost:7051")
 	require.NoError(t, err, "Failed to create gateway server")
 }
```
sykesm: This isn't testing anything.

internal/pkg/gateway/server/api_test.go
```diff
@@ -10,12 +10,15 @@ import (
...
+
 	"github.com/golang/protobuf/proto"
 	cp "github.com/hyperledger/fabric-protos-go/common"
 	pb "github.com/hyperledger/fabric-protos-go/gateway"
 	"github.com/hyperledger/fabric-protos-go/peer"
-	"github.com/hyperledger/fabric/internal/pkg/gateway"
+	endpointMocks "github.com/hyperledger/fabric/internal/pkg/gateway/endpoint/mocks"
 	"github.com/hyperledger/fabric/internal/pkg/gateway/mocks"
+	mocks2 "github.com/hyperledger/fabric/internal/pkg/gateway/server/mocks"
 	idmocks "github.com/hyperledger/fabric/internal/pkg/identity/mocks"
```
sykesm: If we need this many mocks in greenfield code, it's not a good sign.

internal/pkg/gateway/endpoint/endpoint.go
```diff
@@ -0,0 +1,186 @@
...
+	config := comm.ClientConfig{
+		SecOpts: secOpts,
+		Timeout: 5 * time.Second, // TODO from config
+	}
+	client, err := comm.NewGRPCClient(config)
+	if err != nil {
+		return nil, err
+	}
+
+	return client.NewConnection(url)
```
sykesm: address != url and `NewConnection` takes an address.

internal/pkg/gateway/server/api.go
```diff
@@ -79,10 +85,33 @@ func (gs *Server) Endorse(ctx context.Context, proposedTransaction *pb.ProposedT
 	return preparedTxn, nil
 }
 
+//go:generate counterfeiter -o mocks/submitserver.go --fake-name SubmitServer . submitServer
+type submitServer interface {
+	pb.Gateway_SubmitServer
+}
+
```
sykesm: looks like this belongs in the test files.

internal/pkg/gateway/registry/registry_test.go
```diff
@@ -7,18 +7,188 @@ SPDX-License-Identifier: Apache-2.0
...
+Ew1TYW4gRnJhbmNpc2NvMRkwFwYDVQQKExBvcmcxLmV4YW1wbGUuY29tMR8wHQYD
+VQQDExZ0bHNjYS5vcmcxLmV4YW1wbGUuY29tMFkwEwYHKoZIzj0CAQYIKoZIzj0D
+AQcDQgAE3SvDM33N+yqTj5fjRcAAaWj8hSLb0eeG6hxWxEcTxI5Gf1tWA/pathfN
+6BiFzqBGSsWMd/X6e4HLvenzGsxW5KNtMGswDgYDVR0PAQH/BAQDAgGmMB0GA1Ud
+JQQWMBQGCCsGAQUFBwMCBggrBgEFBQcDATAPBgNVHRMBAf8EBTADAQH/MCkGA1Ud
+DgQiBCD/jZ/5NK3W/rQ/wlwRGazJF0TrnPkDQzGVlLWGYYqsGzAKBggqhkjOPQQD
+AgNJADBGAiEA78aGl4HuMb96Yy5JGZb6oHxnGntb/wFD/pP8PVZn6oACIQDdKUu2
+lUxu8s3zN5LF5eUcH7bkDfBEH0riJ/N8tqDzTA==
+-----END CERTIFICATE-----
+`
```
sykesm: Please see the many commits added to Fabric to move away from static certificates to certs generated in test and use the `tlsgen` package to create the certs that are needed to a temporary directory that lives for the duration of the test.

internal/pkg/gateway/registry/registry.go
```diff
@@ -7,40 +7,218 @@ SPDX-License-Identifier: Apache-2.0
...
+			}
+		}
+	}
+
+	return orderers, nil
+}
+
+func (reg *Registry) addPeer(mspid string, url string) error {
+	// assumes that the msp registry has already been populated with this mspid
+	if _, ok := reg.peers[url]; !ok && strings.Contains(url, ":") {
```
sykesm: This is an interesting check.  What does a `:` signify? I can guess but it seems the significance should be made explicit.

internal/pkg/gateway/registry/registry.go
```diff
@@ -7,40 +7,218 @@ SPDX-License-Identifier: Apache-2.0
...
+		// get the peers for the channel
+		peers := map[string]string{}
+		members := reg.discovery.PeersOfChannel(common.ChannelID(channel))
+		for _, member := range members {
+			id := member.PKIid.String()
+			peers[id] = member.PreferredEndpoint()
+		}
+		for org, infoset := range reg.discovery.IdentityInfo().ByOrg() {
+			for _, info := range infoset {
+				pkid := info.PKIId.String()
```
sykesm: This is a landmine. We're hoping that the `fmt.Stringer` implementation on a byte slice that is documented as
```
// PKIidType defines the type that holds the PKI-id
// which is the security identifier of a peer
```
(ie. opaque) is going to be an address? And we're assuming that it's going to stay that way?

sykesm (changes requested):
I know this has been out here for a while pending a review and I've started to look at it a couple of times and then put it down because I didn't really know how best to start the conversation.

After spending a few minutes on it this morning, there are a couple of general themes to the comments I've made.

1. The majority of the code here is generated mocks in support of tests that don't test anything.
2. The generated mocks are largely a result of introducing many, many interfaces. These are adding a lot of noise and complexity
3. The tests that are driving mocks aren't even verifying behavior. There are no assertions on arguments, no assertions on calls, and no assertions on error processing.

Please break this PR into something smaller that we can iterate on. The core of this seems to be the registry. I'd like to iterate on that in isolation.

internal/pkg/gateway/endpoint/endpoint.go
```diff
@@ -0,0 +1,186 @@
...
+
+// An Orderer is a minimal abstraction of an ordering node for the purposes of submitting a transaction.
+type Orderer interface {
+	Send(*common.Envelope) error
+	Recv() (*gatewayProto.Event, error)
+}
+
+// A Factory is used to create instances of endorser clients to remote peers and AtomicBroadcast clients
+// to ordering nodes.
+type Factory interface {
```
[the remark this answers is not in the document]
andrew-coleman: Removed and rewritten

internal/pkg/gateway/endpoint/endpoint.go
```diff
@@ -0,0 +1,186 @@
...
+package endpoint
+
+import (
+	"context"
+	"fmt"
+	"io"
+	"time"
+
+	"github.com/hyperledger/fabric-protos-go/common"
+	gatewayProto "github.com/hyperledger/fabric-protos-go/gateway"
```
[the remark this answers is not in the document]
andrew-coleman: ok

internal/pkg/gateway/endpoint/endpoint.go
```diff
@@ -0,0 +1,186 @@
...
+type remote struct {
+	conn connectionProvider
+}
+
+// NewRemoteEndpointFactory creates an endpoint factory to connect to other nodes in the network.
+func NewRemoteEndpointFactory() Factory {
+	return &remote{
+		conn: &remoteConnection{},
+	}
+}
```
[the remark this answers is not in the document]
andrew-coleman: removed

internal/pkg/gateway/endpoint/endpoint.go
```diff
@@ -0,0 +1,186 @@
...
+	if err != nil {
+		return nil, err
+	}
+	response, err := bc.Recv()
+	if err != nil {
+		return nil, err
+	}
+	if err == io.EOF {
+		return nil, err
+	}
```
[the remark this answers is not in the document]
andrew-coleman: removed

internal/pkg/gateway/endpoint/endpoint.go
```diff
@@ -0,0 +1,186 @@
...
+	if err != nil {
+		return nil, fmt.Errorf("failed to to get response from orderer: %w", err)
+	}
```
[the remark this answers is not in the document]
andrew-coleman: removed

internal/pkg/gateway/endpoint/endpoint.go
```diff
@@ -0,0 +1,186 @@
...
+type remoteEndorser struct {
+	client peer.EndorserClient
+}
+
+// ProcessProposal sends a signed proposal to the endorsing peer.
+func (r *remoteEndorser) ProcessProposal(ctx context.Context, in *peer.SignedProposal) (*peer.ProposalResponse, error) {
+	return r.client.ProcessProposal(ctx, in)
+}
+
+//go:generate counterfeiter -o mocks/atomicbroadcastclient.go --fake-name AtomicBroadcastClient . atomicBroadcastClient
```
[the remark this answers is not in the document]
andrew-coleman: thanks

internal/pkg/gateway/endpoint/endpoint.go
```diff
@@ -0,0 +1,186 @@
...
+		return nil, err
+	}
+	if err != nil {
+		return nil, fmt.Errorf("failed to to get response from orderer: %w", err)
+	}
+	if response == nil {
+		return nil, fmt.Errorf("received nil response from orderer")
+	}
+
+	status := response.Info
```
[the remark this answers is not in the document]
andrew-coleman: no good reason - removed

internal/pkg/gateway/endpoint/endpoint.go
```diff
@@ -0,0 +1,186 @@
...
+		Value: []byte(status),
+	}, nil
+}
+
+//go:generate counterfeiter -o mocks/connectionProvider.go --fake-name ConnectionProvider . connectionProvider
+type connectionProvider interface {
+	NewConnection(url string, tlsRootCerts [][]byte) (*grpc.ClientConn, error)
+}
+
+type remoteConnection struct{}
```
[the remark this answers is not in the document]
andrew-coleman: rewritten

internal/pkg/gateway/endpoint/endpoint.go
```diff
@@ -0,0 +1,186 @@
...
+
+func (rc *remoteConnection) NewConnection(url string, tlsRootCerts [][]byte) (*grpc.ClientConn, error) {
+	secOpts := comm.SecureOptions{
```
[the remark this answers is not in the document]
andrew-coleman: rewritten

internal/pkg/gateway/endpoint/endpoint.go
```diff
@@ -0,0 +1,186 @@
...
+		UseTLS:            true,
+		ServerRootCAs:     tlsRootCerts,
+		RequireClientCert: false,
+	}
+	config := comm.ClientConfig{
+		SecOpts: secOpts,
+		Timeout: 5 * time.Second, // TODO from config
```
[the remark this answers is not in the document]
andrew-coleman: understood

internal/pkg/gateway/endpoint/endpoint_test.go
```diff
@@ -0,0 +1,54 @@
+/*
```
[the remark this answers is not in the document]
andrew-coleman: removed and rewritten

internal/pkg/gateway/registry/registry.go
```diff
@@ -7,40 +7,218 @@ SPDX-License-Identifier: Apache-2.0
...
+	logger              *flogging.FabricLogger
+	peers               map[string]endpoint.Endorser
+	orderers            map[string]endpoint.Orderer
+	msps                map[string][]byte
+	channelsInitialized map[string]bool
 }
 
 // Construct a New registry representing the gateway's view of the network.
-func New(localEndorser gateway.Endorser) (*Registry, error) {
+func New(localEndorser endpoint.Endorser, discovery discovery.Support, selfEndpoint string, factory endpoint.Factory) (*Registry, error) {
```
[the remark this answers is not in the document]
andrew-coleman: removed most interfaces now

internal/pkg/gateway/registry/registry.go
```diff
@@ -7,40 +7,218 @@ SPDX-License-Identifier: Apache-2.0
...
-		logger:        logger,
+		localEndorser:       localEndorser,
+		discovery:           discovery,
+		selfEndpoint:        selfEndpoint,
+		endpointFactory:     factory,
+		logger:              logger,
+		peers:               map[string]endpoint.Endorser{},
+		orderers:            map[string]endpoint.Orderer{},
+		msps:                map[string][]byte{},
+		channelsInitialized: map[string]bool{},
```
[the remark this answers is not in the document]
andrew-coleman: oops - I've added a RWMutex to protect these now

internal/pkg/gateway/gateway.go
andrew-coleman: removed interface

internal/pkg/gateway/server/server_test.go
andrew-coleman: removed and rewritten

internal/pkg/gateway/server/api_test.go
andrew-coleman: yep - they're mostly gone now

internal/pkg/gateway/endpoint/endpoint.go
andrew-coleman: ok

internal/pkg/gateway/server/api.go
andrew-coleman: moved

internal/pkg/gateway/registry/registry_test.go
andrew-coleman: thanks - I've replaced this with tlsgen

internal/pkg/gateway/registry/registry.go
andrew-coleman: ah ok - it seems the local (host) peer identifies itself with an empty string.  I've changed the code to be more explicit and added a comment.

internal/pkg/gateway/registry/registry.go
andrew-coleman: It's only using the pkid to correlate the peers into groups of the same mspid.  It makes no assumption as to whether it is an address or not.  I copied this code from the discovery service which does something similar.

andrew-coleman:
Thanks Matt.  The separation of the Registry into a separate package with an interface is something we designed into the standalone gateway with the idea that the embedded gateway and the standalone gateway would each provide their implementations, and the rest of the gateway server code common to both.  However, in the interest of not having 'premature interfaces', I have flattened all the code into a single package, and removed many of the interfaces.  The only one that remains in the code is the 'Endorser' which provides a common interface for the host peer (EndorserServer) and remote peers (EndorserClient).  Another couple of private interfaces are in the test code for the purposes of mocking the endpoint connections.

internal/pkg/gateway/registry/registry.go
yacovm: A PKI-ID is not an address. It's a hash of the serialized identity.

internal/pkg/gateway/api.go
```diff
@@ -70,19 +85,49 @@ func (gs *Server) Endorse(ctx context.Context, proposedTransaction *pb.ProposedT
...
+	if err != nil {
+		return err
+	}
+
+	if len(orderers) == 0 {
+		return fmt.Errorf("no broadcastClients discovered")
+	}
+
+	// send to first orderer for now
+	logger.Info("Submitting txn to orderer")
```
sykesm, line 116: We'll want to revisit this. An info message with no context isn't particularly helpful to anyone.

internal/pkg/gateway/apiutils.go
```diff
@@ -78,6 +78,9 @@ func getValueFromResponse(response *peer.ProposalResponse) (*gateway.Result, err
 }
 
 func getChannelAndChaincodeFromSignedProposal(signedProposal *peer.SignedProposal) (string, string, error) {
+	if signedProposal == nil {
```
sykesm, line 81: I think this should find a home in protoutils/proputils.go (proposal utils). Some of the helpers over there do a fair bit of this and it would make sense to keep this sort of goo in that package since we already have a dependency on it.

internal/pkg/gateway/endpoint.go
```diff
@@ -0,0 +1,52 @@
...
+	conn, err := newConnection(address, tlsRootCerts)
+	if err != nil {
+		return nil, err
+	}
+	abc := ab.NewAtomicBroadcastClient(conn)
+	return abc.Broadcast(context.TODO()) // TODO build a context using timeouts specified in the gateway config (future user story)
+}
+
+func newConnection(address string, tlsRootCerts [][]byte) (*grpc.ClientConn, error) {
+	secOpts := comm.SecureOptions{
```
sykesm, line 37: As we discussed, we're going to want to revisit some of this. I think we need to look at what we can do to the `comm` package to make it easier to reuse what's there.

sykesm (approved):
Let's get this in and plan to revisit a bit of this before moving to the next bit of the implementation.

The test coverage of the server is much better and we're using the mocks to assert on the interactions and behavior so that adds a level of comfort. On the other hand, we have absolutely no coverage on the contents of `endpoint.go` so we need to backfill that.

PR #2410, opened 2021-02-17
Add Throttle to ListOne

lindluni:
The ListOne function queries the Channel Participation API using Gomega Eventually. The default polling interval is 10ms results in many transient http clients being created and on systems with low Open File Limits this results in too many open files. This change adds a one second throttle to the ListOne function.

### DIFF
--- a/integration/channelparticipation/channel_participation.go
+++ b/integration/channelparticipation/channel_participation.go
@@ -13,6 +13,7 @@ import (
 	"io/ioutil"
 	"mime/multipart"
 	"net/http"
+	"time"
 
 	"github.com/golang/protobuf/proto"
 	"github.com/hyperledger/fabric-protos-go/common"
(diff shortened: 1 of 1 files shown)


lindluni:
Looking at the log output of the CP API flake, I think this will fix the flake as well, looks like the flake is caused by the endpoint being overwhelmed.

sykesm:
If we need to throttle, can we throttle in the implementation like we've done for sessions instead of at the call site?

Also, the polling probably isn't the file descriptor leak; it's probably the transient creation of two clients each time the function is called that is creating creating a new transport with an associated idle connection pool.

lindluni:
> Also

Yep, its the clients being created thats causing the issue, thats apparent from the log. I'll change the wording of the commit message.

lindluni:
How far should I wire the throttle. Would you rather I expose the `sessionLastExecuted` map on the `Network` object (and rename it to something more agnostic) and make use of that since I have access to the the network object in the `ListOne` function, or should I just create something locally to manage it.

lindluni:
I added it to the network object, if you'd rather I do it locally, I can change it

lindluni:
Decided to do it locally

guoger:
#2412 for similar reason

lindluni:
#2411 resolved the issue this was attempting to fix, thanks @sykesm

sykesm:
I think there's still value in exposing a throttling mechanism on the network. It's worth considering what that would look like.

Is it Network#Eventually(...)?
Is it exposing a throttle object?
Is it exporting what we have as a function with a wait time or channel instead of a map?

Basically, it may have been premature to close this as it solves a problem - just not necessarily the file descriptor one.

PR #2412, opened 2021-02-18
Enlarge gomega default eventually poll interval for FetchBlock

guoger:
This is to avoid excessive creation of deliver client, that may
exhaust file descriptors on certain system.

### DIFF
--- a/integration/raft/client.go
+++ b/integration/raft/client.go
@@ -25,7 +25,7 @@ func FetchBlock(n *nwo.Network, o *nwo.Orderer, seq uint64, channel string) *com
 		var err error
 		blk, err = ordererclient.Deliver(n, o, denv)
 		return err
-	}, n.EventuallyTimeout).ShouldNot(HaveOccurred())
+	}, n.EventuallyTimeout, n.SessionCreateInterval).ShouldNot(HaveOccurred())
 
 	return blk
 }


sykesm:
The file descriptor issue should be resolved by https://github.com/hyperledger/fabric/pull/2411. Any throttling would simply rate limiting to reduce cpu and, to a lesser extent, memory.

guoger:
> The file descriptor issue should be resolved by https://github.com/hyperledger/fabric/pull/2411. Any throttling would simply rate limiting to reduce cpu and, to a lesser extent, memory.

IIUC, orderer's deliver Api is not throttled, and `FetchBlock` currently creates new client every 10ms, which leaves file descriptors behind as well pollutes test logs, therefore worth reducing IMO.

sykesm:
> orderer's deliver Api is not throttled, and FetchBlock currently creates new client every 10ms, which leaves file descriptors behind

If true, it's a DoS vector that needs to be addressed in the production code (but I'm not sure it's real).

With respect to throttling in the test, I'd suggest you look at the comments I made on #2410. Changing this interval targets one instance and I would prefer a solution at a lower level.

integration/raft/client.go
```diff
@@ -25,7 +25,7 @@ func FetchBlock(n *nwo.Network, o *nwo.Orderer, seq uint64, channel string) *com
 		var err error
 		blk, err = ordererclient.Deliver(n, o, denv)
```
sykesm, line 26: @guoger The creation of the orderer client inside of Deliver is throttled by the network.

Deliver: https://github.com/hyperledger/fabric/blob/76763d1b8900df79080765453f944122dd7d68aa/integration/ordererclient/orderer_client.go#L20
OrdererClientConn: https://github.com/hyperledger/fabric/blob/76763d1b8900df79080765453f944122dd7d68aa/integration/nwo/network.go#L1486-L1501

PR #2414, opened 2021-02-18
[FAB-18427] Report correct reason of stream abort in orderer cluster

yacovm:
This PR fixes a bug that makes the cluster communication infrastructure
always report an "aborted" reason after a stream terminates.

The reason for the bug is that the serviceStream() method was always
overriding the real reason the stream was terminated, with the same "aborted" reason.

Moving the stream termination reason to reside inside the sync.Once block along with
the rest of the termination logic solved this bug.

This PR also contains a commit that adds a log for the stream termination and its lifetime duration, that is useful on its own, but is also used in the test of the second commit.

### DIFF
--- a/orderer/common/cluster/comm_test.go
+++ b/orderer/common/cluster/comm_test.go
@@ -522,6 +522,55 @@ func TestUnavailableHosts(t *testing.T) {
 	require.Contains(t, err.Error(), "connection")
 }
 
+func TestStreamAbortReportCorrectError(t *testing.T) {
+	// Scenario: node 1 acquires a stream to node 2 and then the stream
+	// encounters an error and as a result, the stream is aborted.
+	// We ensure the error reported is the first error, even after
+	// multiple attempts of using it.
+
+	node1 := newTestNode(t)
+	defer node1.stop()
+
+	node2 := newTestNode(t)
+	defer node2.stop()
+
+	node1.c.Configure(testChannel, []cluster.RemoteNode{node2.nodeInfo})
+	node2.c.Configure(testChannel, []cluster.RemoteNode{node1.nodeInfo})
+
+	node2.handler.On("OnSubmit", testChannel, node1.nodeInfo.ID, mock.Anything).Return(errors.Errorf("whoops")).Once()
+
+	rm1, err := node1.c.Remote(testChannel, node2.nodeInfo.ID)
+	require.NoError(t, err)
+	var streamTerminated sync.WaitGroup
+	streamTerminated.Add(1)
+
+	stream := assertEventualEstablishStream(t, rm1)
+
+	l, err := zap.NewDevelopment()
+	require.NoError(t, err)
+	stream.Logger = flogging.NewFabricLogger(l, zap.Hooks(func(entry zapcore.Entry) error {
+		if strings.Contains(entry.Message, "Stream 1 to") && strings.Contains(entry.Message, "terminated") {
+			streamTerminated.Done()
+		}
+		return nil
+	}))
+
+	// Probe the stream for the first time
+	err = stream.Send(wrapSubmitReq(testReq))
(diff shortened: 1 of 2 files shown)


orderer/common/cluster/comm_test.go
```diff
@@ -522,6 +522,55 @@ func TestUnavailableHosts(t *testing.T) {
...
+	// Probe the stream for the first time
+	err = stream.Send(wrapSubmitReq(testReq))
+	require.NoError(t, err)
+
+	// We should receive back the crafted error
+	_, err = stream.Recv()
+	require.Contains(t, err.Error(), "whoops")
+
+	// Wait for the stream to be aborted from within the communication infrastructure
+	streamAborted.Wait()
```
jyellick: I don't think this necessarily could cause a test flake, but I don't think this step captures exactly what we want.  The log message is emitted by that go routine before the Cancel is called, so there's no guarantee we've actually hit that cancel before performing the below check.

orderer/common/cluster/comm_test.go
```diff
@@ -522,6 +522,55 @@ func TestUnavailableHosts(t *testing.T) {
...
+	require.NoError(t, err)
+	var streamAborted sync.WaitGroup
+	streamAborted.Add(1)
+
+	stream := assertEventualEstablishStream(t, rm1)
+
+	l, err := zap.NewDevelopment()
+	require.NoError(t, err)
+	stream.Logger = flogging.NewFabricLogger(l, zap.Hooks(func(entry zapcore.Entry) error {
+		if strings.Contains(entry.Message, "Stream 1 to") && strings.Contains(entry.Message, "terminating") {
```
sykesm: Not a fan of using log messages to trigger an event.

sykesm (dismissed):
I don't expect you to remove the logging hook but, based on my experience, those hooks come back to bite.

jyellick (commented):
Thanks for the fix @yacovm, find one note below.

orderer/common/cluster/comm_test.go
yacovm: Got a better idea?

sykesm: Given the code, nope. Which is why I approved as written.

orderer/common/cluster/comm_test.go
```diff
@@ -522,6 +522,55 @@ func TestUnavailableHosts(t *testing.T) {
...
+	defer node2.stop()
+
+	node1.c.Configure(testChannel, []cluster.RemoteNode{node2.nodeInfo})
+	node2.c.Configure(testChannel, []cluster.RemoteNode{node1.nodeInfo})
+
+	node2.handler.On("OnSubmit", testChannel, node1.nodeInfo.ID, mock.Anything).Return(errors.Errorf("whoops")).Once()
+
+	rm1, err := node1.c.Remote(testChannel, node2.nodeInfo.ID)
+	require.NoError(t, err)
+	var streamAborted sync.WaitGroup
```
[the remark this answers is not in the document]
yacovm: Yes, but a wait group conveys better what we want to achieve, which is to wait :)

orderer/common/cluster/comm_test.go
yacovm: Thanks, I had them at 2 defers and then moved to one defer func() but forgot the change the order :)

yacovm: Thanks, I had them at 2 defers (which execute in reverse order due to stacking) and then moved to one defer func() but forgot the change the order :)

This cannot cause a flake, but it can cause a false negative.

jyellick (dismissed):
Thanks!

jyellick (approved):

yacovm:

yacovm:

yacovm:

PR #2424, opened 2021-02-19
integration: Remove yaml serialization of network configurations

sykesm:
When `nwo` was first created, we expected the network model to be be read from a file instead of being constructed at runtime so we included an example that used that mechanism. We've since learned that we can get a way with a small number of "standard" networks (with minor deviations) and that the serialization is not required.

This change removes network definition fixtures from the integration tests and removes the example from `nwo`.


integration/nwo/standard_networks.go
```diff
@@ -76,6 +76,48 @@ func BasicSolo() *Config {
 	}
 }
 
+// ThreeOrgSolo returns a simple configuration with three organizations isntead
+// of two.
```
lindluni: sp: instead

lindluni (dismissed):

lindluni (approved):

PR #2427, opened 2021-02-19
Add Alpine Version to Tools Image

lindluni:
`ALPINE_VERSION` was removed for both the intermediate layer and the final layer, a previous commit added it back to the intermediate layer but missed adding it back to the final layer.

### DIFF
--- a/images/tools/Dockerfile
+++ b/images/tools/Dockerfile
@@ -21,7 +21,7 @@ FROM golang as tools
 ARG GO_TAGS
 RUN make configtxgen configtxlator cryptogen peer discover osnadmin idemixgen GO_TAGS=${GO_TAGS}
 
-FROM golang:${GO_VER}-alpine
+FROM golang:${GO_VER}-alpine${ALPINE_VER}
 # git is required to support `go list -m`
 RUN apk add --no-cache \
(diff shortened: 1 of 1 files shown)


denyeart (approved):
Thanks.

I assume we should move to Alpine 3.13? I see `golang:1.15.7-alpine3.13` is available.

lindluni:
> Thanks.
> 
> I assume we should move to Alpine 3.13? I see `golang:1.15.7-alpine3.13` is available.

Bumped in #2428

denyeart:
> > Thanks.
> > I assume we should move to Alpine 3.13? I see `golang:1.15.7-alpine3.13` is available.
> 
> Bumped in #2428

Cool. Looks like the Dockerfile change is also in #2428

lindluni:
> > > Thanks.
> > > I assume we should move to Alpine 3.13? I see `golang:1.15.7-alpine3.13` is available.
> > 
> > 
> > Bumped in #2428
> 
> Cool. Looks like the Dockerfile change is also in #2428

Yep, I didn't have time to test everything locally so I pushed it with this commit on it as well just to make sure the tests look good. It wont matter in the end.

PR #2429, opened 2021-02-19
Add methods to obtain grpc.ClientConn instances configured to connect to an orderer or peer

sykesm:
Add methods to the network to create gRPC client connections to the orderer and peer processes that do no rely on internal packages. This also moves common functionality to the network.

### DIFF
--- a/integration/ledger/snapshot_test.go
+++ b/integration/ledger/snapshot_test.go
@@ -8,11 +8,11 @@ package ledger
 
 import (
 	"context"
-	"crypto/tls"
 	"encoding/base64"
 	"encoding/json"
 	"fmt"
 	"io/ioutil"
+	"math"
 	"os"
 	"path/filepath"
 	"strconv"
@@ -30,7 +30,6 @@ import (
 	"github.com/hyperledger/fabric/integration/nwo/commands"
 	"github.com/hyperledger/fabric/integration/nwo/runner"
 	"github.com/hyperledger/fabric/integration/pvtdata/marblechaincodeutil"
-	ic "github.com/hyperledger/fabric/internal/peer/chaincode"
 	"github.com/hyperledger/fabric/protoutil"
 	. "github.com/onsi/ginkgo"
 	. "github.com/onsi/gomega"
@@ -41,7 +40,7 @@ import (
 
 const testchannelID = "testchannel"
 
-var _ bool = Describe("Snapshot Generation and Bootstrap", func() {
+var _ = Describe("Snapshot Generation and Bootstrap", func() {
 	var (
 		setup                 *setup
 		helper                *marblesTestHelper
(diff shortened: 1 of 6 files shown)


integration/ledger/snapshot_test.go
```diff
@@ -1038,59 +1036,69 @@ func invokeWithoutPassingOrdererEndPoint(n *nwo.Network, peer *nwo.Peer, channel
...
-		KaOpts:  comm.DefaultKeepaliveOptions,
-		SecOpts: comm.SecureOptions{
-			UseTLS:        true,
-			ServerRootCAs: [][]byte{caPEM},
-		},
-	})
-	Expect(err).NotTo(HaveOccurred())
-
-	return gClient
+	for {
```
denyeart, line 1085: What if the deliver event never arrives? Do we need a timeout?

integration/lifecycle/lifecycle_suite_test.go
```diff
@@ -157,83 +155,58 @@ func SignedProposal(channel, chaincode string, signer *nwo.SigningIdentity, args
...
-	err := dg.Connect(ctx)
+	err = df.Send(deliverEnvelope)
 	Expect(err).NotTo(HaveOccurred())
 
-	err = oc.Send(env)
+	err = oc.Send(tx)
 	Expect(err).NotTo(HaveOccurred())
 
-	return dg.Wait(ctx)
+	for {
```
denyeart, line 191: Same as above.

integration/ledger/snapshot_test.go
```diff
@@ -1038,59 +1036,69 @@ func invokeWithoutPassingOrdererEndPoint(n *nwo.Network, peer *nwo.Peer, channel
 
 // commitTx commits a transaction for a given transaction envelope
 func commitTx(n *nwo.Network, orderer *nwo.Orderer, peer *nwo.Peer, channelID string, env *cb.Envelope, txid string) error {
```
denyeart, line 1038: I like how you changed "env" to "tx" in the other CommitTx() further down, could you do the same here for consistency.

denyeart (commented):
Easier to understand, I like it, just one question.

integration/ledger/snapshot_test.go
```diff
@@ -1037,25 +1036,43 @@ func invokeWithoutPassingOrdererEndPoint(n *nwo.Network, peer *nwo.Peer, channel
...
-	var cancelFunc context.CancelFunc
-	ctx, cancelFunc := context.WithTimeout(context.Background(), 10*time.Second)
-	defer cancelFunc()
+	ctx, cancel := context.WithTimeout(context.Background(), 10*time.Second)
```
sykesm, line 1039: This configures the timeout on the `DeliverFiltered` service. The context will be canceled when if the timeout occurs `Recv` will return with a `context.Canceled` error that is percolated out.

integration/ledger/snapshot_test.go
sykesm: https://github.com/hyperledger/fabric/pull/2429/commits/4ce8b576f6ab1828152764d2ac5a9f5c214a6529#r579672777

integration/ledger/snapshot_test.go
denyeart: Great, thanks.

denyeart (approved):

PR #2441, opened 2021-02-24
Add tools target for the tools docker image

sykesm:
FAB-16348:

Today the fabric tools Dockerfile calls `RUN make configtxgen configtxlator cryptogen peer discover idemixgen`. This should be changed to point to a make target that encapsulates all of these items so we don't have to make changes in multiple places in the future.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -211,6 +211,9 @@ protos: gotool.protoc-gen-go
 .PHONY: native
 native: $(RELEASE_EXES)
 
+.PHONY: tools
+tools: $(TOOLS_EXES)
+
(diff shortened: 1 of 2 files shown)


lindluni (approved):

sykesm:
Hit https://jira.hyperledger.org/browse/FAB-18365; rerunning

sykesm:
And https://jira.hyperledger.org/browse/FAB-16233; rerunning

Two test flakes and no code changes. 🍀

denyeart (approved):

PR #2442, opened 2021-02-24
Remove gossipSupport from the peer package

sykesm:
Replace the "gossipSupport" object used to expose configuration through the `Config` interface with a simple `ConfigUpdate` value object. The interface is not necessary to access static data from a configuration update.

FAB-15576


jyellick (approved):
LGTM

PR #2444, opened 2021-02-25
FABGW-8: Capability to observe transaction commits in ledger

bestbeforetoday:
First step in implementing embedded gateway capability to wait for commit of transactions after submit to the orderer.


internal/pkg/gateway/commit/transactionstatus.go
```diff
@@ -0,0 +1,100 @@
...
+				return
+			}
+		}
+	}()
+
+	return out
+}
+
+func findTransactionStatus(block *common.Block, transactionID string) (peer.TxValidationCode, bool) {
+	envelopes := block.Data.Data
```
bestbeforetoday: Does this need nil checks for blocks committed to the peer ledger?

jyellick: Internally in the peer, the nil check makes little sense.  If you want this code to be portable outside the peer, then it would be a good idea.  My feeling is that we omit the check for now.

internal/pkg/gateway/commit/transactionstatus.go
```diff
@@ -0,0 +1,100 @@
...
+	"github.com/hyperledger/fabric-protos-go/common"
+	"github.com/hyperledger/fabric-protos-go/peer"
+	"github.com/hyperledger/fabric/common/ledger/blockledger"
+	"github.com/pkg/errors"
+)
+
+// TransactionStatus is sent to the returned channel when received from the block iterator. The returned channel will
+// be closed after the status is sent, or if an error occurs that prevents the transaction status from being obtained.
+// The receiver should always check the the closed status of the channel on reads.
+func TransactionStatus(blockIterator blockledger.Iterator, transactionID string, quit <-chan int) <-chan peer.TxValidationCode {
```
jyellick: Since this is not wired, let's go ahead and add the `ctx context.Context` parameter and check for the context ending in the select?  Two modified lines in the function, and a new test, so not too intrusive.

bestbeforetoday: I've pushed up a change with what I think you're asking for. The only difference is the code now needs to pull the done channel out of the context argument rather than having the done channel passed in as an argument. Since the code doesn't require anything else from the context, I'm not convinced it's an improvement

bestbeforetoday:
This is very much work in progress so closing this PR for now to avoid people wasting time reviewing code that will significantly change

PR #2448, opened 2021-02-26
Replace WaitGroups with channels and timeouts

sykesm:
A `sync.WaitGroup` does not have a timeout mechanism. If a test is told to `Wait` on a `WaitGroup` and the required number of `Done` calls never get made, the test will hang until the testing package finally kills it. When that happens, we get some stack traces but we lose a lot of time and a fair bit of information.

After hitting the 20m test time out in common/cluster three times in one week, I decided it's time to add a proper timeout.

This commit switches from a WaitGroup to a channel, a sync.Once, and a select on the channel close and a timeout condition. This should result in test failures instead of termination from the testing package.

### DIFF
--- a/orderer/common/cluster/deliver_test.go
+++ b/orderer/common/cluster/deliver_test.go
@@ -654,19 +654,23 @@ func TestBlockPullerFailover(t *testing.T) {
 
 	// Configure the block puller logger to signal the wait group once the block puller
 	// received the first block.
-	var pulledBlock1 sync.WaitGroup
-	pulledBlock1.Add(1)
+	pulledBlock1Done := make(chan struct{})
 	var once sync.Once
 	bp.Logger = bp.Logger.WithOptions(zap.Hooks(func(entry zapcore.Entry) error {
 		if strings.Contains(entry.Message, "Got block [1] of size") {
-			once.Do(pulledBlock1.Done)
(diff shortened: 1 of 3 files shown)


jyellick (approved):
Failing fast is an improvement at least.

orderer/common/cluster/deliver_test.go
```diff
@@ -654,19 +654,23 @@ func TestBlockPullerFailover(t *testing.T) {
...
 		}
 		return nil
 	}))
 
 	go func() {
 		// Wait for the block puller to pull the first block
-		pulledBlock1.Wait()
+		select {
+		case <-pulledBlock1Done:
+		case <-time.After(10 * time.Second):
```
mastersingh24, line 670: Did you choose 10 seconds as it really should never take longer than that?  I imagine when the test "works", it should be pretty quick?

mastersingh24 (commented):
I like this.  The current pattern does not work well.

orderer/common/cluster/deliver_test.go
sykesm: When the tests work, they take under a second so I went with something an order of magnitude longer. If the times need to change, it will be easy enough to see.

The real goal is to catch the source of the hang in one of the `TestBlockPullerBadBlocks` sub-tests and try to fix it.

PR #2451, opened 2021-02-26
comm: expose gRPC dialer options from ClientConfig to enable reuse in multiple contexts

sykesm:
This PR contains a number of small commits that:
- Use `comm.ClientConfig` as the source of dial options
- Remove the unnecessary `comm.GRPCClient` type
- Remove TLS option callbacks and replace with fields on `comm.SecureOptions`
- Remove dead fixtures
- Remove exported functions and methods that are not called in production code
- Merge the set of gRPC security context related helpers from comm with those in common/util
- Move infrastructure with a single, specialized consumer to its consumer

The extraction of the dial options will be used by the gateway.


jyellick (approved):
This one is obviously a bit of a monster, but, going through commit by commit there's nothing objectionable, and in fact all of the changes seem quite rote.  Seems to remove a lot of unnecessary indirection and consequently unnecessary code and otherwise doesn't change anything functionally, looks good!

PR #2453, opened 2021-03-02
FABGW-6 Use concurrent go routines for endorsement

andrew-coleman:
Replace the current sequential logic for getting multiple endorsements with concurrent fan-out/fan-in using go routines
Add configurable endorsement timeout option to the global config and pass through to gateway

### DIFF
--- a/core/peer/config.go
+++ b/core/peer/config.go
@@ -273,8 +273,7 @@ func (c *Config) load() error {
 		c.DeliverClientKeepaliveOptions.ClientTimeout = viper.GetDuration("peer.keepalive.deliveryClient.timeout")
 	}
 
-	c.GatewayOptions = gateway.DefaultOptions()
-	c.GatewayOptions.Enabled = viper.GetBool("peer.gateway.enabled")
+	c.GatewayOptions = gateway.GetOptions(viper.GetViper())
 
 	c.VMEndpoint = viper.GetString("vm.endpoint")
 	c.VMDockerTLSEnabled = viper.GetBool("vm.docker.tls.enabled")
--- a/internal/peer/node/start.go
+++ b/internal/peer/node/start.go
@@ -825,7 +825,12 @@ func serve(args []string) error {
 			logger.Info("Starting peer with Gateway enabled")
 			gatewayprotos.RegisterGatewayServer(
 				peerServer.Server(),
-				gateway.CreateServer(&gateway.EndorserServerAdapter{Server: serverEndorser}, discoveryService, peerInstance.GossipService.SelfMembershipInfo().Endpoint),
+				gateway.CreateServer(
+					&gateway.EndorserServerAdapter{Server: serverEndorser},
+					discoveryService,
+					peerInstance.GossipService.SelfMembershipInfo().Endpoint,
+					coreConfig.GatewayOptions,
+				),
 			)
 		} else {
 			logger.Warning("Discovery service must be enabled for embedded gateway")
--- a/internal/pkg/gateway/api.go
+++ b/internal/pkg/gateway/api.go
@@ -9,11 +9,17 @@ package gateway
 import (
 	"context"
 	"fmt"
+	"sync"
 
 	gp "github.com/hyperledger/fabric-protos-go/gateway"
 	"github.com/hyperledger/fabric-protos-go/peer"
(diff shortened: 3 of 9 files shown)


core/peer/config.go
```diff
@@ -274,7 +274,12 @@ func (c *Config) load() error {
 	}
 
 	c.GatewayOptions = gateway.DefaultOptions()
-	c.GatewayOptions.Enabled = viper.GetBool("peer.gateway.enabled")
+	if viper.IsSet("peer.gateway.enabled") {
```
sykesm: Since this is new code, the logic to create the gateway configuration should live in the gateway code. If possible (and I think we're finally at a level of the code that supports it), the viper instance should be passed to the constructor to avoid dealing with global state.

internal/peer/node/start.go
```diff
@@ -825,7 +825,7 @@ func serve(args []string) error {
 			logger.Info("Starting peer with Gateway enabled")
 			gatewayprotos.RegisterGatewayServer(
 				peerServer.Server(),
-				gateway.CreateServer(&gateway.EndorserServerAdapter{Server: serverEndorser}, discoveryService, peerInstance.GossipService.SelfMembershipInfo().Endpoint),
+				gateway.CreateServer(&gateway.EndorserServerAdapter{Server: serverEndorser}, discoveryService, peerInstance.GossipService.SelfMembershipInfo().Endpoint, coreConfig.GatewayOptions),
```
sykesm: It's time to break this across multiple lines.

internal/pkg/gateway/api.go
```diff
@@ -66,13 +72,28 @@ func (gs *Server) Endorse(ctx context.Context, proposedTransaction *gp.ProposedT
...
-		response, err := endorser.ProcessProposal(ctx, signedProposal)
-		if err != nil {
-			return nil, fmt.Errorf("failed to process proposal: %w", err)
+		go func(endorser peer.EndorserClient) {
+			response, err := endorser.ProcessProposal(ctx, signedProposal)
+			c <- &endorserResponse{pr: response, err: err}
+		}(endorser)
+	}
+
+	timeout := time.After(gs.options.EndorsementTimeout)
```
sykesm: Under load, this can result in significant memory pressure. This problem is highlighted in the doc: `The underlying Timer is not recovered by the garbage collector until the timer fires`

That said, why do this? Why not decorate the context with a timeout and use the same one across all of the calls? (I'm guessing you copied this code from somewhere...) Once you know the calls have a proper timeout and cancelation, you could choose to keep using a select in a for loop or move to something simpler (like a WaitGroup) to wait for the calls to complete.

internal/pkg/gateway/api_test.go
```diff
@@ -74,6 +76,7 @@ type contextKey string
 func TestGateway(t *testing.T) {
 	const testChannel = "test_channel"
 	const testChaincode = "test_chaincode"
+	const endorsementTimeout = 1 * time.Second
```
sykesm: no need to multiply by `1`

internal/pkg/gateway/api_test.go
```diff
@@ -284,6 +292,20 @@ func TestGateway(t *testing.T) {
...
 			},
+			{
+				name: "endorsement times out",
+				plan: endorsementPlan{
+					"g1": {"localhost:7051"},
+				},
+				signedProposal: validSignedProposal,
+				setupRegistry: func(reg *registry) {
+					reg.localEndorser.(*mocks.EndorserClient).ProcessProposalStub = func(context.Context, *peer.SignedProposal, ...grpc.CallOption) (*peer.ProposalResponse, error) {
+						time.Sleep(endorsementTimeout + 10*time.Millisecond)
```
sykesm: We should rarely see a sleep in tests because there's no way to preempt them.

internal/pkg/gateway/api.go
andrew-coleman: rewritten to use context.WithTimeout, as suggested

internal/pkg/gateway/api_test.go
andrew-coleman: this test is now redundant - removed

sykesm (approved):

PR #2457, opened 2021-03-03
Update Docker Publish Script

lindluni:
We no longer publish multi-arch images, remove the manifest tool and update the script to properly publish two-digit releases.


sykesm (approved):

denyeart (approved):
Thanks, this will resolve the final issue in the release job.

PR #2459, opened 2021-03-04
Add Secondary CI on GitHub Actions

lindluni:
As a stepping stone to migrating from Azure Pipelines to GitHub Actions, this PR enables CI to run on both AZP and GHA. Once GHA has been evaluated and any concerns addressed, we would later disable AZP and run exclusively on GHA, assuming we can reach agreement on GHA.


davidkhala:
@lindluni Dear Bret, would you mind provide some public information about our discussoin on decision to migarete from AZP to GHA?

lindluni:
Will send an overview to the mailing list, as well as split unit tests and integration tests out into separate workflows to allow individual triggering

PR #2470, opened 2021-03-09
Provide commit events from ledger

manish-sethi:

#### Type of change
- New feature

#### Description

This PR introduces a function in ledger that provides a channel that can be to receive the commit events. This function is intended to be used by the embedded gateway to notify the clients that are waiting for a transaction to be committed.

### DIFF
--- a/core/ledger/kvledger/kv_ledger.go
+++ b/core/ledger/kvledger/kv_ledger.go
@@ -887,6 +887,16 @@ func (l *kvLedger) GetMissingPvtDataTracker() (ledger.MissingPvtDataTracker, err
 	return l, nil
 }
 
+// CommitNotificationsChannel returns a read-only channel on which ledger sends a `CommitNotification`
+// when a block is committed. The CommitNotification contains entries for the transactions from the committed block,
+// which are not malformed, carry a legitimate TxID, and in addition, are not marked as a duplicate transaction.
+// The consumer can close the 'done' channel to signal that the notifications are no longer needed. This will cause the
+// CommitNotifications channel to close. There is expected to be only one consumer at a time. The function returns error
+// if already a CommitNotification channel is active.
+func (l *kvLedger) CommitNotificationsChannel(done <-chan struct{}) (<-chan *ledger.CommitNotification, error) {
+	return nil, errors.New("yet-to-implement")
+}
+
 // Close closes `KVLedger`.
 // Currently this function is only used by test code. The caller should make sure no in-progress commit
 // or snapshot generation before calling this function. Otherwise, the ledger may have unknown behavior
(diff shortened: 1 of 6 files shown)


core/ledger/kvledger/kv_ledger.go
```diff
@@ -887,6 +887,16 @@ func (l *kvLedger) GetMissingPvtDataTracker() (ledger.MissingPvtDataTracker, err
 	return l, nil
 }
 
+// CommitNotificationsChannel returns a read-only channel on which ledger sends a `CommitNotification`
+// when a block is committed. The CommitNotification contains entries for the transactions from the committed block,
+// which are not malformed, carry a legitimate TxID, and in addition, are not marked as a duplicate transaction.
+// The consumer can send a struct{} to the signal channel when the notifications are no longer needed. This will cause the
+// CommitNotifications channel to close. There is expected to be only one consumer at a time. The function returns error
+// if already a CommitNotification channel is active.
+func (l *kvLedger) CommitNotificationsChannel() (<-chan *ledger.CommitNotification, chan<- struct{}, error) {
```
sykesm: Why return a channel for the consumer to close instead of accepting the channel from the consumer? Or support passing a context that provides the same signal via `Done`?

IMO,if it's the caller's job to close the channel, the caller should provide the channel.

manish-sethi: Yes, agreed. Fixed in the new patch.

core/ledger/ledger_interface.go
```diff
@@ -215,6 +215,13 @@ type PeerLedger interface {
 	CancelSnapshotRequest(height uint64) error
 	// PendingSnapshotRequests returns a list of heights for the pending (or under processing) snapshot requests.
 	PendingSnapshotRequests() ([]uint64, error)
+	// CommitNotificationsChannel returns a read-only channel on which ledger sends a `CommitNotification`
+	// when a block is committed. The CommitNotification contains entries for the transactions from the committed block,
+	// which are not malformed, carry a legitimate TxID, and in addition, are not marked as a duplicate transaction.
+	// The consumer can close the 'done' channel to signal that the notifications are no longer needed. This will cause the
+	// CommitNotifications channel to close. There is expected to be only one consumer at a time. The function returns error
+	// if already a CommitNotification channel is active.
+	CommitNotificationsChannel(done <-chan struct{}) (<-chan *CommitNotification, error)
```
cendhu, line 224: A better approach than the existing state listeners.

cendhu (approved):

core/ledger/ledger_interface.go
manish-sethi: I don't think that the both are comparable. The existing listeners (e.g., to create couchdb index, update confighistory, and update lifecycle cache are mainly designed to process as part of the transaction. E.g., if they fail, transaction fails. So, they are inline processing. In this case, it's a producer-consumer model, so channel fits naturally.

cendhu: Agree. I raised this point because as soon as I read the commit notification I was wondering whether we can reuse the state listener. However, I understood that it would not fit here. Further, the channel approach looked better than the callback. 

Yes `HandleStateUpdates()` returns an error but it is not required in this notification channel.

PR #2471, opened 2021-03-10
Begin untangling the local MSP singleton - Part I

sykesm:


jyellick (approved):
Definitely an improvement

PR #2475, opened 2021-03-11
Gateway gRPC functions to return google.rpc.Status

andrew-coleman:
Enhances the error reporting in the Gateway.
All gRPC functions return google.rpc.Status style message to the calling client.
If the error is as a result to a failed grpc call to an external peer or orderer, then that/those error(s) are added to the ‘details’ field of the Status message.
If it is caused by a non-success status code in the chaincde response message, then this is added to the details field and returned as an error to the caller.

### DIFF
--- a/internal/pkg/gateway/api.go
+++ b/internal/pkg/gateway/api.go
@@ -11,64 +11,80 @@ import (
 	"fmt"
 	"sync"
 
+	"github.com/golang/protobuf/proto"
 	gp "github.com/hyperledger/fabric-protos-go/gateway"
 	"github.com/hyperledger/fabric-protos-go/peer"
 	"github.com/hyperledger/fabric/protoutil"
+	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/status"
 )
 
 type endorserResponse struct {
 	pr  *peer.ProposalResponse
-	err error
+	err *gp.EndpointError
 }
 
 // Evaluate will invoke the transaction function as specified in the SignedProposal
 func (gs *Server) Evaluate(ctx context.Context, proposedTransaction *gp.ProposedTransaction) (*gp.Result, error) {
 	if proposedTransaction == nil {
-		return nil, fmt.Errorf("a proposed transaction is required")
+		return nil, status.Error(codes.InvalidArgument, "a proposed transaction is required")
 	}
 	signedProposal := proposedTransaction.Proposal
 	channel, chaincodeID, err := getChannelAndChaincodeFromSignedProposal(proposedTransaction.Proposal)
 	if err != nil {
-		// TODO need to specify status codes
-		return nil, fmt.Errorf("failed to unpack channel header: %w", err)
+		return nil, status.Errorf(codes.InvalidArgument, "failed to unpack transaction proposal: %s", err)
 	}
 
 	endorsers, err := gs.registry.endorsers(channel, chaincodeID)
 	if err != nil {
-		return nil, err
+		return nil, status.Errorf(codes.Unavailable, "%s", err)
 	}
 	if len(endorsers) == 0 {
-		return nil, fmt.Errorf("no endorsing peers found for channel: %s", proposedTransaction.ChannelId)
-	}
-	response, err := endorsers[0].client.ProcessProposal(ctx, signedProposal) // TODO choose suitable peer based on block height, etc (future user story)
-	if err != nil {
-		return nil, fmt.Errorf("failed to evaluate transaction: %w", err)
+		return nil, status.Errorf(codes.NotFound, "no endorsing peers found for channel: %s", proposedTransaction.ChannelId)
 	}
 
-	return getValueFromResponse(response)
+	endorser := endorsers[0] // TODO choose suitable peer based on block height, etc (future user story)
+	response, err := endorser.client.ProcessProposal(ctx, signedProposal)
+	if err != nil {
+		return nil, rpcError(
+			codes.Aborted,
+			"failed to evaluate transaction",
+			&gp.EndpointError{Address: endorser.address, MspId: endorser.mspid, Message: err.Error()},
+		)
+	}
+
+	value, err := getValueFromResponse(response)
+	if err != nil {
+		return nil, rpcError(
+			codes.Aborted,
+			"transaction evaluation error",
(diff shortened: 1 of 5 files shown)


internal/pkg/gateway/api.go
```diff
@@ -11,64 +11,78 @@ import (
...
 	}
 
 	endorsers, err := gs.registry.endorsers(channel, chaincodeID)
 	if err != nil {
-		return nil, err
+		return nil, status.Errorf(codes.Unavailable, "%s", err)
 	}
 	if len(endorsers) == 0 {
-		return nil, fmt.Errorf("no endorsing peers found for channel: %s", proposedTransaction.ChannelId)
+		return nil, status.Errorf(codes.Unavailable, "no endorsing peers found for channel: %s", proposedTransaction.ChannelId)
```
sykesm: `Unavailable` or `NotFound`? What you have may be the most appropriate but I'm not sure given the text of the message. This applies to the other blocks that do the same thing below.

internal/pkg/gateway/api.go
```diff
@@ -79,32 +93,45 @@ func (gs *Server) Endorse(ctx context.Context, proposedTransaction *gp.ProposedT
...
-			response, err := endorser.client.ProcessProposal(ctx, signedProposal)
-			responseCh <- &endorserResponse{pr: response, err: err}
+			response, err := e.client.ProcessProposal(ctx, signedProposal)
+			var detail *gp.EndpointError
+			if err != nil {
+				detail = &gp.EndpointError{Address: e.address, MspId: e.mspid, Message: err.Error()}
+			}
+			if response.Response.Status < 200 || response.Response.Status >= 400 {
+				detail = &gp.EndpointError{Address: e.address, MspId: e.mspid, Message: fmt.Sprintf("error %d, %s", response.Response.Status, response.Response.Message)}
+			}
```
sykesm: This block looks quite suspect to me. I think the tests need a bit more focus.

internal/pkg/gateway/api_test.go
```diff
@@ -178,9 +203,33 @@ func TestGateway(t *testing.T) {
 				plan: endorsementPlan{
 					"g1": {"localhost:7051"},
 				},
-				signedProposal:       validSignedProposal,
-				processProposalError: fmt.Errorf("mumbo-jumbo"),
-				errString:            "mumbo-jumbo",
+				signedProposal: validSignedProposal,
+				endpointDefinition: &endpointDef{
+					proposalError: status.Error(codes.Aborted, "mumbo-jumbo"),
```
sykesm, line 212: The same error is being used in multiple places; this complicated my sniff test of some of your error logic earlier. It would be better to have unique strings.

sykesm (changes requested):
- Error handling is needed for the status responses with detailed any's
- There's a clear bug in the code that is not being caught by the tests. This means the code and the tests are suspect.

internal/pkg/gateway/api.go
```diff
@@ -120,33 +147,39 @@ func (gs *Server) Endorse(ctx context.Context, proposedTransaction *gp.ProposedT
 // once the transaction is committed on a sufficient number of remoteEndorsers according to a defined policy.
 func (gs *Server) Submit(txn *gp.PreparedTransaction, cs gp.Gateway_SubmitServer) error {
 	if txn == nil {
-		return fmt.Errorf("a signed prepared transaction is required")
+		return status.Error(codes.InvalidArgument, "a signed prepared transaction is required")
 	}
 	if cs == nil {
-		return fmt.Errorf("a submit server is required")
+		return status.Error(codes.InvalidArgument, "a submit server is required")
```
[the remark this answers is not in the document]
andrew-coleman: True, I've changed it to Internal.

internal/pkg/gateway/api.go
andrew-coleman: Changed to NotFound

internal/pkg/gateway/api.go
andrew-coleman: Which bit is suspect?  If either the grpc call to the remote endorser failed (i.e. returned non-nil error) or the chaincode function returned a non-success status code, then it adds the error detail to the response.
If any of the endorsers produce an error like this, then the whole Endorse() function will abort and return the details to the caller.  It wouldn't be able to continue because the subsequent call to `protoutil.CreateTx` will perform the same check and fail.
I think I'm testing both of these scenarios in the unit tests.
Or have I misunderstood your comment?

sykesm: Please take a look at the client.

https://github.com/hyperledger/fabric-protos-go/blob/e08edaab0493270bf5b386b3d458036fad2b0000/peer/peer.pb.go#L68-L75

internal/pkg/gateway/api.go
```diff
@@ -79,32 +95,47 @@ func (gs *Server) Endorse(ctx context.Context, proposedTransaction *gp.ProposedT
...
+			if err != nil {
+				detail = &gp.EndpointError{Address: e.address, MspId: e.mspid, Message: err.Error()}
+				responseCh <- &endorserResponse{err: detail}
+				return
+			}
+			if response.Response.Status < 200 || response.Response.Status >= 400 {
+				// this is an error case and will be returned in the error details to the client
+				detail = &gp.EndpointError{Address: e.address, MspId: e.mspid, Message: fmt.Sprintf("error %d, %s", response.Response.Status, response.Response.Message)}
+			}
+			responseCh <- &endorserResponse{pr: response, err: detail}
```
sykesm, line 111: This could be easer to read with a helper and a switch. Can you please address this in a follow-up?

```go
response, err := e.client.ProcessProposal(ctx, signedProposal)
switch {
    case err != nil:
        responseCh <- &endorserResponse{err: endpointError(e, err)}
    case response.Response.Status < 200 || response.Response.Status >= 400:
        responseCh <- &endorserResponse{err: endpointError(e, fmt.Errorf("...", ...))
    default:
        responseCh <- &endorserResponse{pr: response}
}
```

sykesm (approved):
I think the tests are getting fairly complicated and I know that there are other PRs coming in on top of this. It's probably worth spending some time looking to see if there are ways to simplify it before too much more is added.

PR #2477, opened 2021-03-12
FABGW-8: Wait for transaction commits

bestbeforetoday:
Initial implementation of commit waiting capability for the embedded Gateway. Not yet integrated into Gateway services.

Signed-off-by: Mark S. Lewis <mark_lewis@uk.ibm.com>


manish-sethi (commented):
Overall LGTM. A few comments - all should be easy to fix. "nit" comments are just a suggestion if you like.

internal/pkg/gateway/commit/notifier_test.go
```diff
@@ -0,0 +1,242 @@
...
+				},
+			}
+			actual := <-commitReceive
+
+			expected := commit.Notification{
+				BlockNumber:    1,
+				TransactionID:  "TX_ID",
+				ValidationCode: peer.TxValidationCode_MVCC_READ_CONFLICT,
+			}
+			require.EqualValues(t, expected, actual)
```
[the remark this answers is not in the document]
bestbeforetoday: There is already a separate test that the channel is closed after a commit is delivered

internal/pkg/gateway/commit/channelnotifier.go
```diff
@@ -0,0 +1,120 @@
...
+	return notifier
+}
+
+func (notifier *channelLevelNotifier) run() {
+	for {
+		blockCommit, ok := <-notifier.commitChannel
+		if !ok {
+			break
+		}
+
```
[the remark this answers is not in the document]
bestbeforetoday: I don't think this is ever supposed to happen (other than when the whole peer process terminates), and there is some tension here with the suggestion to clean up the goroutines at the end of the test by closing the `commitSend` channel. I think I'd like to leave this as just a `break` to aid test (and runtime) termination clean up. I am certainly not keen to implement behaviour if it isn't actually required at runtime

internal/pkg/gateway/commit/channelnotifier.go
```diff
@@ -0,0 +1,120 @@
...
+			}
+
+			listeners[i].close()
+
+			lastIndex := len(listeners) - 1
+			listeners[i] = listeners[lastIndex]
+			listeners = listeners[:lastIndex]
+		}
+
+		notifier.listeners[key] = listeners
```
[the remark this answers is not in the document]
bestbeforetoday: I'd already pushed an update with an equivalent implementation before your review. This is a performance / implementation detail that is not testable externally to the implementation

internal/pkg/gateway/commit/notifier.go
```diff
@@ -0,0 +1,79 @@
...
+type NotificationSupplier interface {
+	CommitNotifications(done <-chan struct{}, channelName string) (<-chan *ledger.CommitNotification, error)
+}
+
+// Notifier provides notification of transaction commits.
+type Notifier struct {
+	lock             sync.Mutex
+	supplier         NotificationSupplier
+	channelNotifiers map[string]*channelLevelNotifier
+}
```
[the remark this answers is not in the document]
bestbeforetoday: 👍

internal/pkg/gateway/commit/channelnotifier.go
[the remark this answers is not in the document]
manish-sethi: Yes, this is not supposed to happen. panic may be too harsh. I would suggest to add `Warning` log at the place (e.g., unexpected channel close, future notifications for this channel will stuck/hang upstream).
About the test, not necessarily in this PR, but good to add a test for this.

internal/pkg/gateway/commit/notifier_test.go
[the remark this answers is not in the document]
manish-sethi: I was just referring to the theme of this test (e.g., even if irrelevant message comes on the channel, this test will pass). Will leave up to you.

internal/pkg/gateway/commit/notifier_test.go
```diff
@@ -0,0 +1,242 @@
...
+
+	t.Run("NewNotifier with nil ledger panics", func(t *testing.T) {
+		f := func() {
+			commit.NewNotifier(nil)
+		}
+		require.Panics(t, f)
+	})
+
+	t.Run("Notify", func(t *testing.T) {
+		t.Run("returns error if channel does not exist", func(t *testing.T) {
```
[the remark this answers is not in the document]
bestbeforetoday: 👍

internal/pkg/gateway/commit/notifier_test.go
```diff
@@ -0,0 +1,242 @@
...
+			notificationSupplier := &mock.NotificationSupplier{}
+			notificationSupplier.CommitNotificationsReturns(nil, errors.New("ERROR"))
+			notifier := commit.NewNotifier(notificationSupplier)
+
+			_, err := notifier.Notify(context.Background().Done(), "CHANNEL_NAME", "TX_ID")
+
+			require.Error(t, err)
```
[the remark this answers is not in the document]
bestbeforetoday: 👍

internal/pkg/gateway/commit/channelnotifier.go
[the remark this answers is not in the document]
manish-sethi: OK, great. nit: still don't need `notifier.listeners[key] = listeners`, as you are doing in-place replacement.
Regarding the tests, yes, you need a test in the same package for this. Again, leaving to you.

internal/pkg/gateway/commit/channelnotifier.go
```diff
@@ -0,0 +1,120 @@
...
+	}
+
+	notifier.lock.Lock()
+	notifier.listeners[transactionID] = append(notifier.listeners[transactionID], listener)
+	notifier.lock.Unlock()
+
+	return notifyChannel
+}
+
+type channelLevelListener struct {
```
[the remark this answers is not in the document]
bestbeforetoday: I don't like `channelLevelNotifier` or `channelLevelListener` (although at least they mirrored each other) and I like this name even less. I'll change it to `transactionListener`

internal/pkg/gateway/commit/notifier_test.go
```diff
@@ -0,0 +1,242 @@
...
+		})
+
+		t.Run("returns notifier on successful registration", func(t *testing.T) {
+			commitSend := make(chan *ledger.CommitNotification)
```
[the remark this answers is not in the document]
bestbeforetoday: 👍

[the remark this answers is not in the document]
bestbeforetoday: On reflection, I'm less comfortable with the test code relying on this side-effect of internal implementation to do clean up. It couples the unit tests to the private implementation and makes them less useful as a regression suite if the implementation is refactored. I've added an explicit `Close()` method instead, which may be legitimately used by real code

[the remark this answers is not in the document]
manish-sethi: yes, even better.

manish-sethi:
Thanks @bestbeforetoday. LGTM, except that I would have preferred a warning message instead of panic, but not a big deal if you want to go this way. Even then, a test would have been desired but it can be added later in a separate PR.

If you squash the commits, I can merge this PR now.

internal/pkg/gateway/commit/channelnotifier.go
```diff
@@ -0,0 +1,96 @@
...
+		commitChannel: commitChannel,
+		listeners:     make(map[*channelListener]bool),
+	}
+	go notifier.run()
+	return notifier
+}
+
+func (channelNotifier *channelNotifier) run() {
+	for {
+		blockCommit, ok := <-channelNotifier.commitChannel
```
sykesm: This looks like a range loop to me.

```go
for blockCommit := range channelNotifier.commitChannel {
...
}
```

internal/pkg/gateway/commit/channelnotifier.go
```diff
@@ -0,0 +1,96 @@
...
+func newChannelNotifier(commitChannel <-chan *ledger.CommitNotification) *channelNotifier {
+	notifier := &channelNotifier{
+		commitChannel: commitChannel,
+		listeners:     make(map[*channelListener]bool),
+	}
+	go notifier.run()
+	return notifier
+}
+
+func (channelNotifier *channelNotifier) run() {
```
sykesm: Why aren't we using idiomatic names? This stutters quite a bit and confuses the self-reference with the type name.

internal/pkg/gateway/commit/channelnotifier.go
```diff
@@ -0,0 +1,96 @@
...
+	}
+}
+
+func (channelNotifier *channelNotifier) notify(notification *Notification) {
+	channelNotifier.lock.Lock()
+	defer channelNotifier.lock.Unlock()
+
+	for listener := range channelNotifier.listeners {
+		complete := listener.Receive(notification)
+		if complete {
```
sykesm: no need for a temporary variable

bestbeforetoday:
> Thanks @bestbeforetoday. LGTM, except that I would have preferred a warning message instead of panic, but not a big deal if you want to go this way. Even then, a test would have been desired but it can be added later in a separate PR.

It’s not so easy to test as the panic happens asynchronously not directly from a test function invocation.

> If you squash the commits, I can merge this PR now.

It’s too late here for me to do that. Can you use the _Squash and merge_ option?

internal/pkg/gateway/commit/notifier.go
```diff
@@ -0,0 +1,72 @@
...
+type Notification struct {
+	BlockNumber    uint64
+	TransactionID  string
+	ValidationCode peer.TxValidationCode
+}
+
+func NewNotifier(ledgerNotifier LedgerNotifier) *Notifier {
+	if ledgerNotifier == nil {
+		panic("nil ledger notifier")
+	}
```
sykesm: this shouldn't be necessary as this isn't library code

internal/pkg/gateway/commit/notifier_test.go
```diff
@@ -193,5 +193,50 @@ func TestNotifier(t *testing.T) {
 
 			require.False(t, ok, "Expected notification channel to be closed but receive was successful")
 		})
+
+		t.Run("multiple listeners receive notifications", func(t *testing.T) {
+			commitSend := make(chan *ledger.CommitNotification, 2)
+			notifier := newTestNotifier(commitSend)
+
+			commitReceive1, _ := notifier.Notify(context.Background().Done(), "CHANNEL_NAME", "TX_ID")
+			commitReceive2, _ := notifier.Notify(context.Background().Done(), "CHANNEL_NAME", "TX_ID")
```
sykesm: Please add assertions on the errors.

```
func (notifier *Notifier) Notify(done <-chan struct{}, channelName string, transactionID string) (<-chan Notification, error) {
```

internal/pkg/gateway/commit/notifier_test.go
```diff
@@ -193,5 +193,50 @@ func TestNotifier(t *testing.T) {
...
+			require.EqualValues(t, expected, actual2)
+		})
+
+		t.Run("multiple listeners can stop listening independently", func(t *testing.T) {
+			commitSend := make(chan *ledger.CommitNotification, 2)
+			notifier := newTestNotifier(commitSend)
+
+			ctx, cancel := context.WithCancel(context.Background())
+			commitReceive1, _ := notifier.Notify(ctx.Done(), "CHANNEL_NAME", "TX_ID")
+			commitReceive2, _ := notifier.Notify(context.Background().Done(), "CHANNEL_NAME", "TX_ID")
```
sykesm: Please add assertions on the errors

internal/pkg/gateway/commit/channelnotifier.go
```diff
@@ -0,0 +1,96 @@
...
+import (
+	"sync"
+
+	"github.com/hyperledger/fabric/core/ledger"
+)
+
+type channelNotifier struct {
+	lock          sync.Mutex
+	commitChannel <-chan *ledger.CommitNotification
+	listeners     map[*channelListener]bool
```
sykesm: Using a pointer as a key is very risky. Why are we doing this?

internal/pkg/gateway/commit/channelnotifier.go
```diff
@@ -12,85 +12,113 @@ import (
...
 		commitChannel: commitChannel,
-		listeners:     make(map[*channelListener]bool),
+		listeners:     make(map[string][]*channelLevelListener),
 	}
 	go notifier.run()
 	return notifier
 }
 
-func (channelNotifier *channelNotifier) run() {
+func (notifier *channelLevelNotifier) run() {
```
sykesm, line 32: better, but still atypically long for a self reference

internal/pkg/gateway/commit/channelnotifier.go
```diff
@@ -13,37 +13,46 @@ import (
...
 
 func (notifier *channelLevelNotifier) run() {
 	for {
-		blockCommit, ok := <-notifier.commitChannel
-		if !ok {
-			break
+		select {
+		case blockCommit, ok := <-notifier.commitChannel:
+			if !ok {
+				panic("commit channel closed unexpectedly")
```
sykesm: I'm not sure I understand this. The done channel should be provided for cancelation; if it's closed, it shouldn't bring down the whole peer, should it?

internal/pkg/gateway/commit/notifier.go
```diff
@@ -58,6 +60,12 @@ func (notifier *Notifier) Notify(done <-chan struct{}, channelName string, trans
 	return notifyChannel, nil
 }
 
+// Close the notifier. This closes all notification channels obtained from this notifier. Behaviour is undefined after
+// closing and the notifier should not be used.
+func (notifier *Notifier) Close() {
+	close(notifier.cancel)
+}
```
sykesm: If Close is to be idempotent, you should consider using a sync.Once on the close.

internal/pkg/gateway/commit/notifier_test.go
```diff
@@ -239,4 +249,18 @@ func TestNotifier(t *testing.T) {
 			require.True(t, ok2, "Expected notification channel to deliver a result but was closed")
 		})
 	})
+
+	t.Run("Close", func(t *testing.T) {
+		t.Run("stops all listeners", func(t *testing.T) {
+			commitSend := make(chan *ledger.CommitNotification)
+			notifier := newTestNotifier(commitSend)
+
+			commitReceive, _ := notifier.Notify(context.Background().Done(), "CHANNEL_NAME", "TX_ID")
```
sykesm: error assertion, please.

internal/pkg/gateway/commit/channelnotifier.go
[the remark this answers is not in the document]
bestbeforetoday: You absolutely **do** need the `notifier.listeners[key] = listeners`. Behaviour is wrong without it

internal/pkg/gateway/commit/channelnotifier.go
bestbeforetoday: If `commitChannel` closes the ledger is dead and transactions can't be processed. I'm told that the design of the API I'm using to obtain `commitChannel` is that it is an error to obtain it more than once. Maybe this can be improved upon. Maybe we can do that in another PR

internal/pkg/gateway/commit/notifier.go
bestbeforetoday: I don't really need to be able to close the notifier right now, and this kind of complexity is why I hadn't implemented one until previous review comments drove this "requirement". However, the use of `sync.Once` is an excellent suggestion. 👍

internal/pkg/gateway/commit/notifier_test.go
bestbeforetoday: I personally am not a fan of single character variables names other than specific, very well established (and generally cross-language) conventions such as `i` as a loop index.

If you are mandating that code must use single-character receiver names to be allowed into Fabric, please can you come out and say that so I can understand clearly, and create linting rules to require this.

internal/pkg/gateway/commit/notifier.go
bestbeforetoday: I prefer failing fast to debugging obscure errors further down the line if the code is accidentally mis-used. Is it a significant issue to leave this in place, or would it be better to change `NewNotifier` to return `(*Notifier, error)` and force all callers to error check even if they are passing valid arguments that will never generate the error?

internal/pkg/gateway/commit/channelnotifier.go
bestbeforetoday: I personally am not a fan of single character variables names other than specific, very well established (and generally cross-language) conventions such as i as a loop index.

If you are mandating that code must use single-character receiver names to be allowed into Fabric, please can you come out and say that so I can understand clearly, and create linting rules to require this.

internal/pkg/gateway/commit/notifier_test.go
bestbeforetoday: The "returns notifier on successful registration" test already explicitly tests the error value. What is the value of duplicating the same test at least once in every other test case?

internal/pkg/gateway/commit/channelnotifier.go
bestbeforetoday: Outdated

internal/pkg/gateway/commit/channelnotifier.go
bestbeforetoday: Outdated

internal/pkg/gateway/commit/channelnotifier.go
bestbeforetoday: Outdated

internal/pkg/gateway/commit/channelnotifier.go
bestbeforetoday: Outdated

internal/pkg/gateway/commit/notifier.go
sykesm: This is not a library. In fact, there's no need for a constructor here at all if the field were embedded or exported. You're choosing to implement a constructor.

Now, do i care that you have a nil check and a panic? Not really. Do I care that it implies we don't trust the caller? Yes.

internal/pkg/gateway/commit/notifier_test.go
sykesm: It's not a _test_, it's an _assertion_. It's a piece of the test to ensure that the expected behavior is maintained throughout the execution of the test.

Ignoring errors is *never* the right thing to do - even in tests. This has been a historical problem in the code base that I have spent *significant time* addressing over the past years. It's an established convention and code that comes into Fabric should honor it.

We've now spent as much time debating it as it would take to add `require.NoError(t, err)` at the call sites.

internal/pkg/gateway/commit/channelnotifier.go
sykesm: That's fine. Where the argument starts to break apart is when there are different patterns within the same component. The gateway has:

- func (ef *endpointFactory)
- func (gs *Server)
- func (e *EndorserServerAdapter)

Seems there should at least be consistency within this gateway effort. Don't you agree?

internal/pkg/gateway/commit/channelnotifier.go
sykesm: Then I guess I question why have the channel at all. If we're going to terminate the process for on optional feature, why wouldn't the ledger, at the heart of the system, terminate the process?

So, I still don't understand why we'd panic but it can be revisited at a later time.

internal/pkg/gateway/commit/notifier.go
bestbeforetoday: You have previously instructed me to follow the guidance in Effective Go. This states:

> it's helpful to arrange when designing your data structures that the zero value of each type can be used without further initialization

and then goes on to say:

> Sometimes the zero value isn't good enough and an initializing constructor is necessary

For this type the zero value is not usable and will generate a panic if used. Usable instances are non-trivial for calling code to construct correctly and would also couple callers to implementation detail of this type that they don’t need to know or care about.

You might be inferring that I don’t trust calling code but I am absolutely not implying that. I am expecting to be writing the calling code and I want to make it as easy as possible for me to get that code right.

I am trying really really hard to follow all of your directives but it seems to be impossible to know when there is some other directive I haven’t been told about that supersedes your earlier directives.

internal/pkg/gateway/commit/channelnotifier.go
bestbeforetoday: I can change it back to just silently stop if that makes you happier

internal/pkg/gateway/commit/channelnotifier.go
bestbeforetoday: I can read the code ok either way so no, I’m not that worried about consistency here. If you are then my personal preference would be to avoid single-character variable names throughout. I am happy to make that change in this PR or a separate one

internal/pkg/gateway/commit/notifier_test.go
bestbeforetoday: Ok, I’ll remove the explicit test for this condition and add assertions to all the other test cases

internal/pkg/gateway/commit/channelnotifier.go
bestbeforetoday: Changed to not panic

internal/pkg/gateway/commit/notifier.go
bestbeforetoday: I have changed it to return nil rather than panic to align exactly with the examples in Effective Go

internal/pkg/gateway/commit/channelnotifier.go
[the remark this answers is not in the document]
bestbeforetoday: Changed to not panic anymore

bestbeforetoday:
@sykesm Please re-review

internal/pkg/gateway/commit/channelnotifier.go
sykesm: Honestly, I'd rather you adopt the conventions of the project instead of your own preferences. There's a reason why there's so much inconsistency in our configuration, our structure, and our tests. We have started to settle on some common preferences and I'd very much like to see green field code adhere to them.

internal/pkg/gateway/commit/notifier.go
```diff
@@ -0,0 +1,90 @@
...
+	BlockNumber    uint64
+	TransactionID  string
+	ValidationCode peer.TxValidationCode
+}
+
+// NewNotifier constructor.
+func NewNotifier(supplier NotificationSupplier) *Notifier {
+	if supplier == nil {
+		return nil
+	}
```
sykesm, line 42: This is just getting silly.

internal/pkg/gateway/commit/notifier.go
```diff
@@ -0,0 +1,90 @@
...
+func (notifier *Notifier) channelNotifier(channelName string) (*channelLevelNotifier, error) {
+	notifier.lock.Lock()
+	defer notifier.lock.Unlock()
+
+	result := notifier.channelNotifiers[channelName]
+	if result != nil {
+		return result, nil
+	}
+
+	commitChannel, err := notifier.supplier.CommitNotifications(context.Background().Done(), channelName)
```
sykesm: Why are we bothering with `context.Background()` to get a channel? I see `context` and I assume it's being used but it's not. You're just passing a channel, so, why context?

internal/pkg/gateway/commit/notifier_test.go
```diff
@@ -0,0 +1,285 @@
...
+			require.ErrorContains(t, err, "MY_ERROR")
+		})
+
+		t.Run("delivers notification for matching transaction", func(t *testing.T) {
+			commitSend := make(chan *ledger.CommitNotification, 1)
+			notifier := newTestNotifier(commitSend)
+			defer notifier.Close()
+
+			commitReceive, err := notifier.Notify(context.Background().Done(), "CHANNEL_NAME", "TX_ID")
+			require.NoError(t, err)
```
sykesm, line 54: Thank you for adding the error assertions.

internal/pkg/gateway/commit/notifier_test.go
```diff
@@ -0,0 +1,285 @@
...
+	})
+
+	t.Run("Notify", func(t *testing.T) {
+		t.Run("returns error from notification supplier", func(t *testing.T) {
+			notificationSupplier := &mock.NotificationSupplier{}
+			notificationSupplier.CommitNotificationsReturns(nil, errors.New("MY_ERROR"))
+			notifier := commit.NewNotifier(notificationSupplier)
+			defer notifier.Close()
+
+			_, err := notifier.Notify(context.Background().Done(), "CHANNEL_NAME", "TX_ID")
```
sykesm: Same question here about the use of context at this call and all of the others in the test

internal/pkg/gateway/commit/notifier_test.go
```diff
@@ -0,0 +1,285 @@
...
+			commitReceive, err := notifier.Notify(context.Background().Done(), "CHANNEL_NAME", "TX_ID")
+			require.NoError(t, err)
+
+			commitSend <- &ledger.CommitNotification{
+				BlockNumber: 1,
+				TxIDValidationCodes: map[string]peer.TxValidationCode{
+					"TX_ID": peer.TxValidationCode_MVCC_READ_CONFLICT,
+				},
+			}
+			actual := <-commitReceive
```
sykesm, line 62: Be aware that we have historically had issues with bare channel receives in tests. They have no timeout so when something goes awry, the tests hang for 20 minutes. You don't need to do anything right now but these patterns require consideration as you build out the code.

sykesm (dismissed):
Please make a follow up to avoid the use of the background context to create a cancelation channel or to properly wire a context through the API.

internal/pkg/gateway/commit/notifier.go
bestbeforetoday: Good spot, it really doesn't need to use `context` here 👍

internal/pkg/gateway/commit/notifier_test.go
bestbeforetoday: I wanted this code to be easy to use for any calling code that was using context, by having the same done channel type. This was my way of ensuring I didn't mess up. It doesn't need to use context so I have removed it 👍

internal/pkg/gateway/commit/notifier_test.go
bestbeforetoday: I like to see my tests fail before I make them pass so I've seen this a lot. I always run my unit tests with `go test`'s built-in `-timeout` parameter so they don't block for more than a couple of seconds before failing. Is it worth including the `-timeout` option in the main test run or is it necessary to write explicit timeout handling code in all the unit tests?

internal/pkg/gateway/commit/notifier.go
bestbeforetoday: For my own education, can you explain why this approach suggested by Effective Go and used in standard Go package implementations is wrong in this case? What would be the correct answer here?

internal/pkg/gateway/commit/channelnotifier.go
bestbeforetoday: You're right, these (any many other PR comments) are personal preference. I can absolutely relate; it would be nice for me if everyone had exactly the same personal preferences as me when it comes to code. But they don't, and for me that's fine provided the code is functionally correct (as demonstrated by passing tests) and fairly sane (along the lines of Kent Beck's rules of simple design). I try not to impose my personal preferences on other people's code. Nobody wants to contribute to a project -- particularly an open source project like this -- if they are allowed no freedom to make their own choices about how they implement their code

bestbeforetoday:
@sykesm Thank you for approving the PR. Since then other PRs have gone in so I've merged those changes to get this PR as up-to-date as possible. I also took the opportunity to address some of your latest round of comments. If you are happy enough to approve these changes, please can you also merge the PR so I can proceed with follow-on changes

sykesm (dismissed):

sykesm (approved):

PR #2479, opened 2021-03-13
[FAB-16951] Alternative mechanisms to find pkcs11 key

ldesrosi:
This modification adds a parameter called AltID to the PKCS11 BCCSP module.
When this parameter is set, the generateECKey and findKeyPairFromSKI functions
will use this alternative ID as the CKA_ID and CKA_LABEL to store and use the
associated private key.
This code change is required in situations where the HSM does not allow
modification of the CKA_ID after creation.
This code change includes corresponding unit-test and has been tested against
the SoftHSM, but still needs to be tested against AWS CloudHSM.

Signed-off-by: Luc Desrosiers <ldesrosi@uk.ibm.com>

#### Type of change

- New feature

#### Description

This pull request addresses an issue associated with using the PKCS11 module with the AWS CloudHSM.  As this HSM does not support the modify or copy operations, the SKI cannot be computed and used to update the key LABEL and ID.  
This modification allows to specify via configuration the ID to look up.  
  
#### Additional details. 
  
Port of the related feature added to the 1.4 release.  
Unit-tests have been added to cover the new capability.  
  
#### Related issues. 
  
[FAB-16951 - Add support for alternative mechanisms for locating pkcs11 keys](https://jira.hyperledger.org/browse/FAB-16951).

### DIFF
--- a/bccsp/pkcs11/pkcs11.go
+++ b/bccsp/pkcs11/pkcs11.go
@@ -42,6 +42,7 @@ type Provider struct {
 	curve      asn1.ObjectIdentifier
 	softVerify bool
 	immutable  bool
+	altID      string
 
 	createSessionRetries    int
 	createSessionRetryDelay time.Duration
@@ -102,6 +103,7 @@ func New(opts PKCS11Opts, keyStore bccsp.KeyStore) (*Provider, error) {
 		keyCache:                map[string]bccsp.Key{},
 		softVerify:              opts.SoftwareVerify,
 		immutable:               opts.Immutable,
+		altID:                   opts.AltID,
 	}
 
 	return csp.initialize(opts)
@@ -470,9 +472,22 @@ func (csp *Provider) generateECKey(curve asn1.ObjectIdentifier, ephemeral bool) 
 	}
 	defer func() { csp.handleSessionReturn(err, session) }()
 
-	id := nextIDCtr()
-	publabel := fmt.Sprintf("BCPUB%s", id.Text(16))
-	prvlabel := fmt.Sprintf("BCPRV%s", id.Text(16))
+	keylabel := ""
+	updateSKI := false
+	if csp.altID == "" {
+		// Generate using the SKI process and then make keypair immutable according to csp.immutable
+		keylabel = fmt.Sprintf("BCP%s", nextIDCtr().Text(16))
+		updateSKI = true
+	} else if csp.altID != "" && csp.immutable {
+		// Generate the key pair using AltID process.
+		// No need to worry about immutable since AltID is used with Write-Once HSMs
+		keylabel = csp.altID
+		updateSKI = false
+	} else if csp.altID != "" && !csp.immutable {
+		// Raise an error since AltID is used with Write-Once HSMs
+		// So cannot make support AltID with immutable = false.
+		return nil, nil, fmt.Errorf("Cannot generate a mutable key using AltID.")
+	}
 
 	marshaledOID, err := asn1.Marshal(curve)
 	if err != nil {
@@ -486,8 +501,8 @@ func (csp *Provider) generateECKey(curve asn1.ObjectIdentifier, ephemeral bool) 
 		pkcs11.NewAttribute(pkcs11.CKA_VERIFY, true),
 		pkcs11.NewAttribute(pkcs11.CKA_EC_PARAMS, marshaledOID),
 
-		pkcs11.NewAttribute(pkcs11.CKA_ID, publabel),
-		pkcs11.NewAttribute(pkcs11.CKA_LABEL, publabel),
+		pkcs11.NewAttribute(pkcs11.CKA_ID, []byte(keylabel)),
+		pkcs11.NewAttribute(pkcs11.CKA_LABEL, keylabel),
 	}
 
 	prvkeyT := []*pkcs11.Attribute{
@@ -497,8 +512,8 @@ func (csp *Provider) generateECKey(curve asn1.ObjectIdentifier, ephemeral bool) 
 		pkcs11.NewAttribute(pkcs11.CKA_PRIVATE, true),
 		pkcs11.NewAttribute(pkcs11.CKA_SIGN, true),
 
-		pkcs11.NewAttribute(pkcs11.CKA_ID, prvlabel),
-		pkcs11.NewAttribute(pkcs11.CKA_LABEL, prvlabel),
+		pkcs11.NewAttribute(pkcs11.CKA_ID, []byte(keylabel)),
+		pkcs11.NewAttribute(pkcs11.CKA_LABEL, keylabel),
 
 		pkcs11.NewAttribute(pkcs11.CKA_EXTRACTABLE, false),
 		pkcs11.NewAttribute(pkcs11.CKA_SENSITIVE, true),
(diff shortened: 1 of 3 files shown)


bccsp/pkcs11/pkcs11_test.go
```diff
@@ -283,6 +296,43 @@ func TestECDSAGetKeyBySKI(t *testing.T) {
...
+	_, _, err = csp.generateECKey(oidNamedCurveP256, true)
+	if err != nil {
+		t.Fatalf("Failed generating Key [%s]", err)
+	}
+
+	k, err = csp.findKeyPairFromSKI(sess, fakeSKI, privateKeyType)
+	if err != nil {
+		t.Fatalf("Found no key after generating an EC Key based on the AltId.")
+	}
+	if k == 0 {
```
sykesm, line 331: This is saying we got _some_ object handle back. It's not asserting on the attributes.

bccsp/pkcs11/pkcs11_test.go
```diff
@@ -283,6 +296,43 @@ func TestECDSAGetKeyBySKI(t *testing.T) {
 	require.Equalf(t, k.SKI(), k2.SKI(), "expected %x got %x", k.SKI(), k2.SKI())
 }
 
+func TestAlternateLabelGeneration(t *testing.T) {
```
sykesm, line 299: I'm confused by this test. It's called `TestAlternateLabelGeneration` but I don't see any calls to get attributes to ensure the keys have the appropriate labels. I see the test using the provider to create a key. What's being tested here that wasn't tested in the table test above?

bccsp/pkcs11/pkcs11_test.go
```diff
@@ -250,10 +254,19 @@ func TestKeyGenECDSAOpts(t *testing.T) {
 				require.Len(t, attr, 1)
 
 				if tt.immutable {
-					require.Equal(t, []byte{0}, attr[0].Value)
+					if tt.altID == "" {
```
sykesm, line 257: Please reduce the nesting as there's only a single assertion.
```
if tt.immutable && tt.altID == "" { ... }
```

bccsp/pkcs11/pkcs11.go
```diff
@@ -687,8 +704,14 @@ func (csp *Provider) clearCaches() {
 }
 
 func (csp *Provider) findKeyPairFromSKI(session pkcs11.SessionHandle, ski []byte, keyType keyType) (pkcs11.ObjectHandle, error) {
+	keyID := ski
```
sykesm, line 707: And what about the layers above this? This is changing the ski stored in the cache but doesn't do anything about how the key is managed in `GetKey` or `getECKey`.

sykesm (changes requested):
I've commented on bits of the code but I have a higher level question. The current implementation assumes that keys are identified with a subject key identifier computed from a certificate/signing identity. The general implication is that this package supports multiple keys.

Unless I'm misreading, with these changes, the package is moving to a model where it only supports a single key when `AltID` is set. This does not seem like a reasonable model.

If this is about the signing key for a peer, why are we addressing the alternate ID mapping at this layer instead of at the identity layer in the MSP?

ldesrosi:
@sykesm Thank you for the feedback.  If you're OK, I'll address the higher-level question first and then I can focus on the implementation feedback.

The first thing to mention is that this is a direct port of what was implemented in 1.4.  Also, as you probably saw, the issue is that HSMs like the AWS CloudHSM don't implement operations like C_CopyObject, C_SetAttribute.  As such SKI, which is computed after key generation cannot be used as CKA_LABEL or CKA_ID.

BCCSP is currently agnostic of the identity. We did consider adding a new identity attribute but it was causing too many changes across many interfaces.  

Even so, we had a look at potential attributes we could use:  
* DN:  we did not want to use the DN of the cert because it was going to cause conflicts when renewing certs.  
* Custom fields: we looked at custom fields but then when we came to test, we realized that some CA will drop custom parameters off the Certificate content.  

Another idea was to create a reference table at the MSP level and use it to map the SKI to the actual label to use.  However, it was creating another set of concerns.

I am attaching an analysis I did when the issue was initially identified:   
[AWS CloudHSM impact analysis v2.2.docx](https://github.com/hyperledger/fabric/files/6144168/AWS.CloudHSM.impact.analysis.v2.2.docx). 
It contains the various options we explored. 
Let me know if you would like to discuss.

sykesm:
> The first thing to mention is that this is a direct port of what was implemented in 1.4.

Yes, I'm aware, but I want to highlight one of the comments:
> `I think we should look at possibly making a different change in 2.0 / master, but for 1.4.x this seems ok.`

We knew this needed a better solution when the item was opened in Gerrit in 2019 and I expect that's part of the reason there was so much discussion and delay on merging on the change.

> Also, as you probably saw, the issue is that HSMs like the AWS CloudHSM don't implement operations like C_CopyObject, C_SetAttribute. As such SKI, which is computed after key generation cannot be used as CKA_LABEL or CKA_ID.

Yes, and we should be able to support that. My concern is that this change implies that we will only ever need a single key reference. It doesn't make sense to me to handle a single use case by neutering the API. IMO, it's either the wrong solution or the wrong API.
 
> BCCSP is currently agnostic of the identity. We did consider adding a new identity attribute but it was causing too many changes across many interfaces.

BCCSP doesn't need to know about identity; it needs to know how to locate a key. The SKI was a convenient way of doing that since it's a small, consistent, unique, and easy to generate from a public key.

I believe the goal should be an extension that maps the key identifier to a key label. This mapping can be a function provided to the `Provider` implementation when it's instantiated. This keeps the mapping logic orthogonal to the use. If everything is mapped to the configured `AltId`, the function can just return a static string. It's a small change but it moves the logic out of the mainline path.

For the key generation flows, the discussion should really be moved over to the Fabric CA. In that environment we'd need changes to the API to enable a user to provide the label or to return the generated label.

> Let me know if you would like to discuss.

That's what we're doing. :)

ldesrosi:
If I understand properly, the idea is to separate the key mapping logic from the PKCS11 Provider.   We would define an interface with a state and two functions:  
* PreKeyPairGenerationFlag (Would need a better descriptive name)
   Static value provided by the implementation. Would be false for SKI and true for the AWS CloudHSM. Indicates if the key identifier is computed/available before the KeyPair creation. This would be used by `generateECKey` to determine if the Key Identifier can be set as a parameter to the GenerateKeyPair or if it needs to be modified following the KeyPair generation.
* GenerateKeyIdentifier(key pkcs11.ObjectHandle)  
   Default implementation would compute the SKI and return it.  The AWS CloudHSM would return the AltID
* GetKeyIdentifier(ski []byte)  
  Default implementation returns the SKI and the AWS CloudHSM implementation returns a static value (ie the AltID).

The PKCS11 BCCSP configuration would have a new parameter which is the "Mapper" with two implementations:
* SKI Mapper (Default) 
* StaticKeyMapper, which would have a single config parameter (called Id or AltID).

The outcome remains the same though, meaning that for the StaticKeyMapper, there is only a single key reference.

Is that what you had in mind?

sykesm:
I had an offline chat with @ldesrosi. To summarize, the key generation path is only used by the FabricCA client and the client is using bccsp from the 1.4 version of Fabric. This is unlikely to change in the near term as the CA contains support for RSA keys and certificates and the current versions of BCCSP have dropped that.

So, since we're not using the key generation functions, we only need to handle finding the key. That's much simpler. I'll put together a small PR that deals with that for Fabric 2.x and @ldesrosi will take it from there. I expect to do that some time over the next two days.

sykesm:
Please see #2486

sykesm:
The PR I referenced above has been merged and back-ported to release-2.2. If we need to make additional changes, we should work from that as a base.

Thank you for putting this together.

PR #2486, opened 2021-03-17
Introduce ability to map subject key identifers to key objects with read-only CKA_ID attributes

sykesm:
Add support for mapping subject key identifiers to cryptoki object CKA_ID attributes. This allows Fabric to use keys from token providers that do not support CKA_ID attribute modification.


jyellick (approved):
LGTM

PR #2488, opened 2021-03-18
FABGW-8: Hook commit notifier into Gateway service

bestbeforetoday:
Still not executed at runtime. Just establishing the integration point and peer adapter to allow use later.

Note that this PR is just reinstating some changes that got lost from #2477, possibly during rebasing around merge conflicts.

### DIFF
--- a/internal/pkg/gateway/commitadapter.go
+++ b/internal/pkg/gateway/commitadapter.go
@@ -1,0 +1,26 @@
+/*
+Copyright IBM Corp. All Rights Reserved.
+
+SPDX-License-Identifier: Apache-2.0
+*/
+
+package gateway
+
+import (
+	"github.com/hyperledger/fabric/core/ledger"
+	"github.com/hyperledger/fabric/core/peer"
+	"github.com/pkg/errors"
+)
+
+type PeerNotifierAdapter struct {
+	Peer *peer.Peer
+}
+
+func (adapter *PeerNotifierAdapter) CommitNotifications(done <-chan struct{}, channelName string) (<-chan *ledger.CommitNotification, error) {
+	channel := adapter.Peer.Channel(channelName)
+	if channel == nil {
+		return nil, errors.Errorf("channel does not exist: %s", channelName)
+	}
+
+	return channel.Ledger().CommitNotificationsChannel(done)
+}
--- a/internal/peer/node/start.go
+++ b/internal/peer/node/start.go
@@ -825,6 +825,9 @@ func serve(args []string) error {
 				gateway.CreateServer(
 					&gateway.EndorserServerAdapter{Server: serverEndorser},
 					discoveryService,
+					&gateway.PeerNotifierAdapter{
+						Peer: peerInstance,
+					},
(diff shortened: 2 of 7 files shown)


internal/pkg/gateway/commitadapter.go
```diff
@@ -0,0 +1,26 @@
...
+)
+
+type PeerNotifierAdapter struct {
+	Peer *peer.Peer
+}
+
+func (adapter *PeerNotifierAdapter) CommitNotifications(done <-chan struct{}, channelName string) (<-chan *ledger.CommitNotification, error) {
+	channel := adapter.Peer.Channel(channelName)
+	if channel == nil {
+		return nil, errors.Errorf("Channel does not exist: %s", channelName)
```
sykesm: From the material we reference in our coding [guidelines](https://github.com/hyperledger/fabric/blob/master/docs/source/style-guides/go-style.rst#coding-in-go):

https://github.com/golang/go/wiki/CodeReviewComments#error-strings

> Error strings should not be capitalized (unless beginning with proper nouns or acronyms) or end with punctuation, since they are usually printed following other context. That is, use fmt.Errorf("something bad") not fmt.Errorf("Something bad"), so that log.Printf("Reading %s: %v", filename, err) formats without a spurious capital letter mid-message. This does not apply to logging, which is implicitly line-oriented and not combined inside other messages.

As with most things Fabric there are plenty of counter-examples in legacy code but this is new code.

bestbeforetoday: I know that rule. My bad, this slipped past my radar

bestbeforetoday: Any idea why the staticcheck rules are not catching this?
https://staticcheck.io/docs/checks#ST1005
Is it because we prefer not to use standard Go error libraries, or are some rules disabled to avoid falling over on existing code (in which case, can we ramp up the rules applied to the Gateway code), or something else...?

sykesm: We disable the checks because the older code is littered with bad error strings.

https://github.com/hyperledger/fabric/blob/13092bbfb631a612091d4ccd3e849b320fa9c6a6/staticcheck.conf#L16-L19

sykesm (approved):
I'm merging this but I want to point out that the reason the notification supplier got dropped is because it wasn't called and there are no tests. That still seems to be the case.

bestbeforetoday:
@sykesm it is really really **not** ok to offer to help rebase and merge a PR that you have approved, and to then unilaterally delete my code and force-push those deletes back to my branch without even doing me the courtesy of telling me

mastersingh24:
I see nothing wrong with the way this was handled.  We could either get stuck in an endless here (there were plenty of conflict/merge issues earlier) or get the code merged.

yacovm:
The commit has @bestbeforetoday 's signed-off-by on it:

> Signed-off-by: Mark S. Lewis <mark_lewis@uk.ibm.com>

It should have been coordinated with him whether to make such a change or not. 

When you contribute to an open source project, the maintainers have a say in what **does not get** merged, but the code contributor has a say in **what** gets merged.

PR #2491, opened 2021-03-18
Introduce ability to map subject key identifers to key objects with read-only CKA_ID attributes (release-2.2)

sykesm:
Add support for mapping subject key identifiers to cryptoki object CKA_ID attributes. This allows Fabric to use keys from token providers that do not support CKA_ID attribute modification.

Backport #2486

### DIFF
--- a/bccsp/factory/pkcs11factory.go
+++ b/bccsp/factory/pkcs11factory.go
@@ -9,6 +9,8 @@ SPDX-License-Identifier: Apache-2.0
 package factory
 
 import (
+	"encoding/hex"
+
 	"github.com/hyperledger/fabric/bccsp"
 	"github.com/hyperledger/fabric/bccsp/pkcs11"
 	"github.com/hyperledger/fabric/bccsp/sw"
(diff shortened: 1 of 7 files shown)


bccsp/factory/pkcs11factory.go
```diff
@@ -35,8 +37,27 @@ func (f *PKCS11Factory) Get(config *FactoryOpts) (bccsp.BCCSP, error) {
...
+	for _, k := range p11Opts.KeyIDs {
+		keyMap[k.SKI] = k.ID
+	}
+
+	return func(ski []byte) []byte {
+		keyID := hex.EncodeToString(ski)
+		if id, ok := keyMap[keyID]; ok {
+			return []byte(id)
+		}
+		if p11Opts.AltID != "" {
```
mastersingh24, line 58: This is for backwards compatibility with 1.4.x?

mastersingh24 (approved):
This is probably overkill for how bccsp is actually currently used within fabric, but LGTM

bccsp/factory/pkcs11factory.go
sykesm: Yes.

PR #2492, opened 2021-03-18
Bugfix: loadLastConfig wrongly returns nil instead of error

dvitas:
Trivial bugfix - should use Errorf() here instead of WithMessagef()

#### Type of change

- Bug fix

#### Description

Should use Errorf() here instead of WithMessagef() because "err" here is always nil, hence WithMessagef() always returned nil instead of the expected error.

### DIFF
--- a/orderer/common/follower/follower_chain.go
+++ b/orderer/common/follower/follower_chain.go
@@ -561,7 +561,7 @@ func (c *Chain) loadLastConfig() error {
 	}
 	lastConfig := c.ledgerResources.Block(index)
 	if lastConfig == nil {
-		return errors.WithMessagef(err, "could not retrieve config block from index %d", index)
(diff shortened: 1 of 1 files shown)


sykesm (approved):
Good catch. We generally like to have a test added to prevent regressions and this doesn't seem to break any existing assertions.

Did you spot this while looking at source or did you hit this at runtime?

dvitas:
@sykesm I am drafting the RFC for orderer pruning/snapshot feature, so I am playing with the code to test different approaches to that, and I hit the bug during one of my experiments.

But having a test is always a plus, so I will come up with a unit test for this particular regression in a new pull request.

sykesm:
> But having a test is always a plus, so I will come up with a unit test for this particular regression in a new pull request.

Very cool. Thank you.

PR #2493, opened 2021-03-18
Regression test for prev commit "Bugfix: loadLastConfig wrongly returns nil instead of error"

dvitas:
Regression unit test for the prev bugfix

#### Type of change

- Test update

#### Related issues

Pull request #2492


sykesm (approved):
LGTM. Thanks for adding the test!

PR #2494, opened 2021-03-19
FABGW-8: Implement Gateway CommitStatus service

bestbeforetoday:
- Add a commit finder, which queries for previously committed transaction status as well as using the previously implemented commit notifier to wait for subsequent transaction commits
- Use the commit finder to supply response values for CommitStatus service
- Unit tests for PeerAdapter
- Integration test for CommitStatus


jyellick (approved):
LGTM

PR #2497, opened 2021-03-22
FABGW-9 Select endorsers with max block height

andrew-coleman:
When selecting the endorsers from the endorsement plan,
sort each group in the layout in order for decreasing ledger height.  If equal height, give priority to the host peer.

### DIFF
--- a/internal/pkg/gateway/registry.go
+++ b/internal/pkg/gateway/registry.go
@@ -8,7 +8,7 @@ package gateway
 
 import (
 	"fmt"
-	"math/rand"
+	"sort"
 	"sync"
 
 	"github.com/golang/protobuf/proto"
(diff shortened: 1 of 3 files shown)


internal/pkg/gateway/registry.go
```diff
@@ -67,37 +73,61 @@ func (reg *registry) endorsers(channel string, chaincode string) ([]*endorser, e
...
-			peerGroup := endorsers[0:quantity]
+			// Select n remoteEndorsers from each group sorted by block height
+
+			// block heights
+			var groupPeers []*endorserState
+			for _, peer := range e[group].Peers {
+				msg := &gossip.GossipMessage{}
+				err := proto.Unmarshal(peer.StateInfo.Payload, msg)
+				if err != nil {
+					return nil, err
```
ale-linux, line 86: IIUC this actually changes the failure model of the function - we are now failing to select an endorser if we can't gather block height related information. Is that expected?

yacovm: If you can't discern channel related information about a peer, it will not show up in discovery queries anyway. 
So, all peers you get from discovery are those that have this information in the first place, assuming gossip doesn't misbehave.

yacovm: Right

jyellick (approved):
LGTM

PR #2500, opened 2021-03-23
FABGW-8: Fix race condition in Finder unit test

bestbeforetoday:
Keep sending messages on the send channel to ensure there is still a message available to read once the listener is established.

### DIFF
--- a/internal/pkg/gateway/commit/finder_test.go
+++ b/internal/pkg/gateway/commit/finder_test.go
@@ -9,6 +9,7 @@ package commit
 import (
 	"context"
 	"testing"
+	"time"
 
 	"github.com/hyperledger/fabric-protos-go/peer"
 	"github.com/hyperledger/fabric/core/ledger"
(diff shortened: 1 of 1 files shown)


internal/pkg/gateway/commit/finder_test.go
```diff
@@ -23,6 +23,21 @@ type queryProvider interface { // Mimic QueryProvider to avoid circular import w
 }
 
 func TestFinder(t *testing.T) {
+	sendUntilDone := func(commitSend chan<- *ledger.CommitNotification, msg *ledger.CommitNotification) chan struct{} {
+		done := make(chan struct{})
+
+		go func() {
+			for {
+				select {
+				case commitSend <- msg:
```
jyellick, line 33: This will result in a tight loop, no?  Probably worth putting at least a small sleep in here? 10ms?

bestbeforetoday: 👍

jyellick (approved):
Not my favorite approach, but since I don't have another and it's a flake that could be hitting others, merging.

PR #2505, opened 2021-03-24
peer: the 'noplugin' build tag disables plugins

sykesm:
Go plugins are not supported on all platforms and architectures and there are complications with different libc implementations. This tag can be used to disable plugins on platforms supported by Go to avoid these complications.


mastersingh24:
Guess it was inevitable for the no plugin build tag to be needed again

mastersingh24 (approved):

PR #2508, opened 2021-03-25
Fix timeout of orderer connection from gateway

andrew-coleman:
The gateway was invoking the Broadcast() method on the orderer and caching the returned stream.  This was causing timeout issues.
The code has been rewritten to open a new stream for each transaction submission.


internal/pkg/gateway/api.go
```diff
@@ -175,16 +175,25 @@ func (gs *Server) Submit(ctx context.Context, request *gp.SubmitRequest) (*gp.Su
 	}
 
 	orderer := orderers[0] // send to first orderer for now
+
+	broadcast, err := orderer.client.Broadcast(ctx)
+	if err != nil {
+		return nil, rpcError(
+			codes.Aborted,
+			"failed to send transaction to orderer",
```
sykesm, line 183: This error is the same as the next one and that could complicate debug. It’s also slightly inaccurate as this is a failure to create an `AtomicBroadcast_BroadcastClient` or “BroadcastClient”.

sykesm (approved):

PR #2528, opened 2021-04-01
FAB-18417 Improve snapshot dir error message

denyeart:
Users upgrading to v2.3 don't always know to configure ledger.snapshots.rootDir.
In cases where peer doesn't have write access to this directory,
the error message now mentions the property to configure.


ale-linux (approved):

denyeart:

PR #2536, opened 2021-04-10
Update deploy_chaincode.md

mode51software:
#### Type of change

- Documentation update

#### Description

The name of the test network in docker appears to be fabric_test rather than net_test.

This name is used by the monitordocker.sh logspout command.

```
docker network ls

NETWORK ID          NAME                DRIVER              SCOPE
4ba75e52d44c        bridge              bridge              local
d64c5715a632        fabric_test         bridge              local
ac3dcef617db        host                host                local
50a642dc26be        none                null                local
```


mode51software:
v2.3 already fixes this

PR #2537, opened 2021-04-12
Create codeql-analysis.yml

jfcg:
#### Type of change
- Test update

#### Description
Add Github CodeQL security checks for Go.

### DIFF
--- a/.github/workflows/codeql-analysis.yml
+++ b/.github/workflows/codeql-analysis.yml
@@ -1,0 +1,28 @@
+#
+# SPDX-License-Identifier: Apache-2.0
+#
+name: "CodeQL"
+
+on:
+  push:
+    branches: [ main, release-2.2, release-1.4 ]
+  pull_request:
+    # The branches below must be a subset of the branches above
+    branches: [ main, release-2.2, release-1.4 ]
+
+jobs:
+  analyze:
+    name: Analyze
+    runs-on: ubuntu-latest
+
+    steps:
+    - name: Checkout repo
+      uses: actions/checkout@v2
+
+    - name: Initialize CodeQL
+      uses: github/codeql-action/init@v1
+      with:
+        languages: 'go'
+
+    - name: CodeQL Analysis
+      uses: github/codeql-action/analyze@v1


.github/workflows/codeql-analysis.yml
```diff
@@ -1,24 +1,11 @@
-# For most projects, this workflow file will not need changing; you simply need
```
lindluni: Please add this to the top of the file to pass the verification step:

```
#
#
```

.github/workflows/codeql-analysis.yml
```diff
@@ -1,24 +1,11 @@
...
-# We have attempted to detect the languages in your repository. Please check
-# the `language` matrix defined below to confirm you have the correct set of
-# supported CodeQL languages.
-#
 name: "CodeQL"
 
 on:
   push:
-    branches: [ main ]
+    branches: [ main, release-2.3, release-2.2, release-2.1, release-2.0, release-1.4 ]
```
lindluni: I think I'd like to only see LTS branches here:
```
branches: [ main, release-2.2, release-1.4 ]
```
Remediating and keeping track of all the branches may be a lot of effort. And since all future LTS branches are created from `main` future branches should only have new issues for changes not introduced from cherry-picks.

lindluni (changes requested):

I would like to see this turned on, since it's available it would be nice to have. That being said:

Please add the license header, and remove all the extraneous comments from the default file.

I think we can remove the autobuild step. The point of the autobuild step is to build the compiled binary which does a force download of the dependencies. Since we vendor our deps they already exist in the repo. CodeQL also doesn't support the build features for Go in general.

lindluni:
Please also squash your commits into a single commit. After squashing you will also need to sign your commit using `git commit --amend --no-edit --sign-off` and then force push the changes

.github/workflows/codeql-analysis.yml
```diff
@@ -0,0 +1,33 @@
...
+
+jobs:
+  analyze:
+    name: Analyze
+    runs-on: ubuntu-latest
+
+    strategy:
+      fail-fast: false
+      matrix:
+        language: [ 'go' ]
```
lindluni: Nit: since this repo is only Go, no real need to matrix the build, we could have just put it in the Initialize step, and since it's only a single language with no autobuild, we can fail-fast. But no reason to block the the merge.

lindluni (dismissed):
I don't see a downside to enabling it on PR, let see what others think

.github/workflows/codeql-analysis.yml
jfcg: should `with.languages` be an array or a single string ?

lindluni: Looks like you got it sorted, I should have been clearer, but you can remove the entire `strategy` stanza as `fail-fast: true` is the default behavior. No need to change it we can address it another PR

lindluni (dismissed):

lindluni:
Let's see what others think, but LGTM

jfcg:
Btw in my fork under `/security/code-scanning` I see:
Fabric

denyeart:
So if this is merged it will automatically do the checks for every PR?

I assume we would want to clean up the existing failures before merging so that we start the checks from a clean slate?  Has anybody looked into the failures to see if they are easily fixable?

sykesm:
> Has anybody looked into the failures to see if they are easily fixable?

Yes. The last time this came up. Many aren't really issues. For example, one of the "zip slip" issues that is flagged is actually mitigated. Other issues around integer overflow are also mitigated.

I think we can get the same information on demand so I'm not sure why we want it on every PR.

https://lgtm.com/projects/g/hyperledger/fabric/?mode=list

jfcg:
> Yes. The last time this came up. Many aren't really issues. For example, one of the "zip slip" issues that is flagged is actually mitigated. Other issues around integer overflow are also mitigated.

the second one seems to miss the validPath func or `..` check, so it found one true & one false positives there.
there are 9 "uncontrolled data" reports, if one or two turns out to be true positives then it is definitely worth to fix it. But I agree sometimes false positives are too many. The tool will probably improve in time.

> I think we can get the same information on demand so I'm not sure why we want it on every PR.
> 
> https://lgtm.com/projects/g/hyperledger/fabric/?mode=list

maybe only on push ? i think these kind of checks should be integrated with development cycle.

lindluni:
> > Has anybody looked into the failures to see if they are easily fixable?
> 
> Yes. The last time this came up. Many aren't really issues. For example, one of the "zip slip" issues that is flagged is actually mitigated. Other issues around integer overflow are also mitigated.
> 
> I think we can get the same information on demand so I'm not sure why we want it on every PR.
> 
> https://lgtm.com/projects/g/hyperledger/fabric/?mode=list

We can mark the ones that are false-positives as such. Then only newly introduced issues will be flagged by the PR's and false-positives can be dismissed directly in the PR UI to bring the check back to green and allow a merge, and prevent it from being flagged on merge:

image
image

Honestly, the argument "we can get this information on demand" doesn't really hold historically. We had the same conversation for go-sec because we ran it and published the results externally. When we debated it we decided to drop it externally and only implement it if it was going to be performed on demand at PR time. We know anything living outside of this repository is almost entirely ignored.

sykesm:
> Honestly, the argument "we can get this information on demand" doesn't really hold historically. We had the same conversation for go-sec because we ran it and published the results externally.

Yet we never added it to the PR process... At least that's a tool that can easily be run locally.

I'm not stopping anything but here are my opinions.

1. A tool that has lots of false positives out of the gate will generate a bunch of work. This is my biggest concern.
2. A tool that cannot be easily executed locally in development will impact workflows and desk checks.
3. If it really takes another 15 minutes to run this (even concurrently), that's a heavy price.
4. Sorry, but not everything should be implemented as a GitHub integration. This inevitably moves the source of truth out of the repo and to some online service. This seems to have happened with "read the docs" and it looks like it would happen when someone "dismisses" a warning with ql/lgtm.

Regardless, any real issues that are flagged should be addressed.

sykesm:
To be a bit more clear, I don't object to adding these checks. I do believe the issues that are flagged should be evaluated and, for those that are real issues, it's best to get them addressed before the checks are integrated into the CI process.

The cost vs value can be determined later.

denyeart:
I think we need a plan for the flagged issues before we merge this (either fix them or mark them as false positives). So for now I will close, but feel free to re-open with a plan provided.

PR #2539, opened 2021-04-14
Compare Snapshots Utility

jcastrence:
#### Type of change

- New feature

#### Description

- A ledger troubleshooting utility that can identify differences between two snapshots from different peers. Compares each key / value pair between the snapshots then outputs the key / value pairs that are dissimilar.

#### Related issues
 - [FAB-18425](https://jira.hyperledger.org/browse/FAB-18425)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -28,11 +28,13 @@
 #   - idemixgen - builds a native idemixgen binary
 #   - integration-test-prereqs - setup prerequisites for integration tests
 #   - integration-test - runs the integration tests
+#   - ledger - builds a native fabric ledger troubleshooting binary
 #   - license - checks go source files for Apache license header
 #   - linter - runs all code checks
 #   - native - ensures all native binaries are available
 #   - orderer - builds a native fabric orderer binary
 #   - orderer-docker[-clean] - ensures the orderer container is available[/cleaned]
+#   - osnadmin - builds a native fabric osnadmin binary
 #   - peer - builds a native fabric peer binary
 #   - peer-docker[-clean] - ensures the peer container is available[/cleaned]
 #   - profile - runs unit tests for all packages in coverprofile mode (slow)
@@ -83,13 +85,14 @@ GO_TAGS ?=
 RELEASE_EXES = orderer $(TOOLS_EXES)
 RELEASE_IMAGES = baseos ccenv orderer peer tools
 RELEASE_PLATFORMS = darwin-amd64 linux-amd64 windows-amd64
-TOOLS_EXES = configtxgen configtxlator cryptogen discover idemixgen osnadmin peer
+TOOLS_EXES = configtxgen configtxlator cryptogen discover idemixgen ledger osnadmin peer
 
 pkgmap.configtxgen    := $(PKGNAME)/cmd/configtxgen
 pkgmap.configtxlator  := $(PKGNAME)/cmd/configtxlator
 pkgmap.cryptogen      := $(PKGNAME)/cmd/cryptogen
 pkgmap.discover       := $(PKGNAME)/cmd/discover
 pkgmap.idemixgen      := $(PKGNAME)/cmd/idemixgen
+pkgmap.ledger		  := $(PKGNAME)/cmd/ledger
 pkgmap.orderer        := $(PKGNAME)/cmd/orderer
 pkgmap.osnadmin       := $(PKGNAME)/cmd/osnadmin
 pkgmap.peer           := $(PKGNAME)/cmd/peer
--- a/cmd/ledger/main.go
+++ b/cmd/ledger/main.go
@@ -1,0 +1,58 @@
+/*
+Copyright IBM Corp. All Rights Reserved.
+
+SPDX-License-Identifier: Apache-2.0
+*/
+
+package main
+
+import (
+	"fmt"
+	"os"
+
+	"github.com/hyperledger/fabric/internal/ledger"
+	"gopkg.in/alecthomas/kingpin.v2"
+)
+
+const (
+	resultFilename = "./result.json"
+)
+
+var (
+	app = kingpin.New("ledger", "Ledger Utility Tool")
+
+	compare       = app.Command("compare", "Compare two ledgers via their snapshots.")
+	snapshotPath1 = compare.Arg("snapshotPath1", "File path to first ledger snapshot.").Required().String()
+	snapshotPath2 = compare.Arg("snapshotPath2", "File path to second ledger snapshot.").Required().String()
+
+	troubleshoot = app.Command("troubleshoot", "Identify potentially divergent transactions.")
+
+	args = os.Args[1:]
+)
+
+func main() {
+	kingpin.Version("0.0.1")
+
+	command, err := app.Parse(args)
+	if err != nil {
+		kingpin.Fatalf("parsing arguments: %s. Try --help", err)
+		return
+	}
+
+	switch command {
+
+	case compare.FullCommand():
+
+		count, err := ledger.Compare(*snapshotPath1, *snapshotPath2, resultFilename)
+		if err != nil {
+			fmt.Println(err)
+			return
+		}
+		fmt.Printf("\nSuccessfully compared snapshots. Result saved to %s. Total differences found: %v\n", resultFilename, count)
+
+	case troubleshoot.FullCommand():
+
+		fmt.Println("Command TBD")
+
+	}
+}
--- a/internal/ledger/compare.go
+++ b/internal/ledger/compare.go
@@ -1,0 +1,414 @@
+/*
+Copyright IBM Corp. All Rights Reserved.
+
+SPDX-License-Identifier: Apache-2.0
+*/
+
+package ledger
+
+import (
+	"bufio"
+	"bytes"
+	"encoding/json"
+	"io/ioutil"
+	"os"
+	"path/filepath"
+	"strings"
+
+	"github.com/golang/protobuf/proto"
+	"github.com/hyperledger/fabric/common/ledger/util"
+	"github.com/hyperledger/fabric/core/ledger/kvledger"
+	"github.com/hyperledger/fabric/core/ledger/kvledger/txmgmt/privacyenabledstate"
+	"github.com/hyperledger/fabric/internal/fileutil"
+	"github.com/pkg/errors"
+)
+
+// Compare - Compares two ledger snapshots and outputs the differences in snapshot records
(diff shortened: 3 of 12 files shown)


internal/ledger/compare.go
```diff
@@ -0,0 +1,413 @@
...
+	"github.com/pkg/errors"
+)
+
+// nsKey type is used to compare between snapshot records using both the namespace and key
+type nsKey struct {
+	namespace string
+	key       []byte
+}
+
+// DiffRecord type for representing diverging records in json
```
manish-sethi: Just a nit-picking: Here and other place - usually, golang comments sound like "DiffRecord represents a diverging record in json"

internal/ledger/compare.go
```diff
@@ -0,0 +1,413 @@
...
+type DiffRecord struct {
```
manish-sethi: Like the other types, this can be unexported.

internal/ledger/compare.go
```diff
@@ -0,0 +1,413 @@
...
+	TxNum    uint64 `json:"txNum,omitempty"`
+}
+
+// jsonArrayFileWriter type is used to write a list of DiffRecords to a json file
+type jsonArrayFileWriter struct {
+	file               *os.File
+	buffer             *bufio.Writer
+	encoder            *json.Encoder
+	firstRecordWritten bool
+}
```
manish-sethi: Better to move this down to where the functions on this are defined.

internal/ledger/compare.go
```diff
@@ -0,0 +1,413 @@
...
+
+// Compare - Compares two ledger snapshots and outputs the differences in snapshot records
+// This function will overwrite the file at outputPath if it already exists
+func Compare(snapshotDir1 string, snapshotDir2 string, outputFile string) (err error) {
+	// Check the hashes between two files
+	hashPath1 := filepath.Join(snapshotDir1, kvledger.SnapshotSignableMetadataFileName)
+	hashPath2 := filepath.Join(snapshotDir2, kvledger.SnapshotSignableMetadataFileName)
+
+	equal, err := verifyAndCompareSnapshotHashes(hashPath1, hashPath2)
```
manish-sethi: Nit: On a second thought, `snapshotsComparable` sounds more meaningful name.

internal/ledger/compare_test.go
```diff
@@ -0,0 +1,597 @@
...
+			"public_state.metadata":         "public_state_hash2",
+			"txids.data":                    "txids_hash2",
+			"txids.metadata":                "txids_hash2",
+		},
+		StateDBType: "testdatabase2",
+	}
+
+	// Expected outputs
+	const (
+		expectedDifferenceResult = `[
```
manish-sethi: just like other variables above
```suggestion
		expectedDifferenceResult := `[
```

internal/ledger/compare.go
```diff
@@ -0,0 +1,413 @@
...
+			r, size := utf8.DecodeRune(b)
+			str += string(r)
+			b = b[size:]
+		}
+	} else {
+		str = string(b)
+	}
+
+	return str
+}
```
manish-sethi: I thought that we had removed this in the previous round as the json encoder internally takes care of this... no?

internal/ledger/compare_test.go
```diff
@@ -0,0 +1,597 @@
...
+			Value:    []byte(sample.value),
+			Metadata: []byte(sample.metadata),
+			Version:  toBytes(sample.blockNum, sample.txNum),
+		})
+		if err != nil {
+			return err
+		}
+	}
+
+	pubStateWriter.Done()
```
manish-sethi: error ignored

internal/ledger/compare_test.go
```diff
@@ -0,0 +1,597 @@
...
+type testRecord struct {
+	namespace string
+	key       string
+	value     string
+	blockNum  uint64
+	txNum     uint64
+	metadata  string
+}
+
+func TestCompare(t *testing.T) {
```
manish-sethi, line 39: Test looks a lot better organized now!

internal/ledger/compare.go
```diff
@@ -0,0 +1,413 @@
...
+
+	equal, err := snapshotsComparable(hashPath1, hashPath2)
+	if err != nil {
+		return err
+	}
+
+	if equal {
+		return errors.New("both snapshots public state hashes are same. Aborting compare")
+	}
+
```
denyeart, line 41: Even if the hashes are equal, we still want to do the comparison, in order to catch the 'silent forks' where just state has deviated but hashes are the same.

internal/ledger/compare.go
```diff
@@ -0,0 +1,413 @@
...
+	err = json.Unmarshal([]byte(f), &mdata2)
+	if err != nil {
+		return false, err
+	}
+
+	if mdata1.ChannelName != mdata2.ChannelName || mdata1.LastBlockNumber != mdata2.LastBlockNumber ||
+		mdata1.LastBlockHashInHex != mdata2.LastBlockHashInHex || mdata1.StateDBType != mdata2.StateDBType {
+		return false, errors.Errorf("the supplied snapshots appear to be non-comparable. The basic details in files %s do not match",
```
denyeart: Let's indicate exactly which fields are different (e.g. if two fields don't match, report the differing values from both fields).

internal/ledger/compare.go
denyeart: Regardless of whether hashes are same or different, we should print the result at this point, and mention that all records will now be compared and it may take some time if there are many records to compare.

cmd/ledger/main.go
```diff
@@ -0,0 +1,54 @@
...
+	switch command {
+
+	case compare.FullCommand():
+
+		err := ledger.Compare(*snapshotPath1, *snapshotPath2, "./result.json")
+		if err != nil {
+			fmt.Println(err)
+			return
+		}
+		fmt.Println("Successfully compared snapshots")
```
denyeart: We'll provide this message whether there were any record differences or not.  I had a record difference, but received this message, and therefore thought the comparison resulted in no differences.  When I actually opened the result.json however, I see that there was one difference identified.

What I would suggest, is count the number of differences and report it here (perhaps separate counters for value differences, missing keys in snapshot1, and missing keys in snapshot2, so that user gets aggregate information). And if there is one or more differences reported, print the JSON file name where the details can be viewed.  If there are no differences, then we don't even need to create the JSON file.

I'm fine if the counters go into the JSON results itself as a header, and the response just points to the file.

jcastrence: That makes a lot of sense to me. I think adding a count would make it clear that "Successfully compared snapshots" means that there were no errors, not that there were no differences found.

Makefile
```diff
@@ -28,11 +28,13 @@
 #   - idemixgen - builds a native idemixgen binary
 #   - integration-test-prereqs - setup prerequisites for integration tests
 #   - integration-test - runs the integration tests
+#   - ledger - builds a native fabric ledger binary
```
manish-sethi: "ledger binary" -> ledger troubleshooting binary

internal/ledger/compare.go
```diff
@@ -0,0 +1,420 @@
...
+				if err != nil {
+					return 0, err
+				}
+				// Add difference to output JSON file
+				err = jsonOutputFile.addRecord(*diffRecord)
+				if err != nil {
+					return 0, err
+				}
+				count++
+
```
manish-sethi: Instead of keeping track of count++ at various places, you can keep track of this inside `jsonOutputFile`

internal/ledger/compare_test.go
```diff
@@ -0,0 +1,599 @@
...
+		return err
+	}
+
+	return nil
+}
+
+// compareSnapshots calls the Compare tool and extracts the result json
+func compareSnapshots(ss1 string, ss2 string, res string) (string, error) {
+	// Run compare tool on snapshots
+	_, err := Compare(ss1, ss2, res)
```
manish-sethi: This function now returns `counts` as well. Please enhance the test to verify the counts as well.

internal/ledger/compare.go
jcastrence: This comparison uses the snapshot hash which is generated when the snapshot is created from the ledger, not the block hash. When the snapshot hash is equal, we know there are no differences between the snapshots.

internal/ledger/compare.go
```diff
@@ -0,0 +1,414 @@
...
+	}
+
+	err = json.Unmarshal([]byte(f), &mdata2)
+	if err != nil {
+		return false, err
+	}
+
+	if mdata1.ChannelName != mdata2.ChannelName {
+		return false, errors.Errorf("the supplied snapshots appear to be non-comparable. Channel names do not match."+
+			"\nSnapshot1 channel name: %s\nSnapshot2 channel name: %s", mdata1.ChannelName, mdata2.ChannelName)
```
denyeart, line 316: This line is not covered by tests.

internal/ledger/compare.go
```diff
@@ -0,0 +1,414 @@
...
+	}
+
+	if mdata1.LastBlockNumber != mdata2.LastBlockNumber {
+		return false, errors.Errorf("the supplied snapshots appear to be non-comparable. Last block numbers do not match."+
+			"\nSnapshot1 last block number: %v\nSnapshot2 last block number: %v", mdata1.LastBlockNumber, mdata2.LastBlockNumber)
```
denyeart, line 321: This line is not covered by tests.

internal/ledger/compare.go
```diff
@@ -0,0 +1,414 @@
...
+	}
+
+	if mdata1.LastBlockHashInHex != mdata2.LastBlockHashInHex {
+		return false, errors.Errorf("the supplied snapshots appear to be non-comparable. Last block hashes do not match."+
+			"\nSnapshot1 last block hash: %s\nSnapshot2 last block hash: %s", mdata1.LastBlockHashInHex, mdata2.LastBlockHashInHex)
```
denyeart, line 326: This line is not covered by tests.

denyeart (approved):
Julian, I tried out the utility, it works well now.

Just two small things...

1) I found a few scenarios that are not covered by unit tests. You can find these uncovered statements by generating a coverage report like this:
```
go test -coverprofile=coverage.out
go tool cover -html=coverage.out -o coverage.html
```

2) I think I found one bug, I've opened https://jira.hyperledger.org/browse/FAB-18508 , please take a look.

But overall, I think this is ready to merge and the final nits can be handled in a follow-on PR.

PR #2545, opened 2021-04-17
Fix Typo

huaweibcs:


mastersingh24 (approved):
While this file was is actually generated by a script and the actual commit message itself will not be updated, this change should be OK as all additions are prepends anyway.

I'll approve unless anyone else has objections

denyeart (approved):
There is no need to update the old changelog, but we appreciate the contribution and will merge it!

PR #2567, opened 2021-04-26
Add log warning for unathorized operations requests

denyeart:
It is not obvious which operations requests require a client cert to be passed
when TLS is enabled on a peer but clientAuthRequired is not enabled.
This new log warning message will help troubleshoot failed connections to operations services
when a client cert is not passed.

Docs are also updated to clarify client cert requirements.

### DIFF
--- a/core/middleware/chain.go
+++ b/core/middleware/chain.go
@@ -8,7 +8,11 @@ package middleware
 
 import (
 	"net/http"
+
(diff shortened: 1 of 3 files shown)


denali49 (approved):
LGTM on the docs update portion.

yacovm (approved):

PR #2570, opened 2021-04-28
FABGW-19: Secure CommitStatus service

bestbeforetoday:
- Receive a SignedCommitStatusRequest message.
- Verify signature and check included identity has CHANNELREADERS permission.
- Add integration tests for success and failure scenarios.


andrew-coleman (dismissed):
looks good to me

core/aclmgmt/resourceprovider_test.go
```diff
@@ -57,17 +57,32 @@ type signerSerializer interface {
...
+	evaluator := &mockPolicyEvaluatorImpl{pmap: map[string]string{"res": "pol"}, peval: map[string]error{"pol": nil}}
+	provider := newPolicyProvider(evaluator)
+
+	t.Run("SignedEndorserProposal", func(t *testing.T) {
+		sProp, _ := protoutil.MockSignedEndorserProposalOrPanic("A", &peer.ChaincodeSpec{}, []byte("Alice"), []byte("msg1"))
+		err := provider.CheckACL("pol", sProp)
+		require.NoError(t, err)
+	})
+
+	t.Run("SignedEndorserProposal", func(t *testing.T) {
```
manish-sethi: Might want to use different names -- `SignedProposal` and `SignedEnvelope`?

internal/pkg/gateway/api.go
```diff
@@ -226,11 +227,25 @@ func (gs *Server) Submit(ctx context.Context, request *gp.SubmitRequest) (*gp.Su
...
+		return nil, status.Error(codes.InvalidArgument, "invalid status request")
+	}
+
+	signedData := &protoutil.SignedData{
+		Data:      signedRequest.Request,
+		Identity:  request.Identity,
+		Signature: signedRequest.Signature,
+	}
+	if err := gs.policyChecker.CheckACL(resources.Event_FilteredBlock, request.ChannelId, signedData); err != nil {
+		return nil, status.Error(codes.FailedPrecondition, err.Error())
```
manish-sethi: - `codes.PermissionDenied` is a better fit here.
- Since you use `resources.Event_FilteredBlock` the underlying error message may be misleading to pass on as is, since that would perhaps say something like "no permission for event/FilteredBlock". Either you may like to define a new resource for this api or decorate the error message here (I personally like the former as it matches the current scheme of one-to-one mapping of api to acl entry).

integration/gateway/gateway_test.go
```diff
@@ -201,17 +209,77 @@ var _ = Describe("GatewayService", func() {
...
+			Expect(err).NotTo(HaveOccurred())
+
+			signedStatusRequest := &gateway.SignedCommitStatusRequest{
+				Request: statusRequestBytes,
+			}
+			signedStatusRequest.Signature, err = signingIdentity.Sign(statusRequestBytes)
+			Expect(err).NotTo(HaveOccurred())
+
+			_, err = gatewayClient.CommitStatus(ctx, signedStatusRequest)
+			Expect(err).To(HaveOccurred())
```
manish-sethi, line 272: In error checks, its desired to check for specific error (response code or error message - whatever makes it unique) so as to make sure that the test does not pass for a wrong reason.

core/aclmgmt/resourceprovider_test.go
bestbeforetoday: Good spot! Bad cut & paste. I was aiming to just use the exact names of the types being tested in the description.

internal/pkg/gateway/api.go
bestbeforetoday: Sounds good to me. Where do I define resources types?

manish-sethi: You can look at this file https://github.com/hyperledger/fabric/blob/main/core/aclmgmt/resources/resources.go and follow the pattern for Event_FilteredBlock

denyeart:
I see where the ACL check was added. Where is the cert check and signature check?  (I see the tests, just didn't see the code that does the checks at first glance, was it already being done, or maybe it is done within the ACL check?).

denyeart:
We could use a PR and commit description :-)

integration/gateway/gateway_suite_test.go
```diff
@@ -94,3 +103,50 @@ func chaincodeArgs(transactionName string, args ...[]byte) [][]byte {
 
 	return result
 }
+
+func NewECDSAPrivateKey() *ecdsa.PrivateKey {
```
manish-sethi: Instead of writing this boiler plate code, my suggestion would be to simply use a built-in user that would not be authorized -- you can use "orderer admin" in this case. 
`network.OrdererUserSigner(network.Orderer("orderer"), "Admin")`

manish-sethi (approved):

PR #2576, opened 2021-05-01
removing -a from docker ps command as it lists stopped containers also

RahulRoy090791:
Signed-off-by: System Administrator <root@Rahuls-MacBook-Air.local>

Removing -a from 'docker ps' commnand, since it lists stopped containers also

#### Type of change

- Bug fix
- New feature
- Test update
- Documentation update

### DIFF
--- a/docs/source/test_network.md
+++ b/docs/source/test_network.md
@@ -132,7 +132,7 @@ components. Run the following command to list all of Docker containers that
 are running on your machine. You should see the three nodes that were created by
 the `network.sh` script:
 ```
-docker ps -a
+docker ps
 ```
 
 Each node and user that interacts with a Fabric network needs to belong to an


lindluni (changes requested):
I think instead of removing the `-a` flag, if you feel strongly about this we should remove the word `running` instead.  It's not uncommon for the containers to crash and removing the flag will prevent users from seeing the crashed containers.

RahulRoy090791:
> I think instead of removing the `-a` flag, if you feel strongly about this we should remove the word `running` instead. It's not uncommon for the containers to crash and removing the flag will prevent users from seeing the crashed containers.

Yes, that is also a good suggestion, we can remove the text 'running'.

denyeart:
I agree with @lindluni , especially for newbies it is good to teach them about `-a` so that they see the full picture of containers.

nikhil550:
Maybe the following language is good enough?

Run the following command to list all of th Docker containers on your machine. You should see the three nodes that were created by the `network.sh` script:
```
docker ps -a 
```

maybe we can add a sentence about crashed nodes, but I think it would be better to assume that everything worked given that this is targeted to beginners.

lindluni:
Closing as abandoned

PR #2578, opened 2021-05-07
FABGW-20 Specify endorsing organizations

andrew-coleman:
Implementation of endorsingOrgs in gateway Endorse method.
- The set of endorsing peers from the discovery service is filtered down to only include those from the specified orgs.
- The discovery handling code has been upgraded to step through all given layouts to find a layout that satisfies the endorsingOrgs constraint.


internal/pkg/gateway/registry.go
```diff
@@ -93,29 +93,51 @@ func (reg *registry) endorsers(channel string, chaincode string) ([]*endorser, e
...
+				} else if e, ok := reg.remoteEndorsers[endpoint]; ok {
+					endorser = e
+				} else {
+					reg.logger.Warnf("Failed to find endorser at %s", endpoint)
+					continue
+				}
+
+				// add to list of endorsing peers, unless endorsingOrgs is specified and it's not in one of the allowed orgs
+				if len(endorsingOrgs) == 0 || contains(endorsingOrgs, endorser.mspid) {
+					groupPeers = append(groupPeers, &endorserState{peer: peer, endorser: endorser, endpoint: endpoint, height: height})
```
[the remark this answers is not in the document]
denyeart: The endorsers service discovery query takes into consideration both the chaincode level and collection level endorsement policy. But Manish is correct, we need to discuss the semantics of endorsingOrgs.

I think the primary use case is when the application knows exactly which orgs to target. In that case, the Service Discovery peers query should be utilized, rather than the more expensive endorsers query. This will often be the case when using state-based endorsement (as SD endorsers query doesn't consider state-based endorsement requirements). See more history in https://jira.hyperledger.org/browse/FABN-1601.

Another use case is when the endorsement policy should be used (for example 3 out of 100 orgs), but you want to target some specific orgs, for example you want to utilize OrgA and OrgB but let service discovery pick any third org.

Bret implemented both approaches for the node SDK, and I think we have to do the same for Gateway.

A note on the endorsers query org filtering approach... Bret used a client side endorsers query filter in node SDK , but that approach doesn't work at all when there are many layouts returned since service discovery may only return a random subset. That means you may entirely miss on the intersection of layouts returned and orgs required. The filter therefore must be done on the service discovery side - we need to consider extending the SD endorsers query to take the list of required orgs.  In fact, this is why we have a single team responsible for Gateway and Service Discovery now.

internal/pkg/gateway/registry.go
```diff
@@ -41,12 +41,13 @@ type registry struct {
 
 type endorserState struct {
 	peer     *dp.Peer
+	endorser *endorser
 	endpoint string
```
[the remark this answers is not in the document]
andrew-coleman: Good spot.  Removed.

internal/pkg/gateway/registry.go
```diff
@@ -93,29 +93,51 @@ func (reg *registry) endorsers(channel string, chaincode string) ([]*endorser, e
...
 
 		for _, peer := range r {
-			if peer.endpoint == reg.localEndorser.address {
-				endorsers = append(endorsers, reg.localEndorser)
-			} else if endorser, ok := reg.remoteEndorsers[peer.endpoint]; ok {
-				endorsers = append(endorsers, endorser)
-			} else {
-				reg.logger.Warnf("Failed to find endorser at %s", peer.endpoint)
-			}
+			endorsers = append(endorsers, peer.endorser)
```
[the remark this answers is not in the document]
andrew-coleman: That's my assumption too.  I followed the usage algorithm documented in the proto file:
https://github.com/hyperledger/fabric-protos/blob/main/discovery/protocol.proto#L175
My observation is that each group contains the peers of a single org, and the layouts give the various permutations of those orgs/groups

internal/pkg/gateway/registry.go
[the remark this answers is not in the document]
andrew-coleman: I've reimplemented this as Dave suggests to return all the peers from the list of 'endorsingOrganizations' that have the given chaincode installed.

internal/pkg/gateway/registry.go
```diff
@@ -93,24 +90,77 @@ func (reg *registry) endorsers(channel string, chaincode string) ([]*endorser, e
...
+		for _, peer := range receivers {
+			endorsers = append(endorsers, peer.endorser)
+		}
+		return endorsers, nil
+	}
+
+	return nil, fmt.Errorf("failed to select a set of endorsers that satisfy the endorsement policy")
+}
+
+// endorsersForOrgs returns a set of endorsers oen by the given orgs for the given chaincode on a channel.
```
manish-sethi: what does "oen" mean?

internal/pkg/gateway/registry.go
```diff
@@ -93,24 +90,77 @@ func (reg *registry) endorsers(channel string, chaincode string) ([]*endorser, e
...
+func (reg *registry) endorsersForOrgs(channel string, chaincode string, endorsingOrgs []string) ([]*endorser, error) {
+	err := reg.registerChannel(channel)
+	if err != nil {
+		return nil, err
+	}
+
+	var endorsers []*endorser
+	members := reg.discovery.PeersOfChannel(common.ChannelID(channel))
```
manish-sethi, line 146: Nothing to be done in this PR, However, just to take a note, this appears to be case to revisit if the discovery api/code could be extended to take the org names as a filter that could improve the efficiency of this as compared to get all the peers and iterate over to filter.

internal/pkg/gateway/registry.go
```diff
@@ -93,24 +90,77 @@ func (reg *registry) endorsers(channel string, chaincode string) ([]*endorser, e
...
+		} else {
+			reg.logger.Warnf("Failed to find endorser at %s", endpoint)
+			continue
+		}
+		// check the org membership
+		if contains(endorsingOrgs, endorser.mspid) && member.Properties != nil {
+			for _, installedChaincode := range member.Properties.Chaincodes {
+				// only consider the peers that have our chaincode installed
+				if installedChaincode.Name == chaincode {
+					endorsers = append(endorsers, endorser)
```
manish-sethi: Like the existing function, I was expecting here as well - only one peer (with the maximum height) for one org. Am I missing something?

internal/pkg/gateway/registry.go
```diff
@@ -93,41 +90,113 @@ func (reg *registry) endorsers(channel string, chaincode string) ([]*endorser, e
...
 		}
 	}
 
+	var endorsers []*endorser
+	for _, es := range endorsersByOrg {
+		// sort by decreasing height and select the highest one from each org
+		sort.Slice(es, sorter(es, reg.localEndorser.address))
+		if len(es) > 0 {
+			endorsers = append(endorsers, es[0].endorser)
+		}
```
manish-sethi: Isn't it a noop check?

manish-sethi (approved):

internal/pkg/gateway/registry.go
andrew-coleman: If, for any reason, the endorser returned by the discovery service is not in the gateway registry, then this _could_ be an empty array.  Perhaps if a new endorser had just come online and not been added to the registry yet.  There's still a TODO comment in `registerChannels()` to handle this type of dynamic update.

manish-sethi: Discussed offline and @andrew-coleman is going to push a new patch. Holding it from merging for now.

manish-sethi (changes requested):
Based on off-line discussion, expecting a minor fix and test for handling the case of insufficient available peers

internal/pkg/gateway/registry.go
andrew-coleman: Thanks Manish, I've added an error check that will report which orgs fail to find endorsing peers, and added extra unit tests

manish-sethi (approved):

PR #2580, opened 2021-05-08
fix typo

huaweibcs:


yacovm (changes requested):
This is not a typo. 

https://en.wikipedia.org/wiki/Memoization

huaweibcs:
> This is not a typo.
> 
> https://en.wikipedia.org/wiki/Memoization

thanks for correct

PR #2581, opened 2021-05-08
Add vendor directories into chaincode code package tar when install golang chaincode

xuchengli:
Signed-off-by: xuchengli <lixucheng@aliyun.com>

If chaincode dependency have multi-level directory, only META-INF metadata directory will be added into chaincode code package tar, other dependency directories will be skipped. So when instantiate chaincode it will throw error of missing file or directory. By adding vendor directory at the root of chaincode directory will avoid the issue.

Jira issue: https://jira.hyperledger.org/browse/FAB-18467

- Bug fix

Test:
1. Compile peer binary based on the patch.
2. Restart peer using the latest peer binary.
3. Install and instantiate the chaincode with multi-level vendor dependency directories. The test chaincode is: https://github.com/xuchengli/fabric-chaincode/tree/main/src/test-1

### DIFF
--- a/core/chaincode/platforms/golang/package.go
+++ b/core/chaincode/platforms/golang/package.go
@@ -105,10 +105,11 @@ func findSource(gopath, pkg string) (SourceMap, error) {
 				return nil
 			}
 
-			// Allow import of META-INF metadata directories into chaincode code package tar.
+			// Allow import of META-INF metadata directories and vendor directories into chaincode code package tar.
 			// META-INF directories contain chaincode metadata artifacts such as statedb index definitions
-			if isMetadataDir(path, tld) {
-				logger.Debug("Files in META-INF directory will be included in code package tar:", path)
+			// vendor directories contain chaincode vendor dependencies
+			if isMetadataDir(path, tld) || isVendorDir(path, tld) {
+				logger.Debug("Files in META-INF and vendor directory will be included in code package tar:", path)
 				return nil
 			}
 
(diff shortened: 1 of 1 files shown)


lindluni:
I don't see this behavior exhibited. I've used a chaincode where I vendor the dependencies with private data and a META-INF directory. When I exec into the container and cd `/var/hyperledger/production/lifecycle` and untar the chaincode package and then the code package the vendor directory is correctly inside the `src` directory.

xuchengli:
The issue appears in version 1.4, and in chaincode dependency directory, all subdirectories will be skipped except META-INF directory.  For example: there is a vendor dir ( vendor/github.com/xx/a.go, vendor/github.com/xx/**subdir**/b.h ), the b.h is the dependency of a.go, but after installed, the "vendor/github.com/xx/**subdir** will skipped. So there will throw missing file or directory error when instantiate chaincode.

lindluni:
Again, I tried this just now on 1.4 with a deeply nested dependency in the vendor directory where b is a sub-dir and a depends on b, and I don't see the behavior. I'll let @sykesm take a look at the actual code, but I wasn't able to recreate this as described.

xuchengli:
@lindluni The sub-dir in the vendor dependency is pseudo C package dependencies, `go list` can not import the `#include <dep>` automatically. I copy the sub-dir of pseudo C package dependencies to the vendor dependency manually.

I write a test chaincode: https://github.com/xuchengli/fabric-chaincode/tree/main/src/test-1 , this chaincode can reproduce the issue.

lindluni:
Ah, this has come up several times in the past. We've elected to punt on supporting CGO directly in chaincode, as users can extend the `ccenv` image themselves to include the `C` libraries they need directly in the image. This was decided as the path forward several years ago.

Fabric 1.4 is effectively EOL, with the announcement going out any day now. In 2.x Fabric you can use the custom external builder as a way around this there, or continue extending the ccenv image.

xuchengli:
@lindluni ok, got it, you can close it. Thanks.

PR #2582, opened 2021-05-09
Correcting a possible typo

realnimish:
#### Type of change

- Documentation update

#### Description

I believe it should be "Now that R3 is able to fully participate in channel C1" instead of "Now that R3 is able to fully participate in channel C2" because the latter one doesn't make sense to me as R3 already had the permission to participate in channel C2 and given the context having C1 makes sense.

P.S: Apologies if perhaps I got something wrong, I just started learning Hyperledger Fabric :)

### DIFF
--- a/docs/source/network/network.md
+++ b/docs/source/network/network.md
@@ -152,7 +152,7 @@ For more information about upgrading a chaincode, check out [Upgrade a chaincode
 
 ### Adding existing components to the newly joined channel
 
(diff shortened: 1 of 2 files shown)


denyeart (approved):

realnimish:
Thanks and I would like to mention my dear friend @namannimmo10 who helped me out clear the DCO check. I kinda messed up in such a simple pull request ;)

PR #2584, opened 2021-05-10
Documentation updated

kemi04:
Signed-off-by: Kestutis Gudynas <44440041+kemi04@users.noreply.github.com>

### DIFF
--- a/docs/source/chaincode4ade.rst
+++ b/docs/source/chaincode4ade.rst
@@ -383,7 +383,7 @@ Now let's compile your chaincode.
 
 .. code:: bash
 
-  go get -u github.com/hyperledger/fabric/core/chaincode/shim
+  go get -u github.com/hyperledger/fabric/core/chaincode/shim@v1.4
   go build
 
 Assuming there are no errors, now we can proceed to the next step, testing


lindluni (changes requested):
Thanks for the contribution!

While I recognize `govendor` is deprecated and specifically calls out `go modules` as the replacement, I'm not opposed to this change, but I don't think it highlights the fact that 1.4 fabric explicitly did not support Go Modules. I think you need to start by stating that "While Fabric doesn't support Go Modules directly, you can use Go Modules in Vendor Mode to version and vendor your chaincode dependencies. Note: your chaincode will not work if you only implement modules and don't vendor your dependencies"

And then in the paragraph where you explain `go mod vendor` I would again call out the EXPLICIT requirement to vendor your dependencies as Fabric does not support Go Modules outside of Vendor Mode

denyeart:
@kemi04 sorry for the confusion... when we talked about updating the docs, here's what i was thinking...

- keep the `govendor` command in the release-1.4 docs as-is at https://hyperledger-fabric.readthedocs.io/en/release-1.4/chaincode4ade.html#managing-external-dependencies-for-chaincode-written-in-go
- keep the `go mod vendor` command in the release-2.x docs as-is at https://hyperledger-fabric.readthedocs.io/en/latest/chaincode4ade.html#managing-external-dependencies-for-chaincode-written-in-go
- update the upgrade instructions to mention that you can use either the `govendor` approach or `go mod vendor` approach , at https://hyperledger-fabric.readthedocs.io/en/latest/upgrade_to_newest_version.html#chaincode-shim-changes-go-chaincode-only (so that people coming from 1.4 world that used govendor previously will know that they can continue using govendor, and that they have the option to start using Go modules).

denali49 (approved):
lgtm, doc review only.  Thank you for your contribution!

lindluni (approved):

docs/source/chaincode4ade.rst
```diff
@@ -383,7 +383,7 @@ Now let's compile your chaincode.
 
 .. code:: bash
 
-  go get -u github.com/hyperledger/fabric/core/chaincode/shim
+  go get -u github.com/hyperledger/fabric/core/chaincode/shim@v1.4
```
denyeart, line 386: @kemi04 @lindluni 
Did you guys try this?
I get "no matching versions for query "v1.4".

lindluni: I tried it, it worked for me, let me try again

lindluni: Ah, thats interesting, if you are in your GOPATH, this fails. If you are outside your GOPATH it succeeds by pulling the latest 1.4 release.

lindluni: After some playing around Dave and I discovered this command works but only if you are on the `release-1.4` branch with no `go.mod`

PR #2586, opened 2021-05-10
NewSnapshotReader Nil Error

jcastrence:
#### Type of change

#### Description

- Edited NewSnapshotReader so it now logs if the passed in data file does not exist

#### Related issues

- [FAB-18469](https://jira.hyperledger.org/browse/FAB-18469)

### DIFF
--- a/core/ledger/kvledger/txmgmt/privacyenabledstate/snapshot.go
+++ b/core/ledger/kvledger/txmgmt/privacyenabledstate/snapshot.go
@@ -437,12 +437,17 @@ type snapshotReader struct {
 	cursor   *cursor
 }
 
+// If the passed in file name does not exist a nil response is returned
 func newSnapshotReader(dir, dataFileName, metadataFileName string) (*snapshotReader, error) {
 	dataFilePath := filepath.Join(dir, dataFileName)
 	metadataFilePath := filepath.Join(dir, metadataFileName)
 	exist, _, err := fileutil.FileExists(dataFilePath)
-	if err != nil || !exist {
-		return nil, err
+	if err != nil {
+		return nil, errors.WithMessage(err, "error while checking if data file exists")
+	}
+	if !exist {
+		logger.Infow("Data file does not exist. Nothing to be done.", "filepath", dataFilePath)
+		return nil, nil
 	}
 
 	var dataFile, metadataFile *snapshot.FileReader


C0rWin (dismissed):

core/ledger/kvledger/txmgmt/privacyenabledstate/snapshot.go
```diff
@@ -441,9 +441,12 @@ func newSnapshotReader(dir, dataFileName, metadataFileName string) (*snapshotRea
 	dataFilePath := filepath.Join(dir, dataFileName)
 	metadataFilePath := filepath.Join(dir, metadataFileName)
 	exist, _, err := fileutil.FileExists(dataFilePath)
-	if err != nil || !exist {
+	if err != nil {
 		return nil, err
 	}
+	if !exist {
+		return nil, errors.New("file does not exist")
```
manish-sethi:
```suggestion
		return nil, errors.Errorf("file [%s] does not exist", dataFilePath)
```

core/ledger/kvledger/txmgmt/privacyenabledstate/snapshot_test.go
```diff
@@ -363,6 +363,9 @@ func TestSnapshotReaderNextFunction(t *testing.T) {
 	require.NoError(t, err)
 	require.Equal(t, "", retrievedNs)
 	require.Nil(t, retrievedSr)
+
+	_, err = newSnapshotReader("test-dir", "test-data-filename", "test-metadata-filename")
+	require.EqualError(t, err, "file does not exist")
```
manish-sethi: This test is misplaced, under `TestSnapshotReaderNextFunction`. Moving this under `TestSnapshotImportErrorPropagation` makes a better fit.

manish-sethi:
My suggestion was to simply add the following code in the test file at [line 535](https://github.com/hyperledger/fabric/blob/main/core/ledger/kvledger/txmgmt/privacyenabledstate/snapshot_test.go#L535).

```
		t.Run("error_missing_data_file_"+f, func(t *testing.T) {
			init()
			defer cleanup()

			dataFile := filepath.Join(snapshotDir, f)
			require.NoError(t, os.Remove(dataFile))
			err := dbEnv.GetProvider().ImportFromSnapshot(
				generateLedgerID(t), version.NewHeight(10, 10), snapshotDir)
			require.Contains(t, err.Error(), fmt.Sprintf("file [%s] does not exist", dataFile))
		})
```

jcastrence:
> My suggestion was to simply add the following code in the test file at [line 535](https://github.com/hyperledger/fabric/blob/main/core/ledger/kvledger/txmgmt/privacyenabledstate/snapshot_test.go#L535).
> 
> ```
> 		t.Run("error_missing_data_file_"+f, func(t *testing.T) {
> 			init()
> 			defer cleanup()
> 
> 			dataFile := filepath.Join(snapshotDir, f)
> 			require.NoError(t, os.Remove(dataFile))
> 			err := dbEnv.GetProvider().ImportFromSnapshot(
> 				generateLedgerID(t), version.NewHeight(10, 10), snapshotDir)
> 			require.Contains(t, err.Error(), fmt.Sprintf("file [%s] does not exist", dataFile))
> 		})
> ```

Right, however the changes I've made to snapshot.go have caused some of the existing test cases to fail since they are missing either public state or private state hashes. Because of the addition of a new error being thrown if files are missing, they will fail since they are not expecting this new error. I thought that it'd be best to then test these test cases with missing files separately which ends up covering the addition you suggested. What are your thoughts on this?

manish-sethi:
> What are your thoughts on this?

My suggestion would be to fix the input to the existing test.

denyeart:
Upon further discussion the nil response is expected in some scenarios (such as no private state data) so we can simply leave the functionality as-is and improve the comments and logging to make the intended use more clear.

core/ledger/kvledger/txmgmt/privacyenabledstate/snapshot.go
```diff
@@ -437,12 +438,17 @@ type snapshotReader struct {
...
 	dataFilePath := filepath.Join(dir, dataFileName)
 	metadataFilePath := filepath.Join(dir, metadataFileName)
 	exist, _, err := fileutil.FileExists(dataFilePath)
-	if err != nil || !exist {
-		return nil, err
+	if err != nil {
+		return nil, errors.WithMessage(err, "error while checking if data file exists")
+	}
+	if !exist {
+		log.Printf("Snapshot data file name does not exist")
```
manish-sethi:
```suggestion
	        logger.Infow("Data file does not exist. Nothing to be done.", "filepath", dataFilePath)
```

manish-sethi (approved):

PR #2588, opened 2021-05-13
Govendor added to documentation

kemi04:
Signed-off-by: Kestutis Gudynas <kestutis.gudynas@ibm.com>

### DIFF
--- a/docs/source/upgrade_to_newest_version.md
+++ b/docs/source/upgrade_to_newest_version.md
@@ -33,7 +33,8 @@ For information about how to edit the relevant channel configurations to enable 
 
 ### Chaincode shim changes (Go chaincode only)
 
-The recommended approach is to vendor the shim in your v1.4 Go chaincode before making upgrades to the peers and channels. If you are already using an existing tool such as ``govendor`` you may continue using it to vendor the new chaincode shim. Best practice, however, would be to use Go modules to vendor the new chaincode shim, as modules are now the de facto standard for dependency management in the Go ecosystem. Note that since Fabric 2.0.0, chaincode using Go modules without vendored dependencies is also supported. If you do this, you do not need to make any additional changes to your chaincode.
+The v2.x `ccenv` image that is used to build Go chaincodes no longer automatically vendors the Go chaincode shim dependency like the v1.4 `ccenv` image did. 
+The recommended approach is to vendor the shim in your v1.4 Go chaincode before making upgrades to the peers and channels, since this approach works with both a v1.4.x and v2.x peer. If you are already using an existing tool such as ``govendor`` you may continue using it to vendor the chaincode shim. Best practice, however, would be to use Go modules to vendor the chaincode shim, as modules are now the de facto standard for dependency management in the Go ecosystem. Note that since Fabric v2.0, chaincode using Go modules without vendored dependencies is also supported. If you do this, you do not need to make any additional changes to your chaincode.
 
 If you did not vendor the shim in your v1.4 chaincode, the old v1.4 chaincode images will still technically work after upgrade, but you are in a risky state. If the chaincode image gets deleted from your environment for whatever reason, the next invoke on v2.x peer will try to rebuild the chaincode image and you'll get an error that the shim cannot be found.
 
--- a/docs/source/upgrade_to_newest_version.md
+++ b/docs/source/upgrade_to_newest_version.md
@@ -33,7 +33,7 @@ For information about how to edit the relevant channel configurations to enable 
 
 ### Chaincode shim changes (Go chaincode only)
 
-The recommended approach is to vendor the shim in your v1.4 Go chaincode before making upgrades to the peers and channels. If you do this, you do not need to make any additional changes to your chaincode.
+The recommended approach is to vendor the shim in your v1.4 Go chaincode before making upgrades to the peers and channels. If you are already using an existing tool such as ``govendor`` you may continue using it to vendor the new chaincode shim. Best practice, however, would be to use Go modules to vendor the new chaincode shim, as modules are now the de facto standard for dependency management in the Go ecosystem. Note that since Fabric 2.0.0, chaincode using Go modules without vendored dependencies is also supported. If you do this, you do not need to make any additional changes to your chaincode.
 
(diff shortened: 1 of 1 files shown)


docs/source/upgrade_to_newest_version.md
```diff
@@ -39,7 +39,7 @@ If you did not vendor the shim in your v1.4 chaincode, the old v1.4 chaincode im
 
 At this point, you have two options:
 
-1. If the entire channel is ready to upgrade chaincode, you can upgrade the chaincode on all peers and on the channel (using either the old or new lifecycle depending on the `Application` capability level you have enabled). The best practice at this point would be to vendor the new Go chaincode shim using modules.
+1. If the entire channel is ready to upgrade chaincode, you can upgrade the chaincode on all peers and on the channel (using either the old or new lifecycle depending on the `Application` capability level you have enabled). The best practice at this point would be to vendor the new Go chaincode shim using modules, however, you can also use ``govendor``.
```
lindluni:
```suggestion
1. If the entire channel is ready to upgrade chaincode, you can upgrade the chaincode on all peers and on the channel (using either the old or new lifecycle depending on the `Application` capability level you have enabled). The best practice at this point would be to vendor the new Go chaincode shim. If you are already using an existing tool such as ``govendor`` you may continue using it to vendor the new chaincode shim. Best practice, however, would be to use Go modules to vendor the new chaincode shim, as modules are now the de facto standard for dependency management in the Go ecosystem. Since Fabric 2.0.0, chaincode using Go modules without vendored dependencies is also supported.
```

I think we need to be a little more explicit here, as Fabric 2.0.0 added support for both Go Modules in Vendor Mode, or without Vendor mode. Users should be able to chose which approach they want to follow. Especially since our chaincode samples use modules without vendored dependencies.

lindluni (changes requested):

docs/source/upgrade_to_newest_version.md
kemi04: Thank you for the suggestion. I updated the PR according to your comment.

lindluni (approved):

lindluni:
@denyeart are you okay with this as the 2.x change?

denali49 (dismissed):
lgtm, thanks!

lindluni (dismissed):
I just noticed you are using your github no-reply email address to sign your commits. The Linux Foundation DCO requirements require that users sign commits with valid email signatures containing valid contact email addresses. You will need to update this commit using a valid email address

docs/source/upgrade_to_newest_version.md
```diff
@@ -39,7 +39,7 @@ If you did not vendor the shim in your v1.4 chaincode, the old v1.4 chaincode im
 
 At this point, you have two options:
 
-1. If the entire channel is ready to upgrade chaincode, you can upgrade the chaincode on all peers and on the channel (using either the old or new lifecycle depending on the `Application` capability level you have enabled). The best practice at this point would be to vendor the new Go chaincode shim using modules.
+1. If the entire channel is ready to upgrade chaincode, you can upgrade the chaincode on all peers and on the channel (using either the old or new lifecycle depending on the `Application` capability level you have enabled). The best practice at this point would be to vendor the new Go chaincode shim. If you are already using an existing tool such as ``govendor`` you may continue using it to vendor the new chaincode shim. Best practice, however, would be to use Go modules to vendor the new chaincode shim, as modules are now the de facto standard for dependency management in the Go ecosystem. Since Fabric 2.0.0, chaincode using Go modules without vendored dependencies is also supported.
```
denyeart: I'm good with the technical content, this is exactly what I had in mind, but I think it would be better in the first paragraph of this section to help people get in front of the issue before they upgrade, rather than put it down here which is the section about how to recover if you get into this unfortunate situation after upgrade.

denyeart:
Beyond the signature issue that Brett mentioned, would be nice to have a descriptive commit/PR title and message.

After we merge this in main, we'll also backport to release-2.3 and release-2.2.

docs/source/upgrade_to_newest_version.md
kemi04: Moved it, thank you.

docs/source/upgrade_to_newest_version.md
```diff
@@ -33,7 +33,7 @@ For information about how to edit the relevant channel configurations to enable
 
 ### Chaincode shim changes (Go chaincode only)
 
-The recommended approach is to vendor the shim in your v1.4 Go chaincode before making upgrades to the peers and channels. If you do this, you do not need to make any additional changes to your chaincode.
+The recommended approach is to vendor the shim in your v1.4 Go chaincode before making upgrades to the peers and channels. If you are already using an existing tool such as ``govendor`` you may continue using it to vendor the new chaincode shim. Best practice, however, would be to use Go modules to vendor the new chaincode shim, as modules are now the de facto standard for dependency management in the Go ecosystem. Note that since Fabric 2.0.0, chaincode using Go modules without vendored dependencies is also supported. If you do this, you do not need to make any additional changes to your chaincode.
```
denyeart:
```suggestion
The v2.x `ccenv` image that is used to build Go chaincodes no longer automatically vendors the Go chaincode shim dependency like the v1.4 `ccenv` image did. 
The recommended approach is to vendor the shim in your v1.4 Go chaincode before making upgrades to the peers and channels, since this approach works with both a v1.4.x and v2.x peer. If you are already using an existing tool such as ``govendor`` you may continue using it to vendor the chaincode shim. Best practice, however, would be to use Go modules to vendor the chaincode shim, as modules are now the de facto standard for dependency management in the Go ecosystem. Note that since Fabric v2.0, chaincode using Go modules without vendored dependencies is also supported. If you do this, you do not need to make any additional changes to your chaincode.
```

denyeart (commented):
Thanks Kes, looking good, I'm also suggesting to add one more sentence at the beginning to explain the situation.

denali49 (approved):
lgtm on docs.  Thanks for your contribution!

denyeart (approved):

denyeart:

denyeart:

PR #2594, opened 2021-05-17
Adding block height check into orderer bootstrap

Param-S:
#### Type of change
- Bug fix

#### Description
1. Channel with empty ledger led to crash in 2.2 release
Higher version added a check to skip the channels with
empty ledger & continue to start up. The same changes
backported. 

2. Updated GetBlock construct to retrieve the block by number 
rather than using Iterator.

### DIFF
--- a/orderer/common/multichannel/registrar.go
+++ b/orderer/common/multichannel/registrar.go
@@ -113,16 +113,18 @@ type Registrar struct {
 // ConfigBlock retrieves the last configuration block from the given ledger.
 // Panics on failure.
 func ConfigBlock(reader blockledger.Reader) *cb.Block {
-	lastBlock := blockledger.GetBlock(reader, reader.Height()-1)
+	lastBlock, err := blockledger.GetBlockByNumber(reader, reader.Height()-1)
+	if err != nil {
+		logger.Panicw("Failed to retrieve block", "blockNum", reader.Height()-1, "error", err)
+	}
 	index, err := protoutil.GetLastConfigIndexFromBlock(lastBlock)
 	if err != nil {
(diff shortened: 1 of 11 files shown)


yacovm:
Shouldn't we fix instead the core issue which is that the ledger creation for a channel isn't atomic?

orderer/common/multichannel/registrar.go
```diff
@@ -113,16 +113,18 @@ type Registrar struct {
 // ConfigBlock retrieves the last configuration block from the given ledger.
 // Panics on failure.
 func ConfigBlock(reader blockledger.Reader) *cb.Block {
```
manish-sethi, line 115: Can you add a UT for this function?

Param-S: Yes, I will update the PR with UT

manish-sethi: The main intention of adding the UT was to cover the panic paths. Given that this function can take a mock as input, it should be straight forward to write this.

manish-sethi: Also, as you may want to make these changes into master as well. I'll suggest to brake these changes into two. One the original backporting PR as is and new changes in the master first.

Param-S: I corrected the UT now covering all error scenarios.  Please check.

Should i need to create a separate PR for master branch?

Param-S: https://github.com/hyperledger/fabric/pull/2635 created a separate PR for main

Param-S:
Opened different PRs. Closing this one

PR #2598, opened 2021-05-18
Bump vmImage to Ubuntu-20.04

lindluni:
Just keeping up with latest deps


mastersingh24 (approved):
LGTM

PR #2600, opened 2021-05-18
Remove Excessive Unit Test Logging

jcastrence:
#### Type of change

- Test update

#### Description

- Removed excessive logging in unit tests by changing logging from debug to info in multiple test files

### DIFF
--- a/common/ledger/blkstorage/pkg_test.go
+++ b/common/ledger/blkstorage/pkg_test.go
@@ -24,7 +24,7 @@ import (
 )
 
 func TestMain(m *testing.M) {
-	flogging.ActivateSpec("blkstorage=debug")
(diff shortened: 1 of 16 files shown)


manish-sethi (dismissed):

lindluni:
Just curious, why decrease the log level, unit tests don't dump their logs unless a test fails, why wouldn't we want the tests in debug mode on failure?

jcastrence:
> Just curious, why decrease the log level, unit tests don't dump their logs unless a test fails, why wouldn't we want the tests in debug mode on failure?

We thought it might be better to allow for the developer to turn on debugging for only specific tests to reduce the clutter when unit tests fail, but we've decided this isn't a change worth pursuing anymore.

PR #2604, opened 2021-05-25
FABGW-21: Realtime implementation of ChaincodeEvents service

bestbeforetoday:
Initial implementation of Fabric Gateway's ChaincodeEvents service, allowing only listening for realtime chaincode events.

Also add block number into CommitStatus response.


internal/pkg/gateway/commit/commitadapter.go
```diff
@@ -28,13 +27,32 @@ func (adapter *PeerAdapter) CommitNotifications(done <-chan struct{}, channelNam
...
 	}
 
-	return channel.Ledger().GetTxValidationCodeByTxID(transactionID)
+	block, err := ledger.GetBlockByTxID(transactionID)
+	if err != nil {
+		return nil, err
+	}
+	if block.Header == nil {
+		return nil, errors.Errorf("missing header for block containing transaction %s", transactionID)
+	}
```
C0rWin: This appears a bit weird to me, as you check whenever the header is not nil, while not checking whenever a block is not nil. not something expected to happen here. However, since you are already adding such a check for the header, I would expect to have a similar check for the block itself.

bestbeforetoday: Won't `GetBlockByTxID()` return a non-nil error if there is no block?

C0rWin: Well, it will also return a valid non-nil header in case there are no errors. I guess it depends on what was your intent while adding this check for the non-nil header.

See: 

```go
func deserializeBlock(serializedBlockBytes []byte) (*common.Block, error) {
	block := &common.Block{}
	var err error
	b := newBuffer(serializedBlockBytes)
	if block.Header, err = extractHeader(b); err != nil {
		return nil, err
	}
	if block.Data, _, err = extractData(b); err != nil {
		return nil, err
	}
	if block.Metadata, err = extractMetadata(b); err != nil {
		return nil, err
	}
	return block, nil
}
```

How the block is actually being fetched. Now, this is a nit comment, not a real show stopper, just for the sake of consistency and clarity. Like, why did you check for non-nil header?

bestbeforetoday: It's there because any time I deliver code that deals with protobufs and doesn't explicitly check for nil, it gets flagged in review comments. I'm quite happy to remove the check if that's the preferred approach.

bestbeforetoday: I've pushed an update that removes the nil check on the block header

C0rWin: My initial comment and intent were not to remove the check, but for sanity and clearness, add the check to ensure the block is not nil neither. ;)

bestbeforetoday: I'm not convinced we can reasonably expect `GetBlockByTxID()` to return `nil, nil`. The error response is there to indicate that a block isn't returned. It seems confusing and non-idiomatic to check both the value and the associated error response for nil.

bestbeforetoday:
Refactored to break cyclic dependency and enable the use of Counterfeiter as the mocking framework again.

integration/gateway/gateway_test.go
```diff
@@ -238,13 +238,11 @@ var _ = Describe("GatewayService", func() {
...
-			actualStatus, err := gatewayClient.CommitStatus(ctx, signedStatusRequest)
+			status, err := gatewayClient.CommitStatus(ctx, signedStatusRequest)
 			Expect(err).NotTo(HaveOccurred())
 
-			expectedStatus := &gateway.CommitStatusResponse{
-				Result: peer.TxValidationCode_VALID,
-			}
-			Expect(proto.Equal(actualStatus, expectedStatus)).To(BeTrue(), "Expected\n\t%#v\nto proto.Equal\n\t%#v", actualStatus, expectedStatus)
+			Expect(status.Result).To(Equal(peer.TxValidationCode_VALID))
+			Expect(status.BlockNumber).To(BeNumerically(">", 0))
```
manish-sethi: Better to use the exact block number here.

internal/pkg/gateway/commit/finder.go
```diff
@@ -30,26 +36,34 @@ type Finder struct {
...
 	notifyDone := make(chan struct{})
 	defer close(notifyDone)
 	commitReceive, err := finder.Notifier.notify(notifyDone, channelName, transactionID)
 	if err != nil {
-		return 0, err
+		return nil, err
 	}
 
-	if status, err := finder.Query.TransactionStatus(channelName, transactionID); err == nil {
+	if code, blockNumber, err := finder.Query.TransactionStatus(channelName, transactionID); err == nil {
```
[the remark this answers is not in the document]
bestbeforetoday: There should only be an error here if the transaction does not exist (i.e. is not already committed), in which case we want to wait for it to be committed, not return an error

internal/pkg/gateway/commit/commitadapter.go
```diff
@@ -28,13 +28,28 @@ func (adapter *PeerAdapter) CommitNotifications(done <-chan struct{}, channelNam
...
 
-	return channel.Ledger().GetTxValidationCodeByTxID(transactionID)
+	ledger := channel.Ledger()
+
+	status, err := ledger.GetTxValidationCodeByTxID(transactionID)
+	if err != nil {
+		return 0, 0, err
+	}
+
+	block, err := ledger.GetBlockByTxID(transactionID)
```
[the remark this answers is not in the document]
bestbeforetoday: Logging and debugging are the primary uses. It might also be useful as a constraint for future interactions, since we know the minimum block height required to see this transaction, but the Gateway is not making use of this today so not a showstopper.

Being able to grab just the block number out of the ledger index would be great, either along with the transaction ID or separately. If that's not practical then I'll give up on including the block number along with the validation code so I can make some progress on chaincode eventing, which also requires changes to this code.

integration/gateway/gateway_test.go
bestbeforetoday: How do we know what the block number will be without introducing a fragile dependency on other integration tests and their ordering?

internal/pkg/gateway/commit/commitadapter.go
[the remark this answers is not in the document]
manish-sethi: Yes, I was thinking on the lines of enhancing `GetTxValidationCodeByTxID` to return the block number along with the validation code. However, your PR will have to wait for that change. So one option could be to do that outside this PR. I leave the choice to you. If you prefer to do in this PR, I can spend some time on enhancing the above API

integration/gateway/gateway_test.go
manish-sethi: Not sure I understand the scope when you say other integration tests but before `Describe("CommitStatus"`, it sets up a fresh network...no?

bestbeforetoday: My worry is, what is the block number I should pick? And are we guaranteed that will not change in the future, either because of changes to Fabric implementation or because somebody comes along and adds another test case to this Describe block? I think I'd prefer to test that some sane block number came back rather than deliver fragile tests that may break in the future for reasons not connected to this particular test. It's a moot point for now as we're not happy with the approach for obtaining the block number so I'll remove the assertion for now

internal/pkg/gateway/commit/finder.go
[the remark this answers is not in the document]
bestbeforetoday: OK, I buy that `GetTxValidationCodeByTxID` may return an error for situation in which it makes more sense to return an error. How can I distinguish between them? What I definitely can't do is never wait for transactions that have not yet been committed, so I either need a completely robust way of distinguishing the different error conditions, or just go ahead with the listening anyway and hope for the best. The worst that happens is eventually the client gives up waiting and I stop listening anyway.

bestbeforetoday:
@manish-sethi Moving back to draft to continue changes to support chaincode event listening, at least until there is a suitable mechanism to grab block number along with transaction status and a satisfactory interim PR can be delivered.

manish-sethi:
> @manish-sethi Moving back to draft to continue changes to support chaincode event listening, at least until there is a suitable mechanism to grab block number along with transaction status and a satisfactory interim PR can be delivered.

@bestbeforetoday - The [PR](https://github.com/hyperledger/fabric/pull/2614) for block number is merged now.

internal/pkg/gateway/commit/commitadapter.go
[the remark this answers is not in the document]
bestbeforetoday: Now using the new implementation to get block number directly from the `GetTxValidationCodeByTxID` call

integration/gateway/gateway_test.go
bestbeforetoday: I've changed the test to check the commit status of two subsequent transactions so I can test that the block number of the second commit is one higher than the block number of the first

internal/pkg/gateway/commit/blocknotifier.go
```diff
@@ -0,0 +1,76 @@
...
+import (
+	"sync"
+
+	"github.com/hyperledger/fabric/core/ledger"
+)
+
+type blockListener interface {
+	receiveBlock(*ledger.CommitNotification)
+	close()
+}
```
manish-sethi: Can we remove this interface and straight away use the two concrete types - to make it more explicit. The current code may have saved a few lines of code but lessen the readability.

internal/pkg/gateway/commit/blocknotifier.go
```diff
@@ -0,0 +1,76 @@
...
+
+type blockNotifier struct {
+	commitChannel <-chan *ledger.CommitNotification
```
manish-sethi, line 21: I feel that too many `notifiers` are making it hard to read code. Can we club the functionality in this into the main `Notifier`. That will make the main `Notifier` act as a central piece that glues together the lower level events received from ledger and the other two notifiers handle the event specific delivery.

manish-sethi (commented):
@bestbeforetoday - I feel that the code in this area is already becoming complex enough. Primarily because of how dependencies are factored out and a number of notifiers. I have listed a few suggestions to simplify things, see if they makes sense to you.

internal/pkg/gateway/api.go
```diff
@@ -258,11 +258,55 @@ func (gs *Server) CommitStatus(ctx context.Context, signedRequest *gp.SignedComm
...
+// within the block.
+func (gs *Server) ChaincodeEvents(signedRequest *gp.SignedChaincodeEventsRequest, stream gp.Gateway_ChaincodeEventsServer) error {
+	if signedRequest == nil {
+		return status.Error(codes.InvalidArgument, "a chaincode events request is required")
+	}
+
+	request := &gp.ChaincodeEventsRequest{}
+	if err := proto.Unmarshal(signedRequest.Request, request); err != nil {
+		return status.Error(codes.InvalidArgument, "invalid chaincode events request")
+	}
```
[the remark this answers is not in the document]
bestbeforetoday: How will an empty string or empty slice cause a nil-pointer panic? Could you give me an example of the change you are requiring here?

internal/peer/node/start.go
```diff
@@ -825,14 +825,18 @@ func serve(args []string) error {
...
 					&gateway.EndorserServerAdapter{Server: serverEndorser},
 					discoveryService,
 					&commit.Finder{
 						Query:    peerAdapter,
-						Notifier: commit.NewNotifier(peerAdapter),
+						Notifier: notifier,
+					},
+					&commit.Eventer{
+						Notifier: notifier,
 					},
```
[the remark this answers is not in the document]
bestbeforetoday: I started with the narrower interface but was required to do it this way by another maintainer. The way the code is currently structured (at the request of another maintainer), we can't pass concrete implementations in to `gateway.CreateServer` since mocks are passed in there by unit tests.

I'll look at some changes back towards my original approach for the `Finder` and `Eventer`, and I propose we look at changes to other areas of the code in a different PR rather than shoe-horn a bunch of unrelated code refactor into this one.

internal/pkg/gateway/commit/eventer.go
```diff
@@ -0,0 +1,17 @@
...
+
+import "context"
+
+type Eventer struct {
+	Notifier *Notifier
+}
+
+func (e *Eventer) ChaincodeEvents(ctx context.Context, channelName string, chaincodeName string) (<-chan *BlockChaincodeEvents, error) {
+	return e.Notifier.notifyChaincodeEvents(ctx.Done(), channelName, chaincodeName)
+}
```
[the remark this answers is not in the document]
bestbeforetoday: Correct

internal/pkg/gateway/commit/finder.go
```diff
@@ -12,13 +12,20 @@ import (
...
+	TransactionID string
+	Code          peer.TxValidationCode
+}
+
 // QueryProvider provides status of previously committed transactions on a given channel. An error is returned if the
 // transaction is not present in the ledger.
 type QueryProvider interface {
-	TransactionStatus(channelName string, transactionID string) (peer.TxValidationCode, error)
+	TransactionStatus(channelName string, transactionID string) (peer.TxValidationCode, uint64, error)
 }
```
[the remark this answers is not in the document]
bestbeforetoday: This describes an interface for a subset of capability provided by the PeerAdapter struct, which allows me to use a mock implementation in unit tests. Trying to make the PeerAdapter work purely with interfaces (that I could then mock) would be even more convoluted because of the way to Go type system works

internal/pkg/gateway/commit/blocknotifier.go
bestbeforetoday: This abstraction isn't there to save a few lines of code. It's to:
1. Keep the blockListener [focused on the task](https://en.wikipedia.org/wiki/Single-responsibility_principle) of receiving and dispatching events from the ledger
2. Allow it to be [extended with other listeners without requiring code changes](https://en.wikipedia.org/wiki/Open%E2%80%93closed_principle)
3. Avoid it needing [dependencies on listener implementations](https://en.wikipedia.org/wiki/Dependency_inversion_principle)

internal/peer/node/start.go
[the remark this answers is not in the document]
bestbeforetoday: I've made some changes to reduce the leakage of internal abstractions outside of the gateway package, which I think is along the lines of what you are asking.

Note that even within the gateway packages there are deliberately different layers of abstraction. The core Gateway code is intended to be agnostic to whether it is an embedded or standalone implementation, and so interacts with the embedded implementation-specific elements using interfaces, which could be replaced standalone implementations.

internal/pkg/gateway/api.go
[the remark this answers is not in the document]
manish-sethi: In downstream code this byte array is used to unmarshal and I am not certain that every place all the checks have been in place. So, it was just to fail early if this byte slice is nil - kind of basic input check.

internal/pkg/gateway/commit/finder.go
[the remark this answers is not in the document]
manish-sethi: Decoupling code within a package for the sake of using mocks for UTs does not sound a good idea to me. I am not sure I understand what prevents from writing UTs as long as external dependencies are mocked.

internal/pkg/gateway/commit/blocknotifier.go
manish-sethi: I was thinking in the lines of launching a gorouting for each channel directly into the `Notifier` and the notifier just keeps track of the two event specific notifiers. In the current structure of the code it sounded like all three notifiers are kind of siblings when they are actually not. The other approach could be to have the both event specific notifiers only inside this `blocknotifier` - so this becomes full controller of channel specific notifier and hierarchy is clearer.

internal/peer/node/start.go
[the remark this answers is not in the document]
manish-sethi: We can definitely do it in a separate PR. But I feel that there are way to many layers of abstraction for a simple enough functionality.

internal/pkg/gateway/commit/blocknotifier.go
manish-sethi: These objectives sounds good if they are between course level modules. Applying these to micro level of granularity has led to Fabric code at many places as hard to read IMO.

internal/peer/node/start.go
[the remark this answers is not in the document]
manish-sethi: > we can't pass concrete implementations in to gateway.CreateServer

contrary to this you were actually passing the peer instance to `commit package`. What my suggestion was that only external dependency of gateway package is to get an object on which it can call "RegisterForNotification" for a given channel and that should be easy enough to reflect in first-order via an interface.

internal/pkg/gateway/api.go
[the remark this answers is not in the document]
bestbeforetoday: You can unmarshall a zero-length `[]byte` fine. I don't understand the benefit of adding code bloat for a special check for a zero-length slice rather than letting a meaningful error be raised, such as the requested channel doesn't exist or the supplied identity is not valid. The client APIs don't send empty protobufs so it's not a scenario that can occur by accident

internal/peer/node/start.go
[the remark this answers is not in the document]
bestbeforetoday: > What my suggestion was that only external dependency of gateway package is to get an object on which it can call "RegisterForNotification" for a given channel and that should be easy enough to reflect in first-order via an interface.

No.

`RegisterForNotification()` implies the peer embedded implementation, which is in the commit package. The commit package uses the `NotificationSupplier` interface to define `CommitNotifications()` which is the same as `RegisterForNotification()`, so it's already doing as you suggest.

The gateway server is not intended to be tied to either a peer embedded or standalone implementation, so it uses a `CommitFinder` interface to obtain transaction validation codes by some means that it doesn't need to know about, and which would be significantly different in the embedded and standalone implementations.

internal/pkg/gateway/commit/finder.go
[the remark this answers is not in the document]
bestbeforetoday: QueryProvider decouples the PeerAdapter (i.e. the thin layer of code not easily untangled from the peer implementation, and so you could think of as being the external dependency) from the implementation in this package

internal/pkg/gateway/commit/blocknotifier.go
bestbeforetoday: Feel free to make that change

internal/pkg/gateway/commit/blocknotifier.go
bestbeforetoday: Feel free to make that change

bestbeforetoday:
@manish-sethi Your outstanding comments seem to be primarily objections to code structure that was already there prior to this PR, and was there at the request of other maintainers. I have addressed as many of these structural issues as I think I reasonably can in this PR, which is delivering chaincode event listening, not a refactor of existing code

integration/gateway/gateway_test.go
```diff
@@ -129,175 +187,104 @@ var _ = Describe("GatewayService", func() {
...
+			grpcErr, _ := status.FromError(err)
+			Expect(grpcErr.Code()).To(Equal(codes.PermissionDenied))
 		})
+	})
 
-		It("should fail on unauthorized identity", func() {
-			badSigningIdentity := network.OrdererUserSigner(network.Orderer("orderer"), "Admin")
-			badIdentity, err := badSigningIdentity.Serialize()
+	Describe("ChaincodeEvents", func() {
+		It("should respond with emitted chaincode events", func() {
```
manish-sethi, line 248: I guess that you still plan to add security related tests, like the one in `CommitStatus` api ? However, that can always be added later.

manish-sethi (approved):

manish-sethi:
> @manish-sethi Your outstanding comments seem to be primarily objections to code structure that was already there prior to this PR, and was there at the request of other maintainers. I have addressed as many of these structural issues as I think I reasonably can in this PR, which is delivering chaincode event listening, not a refactor of existing code

Yes, As I mentioned in the previous comment, the refactoring can be taken up in a separate PR

bestbeforetoday:
@manish You were not keen on having the `blockNotifier` in addition to the other notifiers. This is really there because you imposed the limitation that there could only be one listener to ledger notifications, so the `blockNotifier` provides a mechanism to workaround that limitation and allow multiple listeners. One approach to simplifying this code might be to remove your restriction on only allowing a single listener for ledger notifications, in which case the `blockNotifier` and the logic there could be completely removed.

PR #2606, opened 2021-05-25
Optionally disable gossip block forwarding

yacovm:
This commit adds a new configuration option to the peer which makes peers
not forward blocks that they pull from the ordering service.

If all peers in an organization explictly set "peer.deliveryclient.blockGossipEnabled" to false, 
no peer in the organization gossips blocks to any other peer in that organization.

Change-Id: I5d9b278ae72f239129827c044fa78179f6ba87ab


sampleconfig/core.yaml
```diff
@@ -348,6 +348,9 @@ peer:
 
     # Delivery service related config
     deliveryclient:
+        # Prevents this peer from further disseminating blocks it pulled from the ordering service
+        # via gossip. Note that 'gossip.state.enabled' controls point to point block replication.
+        blockGossipDisabled: false
```
denyeart: The config precedent is to use 'enabled' instead of 'disabled', so ideally it would be `blockGossipEnabled` and default to true for backwards compatibility.
I assume you made it `blockGossipDisabled` to handle the scenario where it is not set and we get the zero value "false" from viper. But I think this can be handled by calling viper's IsSet() and if not set make enabled=true for backwards compatibility. That would make the config more rational for end user, wdyt?

C0rWin (dismissed):

denyeart (dismissed):

C0rWin (approved):

denyeart:

denyeart:

PR #2614, opened 2021-05-28
Add block-number along with validation code for a given TXID

manish-sethi:

#### Type of change

#### Description
This allows gateway to pass on additional relevant info to client and for using it for logging that could be helpful in debugging

### DIFF
--- a/common/ledger/blkstorage/blockstore.go
+++ b/common/ledger/blkstorage/blockstore.go
@@ -93,7 +93,7 @@ func (store *BlockStore) RetrieveBlockByTxID(txID string) (*common.Block, error)
 }
 
 // RetrieveTxValidationCodeByTxID returns the validation code for the specified txID
-func (store *BlockStore) RetrieveTxValidationCodeByTxID(txID string) (peer.TxValidationCode, error) {
+func (store *BlockStore) RetrieveTxValidationCodeByTxID(txID string) (peer.TxValidationCode, uint64, error) {
 	return store.fileMgr.retrieveTxValidationCodeByTxID(txID)
 }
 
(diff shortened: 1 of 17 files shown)


bestbeforetoday (approved):

common/ledger/blkstorage/blockstore.go
```diff
@@ -93,7 +93,7 @@ func (store *BlockStore) RetrieveBlockByTxID(txID string) (*common.Block, error)
 }
 
 // RetrieveTxValidationCodeByTxID returns the validation code for the specified txID
-func (store *BlockStore) RetrieveTxValidationCodeByTxID(txID string) (peer.TxValidationCode, error) {
+func (store *BlockStore) RetrieveTxValidationCodeByTxID(txID string) (peer.TxValidationCode, uint64, error) {
```
mastersingh24, line 96: This is a change to a public method, but given Fabric is not designed to be a library, I don't think we need to document the API change.

core/committer/txvalidator/v14/validator_test.go
```diff
@@ -1450,9 +1450,9 @@ func (m *mockLedger) GetBlockByTxID(txID string) (*common.Block, error) {
 }
 
 // GetTxValidationCodeByTxID returns validation code of give tx
-func (m *mockLedger) GetTxValidationCodeByTxID(txID string) (peer.TxValidationCode, error) {
+func (m *mockLedger) GetTxValidationCodeByTxID(txID string) (peer.TxValidationCode, uint64, error) {
```
mastersingh24, line 1453: This is a change to a public method, but given Fabric is not designed to be a library, I don't think we need to document the API change.

mastersingh24 (approved):
One comment about a single-use function, but other than that change seems fine to me

common/ledger/blkstorage/blockstore.go
manish-sethi: Good Catch! I'll fix this in a separate PR to avoid delay merging in this.

PR #2624, opened 2021-05-30
[FAB-18479] Log error if orderer can't forward SubmitRequest

denyeart:
[FAB-18479] Log error if orderer can't forward SubmitRequest to Raft leader

If order cannot forward client transaction SubmitRequest to Raft leader,
it only logged it in a debug message. And since the error does
not get returned to the client either, it is difficult to identify
and troubleshoot.

This commit changes the message to an error in the orderer log.


C0rWin (approved):

denyeart:

denyeart:

PR #2573, opened 2021-04-29
Feature/fix trouble with pkcs11

afrancoc2000:
#### Type of change

- Bug fix

#### Description

Simple changes that fix these errors on 1.4 Path:
- When setting ORDERER_GENERAL_BCCSP_PKCS11_SECURITY orderer variable the orderer fails saying that the security value is a string and not an int.
- When running `GO_TAGS=pkcs11 make docker` the fabric-tools docker image doesn't include a PKCS11 provider.

#### Related issues

https://jira.hyperledger.org/browse/FAB-18457

PR #2631, opened 2021-06-02
[FAB-17900] Fixes numeric env variable override bug

afrancoc2000:
#### Type of change
- Bug fix

#### Description

Backport fix for version 1.4 for error: when setting ORDERER_GENERAL_BCCSP_PKCS11_SECURITY orderer variable, the orderer fails saying that the security value is a string and not an int.

#### Related issues

[FAB-17900](https://jira.hyperledger.org/browse/FAB-17900)
[FAB-18457](https://jira.hyperledger.org/browse/FAB-18457)
https://github.com/hyperledger/fabric/pull/2573

### DIFF
--- a/common/viperutil/config_util.go
+++ b/common/viperutil/config_util.go
@@ -313,7 +313,7 @@ func bccspHook(f reflect.Type, t reflect.Type, data interface{}) (interface{}, e
 
 	config := factory.GetDefaultOpts()
 
-	err := mapstructure.Decode(data, config)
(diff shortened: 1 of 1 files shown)


denyeart (approved):
Thanks, while we don't expect to release additional v1.4.x releases since it is no longer LTS, we could still merge if this helps you. I'll approve, but let's see if anybody else has comments.

mastersingh24 (approved):
This does not hurt anything but we aren't going to be producing new images or releases.

PR #2633, opened 2021-06-02
Clarify Verify behaviour in PKCS11 Impl

davidkel:
Signed-off-by: D <d_kelsey@uk.ibm.com>

#### Type of change

- Test update

#### Description

the pkcs11 verify implementation will never receive a pkcs11 key to be used to verify a signature, this change clarifies that in the code and adds tests for the verify implementation

### DIFF
--- a/bccsp/pkcs11/pkcs11.go
+++ b/bccsp/pkcs11/pkcs11.go
@@ -292,12 +292,11 @@ func (csp *Provider) Verify(k bccsp.Key, signature, digest []byte, opts bccsp.Si
 		return false, errors.New("Invalid digest. Cannot be empty")
 	}
 
-	// Check key type
-	switch key := k.(type) {
-	case *ecdsaPrivateKey:
-		return csp.verifyECDSA(key.pub, signature, digest)
-	case *ecdsaPublicKey:
-		return csp.verifyECDSA(*key, signature, digest)
+	// key (k) will never be a pkcs11 key, do verify using the software implementation
+	// but validate it just in case
+	switch k.(type) {
+	case *ecdsaPrivateKey, *ecdsaPublicKey:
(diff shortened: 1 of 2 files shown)


mastersingh24 (changes requested):
Hmm ... not looking for changes, but I'm not sure this assertion is true.  There is supposed to be a switch for SW verify versus verifying signatures via the PKCS11 provider.

mastersingh24:
@davidkel  What's the genesis of this change?  We are supposed to be able to verify via the PKCS11 provider not always in SW.

sykesm:
> @davidkel What's the genesis of this change? We are supposed to be able to verify via the PKCS11 provider not always in SW.

So, in Fabric, we never get a pkcs11 key because of the MSP implementations. I walked the path multiple times to verify that because I could not believe it at first. This is not to say that _outside_ of Fabric this never happens, but, given the CA is using an old version of this with RSA support, this change will not impact that code.

sykesm:
> So, in Fabric, we never get a pkcs11 key because of the MSP implementations. I walked the path multiple times to verify that because I could not believe it at first. This is not to say that _outside_ of Fabric this never happens, but, given the CA is using an old version of this with RSA support, this change will not impact that code.

I'll actually go one step further. If we ever did use PKCS#11 for verification and we had any hiccup at all with a CloudHSM provider, we would have all sorts of issues in validation. That's how I started walking down the thread in the first place.

denyeart:
@mastersingh24 Based on Matt's assessment, this was a change he suggested so that the implicit assumptions in the code would become explicit, reducing any confusion in this space. As such, are you ok dismissing your comment?

mastersingh24 (approved):

mastersingh24:
LGTM based on the comments.  @sykesm ?

PR #2664, opened 2021-06-10
Update discovery PeersOfChannel

jt-nti:
Update PeersOfChannel behaviour to match Peers function,
i.e. include endpoint in self memeber, only the peers that
have an external endpoint, and sanitizes the envelopes

### DIFF
--- a/discovery/support/gossip/support.go
+++ b/discovery/support/gossip/support.go
@@ -54,12 +54,15 @@ func (s *DiscoverySupport) PeersOfChannel(chain common.ChannelID) discovery.Memb
 		return nil
 	}
 	stateInf := msg.GetStateInfo()
-	selfMember := discovery.NetworkMember{
-		Properties: stateInf.Properties,
-		PKIid:      stateInf.PkiId,
-		Envelope:   msg.Envelope,
-	}
-	return append(s.Gossip.PeersOfChannel(chain), selfMember)
+
+	selfMember := s.Gossip.SelfMembershipInfo()
+	selfMember.Properties = stateInf.Properties
+	selfMember.Envelope = msg.Envelope
+
+	peers := s.Gossip.PeersOfChannel(chain)
+	peers = append(peers, selfMember)
+	// Return only the peers that have an external endpoint, and sanitizes the envelopes.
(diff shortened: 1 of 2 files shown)


yacovm:
I think that we still filter out peers without an external endpoint, regardless of this function. 
It is sufficient to filter them out in the [alive memebership](https://github.com/hyperledger/fabric/blob/main/discovery/support/gossip/support.go#L70) function because we always intersect all peers with the alive membership view.

yacovm:
> > I think that we still filter out peers without an external endpoint, regardless of this function.
> > It is sufficient to filter them out in the [alive memebership](https://github.com/hyperledger/fabric/blob/main/discovery/support/gossip/support.go#L70) function because we always intersect all peers with the alive membership view.
> 
> The immediate problem is that the external endpoint is not being included for the self member, [which means the new gateway code does not find the local endorser](https://github.com/hyperledger/fabric/blob/b4efe857ffad1e84877650f192e6ac27d143716f/internal/pkg/gateway/registry.go#L151).
> 
> Maybe that isn't the best way to identify the local endorser- are there situations that there wouldn't be an external endpoint rather than it just not being returned by discovery? (Even so, it doesn't seem that unreasonable for Peers and PeersOfChannel to behave consistently, or at least not to drop the external endpoint if there is one.)

Not all peers have external endpoints. Only those which are configured to have such. 
In Fabric, peers are identified by their PKI-IDs which is a hash over the serialized identity.

You need to know your PKI-ID which can be computed from the [serialized identity](https://github.com/hyperledger/fabric/blob/main/internal/peer/node/start.go#L307).

yacovm (changes requested):
I don't see why we need to change the implementation of this method

yacovm (changes requested):
The purpose of the package you are changing, is to be used as an implementation of an interface (a dependency) that discovery consumes.
You are not supposed to change it in order to fit the implementation of the gateway. As I said in the PR comments, if you need some information that the method doesn't return, you are free to get it from other places in the peer which I pointed to. 

Moreover, your commit message may explain what you're changing but it doesn't explain intent - why this change is required.

PR #2666, opened 2021-06-11
[FAB-18484] Return transaction forwarding result back to the client synchronously

yacovm:
This commit makes a Raft follower wait for the transaction forwarded to the leader to be sent into
the gRPC stream, and returns the result (success or failure) back to the client accordingly.

Before this commmit, the behavior was that it returns success after enqueueing it into the message queue,
which might have resulted in the transaction being dropped but a success being returned to the client.

Change-Id: I0cd45540be4988845663eb0c68f76fed2ff25b94

### DIFF
--- a/orderer/common/cluster/comm.go
+++ b/orderer/common/cluster/comm.go
@@ -461,8 +461,11 @@ type RemoteContext struct {
 
 // Stream is used to send/receive messages to/from the remote cluster member.
 type Stream struct {
-	abortChan    <-chan struct{}
-	sendBuff     chan *orderer.StepRequest
+	abortChan <-chan struct{}
+	sendBuff  chan struct {
+		request *orderer.StepRequest
+		report  func(error)
+	}
 	commShutdown chan struct{}
 	abortReason  *atomic.Value
 	metrics      *Metrics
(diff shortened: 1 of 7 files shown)


denyeart:
LGTM. I've also verified in my local environment, client receives the submission error.

orderer/common/cluster/comm.go
```diff
@@ -615,11 +624,13 @@ func (stream *Stream) operateWithTimeout(invoke StreamOperation) (*orderer.StepR
 
 	select {
 	case r := <-responseChan:
+		report(r.err)
```
C0rWin, line 627: shouldn't you report only if there is an error?

yacovm: Well if err is nil, then it will report a nil. 

Keep in mind that the actual consumer of this mechanism, is the consensus (etcdraft) layer which actually waits for the result of this function, be it error or nil:


```
report := func(err error) {
		if err != nil {
			atomicErr.Store(err.Error())
			c.Metrics.ProposalFailures.Add(1)
		}
		close(sentChan) // <---- here we close the channel to signal to the goroutine which forward to the leader that we have a result (success or error)
	}

c.rpc.SendSubmit(lead, req, report)

select {
case <-sentChan: // <---- here we wait for the result
case <-c.doneC:
	return errors.Errorf("chain is stopped")
case <-timer.C:
	return errors.Errorf("timed out (%v) waiting on forwarding to %d", c.opts.RPCTimeout, lead)
}
```

C0rWin (approved):

yacovm:

yacovm:

sergefdrv:
> the transaction being dropped but a success being returned to the client

@yacovm could you please expand a bit on how this could happen?

sergefdrv:
@yacovm never mind, I found the details in [FAB-18484](https://jira.hyperledger.org/browse/FAB-18484)

PR #2689, opened 2021-06-17
Clarify orderers seeing the transaction data

arsulegai:
#### Type of change

- Documentation update

#### Description

Having the answer start with a `No` is misleading. Beginners
who generally rely on FAQ section to understand the fundamentals
thought that orderers do not receive any information about a
transaction.

With the explanation on what is available with orderers clarifies
the intent served in this FAQ. i.e. Orderers do get the
transaction but at their goodwill they will not open the
transaction, as they do not have a need to open them.

#### Additional details

NA, documentation update

#### Related issues

None

### DIFF
--- a/docs/source/Fabric-FAQ.rst
+++ b/docs/source/Fabric-FAQ.rst
@@ -56,7 +56,12 @@ Security & Access Control
   Do the orderers see the transaction data?
 
 :Answer:
-  No, the orderers only order transactions, they do not open the transactions.
+  Orderers receive endorsed transactions that are submitted from application
+  clients. The endorsed payload contains the chaincode execution results
+  including the ReadSet and WriteSet information. The orderers only validate
+  the submitter's identity and order transactions, they do not open the
+  endorsed transactions.
+
   If you do not want the data to go through the orderers at all, then utilize
   the private data feature of Fabric.  Alternatively, you can hash or encrypt
   the data in the client application before calling chaincode. If you encrypt


docs/source/Fabric-FAQ.rst
```diff
@@ -56,7 +56,10 @@ Security & Access Control
   Do the orderers see the transaction data?
 
 :Answer:
-  No, the orderers only order transactions, they do not open the transactions.
+  Orderers receive endorsements from the client. The endorsement payload has
```
C0rWin: Well, if you would like to expand with more details I think it has to be more accurate and reflect the actual flow, for example:

Orderers receive a transaction, which includes endorsements from the peers collected by the client, moreover endorsement payload contains not only RWSet, but also signatures of the endorsing peers over the hashed value of the simulation results including endorsement proposal. 

That being said, I do not see real issue with previous answer which is very concrete and precisely answering the question.

arsulegai: I found this to be misleading because the community misinterpreted "do orderers read?" to be "do orderers have access?". This misinterpretation could lead to designs where privacy is assumed. Wordings here carry more weightage because this question is under FAQ for data privacy. There can either be another question on the flow or this question can be clarified with additional details.

Let me know your thoughts.

arsulegai: Thanks Dave for the feedback. I take it as a yes to providing clarification on this answer itself. Edited the wordings as per suggestion.

C0rWin: > I found this to be misleading because the community misinterpreted "do orderers read?" to be "do orderers have access?". This misinterpretation could lead to designs where privacy is assumed. Wordings here carry more weightage because this question is under FAQ for data privacy. There can either be another question on the flow or this question can be clarified with additional details.
> 
> Let me know your thoughts.

I do agree clarification is needed, just thought that it could be a bit more accurate, and Dave's suggestion is actually a good one.

denyeart (approved):

nikhil550:
@mergifyio backport release-2.3

nikhil550:

PR #2690, opened 2021-06-17
Handle missing endpoints from discovery

jt-nti:
For some reason the discovery PeersOfChannel function does not
return an endpoint for the local peer, which the gateway was
relying on

Rather than updating PeersOfChannel to return the endpoint when
there is one, to match the Peers function, this change updates
how the gateway identifies the local peer

### DIFF
--- a/integration/gateway/endorsing_orgs_test.go
+++ b/integration/gateway/endorsing_orgs_test.go
@@ -1,0 +1,151 @@
+/*
+Copyright IBM Corp All Rights Reserved.
+
+SPDX-License-Identifier: Apache-2.0
+*/
+
+package gateway
+
+import (
+	"context"
+	"io/ioutil"
+	"os"
+	"path/filepath"
+	"syscall"
+
+	docker "github.com/fsouza/go-dockerclient"
+	"github.com/golang/protobuf/proto"
+	"github.com/hyperledger/fabric-protos-go/gateway"
+	"github.com/hyperledger/fabric-protos-go/peer"
+	"github.com/hyperledger/fabric/integration/nwo"
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+	"github.com/tedsuo/ifrit"
+)
+
+var _ = Describe("GatewayService with endorsing orgs", func() {
+	var (
+		testDir   string
+		network   *nwo.Network
+		orderer   *nwo.Orderer
+		org1Peer0 *nwo.Peer
+		org2Peer0 *nwo.Peer
+		org3Peer0 *nwo.Peer
+		process   ifrit.Process
+	)
+
+	BeforeEach(func() {
+		var err error
+		testDir, err = ioutil.TempDir("", "gateway")
+		Expect(err).NotTo(HaveOccurred())
+
+		client, err := docker.NewClientFromEnv()
+		Expect(err).NotTo(HaveOccurred())
+
+		config := nwo.ThreeOrgRaft()
+		network = nwo.New(config, testDir, client, StartPort(), components)
+
+		network.GatewayEnabled = true
+
+		network.GenerateConfigTree()
+		network.Bootstrap()
+
+		networkRunner := network.NetworkGroupRunner()
+		process = ifrit.Invoke(networkRunner)
+		Eventually(process.Ready(), network.EventuallyTimeout).Should(BeClosed())
+
+		orderer = network.Orderer("orderer")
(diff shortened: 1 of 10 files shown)


internal/pkg/gateway/registry.go
```diff
@@ -165,12 +166,22 @@ func (reg *registry) endorsersByOrg(channel string, chaincode string) map[string
 	defer reg.configLock.RUnlock()
 
 	for _, member := range members {
+		pkiid := member.PKIid
+		// TODO use member.PreferredEndpoint() or go back to member.Endpoint?
```
andrew-coleman: I think we should stick to member.PreferredEndpoint()

internal/pkg/gateway/registry.go
```diff
@@ -165,12 +166,22 @@ func (reg *registry) endorsersByOrg(channel string, chaincode string) map[string
...
 		// find the endorser in the registry for this endpoint
 		var endorser *endorser
-		if endpoint == reg.localEndorser.address {
+		if bytes.Equal(pkiid, reg.localEndorser.pkiid) {
+			logger.Debugw("Found local endorser", "pkiid", pkiid)
 			endorser = reg.localEndorser
+		} else if endpoint == "" {
+			// TODO the endpoint may be empty because discovery!!
+			// include seperate check/warning or just fall through to else???
+			reg.logger.Warnf("Failed to find endorser: no endpoint from discovery")
```
andrew-coleman: Keep the warning in.  We'd need to understand why we had an endorser with no endpoint if it's not the local peer.  Might be worth logging the pkiid of this endorser.

internal/pkg/gateway/registry.go
```diff
@@ -312,18 +323,19 @@ func (reg *registry) registerChannel(channel string) error {
 	peers := map[string]string{}
 	members := reg.discovery.PeersOfChannel(common.ChannelID(channel))
 	for _, member := range members {
+		// TODO why is this fragile? Make it less fragile?
```
andrew-coleman: Matt was never happy about this - but this code was copied from the discovery service

andrew-coleman (commented):
I'm wondering if we should also add a unit test that demonstrates how this fixes the empty endpoint problem.

internal/pkg/gateway/registry.go
jt-nti: I'll leave it as is for now then

andrew-coleman (dismissed):

denyeart:
@jt-nti Please add PR and commit message explaining the motivation. Is this a workaround for something missing from discovery?

integration/gateway/endorsing_orgs_test.go
```diff
@@ -0,0 +1,151 @@
...
+			for _, o := range peers {
+				mspid := network.Organization(o.Organization).MSPID
+
+				submitCheckEndorsingOrgTransaction(ctx, gatewayClient, signingIdentity, mspid)
+			}
+		}
+	})
+})
+
+func submitCheckEndorsingOrgTransaction(ctx context.Context, client gateway.GatewayClient, signingIdentity *nwo.SigningIdentity, mspids ...string) {
```
denyeart, line 124: The function name is singular (Org) while the input allows for passing multiple orgs (mspids). This causes some confusion.

integration/gateway/endorsing_orgs_test.go
```diff
@@ -0,0 +1,151 @@
...
+
+		for _, p := range peers {
+			conn := network.PeerClientConn(p)
+			defer conn.Close()
+			gatewayClient := gateway.NewGatewayClient(conn)
+			ctx, cancel := context.WithTimeout(context.Background(), network.EventuallyTimeout)
+			defer cancel()
+
+			signingIdentity := network.PeerUserSigner(p, "User1")
+			for _, o := range peers {
```
denyeart, line 115: It took me a while to figure out why you have an outer and an inner peers loop. I finally concluded that you wanted to test every peer's gateway with every peer as the endorsing org.
Would be good to indicate this intent in a comment prior to the loop so that people know what to expect.

denyeart (approved):
The majority of the PR relates to testing the specification of endorsing org, while the PR and commit message talk only of the local peer workaround.  This would have been better as two distinct PRs, or at least two separate commits in a single PR.

I won't hold it up with my minor comments however.

jt-nti:
@denyeart it was the new tests which uncovered the bug :)

PR #2708, opened 2021-06-23
Compare Snapshots Utility First 10

jcastrence:
#### Type of change

- New feature
- Test update

#### Description

- Snapshot comparison tool now only finds the first 10 divergences by default. A flag has been added to let the user allow for all divergences to be found.

#### Related issues

- [FAB-18425](https://jira.hyperledger.org/browse/FAB-18425)

### DIFF
--- a/cmd/ledger/main.go
+++ b/cmd/ledger/main.go
@@ -15,7 +15,8 @@ import (
 )
 
 const (
-	resultFilename = "./result.json"
+	resultFilename      = "./result.json"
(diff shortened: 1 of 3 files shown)


manish-sethi:
I don't think that this flag adds any value. We can close this PR. Instead what is desired is to produce in a separate file, top 10 entries from the list of divergent records sorted by blkNum:TxNum. So, it could be used as a point from where divergence started.

PR #2722, opened 2021-06-29
Add function to delete the ledger data for a channel

manish-sethi:

#### Type of change
- New feature

#### Description
This PR adds a function that takes the ledger config and deletes the data for a channel

#### Additional details
This is a basic function that would be invoked in the future PR and the tests will be added in that PR

#### Related issues
[FAB-11334](https://jira.hyperledger.org/browse/FAB-11334)

### DIFF
--- a/core/ledger/kvledger/unjoin_channel.go
+++ b/core/ledger/kvledger/unjoin_channel.go
@@ -1,0 +1,135 @@
+/*
+Copyright IBM Corp. All Rights Reserved.
+
+SPDX-License-Identifier: Apache-2.0
+*/
+
+package kvledger
+
+import (
+	"github.com/hyperledger/fabric-lib-go/healthz"
+	"github.com/hyperledger/fabric-protos-go/ledger/rwset/kvrwset"
+	"github.com/hyperledger/fabric-protos-go/peer"
+	"github.com/hyperledger/fabric/common/ledger/blkstorage"
+	"github.com/hyperledger/fabric/common/metrics/disabled"
+	"github.com/hyperledger/fabric/core/ledger"
+	"github.com/hyperledger/fabric/core/ledger/confighistory"
+	"github.com/hyperledger/fabric/core/ledger/kvledger/bookkeeping"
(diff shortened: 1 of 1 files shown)


core/ledger/kvledger/unjoin_channel.go
```diff
@@ -0,0 +1,135 @@
...
+	"github.com/hyperledger/fabric/core/ledger"
+	"github.com/hyperledger/fabric/core/ledger/confighistory"
+	"github.com/hyperledger/fabric/core/ledger/kvledger/bookkeeping"
+	"github.com/hyperledger/fabric/core/ledger/kvledger/history"
+	"github.com/hyperledger/fabric/core/ledger/kvledger/txmgmt/privacyenabledstate"
+	"github.com/hyperledger/fabric/core/ledger/pvtdatastorage"
+)
+
+// removeLedgerData removes the data for a given ledger. This function should be invoked when the peer is not running and the caller should hold the file lock for the KVLedgerProvider
+func removeLedgerData(config *ledger.Config, ledgerID string) error {
```
cendhu, line 24: As this function creates all providers and removes the data associated with a given `ledgerID`, shouldn't it take care of deleting the `ledgerID` from the `idStore` too?

If not, are we expecting the caller of this function to open the `idStore` and delete the `ledgerID` once this function returns success?

cendhu (commented):
LGTM. Just a doubt.

core/ledger/kvledger/unjoin_channel.go
cendhu: Got the answer by looking at the code of other commands such as `pause`, `resume`. So the caller should take care of acquiring the lock as well as deleting the `ledgerID` from the `idStore` https://github.com/hyperledger/fabric/blob/7151302be2dff77a9d78cdc8c7a04c4fee352914/core/ledger/kvledger/pause_resume.go#L33-L47 as done here

cendhu (approved):

PR #2724, opened 2021-06-29
[FAB-18508] ledger utility always outputs txNum

shimos:
#### Type of change

- Bug fix

#### Description

This patch fixes a bug reported in FAB-18508, and makes the ledger troubleshooting utility output always the record "txNum".

#### Additional details

The bug occurred when the target transaction is the first transaction of a block (txNum == 0). This patch removes the `omitempty` attributes from all the members of a record.

Note that this patch also changes the output when a key is missing in either snapshot. Instead of an empty object ({}) for the missing snapshot, the entire key (either "snapshotrecord1" or "snapshotrecord2") will be omitted.
(Refer to the changes in the test code)

#### Related issues

https://jira.hyperledger.org/projects/FAB/issues/FAB-18508

### DIFF
--- a/internal/ledger/compare.go
+++ b/internal/ledger/compare.go
@@ -188,21 +188,14 @@ func Compare(snapshotDir1 string, snapshotDir2 string, outputFile string) (count
 type diffRecord struct {
 	Namespace string          `json:"namespace,omitempty"`
 	Key       string          `json:"key,omitempty"`
-	Record1   *snapshotRecord `json:"snapshotrecord1,omitempty"`
-	Record2   *snapshotRecord `json:"snapshotrecord2,omitempty"`
+	Record1   *snapshotRecord `json:"snapshotrecord1"`
+	Record2   *snapshotRecord `json:"snapshotrecord2"`
 }
 
 // Creates a new diffRecord
 func newDiffRecord(namespace string, record1 *privacyenabledstate.SnapshotRecord,
 	record2 *privacyenabledstate.SnapshotRecord) (*diffRecord, error) {
-	// Empty snapshot
-	s0 := snapshotRecord{
-		Value:    "",
-		BlockNum: 0,
-		TxNum:    0,
-	}
-
-	var s1, s2 *snapshotRecord = &s0, &s0 // snapshot records
+	var s1, s2 *snapshotRecord = nil, nil // snapshot records
 	var k string                          // key
 	var err error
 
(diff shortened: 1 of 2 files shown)


internal/ledger/compare_test.go
```diff
@@ -498,13 +484,11 @@ func TestJSONArrayFileWriter(t *testing.T) {
 				"value" : "green",
 				"blockNum" : 472,
 				"txNum" : 61
-			},
-			"snapshotrecord2" : {}
```
manish-sethi, line 502: This was an intentional entry that was chosen to represent a missing record. Unless there is a strong reason, I would prefer to keep it as is.

manish-sethi (commented):
Thanks @shimos for your contribution. Left you a comment.

shimos:
@manish-sethi Thank you for the review. I have revised the patch to keep the JSON structure.

internal/ledger/compare.go
```diff
@@ -186,24 +186,20 @@ func Compare(snapshotDir1 string, snapshotDir2 string, outputFile string) (count
...
 // diffRecord represents a diverging record in json
 type diffRecord struct {
-	Namespace string          `json:"namespace,omitempty"`
-	Key       string          `json:"key,omitempty"`
-	Record1   *snapshotRecord `json:"snapshotrecord1,omitempty"`
-	Record2   *snapshotRecord `json:"snapshotrecord2,omitempty"`
+	Namespace string      `json:"namespace,omitempty"`
+	Key       string      `json:"key,omitempty"`
+	Record1   interface{} `json:"snapshotrecord1,omitempty"`
+	Record2   interface{} `json:"snapshotrecord2,omitempty"`
```
manish-sethi, line 192: It was better to use the concrete type before. On a second thought, as an alternate we could remove "omitempty" from these two lines. That will force to print a nil object as `null` in the final JSON -- which I feel would be even clearer than `{}`. 

Sorry for an additional round of change on this.

manish-sethi: It was better to use the concrete type, as before. On a second thought, as an alternate we could remove "omitempty" in the json annotation for `Record1` and `Record2`. That will force to print a nil object as `null` in the final JSON -- which I feel would be even clearer than `{}`.

Sorry for an additional round of change on this.

shimos:
@manish-sethi ok, so I have reverted to the first patch, and deleted 'omitempty' from `snapshotrecord1` and `snapshotrecord2` so that it will produce `null` for the missing key rather than deleting the key itself.

manish-sethi:
Thanks @shimos for your contribution!

manish-sethi (approved):

PR #2726, opened 2021-06-30
[FAB-18509] Stop panic if collection index path is wrong

mbwhite:

#### Type of change
- Bug fix

### DIFF
--- a/core/ledger/kvledger/txmgmt/privacyenabledstate/db.go
+++ b/core/ledger/kvledger/txmgmt/privacyenabledstate/db.go
@@ -418,27 +418,30 @@ type indexInfo struct {
 
 const (
 	// Example for chaincode indexes:
-	// "META-INF/statedb/couchdb/indexes/indexColorSortName.json"
+	// "META-INF/statedb/couchdb/indexes"
 	chaincodeIndexDirDepth = 3
+
 	// Example for collection scoped indexes:
-	// "META-INF/statedb/couchdb/collections/collectionMarbles/indexes/indexCollMarbles.json"
+	// "META-INF/statedb/couchdb/collections/collectionMarbles/indexes"
 	collectionDirDepth      = 3
 	collectionNameDepth     = 4
 	collectionIndexDirDepth = 5
 )
 
+// Note previous functions will have ensured that the path starts
+// with 'META-INF/statedb' and does not have leading or trailing
+// path deliminators.
 func getIndexInfo(indexPath string) *indexInfo {
 	indexInfo := &indexInfo{}
-	dirsDepth := strings.Split(indexPath, "/")
+	pathParts := strings.Split(indexPath, "/")
+	pathDepth := len(pathParts)
+
 	switch {
-	case len(dirsDepth) > chaincodeIndexDirDepth &&
-		dirsDepth[chaincodeIndexDirDepth] == "indexes":
+	case pathDepth > chaincodeIndexDirDepth && pathParts[chaincodeIndexDirDepth] == "indexes":
 		indexInfo.hasIndexForChaincode = true
-	case len(dirsDepth) > collectionDirDepth &&
-		dirsDepth[collectionDirDepth] == "collections" &&
-		dirsDepth[collectionIndexDirDepth] == "indexes":
+	case pathDepth > collectionIndexDirDepth && pathParts[collectionDirDepth] == "collections" && pathParts[collectionIndexDirDepth] == "indexes":
 		indexInfo.hasIndexForCollection = true
-		indexInfo.collectionName = dirsDepth[collectionNameDepth]
+		indexInfo.collectionName = pathParts[collectionNameDepth]
 	}
 	return indexInfo
 }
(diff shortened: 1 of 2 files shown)


core/ledger/kvledger/txmgmt/privacyenabledstate/db.go
```diff
@@ -420,21 +420,26 @@ const (
 	// Example for chaincode indexes:
 	// "META-INF/statedb/couchdb/indexes/indexColorSortName.json"
 	chaincodeIndexDirDepth = 3
+	pathLengthForCCindexes = 5
```
manish-sethi:
```suggestion
	pathLengthForCCindexes = 4
```
Path that comes here is of [directory only](https://github.com/hyperledger/fabric/blob/main/core/ledger/kvledger/txmgmt/privacyenabledstate/db.go#L323). The file name is not part of this.

core/ledger/kvledger/txmgmt/privacyenabledstate/db.go
```diff
@@ -420,21 +420,26 @@ const (
...
+
 	// Example for collection scoped indexes:
 	// "META-INF/statedb/couchdb/collections/collectionMarbles/indexes/indexCollMarbles.json"
-	collectionDirDepth      = 3
-	collectionNameDepth     = 4
-	collectionIndexDirDepth = 5
+	collectionDirDepth             = 3
+	collectionNameDepth            = 4
+	collectionIndexDirDepth        = 5
+	pathLengthForCollectionindexes = 7
```
manish-sethi:
```suggestion
	pathLengthForCollectionindexes = 6
```
same as above

manish-sethi (changes requested):

core/ledger/kvledger/txmgmt/privacyenabledstate/db.go
mbwhite: thanks for the review Manish - this is the output I got from the unit test - with printing the args to the `getIndexInfo` function. Includes the index file name
```
=== RUN   TestGetIndexInfo
[getIndexInfo] indexPath=META-INF/statedb/couchdb/indexes/indexColorSortName.json
[getIndexInfo] indexPath=META-INF/statedb/couchdb/collections/collectionMarbles/indexes/indexCollMarbles.json
[getIndexInfo] indexPath=META-INF/statedb/couchdb/indexColorSortName.json
[getIndexInfo] indexPath=META-INF/statedb/couchdb/collections/indexes/indexCollMarbles.json
[getIndexInfo] indexPath=META-INF/statedb/couchdb/collections/indexCollMarbles.json
[getIndexInfo] indexPath=META-INF/statedb/
--- PASS: TestGetIndexInfo (0.00s)
```

cendhu: @mbwhite Existing `TestGetIndexInfo()` has a bug too. It should not pass the path including the index def, i.e., `*.json`. The input to `getIndexInfo()` comes from the `HandleDeployChaincode()` which passes only the directory excluding the actual filename. 

We need to fix

1. `TestGetIndexInfo()`
2. Comments  (provided in the const block)
3. `getIndexInfo()`

cendhu: Some tests in `kvledger` package and integration test also fail after making `getIndexInfo()` to consider the full path including the index file.

mbwhite: @cendhu ah.. that makes more sense now thanks... what can we say about the directory path that is passed in?
The code as-is is uses '/' to split the path, is it safe to assume the input will not have leading or trailing directory separators?

(aside I assume as well that Fabric is solely intended to work on a Linux filesystem?  / not being the universal file separator :-) )

cendhu: @mbwhite 
https://github.com/hyperledger/fabric/blob/7151302be2dff77a9d78cdc8c7a04c4fee352914/core/common/ccprovider/cc_statedb_artifacts_provider.go#L64 This function extracts files entries from the chaincode package (tar bytes). Hence, I assume no leading or trailing directory separator unless there is a bug in `ExtractFileEntries()`. I do not see a unit test for `ExtractFileEntries()`. Not sure whether it is tested somewhere and I miss it. Hence, it is better to verify that. 

As far as I know, even on Windows, we recommend docker to be used (https://hyperledger-fabric.readthedocs.io/en/release-1.4/prereqs.html) and our docker base image is alpine linux. Hence, `/` as a file separator is okay.

mbwhite: After a bit more research I believe there are some in db_test.go.  My reading of that function is that it later functions can safely assume there are no leading or trailing /

My background in cross-platform development (pre Docker) has made me question whenever I see anything about a filesystem 'hard coded' like a /

cendhu: I see. Thanks for confirming.

manish-sethi (approved):

manish-sethi:

manish-sethi:

PR #2727, opened 2021-06-30
fix invalid memory panic when release db in high concurrency of chaincode query

huaweibcs:

Type of change
- Bug fix

Description
In the condition of high concurrency of chaincode query, peer will crash after TXSimulator finish SimulateProposal, because queryExecutor's dbItr release db again and the function AddInt32 in Release will panic.
Hence, inspect whether db is nil before dbIter release db.

### DIFF
--- a/vendor/github.com/syndtr/goleveldb/leveldb/db_iter.go
+++ b/vendor/github.com/syndtr/goleveldb/leveldb/db_iter.go
@@ -335,7 +335,7 @@ func (i *dbIter) Value() []byte {
 }
 
 func (i *dbIter) Release() {
-	if i.dir != dirReleased {
+	if i.dir != dirReleased && i.db != nil {
 		// Clear the finalizer.
 		runtime.SetFinalizer(i, nil)
 


ale-linux:
I'm not sure I understand the scenario - but it seems to me we're fixing the effect (nil pointer) rather than the cause (why that pointer is not valid). @manish-sethi - your thoughts?

cendhu (changes requested):
@wuqiaomin Can you create a JIRA and add some logs to explain why/when the panic occurs? It would be good to get the complete call stack if possible. 

Moreover, we cannot directly change a file in the dependency (https://github.com/syndtr/goleveldb). 

Note that we can safely call `iter.Release()` multiple times. Though it is not safe to call `iter.Relealse()` after closing the database, i.e., `db.Close()`, still it would not panic. Hence, I am wondering how the `db` became `nil` and caused panic in `iter.Release()`.

manish-sethi:
Not exactly this but I had come across something in these lines that caused a test flake. In that case, it was a race condition between Iterator opening and closing of goleveldb. This was fixed in [this](https://github.com/hyperledger/fabric/commit/08aabe9187b978e70455cc08a3dfac9ef5836363) commit.

This reported issue seem some other concurrency issue inside goleveldb, because we are not calling DB close ever in a regular running peer. Full panic stack trace will be helpful here.

Also, as @cendhu mentioned, if it turns out to be an issue that cannot be controlled from outside by changing the way we use the APIs and needs a fix in goleveldb, the fix should be submitted to goleveldb repo and we can upgrade the goleveldb.

I think that this PR can be closed and the bug and potential solution be discussed in a Jira.

huaweibcs:
It's important to add that it is an occasional phenomenon in the case of high concurrency and high CPU and memory rate of utilization, and no sooner panic hit than peer restart. 
Related Jira[ issue](https://jira.hyperledger.org/projects/FAB/issues/FAB-18510?filter=allopenissues&orderby=created+DESC%2C+priority+DESC%2C+updated+DESC) has been created.
So I close this PR.

PR #2732, opened 2021-07-01
[FAB-11334] Adds a new 'peer node unjoin' feature

jkneubuh:
The UnjoinChannel routine will mark a ledger as pending delete, removing all
channel specific data from the peer.  The routine must be invoked while the
peer is down.  Peer operators may issue the deletion by running the companion
'peer node unjoin' command from the console.

#### Type of change

- New feature

#### Description

[As a peer admin, I need a way to pause/resume a channel](https://jira.hyperledger.org/browse/FAB-16035)

Unjoining a is a helpful routine for operators managing channel lifecycles.  In some cases, a network participant may opt out of a channel, requesting a complete deletion of all channel specific data from the ledger.  While an operator may issue a node reset after pausing a channel, this is not viable in cases where the re-initialization requires the synchronization of a large (millions / billions) number of transactions. 

#### Related issues

[FAB-16035](https://jira.hyperledger.org/browse/FAB-16035)
[FAB-4481](https://jira.hyperledger.org/browse/FAB-4481)
[FAB-17787](https://jira.hyperledger.org/browse/FAB-17787)
[FAB-17801](https://jira.hyperledger.org/browse/FAB-17801)


yacovm:
> The routine must be invoked while the peer is down.


Can you explain what is the fundamental reason we are shutting down the peer for that? Can't this be done while the peer is running and servicing other channels?

jkneubuh:
Hi @yacovm, thanks for the inquiry and this is a good question.  There is no _fundamental_ reason why unjoin can not operate while the peer is running.  @manish-sethi articulated the pros/cons of running the [unjoin in both online and offline](https://jira.hyperledger.org/browse/FAB-11334?focusedCommentId=71881&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-71881) modes - after review of approach alternatives we elected to run the unjoin in offline mode for simplicity and as a compromise to the current operational practice of pause+reset.  Everyone is in agreement that there is still a need for the online unjoin channel operation.

core/ledger/kvledger/unjoin_channel_test.go
```diff
@@ -0,0 +1,159 @@
...
+	provider = testutilNewProvider(conf, t, &mock.DeployedChaincodeInfoProvider{})
+	activeLedgerIDs, err := provider.List()
+	require.NoError(t, err)
+	require.Len(t, activeLedgerIDs, 0)
+	provider.Close()
+
+	// Subsequent unjoins will not throw an error
+	require.NoError(t, UnjoinChannel(conf, ledgerID))
+	require.NoError(t, UnjoinChannel(conf, ledgerID))
+	require.NoError(t, UnjoinChannel(conf, ledgerID))
```
manish-sethi: - This is not a desired behavior. You got this here because you missed deleting the entry from the idStore. Otherwise, it would have rightly returned this [error](https://github.com/hyperledger/fabric/blob/4c4e58c1228968cefc9d5c027f10884fb3707174/core/ledger/kvledger/kv_ledger_provider.go#L634).

- Another aspect of the behavior would be that it should be able to create the ledger with the same ID again. But this test perhaps fits better in kvledger/tests folder - where we maintain the tests for testing the behavior under various scenarios (including when the ledger contains private data) unlike in this package where the tests are more for very basic code coverage. Keeping that in mind, I would suggest that the two tests that you may want to retain here in package would be the first and last only as the rest of the tests would be covered in kvledger/tests package.

core/ledger/kvledger/unjoin_channel.go
```diff
@@ -11,15 +11,58 @@ import (
...
+	// Set the ledger to a pending deletion status.  At the next startup, the peer will complete the removal of any partially deleted ledgers.
+	if err := updateLedgerStatus(config, ledgerID, msgs.Status_UNDER_DELETION); err != nil {
+		return errors.Wrapf(err, "Unjoin channel [%v]", ledgerID)
+	}
+
+	// remove the ledger data
+	if err := removeLedgerData(config, ledgerID); err != nil {
+		return errors.Wrapf(err, "deleting ledger [%v]", ledgerID)
+	}
+
```
manish-sethi, line 49: - After having the data deleted successfully, we need to delete the entry of the ledger from the idStore; leaving the peer in the state that it can join a channel with same name again.
- We would need make change in the boot process so that we continue to delete the partially deleted ledgers, perhaps because of a crash during execution of this commend - as we do for partially constructed [here](https://github.com/hyperledger/fabric/blob/4c4e58c1228968cefc9d5c027f10884fb3707174/core/ledger/kvledger/kv_ledger_provider.go#L122). But this change can be done in a separate PR as it would need it's own tests to make sure the crash consistency of the deletion.

core/ledger/kvledger/unjoin_channel_test.go
```diff
@@ -0,0 +1,159 @@
...
+}
+
+func TestUpdateLedgerStatus(t *testing.T) {
```
manish-sethi: Though it does not harm having this as a separate test but in any case, the assertions made in this tests should be made in the test `TestUnjoinChannel` above and as such this test can be removed.

core/ledger/kvledger/unjoin_channel_test.go
```diff
@@ -0,0 +1,159 @@
...
+	defer provider.Close()
+
+	ledgerID := constructTestLedgerID(1)
+	genesisBlock, _ := configtxtest.MakeGenesisBlock(ledgerID)
+	_, err := provider.CreateFromGenesisBlock(genesisBlock)
+	require.NoError(t, err)
+
+	// Fail when provider is open (e.g. peer is running)
+	require.Error(t, UnjoinChannel(conf, ledgerID),
+		"as another peer node command is executing, wait for that command to complete its execution or terminate it before retrying")
```
manish-sethi: For matching the errors, either we use `EqualError` or `Contains` funciton.

manish-sethi (commented):
Thanks @jkneubuh for your first PR. Left you a few comments.

yacovm:
> Hi @yacovm, thanks for the inquiry and this is a good question. There is no _fundamental_ reason why unjoin can not operate while the peer is running. @manish-sethi articulated the pros/cons of running the [unjoin in both online and offline](https://jira.hyperledger.org/browse/FAB-11334?focusedCommentId=71881&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-71881) modes - after review of approach alternatives we elected to run the unjoin in offline mode for simplicity and as a compromise to the current operational practice of pause+reset. Everyone is in agreement that there is still a need for the online unjoin channel operation.

Understood, thanks.

So the reason I asked that, is, that if the peer boots after being operated on and having the channel being marked as under deletion, it will obviously not initialize the in-memory structures and goroutines of that channel, because it will first delete the ledger and then resume bootstrapping.
However, other peers will then still think that peer is in the channel, and will try to replicate blocks from it, and send blocks to it, and the peer will still appear in discovery queries for a while. While it won't really harm any operations, it will pollute the logs, confuse clients (incorrect discovery results) and will surely raise questions from operators and administrators.

In gossip, I designed the ability to make the peer [leave a channel](https://github.com/hyperledger/fabric/commit/e2375ff0c04ff7b6565961247847e2d4c3866213) (many years ago, in 2017) and it was under the assumption the channel leave API will be triggered when the peer is online, which will give the peer time to broadcast a message that makes other peers ignore it is in the channel. 

While it is possible to still artificially call [JoinChan](https://github.com/hyperledger/fabric/blob/b5a4fe942ea819011e403aeb843ad6f62564ffd4/gossip/gossip/gossip_impl.go#L191) after bootstrapping, on gossip, just for the channel, and then immediately call LeaveChannel, to make that message broadcasted, it is not effective as the membership view will most likely be empty when the peer boots up. 

So, maybe this gap should be documented at least.

core/ledger/kvledger/unjoin_channel.go
```diff
@@ -11,15 +11,58 @@ import (
...
+	if err != nil {
+		return err
+	}
+	defer idStore.db.Close()
+
+	if err := idStore.updateLedgerStatus(ledgerID, status); err != nil {
+		return err
+	}
+
+	return nil
```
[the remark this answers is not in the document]
jkneubuh: Hm... I mildly disagree with this pattern / convention.  It's OK but I actually prefer the explicit assignment and test of the error routine for absolute consistency.  It seems like many (all?) golang routines consist of a serial flow of do-something+check-error+do-something+check-error+... invocations.    When then final invocation in a series of calls is used as the error for the outer function, the reader needs to stop and think about the return type of the last call in the chain, rather than recognize the pattern as a a cookie cutter linear flow.  For instance, in the two examples above, the explicit 'return nil' case is 100% clear that the routine is returning an OK result.  In the latter, looking at the bottom of the routine it's not 100% clear that the method is returning an error, the new status, or the previous status of the ledger.   Call me simple but I prefer the explicit style - the intent is clearer, and leaves opportunities for post-call logging, future method calls without restructuring, and ... ease of use when attached to a debugger.  IMHO it's worth the extra typing. :)

core/ledger/kvledger/unjoin_channel.go
```diff
@@ -11,15 +11,56 @@ import (
...
+		return errors.WithMessagef(err, "deleting ledger [%s]", ledgerID)
+	}
+
+	// Delete the ledger from the ID storage after the contents have been purged.
+	if err := idStore.deleteLedgerID(ledgerID); err != nil {
+		return errors.WithMessagef(err, "deleting ledger [%s]", ledgerID)
+	}
+
+	logger.Infow("channel has been successfully unjoined", "ledgerID", ledgerID)
+	return nil
```
manish-sethi, line 61: Unlike error strings, log statements would typically start with a full sentence.

core/ledger/kvledger/unjoin_channel.go
```diff
@@ -11,15 +11,56 @@ import (
...
 	"github.com/hyperledger/fabric/core/ledger/kvledger/bookkeeping"
 	"github.com/hyperledger/fabric/core/ledger/kvledger/history"
+	"github.com/hyperledger/fabric/core/ledger/kvledger/msgs"
 	"github.com/hyperledger/fabric/core/ledger/kvledger/txmgmt/privacyenabledstate"
 	"github.com/hyperledger/fabric/core/ledger/pvtdatastorage"
+	"github.com/pkg/errors"
 )
 
+// UnjoinChannel removes the data for a ledger and sets the status to UNDER_DELETION.  This function is to be
+// invoked while the peer is shut down.
```
manish-sethi, line 27: Stale function comment as this function now clears the UNDER_DELETION status.

manish-sethi (approved):
A couple nit comments. Can be fixed in a separate PR or with next PR.

PR #2748, opened 2021-07-10
[FAB-18521] Replicate block metadata with block while OSN catching up

C0rWin:
#### Type of change

- Bug fix

#### Description

While OSN catches up replicating block from the up-to-date replica the consenters' metadata information omitted, i.e.

```
c.support.WriteBlock(block, nil)
```

where `nil` substitutes for block's metadata.

In this commit, the consenters' metadata extracted from the replicated block and being written with the block.

### DIFF
--- a/integration/raft/cft_test.go
+++ b/integration/raft/cft_test.go
@@ -27,6 +27,7 @@ import (
 	conftx "github.com/hyperledger/fabric-config/configtx"
 	"github.com/hyperledger/fabric-protos-go/common"
 	"github.com/hyperledger/fabric-protos-go/msp"
+	"github.com/hyperledger/fabric-protos-go/orderer/etcdraft"
 	"github.com/hyperledger/fabric/common/configtx"
 	"github.com/hyperledger/fabric/common/util"
 	"github.com/hyperledger/fabric/integration/nwo"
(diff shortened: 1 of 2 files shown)


yacovm:
I think this bug deserves a JIRA so we can track backporting, no? 

Also how did we never encountered this before? How do orderers find the consenter mappings without that metadata if they catch up? 

Perhaps consider making the integration test reproduce the problem so we can ensure we're not missing anything

C0rWin:
> I think this bug deserves a JIRA so we can track backporting, no?

No problem, I will open JIRA with a more detailed explanation of the issue ;)

> Also, how did we never encountered this before? How do orderers find the consenter mappings without that metadata if they catch up?

I am asking myself the same question. 

> Perhaps consider making the integration test reproduce the problem so we can ensure we're not missing anything

Sure, I will make changes to IT to stress out the issue.

yacovm:
So when I remove your fix from the production code and re-run the test, it your metadata equality assertion fails, but... the orderer1 still manages to reconstruct the node identities correctly and as a result, to locate the leader and the rest of the nodes. 

I guess, this is because we never do a configuration change in the test, therefore the "empty" metadata is equivalent to the non empty metadata in the case of membership construction.

What I think we should strive for, is a test that if your production code change is removed, then the orderer never manages to recover.

C0rWin:
> So when I remove your fix from the production code and re-run the test, it your metadata equality assertion fails, but... the orderer1 still manages to reconstruct the node identities correctly and as a result, to locate the leader and the rest of the nodes.
> 
> I guess, this is because we never do a configuration change in the test, therefore the "empty" metadata is equivalent to the non empty metadata in the case of membership construction.
> 
> What I think we should strive for, is a test that if your production code change is removed, then the orderer never manages to recover.

Yes, I realized after pushed the update, that as long as there is no reconfiguration and metadata is absent during the bootstrap OSN will take info from config, this is why I think we never saw this issue manifested before (cause case is quite rare).

PS. Will add reconfiguration.

yacovm (approved):
LGTM, the test fails now without the fix

yacovm:

yacovm:

denyeart:
Looks like this merge is causing integration test failures, see #2759 CI.

C0rWin:
> Looks like this merge is causing integration test failures, see #2759 CI.

@denyeart https://github.com/hyperledger/fabric/pull/2761 should fix that

PR #2750, opened 2021-07-12
Ouput Directory For Divergent Results

jcastrence:
#### Type of change

#### Description

- Several changes were made to the ledger utility tool in preparation for future updates, most notably: output files are now created in a new directory, compare tool no longer overwrites existing files and will throw an error instead, and improved usability and consistency for the outputDir flag

### DIFF
--- a/cmd/ledgerutil/main.go
+++ b/cmd/ledgerutil/main.go
@@ -9,14 +9,13 @@ package main
 import (
 	"fmt"
 	"os"
-	"path/filepath"
 
 	"github.com/hyperledger/fabric/internal/ledgerutil"
 	"gopkg.in/alecthomas/kingpin.v2"
 )
 
 const (
-	resultFilename = "result.json"
+	compareErrorMessage = "Ledger Compare Error: "
 )
 
 var (
@@ -41,30 +40,31 @@ func main() {
 		return
 	}
 
-	// Determine result json file location
-	var resultFilepath string
-	if outputDir != nil {
-		resultFilepath = *outputDir
-	} else {
-		resultFilepath, err = os.Getwd()
+	// Set results directory location; an empty location sets the current directory by default
+	if *outputDir != "" {
+		err = os.Chdir(*outputDir)
 		if err != nil {
-			fmt.Println(err)
+			fmt.Printf("%s%s\n", compareErrorMessage, err)
 			return
 		}
 	}
-	resultFilepath = filepath.Join(resultFilepath, resultFilename)
+	*outputDir, err = os.Getwd()
+	if err != nil {
+		fmt.Printf("%s%s\n", compareErrorMessage, err)
+		return
+	}
 
 	// Command logic
 	switch command {
 
 	case compare.FullCommand():
 
-		count, err := ledgerutil.Compare(*snapshotPath1, *snapshotPath2, resultFilepath)
+		count, _, err := ledgerutil.Compare(*snapshotPath1, *snapshotPath2, *outputDir)
 		if err != nil {
-			fmt.Println(err)
+			fmt.Printf("%s%s\n", compareErrorMessage, err)
 			return
 		}
-		fmt.Printf("\nSuccessfully compared snapshots. Result saved to %s. Total differences found: %d\n", resultFilepath, count)
+		fmt.Printf("\nSuccessfully compared snapshots. Result saved to %s. Total differences found: %d\n", *outputDir, count)
 
 	case troubleshoot.FullCommand():
 
--- a/internal/ledgerutil/compare.go
+++ b/internal/ledgerutil/compare.go
@@ -23,48 +23,67 @@ import (
 	"github.com/pkg/errors"
 )
 
+const (
+	resultsDir      = "ledger_comparison_results"
+	allDivsFilename = "all_divergences.json"
+)
+
 // Compare - Compares two ledger snapshots and outputs the differences in snapshot records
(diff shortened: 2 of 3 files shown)


internal/ledgerutil/compare.go
```diff
@@ -164,24 +183,24 @@ func Compare(snapshotDir1 string, snapshotDir2 string, outputFile string) (count
...
 		}
 	}
 
 	err = jsonOutputFile.close()
 	if err != nil {
-		return 0, err
+		return 0, "", err
 	}
-	return jsonOutputFile.count, nil
+	return jsonOutputFile.count, allRes, nil
```
manish-sethi, line 203: I don't think that we need to return the file name from here. This is not used anywhere anyways!

internal/ledgerutil/compare.go
```diff
@@ -23,48 +23,67 @@ import (
...
+	}
+	err = os.Mkdir(resultsDir, 0o777)
+	if err != nil {
+		return 0, "", err
 	}
 
-	// Create the output file
-	jsonOutputFile, err := newJSONFileWriter(outputFile)
+	allRes = filepath.Join(wd, resultsDir, allDivsFilename)
+	jsonOutputFile, err := newJSONFileWriter(allRes)
```
manish-sethi, line 62: I find this code a bit ad-hoc. Earlier, the file name was being passed to this function. All we need to change is that this function will take the outputdir (instead of file) and make the file name a constant in this file.

internal/ledgerutil/compare.go
```diff
@@ -23,48 +23,67 @@ import (
 	"github.com/pkg/errors"
 )
 
+const (
+	resultsDir      = "ledger_comparison_results"
+	allDivsFilename = "all_divergences.json"
```
manish-sethi, line 28: I understand that we would produce more than one file, e.g.,  `all_divergences.json` would contain all diversions and perhaps `top_earliest_divergences.json` would contain top-10. But I am not sure, what is the significant of one extra dir `ledger_comparison_results`. I think that you can produce the output files directly in the user supplied outDir

internal/ledgerutil/compare.go
```diff
@@ -23,48 +23,67 @@ import (
...
+	// Create the output files
+	err = os.Chdir(outputDir)
+	if err != nil {
+		return 0, "", err
+	}
+	wd, err := os.Getwd()
+	if err != nil {
+		return 0, "", err
+	}
+	err = os.Mkdir(resultsDir, 0o777)
```
manish-sethi, line 56: 755?

internal/ledgerutil/compare.go
```diff
@@ -23,48 +23,67 @@ import (
...
 
+const (
+	resultsDir      = "ledger_comparison_results"
+	allDivsFilename = "all_divergences.json"
+)
+
 // Compare - Compares two ledger snapshots and outputs the differences in snapshot records
-// This function will overwrite the file at outputPath if it already exists
-func Compare(snapshotDir1 string, snapshotDir2 string, outputFile string) (count int, err error) {
+// This function will not overwrite existing files
```
manish-sethi, line 32: This is worth adding a UT.

cmd/ledgerutil/main.go
```diff
@@ -41,30 +40,31 @@ func main() {
...
 
-	// Determine result json file location
-	var resultFilepath string
-	if outputDir != nil {
-		resultFilepath = *outputDir
-	} else {
-		resultFilepath, err = os.Getwd()
+	// Set results directory location; an empty location sets the current directory by default
+	if *outputDir != "" {
+		err = os.Chdir(*outputDir)
```
manish-sethi, line 45: I don't think that we should be changing the working dir.

manish-sethi (changes requested):

internal/ledgerutil/compare.go
jcastrence: This was done for the tests because they require the full file path

internal/ledgerutil/compare.go
jcastrence: The code is implemented this way to mimic the way the cd command works on unix based shells, the goal being to improve usability. Otherwise, the user would have to specifically use the full file path for consistent results.

internal/ledgerutil/compare.go
jcastrence: I feel that creating a directory for both results organizes everything a lot better. This is similar to how the snapshot tool works. Maybe we can get some more opinions on this from other maintainers?

cmd/ledgerutil/main.go
jcastrence: The reasoning for this change is the same as the comment above

internal/ledgerutil/compare.go
manish-sethi: Not sure which snapshot tool are you referring to. Assuming you are referring to the snapshot generation, all the files are produced in a single folder.

internal/ledgerutil/compare.go
manish-sethi: Not sure I understand what you mean by "user would have to specifically use the full file path for consistent results". Isn't outDir already contains the path, either full or relative, whatever user wants?

manish-sethi (changes requested):

jcastrence:
Since most of these changes have been deemed unnecessary, this PR will be closed and the remaining changes will be rolled into a new PR.

PR #2754, opened 2021-07-14
[FAB-11334] Scrubs partially constructed/deleted ledgers at peer init

jkneubuh:
Peer init will now remove any ledger with status `UNDER_CONSTRUCTION` or `UNDER_DELETION` 
at startup. 

#### Type of change

#### Description

Unjoining a peer from a channel will mark the ledger with status UNDER_DELETION
and proceed with the ledger removal.  If the unjoin operation fails, the ledger
will include residue from the partial deletion, leaving the kvledger in a
questionable state.  This commit forces the peer initialization to scan for any
ledgers with an UNDER_DELETION status, removing partial ledgers from the peer 
at startup.

#### Related issues

- FAB-16035
- FAB-4481
- FAB-17787
- FAB-17801

### DIFF
--- a/core/ledger/kvledger/delete_partial_ledgers_test.go
+++ b/core/ledger/kvledger/delete_partial_ledgers_test.go
@@ -1,0 +1,117 @@
+/*
+Copyright IBM Corp. All Rights Reserved.
+
+SPDX-License-Identifier: Apache-2.0
+*/
+
+package kvledger
+
+import (
+	"testing"
+
+	"github.com/hyperledger/fabric/core/ledger/kvledger/msgs"
+	"github.com/hyperledger/fabric/core/ledger/mock"
+	"github.com/stretchr/testify/require"
+)
+
+func TestDeleteUnderDeletionLedger(t *testing.T) {
+	conf, cleanup := testConfig(t)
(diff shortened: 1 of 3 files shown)


core/ledger/kvledger/delete_partial_ledgers_test.go
```diff
@@ -0,0 +1,117 @@
...
+	// doom the newly created ledger
+	require.NoError(t, provider.idStore.updateLedgerStatus(ledgerID, msgs.Status_UNDER_DELETION))
+	verifyLedgerIDExists(t, provider, ledgerID, msgs.Status_UNDER_DELETION)
+
+	// explicitly call the delete on the provider.
+	provider.deletePartialLedgers()
+
+	// goodbye, ledger.
+	verifyLedgerDoesNotExist(t, provider, ledgerID)
+}
```
manish-sethi, line 40: Though, the extra test does not harm and in this particular case it's very small enough addition that could easily be ignored. However, in general we should prefer to be thoughtful of the purpose of an additional test. 
- It's hard to think of a bug/issue which would make this test pass while the fail the next one. 
- Neither does this test tests a separate section of the code.

manish-sethi (approved):

PR #2760, opened 2021-07-16
Output File Exists Error

jcastrence:
#### Type of change

- Test update

#### Description

- Compare tool no longer overwrites existing files and will throw an error instead. Compare function now takes in the output directory location instead of the output file as an argument and outputs results to a generated output directory.

### DIFF
--- a/cmd/ledgerutil/main.go
+++ b/cmd/ledgerutil/main.go
@@ -16,7 +16,7 @@ import (
 )
 
 const (
-	resultFilename = "result.json"
+	compareErrorMessage = "Ledger Compare Error: "
 )
 
 var (
@@ -41,30 +41,42 @@ func main() {
 		return
 	}
 
-	// Determine result json file location
-	var resultFilepath string
-	if outputDir != nil {
-		resultFilepath = *outputDir
-	} else {
-		resultFilepath, err = os.Getwd()
-		if err != nil {
-			fmt.Println(err)
-			return
-		}
-	}
-	resultFilepath = filepath.Join(resultFilepath, resultFilename)
-
 	// Command logic
 	switch command {
 
 	case compare.FullCommand():
 
-		count, err := ledgerutil.Compare(*snapshotPath1, *snapshotPath2, resultFilepath)
+		cleanpath := filepath.Clean(*outputDir)
+
+		// Determine result json file location
+		if *outputDir == "" {
+			*outputDir, err = os.Getwd()
+			if err != nil {
+				fmt.Printf("%s%s\n", compareErrorMessage, err)
+				return
+			}
+			cleanpath = "the current directory"
+		}
+		// Clean output
+		switch cleanpath {
+		case ".":
+			cleanpath = "the current directory"
+		case "..":
+			cleanpath = "the parent directory"
+		}
+
+		count, err := ledgerutil.Compare(*snapshotPath1, *snapshotPath2, *outputDir)
 		if err != nil {
-			fmt.Println(err)
+			fmt.Printf("%s%s\n", compareErrorMessage, err)
 			return
 		}
-		fmt.Printf("\nSuccessfully compared snapshots. Result saved to %s. Total differences found: %d\n", resultFilepath, count)
+
+		fmt.Print("\nSuccessfully compared snapshots. ")
+		if count == -1 {
+			fmt.Println("Both snapshot public state hashes were the same. No results were generated.")
+		} else {
+			fmt.Printf("Results saved to %s. Total differences found: %d\n", cleanpath, count)
+		}
 
 	case troubleshoot.FullCommand():
 
--- a/internal/ledgerutil/compare.go
+++ b/internal/ledgerutil/compare.go
@@ -10,6 +10,7 @@ import (
 	"bufio"
 	"bytes"
 	"encoding/json"
+	"fmt"
 	"io/ioutil"
 	"os"
 	"path/filepath"
@@ -23,24 +24,48 @@ import (
 	"github.com/pkg/errors"
 )
 
+const (
+	// AllDiffsByKey - Filename for the json output that contains all differences ordered by key
+	AllDiffsByKey = "all_diffs_by_key.json"
+)
+
 // Compare - Compares two ledger snapshots and outputs the differences in snapshot records
-// This function will overwrite the file at outputPath if it already exists
-func Compare(snapshotDir1 string, snapshotDir2 string, outputFile string) (count int, err error) {
+// This function will throw an error if the output directory already exist in the outputDirLoc
+// Function will return count of -1 if the public state hashes are the same
+func Compare(snapshotDir1 string, snapshotDir2 string, outputDirLoc string) (count int, err error) {
 	// Check the hashes between two files
 	hashPath1 := filepath.Join(snapshotDir1, kvledger.SnapshotSignableMetadataFileName)
 	hashPath2 := filepath.Join(snapshotDir2, kvledger.SnapshotSignableMetadataFileName)
 
-	equal, err := snapshotsComparable(hashPath1, hashPath2)
+	equal, channelName, blockHeight, err := snapshotsComparable(hashPath1, hashPath2)
 	if err != nil {
 		return 0, err
 	}
 
 	if equal {
-		return 0, errors.New("both snapshots public state hashes are same. Aborting compare")
+		return -1, nil
+	}
+
+	// Output directory creation
+	outputDirName := fmt.Sprintf("%s_%d_comparison", channelName, blockHeight)
+	outputDirPath := filepath.Join(outputDirLoc, outputDirName)
+
+	empty, err := fileutil.CreateDirIfMissing(outputDirPath)
+	if err != nil {
+		return 0, err
+	}
+	if !empty {
+		switch outputDirLoc {
+		case ".":
+			outputDirLoc = "the current directory"
+		case "..":
+			outputDirLoc = "the parent directory"
+		}
+		return 0, errors.Errorf("%s already exists in %s. Choose a different location or remove the existing results. Aborting compare", outputDirName, outputDirLoc)
 	}
 
 	// Create the output file
-	jsonOutputFile, err := newJSONFileWriter(outputFile)
+	jsonOutputFile, err := newJSONFileWriter(filepath.Join(outputDirPath, AllDiffsByKey))
 	if err != nil {
 		return 0, err
 	}
@@ -57,7 +82,7 @@ func Compare(snapshotDir1 string, snapshotDir2 string, outputFile string) (count
 		return 0, err
 	}
 
-	// Read each snapshot record  to begin looking for divergences
+	// Read each snapshot record  to begin looking for differences
 	namespace1, snapshotRecord1, err := snapshotReader1.Next()
 	if err != nil {
 		return 0, err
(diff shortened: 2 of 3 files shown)


internal/ledgerutil/compare.go
```diff
@@ -39,8 +44,19 @@ func Compare(snapshotDir1 string, snapshotDir2 string, outputFile string) (count
...
+	allDivsFilepath := filepath.Join(outputDir, AllDivsFilename)
+	outputFullpath, err := filepath.Abs(outputDir)
+	if err != nil {
+		return 0, err
+	}
+
+	// Check for existing output files
+	if _, err := os.Stat(allDivsFilepath); !os.IsNotExist(err) {
+		return 0, errors.Errorf("%s already exists in %s. Aborting compare", AllDivsFilename, outputFullpath)
+	}
```
manish-sethi: A better check would be just to test that the dir is not empty. Going forward, you would produce more files and  mixing of files from different run would not be desired.

cmd/ledgerutil/main.go
```diff
@@ -42,29 +42,29 @@ func main() {
...
 		if err != nil {
-			fmt.Println(err)
+			fmt.Printf("%s%s\n", compareErrorMessage, err)
 			return
 		}
-		fmt.Printf("\nSuccessfully compared snapshots. Result saved to %s. Total differences found: %d\n", resultFilepath, count)
+		fullpath, err := filepath.Abs(*outputDir)
+		if err != nil {
+			fmt.Printf("%s%s\n", compareErrorMessage, err)
+		}
```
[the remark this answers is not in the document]
jcastrence: Had it print absolute path because it prints the literal input the user provides, so if they used the "." for current directory, that's what gets printed back. I'll have it print relative paths and just include edge cases for "." and "..".

internal/ledgerutil/compare.go
jcastrence: I'm not sure what you mean. Since the user can choose to produce the files in any directory, are you saying the command should fail if the directory they have given the tool is not empty? If that's the case, then shouldn't we also create this directory for the user?

jcastrence: Also, files would not be mixed as long as we check for all output files and throw an error if any of them exist.

manish-sethi: Why would you want to do the work for first file and then return error if the second file is present. What's the point of allowing user to write to a dir that may have some unrelated files. 

By mixing what I meant was that as this tool is still evolving, going forward some of the files are going to be optional. For example, when you would add the files for the private data hashes comparison. If your code is not going to produce these files, likely, you will miss checking for the present of those as well and some files from previous runs may get mixed up. Not sure what is the issue in expecting an empty dir to produce the output. If you want to create a new dir, I am fine with that as well if we can come up with some unique name (e.g., append the two input snapshot folders name).
 What I did not find maintainable was testing for file by file one at a time while producing that file.

cmd/ledgerutil/main.go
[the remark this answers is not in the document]
manish-sethi: I think that it should be fine to print in the format user gave the input. IMO, lets not think too much in beautification of output at this stage.

internal/ledgerutil/compare.go
jcastrence: I think concatenating the two snapshot directory names to create the result directory name should work.

internal/ledgerutil/compare.go
```diff
@@ -39,8 +44,22 @@ func Compare(snapshotDir1 string, snapshotDir2 string, outputFile string) (count
...
+	// Output directory creation
+	outputDirName := filepath.Base(snapshotDir1) + "_and_" + filepath.Base(snapshotDir2) + "_compared"
+	outputDirPath := filepath.Join(outputDirLoc, outputDirName)
+
+	// Check for existing output directory
+	if _, err := os.Stat(outputDirPath); !os.IsNotExist(err) {
+		return 0, errors.Errorf("%s already exists in %s. Choose a different location or remove the existing results. Aborting compare", outputDirName, outputDirLoc)
+	}
+
+	err = os.Mkdir(outputDirPath, 0o755)
```
manish-sethi: - This is not a correct check. Any general I/O error will cause to proceed inside this `if` block - which is not your intention. You can use use[ this function](https://github.com/hyperledger/fabric/blob/240cf0e182938a09832ba4162c3c229c81848e3f/internal/fileutil/fileutil.go#L120) here.

- Even simpler would be to simply use [this function](https://github.com/hyperledger/fabric/blob/240cf0e182938a09832ba4162c3c229c81848e3f/internal/fileutil/fileutil.go#L77) which will create the whole path and will return to you whether the dir is empty and you can simply check that the return value is `true`

jcastrence: Thanks for the tip

internal/ledgerutil/compare.go
```diff
@@ -39,8 +44,20 @@ func Compare(snapshotDir1 string, snapshotDir2 string, outputFile string) (count
 		return 0, errors.New("both snapshots public state hashes are same. Aborting compare")
 	}
 
+	// Output directory creation
+	outputDirName := filepath.Base(snapshotDir1) + "_and_" + filepath.Base(snapshotDir2) + "_compared"
```
denyeart: In my trial, I used snapshots of block height 4. This resulted in a directory of the name "4_and_4_compared". The block height will typically be the same for comparisons, so including it in the directory name doesn't make sense to me. Additionally, I think Manish mentioned earlier that we don't need to create new directories, we can just put the results directly in the output directory that the user specified.

internal/ledgerutil/compare.go
```diff
@@ -23,9 +23,14 @@ import (
 	"github.com/pkg/errors"
 )
 
+const (
+	// AllDivsFilename - Filename for the json output that contains all divergences
+	AllDivsFilename = "all_divergences.json"
```
denyeart: Given that the other file will be something like top_divergences_by_height.json, I was thinking this file should be something like all_divergences_by_key.json (since the difference between the files is that this one is sorted by key and the other one is sorted by height).

internal/ledgerutil/compare.go
```diff
@@ -39,8 +44,20 @@ func Compare(snapshotDir1 string, snapshotDir2 string, outputFile string) (count
...
+	// Output directory creation
+	outputDirName := filepath.Base(snapshotDir1) + "_and_" + filepath.Base(snapshotDir2) + "_compared"
+	outputDirPath := filepath.Join(outputDirLoc, outputDirName)
+
+	empty, err := fileutil.CreateDirIfMissing(outputDirPath)
+	if err != nil {
+		return 0, err
+	}
+	if !empty {
+		return 0, errors.Errorf("%s already exists in %s. Choose a different location or remove the existing results. Aborting compare", outputDirName, outputDirLoc)
```
denyeart, line 64: If user doesn't pass and output directory, this leads to an awkward message like
"4_and_4_compared already exists in . Choose a different location or remove the existing results."

cmd/ledgerutil/main.go
```diff
@@ -16,7 +16,7 @@ import (
 )
 
 const (
-	resultFilename = "result.json"
+	compareErrorMessage = "Ledger Compare Error: "
```
denyeart, line 19: If the snapshots are the same, it results in a message like:

"Ledger Compare Error: both snapshots public state hashes are same. Aborting compare"

As an end user, I wouldn't consider a consistent snapshot comparison to be a "Ledger Compare Error".  We should only indicate that it is an error if it is an unexpected scenario. Happy endings should result in a message rather than an error message.

internal/ledgerutil/compare.go
jcastrence: The reason I decided to add back code for creating a new directory was because eventually we are going to be outputting more files like the early divergences results and the private hash comparison results. If we want to give the responsibility of creating and naming the directory to the user, wouldn't it make more sense to create the directory within the tool by allowing the user to pass in the name they want to use? We could also have a generic name for cases where the user doesn't pass anything in. This would allow us to keep the same current functionality and eliminate confusing output directory names.

internal/ledgerutil/compare.go
jcastrence: I agree

internal/ledgerutil/compare.go
jcastrence: This is also what used to happen if the command was successful. Manish felt it wasn't necessary to focus on those edge cases but I can do a quick fix.

cmd/ledgerutil/main.go
```diff
@@ -41,30 +41,42 @@ func main() {
...
 	case compare.FullCommand():
 
-		count, err := ledgerutil.Compare(*snapshotPath1, *snapshotPath2, resultFilepath)
+		cleanpath := filepath.Clean(*outputDir)
+
+		// Determine result json file location
+		if *outputDir == "" {
+			*outputDir, err = os.Getwd()
+			cleanpath = "the current directory"
+			if err != nil {
```
denyeart, line 54: err checks should be done immediately after the statement that returned the error (move this up one line).

internal/ledgerutil/compare.go
```diff
@@ -23,9 +23,15 @@ import (
 	"github.com/pkg/errors"
 )
 
+const (
+	// AllDivsByKey - Filename for the json output that contains all divergences ordered by key
+	AllDivsByKey = "all_divergences_by_key.json"
```
denyeart:
```suggestion
	allDivsByKey = "all_divergences_by_key.json"
```

internal/ledgerutil/compare.go
```diff
@@ -23,9 +23,15 @@ import (
...
+)
+
 // Compare - Compares two ledger snapshots and outputs the differences in snapshot records
-// This function will overwrite the file at outputPath if it already exists
-func Compare(snapshotDir1 string, snapshotDir2 string, outputFile string) (count int, err error) {
+// This function will throw an error if the output directory already exist in the outputDirLoc
+// Function will return count of -1 if the public state hashes are the same
```
denyeart, line 34: If there are 0 differences, wouldn't the logical count to return be 0? Is there a reason to use the magic value of -1 instead of the logical value of 0?
You were probably trying to differentiate between the error case which also returns 0. But in the error case you can check for existence of error returned rather than rely on the count.

internal/ledgerutil/compare.go
denyeart: Yes, we will be outputting multiple files, but all those files can go into the single output directory that the user specified (or in the current directory if they did not specify an output directory).  I'm still not seeing a need to create a subdirectory.

EDIT: Julian explained that one benefit of the subdirectory is that it is easy to detect if the files already exist and error out if so. I agreed that if the subdirectory is based on <channelname>_<height>_comparison then it would make sense.

internal/ledgerutil/compare.go
denyeart: actually we could call them "differences" rather than "divergences"... I think that is a little easier to understand and say and shorten "diff" (unless you and Manish already landed on "divergences"). Also "div" makes me think of the html div tag, whereas diff is well understood.

internal/ledgerutil/compare.go
jcastrence: I thought we decided we weren't going to throw an error if the public state hashes were the same. If the user is simply checking to see if two snapshots are the same, and they end up being the same, this technically shouldn't be an error right?

internal/ledgerutil/compare.go
jcastrence: If we go this route, should we also follow Manish's suggestion and only allow the command to successfully generate output files if the user specified directory isn't empty? If so, this was the other purpose of creating an output directory, to ensure the output directory would be empty as to not overwrite existing output files. Otherwise, we would need to check for the existence of each output file individually, which Manish was against.

internal/ledgerutil/compare.go
denyeart: Right, no error, I'm simply suggesting if they are the same and no errors, the logical way to reflect that would be to return count of 0 rather than -1.

denyeart: Ah, I see what you were thinking now... you want to distinguish between no comparison done and (the unexpected) comparison done and 0 differences found... in that case -1 response is ok.

cmd/ledgerutil/main.go
```diff
@@ -41,30 +41,42 @@ func main() {
...
+			fmt.Printf("%s%s\n", compareErrorMessage, err)
 			return
 		}
-		fmt.Printf("\nSuccessfully compared snapshots. Result saved to %s. Total differences found: %d\n", resultFilepath, count)
+
+		fmt.Print("\nSuccessfully compared snapshots. ")
+		if count == -1 {
+			fmt.Println("Both snapshot public state hashes were the same. No results were generated.")
+		} else {
+			fmt.Printf("Results saved to %s. Total differences found: %d\n", cleanpath, count)
```
denyeart, line 78: Now that we set the subdirectory to `channelname_height_comparison`, that should be reported here instead of just the cleanpath. That is, return outputDirPath from Compare() so that we can report it here (ideally as an absolute path so that there is no doubt where the output is). That would actually eliminate the need for the cleanpath logic altogether... just report the actual path written including subdirectory.

The rest of the logic looks good, so we can defer that to next PR.

denyeart (approved):

PR #2761, opened 2021-07-16
[FAB-18521] Fixing flaky IT, send remove tx to another node

C0rWin:
PR #2748, introduced new IT to ensure the fix. However, there is some flakiness manifested with this IT, caused by sending the remove consenter transaction to the "to be removed" node. Removing the consenter is a config transaction where codes after sending it ensure a new block with the config update successfully committed, which is the root cause for the flakiness. Once OSN is removed from the channel, it no longer can server deliver requests for clients trying to fetch from it.

This commit, fixes it by sending remove OSN config updated transaction
to a different node instead.


yacovm (approved):

denyeart:

PR #2767, opened 2021-07-20
FABGW-25 Build ChaincodeInterest in TX simulator

andrew-coleman:
During transaction simulation, capture extra information to build a ChaincodeInterest structure specific to the chaincode invocation as follows:
- Which private data collections are read from.
- The SBE policies associated with any states that a written to.

Following simulations, the ChaincodeInterest structure is built and returned in the ProposalResponse message.

### DIFF
--- a/core/endorser/endorser.go
+++ b/core/endorser/endorser.go
@@ -174,7 +174,7 @@ func (e *Endorser) callChaincode(txParams *ccprovider.TransactionParams, input *
 }
 
 // SimulateProposal simulates the proposal by calling the chaincode
-func (e *Endorser) SimulateProposal(txParams *ccprovider.TransactionParams, chaincodeName string, chaincodeInput *pb.ChaincodeInput) (*pb.Response, []byte, *pb.ChaincodeEvent, error) {
+func (e *Endorser) simulateProposal(txParams *ccprovider.TransactionParams, chaincodeName string, chaincodeInput *pb.ChaincodeInput) (*pb.Response, []byte, *pb.ChaincodeEvent, *pb.ChaincodeInterest, error) {
 	logger := decorateLogger(endorserLogger, txParams)
 
 	meterLabels := []string{
@@ -186,11 +186,11 @@ func (e *Endorser) SimulateProposal(txParams *ccprovider.TransactionParams, chai
 	res, ccevent, err := e.callChaincode(txParams, chaincodeInput, chaincodeName)
 	if err != nil {
 		logger.Errorf("failed to invoke chaincode %s, error: %+v", chaincodeName, err)
-		return nil, nil, nil, err
+		return nil, nil, nil, nil, err
 	}
 
 	if txParams.TXSimulator == nil {
-		return res, nil, ccevent, nil
+		return res, nil, ccevent, nil, nil
 	}
 
 	// Note, this is a little goofy, as if there is private data, Done() gets called
@@ -201,14 +201,14 @@ func (e *Endorser) SimulateProposal(txParams *ccprovider.TransactionParams, chai
 	simResult, err := txParams.TXSimulator.GetTxSimulationResults()
 	if err != nil {
 		e.Metrics.SimulationFailure.With(meterLabels...).Add(1)
-		return nil, nil, nil, err
+		return nil, nil, nil, nil, err
 	}
 
 	if simResult.PvtSimulationResults != nil {
 		if chaincodeName == "lscc" {
 			// TODO: remove once we can store collection configuration outside of LSCC
 			e.Metrics.SimulationFailure.With(meterLabels...).Add(1)
-			return nil, nil, nil, errors.New("Private data is forbidden to be used in instantiate")
+			return nil, nil, nil, nil, errors.New("Private data is forbidden to be used in instantiate")
 		}
 		pvtDataWithConfig, err := AssemblePvtRWSet(txParams.ChannelID, simResult.PvtSimulationResults, txParams.TXSimulator, e.Support.GetDeployedCCInfoProvider())
 		// To read collection config need to read collection updates before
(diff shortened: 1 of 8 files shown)


core/ledger/kvledger/txmgmt/txmgr/tx_simulator.go
```diff
@@ -157,7 +179,15 @@ func (s *txSimulator) GetTxSimulationResults() (*ledger.TxSimulationResults, err
...
 	s.queryExecutor.addRangeQueryInfo()
-	return s.rwsetBuilder.GetTxSimulationResults()
+	simResults, err := s.rwsetBuilder.GetTxSimulationResults()
+	if err != nil {
+		return nil, err
+	}
+	simResults.PrivateReads = s.privateReads.Clone()
+	for _, policy := range s.keySignaturePolicies {
+		simResults.KeySignaturePolicies = append(simResults.KeySignaturePolicies, policy)
+	}
```
manish-sethi: Is this cloning of both required?

core/ledger/ledger_interface.go
```diff
@@ -470,10 +470,42 @@ func (filter PvtNsCollFilter) Has(ns string, coll string) bool {
...
+// Add a collection to the set of private data collections that are read by the chaincode.
+func (pr PrivateReads) Add(ns, coll string) {
+	if _, ok := pr[ns]; !ok {
+		pr[ns] = map[string]bool{}
+	}
+	pr[ns][coll] = true
+}
+
+// Read returns whether a private data collection has been read during TX simulation.
+func (pr PrivateReads) Read(ns, coll string) bool {
```
manish-sethi: `Exists` may be a more meaningful name.

core/ledger/ledger_interface.go
```diff
@@ -470,10 +470,42 @@ func (filter PvtNsCollFilter) Has(ns string, coll string) bool {
 	return collFilter[coll]
 }
 
+// PrivateReads captures which private data collections are read during TX simulation.
+type PrivateReads map[string]map[string]bool
```
manish-sethi: Not a big deal but better to use canonical type `map[string]map[string]struct{}` for a set.

core/endorser/endorser.go
```diff
@@ -500,3 +507,48 @@ func CreateCCEventBytes(ccevent *pb.ChaincodeEvent) ([]byte, error) {
...
+
+	chaincodes := map[string]struct{}{}
+	if simResult.PvtSimulationResults != nil {
+		// prototype - list collections that are read from (in order to populate NoPublicReads in ChaincodeInterest)
+		for _, prws := range simResult.PvtSimulationResults.GetNsPvtRwset() {
+			for _, cprws := range prws.GetCollectionPvtRwset() {
+				ccInterest.Chaincodes = append(ccInterest.Chaincodes, &pb.ChaincodeCall{
+					Name:            prws.GetNamespace(),
+					CollectionNames: []string{cprws.GetCollectionName()},
+					NoPrivateReads:  !simResult.PrivateReads.Read(prws.Namespace, cprws.CollectionName),
```
manish-sethi: Just a clarification question - so multiple entries for a single chaincode are expected - one for each collection? I figured that though there a slice for specifying multiple collection names but only single flag for `NoPrivateReads` - so is this a workaround to overcome that?

core/endorser/endorser.go
```diff
@@ -500,3 +507,48 @@ func CreateCCEventBytes(ccevent *pb.ChaincodeEvent) ([]byte, error) {
...
+				})
+				chaincodes[prws.GetNamespace()] = struct{}{}
+			}
+		}
+	}
+	// if there are any chaincodes in the public RWset that are not in the private set, add them to the interest
+	if simResult.PubSimulationResults != nil {
+		for _, nsrws := range simResult.PubSimulationResults.GetNsRwset() {
+			if nsrws.GetNamespace() == "_lifecycle" {
+				continue
```
manish-sethi: At other places we avoided this hardcoding via [this function](https://github.com/hyperledger/fabric/blob/9a922fd9c78c01d185bc2b45fb98012a37f7bfb9/core/chaincode/lifecycle/deployedcc_infoprovider.go#L52). see if you could use this here easily. Otherwise, can introduce a function in that package that returns these names - which would cover the name for legacy lifecycle name as well (i.e., `lscc`).

core/endorser/endorser.go
```diff
@@ -500,3 +507,48 @@ func CreateCCEventBytes(ccevent *pb.ChaincodeEvent) ([]byte, error) {
...
+		}
+	}
+	if len(ccInterest.Chaincodes) > 0 {
+		for _, policyBytes := range simResult.KeySignaturePolicies {
+			policy, err := protoutil.UnmarshalSignaturePolicy(policyBytes)
+			if err != nil {
+				return nil, err
+			}
+			ccInterest.Chaincodes[0].KeyPolicies = append(ccInterest.Chaincodes[0].KeyPolicies, policy)
+		}
```
manish-sethi, line 548: Perhaps it won't make any difference functionally but just to confirm that you deliberately mention all the SBE in the first entry for the chaincode?

core/ledger/kvledger/txmgmt/txmgr/tx_simulator.go
andrew-coleman: Yes, I've put a comment in the code explaining.  There is an existing unit test that catches this.

core/endorser/endorser.go
andrew-coleman: Yes exactly.  I've added a comment in the code to this effect

core/endorser/endorser.go
andrew-coleman: Yes, in reality it makes no difference where they are

core/endorser/endorser.go
```diff
@@ -458,9 +464,56 @@ func (e *Endorser) ProcessProposalSuccessfullyOrError(up *UnpackedProposal) (*pb
...
+				// Since each collection in a chaincode could have different values of the NoPrivateReads flag, create a new Chaincode entry for each.
+				ccInterest.Chaincodes = append(ccInterest.Chaincodes, &pb.ChaincodeCall{
+					Name:            prws.GetNamespace(),
+					CollectionNames: []string{cprws.GetCollectionName()},
+					NoPrivateReads:  !simResult.PrivateReads.Exists(prws.Namespace, cprws.CollectionName),
+				})
+				chaincodes[prws.GetNamespace()] = struct{}{}
+			}
+		}
+	}
```
manish-sethi: I missed noticing in this block earlier
- Even in the private read-write, we should skip the _lifecycle namespace (can skip IsSysCC(nsrws.GetNamespace())) -- As the client would specifically target the orgs for these transactions.
- This logic assumes that if a collection is read from, it is written to as well. So, the collections that are only present in the `PrivateReads` and not in the pvt-write-set will get a miss.

manish-sethi (commented):
A few more comments

core/endorser/endorser.go
andrew-coleman: Good catch.  This logic was left over from my prototype which was building a full private read set, so this assumption would have held.  I've now rewritten this entirely in a (hopefully) clearer way.  And added an extra test case for this situation.

core/endorser/endorser.go
```diff
@@ -458,9 +464,92 @@ func (e *Endorser) ProcessProposalSuccessfullyOrError(up *UnpackedProposal) (*pb
...
+			for _, cprws := range prws.GetCollectionPvtRwset() {
+				collections[cprws.CollectionName] = false
+			}
+			chaincodes[prws.Namespace] = collections
+		}
+	}
+
+	// 3. Add/merge collections from the PrivateReads struct
+	for ns, entry := range simResult.PrivateReads {
+		if cc := chaincodes[ns]; cc == nil {
```
manish-sethi, line 510: - Skip systems chaincode in this block as well
- Add a test that asserts that the system chaincodes are indeed skipped

manish-sethi (approved):
LGTM. Just left a pretty minor comment. I am fine merging this as is and you can fix the comments in a separate PR.

PR #2769, opened 2021-07-20
[FAB-11334] Adds a function to purge a ledger's transient storage

### LINKED ISSUE #2786: Add a function to purge a ledger's transient storage

jkneubuh:
#### Type of change

- New feature

#### Description

This introduces a transientstore.Delete(ledgerID) routine to assist with the bookkeeping while unjoining a peer from a channel.  Before removing the transient storage, the store is marked as UNDER_DELETION in a system leveldb, providing an opportunity to recover from a failed deletion at the next peer initialization.

#### Related issues

- [FAB-11334](https://jira.hyperledger.org/browse/FAB-11334)

Resolves #2786


core/transientstore/store.go
```diff
@@ -98,6 +114,173 @@ func (provider *storeProvider) Close() {
...
+func Drop(providerPath, ledgerID string) error {
+	logger.Infof("Dropping ledger [%s] from transient storage path [%s]", ledgerID, providerPath)
+
+	// Ensure the routine is invoked while the peer is down.
+	lockPath := filepath.Join(providerPath, "fileLock")
+	fileLock := leveldbhelper.NewFileLock(lockPath)
+	if err := fileLock.Lock(); err != nil {
+		return errors.WithMessage(err, "as another peer node command is executing,"+
+			" wait for that command to complete its execution or terminate it before retrying")
+	}
```
manish-sethi: This will require to get this lock at peer start as well - i.e., inside the function `NewStoreProvider`.

core/transientstore/delete_store_test.go
```diff
@@ -0,0 +1,291 @@
+/*
+Copyright IBM Corp. All Rights Reserved.
+
+SPDX-License-Identifier: Apache-2.0
+*/
+
+package transientstore
+
```
manish-sethi, line 8: - Can you add one more test that mimics the crash-recovery (i.e., automatically delete at start)
- Can you see if you can cover some of the error paths easily by directly inserting some data into the db, or by closing the db provider at some occasions.
- I feel that some of the tests can be merged into one and have them as a table-driven sub-tests pattern.

core/transientstore/delete_store_test.go
```diff
@@ -0,0 +1,291 @@
...
+	dl, err = sp.getStorageMarkedForDeletion()
+	require.NoError(t, err)
+	require.Equal(t, 2, len(dl))
+	require.Contains(t, dl, doomed1)
+	require.NotContains(t, dl, doomed2)
+	require.Contains(t, dl, doomed3)
+}
+
+func TestClearDeletionTagNotPresent(t *testing.T) {
```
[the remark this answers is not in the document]
jkneubuh: Whoops!   This test was supposed to be checking the boundary cases of adding an invalid ledger to both an empty and populated set of pending deletions.  The magical ledger must have been created by a race condition in the middle of a Josh context switch.    Thanks Manish, good catch.

core/transientstore/delete_store_test.go
jkneubuh: OK.

- `TestProviderRestartAfterFailedDeletionScrubsPendingDeletions` runs a simulated test of the crash recovery at peer (NewProvider) init. 
- The tests do insert some records into the transient storage - are there additional insertions / patterns in addition to the `populateTestStore()` ?

core/transientstore/store.go
jkneubuh: This was a little annoying but there is a bootstrapping issue trying to construct the provider / lock combination.  My solution was to grab the lock at the start of `Drop` (ensure we're not running at the same time as the peer), immediately release the lock, and make the call to NewProvider, where the lock is re-acquired.   

This opens a few nano/pico seconds of time during which the peer may launch concurrently with three simultaneous lightning strikes - for the rough case of ensuring the peer is down while Dropping the transient storage, it should be OK.

core/transientstore/store.go
```diff
@@ -98,6 +114,173 @@ func (provider *storeProvider) Close() {
...
+
+	return nil
+}
+
+// Write the UNDER_DELETE ledger set as a JSON array in the transient system catalog.
+func (provider *storeProvider) markStorageListForDelete(storageList []string) error {
+	json, err := json.Marshal(storageList)
+	if err != nil {
+		return errors.WithMessagef(err, "marshalling delete list")
+	}
```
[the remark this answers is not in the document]
jkneubuh: Latest commit converts from json -> `PendingDeleteStorageList` proto.Message

core/transientstore/store.go
manish-sethi: A cleaner approach to this could be to have a separate unexported function `newStoreProvider` that assumes that the caller has acquired the lock and the have this call from both the functions `NewStoreProvider` and `Drop` after having acquired the lock.

common/ledger/util/leveldbhelper/leveldb_helper.go
```diff
@@ -205,16 +205,25 @@ func (f *FileLock) Lock() error {
...
+	f.db = db
+
 	return nil
 }
 
+// Determine if the lock is currently held open.
+func (f *FileLock) IsLocked() bool {
+	return f.db != nil
+}
+
```
manish-sethi, line 226: I won't over-emphasize this but just to take a note - as the introduction of this function now assumes (at this level of code) that this struct can be used by a more than one go routines, access to field `f.db` should be protected.  Else, a caller can get a wrong result from this function. 

However, we are using this struct in a very limited way and don't foresee any plan to use this function in a test with concurrent go-routines, I am fine leaving it as is; but appreciate if you want to fix this for cleanliness at this level of the code.

core/transientstore/store.go
jkneubuh: The code was grabbing the lock at the wrong path.  Thanks Manish for your help tracking this down.   The new rev in this PR will set up the file lock at `${transientstore}/../fileLock`

common/ledger/util/leveldbhelper/leveldb_helper_test.go
```diff
@@ -157,6 +157,30 @@ func TestFileLock(t *testing.T) {
 	require.NoError(t, os.RemoveAll(fileLockPath))
 }
 
+func TestFileLockLockUnlockLock(t *testing.T) {
+	// create an open lock
+	lockPath := testDBPath + "/fileLock"
+	lock := NewFileLock(lockPath)
```
manish-sethi, line 163: The dir `lockPath` not cleaned up after the test.  In a `defer` you can release the lock and delete the dir

core/transientstore/store.go
```diff
@@ -80,11 +86,45 @@ type RwsetScanner struct {
 
 // NewStoreProvider instantiates TransientStoreProvider
 func NewStoreProvider(path string) (StoreProvider, error) {
-	dbProvider, err := leveldbhelper.NewProvider(&leveldbhelper.Conf{DBPath: path})
+	// Ensure the routine is invoked while the peer is down.
+	lockPath := filepath.Join(filepath.Dir(path), "fileLock")
```
manish-sethi: `transientStoreFileLock` would be a clearer name now - as this dir will directly be created inside `/var/production`, it may be hard to decipher by looking at the dir structure about the context of this dir.

core/transientstore/store.go
```diff
@@ -95,7 +135,179 @@ func (provider *storeProvider) OpenStore(ledgerID string) (*Store, error) {
...
+	}
+	if provider.fileLock != nil {
+		provider.fileLock.Unlock()
+	}
+}
+
+// DeleteStore deletes the TransientStore for a given ledger.
+func (provider *storeProvider) DeleteStore(ledgerID string) error {
+	return provider.dbProvider.Drop(ledgerID)
+}
```
manish-sethi: This seems unnecessarily exported and adding to the interface and causing the mock regenration. I figured that you use this in test. For that you can simply cast to type `*storeProvider` in the test.

core/transientstore/store.go
```diff
@@ -80,11 +86,45 @@ type RwsetScanner struct {
...
+	lock := leveldbhelper.NewFileLock(lockPath)
+	if err := lock.Lock(); err != nil {
+		return nil, errors.WithMessage(err, "as another peer node command is executing,"+
+			" wait for that command to complete its execution or terminate it before retrying")
+	}
+
+	provider, err := newStoreProvider(path, lock)
 	if err != nil {
-		return nil, err
+		return nil, errors.WithMessagef(err, "could not construct storage provider in folder [%s]", path)
```
manish-sethi: For code sanity you could add `lock.Unlock()` before returning error -- as at this level of code it's not fair to assume that the consumer may not retry to reopen the store.

manish-sethi (commented):
LGTM now. A few final clean up comments.

core/transientstore/store.go
jkneubuh: 👍

core/transientstore/store.go
jkneubuh: I agree 110%

core/transientstore/store.go
jkneubuh: got it.

jkneubuh:
OK.  I think this one is a wrap.  I agree we should look to find a wider grain / common lock to handle "peer is running" concurrently with CLI activities requiring the peer to be shut down.

common/ledger/util/leveldbhelper/leveldb_helper_test.go
```diff
@@ -157,6 +157,33 @@ func TestFileLock(t *testing.T) {
...
+	require.Nil(t, lock.db)
+	require.Equal(t, lock.filePath, lockPath)
+	require.False(t, lock.IsLocked())
+
+	defer lock.Unlock()
+	defer os.RemoveAll(lockPath)
```
manish-sethi, line 169: I think that your intent was the following. The `defers` are executed in reverse order.
```suggestion
	defer func(){
	     lock.Unlock()
	     os.RemoveAll(lockPath)
	}()
```

manish-sethi (approved):
Approving with a minor comment that can be fixed later.

PR #2773, opened 2021-07-22
FABGW-25 Endorse using generated ChaincodeInterest

andrew-coleman:
Change the Endorse() method logic as described in the design document attached to the Jira.
- Process proposal on local org peer.
- Using the ChaincodeInterest in the propsal response, invoke discovery to obtain a layout that contains the local org
- Obtain the remaining endorsements
- Build the transaction to send back to the client.

Extra unit tests added to test the new logic.

### DIFF
--- a/internal/pkg/gateway/api.go
+++ b/internal/pkg/gateway/api.go
@@ -32,9 +32,14 @@ func (gs *Server) Evaluate(ctx context.Context, request *gp.EvaluateRequest) (*g
 		return nil, status.Error(codes.InvalidArgument, "an evaluate request is required")
 	}
 	signedProposal := request.GetProposedTransaction()
-	channel, chaincodeID, err := getChannelAndChaincodeFromSignedProposal(signedProposal)
+	channel, chaincodeID, _, err := getChannelAndChaincodeFromSignedProposal(signedProposal)
 	if err != nil {
 		return nil, status.Errorf(codes.InvalidArgument, "failed to unpack transaction proposal: %s", err)
+	}
+
+	err = gs.registry.registerChannel(channel)
+	if err != nil {
+		return nil, status.Errorf(codes.Unavailable, "%s", err)
 	}
 
 	endorser, err := gs.registry.evaluator(channel, chaincodeID, request.GetTargetOrganizations())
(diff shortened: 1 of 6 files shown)


internal/pkg/gateway/api.go
```diff
@@ -86,24 +91,90 @@ func (gs *Server) Endorse(ctx context.Context, request *gp.EndorseRequest) (*gp.
...
+		if firstResponse.Response.Status < 200 || firstResponse.Response.Status >= 400 {
+			return nil, rpcError(codes.Aborted, "failed to endorse transaction", endpointError(firstEndorser, fmt.Errorf("error %d, %s", firstResponse.Response.Status, firstResponse.Response.Message)))
+		}
+
+		// 3. Extract ChaincodeInterest and SBE policies
+		// The chaincode interest could be nil for legacy peers and for chaincode functions that don't produce a read-write set
+		interest := firstResponse.Interest
+		if len(interest.GetChaincodes()) == 0 {
+			interest = defaultInterest
+		}
```
manish-sethi, line 158: A clarification question - Will this condition be ever true unless it's a read-only transaction? If no, then, is there any purpose of proceeding from here? 

Also, that brings another thought, what if it's read-only transaction for pvtdata? That would record some entries in the interest which would give a wrong signal for proceeding.

internal/pkg/gateway/api.go
```diff
@@ -86,24 +91,90 @@ func (gs *Server) Endorse(ctx context.Context, request *gp.EndorseRequest) (*gp.
...
+				endorsers = append(endorsers[:i], endorsers[i+1:]...)
+				responses = append(responses, firstResponse)
+				break
+			}
+		}
+
+		// 5. If transient data is involved, and other (non-local) orgs are required to endorse, return error for privacy reasons
+		if hasTransientData && len(endorsers) > 0 {
+			return nil, status.Error(codes.FailedPrecondition, "non-local organizations are required to endorse; retry specifying endorsing organization(s) to protect transient data")
+		}
```
manish-sethi: Not sure if this caution is need here at this stage. Won't this make this whole approach pretty much unusable with private data as in most of the scenario private data transactions will involve transient data and would require multi-org endorsements.

May be better to set the artificial pvt-reads set for all collections to narrow the scope?

internal/pkg/gateway/api.go
```diff
@@ -86,24 +91,90 @@ func (gs *Server) Endorse(ctx context.Context, request *gp.EndorseRequest) (*gp.
...
+		// 1. Choose an endorser from the gateway's organization
+		var firstEndorser *endorser
+		es, ok := gs.registry.endorsersByOrg(channel, chaincodeID)[gs.registry.localEndorser.mspid]
+		if !ok {
+			// No local org endorsers for this channel/chaincode. If transient data is involved, return error
+			if hasTransientData {
+				return nil, status.Error(codes.FailedPrecondition, "no endorsers found in the gateway's organization; retry specifying endorsing organization(s) to protect transient data")
+			}
+			// Otherwise, just let discovery pick one.
+			endorsers, err = gs.registry.endorsers(channel, defaultInterest, "")
```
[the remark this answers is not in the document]
andrew-coleman: Yes, we need to consider a future proposal for allowing the client to choose the 'initial' endorser

internal/pkg/gateway/api.go
andrew-coleman: This will be true if:
- The Interest structure in the ProposalResponse is nil, which will be the case for legacy peers
- The chaincode invocation doesn't create a RWset.  This happens in one of the integration tests where the chaincode only sets an event

internal/pkg/gateway/api.go
andrew-coleman: Sure, I've implemented your suggestion

manish-sethi (approved):

PR #2780, opened 2021-07-23
FAB 18365 IT flake evictionsuspector failed to read eviction 

Param-S:
#### Description
In failing case, the OSN being removed has committed the config change but failed to receive apply_config msg to halt the chain. This state is defined to be handled as part of the eviction suspicion logic. But it failed when evictionsuspector failed to check from lastconfig block if chain is up-to-date. 

Fix: Updated the order of statements which looks at the latest config block and block height validation.

#### Type of change
- Bug fix

### DIFF
--- a/orderer/consensus/etcdraft/eviction.go
+++ b/orderer/consensus/etcdraft/eviction.go
@@ -113,13 +113,6 @@ func (es *evictionSuspector) confirmSuspicion(cumulativeSuspicion time.Duration)
 
 	es.logger.Infof("Last config block was found to be block [%d]", lastConfigBlock.Header.Number)
 
-	height := es.height()
-
-	if lastConfigBlock.Header.Number+1 <= height {
-		es.logger.Infof("Our height is higher or equal than the height of the orderer we pulled the last block from, aborting.")
-		return
-	}
-
 	err = es.amIInChannel(lastConfigBlock)
 	if err != cluster.ErrNotInChannel && err != cluster.ErrForbidden {
 		details := fmt.Sprintf(", our certificate was found in config block with sequence %d", lastConfigBlock.Header.Number)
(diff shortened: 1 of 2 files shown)


orderer/consensus/etcdraft/eviction.go
```diff
@@ -139,6 +132,12 @@ func (es *evictionSuspector) confirmSuspicion(cumulativeSuspicion time.Duration)
 	es.halted = true
 	es.logger.Infof("Chain has been halted, pulling remaining blocks up to (and including) eviction block.")
```
yacovm, line 133: Please move this:

`es.logger.Infof("Chain has been halted, pulling remaining blocks up to (and including) eviction block.")`

To an else case of the if clause you moved, otherwise this is confusing - it says it is pulling the remaining blocks and then changes its mind about it later on.

orderer/consensus/etcdraft/eviction.go
```diff
@@ -139,6 +132,12 @@ func (es *evictionSuspector) confirmSuspicion(cumulativeSuspicion time.Duration)
...
 
+	height := es.height()
+	if lastConfigBlock.Header.Number+1 <= height {
+		es.logger.Infof("Our height is higher or equal than the height of the orderer we pulled the last block from, aborting.")
+		return
+	}
```
yacovm, line 137:
```suggestion
	} else {
	es.logger.Infof("Chain has been halted, pulling remaining blocks up to (and including) eviction block.")
	}
```

orderer/consensus/etcdraft/eviction.go
yacovm:
```suggestion

```

yacovm:
Please move this:

`es.logger.Infof("Chain has been halted, pulling remaining blocks up to (and including) eviction block.")`

To an else case of the if clause you moved, otherwise this is confusing - it says it is pulling the remaining blocks and then changes its mind about it later on.

Param-S:
Yes, I moved it down now.

yacovm (approved):

orderer/consensus/etcdraft/eviction_test.go
```diff
@@ -194,6 +186,15 @@ func TestEvictionSuspector(t *testing.T) {
 			height:                     9,
 			halt:                       t.Fail,
 		},
+		{
+			description:                "our height is the highest",
```
guoger, line 190: nit - this description probably should be updated to reflect actual semantics

guoger: should this be back-ported to 1.4? (which i know is now EOL, but i saw commits are still being added to it...)

yacovm: I don't think this is important. You can get by without this fix.

PR #2784, opened 2021-07-26
FAB-18067 Discovery support Implicit Collections

andrew-coleman:
Implemented the fix as suggested by Yacov in his Jira comment.

Verified the fix works against the asset-transfer-secured-agreement sample which uses implicit collections.

### DIFF
--- a/core/chaincode/lifecycle/metadata_provider.go
+++ b/core/chaincode/lifecycle/metadata_provider.go
@@ -13,6 +13,8 @@ import (
 	"github.com/hyperledger/fabric-protos-go/peer"
 	"github.com/hyperledger/fabric/common/chaincode"
 	"github.com/hyperledger/fabric/common/policies"
+	"github.com/hyperledger/fabric/common/policydsl"
+	"github.com/hyperledger/fabric/core/chaincode/implicitcollection"
 	"github.com/hyperledger/fabric/protoutil"
 	"github.com/pkg/errors"
 )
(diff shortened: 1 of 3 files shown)


core/chaincode/lifecycle/metadata_provider.go
```diff
@@ -129,6 +131,30 @@ func (mp *MetadataProvider) Metadata(channel string, ccName string, collections
...
+
+			ccInfo.Definition.Collections.Config = append(ccInfo.Definition.Collections.Config, &peer.CollectionConfig{
+				Payload: &peer.CollectionConfig_StaticCollectionConfig{
+					StaticCollectionConfig: &peer.StaticCollectionConfig{
+						Name: col,
+						MemberOrgsPolicy: &peer.CollectionPolicyConfig{
+							Payload: &peer.CollectionPolicyConfig_SignaturePolicy{
+								SignaturePolicy: policydsl.SignedByMspMember(mspID),
+							},
+						},
```
manish-sethi, line 157: Shouldn't you add here `EndorsementPolicy` field as well which would carry the same value as for the `MemberOrgsPolicy`. Look at the following `for` loop that populates the map `ccMetadata.CollectionPolicie` with `EndorsementPolicy`.

manish-sethi: I feel that this eventually requires an integration test somewhere.

andrew-coleman: I've added a test to the discovery integration suite

manish-sethi: Thanks @andrew-coleman for adding the integration test. 

One more point - I am not sure what was the motivation but it appears that we also allow the implicit collection policy to be [configured via channel config](https://github.com/hyperledger/fabric/blob/240cf0e182938a09832ba4162c3c229c81848e3f/core/chaincode/lifecycle/deployedcc_infoprovider.go#L266).

I feel that moving the existing code in a separate function and invoking that function from both the existing code and your code may allow to keep this important logic at one place.

manish-sethi: Merging this as is for now. Opened a new [issue](https://github.com/hyperledger/fabric/issues/2798) so it can be tracked.

manish-sethi (approved):

PR #2799, opened 2021-07-28
Gateway enabled by default

andrew-coleman:
In readiness for the Beta release, set the ‘Enabled’ config option of the gateway to ‘true’.
All peers will now have the gateway server enabled unless it is explicitly disbaled.

### DIFF
--- a/integration/gateway/endorsing_orgs_test.go
+++ b/integration/gateway/endorsing_orgs_test.go
@@ -45,8 +45,6 @@ var _ = Describe("GatewayService with endorsing orgs", func() {
 		config := nwo.ThreeOrgRaft()
 		network = nwo.New(config, testDir, client, StartPort(), components)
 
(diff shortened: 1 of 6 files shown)


jt-nti (dismissed):

denyeart (commented):
I'm fine with the default of enabled, but can you add a core.yaml config option to drive it?

jt-nti:
@andrew-coleman it might be worth updating the integration tests so it's enabled by default there as well...

https://github.com/hyperledger/fabric/blob/main/integration/nwo/network.go#L205

Then you wouldn't need these...

https://github.com/hyperledger/fabric/blob/main/integration/gateway/endorsing_orgs_test.go#L48
https://github.com/hyperledger/fabric/blob/main/integration/gateway/gateway_test.go#L54

denyeart (approved):

PR #2805, opened 2021-08-05
Options for GRPC message size configurable

### LINKED ISSUE #2804: Make grpc max message size configurable
grpc max message size is hardcoded to 100MB for both grpc clients and servers, in both peer and orderer.

Max messages size should be configurable in peer core.yaml and orderer.yaml (right next to the keepalive config options).

To see where max message size is set in the code, grep for DefaultMaxRecvMsgSize and DefaultMaxSendMsgSize.

More history can be found in https://jira.hyperledger.org/browse/FAB-6461

Param-S:
#### Type of change
- Improvement 

#### Description
- Configurable options to update GRPC max message size.  Options added both in peer and orderer node configuration.

#### Related issues
Closes #2804 

### DIFF
--- a/orderer/common/localconfig/config.go
+++ b/orderer/common/localconfig/config.go
@@ -15,6 +15,7 @@ import (
 	"github.com/hyperledger/fabric/common/flogging"
 	"github.com/hyperledger/fabric/common/viperutil"
 	coreconfig "github.com/hyperledger/fabric/core/config"
+	"github.com/hyperledger/fabric/internal/pkg/comm"
 )
 
 var logger = flogging.MustGetLogger("localconfig")
@@ -49,6 +50,8 @@ type General struct {
 	LocalMSPID        string
 	BCCSP             *bccsp.FactoryOpts
 	Authentication    Authentication
+	MaxRecvMsgSize    int32
+	MaxSendMsgSize    int32
 }
 
 type Cluster struct {
@@ -240,6 +243,8 @@ var Defaults = TopLevel{
 		Authentication: Authentication{
 			TimeWindow: time.Duration(15 * time.Minute),
 		},
+		MaxRecvMsgSize: comm.DefaultMaxRecvMsgSize,
+		MaxSendMsgSize: comm.DefaultMaxSendMsgSize,
 	},
 	FileLedger: FileLedger{
 		Location: "/var/hyperledger/production/orderer",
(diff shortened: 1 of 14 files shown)


orderer/common/localconfig/config.go
```diff
@@ -212,6 +214,12 @@ type ChannelParticipation struct {
 	MaxRequestBodySize uint32
 }
 
+// Message size for the GRPC server and client.
+type Message struct {
```
jyellick: We already have connection configuration elements at the top level -- Keepalive, ConnectionTimeout -- any reason to push this into its own structure instead of hanging them off the top level?  As written, the name stutters as "Message.MaxRecvMsgSize".

jyellick: That makes more sense to me. Thanks!

sampleconfig/core.yaml
```diff
@@ -468,6 +468,11 @@ peer:
             # deliverService limits concurrent event listeners registered to deliver service for blocks and transaction events.
             deliverService: 2500
 
+    # Max message size in bytes GRPC server and client can receive
```
denyeart, line 473:
```suggestion
    # Max message size in bytes GRPC server and client can receive.
    # Since all nodes should be consistent it is recommended to keep the default value of 100MB.
```

sampleconfig/core.yaml
```diff
@@ -468,6 +468,11 @@ peer:
...
+    maxRecvMsgSize: 104857600
+    # Max message size in bytes GRPC server and client can send
```
denyeart, line 475:
```suggestion
    # Max message size in bytes GRPC server and client can send.
    # Since all nodes should be consistent it is recommended to keep the default value of 100MB.
```

sampleconfig/orderer.yaml
```diff
@@ -50,6 +50,12 @@ General:
         # ServerTimeout is the duration the server waits for a response from
         # a client before closing the connection.
         ServerTimeout: 20s
+
+    # Max message size in bytes the GRPC server and client can receive
```
denyeart, line 56:
```suggestion
    # Max message size in bytes the GRPC server and client can receive.
    # Since all nodes should be consistent it is recommended to keep the default value of 100MB.
```

sampleconfig/orderer.yaml
```diff
@@ -50,6 +50,12 @@ General:
...
+    MaxRecvMsgSize: 104857600
+    # Max message size in bytes the GRPC server and client can send
```
denyeart, line 58:
```suggestion
    # Max message size in bytes the GRPC server and client can send.
    # Since all nodes should be consistent it is recommended to keep the default value of 100MB.
```

denyeart (commented):
I'd also put a comment in configtx.yaml AbsoluteMaxBytes setting, stating that increasing the value beyond the default of 10MB may cause issues depending on networking infrastructure, and to consider 49MB an upper limit to account for expansion and the node grpc limit of 100MB.

https://github.com/hyperledger/fabric/blob/main/sampleconfig/configtx.yaml#L294

denyeart (approved):

denyeart:

PR #2822, opened 2021-08-10
Add slash command for invalid issue

lindluni:
Adds a "/invalid" command that when typed in an issue will let users know the purpose of GitHub issues and close the issue.

This can be used when a user opens an issue seeking help, rather than to report a bug or request a feature. Simply typing `/invalid` will reply to the issue with the below comment and automatically close the issue:

```
Thank you for opening this issue.

GitHub Issues is a tool for tracking bugs, feature requests, and work in general that 
relates directly to the Fabric codebase. It is not for general help requests. Please 
use one of the following forums to request help for your issue:

- RocketChat: https://chat.hyperledger.org
- Fabric Mailing List: fabric@lists.hyperledger.org
```

### DIFF
--- a/.github/workflows/slash-commands.yml
+++ b/.github/workflows/slash-commands.yml
@@ -1,0 +1,31 @@
+# Copyright the Hyperledger Fabric contributors. All rights reserved.
+#
+# SPDX-License-Identifier: Apache-2.0
+
+name: Slash Commands
+on:
+  issue_comment:
+    types:
+      - created
+      - edited
+jobs:
+  notify:
+    name: Invalid Issue Usage
+    if: contains(github.event.comment.body, '/invalid') && github.event.issue.state == 'open'
+    runs-on: ubuntu-latest
+    steps:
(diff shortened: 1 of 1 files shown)


denyeart:
@lindluni any idea why /ci-run isn't working?
I've re-triggered the failing test in Azure Pipelines (BTW, the orderer unit tests flakes are next to be improved... I think we've solved the orderer integration test flakes now... don't give up hope!)

denyeart (approved):

denyeart:
I've tested the invalid function with https://github.com/hyperledger/fabric/issues/2831 and it works, thanks @lindluni !
Now to figure out why ci-run doesn't work.

lindluni:
@denyeart 

My Azure DevOps token expired :(

I tried to generate a new one but my credentials are screwed up because of my Microsoft account. If you generate a new token in AZP and then replace the existing AZP_TOKEN in the Settings -> Secrets here in the repo that should fix it

PR #2824, opened 2021-08-11
Fix FAB-18528: remove panic in ifConfig func

sardChen:
Fix issues: FAB-18528. When received the constructed message from the malicious node (through the interface "chain.rpc.SendSubmit(dest uint64, request *orderer.SubmitRequest, report func(err error))"), all orderers will breakdown immediately.


yacovm:
Code looks good but you need to [append your signed-off](https://github.com/apps/dco).

Do `git commit -s --amend`

PR #2847, opened 2021-08-19
[FAB-11334] - Adds a functional / integration test for peer unjoin 

jkneubuh:
This integration test launches the peer unjoin command, simulating normal operation and some error / negative test cases.

#### Type of change
- Test update

#### Description

This integration test targets the `peer node unjoin` CLI, both with positive and negative examples: 

- checks for error when unjoining a channel while the peer is running.
- checks for double-unjoin (unjoin unjoined)
- checks for partial unjoin:  If the unjoin fails, the next peer restart will attempt to resume the ledger purge.
- checks for normal unjoin

### DIFF
--- a/integration/ledger/reset_rollback_test.go
+++ b/integration/ledger/reset_rollback_test.go
@@ -28,7 +28,7 @@ import (
 	"github.com/tedsuo/ifrit"
 )
 
-var _ = Describe("rollback, reset, pause and resume peer node commands", func() {
+var _ = Describe("rollback, reset, pause, resume, and unjoin peer node commands", func() {
 	// at the beginning of each test under this block, we have defined two collections:
 	// 1. collectionMarbles - Org1 and Org2 have access to this collection
 	// 2. collectionMarblePrivateDetails - Org2 and Org3 have access to this collection
@@ -73,7 +73,7 @@ var _ = Describe("rollback, reset, pause and resume peer node commands", func() 
 		By("creating 5 blocks")
 		for i := 1; i <= 5; i++ {
 			helper.addMarble("marblesp", fmt.Sprintf(`{"name":"marble%d", "color":"blue", "size":35, "owner":"tom", "price":99}`, i), org2peer0)
-			helper.waitUntilEqualLedgerHeight(height + i)
+			helper.waitUntilAllPeersEqualLedgerHeight(height + i)
 		}
 
 		By("verifying marble1 to marble5 exist in collectionMarbles & collectionMarblePrivateDetails in Org2.peer0")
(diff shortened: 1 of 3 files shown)


denyeart (dismissed):
nice!

integration/ledger/reset_rollback_test.go
```diff
@@ -481,3 +544,15 @@ func (th *testHelper) assertPausedChannel(peer *nwo.Peer) {
...
+func (th *testHelper) assertUnjoinedChannel(peer *nwo.Peer) {
+	sess, err := th.PeerUserSession(peer, "User1", commands.ChannelInfo{
+		ChannelID: th.channelID,
+	})
+	Expect(err).NotTo(HaveOccurred())
+	Eventually(sess, th.EventuallyTimeout).Should(gexec.Exit(1))
+	Expect(sess.Err).To(gbytes.Say("Invalid chain ID"))
+
+	channelLedgerDir := filepath.Join(th.Network.PeerLedgerDir(peer), "chains/chains", th.channelID)
+	Expect(channelLedgerDir).NotTo(BeADirectory())
```
manish-sethi, line 572: While it does not harm to check for lower levels such as this, it's better to execute a channel specific command to and expect to receive an error.

integration/ledger/reset_rollback_test.go
```diff
@@ -191,6 +191,57 @@ var _ = Describe("rollback, reset, pause and resume peer node commands", func()
...
+		helper.unjoin(org3peer0, expectedErrMessage, false)
+		Expect(os.Chmod(ledgerStoragePath, 0o755)).NotTo(HaveOccurred())
+
+		By("restarting peers")
+		setup.startPeer(org1peer0)
+		setup.startPeer(org2peer0)
+		setup.startPeer(org3peer0)
+
+		helper.assertUnjoinedChannel(org1peer0)
+		helper.assertUnjoinedChannel(org3peer0)
```
manish-sethi, line 243: Would it be easy to add after this -- for making one of these peers to join the channel again?

integration/ledger/reset_rollback_test.go
jkneubuh: Agree.  At least in this case it was a bit annoying to have the unjoin halt immediately while the peer is running.  (There's at least a little coverage for the error case with the "double unjoin.")  This is a small motivating case but it does seem like ALL of the `peer` commands should be executed against a running instance - in the case of cloud--native fabric deployments, it's a challenge to run things "next to" where the peer would be running.

denyeart (approved):
nicer!

PR #2852, opened 2021-08-24
2 issues - "Unrecognized archive format" and tag version mismatch

WhitneyMFoster:
#### Type of change

- Bug fix

#### Description

Modified script to look for v2.2.3 in fabric-samples and v2.3.2 in fabric. Assuming fabric-samples was tagged incorrectly. Also, the archive couldn't be read and the addition of -O to command solved the issue.

Motivation from getting "Unrecognized archive format," digging through the script to debug led me to realizing the fabric/fabric-samples versions were tagged incorrectly.

### DIFF
--- a/scripts/bootstrap.sh
+++ b/scripts/bootstrap.sh
@@ -7,6 +7,7 @@
 
 # if version not passed in, default to latest released version
 VERSION=2.3.2
+SAMPLE_VERSION_BANDAID=2.2.3
 # if ca version not passed in, default to latest released version
 CA_VERSION=1.5.1
 ARCH=$(echo "$(uname -s|tr '[:upper:]' '[:lower:]'|sed 's/mingw64_nt.*/windows/')-$(uname -m | sed 's/x86_64/amd64/g')")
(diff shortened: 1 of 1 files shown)


denyeart:
Thank you for the interest and contribution.
The download script is actually working as intended though, since recent versions are intentionally not tagged in fabric-samples so that users of recent versions can always get the latest fabric-samples. I've opened a PR to clarify the message and explain the rationale, please see https://github.com/hyperledger/fabric/pull/2855.

denyeart:
Concerning -O , we have not heard this from other users, but you could open another PR for it and explain the rationale. Thank you.

WhitneyMFoster:
Thank you for the response and clarification in the comments but I had a different reason I should have included in this PR. Here's my issue and how this pr unblocked me:
 
I was going through the [Hyperledger Fabric installation instructions](https://hyperledger-fabric.readthedocs.io/en/latest/install.html). 
For the step that runs `curl -sSL https://bit.ly/2ysbOFE | bash -s`, I was blocked with this error:
```
===> Downloading:  https://github.com/hyperledger/fabric/releases/download/v2.3.2/hyperledger-fabric-darwin-arm64-2.3.2.tar.gz
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100     9  100     9    0     0     17      0 --:--:-- --:--:-- --:--:--    17
tar: Error opening archive: Unrecognized archive format
==> There was an error downloading the binary file.

------> 2.3.2 platform specific fabric binary is not available to download <----
```
So the issue is originating from `download`, not `cloneSamplesRepo` function so that bandaid variable is pointless - my blunder

The thing that fixes my issue is adding the -O on line 79 in the `download` function. Is there a step I may be missing that will make this script work without adding the -O? @denyeart

denyeart:
-O actually breaks the download for me with the following error:
```
gzip: (stdin): unexpected end of file
tar: Child returned status 1
tar: Error is not recoverable: exiting now
```

Not sure why the original does not work for you...

PR #2857, opened 2021-08-25
Implement DisregardNamespacePolicy for gateway

### LINKED ISSUE #2856: Support 'DisregardNamespacePolicy' hint in Gateway
The Gateway automatically generates the ChaincodeInterest structure to pass to the discovery service to find the optimal set of endorsers.  Following the gateway's implementation of this, an extra flag was added to this structure which instructs the discovery service to ignore the chaincode policy if all the state changes are validated by key (SBE) policies.  See https://github.com/hyperledger/fabric/commit/e5e623d92b74772f30246ecd0ebfd0274ae4bae2.

In this story, the ChaincodeInterest generation logic in the transaction simulator will be enhanced to support this new  'DisregardNamespacePolicy' flag.

andrew-coleman:
Add support for the ‘DisregardNamespacePolicy’ flag when generating the ChaincodeInterest structure in the transaction simulator.

resolves #2856 

### DIFF
--- a/core/endorser/endorser.go
+++ b/core/endorser/endorser.go
@@ -14,6 +14,7 @@ import (
 
 	"github.com/golang/protobuf/proto"
 	"github.com/hyperledger/fabric-chaincode-go/shim"
+	"github.com/hyperledger/fabric-protos-go/common"
 	pb "github.com/hyperledger/fabric-protos-go/peer"
 	"github.com/hyperledger/fabric-protos-go/transientstore"
 	"github.com/hyperledger/fabric/common/flogging"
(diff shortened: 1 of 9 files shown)


manish-sethi (commented):
I feel that the code for building chaincode interest is not easy to follow so as to ascertain the correctness. Made a few comments that relates to correctness.

Just a suggestion and see if it is feasible 
- Now it should be possible to build the chaincode interest straight out of `WritesetMetadata` and `PrivateReads` as the former structure contains all the info about writes and the latter for the private reads. Finding intersection by iterating over the read-write set for preparing an intermediate structure causes an avoidable indirection.

- The logic whether chaincode policy could be discarded is spread at places. Can we decided this at the end?

core/endorser/endorser.go
```diff
@@ -472,86 +473,124 @@ func (e *Endorser) ProcessProposalSuccessfullyOrError(up *UnpackedProposal) (*pb
...
+			}
 		}
 	}
 
-	// now build the chaincode interest
 	ccInterest := &pb.ChaincodeInterest{}
-	for chaincode, collections := range chaincodes {
-		if collections == nil {
-			ccInterest.Chaincodes = append(ccInterest.Chaincodes, &pb.ChaincodeCall{
+	for chaincode, collections := range policies.policyRequired {
```
manish-sethi, line 501: Don't you need to skip system chaincodes here?

core/endorser/endorser.go
```diff
@@ -472,86 +473,124 @@ func (e *Endorser) ProcessProposalSuccessfullyOrError(up *UnpackedProposal) (*pb
...
-			// chaincode either not already in the map, or nil entry - add it
-			chaincodes[ns] = map[string]bool{}
-		}
-		for name := range entry {
-			// even if this already exists from the write-set, it will override the value with true to indicate it was read
-			chaincodes[ns][name] = true
+		for collection := range entry {
+			if !policies.requireCollectionPolicy(chaincode, collection) {
+				policies.add(chaincode, "", true)
+			}
```
manish-sethi: Do you need this block at all? If there would be an entry in `PrivateReads` then the corresponding entry would be present in the hashed form in the public rwset which you have already looped through above.

core/endorser/endorser.go
```diff
@@ -472,86 +473,124 @@ func (e *Endorser) ProcessProposalSuccessfullyOrError(up *UnpackedProposal) (*pb
...
+				}
+			} else {
 				// Since each collection in a chaincode could have different values of the NoPrivateReads flag, create a new Chaincode entry for each.
-				ccInterest.Chaincodes = append(ccInterest.Chaincodes, &pb.ChaincodeCall{
-					Name:            chaincode,
-					CollectionNames: []string{collectionName},
-					NoPrivateReads:  !isRead,
-				})
+				ccCall.CollectionNames = []string{collection}
+				ccCall.NoPrivateReads = !simResult.PrivateReads.Exists(chaincode, collection)
```
manish-sethi, line 524: PrivateReads may have additional entries, so you would need to loop through at the end and add the entries that are not present in the policies map.

manish-sethi (commented):
Looks much clearer in the current form. Thanks @andrew-coleman for your patience! A few final comments.

core/endorser/endorser.go
```diff
@@ -472,86 +473,124 @@ func (e *Endorser) ProcessProposalSuccessfullyOrError(up *UnpackedProposal) (*pb
...
+				if policyBytes, sbeExists := stateMetadata[pb.MetaDataKeys_VALIDATION_PARAMETER.String()]; sbeExists {
+					policy, err := protoutil.UnmarshalSignaturePolicy(policyBytes)
+					if err != nil {
+						return nil, err
+					}
+					sbePolicies = append(sbePolicies, policy)
+					if !mp.policyRequired[ns][""] {
+						// the SBE policies need a container in the proto message - false means namespace policy will be disregarded (unless later set to true)
+						mp.policyRequired[ns][""] = false
+					}
```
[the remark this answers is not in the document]
andrew-coleman: You're right, it doesn't need to be in the inner loop.  But it can't be the default because a 'false' entry here is only required if there are SBE policies present.  I've moved this to the end and put it in a conditional.  I've added more comments to explain the data structure and logic a bit better.

core/endorser/endorser.go
andrew-coleman: That's a good observation.  I had a spurious unit test that required this block, but I've fixed the test and removed this block.  The integration tests continue to pass which proves you are right :)

core/endorser/endorser.go
andrew-coleman: I'm struggling to understand this one.  If there are entries here that are not in the policies map, then it's for private reads with no writes.  But they got added to the policy map due to the existence of the hashed reads in the public RWset (as per your previous comment).  And my observation is that there require the chaincode policy.  There's an integration test for this: https://github.com/andrew-coleman/fabric/blob/DisregardNamespacePolicy/integration/gateway/gateway_discovery_test.go#L546
I'll push my latest commit for now and we can revisit this.

core/endorser/endorser.go
```diff
@@ -472,86 +473,112 @@ func (e *Endorser) ProcessProposalSuccessfullyOrError(up *UnpackedProposal) (*pb
...
+		mp.policyRequired[ns] = map[string]bool{"": false}
+		for coll, kmap := range cmap {
+			// look through each of the states that were written to
+			for _, stateMetadata := range kmap {
+				if policyBytes, sbeExists := stateMetadata[pb.MetaDataKeys_VALIDATION_PARAMETER.String()]; sbeExists {
+					policy, err := protoutil.UnmarshalSignaturePolicy(policyBytes)
+					if err != nil {
+						return nil, err
+					}
+					mp.sbePolicies[ns] = append(mp.sbePolicies[ns], policy)
```
manish-sethi, line 559: Won't it be better to de-duplicate policies, either here (while adding) or later? However, as this does not affect correctness, it can be enhanced in a future PR.

core/endorser/endorser_test.go
```diff
@@ -1174,38 +1224,40 @@ var _ = Describe("Endorser", func() {
...
+							NoPrivateReads:  true,
+						},
+					},
 				},
 			)).To(BeTrue())
 		})
 
-		It("add SBE only", func() {
-			sbe1 := &cb.SignaturePolicyEnvelope{
+		It("SBE only, no chaincode policy updates", func() {
```
manish-sethi, line 1243: I hate to nit-pick of this level but I noticed at a couple of places and just for your reference -  When following ginkgo pattern, an `It` is expected to sound like it makes sense as an sentence, in conjunction with enclosing `Describe` and `Context`. In other word, `It` is expected to be part of the sentence :-)

For example, in this case, following may make sense
```
Describe("Endorser",
Context("when building the ChaincodeInterest"
It("disregards chaincode policy if all updates are driven by SBE", func() {
```

manish-sethi (approved):
Given the new structure, do you thing that the field `NoPublicWrites` in the proto `ChaincodeCall` is expected to be used anywhere? Else, you may like to clean that up as well.

core/endorser/endorser.go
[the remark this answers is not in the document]
manish-sethi: Fixed in the latest patch.

core/endorser/endorser.go
manish-sethi: Fixed in the latest patch.

andrew-coleman:
> Given the new structure, do you thing that the field NoPublicWrites in the proto ChaincodeCall is expected to be used anywhere?

As far as I can see, NoPublicWrites is not handled anywhere in Fabric.  I'm guessing it was a feature that was never implemented.

core/endorser/endorser.go
andrew-coleman: Possibly.  I wanted to minimise the processing overhead of this function as much as possible since this gets invoked for all endorsers whereas it's only the first one that gets looked at by the gateway.  I intended the gateway to do the bulk of this type of processing.  If we find, in practice, that there are situations where large numbers of keys all have the same SBE policy, then could look to merge them in this function.

PR #2859, opened 2021-08-25
Nominate Andrew Coleman as Fabric maintainer

denyeart:
I would like to nominate Andrew Coleman as a Fabric maintainer.
Andrew has many years experience with Fabric, especially Fabric
SDKs where he has been a maintainer and lead developer.
Last year, Andrew demonstrated his Go expertise while working
on the fabric-sdk-go contract programming model.
In the past year Andrew has been working in the core Fabric
repository, leading the Fabric Gateway feature from RFC and community
discussions through implementation, test, and a successful beta preview.
Andrew has also picked up skill in other areas of the peer, and
has been reviewing PRs, extending tests, and coaching others.

### DIFF
--- a/MAINTAINERS.md
+++ b/MAINTAINERS.md
@@ -8,6 +8,7 @@ See [the documentation on Maintainers](https://hyperledger-fabric.readthedocs.io
 | Name | GitHub | Chat | email
 |------|--------|------|----------------------
 | Alessandro Sorniotti | [ale-linux][ale-linux] | aso | <ale.linux@sopit.net>
+| Andrew Coleman | [andrew-coleman][andrew-coleman] | andrew-coleman | <andrew_coleman@uk.ibm.com>
 | Artem Barger | [c0rwin][c0rwin] | c0rwin | <artem@bargr.net>
 | Senthilnathan Natarajan | [cendhu][cendhu] | Senthil1 | <cendhu@gmail.com>
 | Dave Enyeart | [denyeart][denyeart] | dave.enyeart | <enyeart@us.ibm.com>
@@ -57,6 +58,7 @@ See [the documentation on Maintainers](https://hyperledger-fabric.readthedocs.io
 | Will Lahti | [wlahti][wlahti] | wlahti | <will.lahti@gmail.com>
 
 [ale-linux]: https://github.com/ale-linux
+[andrew-coleman]: https://github.com/andrew-coleman
 [binhn]: https://github.com/binhn
 [lindluni]: https://github.com/lindluni
 [c0rwin]: https://github.com/c0rwin


denali49:
+1

manish-sethi (approved):
Welcome aboard Andy!

yacovm:
> Andrew has many years experience with Fabric, especially Fabric
> SDKs where he has been a maintainer and lead developer.
> Last year, Andrew demonstrated his Go expertise while working
> on the fabric-sdk-go contract programming model.

Experience in other projects doesn't count as experience with the Fabric core. That's the reason I removed myself from being a maintainer of fabric node SDK and fabric CA back in the day, right when we moved to separate maintainer lists for these projects.


> In the past year Andrew has been working in the core Fabric repository, leading the Fabric Gateway feature

Correct me I'm wrong, but this amounts to [24 commits in total](https://github.com/hyperledger/fabric/commits?author=andrew-coleman). 

I think it is **unjust** to nominate someone when you are clearly skipping over [candidates](https://github.com/hyperledger/fabric/graphs/contributors) which have made much more code contribution.

manish-sethi:
> Correct me I'm wrong, but this amounts to 24 commits in total.
> 
> I think it is **unjust** to nominate someone when you are clearly skipping over [candidates](https://github.com/hyperledger/fabric/graphs/contributors) which have made much more code contribution.

I believe that number of lines (modulo `vendor dir`) is a better measure than number of commits. Also, I guess that for a new proposed maintainer, accepted perception of continue involvement for a reasonable future time is also an important factor -- that's why voting, else it could be automatic.

yacovm:
> > Correct me I'm wrong, but this amounts to 24 commits in total.
> > I think it is **unjust** to nominate someone when you are clearly skipping over [candidates](https://github.com/hyperledger/fabric/graphs/contributors) which have made much more code contribution.
> 
> I believe that number of lines (modulo `vendor dir`) is a better measure than number of commits. Also, I guess that for a new proposed maintainer, accepted perception of continue involvement for a reasonable future time is also an important factor -- that's why voting, else it could be automatic.

Counting lines of code to measure contribution, is wrong, because you measure the complexity of the solution, and not the complexity of the problem. Even [Dijkstra himself said](https://www.cs.utexas.edu/users/EWD/transcriptions/EWD10xx/EWD1036.html):

`My point today is that, if we wish to count lines of code, we should not regard them as "lines produced" but as "lines spent": the current conventional wisdom is so foolish as to book that count on the wrong side of the ledger.`

And even if it was, then my claim about other people more deserving still holds.

Don't get me wrong, I think Andrew is definitely on the right path, but I think that this is a slippery slope, and if we approve this nomination then the bar (of minimal contribution requirement) will only get lower and lower.

denyeart:
@yacovm As you know we have lost some active maintainers and contributors this year, so I'm not looking at lifetime number of commits, but at activity and quality of commits this year. On both measures Andy stands out as the top candidate in this repository. You are right that the activity in the other repositories doesn't count towards commits or a free pass in this repository, however the experience gained in the other repositories has clearly been a factor in Andy's ability to lead and deliver the largest feature in Fabric this year from community RFC through implementation. That is the type of leadership and results that we are looking for in maintainers.

mastersingh24:
I tend to agree with @denyeart on this one.  Nominations are always open (even self-nomination) so the fact that Dave has nominated Andy does not mean others cannot be nominated as well.

And I fully agree that we need at least one more maintainer who is actually focused on Fabric and will be incented to put in the time required.   Andy seems like a good choice to me as well.

mastersingh24 (approved):

cendhu (approved):

ale-linux (approved):

jyellick (approved):
I share some of Yacov's concerns with quantity of commits, but having truly active maintainers is also important (certainly I haven't had the cycles I'd like to have), so I'm inclined to approve.

denyeart:
Ok, 6 of 10 approving so far, I'll go ahead and merge.

PR #2878, opened 2021-08-30
Clean up Go modules (main)

denyeart:
Run "go mod tidy" and "go mod vendor" on main.


yacovm:
Why don't we add `go mod tidy` diff check to CI? We used to have something like that for dep no?

yacovm (approved):

denyeart:
@yacovm there were some checks added here:
https://github.com/hyperledger/fabric/commit/756a1338f94571ef97f134c2505eb3f5072588ae

In the `main` case go.sum was actually ok, but there were some old vendored dependencies that are not checked.

PR #2896, opened 2021-09-06
Implement chaincode event replay for Fabric Gateway

### LINKED ISSUE #2792: Contract event listening - replay

bestbeforetoday:
Resolves #2792


internal/pkg/gateway/api.go
```diff
@@ -361,12 +362,22 @@ func (gs *Server) ChaincodeEvents(signedRequest *gp.SignedChaincodeEventsRequest
 		return status.Error(codes.InvalidArgument, err.Error())
 	}
 
-	ledgerInfo, err := ledger.GetBlockchainInfo()
-	if err != nil {
-		return status.Error(codes.Unavailable, err.Error())
+	var startBlock uint64
+	switch seek := request.StartPosition.Type.(type) {
```
andrew-coleman: If StartPosition hasn't been specified in the request proto, then this will barf.  Should we default to NextCommit in this case?

andrew-coleman: Like you've done in the unit test :)

andrew-coleman (changes requested):

internal/pkg/gateway/api.go
bestbeforetoday: I think I tried this and the protobuf unmarshalling failed because a oneof field wasn't defined in the message

bestbeforetoday: Having said that... the Golang protobuf docs have an example of switching on a oneof with `nil` as a possible value, so let me try that again

https://developers.google.com/protocol-buffers/docs/reference/go-generated#oneof

andrew-coleman (approved):

PR #2897, opened 2021-09-06
install-fabric.sh script - updated version of bootstrap.

mbwhite:
Background:  I've always found the existing bootstrap.sh script to confusing... do the versions come first or last.  Having to think what components I want to NOT have.

I wanted to offer up this alternative way of working for discussion.

Uses positional arguments and opt-in logic to control the components to install
Uses options to define the optional fabric and ca versions

Order of arguments and options is not important

eg
```
install-fabric.sh docker binary samples         # install docker images, binaries and clones samples dir
install-fabric.sh d b s                         # install all components but with short-hand
install-fabric.sh -f 2.4.0-beta b               # 2.4.0-beta for binaries only
install-fabric.sh docker -c 1.5.1 -f 2.4.0-beta # docker images but for 1.5.1 CA and 2.4.0-beta
```
it is an error to NOT specify a component

#### Type of change

- Bug fix
- New feature
- Test update
- Documentation update


denyeart:
I agree, this is better.

I believe the flags were added to indicate what components are not needed so that it would be compatible with prior behavior when you specified no components. But you're right, specifying what components to download is better and users will immediately understand how to use the command. And if we introduce a new script there are no issues with compatibility with the old script.

We can also take the opportunity to retire the https://bit.ly/2ysbOFE URL (at least from current docs), which I always found a little odd... users should know what is happening rather than hiding it behind a shortened URL, and they should be able to find and inspect the script if desired. I think the shortened URL was introduced to track downloads, which can now be tracked directly from github release downloads.

lindluni:
Two years or so ago I wrote a Go program to do this, where you do a `go get github.com/hyperledger/fabric/installer` and it would install a binary that performed all of these functions. I should dig that up, as I agree, the bash script is dated and confusing.

mbwhite:
@lindluni that could be a very useful tool for use in CI pipelines/docker builds etc.

@denyeart  I'd agree with the bit.ly link.... I'd much rather see the proper URL so I can check it before downloading. Is it an idea to update the docs as part of this PR?

denyeart:
I think we'll want both scripts to co-exist for a time, so we can first merge this, ensure that it works well on different platforms, and then update the docs to point to the new script.

I'm fine with the script approach as it will be obvious what it is doing... I'm thinking it may be too much magic (downloading images, cloning and checking out a repository) to bury in a Go executable. Users can easily inspect a script and decide for themselves if they want to use it or adapt the script for their own purposes (potentially in their own CI).

mbwhite:
@denyeart could this merged?

denyeart:
@mbwhite what platforms have you tested it on?

I think we'll want the old script and the new script to co-exist for a time, but only a short time, otherwise people will forget to update one or the other and they'll diverge. I'm good with this one and updating the install and sample docs to point to it. The trickier question is when to retire the old one. I suspect people have baked it into their scripts or CI and the removal will therefore break them. After this one is merged and the docs are updated, would you want to mention it on the mailing list for people to try it out, as well as mention the planned retirement of the old script, to see if there is any violent pushback?

mbwhite:
@denyeart Ubuntu only at this time... will try and recruit some beta testers...

... and yes to the Mailing list question

scripts/install-fabric.sh
```diff
@@ -0,0 +1,286 @@
...
+# Generated online by https://argbash.io/generate
+
+
+# Default values
+_positionals=()
+_arg_comp=('' )
+
+# if version not passed in, default to latest released version
+# if ca version not passed in, default to latest released version
+_arg_fabric_version="2.3.4"
```
denyeart:
```suggestion
_arg_fabric_version="2.3.3"
```

scripts/install-fabric.sh
```diff
@@ -0,0 +1,286 @@
...
+_arg_ca_version="1.5.1"
```
denyeart:
```suggestion
_arg_ca_version="1.5.2"
```

denyeart (commented):
I tested on mac, it worked but it untarred `bin` and `config` directories in the present directory, whereas the old bootstrap.sh script untars them within the cloned `/fabric-samples` as expected. The samples expect `bin` and `config` to be within `/fabric-samples` directory.

If no components are passed, it gives an error. I see that was your intention, but it may be simpler for new users to automatically download samples, binaries, and docker images if no components are passed?

mbwhite:
> I tested on mac, it worked but it untarred `bin` and `config` directories in the present directory, whereas the old bootstrap.sh script untars them within the cloned `/fabric-samples` as expected. The samples expect `bin` and `config` to be within `/fabric-samples` directory.

Ah thanks! I've updated so that they will go into that directory if there.  

> If no components are passed, it gives an error. I see that was your intention, but it may be simpler for new users to automatically download samples, binaries, and docker images if no components are passed?

That was the intent, my logic being that with the previous script I find it all too easy to get stuff I didn't really want, and have to abort, and make good.

But I take your point - if nothing is specified to default to all.

denyeart (approved):

jkneubuh:
I ran this on a CentOS8 VM, it ran without error on the first, second, third attempts.   The script looks OK.  On the first time launching the routine, I hadn't set up docker (or forgot to start the docker daemon), and the installer script correctly skipped the docker images but proceeded to pull the samples, binaries, and config folders.

At first I was expecting to see an error when specifying no arguments.  I agree this is counter-intuitive, especially for newcomers (like me) that just want to pull down the kitchen sink by default.  I prefer the new behavior where it just "does the right thing" by default, even if it takes a little longer to extract the images and fabric binaries.

While the script is "doing the correct thing," I do think that it's _wrong_ to base the fabric installation on a pipeline whereby we install binaries at all.  For members in the development community, they can build the binaries locally.  For everyone else.... it's far simpler to work directly with the containers.   The expectation and operational/scripting burden of the "install from binaries/config" is too high.   No surprise on my bias here, but we should be encouraging the community to migrate towards a more _cloud native_ development process, especially in CI/CD contexts.

Also I prefer the tiny.url route, or _some_ mechanism by which an end-user can install fabric without needing to clone the git.   If we want to track installation count metrics, there should be either another proxy (e.g. hit counts for assets loaded by the script) to gauge the usage, or simply call a metrics-tracking URL directly from the installation script. 

I understand the review comments come post-merge, but it still looks good.  I definitely prefer this usage pattern to the older "opt-out" style for ala carte fabric installs.

mbwhite:
thanks @jkneubuh...  

_On the subject of the binaries_  this does need some work I agree.  I can see that there is a requirement for a 'cli' of some form to install chaincode etc.. which should be small and compact - and if the  is required should act much as `kubectl` does. Well defined locations. the installer creates them etc

Having to have a config directory set for just packaging chaincode is wrong.. plus it doesn't in all cases do the 'correct' thing.

_On the subject of tiny.url_  this is, to use a UK expression 'a bit marmite' Some people like it, some people don't.  For me, I'd prefer to know where that link was resolving to first. 

 On the practical side I'm not 100% sure who has access to the stats now anyway.

PR #2904, opened 2021-09-08
Namespace Filter

### LINKED ISSUE #2819: Ledger comparison utility - Target namespace flag
Ability to filter ledger channel state comparison by chaincode namespace. This can be achieved with addition of a new command flag for the namespace.

jcastrence:
Resolves #2819

Signed-off-by: Julian Castrence <juliancastrence@ibm.com>

#### Type of change

- New feature

#### Description

- Namespace filter flag allows the user to produce only ledger differences of a given namespace. Compatible with existing flags.
- Since it is now possible to compare two different snapshots and find no differences due to the namespace flag being active, output files with no relevant differences (empty files) will not be generated, and if no output files are generated the output directory will not be generated.

### DIFF
--- a/cmd/ledgerutil/main.go
+++ b/cmd/ledgerutil/main.go
@@ -21,6 +21,7 @@ const (
 		". Generating a file of more than the first 10 differences will result in a large amount " +
 		"of memory usage and is not recommended. Defaults to 10. If set to 0, will not produce " +
 		ledgerutil.FirstDiffsByHeight + "."
+	namespaceDesc = "Namepace filter. Only differences of the given namespace will be compared."
 )
 
 var (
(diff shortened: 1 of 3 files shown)


cmd/ledgerutil/main.go
```diff
@@ -60,19 +62,24 @@ func main() {
...
 		if err != nil {
 			fmt.Printf("%s%s\n", compareErrorMessage, err)
 			return
 		}
 
 		fmt.Print("\nSuccessfully compared snapshots. ")
-		if count == -1 {
+		if *namespace != "" {
+			fmt.Printf("The namespace \"%s\" was used as a filter. ", *namespace)
+		}
```
mastersingh24, line 74: Perhaps this should be moved before you call `ledgerutil.Compare`  above?  And maybe something like `Comparing namespace ...` and if not set `Comparing all namespaces` ?

jcastrence: So some logging before the tool starts the comparison to let the user know it's running and under what parameters? Good suggestion. I think it might also be worth logging to the user the `firstDiffs` value as well.

lindluni: Small Go trick: You can use backticks so you dont have to escape quotes: 
```
`The namespace "%s" was used as a filter.`
```

denyeart:
There is some discussion about not merging this one... rather just leave the utility very simple and always collect all differences since that is what will be required when troubleshooting.

jcastrence:
As @denyeart mentioned, we've decided to drop this feature to keep the tool simple. This PR will be closed.

PR #2909, opened 2021-09-14
Write config blocks synchronously in Orderer

Alex-duzhichao:
#### Type of change
- Bug fix

#### Description
Assume the config block contains a transaction which removes the
node from consenters, the consensus chain will switch to follower
chain which reads last config block from ledger directly. Because of
writing config block asynchronously, follower chain maybe read a
stale config block in which the node is still a consenter. So the
node will switch back to a consenter again.

Writting config blocks synchronously would fix this bug.

#### Related issues
Resolves #<2908>

### DIFF
--- a/orderer/common/multichannel/blockwriter.go
+++ b/orderer/common/multichannel/blockwriter.go
@@ -156,7 +156,7 @@ func (bw *BlockWriter) WriteConfigBlock(block *cb.Block, encodedMetadataValue []
 		logger.Panicf("Told to write a config block with unknown header type: %v", chdr.Type)
 	}
 
-	bw.WriteBlock(block, encodedMetadataValue)
+	bw.WriteBlockSync(block, encodedMetadataValue)
 }
 
 // WriteBlock should be invoked for blocks which contain normal transactions.
@@ -173,6 +173,23 @@ func (bw *BlockWriter) WriteBlock(block *cb.Block, encodedMetadataValue []byte) 
 		defer bw.committingBlock.Unlock()
 		bw.commitBlock(encodedMetadataValue)
 	}()
+}
+
+// WriteBlockSync is same as WriteBlock, but commits block synchronously.
+// Note: WriteConfigBlock should use WriteBlockSync instead of WriteBlock.
+//       If the block contains a transaction that remove the node from consenters,
+//       the node will switch to follower and pull blocks from other nodes.
+//       Suppose writing block asynchronously, the block maybe not persist to disk
+//       when the follower chain starts working. The follower chain will read a block
+//       before the config block, in which the node is still a consenter, so the follower
+//       chain will switch to the consensus chain. That's a dead loop!
+//       So WriteConfigBlock should use WriteBlockSync instead of WriteBlock.
+func (bw *BlockWriter) WriteBlockSync(block *cb.Block, encodedMetadataValue []byte) {
+	bw.committingBlock.Lock()
+	bw.lastBlock = block
+
+	defer bw.committingBlock.Unlock()
+	bw.commitBlock(encodedMetadataValue)
 }
 
 // commitBlock should only ever be invoked with the bw.committingBlock held
--- a/orderer/common/multichannel/blockwriter_test.go
+++ b/orderer/common/multichannel/blockwriter_test.go
@@ -252,6 +252,50 @@ func TestGoodWriteConfig(t *testing.T) {
 	require.Equal(t, consenterMetadata, omd.Value)
 }
 
+func TestWriteConfigSynchronously(t *testing.T) {
+	confSys := genesisconfig.Load(genesisconfig.SampleInsecureSoloProfile, configtest.GetDevConfigDir())
+	genesisBlockSys := encoder.New(confSys).GenesisBlock()
+
+	tmpdir, err := ioutil.TempDir("", "file-ledger")
+	require.NoError(t, err)
+	defer os.RemoveAll(tmpdir)
+
+	_, l := newLedgerAndFactory(tmpdir, "testchannelid", genesisBlockSys)
+
+	fakeConfig := &mock.OrdererConfig{}
+	fakeConfig.ConsensusTypeReturns("solo")
+
+	cryptoProvider, err := sw.NewDefaultSecurityLevelWithKeystore(sw.NewDummyKeyStore())
+	require.NoError(t, err)
+
+	mockValidator := &mocks.ConfigTXValidator{}
+	mockValidator.ChannelIDReturns("testchannelid")
+	bw := newBlockWriter(genesisBlockSys, nil,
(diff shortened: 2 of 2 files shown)


ale-linux (commented):
good catch - but it sounds like there was no test that showed the bug before, and no test that asserts that the bug is fixed now. Imho this should be fixed

yacovm:
This exposes another, more critical bug, no?
When you're a member of the cluster, you use an asynchronous block writer, but then when you switch to a follower, you actually don't use it anymore, and you write blocks directly to the ledger, bypassing the block writer, and then in theory you may write blocks out of order to the ledger (which leads to a crash).

@tock-ibm  what do you think?

Alex-duzhichao:
> good catch - but it sounds like there was no test that showed the bug before, and no test that asserts that the bug is fixed now. Imho this should be fixed

@ale-linux I add a unit test case to `BlockWriter` for testing whether the config block has been written persistently to ledger.
Please re-review the PR :)

C0rWin:
> > good catch - but it sounds like there was no test that showed the bug before, and no test that asserts that the bug is fixed now. Imho this should be fixed
> 
> @ale-linux I add a unit test case to `BlockWriter` for testing whether the config block has been written persistently to ledger.
> Please re-review the PR :)

I think @yacovm has a point, can you please also add a test to prove to disprove his comment? I.e. write simultaneously normal blocks and config blocks and make sure they are committed in the same order.

Alex-duzhichao:
> This exposes another, more critical bug, no?
> When you're a member of the cluster, you use an asynchronous block writer, but then when you switch to a follower, you actually don't use it anymore, and you write blocks directly to the ledger, bypassing the block writer, and then in theory you may write blocks out of order to the ledger (which leads to a crash).
> 
> @tock-ibm what do you think?

https://github.com/hyperledger/fabric/blob/8767ceddf604575876be479e77c11d818425e06f/orderer/common/multichannel/blockwriter.go#L168-L176

I think this bug should not happen if this PR can be merged, because `WriteBlock` is not totally asynchronously, it should acquire `committingBlock` before writing block asynchronously and return. The switch from consensus chain to follower chain must happen after a config block, so if we write config block synchronously, we can guarantee all blocks have been written to ledger before the follower chain starting.

Alex-duzhichao:
> > > good catch - but it sounds like there was no test that showed the bug before, and no test that asserts that the bug is fixed now. Imho this should be fixed
> > 
> > 
> > @ale-linux I add a unit test case to `BlockWriter` for testing whether the config block has been written persistently to ledger.
> > Please re-review the PR :)
> 
> I think @yacovm has a point, can you please also add a test to prove to disprove his comment? I.e. write simultaneously normal blocks and config blocks and make sure they are committed in the same order.

I add a test case as you say

C0rWin:
> it should acquire `committing blocks before writing block asynchronously and return.

Would you agree that who gets to the lock first now depends on the routine scheduling if there is a flight config change?

Alex-duzhichao:
> > it should acquire `committing blocks before writing block asynchronously and return.
> 
> Would you agree that who gets to the lock first now depends on the routine scheduling if there is a flight config change?

No, I don't think so. In my opinion, all blocks(normal blocks and config blocks) are written sequentially in a single goroutine, so the previous block must get the lock first no matter how the goroutine is scheduled.

C0rWin:
> > > it should acquire `committing blocks before writing block asynchronously and return.
> > 
> > 
> > Would you agree that who gets to the lock first now depends on the routine scheduling of a flight config change?
> 
> No, I don't think so. In my opinion, all blocks(normal blocks and config blocks) are written sequentially in a single goroutine, so the previous block must get the lock first, no matter how the goroutine is scheduled.

The point is, that the following code:

```
func (bw *BlockWriter) WriteBlock(block *cb.Block, encodedMetadataValue []byte) {
	bw.committingBlock.Lock()
	bw.lastBlock = block
	go func() {
		defer bw.committingBlock.Unlock()
		bw.commitBlock(encodedMetadataValue)
	}()
}
```

It cannot be executed in the same go-routine with the sync write you have added. I'm still pretty sure, that there is a problem here with your fix, as @yacovm pointed out. WriteBlock has the explicit call for a new go-routine.

Alex-duzhichao:
> > > > it should acquire `committing blocks before writing block asynchronously and return.
> > > 
> > > 
> > > Would you agree that who gets to the lock first now depends on the routine scheduling of a flight config change?
> > 
> > 
> > No, I don't think so. In my opinion, all blocks(normal blocks and config blocks) are written sequentially in a single goroutine, so the previous block must get the lock first, no matter how the goroutine is scheduled.
> 
> The point is, that the following code:
> 
> ```
> func (bw *BlockWriter) WriteBlock(block *cb.Block, encodedMetadataValue []byte) {
> 	bw.committingBlock.Lock()
> 	bw.lastBlock = block
> 	go func() {
> 		defer bw.committingBlock.Unlock()
> 		bw.commitBlock(encodedMetadataValue)
> 	}()
> }
> ```
> 
> It cannot be executed in the same go-routine with the sync write you have added. I'm still pretty sure, that there is a problem here with your fix, as @yacovm pointed out. WriteBlock has the explicit call for a new go-routine.

https://github.com/hyperledger/fabric/blob/70ff46ad0181e5e52e25b3566281aa71c05bcdde/orderer/consensus/etcdraft/chain.go#L1049-L1157

I think all `WriteBlock` is called in the `apply` function, and `apply` is called in a single go-routine. 
The point is, before `WriteBlock` creates a new go-routine, `WriteBlock` acquires the `committingBlock` lock first, and after the new go-routine finish, the lock is released.  This lock can guarantee that blocks won't be written out of order.
For example, if we have two blocks `block1` and `block2`, `block1` calls `WriteBlock` that acquires the lock and returns, leaving the new created `go-routine1` run asynchronously. Then  `block2` calls `WriteBlock` and it should be blocked, because the lock hasn't been release by the background `go-rountine1`. Only after the `go-routine1` finished, it releases the lock, the `WriteBlock` called by `block2` can acquire the lock and creates the new go-routine `go-routine2`. It seems like a buffered channel and it can guarantee the blocks is written sequentially.

denyeart:
@yacovm @C0rWin could you take another look?

C0rWin:
@yacovm @denyeart I read the code and @Alex-duzhichao explanation, IMO it's good to go. I also cloned the patch and checked it locally. It seems working as expected.

yacovm:
> @yacovm @C0rWin could you take another look?

I never said the proposed code has a problem. I said that the *current* code has a problem.

But I am not sure that @ale-linux 's comment is addressed:

> good catch - but it sounds like there was no test that showed the bug before, and no test that asserts that the bug is fixed now. Imho this should be fixed

C0rWin (approved):

PR #2927, opened 2021-09-16
Cache channel orderers in registry

### LINKED ISSUE #2911: Gateway: caching ordering nodes
Currently, the gateway's submit function makes an internal request to the peer's discovery service to determine the list of available orderers for a channel.  
It is inefficient to do this on every submit call since the list of orderers is relatively static.
Instead, a map of ordering nodes per channel should be cached in the gateway, and a callback should be registered for config update messages to trigger cache refresh.

andrew-coleman:
Currently discovery is invoked to get the list of channel orderers on each Submit() request (the orderer connections are cached).
This enhancement caches the list of orderers and registers a callback with the channel config update bundle to trigger cache refresh

resolves #2911 

### DIFF
--- a/core/peer/peer_test.go
+++ b/core/peer/peer_test.go
@@ -16,6 +16,8 @@ import (
 	"testing"
 	"time"
 
+	"github.com/hyperledger/fabric/common/channelconfig"
+
 	"github.com/hyperledger/fabric-protos-go/common"
 	pb "github.com/hyperledger/fabric-protos-go/peer"
 	"github.com/hyperledger/fabric/bccsp/sw"
@@ -296,6 +298,42 @@ func TestDeliverSupportManager(t *testing.T) {
 	peerInstance.channels = map[string]*Channel{"testchain": {}}
 	chainSupport = manager.GetChain("testchain")
 	require.NotNil(t, chainSupport, "chain support should not be nil")
+}
+
+func TestConfigCallback(t *testing.T) {
+	peerInstance, cleanup := NewTestPeer(t)
+	defer cleanup()
+
+	var callbackInvoked bool
+	peerInstance.AddConfigCallbacks(func(bundle *channelconfig.Bundle) {
+		callbackInvoked = true
+		orderer, exists := bundle.OrdererConfig()
+		require.True(t, exists)
+		require.NotEmpty(t, orderer.Organizations()["SampleOrg"].Endpoints)
+	})
+
+	peerInstance.Initialize(
+		nil,
+		nil,
+		plugin.MapBasedMapper(map[string]validation.PluginFactory{}),
+		&ledgermocks.DeployedChaincodeInfoProvider{},
+		nil,
+		nil,
+		runtime.NumCPU(),
+	)
+
+	testChannelID := fmt.Sprintf("mytestchannelid-%d", rand.Int())
+	block, err := configtxtest.MakeGenesisBlock(testChannelID)
+	require.NoError(t, err)
+
+	// We expect the callback to be invoked when the channel is created
+	require.False(t, callbackInvoked)
+
+	err = peerInstance.CreateChannel(testChannelID, block, &ledgermocks.DeployedChaincodeInfoProvider{}, nil, nil)
+	require.NoError(t, err)
+
+	// the callback should have been invoked
+	require.True(t, callbackInvoked)
 }
 
 func constructLedgerMgrWithTestDefaults(ledgersDataDir string) (*ledgermgmt.LedgerMgr, error) {
--- a/internal/pkg/gateway/registry.go
+++ b/internal/pkg/gateway/registry.go
@@ -13,11 +13,11 @@ import (
 	"strings"
 	"sync"
 
-	"github.com/hyperledger/fabric-protos-go/peer"
-
 	"github.com/golang/protobuf/proto"
 	dp "github.com/hyperledger/fabric-protos-go/discovery"
 	"github.com/hyperledger/fabric-protos-go/gossip"
+	"github.com/hyperledger/fabric-protos-go/peer"
+	"github.com/hyperledger/fabric/common/channelconfig"
 	"github.com/hyperledger/fabric/common/flogging"
 	gossipapi "github.com/hyperledger/fabric/gossip/api"
 	gossipcommon "github.com/hyperledger/fabric/gossip/common"
@@ -32,15 +32,15 @@ type Discovery interface {
 }
 
 type registry struct {
-	localEndorser       *endorser
-	discovery           Discovery
-	logger              *flogging.FabricLogger
-	endpointFactory     *endpointFactory
-	remoteEndorsers     map[string]*endorser
-	broadcastClients    map[string]*orderer
-	tlsRootCerts        map[string][][]byte
-	channelsInitialized map[string]bool
-	configLock          sync.RWMutex
+	localEndorser      *endorser
+	discovery          Discovery
+	logger             *flogging.FabricLogger
+	endpointFactory    *endpointFactory
+	remoteEndorsers    map[string]*endorser
+	broadcastClients   sync.Map // orderer address (string) -> client connection (orderer)
+	channelInitialized map[string]bool
+	configLock         sync.RWMutex
+	channelOrderers    sync.Map // channel (string) -> orderer addresses (endpointConfig)
 }
 
 type endorserState struct {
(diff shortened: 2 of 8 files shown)


internal/pkg/gateway/registry.go
```diff
@@ -348,7 +341,14 @@ func (reg *registry) registerChannel(channel string) error {
 			}
 		}
 	}
-	reg.channelsInitialized[channel] = true
+	reg.endorsersInitialized[channel] = true
 
 	return nil
 }
+
+func (reg *registry) configUpdate(bundle *channelconfig.Bundle) {
```
yacovm, line 372: 1. The goroutine that calls the callback writes to reg.configValid but the goroutine that reads it in line 282 in the registry, does not see that write, because there is no memory happens-before relation. You need to lock here with the configLock. Furthermore, you may have several invocations of this method in parallel, which is a problem in itself.
If you want to avoid locking explictly, you can use the sync.Map struct which is thread safe. I believe it is cheaper and faster as it's optimized for read intensive workloads.
2. Why do you invalidate the entire map if only a specific channel was updated? You can extract the channel name from the bundle.

andrew-coleman: Thanks, I hadn't spotted how to extract the channel name.  I've made that change and switched to a sync.Map so that the callback won't get potentially blocked by the registry making peer/orderer connections

core/peer/peer_test.go
```diff
@@ -298,6 +300,47 @@ func TestDeliverSupportManager(t *testing.T) {
...
+		runtime.NumCPU(),
+	)
+
+	testChannelID := fmt.Sprintf("mytestchannelid-%d", rand.Int())
+	block, err := configtxtest.MakeGenesisBlock(testChannelID)
+	if err != nil {
+		fmt.Printf("Failed to create a config block, err %s\n", err)
+		t.FailNow()
+	}
+
```
yacovm, line 328: Isn't it a contradiction? 
```

// We expect the callback to be invoked when the channel is created
	require.False(t, callbackInvoked)
```

core/peer/peer_test.go
```diff
@@ -298,6 +300,47 @@ func TestDeliverSupportManager(t *testing.T) {
...
+	// We expect the callback to be invoked when the channel is created
+	require.False(t, callbackInvoked)
+
+	err = peerInstance.CreateChannel(testChannelID, block, &ledgermocks.DeployedChaincodeInfoProvider{}, nil, nil)
```
yacovm, line 332: why use t.Fatalf if you have require somewhere else? Let's be consistent with the pattern

internal/pkg/gateway/registry.go
```diff
@@ -264,63 +266,54 @@ func (reg *registry) orderers(channel string) ([]*orderer, error) {
...
-
-	for _, eps := range config.GetOrderers() {
-		for _, ep := range eps.Endpoint {
-			url := fmt.Sprintf("%s:%d", ep.Host, ep.Port)
-			if ord, ok := reg.broadcastClients[url]; ok {
-				orderers = append(orderers, ord)
-			}
-		}
+	var orderers []*orderer
+	for _, url := range reg.channelOrderers[channel] {
```
yacovm: Can't this map be accessed concurrently while **registerChannel()** is invoked at the same time? You're only locking in **registerChannel** but not here.

internal/pkg/gateway/registry.go
```diff
@@ -264,63 +266,54 @@ func (reg *registry) orderers(channel string) ([]*orderer, error) {
...
+		orderers = append(orderers, reg.broadcastClients[url])
```
yacovm: ditto for **broadcastClients**.

internal/pkg/gateway/registry.go
```diff
@@ -264,63 +266,54 @@ func (reg *registry) orderers(channel string) ([]*orderer, error) {
...
+					reg.broadcastClients[address] = newOrd
+					reg.logger.Infof("Added orderer to registry: %s", address)
 				}
-				reg.broadcastClients[address] = orderer
-				reg.logger.Infof("Added orderer to registry: %s", address)
+				channelOrderers = append(channelOrderers, address)
 			}
 		}
+		reg.channelOrderers[channel] = channelOrderers
+		reg.configValid.Store(channel, true)
```
yacovm: Suppose you have 2 config blocks processed serially, or in a very small amount of time.

The first config block comes in, makes **configValid** be false. 
Further suppose that you get an invocation of **registerChannel** in parallel, and that invocation reaches line 290, so it has read the first config.
Then, suppose the **registerChannel** invocation is delayed, and the second block commits and the callback makes **configValid** be false again. 
Then, when the **registerChannel** reaches line 312, it will set **configValid** for that channel to true, but the cached config will be stale.

internal/pkg/gateway/registry.go
```diff
@@ -264,63 +266,54 @@ func (reg *registry) orderers(channel string) ([]*orderer, error) {
...
-					return err
+		// Get the current set of orderers.  Use existing connections, if available.  Some may have been added, some removed.
+		var channelOrderers []string
+		for mspid, eps := range config.GetOrderers() {
+			for _, ep := range eps.Endpoint {
+				address := fmt.Sprintf("%s:%d", ep.Host, ep.Port)
+				if _, ok := reg.broadcastClients[address]; !ok {
+					// this orderer is new - connect to it and add to the broadcastClients registry
+					tlsRootCerts := reg.tlsRootCerts[mspid]
+					newOrd, err := reg.endpointFactory.newOrderer(address, mspid, tlsRootCerts)
```
yacovm: I know this code was there previously and isn't introduced in this commit. 
Regardless, you may want to know that there is a problem with the current code, and you may want to fix it.

the **newOrderer** method will be called upon every channel config, and that method calls **newConnection** which in turn performs a **synchronous** gRPC **Dial**.
This means that if there is an orderer which you have never connected to before, the method **Submit** of the gateway will **freeze** for the duration of the timeout. 

Perhaps it would be better to make the gRPC connection asynchronous and let gRPC handle it, via the **AsyncConnect** flag.

Another problem is that I don't see anywhere that you close connections to orderers that do not exist anymore. 
It can be easily inferred from the fact that the gRPC connection pointer is swallowed by a gRPC client, and can never be accessed ever again.

yacovm: By the way, there is an even worse problem with the endorser approach, and if a peer is simply shut down forever, you will have a file descriptor leak because the gRPC connection can never be closed from the code.

internal/pkg/gateway/registry.go
yacovm: What I suggest is that you decouple the config parsing and the connection creation (network I/O):

1. Upon a config change, parse the config in the callback and put in the sync.Map the orderer addresses. 
2. Upon a submit (not register), take the latest config, and do the bookkeeping of the connections, while the connections are created asynchronously, and this would let you create them and close them in an almost non-blocking manner.

internal/pkg/gateway/registry.go
yacovm: See my [comment](https://github.com/hyperledger/fabric/pull/2927#discussion_r710454334) about that problem.

core/peer/peer_test.go
andrew-coleman: I don't think so.  It's asserting that the callback gets invoked while processing the `CreateChannel` method.
So it checks it hasn't been invoked beforehand, then checks it has been invoked afterwards.

core/peer/peer_test.go
andrew-coleman: Good point.  Careless cut and paste from a previous test.  I've changed to just using require.

internal/pkg/gateway/registry.go
andrew-coleman: Yes, I'd accidentally removed the read lock in this commit.
I've now modified the logic so that it doesn't access/change anything other than sync.Maps

internal/pkg/gateway/registry.go
andrew-coleman: As above.

internal/pkg/gateway/registry.go
andrew-coleman: Thanks for this.  I've reimplemented it along the lines you suggest.

internal/pkg/gateway/registry.go
andrew-coleman: For now, I've kept the connection synchronous, and I'll create a separate issue to change it to asynchronous.  That will require some extra work (not related to this commit) in order to potentially wait on connections when they are needed.

I'll also create an issue to handle connection closure.  Again, beyond the scope of this commit.

internal/pkg/gateway/registry.go
```diff
@@ -260,68 +262,55 @@ func contains(slice []string, entry string) bool {
...
+			if ok {
+				tlsRootCerts = cert.([][]byte)
 			}
+			client, err = reg.endpointFactory.newOrderer(ep.address, ep.mspid, tlsRootCerts)
+			if err != nil {
+				// Failed to connect to this orderer for some reason.  Log the problem and skip to the next one.
+				reg.logger.Warnw("Failed to connect to orderer", "address", ep.address, "err", err)
+				continue
+			}
+			reg.broadcastClients.Store(ep.address, client)
```
yacovm: You need to lock. Otherwise two goroutines, each invoking **Submit()** can race each other to this point, and they will both create a new gRPC connection, and only one of them will be accessible in the map, and the other one will never be garbage collected and cause a resource leak.

internal/pkg/gateway/registry.go
```diff
@@ -348,7 +346,41 @@ func (reg *registry) registerChannel(channel string) error {
...
 }
+
+func (reg *registry) updateConfig(channel string) ([]*endpointConfig, error) {
+	config, err := reg.discovery.Config(channel)
+	if err != nil {
+		return nil, fmt.Errorf("failed to get config for channel [%s]: %w", channel, err)
+	}
+	// get the tlscerts
+	for msp, info := range config.GetMsps() {
+		reg.tlsRootCerts.Store(msp, info.GetTlsRootCerts())
```
yacovm: I don't understand. Why are we aggregating all TLS root CAs? Discovery, and also the channel config, separates TLS CAs by their MSP, and here you are aggregating them together.

This means that a peer issued by the CA of organization A can be impersonated by a rogue admin of organization B.

internal/pkg/gateway/registry.go
andrew-coleman: Yep.  Rather than add a broader lock, I have followed the pattern described here https://github.com/golang/go/issues/44159#issuecomment-780774977

internal/pkg/gateway/registry.go
andrew-coleman: This cache is not needed anymore - I've removed it

internal/pkg/gateway/registry.go
```diff
@@ -260,68 +260,56 @@ func contains(slice []string, entry string) bool {
...
-	}
 	var orderers []*orderer
-
-	// Get the config
-	config, err := reg.discovery.Config(channel)
-	if err != nil {
-		return nil, err
+	var ordererEndpoints []*endpointConfig
+	var err error
+	addr, exists := reg.channelOrderers.Load(channel)
```
yacovm, line 265: I am not sure you can achieve the task in a lock free manner. 

At some point you will need to disconnect orderers when they are no longer in your config, right? 
When that happens, you will have to lock, otherwise you are risking that a goroutine will load the orderers from the map, and then in parallel the orderer's connection will be closed, and right afterwards, that orderer will be connected to, and you will end up with a connection to an orderer that doesn't exist anymore. 

What do you think?

andrew-coleman: I haven't implemented orderer/peer disconnection yet.  When I do, locking may well have to be added here.  But I'd rather do that as part of that future PR rather than in anticipation of something that hasn't been implemented.

internal/pkg/gateway/registry.go
```diff
@@ -348,7 +343,41 @@ func (reg *registry) registerChannel(channel string) error {
...
+
+func (reg *registry) updateConfig(channel string) ([]*endpointConfig, error) {
+	config, err := reg.discovery.Config(channel)
+	if err != nil {
+		return nil, fmt.Errorf("failed to get config for channel [%s]: %w", channel, err)
+	}
+	var channelOrderers []*endpointConfig
+	for mspid, eps := range config.GetOrderers() {
+		// This mspid should be in the tlsRootCerts map, but if not, pull that from the config message too
+
```
manish-sethi: This comment does not seem to match with the code.

internal/pkg/gateway/registry.go
```diff
@@ -348,7 +343,41 @@ func (reg *registry) registerChannel(channel string) error {
 			}
 		}
 	}
-	reg.channelsInitialized[channel] = true
+	reg.channelInitialized[channel] = true
 
 	return nil
 }
+
+func (reg *registry) updateConfig(channel string) ([]*endpointConfig, error) {
```
manish-sethi: As far as I understood, this function gets invoked from two places. (1) from the function that does lazy loading the orderers details and (2) via config update callback. Won't you need a locking mechanism to make sure that it's (2) that always wins if there is a race? In other word's (1) should never be able to overwrite what is written by (2) for a particular channel?

yacovm: Good catch Manish. I think that maybe there should be a critical section with sync.Once where you do a preliminary loading of the channel's config and this would ensure that you can never get to a state where the channel's orderers aren't populated outside the sync.Once block.

Something like this:
1. Check if **reg.channelOrderers.Load(channel)** is empty
2. If not, great.
3. Else, it's empty. Load from discovery.Config inside a sync.Once block where you check if it has already been populated. If yes, then return the already populated. Else, **reg.channelOrderers.Store()** and do not store it anywhere else in the case of (1) but only in the case of (2)

wdyt?

internal/pkg/gateway/registry.go
andrew-coleman: I meant to delete that comment when I deleted the code

internal/pkg/gateway/registry.go
andrew-coleman: Alternatively, change `updateConfig()` so that it only reads and returns the orderer config rather than store it, then when invoked from (1), follow it with LoadOrStore(), and when invoked from (2), follow it with Store().  That would ensure that in a race condition, the channelOrderers sync.Map would always end up with the config from (2).

manish-sethi: Yes, Both should work. I guess, the one that Andy suggested causes minimum changes in the current code.

manish-sethi (approved):

PR #2933, opened 2021-09-17
Apply the style only the key in readwrite.rst

justin-themedium:
Fix the style from '``k5,``' to '``k5``',

Signed-off-by: Justin Yang <justin.yang@themedium.io>

#### Type of change
- Documentation update

#### Description
Apply the style only the key in readwrite.rst

Fix the style from '```k5,```' to '```k5```,' in 

> T5 passes validation because it reads a key, k5, which was not modified by any of the preceding transactions

in [Read-Write set semantics](https://hyperledger-fabric.readthedocs.io/en/latest/readwrite.html)


yacovm (approved):

denali49 (approved):
lgtm, thanks @justin-themedium

Issue #2931, opened 2021-09-16
Orderer fails to start after some time being down. failed to open WAL: fileutil: file already locked


kopaygorodsky:
It happened on my network when an orderer crashed because no space was left of a device. I'm running nodes in k8s, so I just resized related PersistentVolumeClaim and restarted the node.  It fails to start with this error `fileutil: file already locked`. It happens only to a specific channel, the other ones are ok and replicated just fine.

Logs : https://pastebin.com/pb1XSGn9

For some reason orderer tries to HandleChain two times for the same channel and it fails on the second attempt. 
Maybe when it `Restarting raft node channel=channel6` a race condition happens and the previous node hadn't closed the storage and new one already tried to use it.

yacovm:
> 2021-09-16 23:14:04.549 UTC [orderer.common.cluster] Configure -> INFO 079 Entering, channel: channel6, nodes: []
> 2021-09-16 23:14:04.549 UTC [orderer.consensus.etcdraft] start -> DEBU 07b Starting raft node: #peers: 1 channel=channel6 node=1

what is going on here? Is that node the only one in that channel?

kopaygorodsky:
yes, it was some kind of dev env with a single orderer node. I managed to replicate the same on the network with 3 orderers.

kopaygorodsky:
@yacovm pointed 
`a channel was created and space ran out, so the orderer crashed. When it boots, the system channel initialized before the application channel and it sees it hasn't finished creating the channel properly
so it creates the channel again, and also initialized the chain
afterwards, the application channel is started`

yacovm:
Merged in 2.3, 2.2, 1.4 and main.

guoger:
I think there's another issue caused by this problem: if we have two consecutive blocks creating the same channel, and if a new orderer is being added to this channel, it will be confused and recognize the second block as genesis block, and crashes later due to hash pointer mismatch.

possible related code https://github.com/hyperledger/fabric/blob/c3f6ef94926ac6560bf2776fc65c85ede50bb0fb/orderer/common/cluster/replication.go#L547-L573

possible fix would be to ignore second channel creation block during the loop.

cc @yacovm @kopaygorodsky

yacovm:
> I think there's another issue caused by this problem: if we have two consecutive blocks creating the same channel, and if a new orderer is being added to this channel, it will be confused and recognize the second block as genesis block, and crashes later due to hash pointer mismatch.
> 
> possible related code
> 
[...]

PR #2934, opened 2021-09-18
FAB-2931: do not create a chain if it's already created

kopaygorodsky:
Signed-off-by: Vladyslav Kopaihorodskyi <vlad.kopaygorodsky@gmail.com>

#### Type of change
- Bug fix

#### Description

Do not create a new chain of type `etcdraft.Chain` if such exists in the map of chains. This can happen when in Raft protocol a channel was created, but not marked as done in WAL logs. So at orderer startup, it tried to create another instance of a chain and panicked because that instance startup failed. 

#### Related issues

#2931 

#### Release Note
Fixed bug when an orderer crashed at channel creation and after restart couldn't bootstrap because of desynchronization between WAL logs and ledger state.

### DIFF
--- a/orderer/common/multichannel/registrar.go
+++ b/orderer/common/multichannel/registrar.go
@@ -32,6 +32,7 @@ import (
 	"github.com/hyperledger/fabric/orderer/common/msgprocessor"
 	"github.com/hyperledger/fabric/orderer/common/types"
 	"github.com/hyperledger/fabric/orderer/consensus"
+	"github.com/hyperledger/fabric/orderer/consensus/etcdraft"
 	"github.com/hyperledger/fabric/protoutil"
 	"github.com/pkg/errors"
 )
@@ -537,7 +538,10 @@ func (r *Registrar) CreateChain(chainName string) {
 	if chain != nil {
 		logger.Infof("A chain of type %T for channel %s already exists. "+
 			"Halting it.", chain.Chain, chainName)
+		r.lock.Lock()
 		chain.Halt()
+		delete(r.chains, chainName)
+		r.lock.Unlock()
 	}
 	r.newChain(configTx(lf))
 }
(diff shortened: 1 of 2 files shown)


yacovm:
Looks good to me. Let's get more eyes on this.

yacovm:
@jyellick  @guoger

orderer/common/multichannel/registrar.go
```diff
@@ -537,7 +538,10 @@ func (r *Registrar) CreateChain(chainName string) {
 	if chain != nil {
 		logger.Infof("A chain of type %T for channel %s already exists. "+
 			"Halting it.", chain.Chain, chainName)
+		r.lock.Lock()
```
ale-linux, line 541: Not strictly related to this PR, but I had a look and noticed that there are several accesses to `chains` that are also not gated by a lock. Could you perhaps double-check?

kopaygorodsky: There are two methods that use the locking mechanism to store `r.chain`s. Seems all new/refactored methods started to lock the chains map. I can go over whole register and add locks. 
@yacovm wdut ?

yacovm: Let's not do this in this PR.

kopaygorodsky: I can create another issue and do that if it's really needed. ok?

yacovm: The fact that we don't lock in every access doesn't mean we should. We don't need to lock in the main goroutine for memory access that chronologically precedes memory access by other goroutines, since all goroutines are children of the main goroutine, hence there is a happens-before memory relation.

I don't believe you will find a problem, but, I could be wrong, so if you find a missing lock, please do create an issue and fix it.

kopaygorodsky: Ok, marking it as resolved for now.

ale-linux: Fine by me not to do it in this PR. I just find code that makes seemingly random use of locking for maps (with no explanation) a little hard to maintain. Ideally we'd always use locks (so it's easy to find issues) or carefully document things in the comments. We don't do either..

jyellick:
Looks good to me

denyeart (approved):

denyeart:

denyeart:

yacovm:
@kopaygorodsky do you want to backport it to 1.4 and 2.2 branches?

denyeart:
We'll have to backport to release-2.2 and release-1.4 by hand (cherry-pick), auto backport only worked on release-2.3.

kopaygorodsky:
@yacovm @denyeart yes, ofc. I'll provide PRs tomorrow.

guoger:
this should've fixed https://jira.hyperledger.org/browse/FAB-18526 as well

denyeart:
@kopaygorodsky Reminder before we lose track of this one... if you can't find time for the remaining backports we can likely find another person...

kopaygorodsky:
@denyeart I'll definitely find some time this weekend, had an unexpectedly busy week, sorry.

Issue #2835, opened 2021-08-17
Resolve fabric/orderer/common/cluster unit test flakes


denyeart:
See details and occurrences in https://jira.hyperledger.org/browse/FAB-18506

denyeart:
#2935 fixed the most frequent flake.

The other flake we see sometimes is in TestBlockPullerBadBlocks.

Example here:

https://dev.azure.com/Hyperledger/0227bd2c-c4f6-478f-be00-ee519f115180/_apis/build/builds/39770/logs/9

```
2021-08-24T16:44:15.7497999Z goroutine 1926 [chan receive, 19 minutes]:
2021-08-24T16:44:15.7498361Z testing.(*T).Run(0xc000252900, 0x1191b40, 0x8, 0xc00023c000, 0x1)
2021-08-24T16:44:15.7498748Z 	/opt/hostedtoolcache/go/1.16.7/x64/src/testing/testing.go:1239 +0x610
2021-08-24T16:44:15.7499163Z github.com/hyperledger/fabric/orderer/common/cluster_test.TestBlockPullerBadBlocks(0xc000252900)
2021-08-24T16:44:15.7499646Z 	/home/vsts/work/1/go/src/github.com/hyperledger/fabric/orderer/common/cluster/deliver_test.go:1047 +0x485
2021-08-24T16:44:15.7500044Z testing.tRunner(0xc000252900, 0x11ccd90)
2021-08-24T16:44:15.7500379Z 	/opt/hostedtoolcache/go/1.16.7/x64/src/testing/testing.go:1193 +0x203
2021-08-24T16:44:15.7500716Z created by testing.(*T).Run
2021-08-24T16:44:15.7501050Z 	/opt/hostedtoolcache/go/1.16.7/x64/src/testing/testing.go:1238 +0x5d8

2021-08-24T16:44:15.7541570Z goroutine 1940 [semacquire, 1 minutes]:
2021-08-24T16:44:15.7541887Z sync.runtime_Semacquire(0xc000175eb8)
2021-08-24T16:44:15.7542213Z 	/opt/hostedtoolcache/go/1.16.7/x64/src/runtime/sema.go:56 +0x45
2021-08-24T16:44:15.7542548Z sync.(*WaitGroup).Wait(0xc000175eb0)
2021-08-24T16:44:15.7542886Z 	/opt/hostedtoolcache/go/1.16.7/x64/src/sync/waitgroup.go:130 +0xe5
[...]

PR #2935, opened 2021-09-20
Unit test flake when rpc server stream not closed

Param-S:
#### Type of change
- Test update

#### Description
This Unit test fails randomly in more recent PR builds(https://dev.azure.com/Hyperledger/Fabric/_build/results?buildId=40922&view=logs&j=e306c17a-d139-54bf-a475-f5a11259cee7&t=1e3023a5-584f-52f3-49bc-66bd27d27b6d, https://dev.azure.com/Hyperledger/Fabric/_build/results?buildId=40557&view=logs&j=6b58850f-3858-5a05-33e2-5e41cbf03c4e&t=bddec1cf-ba37-5883-9c3e-fd1e8608f9a1, https://dev.azure.com/Hyperledger/Fabric/_build/results?buildId=40922&view=logs&j=e306c17a-d139-54bf-a475-f5a11259cee7&t=1e3023a5-584f-52f3-49bc-66bd27d27b6d)

Testcase: 
orderer/common/cluster_test.TestSendSubmitWithReport

Analysis/Observation: 
Problem 1: Testcase expects to fail at SubmitRequest after node server restarts with error EOF, but if restart not closed the streams before sendSubmit submitted; test fails. It is random failure. This PR introduces a delay between restart and submitRequest.

Problem 2: Calling t.FailNow() from other than main test go-routine, won't terminate all other subroutines. (which is called in this testcase from require.EqualError(t, err, io.EOF.Error())) This is the reason for the testcase hangs/blocked. https://pkg.go.dev/testing#B.FailNow.  Updated the testcase to signal & terminate the testcase from test main go-routine.

#### Related issues
#2835 

### DIFF
--- a/orderer/common/cluster/rpc_test.go
+++ b/orderer/common/cluster/rpc_test.go
@@ -66,12 +66,20 @@ func TestSendSubmitWithReport(t *testing.T) {
 	node2.stop()
 	node2.resurrect()
 
+	/*
+	 * allow the node2 to restart completely
+	 * if restart not complete, the existing stream able to successfully send
+	 * the next SubmitRequest which makes the testcase fails. Hence this delay
+	 * required
+	 */
+	time.Sleep(time.Second * 5)
+
 	var wg2 sync.WaitGroup
 	wg2.Add(1)
 
 	reportSubmitFailed := func(err error) {
-		require.EqualError(t, err, io.EOF.Error())
 		defer wg2.Done()
+		require.EqualError(t, err, io.EOF.Error())
 	}
 
 	err = node1RPC.SendSubmit(node2.nodeInfo.ID, &orderer.SubmitRequest{Channel: testChannel, Payload: &common.Envelope{Payload: []byte("2")}}, reportSubmitFailed)


yacovm:
Looks like it still failed:


```
--- FAIL: TestSendSubmitWithReport (10.03s)
FAIL
coverage: 95.8% of statements
FAIL	github.com/hyperledger/fabric/orderer/common/cluster	47.830s

```

Param-S:
Yes, there is mistake in the if condition. i will correct it.

orderer/common/cluster/rpc_test.go
```diff
@@ -66,18 +67,31 @@ func TestSendSubmitWithReport(t *testing.T) {
 	node2.stop()
 	node2.resurrect()
 
-	var wg2 sync.WaitGroup
-	wg2.Add(1)
```
yacovm, line 70: > Problem 2: Calling t.FailNow() from other than main test go-routine, won't terminate all other subroutines. (which is called in this testcase from require.EqualError(t, err, io.EOF.Error())) This is the reason for the testcase hangs/blocked. https://pkg.go.dev/testing#B.FailNow. Updated the testcase to signal & terminate the testcase from test main go-routine.

But the previous code was waiting for a wait group in the main goroutine, and that was facilitated using the defer **wg.Done()** in the callback. 

What you did, is replace the waitgroup with a channel and now you're waiting on that channel. 
How does that help the test not to get stuck?

Every execution that writes either 1 or 0 to the channel **c** in the new verson of the test, also releases the waitgroup in the old version of the test, and thus both versions of the test are then not stuck.

Moreover, every execution in which the test does not get stuck (in both the old version of the test and the new version), has to execute **reportSubmitFailed**, thus every execution that never finishes, never executes **reportSubmitFailed**. 
So I'm not sure how your code change helps the test getting unstuck. 

Maybe if you would add a select with a timeout, it would've solved the problem of the test getting stuck:
```
	reportSubmitFailed := func(err error) {
		require.EqualError(t, err, io.EOF.Error())
		close(c)
	}

	err = node1RPC.SendSubmit(node2.nodeInfo.ID, &orderer.SubmitRequest{Channel: testChannel, Payload: &common.Envelope{Payload: []byte("2")}}, reportSubmitFailed)
	require.NoError(t, err)

	select {
            case <- time.After(time.Second * 30):
                  t.Fail("timed out")
            case <- c:
        }

```

Param-S: **go subroutine B**
>       reportSubmitFailed := func(err error) {
> 		require.EqualError(t, err, io.EOF.Error())
> 		defer wg2.Done()
> 	}

	
**go subroutine A (testcase main routine)**
> 	wg2.Wait()

as per this existing code, when the condition is not met `require.EqualError(t, err, io.EOF.Error())`, it is expected to terminate the test. But it is called from`go subroutine B`, it terminates only the subroutine B, other subroutines continue to execute(which resulted into hang & test timeout). As per my understanding, calling this `require.EqualError(t, err, io.EOF.Error())` from subroutine B is not correct. Please correct me if I am wrong.

```
func EqualError(t TestingT, theError error, errString string, msgAndArgs ...interface{}) {
	if h, ok := t.(tHelper); ok {
		h.Helper()
	}
	if assert.EqualError(t, theError, errString, msgAndArgs...) {
		return
	}
	t.FailNow()     <<<<<<<<<<< this is called from go subroutine B, which terminates subroutine B, not others.
}
```

In the PR, it is modified to terminate(when condition is not met) or  continue the test(when condition is met) from the subroutine main.

yacovm: The current change that you have proposed, is only useful to scenarios where **other** tests are stuck, **and** this test fails. 

What I am trying to say, is that after your change - every time this test doesn't get stuck, and fails, it terminates the entire test suite. Previously, before your change, if this test didn't get stuck and failed, it didn't terminate the entire test suite. 

However, both the previous code and your code, lack a way of stopping the test after a timeout, and that can be achieved with a select statement as I outlined in my previous comment.

When a require assertion fails, it prints to the console the failure and the line in the code. In the CI logs that you have attached their links, I searched for a require failure, and didn't find one. Furthermore, I see the following stack trace:
```

goroutine 3131 [semacquire, 19 minutes]:
sync.runtime_Semacquire(0xc0002719f8)
	/opt/hostedtoolcache/go/1.16.7/x64/src/runtime/sema.go:56 +0x45
sync.(*WaitGroup).Wait(0xc0002719f0)
	/opt/hostedtoolcache/go/1.16.7/x64/src/sync/waitgroup.go:130 +0xe5

```

which means that the callback **reportSubmitFailed** was never called, and this will happen again in the new version of the code, because you do not cap the upper limit of waiting for the channel.

Param-S: > which means that the callback reportSubmitFailed was never called, and this will happen again in the new version of the code, because you do not cap the upper limit of waiting for the channel.

No, I am able to recreate this failure locally with existing code. By adding debug statements, I can see that the callback gets called each time.

```
	reportSubmitFailed := func(err error) {
		fmt.Println("reportSubmitFailed called successfully", time.Now(), err)
		debug.PrintStack()
		require.EqualError(t, err, io.EOF.Error())
```

Writing a stand-alone program, I am able to depict the same testcase behaviour. 

```
package conv

import (
	"fmt"
	"runtime/debug"
	"testing"
)

func FailTesting(t *testing.T, c chan int) {
	debug.PrintStack()
	fmt.Printf("printing from gorountine\n")
	// c <- 1
	t.FailNow()
}

func TestConvert(t *testing.T) {
	t.Log("Hello world")
	c := make(chan int)

	func() {
		fmt.Printf("calling failnow goroutine\n")
		go FailTesting(t, c)
		fmt.Printf("lanuched FailNow goroutine\n")
	}()

	val := <-c
	fmt.Printf("printing from main goroutine %v\n", val)
}
```

In the updated PR code, I am trying to avoid calling the FailNow from other goroutines.

yacovm: Oh I haven't noticed that the _defer wg2.Done()_ is **after** the require and not **before** it. That's the reason it is never called. 
```
	var wg2 sync.WaitGroup
	wg2.Add(1)

	reportSubmitFailed := func(err error) {
		require.EqualError(t, err, io.EOF.Error())
		defer wg2.Done()
	}

	err = node1RPC.SendSubmit(node2.nodeInfo.ID, &orderer.SubmitRequest{Channel: testChannel, Payload: &common.Envelope{Payload: []byte("2")}}, reportSubmitFailed)
	require.NoError(t, err)

	wg2.Wait()
```

Now if we just move the wg.Done() one line above, then it would solve the problem you are describing, look at this simple example:
```
func TestFailNow(t *testing.T) {
	var wg sync.WaitGroup
	wg.Add(1)

	go func() {
		defer wg.Done()
		t.FailNow()
	}()
	wg.Wait()
}
```

It's much simpler than your proposition and it also fails despite the fail directive not being in the main goroutine.

Param-S: Yes, It solves the problem. Updated the PR.

yacovm (approved):

denyeart:

denyeart:

PR #2643, opened 2021-06-04
FAB-18482 Unable to specify peer's chaincode.externalBuilders as an env variable

kopaygorodsky:
#### Type of change

- Bug fix

#### Description

As the HLF doc says, it's possible to override each config param via env variables, however, it isn't true for chaincode.externalBuilders 

Also, added a small expectation to a test for `chaincode.system` configuration

#### Related issues

https://jira.hyperledger.org/browse/FAB-18482

PR #1511, opened 2020-07-01
[FAB-17508] Removing viper library from viperutil

nbasker:
Type of change

Improvement (improvement to code by avoiding viper library that is not
needed in this context)

Description

Avoiding the overlap in the implementation of getKeysRecursively and
viper.Get() function. The sub-keys needed by getKeysRecursively is
readily available and the usage of viper.Get() can be avoided.

The reference to viper library is removed from the following modules
1. orderer/common/localconfig
2. internal/configtxgen/genesisconfig
3. common/viperutil

A ConfigParser class is created in viperutil/config_utils.go to achieve
the needed funtionality.

Additional details

Unit tests done are the following

go test -race -count 1 -v -run Test github.com/hyperledger/fabric/orderer/common/localconfig | tail -2
PASS
ok      github.com/hyperledger/fabric/orderer/common/localconfig
1.260s
go test -race -count 1 -v -run Test github.com/hyperledger/fabric/internal/configtxgen/genesisconfig | tail -2
PASS
ok      github.com/hyperledger/fabric/internal/configtxgen/genesisconfig
1.115s
go test -race -count 1 -v -run Test github.com/hyperledger/fabric/common/viperutil | tail -2
PASS
ok      github.com/hyperledger/fabric/common/viperutil  0.793s
go test -race -count 1 -v -run Test github.com/hyperledger/fabric/orderer/consensus/etcdraft | tail -2
PASS
ok      github.com/hyperledger/fabric/orderer/consensus/etcdraft
52.528s
go test -race -count 1 -v -run Test github.com/hyperledger/fabric/orderer/consensus/etcdraft | grep SUCCESS!
SUCCESS! -- 107 Passed | 0 Failed | 0 Pending | 0 Skipped

Related issues

https://jira.hyperledger.org/browse/FAB-17508

Signed-off-by: NicholasB <nbasker@gmail.com>

#### Type of change

- Improvement (improvement to code, performance, etc)

#### Description

Avoiding the overlap in the implementation of getKeysRecursively and
viper.Get() function. 

#### Additional details

Unit Tests Done

#### Related issues

https://jira.hyperledger.org/browse/FAB-17508

PR #2945, opened 2021-09-25
[Release-2.2 BP FAB-2643]

kopaygorodsky:
This is a backport of pull request #2643

This backport required upgrading viper to `v1.1.1` as in the main branch, cherry-pick some related fixes to this upgrade. 

#1523  (only https://github.com/hyperledger/fabric/pull/1523/commits/7dee9f7f0403e4ce993a601a6cf2545a65ebbf0b "Use UnmarshalKey to retrieve peer BCCSP config" to fix pkcs11 suite)
#1511 (fully)

nothing new was added, just exact backport of #2643 with a bit of #1523 and full #1511


kopaygorodsky:
PR is ready to be reviewed.

@yacovm @denyeart This failing unit test passes on my local machine + it passed one time in the pipeline. Could you check it pls? https://dev.azure.com/Hyperledger/Fabric/_build/results?buildId=41478&view=logs&j=6b58850f-3858-5a05-33e2-5e41cbf03c4e&t=bddec1cf-ba37-5883-9c3e-fd1e8608f9a1

kopaygorodsky:
Unit tests were failing for some unknown reason to me.  Locally everything passed, in the pipeline just 1 out of 10 times. 
The second commit fixes the problem somehow, I partially cherry-picked it from https://github.com/hyperledger/fabric/commit/694bbca07201e6f8e7f07771260f67dd5d7226a4

yacovm:
Thanks a lot for your effort, but, I'm honestly not 100% sure that we should cherry pick this.

Even though the documentation says that is possible, it seems to me more like a corner case of a feature.

Is there a specific reason we should cherry pick this to 2.2, and not let users just move to 2.3 if they really need it?

@kopaygorodsky do you have a specific reason you're backporting this? Do you need it in your environment or something?

 @denyeart I saw you initiated the auto-backport but I'm not sure...

kopaygorodsky:
@yacovm originally I made this fix because it was so annoying mounting `core.yaml` as a file.  I needed to keep it up to date with the other values after patches. It looked like a bug to me which should be fixed in the LTS version. 
Also, yes, @denyeart initiated this backport :)

denyeart:
My thinking was if the backport is simple and focused, why not... but if it has to touch 125 files then it is no longer a simple and focused fix, and the risk of destabilizing the LTS release seems to outweigh the benefit. My preference therefore would be not to merge it, apologies for not being more clear up front!

kopaygorodsky:
@denyeart what I can do is slightly different implementation with manual parsing of a YAML string, rather than using decode hook.  Does it make sense?  The only problem is that [this](https://github.com/kopaygorodsky/fabric/blob/dd57d645ab29b31c7ae8f95674cb390f45097cc5/core/chaincode/config_test.go#L50) test won't pass because of a bug in viper.  Currently, it's not working as well.

denyeart:
@kopaygorodsky I was willing to simply close it, but if this issue is impacting you and you'd like to invest the time, then go ahead with a smaller and more focused fix.

mastersingh24 (changes requested):
This is way too many files to touch for a backport.  Maybe you can try to make a more focused change?

kopaygorodsky:
@mastersingh24 yes, I've already found a way to avoid the upgrade of viper, but I'm stuck with a test case that does not pass on the old viper. 
https://github.com/hyperledger/fabric/blob/2f9c167b65edbd99ebb6512e6de077487f907230/core/chaincode/config_test.go#L50 
I added this test case as part of #2643 and it fails on 2.2. I'll try to find some middle ground to fix it.

ale-linux:
This PR Is a tad old and has received no activity in a while. Should we close it? @kopaygorodsky what's the status? Thanks

kopaygorodsky:
Closing this PR in favor of #3534

PR #2947, opened 2021-09-28
Improve an error message in `InstallChaincode`

satota2:
This patch adds the package ID to the error message in the case of 'chaincode already successfully installed'.

#### Type of change

#### Description

In case a 'chaincode already successfully installed' error occurs when installing a chaincode, Fabric admins cannot to know the package ID. In this case, if multiple chaincode packages with the same label are installed, it is not possible to identify them later by using `queryinstalled`.

So, this patch adds the package ID to the error message in the case of 'chaincode already successfully installed'.


denyeart (approved):

denyeart:
Thank you!

PR #2951, opened 2021-09-28
Randomize selection of orderer nodes with retry

### LINKED ISSUE #2912: Gateway: randomize selection of orderer nodes with retry
Currently, the gateway's submit function simply selects the first orderer node that is returned by discovery.
This list or orderers should be randomized to support improved load balancing.
Retry logic should be implemented such that if the selected orderer fails to return a success code, then the next orderer in the list should be tried.  If no orderers succeed, then return the error to the client.

andrew-coleman:
In the Submit() API method, the list of available orderers has been randomized to support improved load balancing.
Retry logic has been added such that if the selected orderer fails to return a success code, then the next orderer in the list is tried. If no orderers succeed, then return an error (containing details from each orderer) to the client.

Resolves #2912 

### DIFF
--- a/internal/pkg/gateway/api.go
+++ b/internal/pkg/gateway/api.go
@@ -9,6 +9,7 @@ package gateway
 import (
 	"context"
 	"fmt"
+	"math/rand"
 	"sync"
 
 	"github.com/golang/protobuf/proto"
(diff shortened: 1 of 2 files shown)


internal/pkg/gateway/api.go
```diff
@@ -279,31 +280,46 @@ func (gs *Server) Submit(ctx context.Context, request *gp.SubmitRequest) (*gp.Su
 		return nil, status.Errorf(codes.Unavailable, "no orderer nodes available")
 	}
 
-	orderer := orderers[0] // send to first orderer for now
+	// randomly shuffle orderers
+	rand.Shuffle(len(orderers), func(i, j int) { orderers[i], orderers[j] = orderers[j], orderers[i] })
```
yacovm: Instead of shuffling perhaps use [rand.Perm](https://pkg.go.dev/math/rand#Perm) ? It's shorter syntax and conveys better what we want to accomplish

andrew-coleman: Yes, that's definitely neater.  Thanks.

yacovm (approved):

PR #2955, opened 2021-09-28
Refactor idemix implementation

ale-linux:
This PR removes the internal implementation of the idemix protocol and uses
an external dependency that contains the same implementation.

Signed-off-by: Alessandro Sorniotti <aso@zurich.ibm.com>

#### Type of change

#### Description

The idemix is factored out of the repo and now imported as a dependency.


integration/nwo/components.go
```diff
@@ -32,7 +34,7 @@ func (c *Components) Discover() string {
 }
 
 func (c *Components) Idemixgen() string {
-	return c.Build("github.com/hyperledger/fabric/cmd/idemixgen")
```
yacovm, line 35: why don't we build it anymore though? can't we do `go install` or something instead?

ale-linux: you mean we could go get the dependency in the makefile and then build it here like we do the others? Yeah, that actually isn't a bad idea.. I'll do it - thx

yacovm:
LGTM, perhaps @adecaro can also take a look. 

Do we have any concern of non-determinism here, or are we OK since:

> and uses an external dependency that contains the same implementation.

can you elaborate on that?

integration/nwo/components.go
yacovm: Yeah something like that... but I tried with `go install` and it didn't like it

ale-linux: Fixed now - wdyt?

ale-linux:
> LGTM, perhaps @adecaro can also take a look.
> 
> Do we have any concern of non-determinism here, or are we OK since:
> 
> > and uses an external dependency that contains the same implementation.
> 
> can you elaborate on that?

good points. So:

- the new module does indeed add new functionalities (e.g. extensions to what's being proved/verified), but these are only usable if the caller of the affected functions (e.g. `Sign` or `Verify`) requests it by supplying a different signer/verifier op; this isn't done by the idemix msp, which keeps using the "old" opts which ensure the "old" behaviour
- furthermore, we have added a set of tests in the new repo to assert that all artefacts (CA PK, params, CA SK, user PK, user SK, nyms, signatures etc) produced by the "old" implementation would cause the same behaviour when supplied to the appropriate function

yacovm:
Looks good, let's wait for @adecaro to make a pass too.

adecaro:
LGTM. This PR will make Fabric lighter and will give more opportunities to enhance the idemix stack.

yacovm (approved):

bestbeforetoday:
@ale-linux Would you be interested in adding a client-side Idemix signing implementation to the [fabric-gateway](https://github.com/hyperledger/fabric-gateway) client API?

All that needs to be provided at the client end is a mechanism to create a signing function of the form: `func(digest []byte) ([]byte, error)`

Existing client signing implementations and the API provided to create them can be found at [https://github.com/hyperledger/fabric-gateway/tree/main/pkg/identity](https://github.com/hyperledger/fabric-gateway/tree/main/pkg/identity):

- [identity.NewPrivateKeySign()](https://github.com/hyperledger/fabric-gateway/blob/a92971dcfd1e7f9fd050b1f3b07174cbf3822cc9/pkg/identity/sign.go#L19-L20) for a private key signer.
- [identity.NewHSMSignerFactory()](https://github.com/hyperledger/fabric-gateway/blob/a92971dcfd1e7f9fd050b1f3b07174cbf3822cc9/pkg/identity/hsmsign.go#L37-L39) (and [NewHSMSigner()](https://github.com/hyperledger/fabric-gateway/blob/a92971dcfd1e7f9fd050b1f3b07174cbf3822cc9/pkg/identity/hsmsign.go#L56-L60) on the factory) to create a signer backed by a Hardware Security Module.

If a specific hash algorithm is required for an Idemix signer, that can be added [here](https://github.com/hyperledger/fabric-gateway/tree/main/pkg/hash).

We would also want a scenario test to ensure an Idemix signer worked at runtime with a real Fabric network, but this should require very little code to extend the existing [Cucumber scenario tests](https://github.com/hyperledger/fabric-gateway/tree/main/scenario), similar to how the HSM signer is tested.

ale-linux:
@bestbeforetoday that sounds indeed like an interesting feature to add. It should be simple enough - see this test for example

https://github.com/IBM/idemix/blob/main/bccsp/bccsp_test.go#L98

lines 98 through 189 basically show all the calls needed to fully set up an idemix stack and sign with it. 

Here instead

https://github.com/IBM/idemix/blob/main/bccsp/bccsp_test.go#L853

we show how to load an idemix configuration from a client local msp store.

PR #2975, opened 2021-10-07
Limit TestBlockPullerBadBlocks pullblock time

Param-S:
#### Type of change
- Test update

#### Description
Testcase times out after 20 minutes as the client is not able
to connect to the server.  The rootcause for the server connection
failure is not known yet.  

BlockPuller provides an option to limit the retries to get the blocks
This will help here when the client failed to connect & fail the testcase
fast. This PR set the retries to 5 times. 

#### Related issues
#2835 

### DIFF
--- a/orderer/common/cluster/deliver_test.go
+++ b/orderer/common/cluster/deliver_test.go
@@ -1086,12 +1086,24 @@ func TestBlockPullerBadBlocks(t *testing.T) {
 				return nil
 			}))
 
-			bp.PullBlock(10)
+			endPullBlock := make(chan struct{})
+			go func() {
+				bp.PullBlock(10)
+				close(endPullBlock)
+			}()
 
 			select {
 			case <-detectedBadBlockDone:
 			case <-time.After(10 * time.Second):
 				t.Fatalf("expected %q to be logged but it was not seen", testCase.expectedErrMsg)
+			}
+
+			select {
+			case <-endPullBlock:
+			case <-time.After(10 * time.Second):
+				// Signal PullBlock to give up on the retries & mark the test as fail
+				close(bp.StopChannel)
+				t.Fatalf("PullBlock did not complete within time")
 			}
 
 			bp.Close()


yacovm (changes requested):
Why not simply have the block puller pull the blocks asynchronously and have a select with a timeout statement that waits for either it to finish or for a timeout to expire? 

Also the title is not suitable to the content. "Orderer unittest flake in deliver_test" - you are not fixing here a flake.  Maybe say instead that you are bounding the maximum execution time of the test?

orderer/common/cluster/deliver_test.go
```diff
@@ -1069,11 +1069,10 @@ func TestBlockPullerBadBlocks(t *testing.T) {
 
 			// Wait until the block is pulled and the malleability is detected
 			detectedBadBlockDone := make(chan struct{})
+
 			var closeOnce sync.Once
 			bp.Logger = bp.Logger.WithOptions(zap.Hooks(func(entry zapcore.Entry) error {
 				if strings.Contains(entry.Message, fmt.Sprintf("Failed pulling blocks: %s", testCase.expectedErrMsg)) {
-					closeOnce.Do(func() { close(detectedBadBlockDone) })
```
yacovm: I don't understand why you moved this out of the logger detection closure. 
The code detects a failure and closes the channel. 

Now in the new version, you close it yourself, so you make the test artificially succeed :(

Param-S: As per my understanding, `bp.PullBlock(10)` completes only when successful execution of this block of code`bp.Logger.WithOptions(zap.Hooks(func(entry zapcore.Entry) error`. Hence, I considered that as implicit successful bad block detection and moved down `close(detectedBadBlockDone)` after PullBlock.

yacovm: But what if there is a bug, and it completes without having that callback be executed? We can't know for sure that a piece of code will end its execution.

Param-S: Agree, now I moved down the pullBlock into go routine.

orderer/common/cluster/deliver_test.go
```diff
@@ -1086,7 +1086,7 @@ func TestBlockPullerBadBlocks(t *testing.T) {
 				return nil
 			}))
 
-			bp.PullBlock(10)
```
yacovm, line 1089: I think we need to synchronize memory coherency otherwise we'll have a data race. 

So let's put the asynchronous block pulling in a goroutine that ends with a waitgroup waiting and then wait for that waitgroup to end, so that we have a happens-before relation.

Param-S: Yes, Updated the PR.

orderer/common/cluster/deliver_test.go
```diff
@@ -1086,10 +1086,22 @@ func TestBlockPullerBadBlocks(t *testing.T) {
...
+			endPullBlock := make(chan struct{})
+			var pullOnce sync.Once
```
yacovm: Why do you need sync.Once? You close the edPullBlock channel only once anyway, there is no way it can be closed twice.

orderer/common/cluster/deliver_test.go
```diff
@@ -1086,10 +1086,22 @@ func TestBlockPullerBadBlocks(t *testing.T) {
...
+				pullOnce.Do(func() { close(endPullBlock) })
+			}()
 
 			select {
 			case <-detectedBadBlockDone:
+				select {
+				case <-endPullBlock:
+				case <-time.After(10 * time.Second):
+					// Signal PullBlock to give up on the retries & mark the test as fail
+					close(bp.StopChannel)
```
yacovm: Why do you need to close the StopChannel if you already have the bp.Close down below in line 1109?

orderer/common/cluster/deliver_test.go
```diff
@@ -1086,10 +1086,22 @@ func TestBlockPullerBadBlocks(t *testing.T) {
...
+			endPullBlock := make(chan struct{})
+			var pullOnce sync.Once
+			go func() {
+				bp.PullBlock(10)
+				pullOnce.Do(func() { close(endPullBlock) })
+			}()
 
 			select {
 			case <-detectedBadBlockDone:
+				select {
```
yacovm: I don't think we need a select inside a select, it's too complex. 

All we need is to synchronize the memory. Notice that when the select finishes in line 1096 we can then do another select. An alternative for memory synchronization can also be a lock. You don't have to use a channel.

orderer/common/cluster/deliver_test.go
Param-S: when times out, following statement `t.Fatalf("PullBlock did not complete within time")` terminates the current go routine, hence, it does not reach to line `bp.Close()`

orderer/common/cluster/deliver_test.go
Param-S: Since, the code block is within an testcase loop, i enclosed it within sync.Once.  I removed it now.

orderer/common/cluster/deliver_test.go
yacovm: So what? the test terminates doesn't it? Why do we care about other goroutines that are not the main goroutine if the test fails?

Param-S: Sorry, If i have not explained it clearly here.  When the PullBlock is not completed within the time, we want to mark that particular test(subtest) as failed.  Hence, in the select case, signaled the blocking PullBlock & then Fail the testcase by calling `t.Fatalf`

We see the following output when the PullBlock failed to return/blocked:

```
--- FAIL: TestBlockPullerBadBlocks (60.02s)
    --- FAIL: TestBlockPullerBadBlocks/nil_header (10.00s)
        deliver_test.go:1106: PullBlock did not complete within time
    --- FAIL: TestBlockPullerBadBlocks/nil_data (10.00s)
        deliver_test.go:1106: PullBlock did not complete within time
```

yacovm (approved):

orderer/common/cluster/deliver_test.go
yacovm: I'm not sure I understand, but we need to make progress

Issue #2976, opened 2021-10-07
Include the chaincode hash when installing an already existing chaincode


varadgit:
When installing a new chaincode on the peer you do get the package id of the chaincode which is good... this allows customers to use that data to go through the approve/commit steps.

However the issue is when they are reinstalling an existing chaincode there is nothing in the response to indicate the package id(hash). Install in that case returns a message that the chaincode already exists. We should include the package id of the chaincode in the response when it is already installed on the peer.

satota2:
@denyeart I want to take the issue.

satota2:
@denyeart @varadgit
I have recently submitted the following patch and it has been merged.
I think this may have solved this Issue, but is it still not enough?
https://github.com/hyperledger/fabric/commit/cf341ee49aabae95f583add859ac9186e0a42bb4
 
I think another good way is to add a command to calculate the package ID for
the packaged source code (e.g., `peer lifecycle chaincode calulatepackageid`).
If you think this command is also useful, I would like to make the patches, what do you think about the idea?

denyeart:
@satota2 It looks like your PR may satisfy the request from @varadgit 

Note that `peer lifecycle chaincode queryinstalled` already can return the package id for installed chaincodes:
https://hyperledger-fabric.readthedocs.io/en/latest/commands/peerlifecycle.html#peer-lifecycle-chaincode-queryinstalled-example

I guess you are asking about adding a new function to find the package id from a packaged chaincode rather than an installed chaincode. That sounds like a good idea. Here is the code to compute the package id:
https://github.com/hyperledger/fabric/blob/main/core/chaincode/persistence/persistence.go#L147-L148

satota2:
@denyeart  Thank you for your comments and useful information.
 
> I guess you are asking about adding a new function to find the package id from a packaged chaincode rather than an installed chaincode. 

Yes, I'm thinking of adding that function.
 
Although `queryinstalled` returns the package ID for installed chaincodes, I think this new function will be useful in some cases.
For example, the following cases: 
- When multiple chaincode packages with the same label name are installed, it is possible to identify which ID corresponds to which package later.
- To check whether a particular chaincode package is installed or not without installing that package (without executing `chaincode install`).
 
> That sounds like a good idea.

Thanks for agreeing with the idea.
I'm going to implement this function and submit the patches, referring to the code you gave me.

satota2:
[...]

PR #2981, opened 2021-10-14
Add `calculatepackageid` command

satota2:
This patch `calculatepackageid` command to calculate the package ID for a packaged chaincode rather than an installed chaincode.

#### Type of change

- New feature

#### Description

This patch `calculatepackageid` command to calculate the package ID for a packaged chaincode rather than an installed chaincode.

The new command will be useful, for example, the following cases:
* When multiple chaincode packages with the same label name are installed,
it is possible to identify which ID corresponds to which package later.
* To check whether a particular chaincode package is installed or not on a peer
without installing that package.

#### Additional details

After this patch is merged, I'm going to submit the following related patches as separate PRs in sequence:

- For the integration tests:
  - https://github.com/satota2/fabric/commit/666a5ca4b941ee92c2da665d619c43df9350399d

#### Related issues

- https://github.com/hyperledger/fabric/issues/2976


internal/peer/lifecycle/chaincode/calculatepackageid.go
```diff
@@ -0,0 +1,114 @@
...
+	}
+
+	metadata, _, err := persistence.ParseChaincodePackage(pkgBytes)
+	if err != nil {
+		return errors.WithMessage(err, "could not parse as a chaincode install package")
+	}
+
+	packageID := persistence.PackageID(metadata.Label, pkgBytes)
+
+	fmt.Fprintf(p.Writer, "package-id: '%s'\n", packageID)
```
denyeart: For consistency with existing commands such as queryinstalled, you could return format like this:
Package ID: basic_1:7db6094164318df5a22fdfb48050a13928f4e3dbb17430575f1699f1b305410b

```suggestion
	fmt.Fprintf(p.Writer, "Package ID: %s\n", packageID)
```

denyeart (commented):
Nice job! Just one nit comment.

Also let's have @mbwhite take a quick look since he oversees chaincode/lifecycle these days.

mbwhite:
@satota2 yes very useful.... (@jkneubuh was wanting this very command a few days ago).

Could I make a suggestion that would be very useful for scripting - a quiet mode. Whereby the only output is the package-id. Reason is you can do this..

```
PACKAGE_ID=$(peer lifecycle chaincode calculatepackageid --quiet  file.tgz)
peer lifecycle chaincode approveformyorg ${PACKAGE_ID}
```

Otherwise we need to either write it out in `json` (`-o json`) and then use JQ to parse out the result.. or use awk to strip off the front part. (something I'd like to get added to the other commands)

internal/peer/lifecycle/chaincode/calculatepackageid.go
satota2: @denyeart Thank you for your comment. I'm going to revise the format with your suggestion.

satota2:
@denyeart @mbwhite Thank you for your comments.

@mbwhite
I agree with you that these options are very useful for operations.
Actually, I planed to add the `-o json` option after this PR and the separated PR for the integration tests are merged.
The quiet option also sounds useful, so I would like to add this option as well.

Could I add these options as separate PRs? If yes, I'd like to a quick minimal fix of this patch and resubmit it first, then submit additional PRs for new options sequentially.

mbwhite:
@satota2  thanks.. that's good with me...

mbwhite (dismissed):

denyeart:
Actually based on the function name, it is clear that a package id will be returned, so we could just strip off the "Package ID:" prefix, even without a separate quiet mode. What do you think @mbwhite ?

satota2:
@denyeart, @mbwhite
I resubmit the patch that changes the prefix format from `package-id: '%s'` to `Package ID: %s`.
I'm going to make some separate PRs to add the output options.

mbwhite (approved):

denyeart (approved):

denyeart:
Thanks @satota2 !

jkneubuh:
I really like this PR;  I am late to the party but I absolutely approve.   
I also would prefer the use of a "silent" mode, preferably by default. 

ありがとうございます Sato-さん!

satota2:
@mbwhite, @denyeart Thank you for your approvals.

@jkneubuh コメントありがとうございます (Thank you for your comments.)
I'm working on a new PR that add output options including a silent mode.
I'm going to submit the above PR after another patch for the integration test is merged.
I would be pleased to discuss the default behavior on the new PR for the output options.
I'll let you and the other reviewers know about that PR using Mention.

denyeart:
@satota2 It appears there is consensus on just returning the package id without any prefix by default. That way we won't need a quiet mode at all. Looking forward to the next PRs!

satota2:
@denyeart
Thank you for your comment. I'm going to modify the default output format based on the consensus and submit the patches after the patch (#3005) is merged.

PR #2985, opened 2021-10-15
Gateway endorsement retry logic

### LINKED ISSUE #2914: Gateway: retry logic for endorse method
Currently, the gateway's endorse method will make a discovery call and select a single layout from the discovery results.  A layout consist of groups of peers where (say) one peer is required from each group.  The current logic is to sort each group by block height and select the peer with the highest block number. If this fails due to connectivity/infrastructure reasons, then a complete set of endorsement won't be created and an error is returned to the client.
This enhancement will handle these errors by attempting to select then next peer in that ordered group, and retry the query. If no peers within a group can successfully process the proposal, then the error is returned to the client.

Note this logic won't apply to errors that are raised by the chaincode since it is assumed that all instances will return this same error (e.g. wrong number/type of arguments, etc).  However, a follow-on enhancement could be designed whereby the gateway repeatedly requests endorsements across available peers until it can form a matching set which satisfy the endorsement policy.

andrew-coleman:
Rather than selecting one layout from the discovery endorsement plan and failing if one of the endorsers fails, this commit attempts to create a set of endorsements by retrying the proposal on other endorser until one of the layouts is satisfied.
Additionally, rather than connect to all peers in a channel once on first usage and then never update that cache, this commit adds support for later additions and removals to/from the cache and closing stale connections to peers.

Resolves #2914 

### DIFF
--- a/internal/pkg/gateway/api.go
+++ b/internal/pkg/gateway/api.go
@@ -26,8 +26,8 @@ import (
 )
 
 type endorserResponse struct {
-	pr  *peer.ProposalResponse
-	err *gp.ErrorDetail
+	endorsementSet []*peer.ProposalResponse
+	err            *gp.ErrorDetail
 }
 
 // Evaluate will invoke the transaction function as specified in the SignedProposal
@@ -41,7 +41,7 @@ func (gs *Server) Evaluate(ctx context.Context, request *gp.EvaluateRequest) (*g
 		return nil, status.Errorf(codes.InvalidArgument, "failed to unpack transaction proposal: %s", err)
 	}
 
-	err = gs.registry.registerChannel(channel)
+	err = gs.registry.connectChannelPeers(channel, false)
 	if err != nil {
 		return nil, status.Errorf(codes.Unavailable, "%s", err)
 	}
@@ -104,11 +104,6 @@ func (gs *Server) Endorse(ctx context.Context, request *gp.EndorseRequest) (*gp.
 		return nil, status.Errorf(codes.InvalidArgument, "failed to unpack transaction proposal: %s", err)
 	}
 
-	err = gs.registry.registerChannel(channel)
-	if err != nil {
-		return nil, status.Errorf(codes.Unavailable, "%s", err)
-	}
-
 	ctx, cancel := context.WithTimeout(ctx, gs.options.EndorsementTimeout)
 	defer cancel()
 
@@ -118,11 +113,11 @@ func (gs *Server) Endorse(ctx context.Context, request *gp.EndorseRequest) (*gp.
 		}},
 	}
 
-	var endorsers []*endorser
-	var responses []*peer.ProposalResponse
+	var plan *plan
+	var endorsements []*peer.ProposalResponse
 	if len(request.EndorsingOrganizations) > 0 {
 		// The client is specifying the endorsing orgs and taking responsibility for ensuring it meets the signature policy
-		endorsers, err = gs.registry.endorsersForOrgs(channel, chaincodeID, request.EndorsingOrganizations)
+		plan, err = gs.registry.planForOrgs(channel, chaincodeID, request.EndorsingOrganizations)
 		if err != nil {
 			return nil, status.Errorf(codes.Unavailable, "%s", err)
 		}
@@ -130,35 +125,43 @@ func (gs *Server) Endorse(ctx context.Context, request *gp.EndorseRequest) (*gp.
 		// The client is delegating choice of endorsers to the gateway.
 
 		// 1. Choose an endorser from the gateway's organization
-		var firstEndorser *endorser
-		es, ok := gs.registry.endorsersByOrg(channel, chaincodeID)[gs.registry.localEndorser.mspid]
-		if !ok {
+		plan, err = gs.registry.planForOrgs(channel, chaincodeID, []string{gs.registry.localEndorser.mspid})
+		if err != nil {
 			// No local org endorsers for this channel/chaincode. If transient data is involved, return error
 			if hasTransientData {
 				return nil, status.Error(codes.FailedPrecondition, "no endorsers found in the gateway's organization; retry specifying endorsing organization(s) to protect transient data")
 			}
 			// Otherwise, just let discovery pick one.
-			endorsers, err = gs.registry.endorsers(channel, defaultInterest, "")
+			plan, err = gs.registry.endorsementPlan(channel, defaultInterest, nil)
 			if err != nil {
 				return nil, status.Errorf(codes.Unavailable, "%s", err)
 			}
-			firstEndorser = endorsers[0]
-		} else {
-			firstEndorser = es[0].endorser
 		}
+		firstEndorser := plan.endorsers()[0]
 
 		gs.logger.Debugw("Sending to first endorser:", "channel", channel, "chaincode", chaincodeID, "MSPID", firstEndorser.mspid, "endpoint", firstEndorser.address)
 
 		// 2. Process the proposal on this endorser
-		firstResponse, err := firstEndorser.client.ProcessProposal(ctx, signedProposal)
-		if err != nil {
-			return nil, wrappedRpcError(err, "failed to endorse transaction", errorDetail(firstEndorser.endpointConfig, err))
+		var firstResponse *peer.ProposalResponse
+		var errDetails []proto.Message
+		for firstResponse == nil && firstEndorser != nil {
+			firstResponse, err = firstEndorser.client.ProcessProposal(ctx, signedProposal)
+			if err != nil {
+				logger.Debugw("Endorse call to endorser failed", "channel", request.ChannelId, "txid", request.TransactionId, "endorserAddress", firstEndorser.endpointConfig.address, "endorserMspid", firstEndorser.endpointConfig.mspid, "error", err)
+				errDetails = append(errDetails, errorDetail(firstEndorser.endpointConfig, err))
+				gs.registry.removeEndorser(firstEndorser)
+				firstEndorser = plan.retry(firstEndorser)
+			} else if firstResponse.Response.Status < 200 || firstResponse.Response.Status >= 400 {
+				logger.Debugw("Endorse call to endorser returned failure", "channel", request.ChannelId, "txid", request.TransactionId, "endorserAddress", firstEndorser.endpointConfig.address, "endorserMspid", firstEndorser.endpointConfig.mspid, "status", firstResponse.Response.Status, "message", firstResponse.Response.Message)
+				err := fmt.Errorf("error %d, %s", firstResponse.Response.Status, firstResponse.Response.Message)
+				endpointErr := errorDetail(firstEndorser.endpointConfig, err)
+				errDetails = append(errDetails, endpointErr)
+				firstEndorser = plan.retry(firstEndorser)
+				firstResponse = nil
+			}
 		}
-		if firstResponse.Response.Status < 200 || firstResponse.Response.Status >= 400 {
-			err := fmt.Errorf("error %d, %s", firstResponse.Response.Status, firstResponse.Response.Message)
-			endpointErr := errorDetail(firstEndorser.endpointConfig, err)
-			errorMessage := "failed to endorse transaction: " + detailsAsString(endpointErr)
-			return nil, rpcError(codes.Aborted, errorMessage, endpointErr)
+		if firstEndorser == nil || firstResponse == nil {
+			return nil, rpcError(codes.Aborted, "failed to endorse transaction, see attached details for more info", errDetails...)
 		}
 
 		// 3. Extract ChaincodeInterest and SBE policies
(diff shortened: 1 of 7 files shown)


internal/pkg/gateway/endpoint.go
```diff
@@ -60,9 +61,16 @@ func (ef *endpointFactory) newEndorser(pkiid common.PKIidType, address, mspid st
 	if connectEndorser == nil {
 		connectEndorser = peer.NewEndorserClient
 	}
+	close := func() error {
+		if conn != nil {
+			return conn.Close()
+		}
```
manish-sethi, line 67: will `conn != nil` ever be false?

internal/pkg/gateway/endorsement.go
```diff
@@ -0,0 +1,112 @@
...
+package gateway
+
+import (
+	"sync"
+
+	"github.com/hyperledger/fabric-protos-go/peer"
+)
+
+type layout struct {
+	required     map[string]int // mspid -> quantity
```
manish-sethi: `mspid -> quantity` seems you wanted to say `groupid -> quantity`

internal/pkg/gateway/endorsement.go
```diff
@@ -0,0 +1,112 @@
...
+		p.nextLayout++
+	}
+	if p.nextLayout >= len(p.layouts) {
+		return nil
+	}
+	var endorsers []*endorser
+	for mspid, qty := range p.layouts[p.nextLayout].required {
+		endorsers = append(endorsers, p.mspEndorsers[mspid][0:qty]...)
+		p.nextEndorser[mspid] = qty
+	}
```
manish-sethi: This use of variable name `mspid` is puzzling me. Shouldn't this be `groupid`?

internal/pkg/gateway/endorsement.go
```diff
@@ -0,0 +1,112 @@
...
+type layout struct {
+	required     map[string]int // mspid -> quantity
+	endorsements []*peer.ProposalResponse
+}
+
+type plan struct {
+	layouts      []*layout
+	mspEndorsers map[string][]*endorser // mspid -> endorsing peers
+	nextLayout   int
+	nextEndorser map[string]int
```
manish-sethi: `nextEndorser` seems to belong to a particular layout. Can you define it in the `layout` struct instead (just like other layout specific fields).

manish-sethi (commented):
The new organization of code is around the structs defined in file `endorsement.go`. However, I am not able to understand the use of terms in this file (see comments). So, either I am missing something very basic or the terms are not correct. I would like to pause here before looking at the changes built on top of this.

internal/pkg/gateway/endpoint.go
andrew-coleman: Yes I don't like this either, but in the unit tests `conn` is nil.  Unfortunately the grpc.ClientConn is a struct rather than an interface, so can't be mocked as I like to.  Happy if you've got other suggestions on this.

internal/pkg/gateway/endorsement.go
andrew-coleman: No, I meant mspid.  The structure is built in this function: https://github.com/andrew-coleman/fabric/blob/endorser_retry/internal/pkg/gateway/registry.go#L53
The group IDs from the discovery service are translated into MSP IDs so that they can be correlated with the endorser objects.  This hasn't really changed, it's just that I'm now building a structure that represents the whole endorsement plan whereas previously I was just choosing the first layout and looking up the mspid for that.

internal/pkg/gateway/endorsement.go
andrew-coleman: No. As above.  Maybe best we have a call to discuss?

internal/pkg/gateway/endorsement.go
andrew-coleman: No, it's common across all layouts.  For example, if we have 3 orgs each of which has 2 peers, and the endorsement policy requires 2 out of 3 to endorse, then the plan will look like this:
```
layouts: [
  {"Org1": 1, "Org2": 1}
  {"Org1": 1, "Org3": 1}
  {"Org2": 1, "Org3": 1}
]
mspEndorsers: {
  "Org1": ["Peer1.Org1", "Peer2.Org1"]
  "Org2": ["Peer1.Org2", "Peer2.Org2"]
  "Org3": ["Peer1.Org3", "Peer2.Org3"]
}
```
Starting with the first layout, the gateway will seek endorsement from Org1 and Org2, and will select Peer1.Org1 and Peer1.Org2 to process the proposal.
Assume that Peer1.Org1 succeeds, it will go through each layout and decrement the number of endorsers (quantity) still required (removing it if gets to zero).  This will result in the following layouts struct:
```
layouts: [
  {"Org2": 1}
  {"Org3": 1}
  {"Org2": 1, "Org3": 1}
]
```
Assume that Peer1.Org2 fails, it will increment the nextEndorser[Org2] and retry on Peer2.Org2.  Assume this retry succeeds, it will go through each layout again as above, this time declaring that the first layout is satisfied and returning the list of endorsements collected by that layout.
However, if that retry on Peer2.Org2 also failed, then because there are no more endorsers for that MSP (org), it will go through the list of layouts and remove all layouts that contain that MSP.  That will leave us with the following layouts struct:
```
layouts: [
  {"Org3": 1}
]
```
The gateway will then select the next (first) peer from Org3 to get endorsement, retrying on its next peer, if necessary.

internal/pkg/gateway/endorsement.go
yacovm: It's actually not *always* MSP ID, but it can also be any kind of principal. The discovery protobuf API is compatible with any abstract policy, for instance, a policy where specific peers own specific assets, etc.

internal/pkg/gateway/endorsement.go
```diff
@@ -0,0 +1,119 @@
...
+
+type layout struct {
+	required     map[string]int // group -> quantity
+	endorsements []*peer.ProposalResponse
+}
+
+type plan struct {
+	layouts        []*layout
+	groupEndorsers map[string][]*endorser // group -> endorsing peers
+	groupIds       map[string]string      // peer pkiid -> group
```
manish-sethi, line 29: So, this assumes that one endorser is present in one group only. If not already mentioned, can you mention this  in the Descriptor proto?

internal/pkg/gateway/endorsement.go
```diff
@@ -0,0 +1,119 @@
...
+	nextLayout     int
+	size           int
```
manish-sethi, line 31: Not sure, if this is the right place for size as it is just the size of returned endorsers in the last invocation. The caller better can get `len(endorsers)`

internal/pkg/gateway/endorsement.go
```diff
@@ -0,0 +1,119 @@
...
+	for p.nextLayout < len(p.layouts) && p.layouts[p.nextLayout] == nil {
+		p.nextLayout++
+	}
+	if p.nextLayout >= len(p.layouts) {
+		return nil
+	}
+	var endorsers []*endorser
+	for group, qty := range p.layouts[p.nextLayout].required {
+		// remove the first qty endorsers from the front of this group
+		endorsers = append(endorsers, p.groupEndorsers[group][0:qty]...)
```
manish-sethi: Can't it cause a `panic` -- If the `retry` function causes reduction in the number of available endorsers in the group and a latter layout has more `qty`?

internal/pkg/gateway/registry.go
```diff
@@ -49,150 +49,149 @@ type endorserState struct {
...
+			quantityByGroup[group] = int(quantity)
+			if group == preferredGroup {
+				hasPreferredGroup = true
 			}
-			// try the next layout
-			continue
 		}
-
-		return endorsers, nil
+		if len(quantityByGroup) == 0 {
```
manish-sethi, line 129: will this ever be true as you already continue to outer loop at line 124?

internal/pkg/gateway/api.go
```diff
@@ -178,79 +181,78 @@ func (gs *Server) Endorse(ctx context.Context, request *gp.EndorseRequest) (*gp.
...
+	var errorDetails []proto.Message
+	for endorsements == nil {
+		// loop through the layouts until one gets satisfied
+		endorsers := plan.endorsers()
+		if endorsers == nil {
+			// no more layouts
+			break
+		}
+		var wg sync.WaitGroup
+		responseCh := make(chan *endorserResponse, plan.size+1)
```
manish-sethi: You create this channel of fixed size while in the goroutines below, one goroutine could write multiple responses (because of retries), eventually leading to block the processing... no?

internal/pkg/gateway/api.go
```diff
@@ -118,47 +113,55 @@ func (gs *Server) Endorse(ctx context.Context, request *gp.EndorseRequest) (*gp.
...
-		if !ok {
+		plan, err = gs.registry.planForOrgs(channel, chaincodeID, []string{gs.registry.localEndorser.mspid})
+		if err != nil {
 			// No local org endorsers for this channel/chaincode. If transient data is involved, return error
 			if hasTransientData {
 				return nil, status.Error(codes.FailedPrecondition, "no endorsers found in the gateway's organization; retry specifying endorsing organization(s) to protect transient data")
 			}
 			// Otherwise, just let discovery pick one.
-			endorsers, err = gs.registry.endorsers(channel, defaultInterest, "")
+			plan, err = gs.registry.endorsementPlan(channel, defaultInterest, nil)
```
manish-sethi, line 135: Though not added in this PR - This appears to cause unnecessarily computing the layouts while you do so below again with the real chaincode interest. Isn't it possible to straight get the plan once. This need not necessarily is to be fixed in this PR.

internal/pkg/gateway/endorsement.go
andrew-coleman: It does assume that.  This is a constraint of the discovery algorithm itself rather than the proto.  I've added a comment.

internal/pkg/gateway/endorsement.go
andrew-coleman: Actually I've modified the code that initialises this size.  It's meant to be the total number of endorsers across all groups so that the channel buffer size can be set accordingly (which you picked up on below)

internal/pkg/gateway/api.go
andrew-coleman: Good catch.  I've modified the calculation of size to be the total number of endorsers across all groups.

internal/pkg/gateway/registry.go
andrew-coleman: Under normal circumstances this should not be hit.  But I have a some tests that have been there for a while which test its behaviour with malformed endorsement descriptors.  In this case the Layout array contains a nil layout object.  I can't remember why this test went in, but we could take it out.

internal/pkg/gateway/endorsement.go
```diff
@@ -0,0 +1,142 @@
...
+	p.planLock.Lock()
+	defer p.planLock.Unlock()
+
+	// Initialize the size with the total number of endorsers across all plans.  This is required to
+	// determine the theoretical maximum channel buffer size when requesting endorsement across multiple goroutines.
+	if p.size == 0 {
+		for _, peers := range p.groupEndorsers {
+			p.size += len(peers)
+		}
+	}
```
manish-sethi: This clearly belongs to a constructor. Better to have a constructor on this type.

internal/pkg/gateway/endorsement.go
```diff
@@ -0,0 +1,142 @@
...
+		for p.nextLayout < len(p.layouts) && p.layouts[p.nextLayout] == nil {
+			p.nextLayout++
+		}
+		if p.nextLayout >= len(p.layouts) {
+			return nil
+		}
+		for group, qty := range p.layouts[p.nextLayout].required {
+			if qty > len(p.groupEndorsers[group]) {
+				// requires more group endorsers than available - abandon this layout
+				endorsers = nil
```
manish-sethi, line 73: Though you never look back but better to set the abandoned layout to nil to avoid `update` on this and for better debugging when you add some logging (see related comment in update function).

internal/pkg/gateway/endorsement.go
```diff
@@ -0,0 +1,142 @@
...
+	group := p.groupIds[endorser.pkiid.String()]
+	endorsers := p.groupEndorsers[group]
+
+	// remove the endorser from this group
+	for i, e := range endorsers {
+		if bytes.Equal(e.pkiid, endorser.pkiid) {
+			p.groupEndorsers[group] = append(endorsers[:i], endorsers[i+1:]...)
+			break
+		}
+	}
```
manish-sethi, line 103: Isn't this a no-op (will `if` be ever true?) as you already remove endorser(s) when you hand out to the caller.

internal/pkg/gateway/endorsement.go
```diff
@@ -0,0 +1,142 @@
...
+	i := p.nextLayout
+	for i < len(p.layouts) {
+		layout := p.layouts[i]
+		if layout != nil {
+			if _, exists := layout.required[group]; exists {
+				p.layouts[i] = nil
+			}
+		}
+		i++
+	}
```
manish-sethi: A more idiomatic approach would be `for i := p.nextLayout; i < len(p.layouts); i++ {`

internal/pkg/gateway/endorsement.go
```diff
@@ -0,0 +1,142 @@
...
+		if bytes.Equal(e.pkiid, endorser.pkiid) {
+			p.groupEndorsers[group] = append(endorsers[:i], endorsers[i+1:]...)
+			break
+		}
+	}
+
+	for _, layout := range p.layouts {
+		if layout == nil {
+			continue
+		}
```
manish-sethi: Related to the comment in the `endorsers` function - Do you need to start here since beginning? Instead, can start with `p.nextLayout`.

manish-sethi (commented):
Looks good to me largely, left a few comments.

internal/pkg/gateway/endorsement.go
andrew-coleman: This becomes true if the endorser was not handed out by this plan, as is always the case for the firstEndorser, which gets used to create the chancode interest that generates the plan, and is subsequently removed from that plan.  I've added a comment and a couple of unit tests which drive this behaviour.

manish-sethi (approved):

PR #2990, opened 2021-10-19
Adding chaincode-as-a-service external builder to peer's docker image

### LINKED ISSUE #2884: Bundle a chaincode-as-a-service external builder into the hyperledger/fabric Docker image
The [Fabric Operations Guide](https://hyperledger-fabric.readthedocs.io/en/latest/cc_service.html#configuring-a-peer-to-process-external-chaincode) describes a process by which a Fabric network administrator may install an external builder to support the lifecycle of external chaincode service endpoints.   This involves updating the peer's config file with the location of the external builder, and installing the `build`, `detect`, and `release` binaries on the peer file system.

The documentation guide provides a reference sample of external builder scripts (bash), but this path is difficult to realize in orchestrated clusters (e.g. Kubernetes.)  A couple of pain points include: 

- The sample bash scripts rely on `jq` for manipulating the metadata and connection json files, but `jq` is not available on the hyperledger/fabric-peer image. 
- Deploying custom scripts and files into the peer image is possible using a combination of Kubernetes configmaps, init containers, or volume shares, but the process adds an additional, tedious configuration step for the network administrator.

The issues outlined above add just enough structural friction to delay or defer the migration to _chaincode-as-a-service_ (_CCaaS_) deployments.  Let's make this trivial for end-users to update to the new pattern by: 

1.  Including reference _chaincode-as-a-service_ builder binaries into the hyperledger/fabric-peer Docker image.

or 

2.  Include a _CCaaS_ build pipelines into the peer's Golang runtime.

With option (1), the network will enable the default CCaaS builder by modifying core.yaml, referencing a known (documented) location for the system builder.  E.g.: 
```
externalBuilders:
  - name: fabric-ccs-builder
    path: /usr/local
    propagateEnvironment:
      - HOME
      - CORE_PEER_ID
      - CORE_PEER_LOCALMSPID
```

The fabric-peer image, in turn, will need to be supplemented with [fabric-ccs-builder](https://github.com/hyperledgendary/fabric-ccs-builder) (or equivalent) binaries on the docker image.  In the configuration above, the image file system includes: 
```
- usr
    +- local
        +- bin
            +- peer 
            +- build 
            +- detect 
            +- release 
```

In option (2) above, instead of distributing external binaries with the hyperledger/fabric-peer image, the `peer` binary will be supplemented with logic suitable for emulating the `build`, `detect`, `release` lifecycle without exec-ing a new process.  With this pattern, the network administrator may register the default builder by specifying an attribute in the `externalBuilders` configuration stanza, e.g.: 

```
  - name: fabric-ccs-builder
    type: ccsas
    propagateEnvironment:
      - HOME
      - CORE_PEER_ID
      - CORE_PEER_LOCALMSPID
```

With the latter configuration, when the peer detects an external builder with the `ccaas` type,  the routines internally bypass the fork/exec, invoking native Golang routines to manipulate the chaincode descriptors.

Choose one of these approaches (or a third...) and bundle an external builder into the base fabric image.  This will be a valuable time-saver and convenience for running chaincode-as-a-service in Fabric networks.

mbwhite:

#### Type of change

- Bug fix
- New feature
- Test update
- Documentation update

#### Related issues

Resolves #2884


jt-nti:
It would be really really nice to get an out of the box external builder, although that's only part of the story... e.g. `peer chaincode package` would need to know about it. Including the external builder in the docker image seems like a good option but it would also be good to update the fabric-samples nano test network to demonstrate how to configure it outside docker. It might also be nice to support prebuilt/binary chaincode in the build. For example, the fabric integration tests include a builder for that -> https://github.com/hyperledger/fabric/tree/main/integration/externalbuilders

mbwhite:
Candidate update to the fabric-samples to use these builders. https://github.com/hyperledger/fabric-samples/pull/507

mbwhite:
**Reference for Running Chaincode-as-a-service**

_node.js_
All the code is part of the node libraries (use >2.2.1). Chaincode-as-a-service can be started by, for example, adding these two lines to the `package.json` scripts section

```
"start:server-nontls": "set -x && fabric-chaincode-node server --chaincode-address=$CHAINCODE_SERVER_ADDRESS --chaincode-id=$CHAINCODE_ID",
"start:server": "set -x && fabric-chaincode-node server --chaincode-address=$CHAINCODE_SERVER_ADDRESS --chaincode-id=$CHAINCODE_ID --chaincode-tls-key-file=/hyperledger/privatekey.pem --chaincode-tls-client-cacert-file=/hyperledger/rootcert.pem --chaincode-tls-cert-file=/hyperledger/cert.pem"
```

_java_
For Java use the 2.3 or main branch libraries. A separate main class needs to be written and set as the main class. (this is instead of the `org.hyperleder.fabric.contract.ContractRouter`)

```
/*
 * SPDX-License-Identifier: Apache-2.0
 */

package org.example;

import java.io.IOException;
import org.hyperledger.fabric.contract.ContractRouter;
import org.hyperledger.fabric.shim.ChaincodeServer;
import org.hyperledger.fabric.shim.ChaincodeServerProperties;
import org.hyperledger.fabric.shim.NettyChaincodeServer;

public class ContractBootstrap {

    private static final String CHAINCODE_SERVER_ADDRESS = "CHAINCODE_SERVER_ADDRESS";
    private static final String CHAINCODE_ID = "CHAINCODE_ID";
    private static final String CORE_PEER_TLS_ENABLED = "CORE_PEER_TLS_ENABLED";
    private static final String CORE_PEER_TLS_ROOTCERT_FILE = "CORE_PEER_TLS_ROOTCERT_FILE";
    private static final String ENV_TLS_CLIENT_KEY_FILE = "CORE_TLS_CLIENT_KEY_FILE";
    private static final String ENV_TLS_CLIENT_CERT_FILE = "CORE_TLS_CLIENT_CERT_FILE";


    public static void main(String[] args) throws Exception {
        ChaincodeServerProperties chaincodeServerProperties = new ChaincodeServerProperties();

        final String chaincodeServerPort = System.getenv(CHAINCODE_SERVER_ADDRESS);
        if (chaincodeServerPort == null || chaincodeServerPort.isEmpty()) {
            throw new IOException("chaincode server port not defined in system env. for example 'CHAINCODE_SERVER_ADDRESS=0.0.0.0:9999'");
        }

        final int port = Integer.parseInt(chaincodeServerPort.split(":")[1]);
        chaincodeServerProperties.setPortChaincodeServer(port);

        final String coreChaincodeIdName = System.getenv(CHAINCODE_ID);
        if (coreChaincodeIdName == null || coreChaincodeIdName.isEmpty()) {
            throw new IOException("core peer address not defined in system env. for example 'CHAINCODE_ID=externalcc:06d1d324e858751d6eb4211885e9fd9ff74b62cb4ffda2242277fac95d467033'");
        }

        boolean tlsEnabled = Boolean.parseBoolean(System.getenv(CORE_PEER_TLS_ENABLED));
        if (tlsEnabled) {
            // String tlsClientRootCertPath = System.getenv(CORE_PEER_TLS_ROOTCERT_FILE);
            String tlsClientKeyFile = System.getenv(ENV_TLS_CLIENT_KEY_FILE);
            String tlsClientCertFile = System.getenv(ENV_TLS_CLIENT_CERT_FILE);

            // set values on the server properties
            chaincodeServerProperties.setTlsEnabled(true);
            chaincodeServerProperties.setKeyFile(tlsClientKeyFile);
            chaincodeServerProperties.setKeyCertChainFile(tlsClientCertFile);
        }

        ContractRouter contractRouter = new ContractRouter(new String[] {"-i", coreChaincodeIdName});
        ChaincodeServer chaincodeServer = new NettyChaincodeServer(contractRouter, chaincodeServerProperties);

        contractRouter.startRouterWithChaincodeServer(chaincodeServer);
    }

}
```

Note - this should be added to the java libraries properly

_golang_
Like Java, this requires a separate 'main' file to be written.. for example
```
/*
*/

package main

import (
	"encoding/base64"
	"encoding/json"
	"fmt"
	"io/ioutil"
	"log"
	"os"
	"strconv"

	"github.com/hyperledger/fabric-chaincode-go/shim"
	"github.com/hyperledger/fabric-contract-api-go/contractapi"
	"github.com/hyperledger/fabric-samples/asset-transfer-basic/chaincode-go/chaincode"
)

func main() {

	fmt.Printf("Starting chaincode\n")

	assetChaincode, err := contractapi.NewChaincode(&chaincode.SmartContract{})
	if err != nil {
		log.Panicf("Error creating asset-transfer-basic chaincode: %v", err)
	}
	fmt.Printf("Created Contract\n")

	tlsProps, err := getTlsProps()
	if err != nil {
		log.Panicf("Error creating getting TLS properties: %v", err)
	}
	fmt.Printf("Creating Server\n")

	server := &shim.ChaincodeServer{
		CCID:     os.Getenv("CHAINCODE_ID"),
		Address:  os.Getenv("CHAINCODE_SERVER_ADDRESS"),
		CC:       assetChaincode,
		TLSProps: tlsProps,
	}

	out, err := json.Marshal(server)
	if err != nil {
		panic(err)
	}
	fmt.Println(string(out))

	fmt.Printf("Starting server\n")
	err = server.Start()
	if err != nil {
		fmt.Printf("Error starting go-contract chaincode: %s", err)
	}

}

func getTlsProps() (shim.TLSProperties, error) {

	tlsEnabled, err := strconv.ParseBool(os.Getenv("CORE_PEER_TLS_ENABLED"))
	if err != nil {
		tlsEnabled = false
	}

	if !tlsEnabled {
		return shim.TLSProperties{Disabled: true}, nil
	}

	var key []byte
	path, set := os.LookupEnv("CORE_TLS_CLIENT_KEY_FILE")
	if set {
		key, err = ioutil.ReadFile(path)
		if err != nil {
			return shim.TLSProperties{}, fmt.Errorf("failed to read private key file: %s", err)
		}

	} else {
		data, err := ioutil.ReadFile(os.Getenv("CORE_TLS_CLIENT_KEY_PATH"))
		if err != nil {
			return shim.TLSProperties{}, fmt.Errorf("failed to read private key file: %s", err)
		}
		key, err = base64.StdEncoding.DecodeString(string(data))
		if err != nil {
			return shim.TLSProperties{}, fmt.Errorf("failed to decode private key file: %s", err)
		}
	}

	var cert []byte
	path, set = os.LookupEnv("CORE_TLS_CLIENT_CERT_FILE")
	if set {
		cert, err = ioutil.ReadFile(path)
		if err != nil {
			return shim.TLSProperties{}, fmt.Errorf("failed to read public key file: %s", err)
		}
	} else {
		data, err := ioutil.ReadFile(os.Getenv("CORE_TLS_CLIENT_CERT_PATH"))
		if err != nil {
			return shim.TLSProperties{}, fmt.Errorf("failed to read public key file: %s", err)
		}
		cert, err = base64.StdEncoding.DecodeString(string(data))
		if err != nil {
			return shim.TLSProperties{}, fmt.Errorf("failed to decode public key file: %s", err)
		}
	}

	root, err := ioutil.ReadFile(os.Getenv("CORE_PEER_TLS_ROOTCERT_FILE"))
	if err != nil {
		return shim.TLSProperties{}, fmt.Errorf("failed to read root cert file: %s", err)
	}

	return shim.TLSProperties{
		Disabled:      false,
		Key:           key,
		Cert:          cert,
		ClientCACerts: root,
	}, nil
}
```

ext_ccs_builder/go.mod
```diff
@@ -0,0 +1,5 @@
+module hyperledgendary/fabric-ext-builder/v2
```
davidkel: wrong module name ?

davidkel:
I really think that this should support non chaincode-as-a-service as well to support existing users with existing chaincode implementations. I feel the initial benefit of this is the removal of dependencies on the internal Docker builder and launcher, plus there are an awful lot of existing chaincode implementations that would need to be changed over to chaincode as a service

ext_ccs_builder/go.mod
mbwhite: ah yes you're correct

mbwhite:
Sorry I'd misrepresented the scope of this suggestion in the title.  this is _only_ about supporting the issue #2884 - that is chaincode-as-a-service. 

The aim is to provide, in the Fabric Peer as standard, an external builder for processing chaincode-as-a-service. rather than have to configure the peer specifically to add this. 

@jt-nti @davidkel both very correct in your points.  The ways an external builder (not using chaincode-as-a-service) are vast, though not directly called from the peer, code needs to be running in the same container as the peer to build/create the chaincode. This is now responsible for orchestrating the starting of the chaincode. It may no longer be docker, but it would need to be something. 

This specific PR is exploring one of the ways that an external builder could be provided as standard - that is via including it preconfigured in the built docker image. This is a simple, and effective solution.  It does not however mean a 'standalone' peer binary would know, though that could be configured to use the same if you wished.

jkneubuh (dismissed):
I really like this approach.  The compromise of installing a chaincode-as-a-service external builder into the peer image strikes the right balance of simplicity vs flexibility.  Even with some rough edges around the docker build, it seems infinitely better than the alternative approach of altering the peer golang routines for cc lifecycle.

Using the new default builder, it was _very easy_ to get up and running with external chaincode services running on Kubernetes.   Matthew's (fabric-samples) PR 527 shows: 

1.  An update to core.yaml to enable the builder: 
```
    externalBuilders:
        - name: ccaas_builder
          propagateEnvironment:
            - EXTERNAL_CHAINCODE_BUILDER_CONFIG
          path: /opt/hyperledger/ccaas_builder
```

2.  yaml updates to launch the chaincode containers (compiled out of band) 

3.  template substitution within connection.json 


In contrast with the previous approach for registering an external chaincode builder, this is _very, very straightforward._   This PR reduces the config burden for switching to external chaincode-as-a-service significantly.  I approve!

Makefile
```diff
@@ -344,3 +344,13 @@ spaces:
...
 docs:
 	@docker run --rm -v $$(pwd):/docs n42org/tox:3.4.0 sh -c 'cd /docs && tox -e docs'
+
+.PHONY: ccaasbuilder-clean
+ccaasbuilder-clean:
+	rm -rf $(MARCH:%=release/%)/bin/ccaas_builder
+
+.PHONY: ccaasbuilder
+ccaasbuilder: ccaasbuilder-clean
+	cd ccaas_builder && go build -o ../$(MARCH:%=release/%)/bin/ccaas_builder/bin/ ./cmd/build/
```
SamYuan1990: Will the binary file, `build`, `detect`, `release` been added in release tar?

ccaas_builder/go.mod
```diff
@@ -0,0 +1,5 @@
+module hyperledgendary/fabric-ext-builder/v2
+
+go 1.16
```
SamYuan1990, line 3: https://github.com/hyperledger/fabric/blob/main/go.mod#L3
em... according to main branch, seems peer use go 1.14?

ccaas_builder/cmd/build/main.go
```diff
@@ -0,0 +1,163 @@
+package main
```
SamYuan1990, line 7: I wanna some comments here, for this file.
Please correct me for any mistake.
This file looks like analysis the file with schema of `type Connection struct`.

ccaas_builder/cmd/detect/main.go
```diff
@@ -0,0 +1,55 @@
...
+		}
+
+		var metadata ChaincodeMetadata
+		err = json.Unmarshal(mdbytes, &metadata)
+		if err != nil {
+			return err
+		}
+
+		switch strings.ToLower(metadata.Type) {
+		case "external":
```
SamYuan1990: if we only have one condition here... why not if else?

SamYuan1990 (commented):
@mbwhite , I added some comments here. please take a look at.

Makefile
mbwhite: That's a good idea; I don't see why they shouldn't be included within that tar file.

ccaas_builder/go.mod
mbwhite: mmm I thought it had all been pushed up to 1.16.. @denyeart ?

ccaas_builder/cmd/build/main.go
mbwhite: some additional comments are needed yes.

ccaas_builder/cmd/detect/main.go
mbwhite: a good question... we also need a better name not external

jkneubuh: - `CCAAS` ? 
- `CCaaS` ? 
- `ccaas` ? 
- `external-service` ? 
- `external-ccaas` ? 
- `external-chaincode` ?
- `service` ? 


(I like `ccaas`, despite it being _impossible_ to type into a form with an autocorrect spellchecker.)

mbwhite: I'll update to use **ccaas** :-)

ccaas_builder/cmd/build/main.go
mbwhite: @SamYuan1990 have updated the comments

mbwhite:
@hyperledger/fabric-core-maintainers this is ready for review.. thankyou.

internal/peer/node/start.go
```diff
@@ -204,7 +204,7 @@ func serve(args []string) error {
 	grpc.EnableTracing = true
 
 	logger.Infof("Starting %s", version.GetInfo())
-
+	logger.Infof("Starting locally rebuilt: mbw01")
```
andrew-coleman: Did you mean to leave this line in?

mbwhite: No; I thought I'd removed that... thanks @andrew-coleman

ccaas_builder/cmd/build/main.go
```diff
@@ -0,0 +1,175 @@
...
+	}
+
+	var metadata ChaincodeMetadata
+	if err := json.Unmarshal(metadataFileContents, &metadata); err != nil {
+		return err
+	}
+
+	if strings.ToLower(metadata.Type) != "external" {
+		return fmt.Errorf("chaincode type should be external, it is %s", metadata.Type)
+	}
```
jkneubuh: This is a mismatch with the types declared in metadata.json.  Should be: 
```
	if strings.ToLower(metadata.Type) != "ccaas" {
  		return fmt.Errorf("chaincode type should be 'ccaas', it is %s", metadata.Type)
  	}
```

ccaas_builder/cmd/detect/main.go
```diff
@@ -0,0 +1,60 @@
...
+var logger = log.New(os.Stderr, "", 0)
+
+func main() {
+	logger.Println("::Detect")
+
+	if err := run(); err != nil {
+		logger.Printf("::Error: %v\n", err)
+		os.Exit(1)
+	} else {
+		logger.Printf("::Type detected as external")
```
jkneubuh: external -> ccaas

ccaas_builder/cmd/detect/main.go
```diff
@@ -0,0 +1,60 @@
...
+			return err
+		}
+
+		var metadata ChaincodeMetadata
+		err = json.Unmarshal(mdbytes, &metadata)
+		if err != nil {
+			return err
+		}
+
+		if strings.ToLower(metadata.Type) == "ccass" {
```
jkneubuh: ccass -> ccaas

ccaas_builder/cmd/release/main.go
```diff
@@ -0,0 +1,103 @@
...
+}
+
+func main() {
+	logger.Println("::Release")
+
+	if err := run(); err != nil {
+		logger.Printf("::Error: %v\n", err)
+		os.Exit(1)
+	} else {
+		logger.Printf("::Type detected as external")
```
jkneubuh: external -> ccaas

images/peer/Dockerfile
```diff
@@ -20,19 +20,23 @@ RUN apk add --no-cache \
...
 	musl-dev
+RUN GO111MODULE=on go get github.com/mikefarah/yq/v4
 ADD . $GOPATH/src/github.com/hyperledger/fabric
 WORKDIR $GOPATH/src/github.com/hyperledger/fabric
 
 FROM golang as peer
 ARG GO_TAGS
 RUN make peer GO_TAGS=${GO_TAGS}
+RUN make ccaasbuilder
+RUN yq e -P '.chaincode.externalBuilders[0]={"name":"ccaas_builder","propagateEnvironment":["EXTERNAL_CHAINCODE_BUILDER_CONFIG"],"path":"/opt/hyperledger/ccaas_builder"}' /go/src/github.com/hyperledger/fabric/sampleconfig/core.yaml > /go/src/github.com/hyperledger/fabric/sampleconfig/core_updated.yaml
```
jkneubuh: Does this need to be `CHAINCODE_AS_A_SERVICE_BUILDER_CONFIG`  ?

jkneubuh (changes requested):
The new approach works really well but there are a couple of issues in this PR: 

- `ccass` -> `ccaas`       (to match target of metadata.json) 

- detect is detecting `ccass` but build/main.go is still checking for `external` 

- The default core.yaml should propagate the updated env var name `CHAINCODE_AS_A_SERVICE_BUILDER_CONFIG` 


After making these updates in the core images it was smooth sailing.

mbwhite:
thank you @jkneubuh; I've grepped the code as well looking for the words external ccaas ccaass ccas etc  make sure there's nothing hidden.. should be good now

jkneubuh:
👍

images/peer/Dockerfile
```diff
@@ -20,19 +20,23 @@ RUN apk add --no-cache \
...
 	musl-dev
+RUN GO111MODULE=on go get github.com/mikefarah/yq/v4
 ADD . $GOPATH/src/github.com/hyperledger/fabric
 WORKDIR $GOPATH/src/github.com/hyperledger/fabric
 
 FROM golang as peer
 ARG GO_TAGS
 RUN make peer GO_TAGS=${GO_TAGS}
+RUN make ccaasbuilder
+RUN yq e -P '.chaincode.externalBuilders[0]={"name":"ccaas_builder","propagateEnvironment":["CHAINCODE_AS_A_SERVICE_BUILDER_CONFIG"],"path":"/opt/hyperledger/ccaas_builder"}' /go/src/github.com/hyperledger/fabric/sampleconfig/core.yaml > /go/src/github.com/hyperledger/fabric/sampleconfig/core_updated.yaml
```
denyeart: I think it would be better to just update sampleconfig/core.yaml, rather than inject the config here. Injecting it here incurs an extra yq dependency and makes the docker image build awkward. The sampleconfig/core.yaml is already intended to be used in the fabric peer docker image as-is.

ccaas_builder/go.mod
denyeart: The project currently uses Go 1.16.
As far as I can tell, the value in the go directive here doesn't really matter since there have been no incompatible language changes, and therefore the Fabric go.mod still says 1.14. For new components such as this one, I would utilize the current version which is go 1.16.

ccaas_builder/go.mod
```diff
@@ -0,0 +1,5 @@
+module hyperledgendary/fabric-ext-builder/v2
```
denyeart: Should this be github.com/hyperledger/fabric/ccaas_builder?

ccaas_builder/cmd/detect/main.go
```diff
@@ -0,0 +1,60 @@
...
+			return err
+		}
+
+		if strings.ToLower(metadata.Type) == "ccaas" {
+			return nil
+		} else {
+			return fmt.Errorf("chaincode type not supported: %s", metadata.Type)
+		}
+
+	} else {
```
denyeart: It is idiomatic to avoid indenting and else statements. Handle the failure first in a short indent, and then the normal flow can be un-indented.
See guidance at https://github.com/golang/go/wiki/CodeReviewComments#indent-error-flow

ccaas_builder/cmd/build/main.go
```diff
@@ -0,0 +1,175 @@
...
+}
+
+func main() {
+	logger.Println("::Build")
+
+	if err := run(); err != nil {
+		logger.Printf("::Error: %v\n", err)
+		os.Exit(1)
+	} else {
+		logger.Printf("::Type detected as ccaas")
```
denyeart: This looks like a copy/paste from detect program.
And no need for an else here.
See guidance at https://github.com/golang/go/wiki/CodeReviewComments#indent-error-flow

ccaas_builder/cmd/detect/main.go
```diff
@@ -0,0 +1,60 @@
...
+
+var logger = log.New(os.Stderr, "", 0)
+
+func main() {
+	logger.Println("::Detect")
+
+	if err := run(); err != nil {
+		logger.Printf("::Error: %v\n", err)
+		os.Exit(1)
+	} else {
```
denyeart: No need for an else. You can handle the normal flow outside an indented else.
See guidance at https://github.com/golang/go/wiki/CodeReviewComments#indent-error-flow

ccaas_builder/cmd/build/main.go
```diff
@@ -0,0 +1,175 @@
...
+	}
+}
+
+func run() error {
+	sourceDir, metadataDir, outputDir := os.Args[1], os.Args[2], os.Args[3]
```
denyeart, line 64: Check number of args available first.

ccaas_builder/cmd/build/main.go
```diff
@@ -0,0 +1,175 @@
...
+	err = ioutil.WriteFile(connectionDestFile, updatedConnectionBytes, fileInfo.Mode())
+	if err != nil {
+		return err
+	}
+
+	return nil
+
+}
+
+// Helper function to process a template against a string, and return a string
```
denyeart: function comments in Go start with the function name, e.g.
// execTempl is a helper function to...

ccaas_builder/cmd/build/main.go
```diff
@@ -0,0 +1,175 @@
...
+func execTempl(cfg map[string]interface{}, inputStr string) (string, error) {
+
+	t, err := template.New("").Parse(inputStr)
+	if err != nil {
+		fmt.Printf("Failed to parse the Address field template: %s", err)
```
denyeart: Not specific to Address.

ccaas_builder/cmd/build/main.go
```diff
@@ -0,0 +1,175 @@
...
+	var cfg map[string]interface{}
+
+	cfgString := os.Getenv("CHAINCODE_AS_A_SERVICE_BUILDER_CONFIG")
+	if cfgString != "" {
+		if err := json.Unmarshal([]byte(cfgString), &cfg); err != nil {
+			return fmt.Errorf("Failed to unmarshal %s", err)
+		}
+	}
+
+	updatedConnection.Address, err = execTempl(cfg, connectionData.Address)
```
denyeart, line 133: I'm not following what value the template use provides.

ccaas_builder/cmd/build/main.go
```diff
@@ -0,0 +1,175 @@
...
+		updatedConnection.RootCert, err = execTempl(cfg, connectionData.RootCert)
+		if err != nil {
+			return fmt.Errorf("Failed to parse the RootCert field template: %s", err)
+		}
+		updatedConnection.ClientKey, err = execTempl(cfg, connectionData.ClientKey)
+		if err != nil {
+			return fmt.Errorf("Failed to parse the ClientKey field template: %s", err)
+		}
+
+	} else {
```
denyeart: No need for an else here.

ccaas_builder/cmd/release/main.go
```diff
@@ -0,0 +1,103 @@
...
+	ClientCert  string `json:"client_cert"`
+}
+
+func main() {
+	logger.Println("::Release")
+
+	if err := run(); err != nil {
+		logger.Printf("::Error: %v\n", err)
+		os.Exit(1)
+	} else {
```
denyeart: No need for an else here.

ccaas_builder/cmd/release/main.go
```diff
@@ -0,0 +1,103 @@
...
+		logger.Printf("::Type detected as ccaas")
+	}
+}
+
+func run() error {
+	builderOutputDir, releaseDir := os.Args[1], os.Args[2]
```
denyeart, line 54: Check for number of args.

internal/peer/node/start.go
```diff
@@ -204,7 +204,6 @@ func serve(args []string) error {
 	grpc.EnableTracing = true
 
 	logger.Infof("Starting %s", version.GetInfo())
-
```
denyeart: No need to touch this file :-)

denyeart:
Each of the programs should have associated tests.

ccaas_builder/cmd/build/main.go
```diff
@@ -0,0 +1,175 @@
...
+
+	t, err := template.New("").Parse(inputStr)
+	if err != nil {
+		fmt.Printf("Failed to parse the Address field template: %s", err)
+		return "", err
+	}
+
+	buf := &bytes.Buffer{}
+	err = t.Execute(buf, cfg)
+	return buf.String(), err
```
denyeart: Check for the err and return the error first, rather than returning both the success response and the error in the same line.

denyeart (commented):
Each of the programs should have associated tests.
What we really need is an integration test using the new ccaas builder. There is an existing ccaas integration test that could utilize the new builder programs. See PR history for the test here:
https://github.com/hyperledger/fabric/pull/1510
https://github.com/hyperledger/fabric/pull/1561

ccaas_builder/cmd/build/main.go
mbwhite: Within each org the deployed chaincode package needs to be identical. Take the case of two peers in an org both with the same package (with then the same connection.json).  Unless this connection.json can be modified by the builder only a single instance of chaincode process can be used.  

Causes problems for scalability and availability.

ccaas_builder/cmd/build/main.go
mbwhite: Can you expand on the thinking here please?  There are two codepaths, one if the `connectionData.TLS` is true and one if it's false.

images/peer/Dockerfile
mbwhite: Yes that would be cleaner; I wasn't sure if updating the sample config would be ok.

internal/peer/node/start.go
mbwhite: I'd updated this to include a log point to prove when testing that it was actually the peer I'd rebuilt.

ccaas_builder/cmd/build/main.go
denyeart: In the If statement you'll do a return with the error.
If you don't fall into the If statement, it will continue on the normal code path. No else statement is needed for the normal code path. This is a fundamental Go philosophy.

mbwhite: Both branches are normal code paths here... 

```
if (tlsEnabled) {
// stuff for tls
} else {
// stuff for non tls
}
```

The only other way this could be written would be 
```
/// assume non-tls and setup tls stuff
if (tlsEnabed){
 // setup tls, assuming any non-tls stuff is 'undone'
}
```

denyeart: Sorry, my second response was incorrect, I was thinking of the other occurrences.

I stand by the initial review comment however. In Go the zero value of a boolean is false. Therefore you don't need to explicitly set `updatedConnection.TLS` to false. You can simply check for the true case in the If statement and completely avoid the else. This makes the code easier to read.

See https://dave.cheney.net/2013/01/19/what-is-the-zero-value-and-why-is-it-useful

Many of my comments are simply Go conventions that may seem a bit arbitrary. But when a whole community of developers share the same practices the code becomes more readable by that community!

sampleconfig/core.yaml
```diff
@@ -585,12 +585,12 @@ chaincode:
     # builders in the order specified below.
     # If you don't need to fallback to the default Docker builder, also unconfigure vm.endpoint above.
     # To override this property via env variable use CORE_CHAINCODE_EXTERNALBUILDERS: [{name: x, path: dir1}, {name: y, path: dir2}]
-    externalBuilders: []
-        # - path: /path/to/directory
-        #   name: descriptive-builder-name
-        #   propagateEnvironment:
-        #      - ENVVAR_NAME_TO_PROPAGATE_FROM_PEER
-        #      - GOPROXY
+    externalBuilders:
```
denyeart, line 588: Above, it would be worthwhile to comment about the new out-of-the-box ccaas, in addition to the default Docker builder.

images/peer/Dockerfile
```diff
@@ -26,13 +26,15 @@ WORKDIR $GOPATH/src/github.com/hyperledger/fabric
...
 
 FROM peer-base
 ENV FABRIC_CFG_PATH /etc/hyperledger/fabric
 VOLUME /etc/hyperledger/fabric
 VOLUME /var/hyperledger
 COPY --from=peer /go/src/github.com/hyperledger/fabric/build/bin /usr/local/bin
 COPY --from=peer /go/src/github.com/hyperledger/fabric/sampleconfig/msp ${FABRIC_CFG_PATH}/msp
-COPY --from=peer /go/src/github.com/hyperledger/fabric/sampleconfig/core.yaml ${FABRIC_CFG_PATH}
+COPY --from=peer /go/src/github.com/hyperledger/fabric/sampleconfig/core.yaml ${FABRIC_CFG_PATH}/core.yaml
+COPY --from=peer /go/src/github.com/hyperledger/fabric/release/linux-amd64/bin/ccaas_builder/bin/ /opt/hyperledger/ccaas_builder/bin/
```
denyeart, line 38: Did you build the image locally and confirm it has the correct content and can be used?
e.g. `make docker-clean docker`

denyeart (commented):
Looking pretty good now, I'll go through the code in some more detail, but looks like it is structured well. Just a few nits for now...

images/peer/Dockerfile
mbwhite: I have done; and will repeat the exercise before final merge

jt-nti (commented):
Looks good to me although it seems a bit odd that ccaas_builder is a completely separate go module whereas there are existing commands under `/cmd` which are part of the top level module. What are the pros and cons of doing it each way?

mbwhite:
@jt-nti yes a good question; I don't have the experience of knowledge of the build/codebase structure philosophy to be able to definitively respond.  Will take the direction of the maintainers here, as to what is the most approraite.

Makefile
```diff
@@ -344,3 +344,13 @@ spaces:
...
+
+.PHONY: ccaasbuilder-clean
+ccaasbuilder-clean:
+	rm -rf $(MARCH:%=release/%)/bin/ccaas_builder
+
+.PHONY: ccaasbuilder
+ccaasbuilder: ccaasbuilder-clean
+	cd ccaas_builder && go test -v ./cmd/detect && go build -o ../$(MARCH:%=release/%)/bin/ccaas_builder/bin/ ./cmd/detect/
+	cd ccaas_builder && go test -v ./cmd/build && go build -o ../$(MARCH:%=release/%)/bin/ccaas_builder/bin/ ./cmd/build/
+	cd ccaas_builder && go test -v ./cmd/release && go build -o ../$(MARCH:%=release/%)/bin/ccaas_builder/bin/ ./cmd/release/
```
denyeart, line 356: I don't think we need `go test` here.

Is there a reason you'd put the built executables in $(BUILD_DIR)/bin with the others?
e.g. /build/bin/ccaas_builder/bin

ccaas_builder/cmd/detect/main.go
```diff
@@ -0,0 +1,61 @@
...
+	"encoding/json"
+	"fmt"
+	"io/ioutil"
+	"log"
+	"os"
+	"path/filepath"
+	"strings"
+)
+
+var logger = log.New(os.Stderr, "", 0)
```
denyeart, line 21: Everywhere else in Fabric Go code we use github.com/hyperledger/fabric/common/flogging, e.g. 
`var logger = flogging.MustGetLogger("kvledger")`
But since this doesn't pull in any Fabric dependencies, I guess it's ok to use the built in logger.

ccaas_builder/cmd/detect/main.go
```diff
@@ -0,0 +1,61 @@
...
+	logger.Printf("::Type detected as ccaas")
+
+}
+
+type ChaincodeMetadata struct {
+	Type string `json:"type"`
+}
+
+func run() error {
+	if len(os.Args) > 2 {
```
denyeart: Using idiomatic Go, we would check for the error condition and return the error in the if statement, then the normal flow would be unindented. No need to have all the normal flow within an if statement... by the time you get to the else statement you've forgotten what it's for :-)

ccaas_builder/cmd/detect/main.go
```diff
@@ -0,0 +1,61 @@
...
+		chaincodeMetaData := os.Args[2]
+		mdbytes, err := ioutil.ReadFile(filepath.Join(chaincodeMetaData, "metadata.json"))
+		if err != nil {
+			return err
```
denyeart: We typically add context to the error so the user knows what was going on instead of just returning the underlying Go error message. Typically `errors.WithMessage()` is used for this:
https://pkg.go.dev/github.com/pkg/errors#WithMessage

Do you want to check file existence like you do in build program?

ccaas_builder/cmd/detect/main.go
```diff
@@ -0,0 +1,61 @@
...
+	if err := run(); err != nil {
+		logger.Printf("::Error: %v\n", err)
+		os.Exit(1)
+	}
+
+	logger.Printf("::Type detected as ccaas")
+
+}
+
+type ChaincodeMetadata struct {
```
denyeart: Only use leading upper case if the type (or function) is exported (used outside of this package).

ccaas_builder/cmd/detect/main.go
```diff
@@ -0,0 +1,61 @@
...
+	Type string `json:"type"`
```
denyeart, line 36: same here.

ccaas_builder/cmd/detect/main.go
```diff
@@ -0,0 +1,61 @@
...
+		var metadata ChaincodeMetadata
+		err = json.Unmarshal(mdbytes, &metadata)
+		if err != nil {
+			return err
+		}
+
+		if strings.ToLower(metadata.Type) != "ccaas" {
+			return fmt.Errorf("chaincode type not supported: %s", metadata.Type)
+		}
+		// returning nil indicates to the peer a success full detection
```
denyeart:
```suggestion
		// returning nil indicates to the peer a successful detection
```

ccaas_builder/cmd/detect/main_test.go
```diff
@@ -0,0 +1,56 @@
...
+			exitCode: 1,
+			args:     []string{"wibble", "wibble", "wibble"},
+		},
+		"validtype": {
+			exitCode: 0,
+			args:     []string{"na", "testdata/validtype"},
+		},
+		"wrongtype": {
+			exitCode: 1,
+			args:     []string{"na", "testdata/wrongtype"},
```
denyeart, line 41: Can you also try one with missing metadata.json, this will also show you whether the error message is good enough.

ccaas_builder/cmd/build/main.go
```diff
@@ -0,0 +1,189 @@
...
+	if _, err := os.Stat(metainfoSrcDir); !os.IsNotExist(err) {
+		if err := copy.Copy(metainfoSrcDir, metainfoDestDir); err != nil {
+			return fmt.Errorf("failed to copy build META-INF folder: %s", err)
+		}
+	}
+
+	// Process and update the connections file
+	fileInfo, err := os.Stat(connectionSrcFile)
+	if err != nil {
+		return fmt.Errorf("connection.json not found in source folder: %s", err)
```
denyeart: Good, this is what I mentioned should be done in detect, but using errors.WithMessage() to augment the error message.

ccaas_builder/cmd/build/main_test.go
```diff
@@ -0,0 +1,188 @@
...
+		},
+		"twomany": {
+			exitCode: 1,
+			args:     []string{"wibble", "wibble", "wibble", "wibble"},
+		},
+	}
+
+	// Build ledger binary
+	gt := NewWithT(t)
+	detectCmd, err := gexec.Build("github.com/hyperledger/fabric/ccaas_builder/cmd/build")
```
denyeart:
```suggestion
	buildCmd, err := gexec.Build("github.com/hyperledger/fabric/ccaas_builder/cmd/build")
```

ccaas_builder/cmd/build/main_test.go
```diff
@@ -0,0 +1,188 @@
...
+			gt.Expect(err).NotTo(HaveOccurred())
+			gt.Eventually(session, 5*time.Second).Should(gexec.Exit(testCase.exitCode))
+		})
+	}
+}
+
+func TestGoodPath(t *testing.T) {
+	gt := NewWithT(t)
+
+	releaseCmd, err := gexec.Build("github.com/hyperledger/fabric/ccaas_builder/cmd/build")
```
denyeart, line 60:
```suggestion
	buildCmd, err := gexec.Build("github.com/hyperledger/fabric/ccaas_builder/cmd/build")
```

ccaas_builder/cmd/build/main_test.go
```diff
@@ -0,0 +1,188 @@
...
+		"tls_required": false
+	  }`)
+
+	metadataJson := path.Join(testPath, "in-metadata-dir", "metadata.json")
+	fileMetadata, err := os.Create(metadataJson)
+	gt.Expect(err).NotTo(HaveOccurred())
+	fileMetadata.WriteString(`{
+		  "type":"ccaas"
+		}`)
+
```
denyeart, line 93: Also add negative test for no metadata.json and no connection.json.

ccaas_builder/cmd/build/main.go
```diff
@@ -0,0 +1,189 @@
...
+	if err != nil {
+		return err
+	}
+
+	// read the connection.json file into structure to process
+	var connectionData Connection
+	if err := json.Unmarshal(connectionFileContents, &connectionData); err != nil {
+		return err
+	}
+
```
denyeart, line 117: Do we need to check for required properties that may not have gotten set in the JSON? Or does the template do that?

denyeart (commented):
I reviewed Detect first. Many of the comments there apply to Build and Release as well.

Makefile
mbwhite: Is it correct to assume that the `make unit-test` will find all the tests needed?

The path you provide in the `core.yaml` needs to point to a directory that contains a `bin` directory with the apps in it.  Hence the `ccaas_builder/bin` but as to where exactly the source lives and the binaries built to. happy to take guidance.

ccaas_builder/cmd/detect/main.go
mbwhite: yes -as you say these aren't part of say the Peer, Orderer etc. they are external applications; all the examples in the docs use bash; but they could be anything.

They could pull in the Fabric dependencies such as the `flogging` if you wish.

Makefile
mbwhite: Update: I ran the `make unit-tests` and it didn't pick up these tests automatically? Should it be updated to do this?

ccaas_builder/cmd/detect/main.go
mbwhite: Idiomatic changes made :-)

ccaas_builder/cmd/detect/main.go
mbwhite: Adding context is very useful thanks. that's for pointing out the `withMessage`.  Will update across the board with this.

Though (a) if these files aren't there then the peer's probably gone wrong and (b) the peer doesn't really report much back to if one these programs fails..

ccaas_builder/cmd/detect/main.go
mbwhite: Ah yes.. sorry forget there was a variant of Hungarian Notation instead of `public/private .. ` keywords

ccaas_builder/cmd/build/main.go
mbwhite: I've added the `error if missing` key option to the template for exactly this case.

ccaas_builder/cmd/build/main_test.go
mbwhite: copy'n'paste again :)

denyeart:
> Looks good to me although it seems a bit odd that ccaas_builder is a completely separate go module whereas there are existing commands under `/cmd` which are part of the top level module. What are the pros and cons of doing it each way?

Since these are completely standalone programs with no Fabric dependencies I think it is appropriate to keep them separate.

Makefile
denyeart: I would have expected all unit tests under /fabric to get picked up automatically.

ccaas_builder/cmd/detect/main.go
denyeart: I was thinking of a scenario where the user forgot to include metadata.json in the chaincode package.

ccaas_builder/cmd/detect/main.go
```diff
@@ -0,0 +1,75 @@
...
+		os.Exit(1)
+	}
+
+	logger.Printf("::Type detected as ccaas")
+
+}
+
+type chaincodeMetadata struct {
+	Type string `json:"type"`
+}
```
denyeart, line 37: Type can be lowercase (not exported) as well.

ccaas_builder/cmd/detect/main.go
```diff
@@ -0,0 +1,75 @@
...
+	if len(os.Args) < 3 {
+		return fmt.Errorf("too few arguments")
+	}
+
+	chaincodeMetaData := os.Args[2]
+
+	metadataFile := filepath.Join(chaincodeMetaData, "metadata.json")
+	_, err := os.Stat(metadataFile)
+	if err != nil {
+		return fmt.Errorf("%s not found ", metadataFile)
```
denyeart, line 49: Even for these we usually use errors.New().

ccaas_builder/cmd/build/main.go
```diff
@@ -0,0 +1,194 @@
...
+type chaincodeMetadata struct {
+	Type string `json:"type"`
+}
+
+// Connection structure is used to represent the
+// connection.json file that is supplied in the
+// chaincode package.
+type connection struct {
+	Address     string `json:"address"`
```
denyeart, line 34: These can all be lowercase (not exported).

ccaas_builder/cmd/detect/main.go
mbwhite: just been updating these, but now have VSCode warnings...

```should use fmt.Errorf(...) instead of errors.New(fmt.Sprintf(...)) (S1028) go-staticcheck```

File: ccaas_builder/cmd/detect/main.go (hunk 41 above)
Comment (mbwhite) [reply ccaas_builder/cmd/detect/main.go:37]: Used to unmarshall JSON data into so the elements need to be exported

File: ccaas_builder/cmd/build/main.go (hunk 43 above)
Comment (mbwhite) [reply ccaas_builder/cmd/build/main.go:34]: Used to unmarshall JSON data into so the elements need to be exported

**Comment by jt-nti** [comment]
> > Looks good to me although it seems a bit odd that ccaas_builder is a completely separate go module whereas there are existing commands under `/cmd` which are part of the top level module. What are the pros and cons of doing it each way?
> 
> Since these are completely standalone programs with no Fabric dependencies I think it is appropriate to keep them separate.

So a follow on question is whether it would be better for the new external builder module to live in a separate repo. This PR would essentially convert the fabric repository into a monorepo. What are the implications of that in Go?

**Comment by denyeart** [comment]
> > > Looks good to me although it seems a bit odd that ccaas_builder is a completely separate go module whereas there are existing commands under `/cmd` which are part of the top level module. What are the pros and cons of doing it each way?
> > 
> > 
> > Since these are completely standalone programs with no Fabric dependencies I think it is appropriate to keep them separate.
> 
> So a follow on question is whether it would be better for the new external builder module to live in a separate repo. This PR would essentially convert the fabric repository into a monorepo. What are the implications of that in Go?

Since users wouldn't import the new programs, I don't see an impact to users.
As maintainers, we'd have to manage the new programs as part of the Fabric release in terms of fixes and release tagging.
It's really no different than delivering the other stand alone command line utilities as part of the Fabric release - it hasn't been an issue.
That being said, in an ideal world the new build programs would go into their own repository, and even the Fabric image build and publish could go into its own repository to keep the core Fabric repo slim and pure. The question is whether that purity is worth the cost of the additional CI machinery that would have to be maintained to tie the various repos together. I've been leaning towards keeping the precedent that everything the image depends on goes in a single repo to keep it simple, but it could be argued either way.

**Comment by mbwhite** [comment]
> That being said, in an ideal world the new build programs would go into their own repository, and even the Fabric image build and publish could go into its own repository to keep the core Fabric repo slim and pure. The question is whether that purity is worth the cost of the additional CI machinery that would have to be maintained to tie the various repos together. I've been leaning towards keeping the precedent that everything the image depends on goes in a single repo to keep it simple, but it could be argued either way.

For the record, I completely agree with the desire to keep things slim/pure; Fabric is a now a product to be used in production environments, so it's consumability is of paramount importance. This PR is intended to make this easier, and make migration to a K8S environment significantly easier.  (agree this is not being disputed).

However I don't think this PR is the place to refactor the repository; the question is it worth not having this consumeability improvement until such a refactor takes place?

**Review by jt-nti — APPROVED** [review]
Since the current precedent is to keep everything the image depends on in a single repo, it makes sense to get this PR in and discuss whether we want to revisit that strategy and refactor the repository later.

I would prefer a minor change to follow the precedent of the existing commands being part of the same Fabric module, rather than introducing a new separate module just for the builder, but also happy for the PR to go in as is if that's the consensus.

File: ccaas_builder/cmd/build/main.go [hunk 44]
```diff
@@ -0,0 +1,194 @@
...
+	return nil
+
+}
+
+// execTempl is a helper function to process a template against a string, and return a string
+func execTempl(cfg map[string]interface{}, inputStr string) (string, error) {
+
+	t, err := template.New("").Option("missingkey=error").Parse(inputStr)
+	if err != nil {
+		fmt.Printf("Failed to parse the template: %s", err)
```
andrew-coleman, line 183: Shouldn't errors be written to stderr?

ccaas_builder/cmd/release/main.go
```diff
@@ -0,0 +1,107 @@
...
+	RootCert    string `json:"root_cert"`
+	ClientKey   string `json:"client_key"`
+	ClientCert  string `json:"client_cert"`
+}
+
+func main() {
+	logger.Println("::Release")
+
+	if err := run(); err != nil {
+		logger.Printf("::Error: %v\n", err)
```
andrew-coleman, line 42: stderr?

ccaas_builder/cmd/build/main.go
mbwhite: good point.. yes they should

ccaas_builder/cmd/release/main.go
mbwhite: yep

ccaas_builder/cmd/build/main.go
mbwhite: @andrew-coleman just went and checked - the `logger` that is created at the top of the file is set to use stderr - so all the messages will go to stdout.

Recall now that the peer only captures stderr for the external builders.

andrew-coleman (approved):

needhamj:
Hi, I'm interested in this. So just to be clear, where does this leave external Java chaincode? Does this PR cover that? Is there any chance of a working image we can use on our project as it's pretty critical for us.

denyeart (approved):
Thanks Matthew. This is a great leap in consumability for chaincode-as-a-service. It does muddy the waters of the Fabric repo a bit, but we can always move it out later if it becomes a problem or if we decide on another strategy for building and publishing the Fabric peer docker image.

denyeart:

denali49:
Well done!!! 

Chris Gabriel, Founder
Hyperchain Labs
***@***.***
918-397-4078

hyperchainlabs.com
Twitter @HyperchainLabs

uqix:
is it possible to use ccaas_builder in fabric v2.2? @mbwhite

mbwhite:
@uqix yes; that should be possible. The  binaries can be used as is, but you will need to edit the config.yaml used by the peer inline with the Fabric 2.4 version. So they pick up the required ccaas binaries.

uqix:
For v2.2, we needed to build custom peer image:

Dockerfile:

```Dockerfile
ARG fabric_version

FROM hyperledger/fabric-peer:2.4.3

FROM hyperledger/fabric-peer:$fabric_version
COPY --from=0 /opt/ /opt/
RUN sed -i 's/externalBuilders: \[\]/externalBuilders:\n        - name: ccaas_builder\n          path: \/opt\/hyperledger\/ccaas_builder/' /etc/hyperledger/fabric/core.yaml
```
Build:

```bash
dockerfile_version=1.0.1
fabric_version=2.2.2
image_name=uqix/hyperledger-fabric-peer:$dockerfile_version-$fabric_version

docker build \
    --build-arg fabric_version=$fabric_version \
    -t $image_name \
    . \
  && docker push $image_name
```

PR #2991, opened 2021-10-21
fix Windows SyncDir issue

### LINKED ISSUE #2984: Unable to use peer/orderer binaries from windows package
The first issue is that the binaries don't have the '.exe' extension so would be good to package them with the extension.

The other problem is the following code from internal/fileutil.go

```
// SyncDir fsyncs the given dir
func SyncDir(dirPath string) error {
	dir, err := os.Open(dirPath)
	if err != nil {
		return errors.Wrapf(err, "error while opening dir:%s", dirPath)
	}
	if err := dir.Sync(); err != nil {
		dir.Close()
		return errors.Wrapf(err, "error while synching dir:%s", dirPath)
	}
	if err := dir.Close(); err != nil {
		return errors.Wrapf(err, "error while closing dir:%s", dirPath)
	}
	return err
}
```
calling Sync on a dir fails with the error "The handle is invalid." on windows as it isn't valid to perform that action

As a workaround I used the following implementation

```
// SyncDir fsyncs the given dir
func SyncDir(dirPath string) error {
	dir, err := os.Open(dirPath)
	if err != nil {
		return errors.Wrapf(err, "error while opening dir:%s", dirPath)
	}
	if err := dir.Sync(); err != nil {
		if !strings.Contains(err.Error(), "The handle is invalid.") {
			dir.Close()
			return errors.Wrapf(err, "error while synching dir:%s", dirPath)
		}
	}
	if err := dir.Close(); err != nil {
		return errors.Wrapf(err, "error while closing dir:%s", dirPath)
	}
	return err
}
```
But this will only work on english versions of windows probably so a better soln should be done
- detect if the platform is windows ?
- remove this function into 2 separate files and use 
```
//go:build linux
// +build linux
```
to embed the correct file for the correct platform

davidkel:
Signed-off-by: D <d_kelsey@uk.ibm.com>

closes #2984 

#### Type of change

- Bug fix

#### Description

On windows it's not valid to invoke Sync on a directory, it results in an error "Handle is invalid"

#### Additional details

As windows is there only as a development platform, removing the attempt to Sync on a dir restores the ability for Peer and Orderer to run natively on Windows Desktop platforms.

Included as well is ensuring the downloadable windows package contains .exe files so they are usable straight away (no extra rename once unpacked is required)

### DIFF
--- a/internal/fileutil/syncdir_windows.go
+++ b/internal/fileutil/syncdir_windows.go
@@ -1,0 +1,12 @@
+/*
+Copyright IBM Corp. All Rights Reserved.
+
+SPDX-License-Identifier: Apache-2.0
+*/
+
+package fileutil
+
+// SyncDir this is a noop on windows
+func SyncDir(dirPath string) error {
+	return nil
+}
--- a/ci/scripts/create_binary_package.sh
+++ b/ci/scripts/create_binary_package.sh
@@ -8,4 +8,7 @@ make "release/${TARGET}"
 mkdir "release/${TARGET}/config"
 mv sampleconfig/*yaml "release/${TARGET}/config"
 cd "release/${TARGET}"
+if [ "$TARGET" == "windows-amd64" ]
+    for FILE in bin/*; do mv $FILE $FILE.exe; done
+fi
 tar -czvf "hyperledger-fabric-${TARGET}-${RELEASE}.tar.gz" bin config
--- a/internal/fileutil/fileutil.go
+++ b/internal/fileutil/fileutil.go
@@ -52,22 +52,6 @@ func CreateAndSyncFile(filePath string, content []byte, perm os.FileMode) error 
 	return nil
 }
 
-// SyncDir fsyncs the given dir
-func SyncDir(dirPath string) error {
-	dir, err := os.Open(dirPath)
-	if err != nil {
-		return errors.Wrapf(err, "error while opening dir:%s", dirPath)
-	}
-	if err := dir.Sync(); err != nil {
-		dir.Close()
-		return errors.Wrapf(err, "error while synching dir:%s", dirPath)
-	}
-	if err := dir.Close(); err != nil {
-		return errors.Wrapf(err, "error while closing dir:%s", dirPath)
-	}
-	return err
-}
-
 // SyncParentDir fsyncs the parent dir of the given path
 func SyncParentDir(path string) error {
 	return SyncDir(filepath.Dir(path))
--- a/internal/fileutil/syncdir.go
+++ b/internal/fileutil/syncdir.go
@@ -1,0 +1,31 @@
+// +build !windows
+
+/*
+Copyright IBM Corp. All Rights Reserved.
+
+SPDX-License-Identifier: Apache-2.0
+*/
+
+package fileutil
+
+import (
+	"os"
+
+	"github.com/pkg/errors"
+)
+
+// SyncDir fsyncs the given dir
+func SyncDir(dirPath string) error {
+	dir, err := os.Open(dirPath)
+	if err != nil {
+		return errors.Wrapf(err, "error while opening dir:%s", dirPath)
+	}
+	if err := dir.Sync(); err != nil {
+		dir.Close()
+		return errors.Wrapf(err, "error while synching dir:%s", dirPath)
+	}
+	if err := dir.Close(); err != nil {
+		return errors.Wrapf(err, "error while closing dir:%s", dirPath)
+	}
+	return err
+}


ale-linux (changes requested):
I'd rather leave things as they are and add some kind of graceful error handling rather than resorting to build tags - since I'm guessing it's not so much a windows/linux issue, but rather a filesystem-specific one. Can we do some kind of typecast of the error and then either err out or log a warning and continue? Also, what's the purpose of that dir sync? Can't we traverse the dir and sync the files individually (in case this makes sense).

davidkel:
@ale-linux I believe fsync on a dir may be required if you rename an existing directory or create one for example. I don't believe it means it will fsync all files in that directory. The point on windows is that you just cannot call Sync on a directory handle it will always return an error saying File handle invalid. Therefore for windows you will never call Sync where on Posix systems you probably always want to call Sync. It would be bad to catch the error message to ignore it for windows, I could check the platform in the code to say if this is windows don't call sync but to me the idomatic approach for Go would be to use build tags and/or filenames which is what I have done here.

For example I think doing this in the code
```
if runtime.GOOS != "windows" {
    dir.Sync()
    ...
}
```
is not nice

internal/fileutil/syncdir_windows.go
```diff
@@ -0,0 +1,33 @@
...
+	// return an error saying the handle is invalid. It is not valid to
+	// call Sync on a windows directory
+
+	// This call still performs a valid check to ensure the directory exists
+	// and can be opened (ie no permission problems)
+
+	if err := dir.Close(); err != nil {
+		return errors.Wrapf(err, "error while closing dir:%s", dirPath)
+	}
+	return err
```
manish-sethi: You can simply return `nil` here. SyncDir is (and is expected to be) called after having created or deleted files inside the dir.

internal/fileutil/syncdir_windows.go
```diff
@@ -0,0 +1,33 @@
+/*
```
manish-sethi, line 1: Don't you need `// +build windows` here?

manish-sethi (commented):
I agree with @davidkel on this. This is idiomatic way of handling platform specific code. In fact, most of the golang based DB projects would have something like this (e.g., https://github.com/syndtr/goleveldb/tree/master/leveldb/storage). However, I am not sure that in general we want to take the burden of claiming that we support windows.

internal/fileutil/syncdir_windows.go
davidkel: @manish-sethi interesting because I couldn't find anything that implied that calling sync on a directory affected the files contained within that directory. It appears to be specific to directories only eg, renaming a directory. The first time it hit the problem was when a directory is created, so it makes sense on linux platforms to sync the directory.

internal/fileutil/syncdir_windows.go
davidkel: Nope, as the _windows in the file name does that for you.

davidkel:
@manish-sethi so the question of support is an interesting one. Only linux is supported for running a fabric network, yet people use a mac to do development and with the recent introduction of test-network-nano-bash, mac users can run fabric natively to help with their development and testing with fabric. Certainly there would not be any support for a fabric network running natively on a mac. Fabric also package binaries which include peer and orderer binaries and that is done for linux, mac and windows. So my intention would be that we offer the same capability afforded to mac users where they can run fabric natively on windows for the purposes of test and development. In fact I think it is testament to both fabric and go lang that it works with only this minor change being required

internal/fileutil/syncdir_windows.go
davidkel: I could just make this a complete no-op for windows, but wondered whether there was some value with doing the open call to check that the directory does indeed exist and that there were no permission problems, which is also what would happen on linux before the sync call was made.

manish-sethi: > @manish-sethi interesting because I couldn't find anything that implied that calling sync on a directory affected the files contained within that directory.

On Linux, the directory is also a file (a special file though) that maintains the list of entries for the files inside the directory. So, now if you create a file inside the dir, synch on dir is required so as to make sure that new dir listing is also persisted. Otherwise, an untimely (or rather a well timed) crash could leave the filesystem in a state where the file contents are persisted (assuming, you called fsync on file already), but the dir does not show that file. In other words, file exists in FS but is not reachable.

>It appears to be specific to directories only eg, renaming a directory.

Again, if you created a dir, you would invoke sync on it's parent dir as well (the same reasoning as above).

manish-sethi: > I could just make this a complete no-op for windows, but wondered whether there was some value with doing the open call to check that the directory does indeed exist and that there were no permission problems, which is also what would happen on linux before the sync call was made.

In the linux specific code, only interesting piece is the `fsync` call. Rest is just for the convenience of the caller, so the caller can pass a string path. Because, anyways, the caller would have added/deleted some files (or dirs) in the dir and any permission problems would have surfaced there.

mastersingh24:
I don't see any issues with this.  `fsync` is not supported on Windows for directories and allowing people who want to try this on Windows for development seems ok.  I'm sure future issues may arise on Windows as well, but this does not hurt anything.

ale-linux:
Thanks for the explanation! I'm still confused though: if the operation is necessary for the correct operations of the peer/orderer, then just silently skipping it on windows doesn't sound to me like the right thing to do, because it may leave the process in an unspecified, unstable state.

Otoh, if the operation can be skipped (as we're prepared to do on windows), then let's skip it.

Should we perhaps consider a fs-wide sync on windows?

mbwhite (approved):

mastersingh24 (approved):

manish-sethi:
> Otoh, if the operation can be skipped (as we're prepared to do on windows), then let's skip it.

@ale-linux - I am not fully sure whether windows performs filesystem meta-operation (create/delete files/dirs) as sync only internally, and hence they don't expose fsync on dirs. I could not find any authentic documentation on this (perhaps, a side-effect of non-opensource).

However, we are not the first one making this choice. Many KV stores have opted for this choice. A few references - [goleveldb](https://github.com/syndtr/goleveldb/blob/2ae1ddf74ef7020251ff1ff0fe8daac21a157761/leveldb/storage/file_storage_windows.go#L78), [badgerdb](https://github.com/dgraph-io/badger/pull/892), [Pebble/Rocksdb](https://github.com/cockroachdb/pebble/blob/bee0c60e96bc1f12308484c9789c10cb4b19f77a/vfs/dir_windows.go#L20). We are anyways dependent on `goleveldb` and already exposed to this behavior. So, I won't prefer to go overboard on this one for windows.

ale-linux (approved):

ale-linux:
Fair enough, thanks @manish-sethi for the investigation.

PR #2707, opened 2021-06-23
WIP: prepare for etcd version bump.

guoger:
Signed-off-by: Jay Guo <guojiannan1101@gmail.com>

PR #2997, opened 2021-10-26
Protobuf and etcd upgrade

atoulme:
#### Type of change

#### Description

Update protobuf and etcd to their latest versions.

#### Additional details

This is a reprisal of the work on updating protobuf by #2185

#### Related issues

FAB-18363


atoulme:
A few regressions from etcd upgrade to look into, very interesting issues to look into. Don't mind me, I'll keep digging.

common/cauthdsl/policy_test.go
```diff
@@ -103,7 +103,8 @@ func TestNewPolicyErrorCase(t *testing.T) {
 	pol1, msg1, err1 := provider.NewPolicy([]byte{0})
 	require.Nil(t, pol1)
 	require.Nil(t, msg1)
-	require.EqualError(t, err1, "Error unmarshalling to SignaturePolicy: proto: common.SignaturePolicyEnvelope: illegal tag 0 (wire type 0)")
+	require.Error(t, err1)
```
Param-S: can we use require.ErrorContains()

orderer/consensus/etcdraft/chain_test.go
```diff
@@ -3243,7 +3243,7 @@ var _ = Describe("Chain", func() {
 						})
 				})
 
-				It("aborts waiting for block to be committed upon leadership lost", func() {
+				FIt("aborts waiting for block to be committed upon leadership lost", func() {
```
Param-S: Because of this Focused testcase, other testcases are skipped. is there any reason for disabling other testcases.

atoulme: That’s not from me, must be from earlier commits I used. That’s an awesome find, maybe we just need to revert that line.

Param-S:
@atoulme can you update your branch & update the PR.

atoulme:
Done - tests running.

Param-S:
there are failures related to protobuf changes:
https://dev.azure.com/Hyperledger/Fabric/_build/results?buildId=46926&view=logs&j=6b58850f-3858-5a05-33e2-5e41cbf03c4e&t=bddec1cf-ba37-5883-9c3e-fd1e8608f9a1&l=3471
, please check here core/chaincode/lifecycle/serializer_test.go

https://dev.azure.com/Hyperledger/Fabric/_build/results?buildId=46926&view=logs&j=6b58850f-3858-5a05-33e2-5e41cbf03c4e&t=bddec1cf-ba37-5883-9c3e-fd1e8608f9a1&l=3726
core/endorser/endorser_test.go:997

This testcase fails with panic. I am seeing the same issue on my local environment. I am currently debugging this one. 
https://dev.azure.com/Hyperledger/Fabric/_build/results?buildId=46926&view=logs&j=6b58850f-3858-5a05-33e2-5e41cbf03c4e&t=bddec1cf-ba37-5883-9c3e-fd1e8608f9a1&l=5373

Param-S:
@atoulme , following change should address the panic(mentioned in the last comment). Please check by updating this PR. 

```
git diff chain_test.go
diff --git a/orderer/consensus/etcdraft/chain_test.go b/orderer/consensus/etcdraft/chain_test.go
index c716fa225..826396287 100644
--- a/orderer/consensus/etcdraft/chain_test.go
+++ b/orderer/consensus/etcdraft/chain_test.go
@@ -212,7 +212,10 @@ var _ = Describe("Chain", func() {
                }

                JustBeforeEach(func() {
-                       chain, err = etcdraft.NewChain(support, opts, configurator, nil, cryptoProvider, noOpBlockPuller, fakeHaltCallbacker.HaltCallback, observeC)
+                       rpc := &mocks.FakeRPC{}
+                       chain, err = etcdraft.NewChain(support, opts, configurator, rpc, cryptoProvider, noOpBlockPuller, fakeHaltCallbacker.HaltCallback, observeC)
```

orderer/consensus/etcdraft/node.go
```diff
@@ -57,8 +58,8 @@ func (n *node) start(fresh, join bool) {
 	var campaign bool
 	if fresh {
 		if join {
-			raftPeers = nil
 			n.logger.Info("Starting raft node to join an existing channel")
+			n.Node = raft.RestartNode(n.config)
```
Param-S, line 62: As per my understanding, fresh + join condition is equal to `adding a new node to existing channel`, it should be StartNode, why it is updated to RestartNode?

atoulme:
Thanks Param, I have applied your changes.

orderer/consensus/etcdraft/node.go
atoulme: The issue is how to add the node to its existing set of peers and config. The docs on the method explains this:
https://github.com/atoulme/fabric/blob/939220f920153ef4f8ac81833f07ad9c85760ebb/vendor/go.etcd.io/etcd/raft/v3/node.go#L238

This commit shows how state is recreated: https://github.com/hyperledger/fabric/pull/2997/commits/6ee57eb420eaaf46704c499ee01f42af648717f0

common/cauthdsl/policy_test.go
atoulme: Yes, of course, changed.

atoulme:
Looks like we have 2 errors left, but both pass locally on my laptop. Any ideas?

Param-S:
I rerun the unit-tests job(with the assumption that the problem related to timing), the failed testcases are passed now in the 2nd run.  It needs bit investigation on what is the issue in the first run. 

There is separate failure, it seems to me related to GRPC msg update. We need to update the testcase wrt latest msg. 

https://dev.azure.com/Hyperledger/Fabric/_build/results?buildId=47086&view=logs&j=e306c17a-d139-54bf-a475-f5a11259cee7&t=1e3023a5-584f-52f3-49bc-66bd27d27b6d&l=130
    zap_test.go:310: 
        	Error Trace:	zap_test.go:310
        	Error:      	Not equal: 
        	            	expected: "grpc DEBUG TestGRPCLogger message\n"
        	            	actual  : "grpc DEBUG callWrapper message\n"

atoulme:
Yes, I was wondering about this. I fixed the problem now by increasing the caller skip level when zap looks for the caller.

gossip/service/gossip_service_test.go
```diff
@@ -298,16 +298,19 @@ func TestWithStaticDeliverClientNotLeader(t *testing.T) {
 	}
 	n := 2
 	gossips := startPeers(serviceConfig, n, 0, 1)
+	fmt.Printf("We have %d gossips\n", len(gossips))
```
yacovm: please remove redundant prints in this test.

gossip/util/logging.go
```diff
@@ -34,7 +34,7 @@ var (
 )
 
 // defaultTestSpec is the default logging level for gossip tests
-var defaultTestSpec = "WARNING"
+var defaultTestSpec = "DEBUG"
```
yacovm: please change it back

gossip/service/gossip_service_test.go
```diff
@@ -50,7 +50,7 @@ import (
 	"google.golang.org/grpc"
 )
 
-const TIMEOUT = 90 * time.Second
+const TIMEOUT = 45 * time.Second
```
yacovm: please revert timeout to what it was

protoutil/txutils_test.go
```diff
@@ -230,7 +230,7 @@ func TestCreateSignedTx(t *testing.T) {
 		},
 	}}
 	_, err = protoutil.CreateSignedTx(prop, signID, responses...)
-	require.Error(t, err, "Expected error with no endorsements")
```
yacovm: I don't understand how this error change is related to protobuf or etcd changes

yacovm:
What is also missing is an explanation of the nature of the changes.

Can you also explain the following?

1. How [Jason's comment](https://github.com/hyperledger/fabric/pull/2452#issuecomment-791130912) and the changes to the Raft chain (in the commits) are tied. Can you explain how the new Raft version behaves and what changes were done to accommodate it? 
2. Whether running mixed Raft clusters (suppose, a cluster of 4) is possible after this change or not.

orderer/consensus/etcdraft/node.go
```diff
@@ -201,6 +202,14 @@ func (n *node) send(msgs []raftpb.Message) {
 
 		status := raft.SnapshotFinish
 
+		// Replace node list in snapshot with CURRENT node list in cluster.
```
yacovm, line 205: can you explain why you're doing this? What if the conf state is the old confState (we committed the fabric config block but didn't commit the conf change entry yet)? 

And we only take snapshots when we commit blocks, right? What if there is no new snapshot created and the latest snapshot will always be a stale one?

gossip/service/gossip_service_test.go
atoulme: fixed

gossip/service/gossip_service_test.go
atoulme: fixed

protoutil/txutils_test.go
atoulme: fixed

orderer/consensus/etcdraft/node.go
atoulme: this is because of changes in the way raft nodes are restarted with this new version. There is a change in the API of etcd raft that requires that we restart without setting peers and set them afterwards.

I fished that change from another in-progress PR that never made it to main, with better explanations: https://github.com/hyperledger/fabric/pull/2452#issuecomment-792312984

atoulme:
@yacovm at the risk of disappointing you, this is just a straight up upgrade of etcd and protobuf to newer versions. There's some API changes, especially in the way a node starts and is configured with existing peers. There's some differences in the errors thrown by protobuf. There are no functional changes besides this.

I have no clue at all as to whether those changes make this code incompatible with previous revisions, ie I have not tried to form a cluster with the code before and after, and I am certainly not the best equipped for that.

yacovm:
> @yacovm at the risk of disappointing you, this is just a straight up upgrade of etcd and protobuf to newer versions. There's some API changes, especially in the way a node starts and is configured with existing peers. There's some differences in the errors thrown by protobuf. There are no functional changes besides this.
> 
> I have no clue at all as to whether those changes make this code incompatible with previous revisions, ie I have not tried to form a cluster with the code before and after, and I am certainly not the best equipped for that.

Sure, I am not blaming or implying the work you did is not valuable.
However I am concerned about:

> There are no functional changes besides this.

I understood that etcd made a functional change in how they handle snapshots and reconfiguration (specifically, removal of nodes). As per @guoger 's [comment](https://github.com/hyperledger/fabric/pull/2452#issuecomment-792312984): 

>  If the snapshot sent was taken before adding new node, then it certainly does not contain new node, and would fail the check aforementioned. etcd server it self handles this by injecting the actual ConfState to snapshot message (produced by raft) before sending them out.

Now, clearly this means that the PR **already contains** a functional change (in the dependencies). 

It is up to Fabric to make sure that the functional change in dependencies doesn't translate to a functional change in operations.

I'm not sure that the trick that Jay pointed out that etcd is doing (which you also attempt to perform) is done correctly, because:

1.  It only takes place if a config change was applied:

```
			state := n.confState.Load()
			if state != nil {
				msg.Snapshot.Metadata.ConfState.Voters = state.(*raftpb.ConfState).Voters
			}
```

What if this node was restarted and then it means the `confState` is nil? In that case won't the voters contain the voters from before the config? 

2. What if since we do the reconfig in two stages, the second entry from Raft is yet to be applied and then we still send the old voters? 


> I have no clue at all as to whether those changes make this code incompatible with previous revisions, ie I have not tried to form a cluster with the code before and after, and I am certainly not the best equipped for that.

I understand, but I think we need to test it before merging this PR so we will know how to advise users.

Param-S:
@yacovm I am also looking at this etcdraft upgrade changes, i can run the mixed configuration cluster validation & confirm the behaviour.

bestbeforetoday:
A naive question, related to this PR although probably not something to be addressed here... should we be looking to migrate to the v2 Go protobuf API?
- https://go.dev/blog/protobuf-apiv2
- https://pkg.go.dev/google.golang.org/protobuf

atoulme:
@bestbeforetoday from the first link:
> Because APIv2 is not backwards compatible with APIv1, we need to use different module paths for each.

So if you read what @yacovm is commenting, that's probably not in the cards.

atoulme:
@yacovm I didn't interpret your comment as a criticism, but as a well founded risk assessment and a set of criteria to look at before this PR is merged. Your feedback is appreciated and on point.

I'd be lying if I didn't say I wasn't impatient to get this in, I can't wait to demonstrate Fabric with complete tracing :)

Besides the points you raised, are there any other points worth of consideration? If Param is able to work through a mixed scenario, will it be enough to get this PR in? What else is needed?

yacovm:
I have nothing else to comment on, apart from what I commented above.

Param-S:
Progress Update: 

So far, I tested the following mixed scenarios:

Orderer 1: Orderer built with etcd v0.5.0-alpha.5.0.20181228115726-23731bf9ba55
Orderer 2: Orderer built with etcd v0.5.0-alpha.5.0.20181228115726-23731bf9ba55
Orderer 3: Orderer built with etcd v3.5.1

| Scenario  | Observation |
|---|---|
| Bring up with test-network-nano-bash. All the orderers started up in the same order  | All orderers able to form quorum and consent on transactions  |
| Bring down the network(all orderers). Restart all of them in reverse order (orderer3, orderer2, orderer1) | All orderers able to form quorum and consent on transactions |
| Submit a config transaction to remove orderer3 from the system channel | Orderer3 evicted & consenters set updated |
| Submit a config transaction to add orderer3 to system channel. Start up the orderer3. | Orderer3 is able to join the quorum with Raft ID 4 |
| Updated the snapshot interval size to 1 KB. Reinitialised the network. | Tried out all above 3 cases, It all worked |

Also Looked at few etcdraft issues wrt StartNode/Restart Node changes: 

* https://github.com/etcd-io/etcd/pull/10892 . As per my understanding, in this PR there are 2 major changes:
1. RawNode specific - it affects only the applications which uses rawnode for raft instantiation. This is not applicable as fabric uses (*node) for raft instantiation. 
2. Bootstrapping specific restructuring - it affects the StartNode invocation with peers = nil. Suggestion is to use RestartNode. We have updated the same in this PR.  (we can refer etcd Raftexample PR https://github.com/etcd-io/etcd/pull/12696/files which updated the example new nodes joining the existing cluster) 

Created an issue in etcd for confirmation on the RestartNode: https://github.com/etcd-io/etcd/issues/13666

I will continue to look at the snapshot specific changes & try to get more understanding.

Param-S:
Further checked on this new condition `the node which receives the snapshot should be part of the quorum/confstate`, It is only additional check introduced in etcdraft PR https://github.com/etcd-io/etcd/commit/b171e1c78b0411c8a2648de0eee2102735bec6b4, https://github.com/etcd-io/etcd/blob/792b7f57d3505b6c87fe46a6b9b4875de0c2c042/raft/raft.go#L1547, there is no functionality difference.  

As Jay pointed out in his comment, etcd made a similar change(https://github.com/etcd-io/etcd/blob/792b7f57d3505b6c87fe46a6b9b4875de0c2c042/server/etcdserver/server.go#L1110) to circumvent the additional check.

Param-S:
> What if this node was restarted and then it means the confState is nil? In that case won't the voters contain the voters from before the config?

As per current understanding, the latest conf-state gets initialized as part of the orderer and raft memory storage initialization. The variable conf-state gets set when the first node becomes the leader, the confstate gets initialized as part of the InFlightConfChange check.

https://github.com/hyperledger/fabric/blob/7b7b6e9248b77be956025cf56eec3e7799f17c72/orderer/consensus/etcdraft/chain.go#L642
https://github.com/hyperledger/fabric/blob/7b7b6e9248b77be956025cf56eec3e7799f17c72/orderer/consensus/etcdraft/chain.go#L223

I verified the behavior with the following steps:

1. Start Orderer1, 2, 3 with SnapshotIntervalSize = 1KB (It expects to take snap on every transaction) 
2. Issue config transaction to add Orderer 4
3. Stop all the 3 orderers
4. Start the Orderer 1, 2 & 4

Added a breakpoints in Send method where the Snapshot updated with latest confstate.  I can see that the leader updates the snapshot with the latest conf state which includes Orderer 4 in the consenters.

Please guide me if i need to validate with any other scenario.

atoulme:
Rebased off main and pushed. Let's see if tests pass - if they do, let's get this one in?

yacovm:
> Rebased off main and pushed. Let's see if tests pass - if they do, let's get this one in?

I'm very busy this week so I'll only be able to take a deep look at it next week. 
Apart from that, why are Brett Logan's commits part of this PR? I don't think rebasing on top of main should have them... how did these commits get there?

atoulme:
The PR is based off his PR. See "additional details" in the PR description.

yacovm:
Hey, sorry I am too busy this week to check this PR :( 

In fact I'll probably not be able to do it this month, I'm too overloaded with other things and I don't want to be sloppy in reviewing this.

atoulme:
Who else is available?

denyeart:
Discussed with Yacov... we concluded that given the potential risk of an incompatible mixed version orderer deployment that we really do need to wait for Yacov's final review in the coming weeks since he has the most expertise in this area.

yacovm:
> Discussed with Yacov... we concluded that given the potential risk of an incompatible mixed version orderer deployment that we really do need to wait for Yacov's final review in the coming weeks since he has the most expertise in this area.

@guoger is also a good candidate (better than me, in fact) to look at it.

Param-S:
@guoger can you review this PR

atoulme:
Hello folks, any update?

denyeart:
@Param-S Please rebase on latest main to resolve the conflicts. Then we still need a final review from @yacovm or @guoger .

Param-S:
@atoulme i don't have permission to rebase on your branch. can you rebase to resolve the conflicts.

atoulme:
@Param-S invite out to you so you can push. Feel free to also send a PR to my branch if that helps.

Param-S:
@denyeart @yacovm merge conflicts are resolved now. can you review the PR

atoulme:
Is anyone available to review? Please let us know.

orderer/consensus/etcdraft/node.go
```diff
@@ -201,6 +202,14 @@ func (n *node) send(msgs []raftpb.Message) {
...
+		if msg.Type == raftpb.MsgSnap {
+			state := n.confState.Load()
+			if state != nil {
+				msg.Snapshot.Metadata.ConfState.Voters = state.(*raftpb.ConfState).Voters
```
yacovm: Why did you decide to only override the `Voters` and not the entire conf state?
I see that conf state has more fields, like voters outgoing:

```
message ConfState {
	// The voters in the incoming config. (If the configuration is not joint,
	// then the outgoing config is empty).
	repeated uint64 voters = 1;
	// The learners in the incoming config.
	repeated uint64 learners          = 2;
	// The voters in the outgoing config.
	repeated uint64 voters_outgoing   = 3;
	// The nodes that will become learners when the outgoing config is removed.
	// These nodes are necessarily currently in nodes_joint (or they would have
	// been added to the incoming config right away).
	repeated uint64 learners_next     = 4;
	// If set, the config is joint and Raft will automatically transition into
	// the final config (i.e. remove the outgoing config) when this is safe.
	optional bool   auto_leave        = 5 [(gogoproto.nullable) = false];
}
```

Etcd overrides the conf state fully, not just the voters.

yacovm:
> > What if this node was restarted and then it means the confState is nil? In that case won't the voters contain the voters from before the config?
> 
> As per current understanding, the latest conf-state gets initialized as part of the orderer and raft memory storage initialization. The variable conf-state gets set when the first node becomes the leader, the confstate gets initialized as part of the InFlightConfChange check.
> 
> https://github.com/hyperledger/fabric/blob/7b7b6e9248b77be956025cf56eec3e7799f17c72/orderer/consensus/etcdraft/chain.go#L642
> 
> https://github.com/hyperledger/fabric/blob/7b7b6e9248b77be956025cf56eec3e7799f17c72/orderer/consensus/etcdraft/chain.go#L223
> 
> I verified the behavior with the following steps:
> 
>     1. Start Orderer1, 2, 3 with SnapshotIntervalSize = 1KB (It expects to take snap on every transaction)
> 
>     2. Issue config transaction to add Orderer 4
> 
>     3. Stop all the 3 orderers
> 
>     4. Start the Orderer 1, 2 & 4
> 
> 
> Added a breakpoints in Send method where the Snapshot updated with latest confstate. I can see that the leader updates the snapshot with the latest conf state which includes Orderer 4 in the consenters.
> 
> Please guide me if i need to validate with any other scenario.

I don't agree with your assessment. The code you mention deals with the corner case of an in-flight config change (implied by the **getInFlightConfChange** method name). 

The scenario you checked worked as you expected because the latest block **is a config block**. 

But I think that if the latest block won't be a config block, then it will not continue down the path you refer to, because it will return early [here](https://github.com/hyperledger/fabric/blob/7b7b6e9248b77be956025cf56eec3e7799f17c72/orderer/consensus/etcdraft/chain.go#L1333).

orderer/consensus/etcdraft/node.go
atoulme: I took this patch from the work by @guoger on https://github.com/hyperledger/fabric/pull/2707/files#diff-b7eaca6d9a2722c748ff597e17957ee6eafac930654420eb9a82d7df7d7bbf0bR209

I can make it take the whole ConfState instead.

yacovm: Let's do it, it can't hurt :)

atoulme: ok, done!

atoulme:
@yacovm I have assigned the whole conf state in my most recent commit.

yacovm:
> @yacovm I have assigned the whole conf state in my most recent commit.

I know I am waiting for Param's answer to my [comment](https://github.com/hyperledger/fabric/pull/2997#issuecomment-1086590654).

Param-S:
@yacovm I am currently checking on the scenario which you have pointed out above. I will update here once i complete it.

Param-S:
@yacovm For the scenario:

> What if this node was restarted and then it means the confState is nil? In that case won't the voters contain the voters from before the config?

> But I think that if the latest block won't be a config block, then it will not continue down the path you refer to, because it will return early [here](https://github.com/hyperledger/fabric/blob/7b7b6e9248b77be956025cf56eec3e7799f17c72/orderer/consensus/etcdraft/chain.go#L1333).

Further studied the scenario with the following steps: (Using 3 node ordering service),

1. Updated the consenters set with a new node(4th node) for a channel (config block n - 1th index)
2. Submitted a normal transaction - (normal block nth index)
3. Stopped all the 3 nodes
4. Started the nodes 1, 2 & 4

**While the node 1 & 2 starts up, the in-memory confState is set based on the WAL entries(raftLog) as part of the raft consenters set initialization.  The below log entries & stack trace was collected when raft applied the node 4 to the consenters set.**

```
(dlv) c
2022-04-10 11:43:33.914 PDT 0759 INFO [orderer.consensus.etcdraft] switchToConfig -> 2 switched to configuration voters=(1 2 3) channel=tmpchannel node=2

2022-04-10 11:43:33.914 PDT 075c INFO [orderer.consensus.etcdraft] apply -> Applied config change to add node 3, current nodes in channel: [1 2 3] channel=tmpchannel node=2

> github.com/hyperledger/fabric/orderer/consensus/etcdraft.(*node).ApplyConfChange() /root/hyperledger/etcdprotobuf/fabric/orderer/consensus/etcdraft/node.go:311 (hits goroutine(102):4 total:7) (PC: 0x1097eca)
   306:	func (n *node) lastIndex() uint64 {
   307:		i, _ := n.storage.ram.LastIndex()
   308:		return i
   309:	}
   310:
=> 311:	func (n *node) ApplyConfChange(cc raftpb.ConfChange) *raftpb.ConfState {
   312:		state := n.Node.ApplyConfChange(cc)
   313:		n.confState.Store(state)
   314:		return state
   315:	}
(dlv) stack
0  0x0000000001097eca in github.com/hyperledger/fabric/orderer/consensus/etcdraft.(*node).ApplyConfChange
   at /root/hyperledger/etcdprotobuf/fabric/orderer/consensus/etcdraft/node.go:311
1  0x0000000001087ebd in github.com/hyperledger/fabric/orderer/consensus/etcdraft.(*Chain).apply
   at /root/hyperledger/etcdprotobuf/fabric/orderer/consensus/etcdraft/chain.go:1087
2  0x0000000001082fa7 in github.com/hyperledger/fabric/orderer/consensus/etcdraft.(*Chain).run
   at /root/hyperledger/etcdprotobuf/fabric/orderer/consensus/etcdraft/chain.go:781
3  0x000000000107f92b in github.com/hyperledger/fabric/orderer/consensus/etcdraft.(*Chain).Start·dwrap·2
   at /root/hyperledger/etcdprotobuf/fabric/orderer/consensus/etcdraft/chain.go:368
4  0x00000000004788c1 in runtime.goexit
   at /usr/local/go/src/runtime/asm_amd64.s:1581
(dlv) print cc
go.etcd.io/etcd/raft/v3/raftpb.ConfChange {
	Type: ConfChangeAddNode (0),
	NodeID: 4,
	Context: []uint8 len: 0, cap: 0, nil,
	ID: 0,}
(dlv) c

2022-04-10 11:43:43.395 PDT 075f INFO [orderer.consensus.etcdraft] switchToConfig -> 2 switched to configuration voters=(1 2 3 4) channel=tmpchannel node=2
```

Based on this debugging, it is understood that the confstate is getting updated based on the latest WAL entries.  I hope this helps to follow the confstate update in the normal flow.  

Please let me know if i need to debug any other scenarios.

yacovm:
> Based on this debugging, it is understood that the confstate is getting updated based on the latest WAL entries. I hope this helps to follow the confstate update in the normal flow.

I don't understand why would Raft make Fabric do apply() after it came from a restart? It seems suspicious to me. Could it be that the nodes were shut down before it had the time to mark that the Raft config was persisted?  How much time did you wait between (1) and (3) ?

Param-S:
> How much time did you wait between (1) and (3) ? 

It should be a couple of minutes b/w each step as I executed each step manually. 

> I don't understand why would Raft make Fabric do apply() after it came from a restart?  

I think, it needs to initialize the in-memory storage with the raft configuration.  It reads from the disk(WAL entries) & creates the in-memory structures. 

To confirm the behaviour, I tried out these steps:

Based on the fabric-samples/test-network-nano-bash - 3 orderers cluster

1. Started the orderer1, 2, 3 & allowed to elect the leader & quorum
2. Killed all the orderers
3. Started the orderer1, 2 & added the breakpoint in ApplyConfChange

```
2022-04-12 03:10:07.117 PDT 02a9 INFO [orderer.consensus.etcdraft] switchToConfig -> 1 switched to configuration voters=() channel=test-system-channel-name node=1

=> 311:	func (n *node) ApplyConfChange(cc raftpb.ConfChange) *raftpb.ConfState {
   312:		state := n.Node.ApplyConfChange(cc)
   313:		n.confState.Store(state)
   314:		return state
   315:	}
(dlv) print cc
go.etcd.io/etcd/raft/v3/raftpb.ConfChange {
	Type: ConfChangeAddNode (0),
	NodeID: 1,
	Context: []uint8 len: 0, cap: 0, nil,
	ID: 0,}
	
2022-04-12 03:10:43.032 PDT 02ac INFO [orderer.consensus.etcdraft] switchToConfig -> 1 switched to configuration voters=(1) channel=test-system-channel-name node=1
2022-04-12 03:10:43.032 PDT 02ad INFO [orderer.consensus.etcdraft] apply -> Applied config change to add node 1, current nodes in channel: [1] channel=test-system-channel-name node=1

=> 311:	func (n *node) ApplyConfChange(cc raftpb.ConfChange) *raftpb.ConfState {
   312:		state := n.Node.ApplyConfChange(cc)
   313:		n.confState.Store(state)
   314:		return state
   315:	}
(dlv) print cc
go.etcd.io/etcd/raft/v3/raftpb.ConfChange {
	Type: ConfChangeAddNode (0),
	NodeID: 2,
	Context: []uint8 len: 0, cap: 0, nil,
	ID: 0,}
	
2022-04-12 03:11:30.358 PDT 02ae INFO [orderer.consensus.etcdraft] switchToConfig -> 1 switched to configuration voters=(1 2) channel=test-system-channel-name node=1
2022-04-12 03:11:30.358 PDT 02af INFO [orderer.consensus.etcdraft] apply -> Applied config change to add node 2, current nodes in channel: [1 2] channel=test-system-channel-name node=1

=> 311:	func (n *node) ApplyConfChange(cc raftpb.ConfChange) *raftpb.ConfState {
   312:		state := n.Node.ApplyConfChange(cc)
   313:		n.confState.Store(state)
   314:		return state
   315:	}
(dlv) print cc
go.etcd.io/etcd/raft/v3/raftpb.ConfChange {
	Type: ConfChangeAddNode (0),
	NodeID: 3,
	Context: []uint8 len: 0, cap: 0, nil,
	ID: 0,}	
	
2022-04-12 03:17:41.496 PDT 02b3 INFO [orderer.consensus.etcdraft] switchToConfig -> 2 switched to configuration voters=(1 2 3) channel=test-system-channel-name node=2
2022-04-12 03:17:41.496 PDT 02b4 INFO [orderer.consensus.etcdraft] apply -> Applied config change to add node 3, current nodes in channel: [1 2 3] channel=test-system-channel-name node=2
```

Based on this observation, I presume that the raft needs to initialize the initial nodes(in memory) as part of the start/restart of the node, which in turn calls back to the application(orderer).

yacovm:
@Param-S I think it works for you just because the latest config block was still relatively "fresh" and as a result it's still in the WAL and hasn't been garbage collected by snapshot. 

If you put many transactions between the last config and the restart of the node, you will see that **ApplyConfChange** will not be called. 

I made a small [test](https://github.com/yacovm/fabric/commit/058b330ae39414baab1146f244a8288d266bc319) where I put 100 transactions to enforce a snapshot that prunes the WAL:
```
			By("performing operation with orderer1")
			for i := 1; i < 100; i++ {
				env := CreateBroadcastEnvelope(network, o1, network.SystemChannel.Name, make([]byte, 1024 * 10))
				resp, err := ordererclient.Broadcast(network, o1, env)

				block := FetchBlock(network, o1, uint64(i), network.SystemChannel.Name)
				Expect(block).NotTo(BeNil())
			}
```

and added a print to **ApplyConfChange**: 

```
func (n *node) ApplyConfChange(cc raftpb.ConfChange) *raftpb.ConfState {
	debug.PrintStack()
	return n.Node.ApplyConfChange(cc)
}
```

and it doesn't print it after a restart.

Param-S:
@yacovm The current implementation of NewChain reads the confstate of latest snapshot and stores it chain object. 
https://github.com/hyperledger/fabric/blob/ccfa8a4a78f51e81bed877c8336e3de6141d3774/orderer/consensus/etcdraft/chain.go#L245
Now I updated the same flow to set the same value to node's confstate attribute which can be used later in the flow.

Param-S:
@yacovm Now, the ConfState picked up from the latest snapshot and at the node initialization time itself, it should address the issue. Could you check & confirm

yacovm (approved):

yacovm:
Looks good, let's get some more eyes on this.

atoulme:
How many eyes needed? Can this be merged please?

yacovm:
> How many eyes needed? Can this be merged please?

At least one more pair of eyes besides mine :-)

atoulme:
Do you have someone in mind?

yacovm:
> Do you have someone in mind?

Probably @guoger is best, afterwards maybe @C0rWin or @manish-sethi

C0rWin:
> > Do you have someone in mind?
> 
> Probably @guoger is best, afterwards maybe @C0rWin or @manish-sethi

I can do it over the weekend

atoulme:
@C0rWin any update?

atoulme:
@C0rWin any update? @yacovm anyone else available or can we please merge?

C0rWin:
> @C0rWin any update? @yacovm anyone else available or can we please merge?



This is quite colossal PR to review. Yes, it takes some time as I'm running several tests to see what output and check it... I know you have been doing it for a long while, but please have patience, and we will merge it.

atoulme:
Sure, any ETA?

atoulme:
@C0rWin any update please? Are the tests not covering enough this change? What can we do here?

C0rWin (approved):
@atoulme sorry, took me some time to review, run locally and seems the PR is fine, though a little bit too bigger than I would go with merge :) 

but given there is not other way, LGTM and thanks

atoulme:
Great, please merge?

SamYuan1990:
Hi guys,

so does this pr merge means we are completed with protobuf upgrade? and which means for PR been blocked by protobuf can be reopened in pr review process?
such as https://github.com/hyperledger/fabric/pull/3202 ?

thanks and regards
Sam

denyeart:
@SamYuan1990 Yes, please proceed now!

SamYuan1990:
> @SamYuan1990 Yes, please proceed now!

ok, @denyeart , please take a look at #3498

PR #2999, opened 2021-10-27
Retry logic for evaluate

### LINKED ISSUE #2913: Gateway: retry logic for evaluate method
Currently, the gateway's evaluate method will select a peer based on maximum block height in order to query the ledger.  If this fails due to connectivity reasons, then the error is returned to the client.
This enhancement will handle these errors by attempting to select an alternative peer, and retry the query.  If no peers can successfully process the query, then the error is returned to the client.
Note this logic won't apply to errors that are raised by the chaincode since it is assumed that all instances will return this same error (e.g. wrong number/type of arguments, etc).

andrew-coleman:
Modify the evaluate method to retry processing of a transaction proposal on another peer in the case of an error.

resolves #2913 

### DIFF
--- a/internal/pkg/gateway/api.go
+++ b/internal/pkg/gateway/api.go
@@ -53,7 +53,7 @@ func (gs *Server) Evaluate(ctx context.Context, request *gp.EvaluateRequest) (*g
 		transientProtected = true
 	}
 
-	endorser, err := gs.registry.evaluator(channel, chaincodeID, targetOrgs)
+	plan, err := gs.registry.evaluator(channel, chaincodeID, targetOrgs)
 	if err != nil {
 		if transientProtected {
 			return nil, status.Errorf(codes.Unavailable, "no endorsers found in the gateway's organization; retry specifying target organization(s) to protect transient data: %s", err)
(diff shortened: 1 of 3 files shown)


internal/pkg/gateway/registry.go
```diff
@@ -205,21 +205,31 @@ func (reg *registry) evaluator(channel string, chaincode string, targetOrgs []st
...
-			evaluator = e[0].endorser
-			maxHeight = e[0].height
+		if es, ok := endorsersByOrg[org]; ok {
+			for _, e := range es {
+				if org == reg.localEndorser.mspid {
+					localOrgEndorsers = append(localOrgEndorsers, e)
+				} else {
+					otherOrgEndorsers = append(otherOrgEndorsers, e)
+				}
+			}
```
manish-sethi: For loop can be avoided

```suggestion
			if org == reg.localEndorser.mspid {
				localOrgEndorsers = es
			}else{
				otherOrgEndorsers = append(otherOrgEndorsers, es...)
			}
```

internal/pkg/gateway/api.go
```diff
@@ -64,24 +64,41 @@ func (gs *Server) Evaluate(ctx context.Context, request *gp.EvaluateRequest) (*g
...
+			err = fmt.Errorf("error %d returned from chaincode %s on channel %s: %s", response.Status, chaincodeID, channel, response.Message)
+			endpointErr := errorDetail(endorser.endpointConfig, err)
+			errDetails = append(errDetails, endpointErr)
+			// this is a chaincode error response - don't retry
+			endorser = nil
+			response = nil
+		}
+	}
+	if endorser == nil || response == nil {
+		return nil, rpcError(codes.Aborted, "failed to evaluate transaction, see attached details for more info", errDetails...)
```
manish-sethi: From the readability perspective, setting variables to some values and checking latter for clubbing all cases of return error is not a preferred way. The suggested way would be to check for return condition in both if clauses and return early.

manish-sethi (commented):
LGTM - A couple cleanup comments

manish-sethi (approved):

PR #3002, opened 2021-10-29
Add explanation of stringArray (--peerAddresses)

gqqnbig:
#### Type of change

- Documentation update

#### Description

I searched the documentations but didn't find they explain stringArray. 

I looked at the source code of github.com/spf13/pflag and guess I should use

 `--peerAddresses localhost:9051--peerAddresses localhost:7051` rather than `--peerAddresses "localhost:9051
localhost:7051"`.

I think this information should be added to the docs.


denyeart (approved):
Thanks!

PR #3006, opened 2021-11-01
Add logging for identity, policy, and signature troubleshooting

denyeart:
Most identity, policy, and signature issues return a fairly generic error
message to the user, e.g. "not authorized".
This is often intentional so as to not disclose information to malicious users that
may be probing for information about the system.

This commit adds logging on the orderer and peer side so that identity issues
can more easily be troubleshooted by users setting up sample networks,
and by administrators and SREs in production networks.

For any identity, policy, or signature error, the failed policy and passed identity is now logged in a warning message.
Additionally, the identity of every signature that is verified can be seen if tracing is enabled.

The new logging can help with the following types of issue resolution:
User cert and MSP membership errors
Determine which user is unauthorized to perform an action
Determine which MSPs and user signatures are included in a config transaction that was invalidated
Determine which peers participated in an endorsement invalidation
Determine which peer signature doesn't match the others in a proposal response


core/common/validation/statebased/validator_keylevel.go
```diff
@@ -268,7 +268,12 @@ func (klv *KeyLevelValidator) Validate(cc string, blockNum, txNum uint64, rwsetB
 	}
 
 	// return the decision of the policy evaluator
-	return policyEvaluator.Evaluate(blockNum, txNum, rwset.NsRwSets, cc, signatureSet)
+	err := policyEvaluator.Evaluate(blockNum, txNum, rwset.NsRwSets, cc, signatureSet)
+	if err != nil {
+		// if endorsement policy check fails, log the endorser identities
+		logger.Warnw("Endorsment policy failure", "error", err, "chaincode", cc, "endorsingIdentities", protoutil.LogMessageForSerializedIdentities(signatureSet))
```
ale-linux: is it worth logging the policy as well?

msp/identities.go
```diff
@@ -182,14 +182,16 @@ func (id *identity) Verify(msg []byte, sig []byte) error {
 	}
 
 	if mspIdentityLogger.IsEnabledFor(zapcore.DebugLevel) {
-		mspIdentityLogger.Debugf("Verify: digest = %s", hex.Dump(digest))
-		mspIdentityLogger.Debugf("Verify: sig = %s", hex.Dump(sig))
+		mspIdentityLogger.Debugf("Verify: signer identity (certificate subject=%s issuer=%s serialnumber=%d)", id.cert.Subject, id.cert.Issuer, id.cert.SerialNumber)
+		// mspIdentityLogger.Debugf("Verify: digest = %s", hex.Dump(digest))
```
ale-linux, line 186: Why drop those?

protoutil/signeddata.go
```diff
@@ -80,3 +84,33 @@ func EnvelopeAsSignedData(env *common.Envelope) ([]*SignedData, error) {
...
+// LogMessageForSerializedIdentity returns a string with seriealized identity information,
+// or a string indicating why the serialized identity information cannot be returned.
+// Any errors are intentially returned in the return strings so that the function can be used in single-line log messages with minimal clutter.
+func LogMessageForSerializedIdentity(serializedIdentity []byte) string {
+	id := &msp.SerializedIdentity{}
+	err := proto.Unmarshal(serializedIdentity, id)
+	if err != nil {
+		return fmt.Sprintf("Could not unmarshal serialized identity: %s", err)
+	}
+	pemBlock, _ := pem.Decode(id.IdBytes)
```
ale-linux, line 97: can we log at least the length of rest?

protoutil/testdata/peer-expired.pem
```diff
@@ -0,0 +1,14 @@
+-----BEGIN CERTIFICATE-----
```
ale-linux, line 1: it would be good to track the command used to generate this fixture

core/common/validation/statebased/validator_keylevel.go
denyeart: yes, good idea, the ccEP is a byte array, is there an existing function to turn that into a human readable policy?

msp/identities.go
denyeart: Digest and signature validation are well-trodden. I can't think of anytime in the last five years where this debug has been useful, but it fills the logs with several lines of random bytes. So I thought while adding useful debug I could also remove un-useful debug. I left it commented so that a developer could easily enable if they like. But if you think it is valuable to keep, I can keep.

protoutil/testdata/peer-expired.pem
denyeart: I copied this cert from the msp testdata... do you know if the msp tests mention how they were generated?

protoutil/signeddata.go
denyeart: Sorry, not understanding, log the length of what?

core/common/validation/statebased/validator_keylevel.go
ale-linux: yeah that is tricky since we support different types of policy. We could try to unmarshal the bytes as 

https://github.com/hyperledger/fabric/blob/main/core/policy/application.go#L148 and print with %+v or %#v

msp/identities.go
ale-linux: No I trust you if you say that it hasn't been useful - SGTM to remove then

protoutil/signeddata.go
ale-linux: of the rest, basically pem.Decode returns block, rest; but here we don't even assign a variable name to rest. However, for malformed PEM data, there might be trailing, undecoded data

protoutil/testdata/peer-expired.pem
ale-linux: ha! no, unfortunately.. I think we would need to script the regeneration of those fixture (or at least document how we generated them). For another PR though

core/common/validation/statebased/validator_keylevel.go
denyeart: Thanks @ale-linux I tried it out and that approach works well for both signature policies and channel config reference policies, I'll push another one...

protoutil/signeddata.go
denyeart: I haven't seen malformed PEM data in the last five years either, I think I'll just keep it simple.

denyeart:
@ale-linux I think it is ready for final review now...

ale-linux (approved):

PR #3010, opened 2021-11-02
Update transaction flow doc for peer gateway

denyeart:
Update transaction flow doc for the new v2.4 flow with the peer gateway.


denyeart:
Note - some of the graphics will also need to be updated.

denali49 (approved):
lgtm, thanks @denyeart

PR #3017, opened 2021-11-04
Better gRPC error on context error from CommitStatus service

bestbeforetoday:
Rather than always return a FailedPrecondition error on failure obtaining commit status, return either a DeadlineExceeded or Canceled error on a context error. This may happen if the call is cancelled from the client end, either by an explicit context cancel or timeout.

### DIFF
--- a/internal/pkg/gateway/apiutils.go
+++ b/internal/pkg/gateway/apiutils.go
@@ -45,7 +45,7 @@ func getChannelAndChaincodeFromSignedProposal(signedProposal *peer.SignedProposa
 	return channelHeader.ChannelId, spec.ChaincodeSpec.ChaincodeId.Name, len(payload.TransientMap) > 0, nil
 }
 
-func rpcError(code codes.Code, message string, details ...proto.Message) error {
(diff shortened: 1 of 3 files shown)


internal/pkg/gateway/apiutils.go
```diff
@@ -45,7 +45,7 @@ func getChannelAndChaincodeFromSignedProposal(signedProposal *peer.SignedProposa
 	return channelHeader.ChannelId, spec.ChaincodeSpec.ChaincodeId.Name, len(payload.TransientMap) > 0, nil
 }
 
-func rpcError(code codes.Code, message string, details ...proto.Message) error {
+func newRpcError(code codes.Code, message string, details ...proto.Message) error {
```
ale-linux, line 48: Out of curiosity, why the change of function name?

ale-linux (approved):

internal/pkg/gateway/apiutils.go
bestbeforetoday: I introduced a `toRpcError` function and just wanted the names to be a little more distinct to avoid confusion

PR #3031, opened 2021-11-09
ProposalResponse error should return error

andrew-coleman:
A transaction proposal can fail for two reasons:
1) The chaincode return an error (status 500) response
2) Some infrastructure error occurs or error generated in the peer.

For case (1), the ProcessProposal() function will return the ProposalResponse and a nil error.
For case (2), most error paths return both a newly generated status 500 response and the error.  There is one path though (the one that invokes the simulateProposal) that just returns a status 500 response, but a nil error.
This means that client applications (via SDKs or Gateway) cannot destinguish between chaincode raised errors (case 1) and infrastructure errors (case 2).  This is a problem for the gateway which needs to destinguish these two situations in its retry logic.

This commit fixes this by returning the original error (instead of nil) as well as the generated ProposalResponse for this error path.

### DIFF
--- a/core/endorser/endorser.go
+++ b/core/endorser/endorser.go
@@ -345,7 +345,7 @@ func (e *Endorser) ProcessProposal(ctx context.Context, signedProp *pb.SignedPro
 	pResp, err := e.ProcessProposalSuccessfullyOrError(up)
 	if err != nil {
 		endorserLogger.Warnw("Failed to invoke chaincode", "channel", up.ChannelHeader.ChannelId, "chaincode", up.ChaincodeName, "error", err.Error())
-		return &pb.ProposalResponse{Response: &pb.Response{Status: 500, Message: err.Error()}}, nil
+		return &pb.ProposalResponse{Response: &pb.Response{Status: 500, Message: err.Error()}}, err
 	}
 
 	if pResp.Endorsement != nil || up.ChannelHeader.ChannelId == "" {
--- a/core/endorser/endorser_test.go
+++ b/core/endorser/endorser_test.go
@@ -286,7 +286,7 @@ var _ = Describe("Endorser", func() {
 
 		It("returns the error, but with no payload encoded", func() {
 			proposalResponse, err := e.ProcessProposal(context.Background(), signedProposal)
-			Expect(err).NotTo(HaveOccurred())
+			Expect(err).To(MatchError("endorsing with plugin failed: fake-endorserment-error"))
 			Expect(proposalResponse.Payload).To(BeNil())
 			Expect(proposalResponse.Response).To(Equal(&pb.Response{
 				Status:  500,
@@ -337,7 +337,7 @@ var _ = Describe("Endorser", func() {
 
 		It("returns a response with the error", func() {
 			proposalResponse, err := e.ProcessProposal(context.Background(), signedProposal)
-			Expect(err).NotTo(HaveOccurred())
+			Expect(err).To(MatchError("fake-simulator-error"))
 			Expect(proposalResponse.Payload).To(BeNil())
 			Expect(proposalResponse.Response).To(Equal(&pb.Response{
 				Status:  500,
@@ -361,7 +361,7 @@ var _ = Describe("Endorser", func() {
 
 		It("returns a response with the error", func() {
 			proposalResponse, err := e.ProcessProposal(context.Background(), signedProposal)
-			Expect(err).NotTo(HaveOccurred())
+			Expect(err).To(MatchError("fake-history-error"))
 			Expect(proposalResponse.Payload).To(BeNil())
 			Expect(proposalResponse.Response).To(Equal(&pb.Response{
 				Status:  500,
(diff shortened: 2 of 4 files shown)


ale-linux (commented):
@jyellick reworked this code path (I remember reviewing the code) - and he was usually very careful with these things. @jyellick - any chance you still remember whether you intentionally did not propagate the error back to the SDK?

yacovm:
I'm not Jason but I think that it might fail some gRPC clients if the RPC returns an error in the Go server-side code

yacovm (changes requested):
I'm not Jason but I think that it will omit the application message (left argument returned) if the RPC returns an error in the Go server-side code so the 500 status won't be propagated to the client application layer.

Specifically, look at [this code](https://github.com/grpc/grpc-go/blob/master/server.go#L1282-L1310) of the gRPC server side processing.

We should have probably [investigated](https://github.com/avinassh/grpc-errors/blob/master/go/server.go#L22) how gRPC error handling should be [done](https://www.grpc.io/docs/guides/error/).

Some Fabric context: https://jira.hyperledger.org/browse/FAB-4347 

> This is a problem for the gateway which needs to destinguish these two situations in its retry logic.

It might be, but, we also need to ensure we are not harming the [interpretation](https://imgs.xkcd.com/comics/workflow.png) of all other existing clients since 2017.

ale-linux (changes requested):
Good point @yacovm, yes. The risk isn't high since there's no way this change might lead to a fork, but you have a point that there was probably a lot of tooling/scripting developed that - for better or worse - interpreted errors as we generate them right now.

I think we need to think this change through a little bit more. Thoughts?

yacovm:
I don't think this should be the solution. We can't change API semantics of (among) the most fundamental gRPC call we have. We should investigate other ways to achieving what he wants, probably via HTTP headers / metadata or expanding a protobuf message or something.

@denyeart  wdyt?

denyeart:
We do need to be very careful here. But we do need to improve the error handling one way or another. Let's start the deeper thinking by analyzing the exact impact to clients in more detail. @andrew-coleman can you provide more details beyond the original description with respect to exact impact to clients. Concerning infrastructure errors, you said "most error paths return both a newly generated status 500 response and the error" already. So does this imply the new error handling will likely be handled well by clients since it simply expands the set infrastructure errors clients already handle?

Do you see any good alternative path using additional metadata or proto extensions? I don't think we want to go with an awkward extension, but if there is one that makes sense we could go that approach.

If we think the impact on clients is none to minor, we could describe the PR change in release notes as a bug fix, and explain what clients should watch out for when upgrading to v2.4.
If we think the impact on clients is significant, then we probably want to wait for a v3.0 release or choose another approach.

yacovm:
> If we think the impact on clients is none to minor, we could describe the PR change in release notes as a bug fix, and explain what clients should watch out for when upgrading to v2.4.
> If we think the impact on clients is significant, then we probably want to wait for a v3.0 release or choose another approach.

You cannot quantify that because you don't know how users use our APIs. 
If we return an error, that suppresses the application layer response and some users may use it for something. 
I think that we should investigate an alternative that doesn't effect existing semantics and in v3.0 we can revisit.

denyeart:
I did some project archeology. In v1.0 proposal processing errors resulted in Go err response, which eventually becomes a grpc error. In v1.2 the behavior was intentionally changed to return a proposal response with 500 and nil err, because users wanted their clients to base error handling off of the error code (500 in these cases).  A similar discussion occurred at that time... Yacov asserted that the change would cause incompatibility for existing clients, but it was eventually decided that it would be better to make the change sooner than later and it was agreed by all to make the change for v1.2.

Andy's PR would revert the intentional change in v1.2 back to the prior v1.0 behavior, so I have to agree with Yacov that we can't make this change.

Especially when looking at the v1.4 codebase, it is clear that the design intent is for any proposal processing errors to return a proposal response with 500 and nil error. Only issues with the proposal itself return an error, such as unpacking and preprocessing validation on the proposal. The design intent for proposal processing errors became less clear in the refactored code in v2.x since there is a single ProcessProposalSuccessfullyOrError() function that encapsulates all the processing errors that ends up returning a single nil error. Turns out this single nil error is very much intentional.

I've added PR https://github.com/hyperledger/fabric/pull/3044 to clarify the design intent.

core/endorser/endorser.go
```diff
@@ -345,7 +345,7 @@ func (e *Endorser) ProcessProposal(ctx context.Context, signedProp *pb.SignedPro
 	pResp, err := e.ProcessProposalSuccessfullyOrError(up)
 	if err != nil {
 		endorserLogger.Warnw("Failed to invoke chaincode", "channel", up.ChannelHeader.ChannelId, "chaincode", up.ChaincodeName, "error", err.Error())
-		return &pb.ProposalResponse{Response: &pb.Response{Status: 500, Message: err.Error()}}, nil
+		return &pb.ProposalResponse{Response: &pb.Response{Status: 500, Message: err.Error()}}, err
```
denyeart, line 348: Turns out the nil error response was an intentional design change in v1.2.

denyeart (changes requested):

andrew-coleman:
Thanks for the explanation (and history lesson :) ).  Where I was hoping to go with this ultimately was to adopt the grpc status error model (https://pkg.go.dev/google.golang.org/grpc/status) and associate the different paths with an appropriate status [code](https://pkg.go.dev/google.golang.org/grpc@v1.42.0/codes#Code).  So, for example, if a chaincode call times out, [the error that gets generated](https://github.com/hyperledger/fabric/blob/main/core/chaincode/handler.go#L1183) could instead be
`err = status.Error(codes.DeadlineExceeded, "timeout expired while executing transaction")`
which would then let client applications (and the gateway) take followup actions based on the type of error (Google gives guidance on this) rather than having to parse the text message in the error.

PR #3036, opened 2021-11-09
Add extra info to error message

andrew-coleman:
When processing an endorsement plan, failure to connect to remote peers could lead to an error saying there aren’t enough endorsers to satisfy the policy.  This commit add the list of peers that failed to commit.

### DIFF
--- a/internal/pkg/gateway/registry.go
+++ b/internal/pkg/gateway/registry.go
@@ -66,6 +66,7 @@ func (reg *registry) endorsementPlan(channel string, interest *peer.ChaincodeInt
 	// Also build a map of endorser pkiid to groupId
 	groupEndorsers := map[string][]*endorser{}
 	var preferredGroup string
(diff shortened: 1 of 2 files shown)


internal/pkg/gateway/registry.go
```diff
@@ -140,7 +142,7 @@ layout:
 	layouts := append(preferredLayouts, otherLayouts...)
 
 	if len(layouts) == 0 {
-		return nil, fmt.Errorf("failed to select a set of endorsers that satisfy the endorsement policy")
+		return nil, fmt.Errorf("failed to select a set of endorsers that satisfy the endorsement policy due to unavailability of peers: %v", unavailableEndorsers)
```
denyeart, line 145: I prefer to use the specific type like %s rather than the generic %v for readability, but that doesn't need to hold up this PR.

denyeart (approved):

PR #3040, opened 2021-11-11
Facilitate unaware threshold signature endorsement

yacovm:
This commit contains the following two seemingly unrelated changes, each on its own preserving current Fabric behavior:

1. **Deduplicate endorsements in createTx**
This commit makes createTx de-duplicate endorsers according to their identity.
Meaning, when multiple but identical endorsements (same identity) are passed to
the createTx function, it returns a transaction with unique endorsements, and without repeated endorsers.
Since at validation, endorsers are anyway de-duplicated, this commit preserves current system behavior.

2. **Support identity based endorsement policy in discovery**
Currently, whenever a chaincode's endorsement policy contains an identity principal, discovery rejects the policy,
saying that it contains a principal it does not support. This commit simply makes discovery classify the principal.

**Making clients and gateways unaware of a threshold signature scheme**
This commit enables advanced use cases, in which case the endorsements collected for transaction
submission contain identical identities that represent a threshold or aggregation signature scheme.

In such a case, peers can be deployed with custom endorsement plugins and generate a threshold/aggregation signature,
but the flow of service discovery, the client, and peer gateway remains the same.

The endorsement policy can to be an OR over two mutually exclusive cases:
1. An AND over a set of organizational principals (such as Org1.Peer and Org2.Peer)
2. An OR over one or more identities (which can only be signed in collaboration among the peers)
Discovery will always return only combinations of the former kind since no peer can identify with an identity of the latter kind.

The flow will then be as follows: 
1. The client and the gateway interact as usual, and the gateway collects endorsements from several peers.
2. The peers, using the endorsement plugin, engage in a protocol at which end they all return the same endorsement.
3. The gateway submits the transaction (this time, with a single endorsement) to the orderer.
4. Upon transaction validation we validate a single endorsement instead of several.

By enabling threshold signatures, we can maintain security (multiple organizations execute the smart contract and sign over the execution results) without sacrificing performance (transaction ends up with having a single endorsement)

Change-Id: I9193187ae6b08791f8762a9d325442d156c4f828

### DIFF
--- a/common/policies/inquire/compare.go
+++ b/common/policies/inquire/compare.go
@@ -21,6 +21,7 @@ type ComparablePrincipal struct {
 	ou        *msp.OrganizationUnit
 	role      *msp.MSPRole
 	mspID     string
+	idBytes   []byte
 }
 
 // Equal returns whether this ComparablePrincipal is equal to the given ComparablePrincipal.
@@ -44,6 +45,8 @@ func NewComparablePrincipal(principal *msp.MSPPrincipal) *ComparablePrincipal {
 		return cp.ToRole()
 	case msp.MSPPrincipal_ORGANIZATION_UNIT:
 		return cp.ToOURole()
+	case msp.MSPPrincipal_IDENTITY:
+		return cp.ToIdentity()
 	}
 	mapping := msp.MSPPrincipal_Classification_name[int32(principal.PrincipalClassification)]
 	logger.Warning("Received an unsupported principal type:", principal.PrincipalClassification, "mapped to", mapping)
@@ -101,6 +104,10 @@ func (cp *ComparablePrincipal) IsA(other *ComparablePrincipal) bool {
 		return this.role.Role == other.role.Role
 	}
 
+	if this.principal.PrincipalClassification == msp.MSPPrincipal_IDENTITY {
+		return bytes.Equal(this.idBytes, other.idBytes) && this.mspID == other.mspID
+	}
+
 	// Else, we can't say anything, because we have no knowledge
 	// about the OUs that make up the MSP roles - so return false
 	return false
@@ -116,6 +123,19 @@ func (cp *ComparablePrincipal) ToOURole() *ComparablePrincipal {
 	}
 	cp.mspID = ouRole.MspIdentifier
 	cp.ou = ouRole
+	return cp
+}
+
+// ToIdentity converts this ComparablePrincipal to Identity principal, and returns nil on failure
+func (cp *ComparablePrincipal) ToIdentity() *ComparablePrincipal {
+	sID := &msp.SerializedIdentity{}
+	err := proto.Unmarshal(cp.principal.Principal, sID)
+	if err != nil {
+		logger.Warning("Failed unmarshalling principal:", err)
+		return nil
+	}
+	cp.mspID = sID.Mspid
+	cp.idBytes = sID.IdBytes
 	return cp
 }
 
--- a/common/policies/inquire/compare_test.go
+++ b/common/policies/inquire/compare_test.go
@@ -47,8 +47,17 @@ func TestNewComparablePrincipal(t *testing.T) {
 		require.Nil(t, NewComparablePrincipal(nil))
 	})
 
-	t.Run("Invalid principal type", func(t *testing.T) {
-		require.Nil(t, NewComparablePrincipal(identity(mspID)))
+	t.Run("Identity", func(t *testing.T) {
+		expectedPrincipal := &ComparablePrincipal{
+			mspID:   mspID,
+			idBytes: []byte("identity"),
+			principal: &msp.MSPPrincipal{
+				PrincipalClassification: msp.MSPPrincipal_IDENTITY,
+				Principal:               protoutil.MarshalOrPanic(&msp.SerializedIdentity{IdBytes: []byte("identity"), Mspid: mspID}),
+			},
+		}
+
+		require.Equal(t, expectedPrincipal, NewComparablePrincipal(identity(mspID, []byte("identity"))))
 	})
 
 	t.Run("Invalid principal input", func(t *testing.T) {
@@ -88,6 +97,9 @@ func TestNewComparablePrincipal(t *testing.T) {
 }
 
 func TestIsA(t *testing.T) {
+	id1 := NewComparablePrincipal(identity("Org1MSP", []byte("identity")))
+	id2 := NewComparablePrincipal(identity("Org2MSP", []byte("identity")))
+	id3 := NewComparablePrincipal(identity("Org1MSP", []byte("identity2")))
 	member1 := NewComparablePrincipal(member("Org1MSP"))
 	member2 := NewComparablePrincipal(member("Org2MSP"))
 	peer1 := NewComparablePrincipal(peer("Org1MSP"))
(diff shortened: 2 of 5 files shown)


denyeart:
@yacovm Could you clarify the description with more background/motivation?
- What is the ultimate goal? I assume faster validation due to a single endorsement check is one goal. Are there others?
- How would threshold/aggregate signatures work? Wouldn't the gateway have to be enhanced to combine the individual endorsements before submitting the transaction?
- The part about "The endorsement policy can to be an OR over two mutually exclusive cases" confuses me... how is that related to the overall motivation?

yacovm:
>     * What is the ultimate goal? I assume faster validation due to a single endorsement check is one goal. Are there others?

Not just faster validation but reduction in transaction size, by only having a single identity. 


>     * How would threshold/aggregate signatures work? Wouldn't the gateway have to be enhanced to combine the individual endorsements before submitting the transaction?

That's the idea behind (1) - the gateway will get from each endorser the same identical certificate and it will then de-duplicate it.


>     * The part about "The endorsement policy can to be an OR over two mutually exclusive cases" confuses me... how is that related to the overall motivation?

You see, the gateway, or other clients - use discovery, right? But there is no way to make discovery convey information about a threshold signature. So we do a trick - we make an endorsement policy that can be satisfied in two disjoint ways:
(a) An AND over the peers that the threshold signature is composed of, to make discovery choose these peers for the client/gateway.
(b) An identity principal who's public key corresponds to a private key that no single peer has, and to make a signature that is verified under that identity you do a threshold/aggregation signature scheme protocol. 

I added: "By enabling threshold signatures, we can maintain security (multiple organizations execute the smart contract and sign over the execution results) without sacrificing performance (transaction ends up with having a single endorsement)" to the description.

denyeart:
Ok, I'm getting closer...

But with threshold signatures I thought each party (each peer) would sign with a partial signature that then gets aggregated. How can each peer return the same endorsement? I guess you are suggesting that the peers coordinate and aggregate the signatures prior to returning endorsement response? How would the peers realistically do this when they don't know which peers the gateway picked for endorsement (assuming there are multiple orgs to choose from and multiple peers per org).
Wouldn't it be more feasible to have each peer return their partial signature to the gateway, and the gateway then aggregates them into a single endorsement? Yes, that would require more changes than you are proposing, but first class support of threshold signatures would be a nice feature wouldn't it? Or maybe not entirely first class but at least expose some patterns to make it easier for users.

We should probably get some consensus on the overall approach before proceeding with a PR. I realize that you were attempting the path with least Fabric changes, but let's make sure we have some consensus on that path.

yacovm:
> But with threshold signatures I thought each party (each peer) would sign with a partial signature that then gets aggregated.

Usually when you say "threshold" you mean **t** out of **n** where **t < n**. 
You can also aggregate signatures but then it's always **n** out of **n**. 
Let me give you an example to demonstrate. Suppose you have a private key **sk=sk1+sk2+sk3** and you can aggregate its public keys **pk1**, **pk2**, **pk3** by adding them into a single public key. This is easy, but the downside is that if one of the private keys is lost, no one can sign anymore. In Fabric it's less of a problem because you can specify several combinations of aggregated keys. 
In a threshold scheme, you have a protocol which generates **t**<**n** private keys (and corresponding public keys) and then if one of the private keys is lost, then it's fine because you still have some private keys left (unless you end up with less than **t** of them). The downside here, is that the threshold public key (and corresponding private keys for all parties) generation algorithm is very complex and in the aggregation case, it is very simple.

Now, some threshold schemes allow you to easily "glue" signatures together, some require you to put some effort (communication rounds), and some require you to put a lot of effort (communication rounds, and all kinds of crypto sprinkled on top).

> How can each peer return the same endorsement? I guess you are suggesting that the peers coordinate and aggregate the signatures prior to returning endorsement response?

Exactly. The endorsement plugin will trigger a protocol in which end it will output a uniform endorser certificate according to the membership of the peers who were involved in the protocol.

> How would the peers realistically do this when they don't know which peers the gateway picked for endorsement (assuming there are multiple orgs to choose from and multiple peers per org).

That's a good question. The peers that are selected, have the same transaction ID. All that is needed is some sort of "distributed [barrier](https://en.wikipedia.org/wiki/Barrier_(computer_science))".

> Wouldn't it be more feasible to have each peer return their partial signature to the gateway, and the gateway then aggregates them into a single endorsement? 

Easier said than done. First of all, Fabric always signs over the identity of the peer as well, so we already have a mismatch. 
Even if that wasn't the case, then among Elliptic Curve based signatures, only BLS is that friendly. ed25519 which exists in Go, requires some communication rounds among the endorsers either directly or indirectly. ECDSA is the worst and it requires a complex full-blown MPC protocol.

> Yes, that would require more changes than you are proposing, but first class support of threshold signatures would be a nice feature wouldn't it? Or maybe not entirely first class but at least expose some patterns to make it easier for users.

Both ed25519 and BLS require changes to Fabric validation. The only scheme that Fabric supports today, as you know, is ECDSA.

> I realize that you were attempting the path with least Fabric changes, but let's make sure we have some consensus on that path.

Precisely, I am working with what I have, because I know that somehow, in the last few years, feature development takes a decade. 

I don't understand why you're opposed to this. These two changes do not influence anyone and make sense individually. 

I will achieve what I need with this change or without it. 
If you don't merge this, I will just use Fabric Smart Client nodes as endorsers. 
I thought that it can be nice if I can show how you can do threshold signatures in Fabric vanilla, using service discovery and gateway. I guess even this tiny change is too much.

adecaro:
Thanks @yacovm for submitting a very simple change that can enable so much. 
Threshold has many benefits and also complexities to be handled with care.
This PR goes in the direction of minimal changes to Fabric to enable a larger eco-system around it.
It looks good to me. Thanks.

denyeart:
> I don't understand why you're opposed to this.

I'm not opposed to this. It just didn't seem like the approach would be very obvious or consumable for users deploying Fabric, so I was poking on whether there may be a more obvious and more consumable approach possible. It sounds like there is not, so I'm fine with it as is, as I agree threshold signature support is a valuable addition.

Could I ask that you extend the https://hyperledger-fabric.readthedocs.io/en/latest/endorsement-policies.html doc topic with an Advanced section that provides some hints as to how more advanced deployments may leverage the new capability?

yacovm:
> Could I ask that you extend the https://hyperledger-fabric.readthedocs.io/en/latest/endorsement-policies.html doc topic with an Advanced section that provides some hints as to how more advanced deployments may leverage the new capability?

No, because the so called new capability hasn't been implemented yet and is entirely based on future implementation of an endorsement plugin. 

After I implement and test an end to end scenario I can then update the bespoken documentation, but in order for the former to happen I need this PR merged.

ale-linux (approved):
Thanks @yacovm and all those who have weighed in. After a thorough review, I only see reasons in favour of merging this PR, which I summarise here:

* the changes are small and uncontroversial
* the changes are safe, they do not impact verification and commit
* the changes are a first step towards the larger goal of making fabric txes smaller
* the changes would make sense even without the supporting use-case of threshold cryptography (in fact I would go as far as to argue that we should push similar changes in the SDKs; maybe they actually already do this?)

For these reasons the changes are okay for me. I'm approving but not merging yet just to give the others a chance to shout if they disagree

denyeart (approved):

denyeart:
Thanks @adecaro and @ale-linux ! I'll go ahead and merge.

PR #3045, opened 2021-11-15
Enhance gateway error logic

andrew-coleman:
Add logic to determine the origin of the error returned by ProcessProposal, and whether the evaluate/endorse methods should close the connection to the peer and/or retry the proposal on another peer.

### DIFF
--- a/internal/pkg/gateway/api.go
+++ b/internal/pkg/gateway/api.go
@@ -10,6 +10,7 @@ import (
 	"context"
 	"fmt"
 	"math/rand"
+	"strings"
 	"sync"
 
 	"github.com/golang/protobuf/proto"
@@ -19,6 +20,7 @@ import (
 	"github.com/hyperledger/fabric-protos-go/peer"
 	"github.com/hyperledger/fabric/common/ledger"
 	"github.com/hyperledger/fabric/core/aclmgmt/resources"
+	"github.com/hyperledger/fabric/core/chaincode"
 	"github.com/hyperledger/fabric/internal/pkg/gateway/event"
 	"github.com/hyperledger/fabric/protoutil"
 	"google.golang.org/grpc/codes"
@@ -71,32 +73,28 @@ func (gs *Server) Evaluate(ctx context.Context, request *gp.EvaluateRequest) (*g
 		gs.logger.Debugw("Sending to peer:", "channel", channel, "chaincode", chaincodeID, "MSPID", endorser.mspid, "endpoint", endorser.address)
 
 		pr, err := endorser.client.ProcessProposal(ctx, signedProposal)
-		if err != nil {
-			logger.Debugw("Evaluate call to endorser failed", "chaincode", chaincodeID, "channel", channel, "txid", request.TransactionId, "endorserAddress", endorser.endpointConfig.address, "endorserMspid", endorser.endpointConfig.mspid, "error", err)
-			errDetails = append(errDetails, errorDetail(endorser.endpointConfig, err))
-			gs.registry.removeEndorser(endorser)
-			endorser = plan.retry(endorser)
-			if endorser == nil {
-				return nil, newRpcError(codes.Aborted, "failed to evaluate transaction, see attached details for more info", errDetails...)
-			}
-		}
-
-		response = pr.GetResponse()
-		if response != nil {
-			if response.Status < 200 || response.Status >= 400 {
-				logger.Debugw("Evaluate call to endorser returned a malformed or error response", "chaincode", chaincodeID, "channel", channel, "txid", request.TransactionId, "endorserAddress", endorser.endpointConfig.address, "endorserMspid", endorser.endpointConfig.mspid, "status", response.Status, "message", response.Message)
-				err = fmt.Errorf("error %d returned from chaincode %s on channel %s: %s", response.Status, chaincodeID, channel, response.Message)
-				endpointErr := errorDetail(endorser.endpointConfig, err)
-				errDetails = append(errDetails, endpointErr)
-				// this is a chaincode error response - don't retry
-				return nil, newRpcError(codes.Aborted, "evaluate call to endorser returned error: "+response.Message, errDetails...)
-			}
-
+		success, message, retry, remove := gs.responseStatus(pr, err)
+		if success {
+			response = pr.Response
 			// Prefer result from proposal response as Response.Payload is not required to be transaction result
 			if result, err := getResultFromProposalResponse(pr); err == nil {
 				response.Payload = result
 			} else {
 				logger.Warnw("Successful proposal response contained no transaction result", "error", err.Error(), "chaincode", chaincodeID, "channel", channel, "txid", request.TransactionId, "endorserAddress", endorser.endpointConfig.address, "endorserMspid", endorser.endpointConfig.mspid, "status", response.Status, "message", response.Message)
+			}
+		} else {
+			logger.Debugw("Evaluate call to endorser failed", "chaincode", chaincodeID, "channel", channel, "txid", request.TransactionId, "endorserAddress", endorser.endpointConfig.address, "endorserMspid", endorser.endpointConfig.mspid, "error", message)
+			errDetails = append(errDetails, errorDetail(endorser.endpointConfig, message))
+			if remove {
+				gs.registry.removeEndorser(endorser)
+			}
+			if retry {
+				endorser = plan.nextPeerInGroup(endorser)
+			} else {
+				return nil, newRpcError(codes.Aborted, "evaluate call to endorser returned error: "+message, errDetails...)
+			}
+			if endorser == nil {
+				return nil, newRpcError(codes.Aborted, "failed to evaluate transaction, see attached details for more info", errDetails...)
 			}
 		}
 	}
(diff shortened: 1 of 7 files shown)


internal/pkg/gateway/api.go
```diff
@@ -273,6 +270,44 @@ func (gs *Server) Endorse(ctx context.Context, request *gp.EndorseRequest) (*gp.
...
+// responseStatus unpacks the proposal response and error values that are returned from ProcessProposal and
+// determines how the gateway should react (retry?, close connection?).
+// Uses the grpc canonical status error codes and their recommended actions.
+// Returns:
+// - response successful (bool)
+// - error message extracted from the err or generated from 500 proposal response
+// - should the gateway retry (only the Evaluate() uses this)
+// - should the gateway close the connection and remove the peer from its registry
+func (gs *Server) responseStatus(response *peer.ProposalResponse, err error) (bool, string, bool, bool) {
+	// if the ProcessProposal function returned an err, then it must have been from the unpackProposal() or preProcess() stages
```
denyeart: grpc issues will also return an err, right?

internal/pkg/gateway/api.go
```diff
@@ -273,6 +270,44 @@ func (gs *Server) Endorse(ctx context.Context, request *gp.EndorseRequest) (*gp.
...
 }
 
+// responseStatus unpacks the proposal response and error values that are returned from ProcessProposal and
+// determines how the gateway should react (retry?, close connection?).
+// Uses the grpc canonical status error codes and their recommended actions.
+// Returns:
+// - response successful (bool)
+// - error message extracted from the err or generated from 500 proposal response
+// - should the gateway retry (only the Evaluate() uses this)
+// - should the gateway close the connection and remove the peer from its registry
```
denyeart: What exactly does remove from registry mean? Just until the next time we check with discovery? Or longer?

internal/pkg/gateway/api.go
```diff
@@ -273,6 +270,44 @@ func (gs *Server) Endorse(ctx context.Context, request *gp.EndorseRequest) (*gp.
...
+	if err != nil {
+		if response == nil {
+			// if there is no ProposalResponse, then this must have been generated by grpc in response to an unavailable peer
+			// - close the connection and retry on another
+			return false, err.Error(), true, true
+		}
+		return false, err.Error(), false, false
+	}
+	if response.Response.Status < 200 || response.Response.Status >= 400 {
+		// if response contains a payload, then it must have been generated by the peer
```
denyeart:
```suggestion
		// if response contains a payload, then it must have been generated by the peer
```
```suggestion
		// if response doesn't contain a payload, then it must have been generated by the peer rather than by chaincode
```

internal/pkg/gateway/api.go
```diff
@@ -273,6 +270,44 @@ func (gs *Server) Endorse(ctx context.Context, request *gp.EndorseRequest) (*gp.
...
+		if response.Payload == nil && response.Response.Status == 500 {
+			// if there's a error 500 response but no payload, then it was generated in the peer
```
denyeart: Half of the comment is above the If statement, half below the If statement. This makes it confusing to read.

internal/pkg/gateway/api.go
```diff
@@ -273,6 +270,44 @@ func (gs *Server) Endorse(ctx context.Context, request *gp.EndorseRequest) (*gp.
...
+			if strings.HasSuffix(response.Response.Message, chaincode.ErrorStreamTerminated) {
+				// chaincode container crashed probably. Close connection and retry on another peer
```
denyeart, line 309: Did you check all possible errors and conclude that this is really the only one that should close the connection and remove the peer? I was hoping that all errors from peer could be handled consistently to eliminate the need to parse error messages, but maybe that is wishful thinking.

internal/pkg/gateway/api.go
```diff
@@ -273,6 +270,44 @@ func (gs *Server) Endorse(ctx context.Context, request *gp.EndorseRequest) (*gp.
 	return endorseResponse, nil
 }
 
+// responseStatus unpacks the proposal response and error values that are returned from ProcessProposal and
+// determines how the gateway should react (retry?, close connection?).
+// Uses the grpc canonical status error codes and their recommended actions.
+// Returns:
+// - response successful (bool)
+// - error message extracted from the err or generated from 500 proposal response
+// - should the gateway retry (only the Evaluate() uses this)
```
denyeart: Why does only Evaluate use this?

internal/pkg/gateway/api.go
```diff
@@ -71,25 +74,24 @@ func (gs *Server) Evaluate(ctx context.Context, request *gp.EvaluateRequest) (*g
...
+		success, message, retry, remove := gs.responseStatus(pr, err)
+		if success {
+			response = pr.Response
+		} else {
+			logger.Debugw("Evaluate call to endorser failed", "chaincode", chaincodeID, "channel", channel, "txid", request.TransactionId, "endorserAddress", endorser.endpointConfig.address, "endorserMspid", endorser.endpointConfig.mspid, "error", message)
+			errDetails = append(errDetails, errorDetail(endorser.endpointConfig, message))
+			if remove {
+				gs.registry.removeEndorser(endorser)
+			}
+			if retry {
```
denyeart, line 91: How many retries? I guess just a single retry?

internal/pkg/gateway/api.go
```diff
@@ -161,16 +163,14 @@ func (gs *Server) Endorse(ctx context.Context, request *gp.EndorseRequest) (*gp.
...
-				errDetails = append(errDetails, endpointErr)
+			success, message, _, remove := gs.responseStatus(firstResponse, err)
+
+			if !success {
+				logger.Debugw("Endorse call to endorser failed", "channel", request.ChannelId, "txid", request.TransactionId, "endorserAddress", firstEndorser.endpointConfig.address, "endorserMspid", firstEndorser.endpointConfig.mspid, "error", message)
+				errDetails = append(errDetails, errorDetail(firstEndorser.endpointConfig, message))
+				if remove {
+					gs.registry.removeEndorser(firstEndorser)
+				}
 				firstEndorser = plan.retry(firstEndorser)
```
denyeart: What is the meaning of this? Are we retrying on the first endorser again?

internal/pkg/gateway/api.go
andrew-coleman: Yes, I think I handle that in the err!=nil && response==nil case

internal/pkg/gateway/api.go
andrew-coleman: The former.  It closes the connection and removes from the registry, but if the next call to discovery returns that peer in its plan, then it will attempt to reconnect and add it back.  This could happen if the peer had gone down, but the gossip has not notified the discovery service yet.

internal/pkg/gateway/api.go
andrew-coleman: yes, these comments could be better. I've reworded them and repositioned them.

internal/pkg/gateway/api.go
andrew-coleman: Yes, as above.  Hopefully a bit clearer now

internal/pkg/gateway/api.go
andrew-coleman: Two situations where the peer connection gets closed and removed from cache:
1. If there's a grpc error but nil response from the peer - this assumes a connectivity problem with the peer (possibly peer crashed).
2. If the peer loses connection with the chaincode container - assumes the chaincode container has crashed or lost connectivity - I need to parse the error message to determine this.

internal/pkg/gateway/api.go
andrew-coleman: Endorse() will retry until all endorsement layouts have been exhausted.  It tries to get a successful endorsement from each org and minimise the changes of a rogue peer scuppering the transaction.  If an org is behaving badly, it can move on to a different layout.
Evaluate() only needs a single result from a local peer, which assumes hasn't gone rogue.

internal/pkg/gateway/api.go
andrew-coleman: It's in a loop.  It will retry until it's either successful, or runs out of peers to try.

internal/pkg/gateway/api.go
andrew-coleman: We're retrying to get the first endorsement (the one that generates the interest) from another peer in the same org

internal/pkg/gateway/api.go
```diff
@@ -71,25 +74,24 @@ func (gs *Server) Evaluate(ctx context.Context, request *gp.EvaluateRequest) (*g
...
+			logger.Debugw("Evaluate call to endorser failed", "chaincode", chaincodeID, "channel", channel, "txid", request.TransactionId, "endorserAddress", endorser.endpointConfig.address, "endorserMspid", endorser.endpointConfig.mspid, "error", message)
+			errDetails = append(errDetails, errorDetail(endorser.endpointConfig, message))
+			if remove {
+				gs.registry.removeEndorser(endorser)
+			}
+			if retry {
+				endorser = plan.retry(endorser)
+			} else {
+				return nil, newRpcError(codes.Aborted, "evaluate call to endorser returned error: "+message, errDetails...)
+			}
```
denyeart, line 95: Can't this and the next return be combined? Why does this one return message and errDetails, and the latter return only errDetails?
Regardless of whether we don't retry, or we've exhausted retries, shouldn't response be identical?

internal/pkg/gateway/api.go
denyeart: Ok, it is not obvious from reading the code, maybe change retry() function name to something like retryOnDifferentPeerInOrg().

internal/pkg/gateway/api.go
```diff
@@ -273,6 +270,43 @@ func (gs *Server) Endorse(ctx context.Context, request *gp.EndorseRequest) (*gp.
...
 
+// responseStatus unpacks the proposal response and error values that are returned from ProcessProposal and
+// determines how the gateway should react (retry?, close connection?).
+// Uses the grpc canonical status error codes and their recommended actions.
+// Returns:
+// - response successful (bool)
+// - error message extracted from the err or generated from 500 proposal response
+// - should the gateway retry (only the Evaluate() uses this)
+// - should the gateway close the connection and remove the peer from its registry
+func (gs *Server) responseStatus(response *peer.ProposalResponse, err error) (bool, string, bool, bool) {
```
denyeart:
```suggestion
func (gs *Server) responseStatus(response *peer.ProposalResponse, err error) (success bool, errorMessage string, retry bool, remove bool) {
```

These functions that return multiple bools can be difficult to keep track of. Adding variable names in the function declaration helps the reader keep track of things and tie them back to the comments above, even if they are not used in the body.

internal/pkg/gateway/api.go
denyeart: Ok, I would mention this in the removeEndorser() function comment.

internal/pkg/gateway/api.go
denyeart: Ok, I'd suggest summarize the retry logic in the Endorse() and Evaluate() function comments.

internal/pkg/gateway/api.go
```diff
@@ -273,6 +270,43 @@ func (gs *Server) Endorse(ctx context.Context, request *gp.EndorseRequest) (*gp.
 	return endorseResponse, nil
 }
 
+// responseStatus unpacks the proposal response and error values that are returned from ProcessProposal and
+// determines how the gateway should react (retry?, close connection?).
+// Uses the grpc canonical status error codes and their recommended actions.
+// Returns:
+// - response successful (bool)
```
denyeart, line 290: Good that you mention this is (bool). I'd recommend do the same for all four parameters so that reader can keep track of them.

internal/pkg/gateway/api.go
bestbeforetoday: Comments that duplicate the logic of the code are a well-known bad practice. They add clutter/duplication and tend to get out of date and become misleading as the code changes. Could we focus less on accurately describing the code in comments, and focus more on clearly expressing the logic in code?

internal/pkg/gateway/api.go
bestbeforetoday: I'm not a big fan of long lists of parameters or return values in general. I wonder if encapsulating the logic and/or resulting information in a separate struct (with nicely named methods, or at least fields) to communicate information back to consuming code might be clearer?

It's a shame that this logic is required to distinguish between different classes of errors at all though. 😞 

Although, given it's a private method only used by other methods on the same struct, I'm not sure this is anything to worry overly about.

internal/pkg/gateway/api.go
denyeart: @bestbeforetoday  Of course the priority should be on expressive logic in code.
I'm talking about a top level function comment on Endorse() and Evaluate(). Retry is such a fundamental and non-obvious aspect of this top-level public contract that it should at least be mentioned in the function comment that serves as the contract documentation.
See the example from the Go authors - https://go.dev/blog/godoc

bestbeforetoday: Endorse and Evaluate are gRPC services. Their documentation comments should be in the protobuf definitions, not in implementation code.

Documentation should describe the external behaviour the service (or function) provides, not the implementation details of how it provides that service.

denyeart: Retry is certainly something clients need to know that the Gateway services are doing, otherwise they will attempt their own retry strategies that overlap with what Gateway does. Agree that it would go best in the protobuf definitions.

denyeart (approved):

denyeart:
Thanks Andy!

PR #3054, opened 2021-11-17
gateway doc updates

joshhus:

- proposed edits to gateway overview topic
- adds to "peer" docs 

#### Type of change

- New feature
- Documentation update

#### Description

#### Additional details

The section at the following link, and its immediately preceding paragraph, have some duplication; I attempted to resolve the duplication but these are also related topics - a) specifying orgs due to transient data and b) specifying orgs generally (for other reasons) (https://github.com/hyperledger/fabric/blob/main/docs/source/gateway.md#targeting-specific-endorsement-peers)

#### Related issues

### DIFF
--- a/docs/source/gateway.md
+++ b/docs/source/gateway.md
@@ -1,12 +1,12 @@
 # Fabric Gateway
 
-Fabric Gateway is a service, introduced in Hyperledger Fabric v2.4 peers, that provides a simplified, minimal API for submitting transactions to a Fabric network. Requirements previously placed on the client SDKs, such as gathering transaction endorsements from peers of various organizations, are delegated to the Fabric Gateway service running within a peer to enable simplified application development and transaction submission in v2.4.
+Fabric Gateway (the gateway) is a service, introduced in Hyperledger Fabric v2.4, which provides a simplified, minimal API for submitting transactions to a Fabric network. Requirements previously placed on the client application, such as managing transaction endorsements, are delegated to a gateway peer to provide simplified application development and transaction submission in v2.4.
 
 ## Writing client applications
 
-Starting with Fabric v2.4, client applications should use one of the gateway SDKs (Go, Node, or Java), which are optimized to interact with the Fabric Gateway. These SDKs expose the same high-level programming model which was initially introduced in Fabric v1.4.
+Starting with Fabric v2.4, client applications should use a gateway SDK (Go, Node, or Java), which are optimized to interact with Fabric Gateway. The gateway SDKs expose the same high-level programming model which was initially introduced in Fabric v1.4.
 
-Fabric Gateway (aka *the gateway*) manages the following transaction steps:
+The gateway manages the following transaction steps:
 
 - **Evaluate** a transaction proposal. This will invoke a smart contract (chaincode) function on a single peer and return the result to the client. This is typically used to query the current state of the ledger without making any ledger updates.
 The gateway will preferably select a peer in the same organization as the gateway peer and choose the peer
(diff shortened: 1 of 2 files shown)


docs/source/gateway.md
```diff
@@ -38,39 +34,32 @@ Hyperledger Fabric currently supports client application development in three la
...
+### Endorsing private data
 
-### Targeting specific endorsement peers
+The gateway endorsement process is more restrictive for private data, passed in a transaction proposal as transient data, because it often contains sensitive or personal information that must never be shared with all organizations. In this case, the gateway will restrict the set of endorsing organizations to only members of the private data collection to be accessed (either read or write). If this restriction for private data would not satisfy the endorsement policy, the gateway returns an error to the client and does not submit the transaction to any organization for endorsement. In this case, the client application must explicitly [specify the organizations](#targeting-specific-endorsement-peers) to request endorsement from.
```
denyeart, line 58: Link needs to be updated as well.

```suggestion
The gateway endorsement process is more restrictive for private data, passed in a transaction proposal as transient data, because it often contains sensitive or personal information that must never be shared with all organizations. In this case, the gateway will restrict the set of endorsing organizations to only members of the private data collection to be accessed (either read or write). If this restriction for private data would not satisfy the endorsement policy, the gateway returns an error to the client and does not submit the transaction to any organization for endorsement. In this case, the client application must explicitly [specify the organizations](#specifying-endorsement-organizations) to request endorsement from.
```

docs/source/peers/peers.md
```diff
@@ -160,6 +167,22 @@ Later in this section, you'll learn more about the detailed nature of this
 ordering process --- and for a really detailed look at this process see the
 [Transaction Flow](../txflow.html) topic.
 
+### Using Fabric Gateway
+
+Hyperledger Fabric v2.4 introduces the optional Fabric Gateway to simplify client
+application development. Fabric Gateway (the gateway) migrates code for transaction
```
denyeart, line 173: I wouldn't use the word `migrates`, that usually implies migrating data. Maybe `delegates` is a better word.
And I wouldn't use the word `code` as that implies the user must provide code. The endorsement management logic is simply shifted from SDK-side to peer-side.

docs/source/peers/peers.md
```diff
@@ -160,6 +167,22 @@ Later in this section, you'll learn more about the detailed nature of this
...
+proposals and endorsements to the gateway peer, which then interacts with the channels
```
denyeart, line 174: No need to mention 'channels', since these requests are in the scope of a single channel.

docs/source/peers/peers.md
```diff
@@ -160,6 +167,22 @@ Later in this section, you'll learn more about the detailed nature of this
...
+and peers required by the transaction. The legacy model of transaction proposal and
+endorsement code residing in the client application remains supported. For those
```
denyeart, line 176: Again, I wouldn't use the word "code" here. The endorsement management logic was in the SDK that the client application depends on, not in the client application itself.

docs/source/peers/peers.md
```diff
@@ -341,6 +370,13 @@ application and a set of peers --- it does not involve orderers. Phase 1 is only
 concerned with an application asking different organizations' endorsing peers to
 agree to the results of the proposed chaincode invocation.
 
+**Attention**: Hyperledger Fabric v2.4 introduces Fabric Gateway, which
+simplifies client application development by migrating transaction proposal and
+endorsement code from the client application to the gateway peer. Refer to
+the [Fabric Gateway](../gateway.html) documentation on the gateway process
+and simplified SDKs. The legacy process, including coding transaction proposal and
+endorsement in the client application, remains supported, as discussed below.
+
```
denyeart, line 379: Instead of just adding the Attention note, I thought we would we would be updating the text throughout to assume that gateway manages the endorsements and transaction submission, rather than the client application managing the endorsements and transaction submission.  We can say that client applications using the legacy SDKs can alternatively manage the endorsements, but I think the main flow of the transaction process in these sections should be based on the gateway now.

The problem with just adding Attention notes is two-fold:
- We want new readers to think of the gateway flow as the default flow.
- We don't want to have to update these docs yet again in a year to make the changes to the main text. Let's just do it now and be done with it.

docs/source/gateway.md
joshhus: done.

docs/source/peers/peers.md
joshhus: ok

docs/source/peers/peers.md
joshhus: ok

docs/source/peers/peers.md
joshhus: ok

docs/source/peers/peers.md
joshhus: It's fair, it's substantially more rewriting and review. The issue is the old way is still supported, so I'll give it a go to show that both models work.

joshhus:
Closing, replaced by https://github.com/hyperledger/fabric/pull/3072

PR #3067, opened 2021-11-21
add http proxy support to core/chaincode/platforms

### LINKED ISSUE #3066: add http proxy support to core/chaincode/platforms
current main branch doesn't support http proxy in chaincode container image build phase (peer lifecycle chaincode package etc).

as the result, user needs to take below work-around at this moment.
choice A)  download  all dependencies before peer lifecycle chaincode package ( go mod vendor etc)
choice B)  build and use custom CORE_CHAINCODE_{{LANG}}_RUNTIME image.

in the above choices,
user needs to care the fabric version and use corresponding version of tool-chain(golang/nodejs etc) or fabric-{{lang}}env, even user just wants to invoke sample chaincode.

When fabric/core/chaincode/platforms cares http proxy in DockerBuildOptions function, like below,
work-around is no more needed.

```patch
# sample patch
diff --git a/core/chaincode/platforms/golang/platform.go b/core/chaincode/platforms/golang/platform.go
--- a/core/chaincode/platforms/golang/platform.go
+++ b/core/chaincode/platforms/golang/platform.go
@@ -228,7 +228,7 @@ echo Done!
 
 func (p *Platform) DockerBuildOptions(path string) (util.DockerBuildOptions, error) {
        env := []string{}
-       for _, key := range []string{"GOPROXY", "GOSUMDB"} {
+       for _, key := range []string{"GOPROXY", "GOSUMDB", "https_proxy", "http_proxy", "no_proxy"} {
                if val, ok := os.LookupEnv(key); ok {
                        env = append(env, fmt.Sprintf("%s=%s", key, val))
                        continue
```

itaru2622:
Signed-off-by: itaru2622 <itaru2622@gmail.com>

close #3066 

#### Type of change

- New feature

#### Description

add http proxy support to chaincode container build phase when fabric-peer container run with  "-e https_proxy=... " 
for details, refer #3066 

#### Additional details

when your devenv is located behind http proxy, additional no_proxy entries may be required to pass unit-test and integration-test.
the reasons is  described in #3066 and end of docs/sources/dev-setup/devenv.rst

### DIFF
--- a/core/chaincode/platforms/golang/platform.go
+++ b/core/chaincode/platforms/golang/platform.go
@@ -228,7 +228,7 @@ echo Done!
 
 func (p *Platform) DockerBuildOptions(path string) (util.DockerBuildOptions, error) {
 	env := []string{}
-	for _, key := range []string{"GOPROXY", "GOSUMDB"} {
+	for _, key := range []string{"GOPROXY", "GOSUMDB", "https_proxy", "http_proxy", "no_proxy", "HTTPS_PROXY", "HTTP_PROXY", "NO_PROXY"} {
 		if val, ok := os.LookupEnv(key); ok {
 			env = append(env, fmt.Sprintf("%s=%s", key, val))
 			continue
--- a/core/chaincode/platforms/golang/platform_test.go
+++ b/core/chaincode/platforms/golang/platform_test.go
@@ -20,7 +20,6 @@ import (
 	"testing"
 
 	pb "github.com/hyperledger/fabric-protos-go/peer"
-	"github.com/hyperledger/fabric/core/chaincode/platforms/util"
 	"github.com/hyperledger/fabric/core/config/configtest"
 	"github.com/pkg/errors"
 	"github.com/spf13/viper"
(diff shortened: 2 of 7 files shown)


denyeart:
@mbwhite @jkneubuh Could you take a look? Given Fabric is shifting to chaincode as an external service, is this something we would still want?

itaru2622:
@denyeart 
Cc: @mbwhite @jkneubuh 

Did you mean chaincode as an external service == PR 2990?

I checked PR 2990 and found that it has still core/chaincode/platforms and it doesn't  have proxy care. So I think my PR needs when user takes legacy way to run chaincode container.

mbwhite:
Hello @itaru2622 ... the preferred direction for chaincode is to run it in a 'external-as-a-service' mode. Where the chaincode is stood up independently of the peer. The peer has no role in bringing up chaincode, and will just connect to the given address when needed. 

Hence this would be our preferred route... 

I'm not 100% sure I follow your scenario sorry.

itaru2622:
@mbwhite @denyeart @jkneubuh 
I understood  your stance,  shifting chaincode as a service.

if you plant to deprecate (remove) core/chaincode/platforms in a couple of months, it may need not my patch.
otherwise,  it improves current user experience.

current code doesn't work even user sets proxy environment correctly to fabric-peer. it makes user confused.
my small patch fixes the above.

ale-linux:
Closing because of the long inactivity

PR #3072, opened 2021-11-22
1121 peers e2e and gateway

joshhus:
- Peers topic e3e edit for gateway, in detail, substantial.
- Minor updates to gateway overview since prior commit.
- Flow diagrams figure 6 and figure 10 are commented out for now - pending design update to reflect the associated text updates. 

#### Type of change

- New feature
- Documentation update

#### Description

Fabric Gateway service - new feature.

#### Related issues


joshhus:
Replaces https://github.com/hyperledger/fabric/pull/3054

docs/source/peers/peers.md
```diff
@@ -1,154 +1,183 @@
...
-more on this later). Smart contracts and ledgers are used to encapsulate the
-shared *processes* and shared *information* in a network, respectively. These
-aspects of a peer make them a good starting point to understand a Fabric network.
-
-Other elements of the blockchain network are of course important: ledgers and
-smart contracts, orderers, policies, channels, applications, organizations,
-identities, and membership, and you can read more about them in their own
-dedicated sections. This section focusses on peers, and their relationship to those
-other elements in a Fabric network.
+A fundamental element of a Hyderledger Fabric blockchain network is the set of
```
denyeart, line 3:
```suggestion
A fundamental element of a Hyperledger Fabric blockchain network is the set of
```

docs/source/peers/peers.md
```diff
@@ -1,154 +1,183 @@
...
+*peer nodes* (or, simply, *peers*). Peers are fundamental because they manage host
```
denyeart, line 4:
```suggestion
*peer nodes* (or, simply, *peers*). Peers are fundamental because they manage
```

docs/source/peers/peers.md
```diff
@@ -1,154 +1,183 @@
...
+ledgers and smart contracts. In Hyperledger Fabric v2.4, gateway peers also manage
```
denyeart, line 5: I'd suggest not introducing a term like "gateway peers".
We expect most users will simply keep the gateway service enabled on all peers, therefore it would be better to say something like "the gateway service running in peers also manages..."

docs/source/peers/peers.md
```diff
@@ -1,154 +1,183 @@
...
+Peers are a flexible and redundant element which can be created, started,
+stopped, reconfigured and deleted. Peers expose a set of APIs that enable administrators
+and client applications to interact with the services that peers provide. We'll
+learn more about these services in this section.
+
+## Fabric Gateway service
+
+Starting in Hyperledger Fabric v2.4, the [Fabric Gateway](../gateway.html) service
+is installed and enabled on each peer by default. This gateway service embeds  
+the transaction proposal and endorsement logic in the peer, as opposed to in client
```
denyeart, line 38: The transaction proposal and endorsement logic has always been embedded within each peer. What is different about v2.4 is that the gateway service can make proposals and gather endorsements from OTHER peers and submit transactions to the ordering service on behalf of a client application.

docs/source/peers/peers.md
```diff
@@ -1,154 +1,183 @@
...
+applications (in Fabric v2.3 and earlier). The v2.4 client SDKs (Go, Node, Java)
+incorporate this peer-centric model of transaction processing, enabling
+simplified application development. To continue running v2.3 or earlier client
+applications on a gateway peer, simply disable the Fabric Gateway service on that peer.
```
denyeart, line 42: You don't "run a client application on a gateway peer". I know what you are trying to say, but this wording may confuse readers.

Also, there is no need to disable the Fabric Gateway service. Client applications can gather endorsements per the prior model, regardless of whether gateway is enabled or not.

docs/source/peers/peers.md
```diff
@@ -1,154 +1,183 @@
...
+
+## Chaincode terminology
+
+Fabric implements **smart contracts** through a type of logic called **chaincode**
+--- code that accesses the ledger --- and is written using the Go, Java or Node SDK.
```
denyeart, line 47: Actually when we say "SDK" we are usually referring to the client application SDK.
For chaincode we don't use the term SDK, we usually just call them smart contract or chaincode APIs. See https://hyperledger-fabric.readthedocs.io/en/latest/sdk_chaincode.html for reference.
Note that both the client SDKs and smart contract APIs are available in Go, Java, and Node.

docs/source/peers/peers.md
```diff
@@ -1,154 +1,183 @@
...
 
 ## Ledgers and Chaincode
 
-Let's look at a peer in a little more detail. We can see that it's the peer that
-hosts both the ledger and chaincode. More accurately, the peer actually hosts
-*instances* of the ledger, and *instances* of chaincode. Note that this provides
-a deliberate redundancy in a Fabric network --- it avoids single points of
-failure. We'll learn more about the distributed and decentralized nature of a
-blockchain network later in this section.
+Now let's look at a gateway peer in a little more detail. We can see that it's the
```
denyeart, line 55:
```suggestion
Now let's look at a peer in a little more detail. We can see that it's the
```

docs/source/peers/peers.md
```diff
@@ -1,154 +1,183 @@
...
-that information might be out of date. In the diagram, you can see that ledger
-query is a simple three-step process.
-
-An update transaction starts in the same way as a query transaction, but has two
-extra steps. Although ledger-updating applications also connect to peers to
+ledger. Ledger-query interactions involve a simple dialogue between
+an application and a peer; ledger-update (write) interactions are more
+involved with additional steps.
+
+A client application connects to the v2.4 Fabric Gateway, through a gateway peer,
```
denyeart, line 124:
```suggestion
A client application connects to the gateway service running on a peer
```

docs/source/peers/peers.md
```diff
@@ -1,154 +1,183 @@
...
-query is a simple three-step process.
-
-An update transaction starts in the same way as a query transaction, but has two
-extra steps. Although ledger-updating applications also connect to peers to
+ledger. Ledger-query interactions involve a simple dialogue between
+an application and a peer; ledger-update (write) interactions are more
+involved with additional steps.
+
+A client application connects to the v2.4 Fabric Gateway, through a gateway peer,  
+in order to access ledgers and chaincodes. The v2.4 Fabric Software Development Kits
```
denyeart, line 125: We are actually calling them Gateway SDKs and starting their numbering at v1.0.0.

docs/source/peers/peers.md
```diff
@@ -1,154 +1,183 @@
...
+(SDKs) make this easy for programmers --- the APIs enable applications to (through
+the gateway) submit transaction proposals, invoke chaincodes, request endorsement,
```
denyeart, line 127: "submit transaction proposals" and "invoke chaincodes" are the same thing.  A transaction proposal specifies which chaincode to invoke. As written it sounds like two different things.

docs/source/peers/peers.md
```diff
@@ -1,154 +1,183 @@
...
+receive events, and forward endorsed transactions to the ordering service for
+validation and commitment to the distributed ledger. (For more detail on the gateway,
+refer to the [Fabric Gateway](../gateway.html) overview on v2.4 transaction management
```
denyeart, line 130: We should avoid specifying "v2.4" like this, as that will be confusing when future versions are released and require us to re-visit the docs.

docs/source/peers/peers.md
```diff
@@ -1,154 +1,183 @@
...
+
+<!--
+
+Figure 6. Applications and Peers
+
+![Peer6](./peers.diagram.6.png) -->
+
+Peers, in conjunction with orderers, ensure that the ledger is kept consistent
+and up-to-date on every peer in a channel. The following sequence describes how
+the interactions between a client application, a gateway peer and orderer nodes
```
denyeart, line 147:
```suggestion
the interactions between a client application, the gateway service running on a peer, and other peers and orderer nodes
```

docs/source/peers/peers.md
```diff
@@ -1,154 +1,183 @@
...
+produce this result:
+
+1. **Proposal submission** --- The client application (A1) submits a transaction proposal by connecting to gateway peer P1 and invoking chaincode (S1) to query or update the ledger (L1). For ledger updates (writes), A1 must either delegate endorsement to the gateway service or explicitly identify the organizations required for transaction endorsement.
```
denyeart, line 150:
```suggestion
1. **Proposal submission** --- The client application (A1) submits a transaction proposal by connecting to the gateway service on peer P1 and invoking chaincode (S1) to query or update the ledger (L1). For ledger updates (writes), A1 must either delegate the selection of endorsers to the gateway service or explicitly identify the organizations required for transaction endorsement.
```

docs/source/peers/peers.md
```diff
@@ -1,154 +1,183 @@
...
+2. **Transaction validation** --- P1 runs chaincode S1 to validate the transaction proposal and returns its response to the client with the proposed ledger update or query result. For a transaction proposal that is not validated successfully by P1, the transaction workflow ends here. For validated transactions (and queries), the process continues.
```
denyeart, line 151: We reserve the term "transaction validation" for the post-ordered validation/commit phase (your step 9 here).
We call this step "transaction execution" or "transaction simulation" (since the transaction isn't really executed yet, we're simply executing the chaincode to get the proposed writes).
You could call it execution failure or simulation failure at this point, rather than invalidation.
For successful executions/simulations, the results are not yet returned to client. The gateway service first gathers the other endorsements as seen in your step 3.

docs/source/peers/peers.md
```diff
@@ -1,154 +1,183 @@
...
+3. **Endorsement delegation** --- If A1 has delegated transaction endorsement for a ledger update, the gateway service on P1: a) requests endorsement from a peer in its own organization (selecting the available peer with the highest block height, which can be P1); b) reads the set of endorsement policies for the updated keys; and c) forwards the proposal to endorsement peers in the organizations required to endorse this transaction.
```
denyeart, line 152: 3a execution and endorsement on your own org peer was already done in step 2.

docs/source/peers/peers.md
```diff
@@ -1,154 +1,183 @@
...
+4. **Endorsement collection** --- P1 collects the transaction endorsement (or query) results from each organization and returns the proposal response to the client application A1. For ledger queries, the process is now complete. For ledger updates, if the proposal is endorsed, P1 forwards the transaction to the channel ordering service.
```
denyeart, line 153: Actually for ledger queries, the gateway service only invokes the chaincode on a peer from your org. It doesn't gather additional endorsements.

docs/source/peers/peers.md
```diff
@@ -1,154 +1,183 @@
...
+5. **Endorsement retries** --- If the transaction proposal is *not* endorsed, the gateway service will attempt retries on peers that would satisfy endorsement for a ledger update (unless private data has been submitted). If the proposal is *not* endorsed due to detection of transient data --- which often contains private data --- P1 returns a request to the client (A1) for explicit specification of the organizations that are members of the private data collection(s). Once A1 is updated and resubmitted in consideration of private data, the gateway repeats the process by requesting endorsement from the specified organizations and collecting the results.
```
denyeart, line 154: The retry and private data nuances may be too much information for this peer Key Concepts topic. We don't want to confused newbies with those details. Those are advanced scenarios and the information is available in the dedicated Gateway and Private Data topics. Perhaps we could just mention to see the Gateway doc for more details about retries and private data.

docs/source/peers/peers.md
```diff
@@ -1,154 +1,183 @@
...
+6. **Endorsement failure** --- If the returned endorsements do not satisfy the endorsement policy requirements, P1 returns a failure message to the client and P1 updates its ledger with *only* a record of the failed transaction proposal. For a failed endorsement, no other ledger or state database update occurs on any peer, and the transaction workflow is now complete.
```
denyeart, line 155: A peer never updates its ledger in the proposal/endorsement phase.
Failures will get returned to client and the transaction will never get into a block.

docs/source/peers/peers.md
```diff
@@ -1,154 +1,183 @@
...
+7. **Endorsement success** --- If the returned endorsements satisfy the endorsement policy requirements, P1 submits the transaction to the ordering service. Once the ordering service validates the transaction, P1 returns a transaction envelope with the endorsements to the client for signing.
```
denyeart, line 156: I have to defer to @andrew-coleman  on how signing works.
But in the scenarios where a client must sign, this occurs prior to submission to ordering service.

docs/source/peers/peers.md
```diff
@@ -1,154 +1,183 @@
...
+8. **Transaction signing** --- The client (A1) returns the signed transaction to P1, which forwards the transaction to O1 for commitment to the ledger (along with other transactions which are first ordered and packaged together in blocks for the ledger update.)
+9. **Ledger commitment** --- O1 collects transactions from across the channel and packages them into ordered blocks and distributes the blocks to all peers, including P1. P1 (and all other channel peers) validates the transaction before committing it to the ledger (L1). The commit is an immutable update written to the channel ledger, and the world state (the effect of the transaction, such as a change in asset ownership) is also updated on each peer.
+10. **Commit event** --- A1 receives a commit status event from P1 with proof of the ledger commit.
+
+### Querying the Ledger
```
denyeart, line 161: "Querying the Ledger" and "Updating the Ledger" sections repeat what was included in steps 1 through 10 above. Can we converge these?
Actually, I would expect the step-by-step details to come later in the "Peers and Orderers" section (Phase 1, 2, 3). Could we avoid repeating it?

docs/source/peers/peers.md
```diff
@@ -1,154 +1,183 @@
...
+
+A peer can promptly return the results of a ledger query to the client application,
+because the information required to respond to the query is (typically) in the gateway
+peer's ledger. By default, the gateway service does not communicate with other peers
+to respond to a query. An application can specify for the gateway service to issue
+the query on one or more additional peers, to corroborate the initial result or verify
```
denyeart, line 167: The gateway service will only query from a peer in its own org.

docs/source/peers/peers.md
```diff
@@ -160,34 +189,51 @@ Later in this section, you'll learn more about the detailed nature of this
 ordering process --- and for a really detailed look at this process see the
 [Transaction Flow](../txflow.html) topic.
 
+### Fabric Gateway
```
denyeart, line 192: This is a repeat of section "Fabric Gateway service" above.

docs/source/peers/peers.md
```diff
@@ -199,53 +245,53 @@ form a blockchain network.
...
+nodes are not shown in the diagram.*
+
+It's important to notice what's happens during the formation of a Fabric
 blockchain network. *The network is both formed and managed by the multiple
-organizations who contribute resources to it.* Peers are the resources that
-we're discussing in this topic, but the resources an organization provides are
-more than just peers. There's a principle at work here --- the network literally
+organizations that contribute resources to it.* Peers are the resources that
+we're discussing in this topic, but organizations also provide other resources,
+such as chaincode and consensus algorithms. There's a principle at work here --- the network literally
```
denyeart, line 269: I would remove "consensus algorithms" here. You could replace with "ordering service nodes" if that is what you mean.

docs/source/peers/peers.md
```diff
@@ -199,53 +245,53 @@ form a blockchain network.
...
+the same, because each organization can decide how it wants to process the ledger.
+Both application and presentation logic can therefore vary across organizations,
+even though all peers host an equivalent copy of the ledger.
+
+Applications can connect to peers in their organization, or to peers in other
+organizations, depending on the nature of the ledger interaction that's required.
+For ledger-query interactions, applications may connect only to their own
+organization's peers. For ledger-updates, however, a gateway peer connects the
+application to peers in *every* organization required to endorse the ledger update
+(write) transaction.
```
denyeart, line 294: Using the gateway service pattern, client applications would not directly connect to peers in other organizations. The client application would connect to a peer in their organization (or a peer in another trusted organization), and that peer's gateway service would connect to other peers as needed.

docs/source/peers/peers.md
```diff
@@ -312,40 +360,67 @@ peer cannot, on its own, update the ledger --- updating requires the consent of
...
-  as transactions and packaged into blocks.
-* In the third and final phase, these blocks are distributed back to every peer where
-  each transaction is validated before being committed to that peer's copy of the ledger.
-
-As you will see, orderer nodes are central to this process, so let's
-investigate in a little more detail how applications and peers use orderers to
-generate ledger updates that can be consistently applied to a distributed,
-replicated ledger.
+Specifically, client applications that request a ledger update are involved in a
+3-phase process which, through the Fabric Gateway service in v2.4, ensures that all
```
denyeart, line 369: Again, reference to v2.4 will not work in future releases. You could say "starting in v2.4".

docs/source/peers/peers.md
```diff
@@ -312,40 +360,67 @@ peer cannot, on its own, update the ledger --- updating requires the consent of
...
-As you will see, orderer nodes are central to this process, so let's
-investigate in a little more detail how applications and peers use orderers to
-generate ledger updates that can be consistently applied to a distributed,
-replicated ledger.
+Specifically, client applications that request a ledger update are involved in a
+3-phase process which, through the Fabric Gateway service in v2.4, ensures that all  
+peers in the blockchain network keep their ledgers consistent with each other:
+
+1. In the first phase, the client application sends a transaction proposal to a
+gateway peer, preferably in its own organization. Assuming the application has
```
denyeart, line 373: Instead of "preferably", I think we can be more prescriptive here... a peer from its own organization or another trusted organization.

docs/source/peers/peers.md
```diff
@@ -312,40 +360,67 @@ peer cannot, on its own, update the ledger --- updating requires the consent of
...
+delegated transaction endorsement to the Fabric Gateway service, the gateway peer
+validates (runs) the transaction and forwards it to endorsing peers in the organizations
```
denyeart, line 375: Again, reserve the term "validates" for Phase 3 validation/commit phase.
The terms "executes", "runs", and "simulates" are all used elsewhere throughout the docs.

docs/source/peers/peers.md
```diff
@@ -312,40 +360,67 @@ peer cannot, on its own, update the ledger --- updating requires the consent of
...
+to the gateway peer.
+3. In the third and final phase, the signed transaction (blocks) are distributed
+to the endorsing peers, which validate the transaction.
+
+In addition to the client application and gateway peers, orderer nodes play a
+critical role in transactions that update the ledger. Before committing transactions
+to the ledger, orderer nodes do just that --- they order all submitted transactions
+to ensure a consistent and accurate sequence of transactions on the shared ledger.
+Note that blockchain manages inconsistent ledgers through a process called forking,
+and the subsequent pruning (discarding and archiving) of the invalidated forks.
```
denyeart, line 390: I wouldn't mention this. This is true in bitcoin, but not in Fabric. Since peers deterministically validate or invalidate transactions, there are no forks possible (unless there is a bug or a tampered peer).

docs/source/peers/peers.md
```diff
@@ -312,40 +360,67 @@ peer cannot, on its own, update the ledger --- updating requires the consent of
...
-ledger, but rather simply signs it and returns it to the application. Once the
+Phase 1 of the transaction workflow involves an interaction between a client
+application and a set of peers --- orderers are not involved yet. In Phase 1,
+the client application initiates a request to endorsing peers, through the
+Fabric Gateway service, to return their responses to the proposed transaction.
+
+To begin phase 1, the client application generates a transaction proposal and
+sends it to a gateway peer in its own organization. The gateway peer with
+the highest block height is selected; if none are discovered, the gateway
+service moves on to select a gateway peer in another organization.
```
denyeart, line 402: It does not "move on to select a gateway peer in another organization".

docs/source/peers/peers.md
```diff
@@ -312,40 +360,67 @@ peer cannot, on its own, update the ledger --- updating requires the consent of
...
+
+Once selected, the gateway peer validates the transaction by invoking the
```
denyeart, line 404:
```suggestion
Once selected, the gateway peer executes the transaction by invoking the
```

docs/source/peers/peers.md
```diff
@@ -312,40 +360,67 @@ peer cannot, on its own, update the ledger --- updating requires the consent of
...
+chaincode without a ledger update --- this can be referred to as simulating
+the transaction, because it runs the application without any effect on the ledger
+or state on any peer. The gateway peer then returns it transaction result to the
+client, and if the application delegates endorsement to the gateway service,
+it also forwards the proposal to the *endorsing peers*.
```
denyeart, line 409: It does not return to client here... it forwards the proposal to other endorsing peers first.

docs/source/peers/peers.md
```diff
@@ -312,40 +360,67 @@ peer cannot, on its own, update the ledger --- updating requires the consent of
...
+
+Each endorsing peer then follows suit, executing the chaincode called by the
+client application to generate a transaction proposal response and return it
+(signed) to the gateway peer. The gateway peer collects all responses, and if
+the responses collectively satisfy the endorsement policy (or policies), it
+forwards the transaction to the ordering service for sequencing. If the
+endorsement policies are not satisfied, or if the transaction would write to
+a private data collection, the gateway peer returns a failed proposal message
+to the client. The client then must explicitly specify the organizations
+required for endorsement and resubmit the updated transaction proposal.
```
denyeart, line 419: The client would not delegate to gateway service, and then upon failure, specify endorsing organizations.
These are two distinct application patterns, an application will EITHER delegate to gateway service OR specify endorsing organizations. It won't do some hybrid based on dynamic error handling.

docs/source/peers/peers.md
```diff
@@ -369,61 +447,79 @@ can be accepted by the network. This is literally what it means to achieve
...
+
+target section or discussion not found -->
+
+An individual peer cannot detect whether its transaction result is
+non-deterministic --- transaction responses must be collected by the gateway
+service for comparison before it can detect non-determinism in chaincode.
+In response to inconsistent transaction responses, caused either by
+nondeterministic chaincode or any other issue, the gateway service terminates
+the transaction workflow and returns a failed proposal message to the client. The
+client can then resubmit the transaction proposal or an amended transaction proposal.
```
denyeart, line 488: I realize this content existed previously, but it becomes more confusing with gateway.
I would remove all this discussion of different endorsements and non-determinism. It is too much information and too nuanced for Key Concepts. It gets in the way more than helps, especially now that gateway is in the middle.

docs/source/peers/peers.md
```diff
@@ -369,61 +447,79 @@ can be accepted by the network. This is literally what it means to achieve
...
-subsequent validation of blocks from the orderer to the peers, where they can be
+The final phase of the transaction workflow involves the distribution (and
+subsequent validation) of blocks from the orderer to the peers, where they are
 committed to the ledger. Specifically, at each peer, every transaction within a
 block is validated to ensure that it has been consistently endorsed by all
-relevant organizations before it is committed to the ledger. Failed transactions
-are retained for audit, but are not committed to the ledger.
+relevant organizations before being committed to the ledger. A different
+validation result across on any peers means endorsements by one or more of  
+those peers are not equivalent. Failed transactions are retained for audit,
```
denyeart, line 514: I would remove this.
The peer validation phase does check that endorsements match, and you can mention this. But note that gateway will not even submit the transaction if they don't match, so mentioning this will confuse people.
Also "different validation result" is not correct here... you mean "different execution result".
Finally, note that there are other potential reasons for transaction invalidation besides executions/endorsements not matching.

docs/source/peers/peers.md
joshhus: So there is no record of a validation failure or an endorsement failure, anywhere? I expected that if a transaction fails, even during proposal phase, there would still be some record of that failure. e.g. Repeated failures could indicate nefarious attempts to access the ledger, making the record valuable.

docs/source/peers/peers.md
joshhus: Yes I agree that the signing comes earlier than stated here, will update.

docs/source/peers/peers.md
joshhus: I find it valuable to have an e2e in one place, rather than splitting it up into phases and details in each phase, in different doc sections. This is arguable however. ... Some redundancy can be a good thing for readers, will re-visit.

docs/source/peers/peers.md
joshhus: okay, so it can query additional peers to confirm the initial query result, but do so only on additional peers in its own org. will clarify.

docs/source/peers/peers.md
joshhus: A judgment call on the value of redundancy; a reader can't be required or presumed to have read everything on the page above it. Sections can be linked to directly, e.g.

docs/source/peers/peers.md
joshhus: yes.

docs/source/peers/peers.md
joshhus: yes, will clarify as done elsewhere.

docs/source/peers/peers.md
joshhus: yes.

docs/source/peers/peers.md
joshhus: ok.

docs/source/peers/peers.md
joshhus: yes.

docs/source/peers/peers.md
joshhus: yes, the current docs seem to equate blockchain with fabric blockchain, I have added "Fabric" blockchain elsewhere.

docs/source/peers/peers.md
joshhus: so what happens then if none are discovered. or is that not a viable scenario.

docs/source/peers/peers.md
joshhus: ok.

docs/source/peers/peers.md
joshhus: right the client application would have to specify the endorsing orgs and resubmit, it says the client must do this. are you saying the gateway service would have to be disabled on the peer first in order to resubmit specified orgs.?

docs/source/peers/peers.md
joshhus: okay will comment out for a future move to another location.

docs/source/peers/peers.md
joshhus: yes so there is a submit by client during proposal phase to the service and then another submit by the service apparently. will look to differentiate. 
will diff the execution/simulation/run vs. validation. they do have to run it to validate it though afaik.

joshhus: Seems there's a submit by the app to the gateway service, of a proposal, and then a submit of by the service to the orderers, for an endorsed transaction. Hence good to distinguish the initial submit by calling it a send instead of submit, if that works.

docs/source/peers/peers.md
joshhus: yes

docs/source/peers/peers.md
joshhus: yes

joshhus: original was host but ok.

docs/source/peers/peers.md
joshhus: ok

docs/source/peers/peers.md
joshhus: Well the point here is that something about proposal and endorsement has moved from the client app to the gateway service. I'll try to get it right.

docs/source/peers/peers.md
joshhus: ok

docs/source/peers/peers.md
joshhus: yes ok.

docs/source/peers/peers.md
joshhus: ok

joshhus: changed all

docs/source/peers/peers.md
joshhus: ok.

docs/source/peers/peers.md
joshhus: ok

docs/source/peers/peers.md
joshhus: ok. I always found chaincodes plural a bit weird sounding as opposed to invoke chaincode, singular. leaving as is for now.

docs/source/peers/peers.md
joshhus: ok.

docs/source/peers/peers.md
joshhus: ok.

docs/source/peers/peers.md
denyeart: No record in ledger for proposal phase failures. The failures are noted in peer log and peer metrics however, so these can be used by ops to detect repeat failures.

docs/source/peers/peers.md
joshhus: ok

docs/source/peers/peers.md
joshhus: ok I think you're saying to move 3a text to step 2.

docs/source/peers/peers.md
denyeart: For queries, only a single peer will execute the query and return response to client. If client wants to get query results from peers of various organizations, they would have to specify the target organization in N individual queries.

docs/source/peers/peers.md
denyeart: I'm saying the client would never delegate to gateway service and then upon failure resubmit to specific endorsing organizations. The choice of whether to delegate to gateway service or target specific endorsing organizations would be baked into the application based on application requirements. For example if you have private data or only trust a subset of organizations, then you would target those organizations. If you don't have these types of requirements, then you would delegate the choice to gateway service.

docs/source/peers/peers.md
denyeart: The gateway peer has ability to execute/endorse on the local peer itself (assuming the chaincode is installed locally), and therefore the local peer is always implicitly "discovered".

docs/source/peers/peers.md
joshhus: ok.

docs/source/peers/peers.md
joshhus: okay will comment out this step 5.

docs/source/peers/peers.md
joshhus: no record exists, per discussion

joshhus: no ledger update, ok.

docs/source/peers/peers.md
joshhus: returning to the 3 phases.

docs/source/peers/peers.md
joshhus: idk. @andrew-coleman When does the client have to sign something that is returned to it by the gateway service?

docs/source/peers/peers.md
joshhus: Maybe should keep the legacy 3 phase sections and just add the clearer steps to it for gateway.

docs/source/peers/peers.md
joshhus: This section also includes the diagram / flow. It's seeming like steps 1-10 should just be moved to the existing other sections as applicable.

docs/source/peers/peers.md
joshhus: deleting "if none are discovered ..."

docs/source/peers/peers.md
joshhus: no disabling of gateway. only v2.3 or v2.4 SDKs

docs/source/peers/peers.md
joshhus: ok

docs/source/peers/peers.md
joshhus: Andy uses submit so to be consistent will use submit ... the submit "to" should clarify it enough.

docs/source/peers/peers.md
joshhus: yes ok.

docs/source/peers/peers.md
joshhus: resolved via slack - client signs on initial submission and then following endorsement.

joshhus:
iterating again so will likely close this PR in favor of a new one. it's a process that works for me in GitHub. Trying to amend gives me out of sync messages etc., despite attempts at syncing.

joshhus:
Closing, replacement PR is https://github.com/hyperledger/fabric/pull/3077

PR #3077, opened 2021-11-24
peers doc iteration

joshhus:

#### Type of change

- New feature
- Documentation update

#### Description

Another iteration on the Peers topic for new Fabric Gateway service.

#### Related issues

### DIFF
--- a/docs/source/peers/peers.md
+++ b/docs/source/peers/peers.md
@@ -319,7 +319,7 @@ peers and orderers manage the consensus process in this section.
 
 Specifically, applications that want to update the ledger are involved in a
 3-phase process, which ensures that all the peers in a blockchain network keep
-their ledgers consistent with each other. 
+their ledgers consistent with each other.
 
 * In the first phase, applications work with a subset of *endorsing peers*, each of
   which provide an endorsement of the proposed ledger update to the application,


joshhus:
So the "issue" with this page/topic is that it was originally written for Peers, and then goes through the relationship in detail of a Peer to EACH other element in a Fabric channel. Hence, the sense of duplication across this very long page/topic regardless of the gateway service adds. Especially in the later Orderers and Peers section, which is very long. I cleaned up that section e2e, BUT also left the Phase 1, 2, 3 discussions in there. ... With Phase 1, 2, 3 discussions ALSO in the previous Applications and Peers section, there remains some duplication - albeit from different perspectives (Apps vs. Orderers).

So rather than just delete legacy text - Orderers and Peers does have some interesting comments in there about security for example - I suggest we leave what I left in for Monday without further re-strategizing the entire page/topic. (But I have deleted substantial parts, such as those on determinism, per the review.)

In summary, the whole Peers page has gotten my e2e today and I think it's pretty good for Monday. Notable exception being the missing graphics for Figures 6 and 10, which are commented out for now. I'll read it again Friday in the formatted view to be sure of all details, and can do another PR per you reviews as necessary.

denyeart:
@joshhus No changes are reflected in this PR!

joshhus:
Replaced by https://github.com/hyperledger/fabric/pull/3079

PR #3085, opened 2021-11-29
Add read version to the example in read-write set

### LINKED ISSUE #3015: Confusing examples on Read-Write set semantics
[docs/source/readwrite.rst](https://github.com/hyperledger/fabric/blob/main/docs/source/readwrite.rst) says each read operation has a key and a version. However, the read operations in the last World state example have no version.

Do you like me adding read version to the example? It would be like this:

> World state: (k1,1,v1), (k2,1,v2), (k3,1,v3), (k4,1,v4), (k5,1,v5)
> T1 -> Write(k1, v1'), Write(k2, v2')
> T2 -> Read(k1, v1), Write(k3, v3')
> T3 -> Write(k2, v2'')
> T4 -> Write(k2, v2'''), read(k2, v2)
> T5 -> Write(k6, v6'), read(k5, v4)

T5 will fail because it reads k5 of version 4 but in world state k5 is of version 5, meaning someone has updated k5 before T5 got committed (or validated?).

gqqnbig:
#### Type of change

- Documentation update

#### Description

Per discussion, we want to add read version to the example in read-write set so that

1. the read version described in the earlier description appears in this example.
2. show readers clearly that a transaction is rejected due to read version mismatch.

#### Related issues

Fixes #3015

### DIFF
--- a/docs/source/readwrite.rst
+++ b/docs/source/readwrite.rst
@@ -121,10 +121,10 @@ each of these transactions.
 
     World state: (k1,1,v1), (k2,1,v2), (k3,1,v3), (k4,1,v4), (k5,1,v5)
     T1 -> Write(k1, v1'), Write(k2, v2')
-    T2 -> Read(k1), Write(k3, v3')
+    T2 -> Read(k1,1), Write(k3, v3')
     T3 -> Write(k2, v2'')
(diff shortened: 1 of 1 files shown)


manish-sethi (approved):

manish-sethi:
Thanks @gqqnbig !

PR #3080, opened 2021-11-27
Other key concepts

joshhus:
#### Type of change

- New feature
- Documentation update

#### Description

Other Key Concepts - Network and Ordering Service topics for gateway

#### Related issues

https://github.com/hyperledger/fabric/issues/2807 
- Other Key Concepts items

PR #3086, opened 2021-11-29
latest PR review comments

joshhus:

#### Type of change

- New feature
- Documentation update

#### Description

Gateway updates to Ordering service topic and Network topic

#### Related issues

This replaces https://github.com/hyperledger/fabric/pull/3080


docs/source/network/network.md
```diff
@@ -84,13 +84,16 @@ While it's now technically possible to drive transactions using the peer CLI, th
...
 
-Once a chaincode has been installed on a peer node and defined on a channel it can be [invoked](../glossary.html#invoke) by a client application. Client applications do this by sending transaction proposals to peers owned by the organizations specified by the endorsement policy. The transaction proposal serves as input to the chaincode, which uses it to generate an endorsed transaction response, which is returned by the peer node to the client application.
+Once a chaincode has been installed on a peer node and defined on a channel it can be [invoked](../glossary.html#invoke) by a client application. Client applications do this by sending transaction proposals to the gateway. The target peer --- a peer in the same organization submitting the proposal --- then runs the transaction and forwards the proposal to peers in the organizations required by the endorsement policy. The transaction proposal serves as input to the chaincode, which uses it to generate an endorsed transaction response, which is returned by the target peer to the client application.
```
andrew-coleman, line 94: The chaincode is not invoked directly by the client - the client instructs the gateway to endorse the proposal.  The gateway does this by selecting peers that have the chaincode installed from the organizations required to satisfy the endorsement policy.  The gateway collects all the proposal responses and builds an endorsed transaction message which can be signed and sent to the ordering service.

docs/source/network/network.md
```diff
@@ -84,13 +84,16 @@ While it's now technically possible to drive transactions using the peer CLI, th
...
+
+Note: Instead of delegating transaction endorsement to the gateway, the client application can specify the endorsing organizations and collect the endorsements --- refer to the [v2.3 Applications and Peers](https://hyperledger-fabric.readthedocs.io/en/release-2.3/peers/peers.html#applications-and-peers) topic for details.
```
andrew-coleman, line 96: They *can*, but if they did, they would have to use these older API for everything.  They can't mix and match.  I.e. we wouldn't support collecting a few proposal responses directly from peers, than then use the gateway for sending to the orderer, for instance.

docs/source/orderer/ordering_service.md
```diff
@@ -74,41 +74,45 @@ process with three phases that ensures all of the peers in a blockchain network
 keep their ledgers consistent with each other.
 
 In the first phase, a client application sends a transaction proposal to
-a subset of peers that will invoke a smart contract to produce a proposed
-ledger update and then endorse the results. The endorsing peers do not apply
-the proposed update to their copy of the ledger at this time. Instead, the
-endorsing peers return a proposal response to the client application. The
+the Fabric Gateway service. The target peer --- a peer in the same organization
+submitting the proposal --- then runs the transaction. The gateway also forwards
```
andrew-coleman, line 78: Just to clarify, the target peer is the one that the client has connected to in order to use the gateway service.  The first endorsement is done by a peer in the target peer's organization.  It could be the target peer itself, but not necessarily (e.g. the target peer might not have the chaincode installed on it).  Also the client might not necessarily be a member of this same organization (e.g. the client's org might not run any peers), but the client *trusts* the target peer's org to handle the transaction.

andrew-coleman: Also, we've lost an important sentence in this change:

"The endorsing peers do not apply the proposed update to their copy of the ledger at this time."

In this phase, the peers are saying what they *would* do with this transaction proposal, given the current state of their individual ledgers.  Only in phase 3, *if and only if all the endorsing peers agree* (i.e. all endorsements match), do the actual changes get applied.

docs/source/orderer/ordering_service.md
```diff
@@ -135,36 +141,36 @@ packaged into multiple different blocks that compete to form a chain.
...
 orderers most definitely do not. Every authorized transaction that arrives at an
-orderer is mechanically packaged in a block --- the orderer makes no judgement
+orderer is then mechanically packaged into a block --- the orderer makes no judgement
 as to the content of a transaction (except for channel configuration transactions,
 as mentioned earlier).
 
 At the end of phase two, we see that orderers have been responsible for the simple
 but vital processes of collecting proposed transaction updates, ordering them,
-and packaging them into blocks, ready for distribution.
+and packaging them into blocks, ready for distribution to the gateway service.
```
andrew-coleman, line 155:
```suggestion
and packaging them into blocks, ready for distribution.
```
No, blocks get distributed to committing peers.  The gateway plays no part in this.

docs/source/network/network.md
joshhus: reviewed and updated on 12-1-21 zoom.

docs/source/network/network.md
joshhus: Note was removed in between iterations.

docs/source/orderer/ordering_service.md
joshhus: yes ok

docs/source/orderer/ordering_service.md
joshhus: ok.

docs/source/orderer/ordering_service.md
joshhus: ok

joshhus:
replaced by  https://github.com/hyperledger/fabric/pull/3090

PR #3087, opened 2021-11-29
Add command reference doc for ledgerutil

denyeart:
Add command reference doc for ledgerutil.
Includes minor edits to the command's user messages.


docs/source/commands/ledgerutil.md
```diff
@@ -0,0 +1,64 @@
...
+
+The `ledgerutil compare` command allows administrators to compare channel snapshots from two different peers.
+Although channel snapshots from the same height should be identical across peers, if the snapshot
+files indicate different state hashes (as seen in `_snapshot_signable_metadata.json` file from each snapshot)
+then it indicates that at least one of the peers state databases is not correct relative to the blockchain.
+The `ledgerutil compare` utility will output a set of JSON files if the snapshots are not identical.
+One output JSON file will include any key/value differences sorted by key, another JSON file will include
+any key/value differences sorted by block and transaction height so that you can identify the height
+where a divergence may have occurred.
+
```
manish-sethi, line 21: I guess adding about potential use of the utility may be helpful  -- e..g., in the case of difference between two snapshots, users may like to figure out which snapshot is diverged from the expected, perhaps either based on application knowledge or comparing with more snapshots from other peers. The logs near the time of initial divergence may help figure out what may have gone wrong.

docs/source/commands/ledgerutil.md
```diff
@@ -0,0 +1,64 @@
...
+ --->
+
+# ledgerutil
+
+The `ledgerutil compare` command allows administrators to compare channel snapshots from two different peers.
+Although channel snapshots from the same height should be identical across peers, if the snapshot
+files indicate different state hashes (as seen in `_snapshot_signable_metadata.json` file from each snapshot)
+then it indicates that at least one of the peers state databases is not correct relative to the blockchain.
+The `ledgerutil compare` utility will output a set of JSON files if the snapshots are not identical.
+One output JSON file will include any key/value differences sorted by key, another JSON file will include
```
jcastrence:
```suggestion
Two output JSON files will include any key/value differences sorted by key (one for public key/value differences and one for private key/value differences), another JSON file will include any key/value differences (public or private) sorted by block and transaction height...
```

docs/wrappers/ledgerutil_postscript.md
```diff
@@ -0,0 +1,20 @@
...
+  * Compare snapshots from two different peers for mychannel at snapshot height 5.
+
+    ```
+    ledgerutil compare -o ./compare_output -f 10 ./peer0.org1.example.com/snapshots/completed/mychannel/5 ./peer1.org1.example.com/snapshots/completed/mychannel/5
+
+    Both snapshot public state and private state hashes were the same. No results were generated.
+
+    ```
+
+    The response above indicates that the snapshots are identical. If the snapshots were not identical, the command results will indicate where the comparison output files are written.
```
jcastrence: Maybe we can also include an example of running the command between two snapshots where there are differences:

'''
ledgerutil compare snapshots/ss1 snapshots/ss2

Successfully compared snapshots. Results saved to mychannel_3_comparison. Total differences found: 3
'''

manish-sethi (approved):

PR #3088, opened 2021-12-01
Corrected container names in commercial paper tutorial

nidhi-singh02:
Corrected container names in commercial paper tutorial

#### Type of change

- Documentation update

#### Description
Corrected container names in commercial paper tutorial. The container names for Org1 and Org2 were swapped with each other.

Signed-off-by: Nidhi Singh <nidhi2894@gmail.com>

### DIFF
--- a/docs/source/tutorial/commercial_paper.md
+++ b/docs/source/tutorial/commercial_paper.md
@@ -170,8 +170,8 @@ a86f50ca1907        hyperledger/fabric-peer:latest      "peer node start"       
 
 See if you can map these containers to the nodes of the test network (you may
 need to horizontally scroll to locate the information):
-* The Org1 peer, `peer0.org1.example.com`, is running in container `a86f50ca1907`
-* The Org2 peer, `peer0.org2.example.com`, is running in container `77d0fcaee61b`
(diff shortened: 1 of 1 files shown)


nikhil550 (commented):
The container names are different each time the test network is run. As a result, the container names are just for reference rather than a value that reflects the user experience.

nidhi-singh02:
> The container names are different each time the test network is run. As a result, the container names are just for reference rather than a value that reflects the user experience.

yes, that is there. The container names for peer of Org1 and Org2 are inter-changed in the documentation.

denyeart:
> > The container names are different each time the test network is run. As a result, the container names are just for reference rather than a value that reflects the user experience.
> 
> yes, that is there. The container names for peer of Org1 and Org2 are inter-changed in the documentation.

But you've only updated the container ids in this PR. They are just examples and will be different every time, so not sure how they could be incorrect and require fixing.

denyeart:
BTW, if you do want to proceed with a PR, you'll need to sign the commit, see:
https://github.com/hyperledger/fabric/pull/3088/checks?check_run_id=4380855094

denyeart:
Haven't heard anything for one month, will close for now...

Issue #3022, opened 2021-11-08
Upgrade Private data store retroactively


manish-sethi:
Upgrade private data store at peer boot time for creating HashedIndex entries retroactively.

denyeart:
#3092 implemented most of the logic, but I think it still needs to be wired into peer startup. @manish-sethi should we keep this issue open for that, or do you expect to handle in a different issue?

PR #3092, opened 2021-12-01
Create HashedIndex in Pvtdata Store retroactively

manish-sethi:

- Add functions in the pvtdata store package that would be used for creating the HashedIndex in Pvtdata Store retroactively.
- We use this opportunity to upgrade the data format from V11 to V12, so we can simplify the reading code in the future releases.
- Small refactoring is included to simplify the usage of update batch.

#### Type of change
- New feature

#### Related issues
- [RFC](https://github.com/hyperledger/fabric-rfcs/blob/main/text/0000-private_data_purge.md)
- [Issue](https://github.com/hyperledger/fabric/issues/3022)

### DIFF
--- a/common/ledger/util/leveldbhelper/leveldb_provider.go
+++ b/common/ledger/util/leveldbhelper/leveldb_provider.go
@@ -55,6 +55,39 @@ type Provider struct {
 	dbHandles map[string]*DBHandle
 }
 
+// DataFormatInfo contains the information about the version of the data format
+type DataFormatInfo struct {
+	FormatVerison string // version of the data format
+	IsDBEmpty     bool   // set to true if the db does not contain any data
+}
+
+// RetrieveDataFormatInfo retrieves the DataFormatInfo for the db at the supplied `dbPath`
+func RetrieveDataFormatInfo(dbPath string) (*DataFormatInfo, error) {
+	db := CreateDB(&Conf{DBPath: dbPath})
+	db.Open()
+	defer db.Close()
+
+	dbEmpty, err := db.IsEmpty()
+	if err != nil {
+		return nil, err
+	}
+
+	internalDB := &DBHandle{
+		db:     db,
+		dbName: internalDBName,
+	}
+
+	formatVersion, err := internalDB.Get(formatVersionKey)
+	if err != nil {
+		return nil, err
+	}
+
+	return &DataFormatInfo{
+		IsDBEmpty:     dbEmpty,
+		FormatVerison: string(formatVersion),
+	}, nil
+}
+
 // NewProvider constructs a Provider
 func NewProvider(conf *Conf) (*Provider, error) {
 	db, err := openDBAndCheckFormat(conf)
@@ -118,6 +151,11 @@ func openDBAndCheckFormat(conf *Conf) (d *DB, e error) {
 func (p *Provider) GetDataFormat() (string, error) {
 	f, err := p.GetDBHandle(internalDBName).Get(formatVersionKey)
 	return string(f), err
+}
+
+func (p *Provider) SetDataFormat(format string) error {
+	db := p.GetDBHandle(internalDBName)
+	return db.Put(formatVersionKey, []byte(format), true)
 }
 
 // GetDBHandle returns a handle to a named db
@@ -229,17 +267,17 @@ func (h *DBHandle) IsEmpty() (bool, error) {
 // NewUpdateBatch returns a new UpdateBatch that can be used to update the db
 func (h *DBHandle) NewUpdateBatch() *UpdateBatch {
 	return &UpdateBatch{
-		dbName: h.dbName,
-		Batch:  &leveldb.Batch{},
+		dbName:       h.dbName,
+		leveldbBatch: &leveldb.Batch{},
 	}
 }
 
 // WriteBatch writes a batch in an atomic way
 func (h *DBHandle) WriteBatch(batch *UpdateBatch, sync bool) error {
-	if batch == nil || batch.Len() == 0 {
+	if batch == nil || batch.leveldbBatch.Len() == 0 {
 		return nil
 	}
-	if err := h.db.WriteBatch(batch.Batch, sync); err != nil {
+	if err := h.db.WriteBatch(batch.leveldbBatch, sync); err != nil {
 		return err
 	}
 	return nil
@@ -273,8 +311,9 @@ func (h *DBHandle) Close() {
 
 // UpdateBatch encloses the details of multiple `updates`
 type UpdateBatch struct {
-	*leveldb.Batch
-	dbName string
+	leveldbBatch *leveldb.Batch
+	dbName       string
+	size         int
 }
 
 // Put adds a KV
@@ -282,12 +321,32 @@ func (b *UpdateBatch) Put(key []byte, value []byte) {
 	if value == nil {
 		panic("Nil value not allowed")
 	}
-	b.Batch.Put(constructLevelKey(b.dbName, key), value)
+	k := constructLevelKey(b.dbName, key)
+	b.leveldbBatch.Put(k, value)
+	b.size += len(k) + len(value)
 }
 
 // Delete deletes a Key and associated value
 func (b *UpdateBatch) Delete(key []byte) {
-	b.Batch.Delete(constructLevelKey(b.dbName, key))
+	k := constructLevelKey(b.dbName, key)
+	b.size += len(k)
+	b.leveldbBatch.Delete(k)
+}
+
+// Size returns the current size of the batch
+func (b *UpdateBatch) Size() int {
+	return b.size
+}
+
+// Len returns number of records in the batch
+func (b *UpdateBatch) Len() int {
+	return b.leveldbBatch.Len()
+}
+
+// Reset resets the batch
+func (b *UpdateBatch) Reset() {
+	b.leveldbBatch.Reset()
+	b.size = 0
 }
 
 // Iterator extends actual leveldb iterator
--- a/core/ledger/pvtdatastorage/retroactive_hashed_index.go
+++ b/core/ledger/pvtdatastorage/retroactive_hashed_index.go
@@ -1,0 +1,176 @@
+/*
+Copyright IBM Corp. All Rights Reserved.
+
+SPDX-License-Identifier: Apache-2.0
+*/
+
+package pvtdatastorage
+
+import (
+	"time"
+
+	"github.com/hyperledger/fabric-protos-go/ledger/rwset"
+	"github.com/hyperledger/fabric/common/ledger/util/leveldbhelper"
+	"github.com/hyperledger/fabric/core/ledger"
+	"github.com/hyperledger/fabric/core/ledger/kvledger/txmgmt/rwsetutil"
+	"github.com/hyperledger/fabric/core/ledger/util"
+	"github.com/pkg/errors"
+)
+
+const (
+	previousDataVersion = ""
+	currentDataVersion  = "2.5"
+	maxUpgradeBatchSize = 4 * 1024 * 1024 // 4 MB
+)
+
+func checkAndConstructHashedIndex(storePath string, ledgerIDs []string) error {
+	info, err := leveldbhelper.RetrieveDataFormatInfo(storePath)
+	if err != nil {
+		return err
+	}
+
+	if info.IsDBEmpty || info.FormatVerison == currentDataVersion {
+		return nil
+	}
+
+	if info.FormatVerison == previousDataVersion {
+		if err := constructHashedIndex(storePath, ledgerIDs); err != nil {
+			return err
+		}
+		return nil
+	}
+
+	return errors.Errorf("unexpected data version - cannot upgrade data format for pvtdatastore from %s to %s", info.FormatVerison, currentDataVersion)
+}
+
+// constructHashedIndex creates the HashedIndex entries for the private data keys and at the end sets the
+// data format version to the current version (2.5)
+func constructHashedIndex(storePath string, ledgerIDs []string) error {
+	p, err := leveldbhelper.NewProvider(
+		&leveldbhelper.Conf{
+			DBPath:         storePath,
+			ExpectedFormat: previousDataVersion,
+		})
+	if err != nil {
+		return err
+	}
+
+	defer p.Close()
+
+	for _, l := range ledgerIDs {
+		db := p.GetDBHandle(l)
+		if err := constructHashedIndexFor(l, db); err != nil {
+			return err
+		}
+	}
+	return p.SetDataFormat(currentDataVersion)
+}
+
+// constructHashedIndexFor creates the HashedIndex entries for a given ledger.
+// In this function we also piggyback to upgrade any private data key from format V11 to V12
+func constructHashedIndexFor(ledgerID string, db *leveldbhelper.DBHandle) error {
+	startKey, endKey := entireDatakeyRange()
+	itr, err := db.GetIterator(startKey, endKey)
+	if err != nil {
+		return err
+	}
+	defer itr.Release()
+
+	batch := db.NewUpdateBatch()
+
+	initialTime := time.Now()
+	numEntriesCreated := 0
+	logger.Infow("Starting creation of HashedIndex entries retroactively", "ledgerID", ledgerID)
+	for itr.Next() {
+		k := itr.Key()
+		v := itr.Value()
+
+		// In pvtdatastore, we used to persist the data of entire transaction as a single KV in V11.
+		// Later, when we introduced BlockToLive, we started storing each collection data as a KV.
+		// Here we leverage this opportunity (when we create HashedIndexKeys retroactively) to convert
+		// the private data in V11 format, if any so in the next available opportunity,
+		// we will remove the code where we have to detect this difference while processing queries on pvtdatastore.
+		v11Fmt, err := v11Format(k)
+		if err != nil {
+			return err
+		}
+
+		if v11Fmt {
+			blockNum, _, err := v11DecodePK(k)
+			if err != nil {
+				return err
+			}
+
+			txPvtData, err := v11DecodeKV(k, v, nil)
+			if err != nil {
+				return err
+			}
+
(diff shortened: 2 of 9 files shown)


common/ledger/util/leveldbhelper/leveldb_provider.go
```diff
@@ -55,6 +55,37 @@ type Provider struct {
 	dbHandles map[string]*DBHandle
 }
 
+type DataFormatInfo struct {
```
denyeart, line 59: Add comments for all exported structs and functions.

core/ledger/pvtdatastorage/retroactive_hashed_index.go
```diff
@@ -0,0 +1,161 @@
...
+			if err != nil {
+				return err
+			}
+
+			txPvtData, err := v11DecodeKV(k, v, nil)
+			if err != nil {
+				return err
+			}
+
+			dataEntries := prepareDataEntries(blockNum, []*ledger.TxPvtData{txPvtData})
```
denyeart, line 109: I find generic names like "dataEntries" confusing. Can we be more specific both in the variable name and function name?
For example I think "dataEntries" spans all collections, while "dataEntry" is for a single collection.

Also, prepareDataEntries() is for v11 format only right? Therefore could we put "v11" in the function name so that we know this is legacy (with a potential to remove in the future), similar to what you've done with v11DecodePK() function name? Same for any other functions related to v11 format.

core/ledger/pvtdatastorage/retroactive_hashed_index.go
```diff
@@ -0,0 +1,161 @@
...
+				if dataVal, err = encodeDataValue(dataEntry.value); err != nil {
+					return err
+				}
+				batch.Put(dataKey, dataVal)
+				if err := addHashedIndexEntriesInto(batch, dataEntry.key, dataEntry.value); err != nil {
+					return err
+				}
+			}
+			batch.Delete(k)
+		} else {
```
denyeart, line 123: For any else statement distant from the original if, it is nice to have a comment reminding the reader what falls into this code branch.

core/ledger/pvtdatastorage/retroactive_hashed_index.go
```diff
@@ -0,0 +1,161 @@
...
+			return err
+		}
+		return nil
+	}
+
+	return errors.Errorf("cannot upgrade data format for pvtdatastore from %s to %s", i.FormatVerison, currentDataVersion)
+}
+
+// constructHashedIndex creates the HashedIndex entries for the private data keys and at the end sets the
+// data format version to the current version (2.6)
```
denyeart:
```suggestion
// data format version to the current version (2.5)
```

core/ledger/pvtdatastorage/retroactive_hashed_index.go
```diff
@@ -0,0 +1,161 @@
...
+
+	"github.com/hyperledger/fabric/core/ledger"
+	"github.com/hyperledger/fabric/core/ledger/kvledger/txmgmt/rwsetutil"
+	"github.com/hyperledger/fabric/core/ledger/util"
+	"github.com/pkg/errors"
+)
+
+const (
+	previousDataVersion = ""
+	currentDataVersion  = "2.6"
```
denyeart:
```suggestion
	currentDataVersion  = "2.5"
```

core/ledger/pvtdatastorage/retroactive_hashed_index.go
```diff
@@ -0,0 +1,161 @@
...
+	maxUpgradeBatchSize = 4 * 1024 * 1024 // 4 MB
+)
+
+func checkAndConstructHashedIndex(storePath string, ledgerIDs []string) error {
+	i, err := leveldbhelper.RetrieveDataFormatInfo(storePath)
```
denyeart: I usually think of an "i" variable as an index or counter, could we have an expressive variable name?

core/ledger/pvtdatastorage/retroactive_hashed_index.go
```diff
@@ -0,0 +1,161 @@
...
+		return err
+	}
+
+	if i.IsDBEmpty || i.FormatVerison == currentDataVersion {
+		return nil
+	}
+
+	if i.FormatVerison == previousDataVersion {
+		if err := constructHashedIndex(storePath, ledgerIDs); err != nil {
+			return err
```
denyeart, line 38: I don't think we need to do it everywhere, but for the important error handling points could we augment the underlying error message with WithMessage() so that we'll get more error context.

core/ledger/pvtdatastorage/retroactive_hashed_index.go
```diff
@@ -0,0 +1,161 @@
...
+		}
+		return nil
+	}
+
+	return errors.Errorf("cannot upgrade data format for pvtdatastore from %s to %s", i.FormatVerison, currentDataVersion)
```
denyeart: Shouldn't the error be something like "unexpected version" rather than "cannot upgrade".

core/ledger/pvtdatastorage/retroactive_hashed_index.go
manish-sethi: Ran ahead of time :-(

core/ledger/pvtdatastorage/retroactive_hashed_index.go
manish-sethi: When I started this practice of following golang style, it seemed odd to me as well. But after a while, I am adjusted and like it better - when the variable lifespan is to short, use of a single letter variable makes it better to read.

core/ledger/pvtdatastorage/retroactive_hashed_index.go
manish-sethi: To me, every error is important. Rather, the way I like to differentiate is whether user can take some action for fixing the error or it will have to come to developer. If it has to come to developer, the stack trace is best and all the context is there in the stack trace. WithMessage is useful, if I could instruct the user to fix something..(like change some env variable, or retry a request, etc..)

core/ledger/pvtdatastorage/retroactive_hashed_index.go
manish-sethi: > I find generic names like "dataEntries" confusing. Can we be more specific both in the variable name and function name?

This is not a generic name. This is very much in accordance with the [naming of various data elements in the current design](https://github.com/hyperledger/fabric/blob/c0498d4f22d52b9c4e1a63fc9239b90bf9dc9d4e/core/ledger/pvtdatastorage/store.go#L123-L128 ).

> For example I think "dataEntries" spans all collections, while "dataEntry" is for a single collection.

Given the above, in the `StoreEntries`, a single `dataEntry` represents one KV in the store for data and `dataEntries` is simply a plural. In fact, we have a [type](https://github.com/hyperledger/fabric/blob/516bc8bab548a484e2e62fee24b0f4fed5bf6873/core/ledger/pvtdatastorage/store.go#L81) for this as well.

> Also, prepareDataEntries() is for v11 format only right? Therefore could we put "v11" in the function name so that we > know this is legacy (with a potential to remove in the future), similar to what you've done with v11DecodePK() function name? Same for any other functions related to v11 format

No, `prepareDataEntries() ` is the existing function that produces data in V12 format.

core/ledger/pvtdatastorage/retroactive_hashed_index.go
manish-sethi: I guess we have different measure of distance. Both, `if` and `else`, are visible together within half screen :-)

core/ledger/pvtdatastorage/retroactive_hashed_index.go
manish-sethi: Agreed!

core/ledger/pvtdatastorage/retroactive_hashed_index.go
davidkel: I personally find it harder to read. I would like to be able to understand a line of code without having to look elsewhere to remind me what a single character actually is, single character variables I think is one of the worst ideas to come out of idiomatic go

core/ledger/pvtdatastorage/retroactive_hashed_index.go
denyeart: I think you've proven my point by pointing to the struct:

```
type storeEntries struct {
	dataEntries             []*dataEntry
	expiryEntries           []*expiryEntry
	elgMissingDataEntries   map[missingDataKey]*bitset.BitSet
	inelgMissingDataEntries map[missingDataKey]*bitset.BitSet
}
```

Three of the fields have descriptive names which makes them easy to understand. One of the fields has a generic name "dataEntry" which makes it impossible to know what kind of data it is. Everything in the ledger is a "dataEntry". I'm asking to update the name of the type as well as the variable so that readers know what kind of data it is.

core/ledger/pvtdatastorage/retroactive_hashed_index.go
denyeart: If it is a well known noun like an "index" I can agree to an i. Same for k for "key" and v for "value" since they are well known nouns in the domain. But an "Info" is a one-off that nobody reading ledger code knows about, and therefore deserves more than a letter (especially a letter that usually means something else).

core/ledger/pvtdatastorage/retroactive_hashed_index.go
denyeart: I'd say it requires either a stack or a WithMessage(). Does this one produce a stack?  If not, simply returning only the underlying error helps neither the user nor the SRE.

denyeart:
@manish-sethi I resolved most of the comments, but for three of them I've clarified the review comment.

core/ledger/pvtdatastorage/retroactive_hashed_index.go
```diff
@@ -0,0 +1,161 @@
...
+			if err := addHashedIndexEntriesInto(batch, dataKey, dataVal); err != nil {
+				return err
+			}
+		}
+
+		if batch.Size() >= maxUpgradeBatchSize {
+			if err := db.WriteBatch(batch, true); err != nil {
+				return err
+			}
+			batch.Reset()
```
denyeart, line 147: Since it may take some time to iterate through all the private data during first peer start after upgrade, it is good to provide some log info statements so that administrator has a clue why server isn't coming online as quick as expected. Otherwise admin may think the server is hung during startup. This may be a convenient place to add the log info statements so that admin knows progress is being made.

core/ledger/pvtdatastorage/retroactive_hashed_index.go
manish-sethi: Every error returned by fabric code is expected to contain the stack trace, so yes nothing to be done here.

core/ledger/pvtdatastorage/retroactive_hashed_index.go
manish-sethi: Not sure what ledger code has to do with variable `i` here. It's the only variable here, defined in first line and used in the next two lines. In fact, single letter variable are encouraged, so you can read the code better by simply overlooking it.

If the variable spans more than handful of lines, I agree, and you would see in a typical golang code as well, you would be encouraged to use longer/expressive names.

 @davidkel - everyone has different preferences that we can keep debating about... That's why it is preferred to adapt the style of the language at hand. If many folks are finding this readable perhaps we need to get adapted rather than writing code in a language that smells like some other language.

core/ledger/pvtdatastorage/retroactive_hashed_index.go
manish-sethi: We are free to draw any conclusion but IMO, it proves other way. `data` means data as `expiry` means expiry, `missingData` mean missing data. I am not sure where the confusion is. I guess that you are applying general meaning to it. Not sure what Reader you have in mind -- These terminologies are there since day one. Anyone who had written code in this package and understood the code/design never raised this issue. Not sure when you say what kind of data - in private store context, of course data means `private data`.  

In any case, changing all over this type and name is beyond the scope of this PR and given that I am not convinced with your observation, we can discuss outside this PR and you are welcome to suggest the change.

core/ledger/pvtdatastorage/retroactive_hashed_index.go
manish-sethi: Agreed! Will add a log statement here.

core/ledger/pvtdatastorage/retroactive_hashed_index.go
denyeart: @manish-sethi It is not as black and white as you indicate. You can't just assume that other viewpoints are merely opinions or from non-adapted developers while your own viewpoints are the correct ones, and that `i` is more readable than `info`.

While it is true that variables in close proximity should be short or even single-letter, there are also suggestions in the same Go guidelines that `i` shouldn't be used for things that aren't loop indexes/incrementers.

https://github.com/golang/go/wiki/CodeReviewComments#variable-names

> Common variables such as loop indices and readers can be a single letter (i, r). More unusual things and global variables need more descriptive names.

In the context of ledger, an Info is a more unusual thing (not a well-known noun as I mentioned earlier).

https://dave.cheney.net/practical-go/presentations/gophercon-singapore-2019.html

> Finally, certain single letter variables have traditionally been associated with loops and counting. For example, i, j, and k are commonly the loop induction variable for simple for loops. [2] n is commonly associated with a counter or accumulator. v is a common shorthand for a value in a generic encoding function, k is commonly used for the key of a map. a and b are generic names for parameters comparing two variables of the same type. x and y are generic names for local variables created for comparison, and s is often used as shorthand for parameters of type string.
> 
> As with the db example above programmers expect i to be a loop induction variable. If you ensure that i is always a loop variable, not used in other contexts outside a for loop. When readers encounter a variable called i, or j, they know that a loop is close by.

This debate has already gone on longer than it should have, it was a minor comment and I'm happy to move on without it. My larger point is that we should try to be receptive to comments from people reading the code (especially when they are backed up by Go guidelines!). Code readability is best judged by people trying to read the code. I actually didn't know about the written Go guidelines about not using `i` outside of loop context due to potential confusions, I only knew that it tripped ME up when first scanning the code for that very reason. If there are two people indicating the readability could be improved, isn't that enough?

core/ledger/pvtdatastorage/retroactive_hashed_index.go
denyeart: The only thing I can agree to here is that `dataEntry` preceded this PR and can be discussed outside.
As the code author, `dataEntry` clearly means something very specific to you. As a non-author reader, I can tell you such a general term (everything in the ledger is a data entry!) doesn't mean something specific to somebody reading the code for the first time. The definition of private data store data entry has changed over time and even I forget which way was original and which way is current, therefore something like `collDataEntry` would reinforce that the data entries are for private data writes and scoped at collection level.

The fact that it was there since day one is insignificant, I thought we all agreed to review code with a more critical eye towards readability than we have in the past, and improve existing code as we go.

Again, it was a minor point and I'm happy to move on.

core/ledger/pvtdatastorage/retroactive_hashed_index.go
manish-sethi: > @manish-sethi It is not as black and white as you indicate. You can't just assume that other viewpoints are merely opinions or from non-adapted developers while your own viewpoints are the correct ones, and that i is more readable than info.

@denyeart - Not sure why you are giving this a spin. All I am saying is that we all need to get over our preferences/view points and try to get use to the convention of the language. E.g., if you apply these golang convention to java code, you would hear other way around. I'll paste the complete text from where you pasted selected text.

> Variable names in Go should be short rather than long. This is especially true for local variables with limited scope. Prefer c to lineCount. Prefer i to sliceIndex.

> The basic rule: the further from its declaration that a name is used, the more descriptive the name must be. For a method receiver, one or two letters is sufficient. Common variables such as loop indices and readers can be a single letter (i, r). More unusual things and global variables need more descriptive names.`

Then this debate gets reduced to whether `info` is unusual or not in the context of this function - which is a 10 line function and the life of this variable is 3-4 line. 

IMO, to learn the meaning of open guidance (like usual v/s unusual), one will have to learn by looking at standard code, beyond guidelines. As an example a [sample code snippet](https://github.com/golang/go/blob/dc50d69119ca8070dd16fb968b46f21f854d89fb/src/cmd/go/internal/list/list.go#L594) from golang src code. This `if` block, (and next `for loop`) we all may find unbearable because local variables are not expressive but that is what is encouraged in this convention.

> If there are two people indicating the readability could be improved, isn't that enough?

It's a matter of choice whether we all collectively stay at what is readable to us or get used to non-expressive local variables of shorter life span?

Yes, this debate has gone overboard which it does not deserve. I'll change this to `info` to keep you both happy but still I would encourage you to look at the usage of shorter/not expressive local variable in standard code - it's full of this.

core/ledger/pvtdatastorage/retroactive_hashed_index.go
manish-sethi: Yes, I would be happy to agree with the disagreement on this :-)

core/ledger/pvtdatastorage/retroactive_hashed_index.go
```diff
@@ -0,0 +1,176 @@
...
+				logger.Infow("Creating HashedIndex entries retroactively...", "total entries created", numEntriesCreated)
+			}
+			batch.Reset()
+		}
+	}
+
+	if err := db.WriteBatch(batch, true); err != nil {
+		return err
+	}
+	logger.Infow("Creatiion of HashedIndex entries retroactively done", "total entries created", numEntriesCreated)
```
denyeart:
```suggestion
	logger.Infow("Creation of HashedIndex entries retroactively done", "total entries created", numEntriesCreated)
```

After the long discussion I wanted to merge quick but found this typo :-)

Might as well add channel name (ledgerId) to each of the new log messages as well.

manish-sethi: > After the long discussion I wanted to merge quick but found this typo :-)

Oh no :-(

denyeart (approved):

PR #3097, opened 2021-12-04
Update peer-chaincode-devmode.md

bellaj:
Adding a new instruction to create /var/hyperledger folder required by Hyperledger

#### Type of change

- Documentation update

#### Description

This PR updates the documentation by adding instructions for creating `/var/hyperledger` directory. If such directory doesn't exist the orderer will not be able to run after executing 
`ORDERER_GENERAL_GENESISPROFILE=SampleDevModeSolo orderer`

### DIFF
--- a/docs/source/peer-chaincode-devmode.md
+++ b/docs/source/peer-chaincode-devmode.md
@@ -33,7 +33,15 @@ Throughout this tutorial, all commands are performed from the `fabric/` folder. 
     ```
     export FABRIC_CFG_PATH=$(pwd)/sampleconfig
     ```
-5. Generate the genesis block for the ordering service. Run the following command to generate the genesis block and store it in `$(pwd)/sampleconfig/genesisblock` so that it can be used by the orderer in the next step when the orderer is started.
+
+5. Create the `hyperledger` subdirectory in the `/var` directory. This is a location Fabric uses. To create the hyperledger subdirectory, execute these commands, replacing the question marks with your username:
+
+```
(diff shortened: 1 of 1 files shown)


docs/source/peer-chaincode-devmode.md
```diff
@@ -33,7 +33,15 @@ Throughout this tutorial, all commands are performed from the `fabric/` folder.
     ```
     export FABRIC_CFG_PATH=$(pwd)/sampleconfig
     ```
-5. Generate the genesis block for the ordering service. Run the following command to generate the genesis block and store it in `$(pwd)/sampleconfig/genesisblock` so that it can be used by the orderer in the next step when the orderer is started.
+
+5. Create the `hyperledger` subdirectory in the `/var` directory. This is a location Fabric uses. To create the hyperledger subdirectory, execute these commands, replacing the question marks with your username:
```
denyeart, line 37: Can you clarify what Fabric uses in /var/hyperledger. I assume you are talking about the orderer and peer ledger location? Assuming the configtxgen step doesn't need it, I'd move this down to right before starting the orderer and peer.

denyeart:
You need to sign the commit as mentioned at:
https://github.com/hyperledger/fabric/pull/3097/checks?check_run_id=4417068867

docs/source/peer-chaincode-devmode.md
bellaj: >  I assume you are talking about the orderer and peer ledger location

Exactly

PR #3012, opened 2021-11-02
Show what do not match

gqqnbig:
Signed-off-by: gqqnbig <gqqnb2005@gmail.com>

#### Type of change

- Improvement (improvement to code, performance, etc)

#### Description

When  `peer chaincode invoke`  submits a request, if chaincode has bugs, it may return different responses on different nodes.

Before, the error message is like

> Error: could not assemble transaction: ProposalResponsePayloads do not match  - proposal response: version:1 response:<status:200 payload:"true" > payload:"\n \265\277\020\265\226D\035\361\001n\371K\013\274\021F\227\002?\261\204\370\327\363r\347\010\221\313\214\031\230\022\213\001\no\0227\n\n_lifecycle\022)\n'..." endorsement:<endorser:"\n\007Org2MSP\022\252\006-----BEGIN CERTIFICATE-----\nMIICKDCCAc+gA..." signature:"0D\002 J\\\300\307~\350\236\026\245H\003\337\2732\020\312\n\017Fa\306~\367x\356\347K\225\..." > 

That is it. Developers have no idea what mismatch what.

Now, with this PR, the error message is like 

> Error: could not assemble transaction: ProposalResponsePayloads do not match with version:1 response:<status:200 payload:"true" > payload:"\n \265\277\020\265\226D\035\361\001n\371K\013\274\021F\227\002?\261\204\370\327\363r\347\010\221\313\214\031\230\022\230\001\n|\0227\n\n_lifecycle\022)\n'\n..." endorsement:<endorser:"\n\007Org1MSP\022\252\006-----BEGIN CERTIFICATE-----\nMIICKDCCAc6gAwIBAgIQEO..." signature:"0E\002!\000\331p\034A\353t\261{\026#\177\211\371g\232\213\331\323\..." >
> &#45; proposal response: version:1 response:<status:200 payload:"true" > payload:"\n \265\277\020\265\226D\035\361\001n\371K\013\274\021F\227\002?\261\204\370\327\363r\347\010\221\313\214\031\230\022\213\001\no\0227\n\n_lifecycle\022)\n'..." endorsement:<endorser:"\n\007Org2MSP\022\252\006-----BEGIN CERTIFICATE-----\nMIICKDCCAc+gA..." signature:"0D\002 J\\\300\307~\350\236\026\245H\003\337\2732\020\312\n\017Fa\306~\367x\356\347K\225\..." > 

With this error message, developers can use text comparison tools to find out where the two responses start to differ, and they may get some cues.

[...]

PR #3104, opened 2021-12-08
Gateway to use simplified createTx function

### LINKED ISSUE #332: client loses connection to gateway peer and cannot recover without a restart when endorsement policy is changed
First I will note that the client could probably recover but would have to be coded to detect the scenario and create a new client and gateway to recover. The problem I see reports 2 errors
```
{"component":"CLIENT","timestamp":"2021-11-25T17:05:58.440Z","txnId":"f4bcbe05d1c9257ebfbaa7492d3a853935ab69789c2d2feb661b413662d10a76","stage":"Endorsing","message":"addUpdateAssets({\"arguments\":[\"1\",\"2000\"]})"}
{"component":"CLIENT","timestamp":"2021-11-25T17:06:28.454Z","txnId":"f4bcbe05d1c9257ebfbaa7492d3a853935ab69789c2d2feb661b413662d10a76","stage":"Failed","message":"4 DEADLINE_EXCEEDED: Deadline exceeded"}
{"component":"CLIENT","timestamp":"2021-11-25T17:05:58.443Z","txnId":"bcd6cc53c262a76f23015a2003af3e72785079f44be01ba28840c6adaa31bc6a","stage":"Endorsing","message":"updateChaosAsset({\"arguments\":[\"cd1\",\"99\"]})"}
{"component":"CLIENT","timestamp":"2021-11-25T17:06:28.454Z","txnId":"bcd6cc53c262a76f23015a2003af3e72785079f44be01ba28840c6adaa31bc6a","stage":"Failed","message":"14 UNAVAILABLE: Stream refused by server"}
```
But either way no further submitted transaction succeeds until the client was restarted

Unfortunately client side grpc logging doesn't tell me much only saying what the error message says. The peer logs don't provide anything helpful and if I turn on too much logging then I struggled to recreate the problem.

I recreated the scenario outside of the endorsement policy by having a long running transaction create MVCC Read Conflicts so it looks related to multiple long running transactions failing which have a large write set, in my case writing to 2000 keys but small amounts of data.

A recent change to fabric here https://github.com/hyperledger/fabric/pull/3012 changed the output for when proposal responses don't match by including base64 version of protobuf bytes into the error which subsequently end up in the logs. If I revert this change then I haven't been able to hit the problem. It looks like this change was intended for when the peer is used as a cli but also occurs when the peer is running as a server.

If I don't use the long running txns with large number of keys then the problem doesn't occur, so this is specific to the contrived application being run to test the scenario, so it's not likely to be a common problem so I will leave it up for discussion about whether the recent change to outputting base64 when proposals don't match is of value and should be output when the peer is running in server mode.

andrew-coleman:
The gateway has been reusing the protoutil.CreateTx function to build the transaction envelope from the set of endorsements.  This function changed recently to add extra information into the error message if the proposal responses don’t all match.  This change was identified as causing a problem for the gateway (https://github.com/hyperledger/fabric-gateway/issues/332).
This commit adds a simplified createTx function in the gateway package that removes this amendment and also removes some other checks that aren’t relevant to its usage in the gateway.

Resolves https://github.com/hyperledger/fabric-gateway/issues/332


internal/pkg/gateway/api.go
```diff
@@ -321,7 +337,7 @@ func (gs *Server) Endorse(ctx context.Context, request *gp.EndorseRequest) (*gp.
 		return nil, newRpcError(codes.Aborted, "failed to collect enough transaction endorsements, see attached details for more info", errorDetails...)
 	}
 
-	env, err := protoutil.CreateTx(proposal, endorsements...)
```
yacovm, line 324: But previous this was the only place in the code that was using protoutil.CreateTx, so shouldn't we also erase CreateTx altogether?

internal/pkg/gateway/apiutils.go
```diff
@@ -57,9 +59,44 @@ func newRpcError(code codes.Code, message string, details ...proto.Message) erro
...
+) (*common.Envelope, error) {
+	// ensure that all actions are bitwise equal
+	payload := resps[0].Payload
+	for _, r := range resps {
+		if !bytes.Equal(payload, r.Payload) {
+			return nil, errors.New("ProposalResponsePayloads do not match")
+		}
+	}
+
+	var endorsements []*peer.Endorsement
```
yacovm: Can we please not revert the logic that was put in place in this [PR](https://github.com/hyperledger/fabric/pull/3040)? I am referring to the de-duplication.

Let's de-duplicate it like in that PR if you wish to move the logic here.

yacovm (dismissed):
See comments. 

In https://github.com/hyperledger/fabric/pull/3040 we put in place a de-duplication mechanism for the case of identical endorsements. Let's keep it that way please.

yacovm:
Also, I don't get it. That [commit](https://github.com/hyperledger/fabric/pull/3012) added the strings to the error returned. 
Why not simply change the code there to log the payload instead of returning it? Surely it will achieve the intended goal and will be a smaller change and not leave us with two different implementations scattered throughout the code base.

ale-linux:
I'm not fully up to speed wrt. the gateway logic, but: it seems to me that a proliferation of code paths doing the same thing - some with a certain amount of error messages and others with fewer such messages - isn't great.

Isn't there a way to just get the gateway to work with the current code path that includes the extra info?

internal/pkg/gateway/api.go
andrew-coleman: Yep, although given the feedback, I'll rework it so that it shares as much common code as possible.  I am trying to avoid the duplicate unmarshalling that's currently happening in the gateway,  but I can rework the 2 wrapper functions in protoutil to avoid that.

internal/pkg/gateway/apiutils.go
andrew-coleman: Sure - I hadn't fully appreciated why that code was there until I read your linked PR.

andrew-coleman:
> Why not simply change the code there to log the payload instead of returning it? Surely it will achieve the intended goal and will be a smaller change and not leave us with two different implementations scattered throughout the code base.

Sure, I'm happy to do that if @gqqnbig is.

yacovm:
> > Why not simply change the code there to log the payload instead of returning it? Surely it will achieve the intended goal and will be a smaller change and not leave us with two different implementations scattered throughout the code base.
> 
> Sure, I'm happy to do that if @gqqnbig is.

We [can't](https://www.evanjones.ca/grpc-is-tricky.html) return errors with too much text in gRPC. 
Given that fact, I don't think we need to have that person's permission to change that line in the code.

gqqnbig:
I'm a CLI user for now, so I don't know much about gateway. I like refactoring and removing duplicate code.

If you change the base64 output to log, do I see the log from `docker logs` when I use [the test network](https://github.com/hyperledger/fabric-samples)? If yes, as long as I can see what is different, I'm fine with your change.

As @andrew-coleman said in hyperledger/fabric-gateway#332, a structured message that clearly shows the keys/values/versions that differ is much better. I'm all for it.

protoutil/txutils.go
```diff
@@ -14,9 +14,12 @@ import (
 	"github.com/golang/protobuf/proto"
 	"github.com/hyperledger/fabric-protos-go/common"
 	"github.com/hyperledger/fabric-protos-go/peer"
+	"github.com/hyperledger/fabric/common/flogging"
 	"github.com/pkg/errors"
 )
 
+var logger = flogging.MustGetLogger("protoutil")
```
yacovm: The logger of the caller may be decorated with all kinds of useful information such as channel name, etc. 

Let's inject a logger into the method call instead of using a static one here.

protoutil/txutils.go
```diff
@@ -203,8 +220,10 @@ func createTx(
 		}
 
 		if !bytes.Equal(a1, r.Payload) {
-			return nil, errors.Errorf("ProposalResponsePayloads do not match (base64): '%s' vs '%s'",
-				b64.StdEncoding.EncodeToString(r.Payload), b64.StdEncoding.EncodeToString(a1))
+			if logger != nil {
+				logger.Infow("ProposalResponsePayloads do not match (base64)", "payload1", b64.StdEncoding.EncodeToString(a1), "payload2", b64.StdEncoding.EncodeToString(r.Payload))
```
yacovm, line 230: why Info and not Warn?

andrew-coleman: I was persuaded by this article: https://dave.cheney.net/2015/11/05/lets-talk-about-logging

denyeart: I'm not persuaded by this article :-)
The precedent in Fabric is basically:
- Normal happy path things that admin might like to know about are logged at Info.
- Unexpected errors that admin should know about are logged at Error.
- Expected transaction issues (like this one) that an admin can usually ignore but might still want to know about are logged at Warning.

internal/pkg/gateway/api.go
```diff
@@ -134,12 +134,28 @@ func (gs *Server) Endorse(ctx context.Context, request *gp.EndorseRequest) (*gp.
 	}
 	proposal, err := protoutil.UnmarshalProposal(signedProposal.ProposalBytes)
 	if err != nil {
-		return nil, status.Errorf(codes.InvalidArgument, "failed to unpack transaction proposal: %s", err)
+		return nil, status.Error(codes.InvalidArgument, err.Error())
 	}
-	channel, chaincodeID, hasTransientData, err := getChannelAndChaincodeFromSignedProposal(signedProposal)
+	header, err := protoutil.UnmarshalHeader(proposal.Header)
```
denyeart, line 139: Can't getChannelAndChaincodeFromSignedProposal() be adapted to be used by both Evaluate() and Endorse()? I liked the symmetry and re-use you had originally, not to mention keeping the Endorse() logic clean.

internal/pkg/gateway/api.go
```diff
@@ -321,7 +337,7 @@ func (gs *Server) Endorse(ctx context.Context, request *gp.EndorseRequest) (*gp.
 		return nil, newRpcError(codes.Aborted, "failed to collect enough transaction endorsements, see attached details for more info", errorDetails...)
 	}
 
-	env, err := protoutil.CreateTx(proposal, endorsements...)
+	env, err := protoutil.CreateTx(header, proposal.Payload, logger.With("channel", channel, "txID", request.TransactionId), endorsements...)
```
denyeart, line 340: Above you changed the function name to CreateSignedTxWithLogger() when you added the logger. Here you add the logger without changing the function name. Better to be consistent.

andrew-coleman:
As discussed with @denyeart, based on performance analysis of the gateway, we have decided to change the endorsement collection code in the gateway to minimise CPU and memory usage.  Currently, the endorsement collection code is storing more information than required (i.e. all of the ProposalResponse objects rather than just the Endorsement objects, and one response payload) and repeating some of the unmarshalling in order to reuse the protoutils.CreateTx() function.  Closing this PR.

PR #3109, opened 2021-12-09
Update config transaction validation logging

denyeart:
Fix misleading debug message.
Log info when channel config is applied.

### DIFF
--- a/core/committer/txvalidator/v20/validator.go
+++ b/core/committer/txvalidator/v20/validator.go
@@ -406,6 +406,7 @@ func (v *TxValidator) validateTx(req *blockValidationRequest, results chan<- *bl
 				return
 			}
 
+			logger.Debugw("Config transaction envelope passed validation checks", "channel", channel)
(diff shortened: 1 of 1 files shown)


core/committer/txvalidator/v20/validator.go
```diff
@@ -406,6 +406,7 @@ func (v *TxValidator) validateTx(req *blockValidationRequest, results chan<- *bl
 				return
 			}
 
+			logger.Debugw("Config transaction passed validation checks", "channel", channel)
```
manish-sethi: This could be a bit misleading as for the config transaction, both the actual config validation and apply happens inside `Apply` function below. May be you want to enhance the message a bit - "Config transaction envelop passed validation checks"

ale-linux: +1 and by extension the next log entry should read "transaction validated and applied.."

denyeart: Thanks guys, done!

manish-sethi (approved):

PR #3110, opened 2021-12-09
Update 'Running a Fabric Application' tutorial for Fabric Gateway

bestbeforetoday:
Also fixed a few build warnings in other documents.

Contributes to #2807

### DIFF
--- a/docs/source/getting_started.rst
+++ b/docs/source/getting_started.rst
@@ -21,7 +21,7 @@ The Fabric application stack has five layers:
 * :doc:`Prerequisite software <prereqs>`: the base layer needed to run the software, for example, Docker.
 * :doc:`Fabric and Fabric samples <install>`: the Fabric executables to run a Fabric network along with sample code.
 * :doc:`Contract APIs <sdk_chaincode>`: to develop smart contracts executed on a Fabric Network.
-* :doc:`Application SDKs <sdk_chaincode>`: to develop your blockchain application.
+* :doc:`Application APIs <sdk_chaincode>`: to develop your blockchain application.
 * The Application: your blockchain application will utilize the Application SDKs to call smart contracts running on a Fabric network.
 
 
--- a/docs/source/write_first_app.rst
+++ b/docs/source/write_first_app.rst
@@ -1,1014 +1,567 @@
 Running a Fabric Application
-==============================
+############################
+.. note:: If you're not yet familiar with the fundamental architecture of a Fabric blockchain network, you may want to
+          visit the :doc:`key_concepts` section prior to continuing.
+          
+          You should also be familiar with the Fabric Gateway service and how it relates to the application transaction
+          flow, documented in the :doc:`gateway` section.
 
-.. note:: If you're not yet familiar with the fundamental architecture of a
-          Fabric network, you may want to visit the :doc:`key_concepts` section
-          prior to continuing.
-
-          It is also worth noting that this tutorial serves as an introduction
-          to Fabric applications and uses simple smart contracts and
-          applications. For a more in-depth look at Fabric applications and
-          smart contracts, check out our
-          :doc:`developapps/developing_applications` section or the
-          :doc:`tutorial/commercial_paper`.
-
-This tutorial provides an introduction to how Fabric applications interact
-with deployed blockchain networks. The tutorial uses sample programs built using the
-Fabric SDKs -- described in detail in the :doc:`/developapps/application` topic --
-to invoke a smart contract which queries and updates the ledger with the smart
-contract API -- described in detail in :doc:`/developapps/smartcontract`.
-We will also use our sample programs and a deployed Certificate Authority to generate
-the X.509 certificates that an application needs to interact with a permissioned
-blockchain.
+This tutorial provides an introduction to how Fabric applications interact with deployed blockchain networks. The
+tutorial uses sample programs built using the Fabric Gateway application API to invoke a smart contract, which queries
+and updates the ledger with the smart contract API -- described in detail in :doc:`/developapps/smartcontract`.
 
 **About Asset Transfer**
 
-This Asset Transfer (basic) sample demonstrates how to initialize a ledger with assets, query those assets, create
-a new asset, query a single asset based on an asset ID, update an existing asset, and transfer an asset to a new owner.
-It involves the following two components:
+The Asset Transfer (basic) sample demonstrates how to create, update, and query assets. It involves the following two
+components:
 
-  1. Sample application: which makes calls to the blockchain network, invoking transactions
-  implemented in the chaincode (smart contract). The application is located in the following ``fabric-samples`` directory:
+  1. **Sample application:** which makes calls to the blockchain network, invoking transactions
+  implemented in the smart contract. The application is located in the following ``fabric-samples`` directory:
 
-  .. code:: bash
+  .. code-block:: text
 
-    asset-transfer-basic/application-javascript
+    asset-transfer-basic/application-gateway-typescript
 
-  2. Smart contract itself, implementing the transactions that involve interactions with the
-  ledger. The smart contract (chaincode) is located in the following ``fabric-samples`` directory:
+  2. **Smart contract:** which implements the transactions that interact with the
+  ledger. The smart contract is located in the following ``fabric-samples`` directory:
 
-  .. code:: bash
+  .. code-block:: text
 
-    asset-transfer-basic/chaincode-(javascript, java, go, typescript)
+    asset-transfer-basic/chaincode-(typescript, go, java)
 
-Please note that for the purposes of this tutorial, the terms chaincode and smart contract are used
-interchangeably. For this example, we will be using the javascript chaincode.
+For this example, we will be using the TypeScript smart contract.
 
-We’ll go through three principle steps:
+This tutorial consists of two principle parts:
 
-  **1. Setting up a development environment.** Our application needs a network
-  to interact with, so we'll deploy a basic network for our smart contracts and
-  application.
+  1. **Set up a blockchain network.**
+  Our application needs a blockchain network to interact with, so we will launch a basic network and deploy a smart
+  contract for our application.
 
   .. image:: images/AppConceptsOverview.png
 
-  **2. Explore a sample smart contract.**
-  We’ll inspect the sample assetTransfer (javascript) smart contract to learn about the transactions within it,
-  and how they are used by an application to query and update the ledger.
+  2. **Run the sample application to interact with the smart contract.**
+  Our application will use the assetTransfer smart contract to create, query, and update assets on the ledger. We will
+  step through the code of the application and the transactions it invokes, including creating some initial assets,
+  querying an asset, querying a range of assets, creating a new asset, and transferring an asset to a new owner.
 
-  **3. Interact with the smart contract with a sample application.** Our application will
-  use the assetTransfer smart contract to create, query, and update assets on the ledger.
-  We'll get into the code of the app and the transactions they create, including initializing
-  the ledger with assets, querying an asset, querying a range of assets, creating a new asset,
-  and transferring an asset to a new owner.
+After completing this tutorial you should have a basic understanding of how Fabric applications and smart contracts
+work together to manage data on the distributed ledger of a blockchain network.
 
-After completing this tutorial you should have a basic understanding of how Fabric
-applications and smart contracts work together to manage data on the distributed
-ledger of a blockchain network.
 
 Before you begin
-----------------
+================
+In addition to the standard :doc:`prereqs` for Fabric, this tutorial leverages the Fabric Gateway application API
+for Node. See the `documentation <https://hyperledger.github.io/fabric-gateway/>`_ for a up to date list of supported
+programming language runtimes and dependencies.
 
-In addition to the standard :doc:`prereqs` for Fabric, this tutorial leverages the Hyperledger Fabric SDK for Node.js. See the Node.js SDK `README <https://github.com/hyperledger/fabric-sdk-node#build-and-test>`__ for a up to date list of prerequisites.
+- Ensure you have a suitable version of Node installed. Instructions for installing Node can be found in the `Node.js
+  documentation <https://nodejs.dev/learn/how-to-install-nodejs>`_.
 
-- If you are using macOS, complete the following steps:
+- If you are on Windows, you can use npm to install the `windows-build-tools <https://github.com/felixrieseberg/windows-build-tools#readme>`_
+  which installs all required compilers and tooling by running the following command:
 
-  1. Install `Homebrew <https://brew.sh/>`_.
-  2. Check the Node SDK `prerequisites <https://github.com/hyperledger/fabric-sdk-node#build-and-test>`_ to find out what level of Node to install.
-  3. Run ``brew install node`` to download the latest version of node or choose a specific version, for example: ``brew install node@10`` according to what is supported in the prerequisites.
-  4. Run ``npm install``.
-
-- If you are on Windows,  you can install the `windows-build-tools <https://github.com/felixrieseberg/windows-build-tools#readme>`_ with npm which installs all required compilers and tooling by running the following command:
-
-  .. code:: bash
+  .. code-block:: bash
 
     npm install --global windows-build-tools
 
-- If you are on Linux, you need to install `Python v2.7 <https://www.python.org/download/releases/2.7/>`_, `make <https://www.gnu.org/software/make/>`_, and a C/C++ compiler toolchain such as `GCC <https://gcc.gnu.org/>`_. You can run the following command to install the other tools:
+- If you are on Linux, you need to install `Python v2.7 <https://www.python.org/download/releases/2.7/>`_, `make <https://www.gnu.org/software/make/>`_,
+  and a C/C++ compiler toolchain such as `GCC <https://gcc.gnu.org/>`_. You can run the following command to install
+  these tools:
 
-  .. code:: bash
+  .. code-block:: bash
 
     sudo apt install build-essential
 
+
 Set up the blockchain network
+=============================
+If you've already run through :doc:`test_network` tutorial and have a network up and running, this tutorial will bring
+down your running network before bringing up a new one, to ensure you start with an empty ledger.
+
+
+Launch the blockchain network
 -----------------------------
+Navigate to the ``test-network`` subdirectory within your local clone of the ``fabric-samples`` repository.
 
-If you've already run through :doc:`test_network` tutorial and have a network up
-and running, this tutorial will bring down your running network before
-bringing up a new one.
-
-
-Launch the network
-^^^^^^^^^^^^^^^^^^
-
-.. note:: This tutorial demonstrates the JavaScript versions of the Asset Transfer
-          smart contract and application, but the ``fabric-samples`` repository also
-          contains Go, Java and TypeScript versions of this sample smart contract. To try the
-          Go, Java or TypeScript versions, change the ``javascript`` argument
-          for ``./network.sh deployCC -ccl javascript`` below to either ``go``, ``java`` or ``typescript``
-          and follow the instructions written to the terminal. You may use any chaincode language sample with
-          the javascript application sample (e.g javascript application calling go chaincode functions or
-          javascript application calling typescript chaincode functions, etc.)
-
-Navigate to the ``test-network`` subdirectory within your local clone of the
-``fabric-samples`` repository.
-
-.. code:: bash
+.. code-block:: bash
 
   cd fabric-samples/test-network
 
 If you already have a test network running, bring it down to ensure the environment is clean.
 
-.. code:: bash
+.. code-block:: bash
 
   ./network.sh down
 
 Launch the Fabric test network using the ``network.sh`` shell script.
 
-.. code:: bash
+.. code-block:: bash
 
   ./network.sh up createChannel -c mychannel -ca
 
-This command will deploy the Fabric test network with two peers, an ordering service, and three certificate authorities (Orderer, Org1, Org2).
-Instead of using the cryptogen tool, we bring up the test network using Certificate Authorities,
-hence the ``-ca`` flag. Additionally, the org admin user registration is bootstrapped when the Certificate Authority is started.
-In a later step, we will show how the sample application completes the admin enrollment.
+This command will deploy the Fabric test network with two peers, an ordering service, and three certificate authorities
+(Orderer, Org1, Org2). Instead of using the cryptogen tool, we bring up the test network using certificate authorities,
+hence the ``-ca`` flag. Additionally, the org admin user registration is bootstrapped when the certificate authority is
+started.
 
-Next, let's deploy the chaincode by calling the ``./network.sh`` script with the chaincode name and language options.
 
-.. code:: bash
+Deploy the smart contract
+-------------------------
+.. note:: This tutorial demonstrates the TypeScript versions of the Asset Transfer smart contract and application, but
+          you may use any smart contract language sample with the TypeScript application sample (e.g TypeScript
+          application calling Go smart contract functions or TypeScript application calling Java smart contract
+          functions, etc.). To try the Go or Java versions of the smart contract, change the ``typescript`` argument
+          for the ``./network.sh deployCC -ccl typescript`` command below to either ``go`` or ``java`` and follow the
+          instructions written to the terminal.
 
-  ./network.sh deployCC -ccn basic -ccp ../asset-transfer-basic/chaincode-javascript/ -ccl javascript
+Next, let's deploy the chaincode package containing the smart contract by calling the ``./network.sh`` script with the
+chaincode name and language options.
 
-.. note:: Behind the scenes, this script uses the chaincode lifecycle to package, install,
-          query installed chaincode, approve chaincode for both Org1 and Org2, and finally commit the chaincode.
+.. code-block:: bash
 
-If the chaincode is successfully deployed, the end of the output in your terminal should look similar to below:
+  ./network.sh deployCC -ccn basic -ccp ../asset-transfer-basic/chaincode-typescript/ -ccl typescript
 
-.. code:: bash
+This script uses the chaincode lifecycle to package, install, query installed chaincode, approve chaincode for both
+Org1 and Org2, and finally commit the chaincode.
+
+If the chaincode package is successfully deployed, the end of the output in your terminal should look similar to below:
+
+.. code-block:: text
 
   Committed chaincode definition for chaincode 'basic' on channel 'mychannel':
   Version: 1.0, Sequence: 1, Endorsement Plugin: escc, Validation Plugin: vscc, Approvals: [Org1MSP: true, Org2MSP: true]
-  ===================== Query chaincode definition successful on peer0.org2 on channel 'mychannel' =====================
-
-  ===================== Chaincode initialization is not required =====================
-
-
-Sample application
-^^^^^^^^^^^^^^^^^^
-Next, let's prepare the sample Asset Transfer Javascript application that will be used to interact with the deployed chaincode.
-
-- `JavaScript application <https://github.com/hyperledger/fabric-samples/blob/main/asset-transfer-basic/application-javascript>`__
-
-Note that the sample application is also available in Go and Java at the links below:
-
-- `Go application <https://github.com/hyperledger/fabric-samples/blob/main/asset-transfer-basic/application-go>`__
-- `Java application <https://github.com/hyperledger/fabric-samples/blob/main/asset-transfer-basic/application-java>`__
+  Query chaincode definition successful on peer0.org2 on channel 'mychannel'
+  Chaincode initialization is not required
 
 
-Open a new terminal, and navigate to the ``application-javascript`` folder.
+Prepare the sample application
+------------------------------
+Now, let's prepare the sample Asset Transfer `TypeScript application <https://github.com/hyperledger/fabric-samples/tree/main/asset-transfer-basic/application-gateway-typescript>`_
+that will be used to interact with the deployed smart contract.
 
-.. code:: bash
+Open a new terminal, and navigate to the ``application-gateway-typescript`` directory.
 
-  cd asset-transfer-basic/application-javascript
+.. code-block:: bash
 
-This directory contains sample programs that were developed using the Fabric
-SDK for Node.js. Run the following command to install the application dependencies.
-It may take up to a minute to complete:
+  cd asset-transfer-basic/application-gateway-typescript
 
-.. code:: bash
+This directory contains a sample application developed using the Fabric Gateway application API for Node.
+
+Run the following command to install the dependencies and build the application. It may take some time to complete:
+
+.. code-block:: bash
 
   npm install
 
-This process is installing the key application dependencies defined in the application's
-``package.json``. The most important of which is the ``fabric-network`` Node.js module;
-it enables an application to use identities, wallets, and gateways to connect to
-channels, submit transactions, and wait for notifications. This tutorial also
-uses the ``fabric-ca-client`` module to enroll users with their respective
-certificate authorities, generating a valid identity which is then used by
-the ``fabric-network`` module to interact with the blockchain network.
+This process installs the application dependencies defined in the application's ``package.json``. The most important
+of which is the ``@hyperledger/fabric-gateway`` Node.js package; this provides the Fabric Gateway application API used
+to connect a Fabric Gateway and, using a specific client identity, to submit and evaluate transactions, and receive
+events.
 
 Once ``npm install`` completes, everything is in place to run the application.
-Let's take a look at the sample JavaScript application files we will be using
-in this tutorial. Run the following command to list the files in this directory:
 
-.. code:: bash
+Let's take a look at the sample TypeScript application files we will be using in this tutorial. Run the following
+command to list the files in this directory:
+
+.. code-block:: bash
 
   ls
 
 You should see the following:
 
-.. code:: bash
+.. code-block:: text
 
-  app.js                  node_modules            package.json       package-lock.json
+  dist
+  node_modules
+  package-lock.json
+  package.json
+  src
+  tsconfig.json
 
-.. note:: The first part of the following section involves communication with the Certificate
-          Authority. You may find it useful to stream the CA logs when running
-          the upcoming programs by opening a new terminal shell and running
-          ``docker logs -f ca_org1``.
+The ``src`` directory contains the client application source code. The JavaScript output generated from this source
+code during the install process is located in the ``dist`` directory, and can be ignored.
 
-When we started the Fabric test network back in the first step, an admin user --- literally called ``admin`` ---
-was created as the **registrar** for the Certificate Authority (CA). Our first
-step is to generate the private key, public key, and X.509 certificate for
-``admin`` by having the application call the ``enrollAdmin`` . This process uses a **Certificate
-Signing Request** (CSR) --- the private and public key are first generated
-locally and the public key is then sent to the CA which returns an encoded
-certificate for use by the application. These credentials are then stored
-in the wallet, allowing us to act as an administrator for the CA.
+
+Run the sample application
+==========================
+When we started the Fabric test network earlier in this tutorial, several identities were created using the Certificate
+Authorities. These include a user identity for each of the organizations. The application will use the credentials
+of one of these user identities to transact with the blockchain network.
 
 Let's run the application and then step through each of the interactions with the smart contract functions. From the
-``asset-transfer-basic/application-javascript`` directory, run the following command:
+``asset-transfer-basic/application-gateway-typescript`` directory, run the following command:
 
-.. code:: bash
+.. code-block:: bash
 
-  node app.js
+  npm start
 
 
-First, the application enrolls the admin user
----------------------------------------------
+First, establish a gRPC connection to the Gateway
+-------------------------------------------------
+The client application establishes a `gRPC <https://grpc.io/>`_ connection to the Fabric Gateway service that it will
+use to transact with the blockchain network. To do this, it only requires the Fabric Gateway's endpoint address and, if
+it is configured to use TLS, appropriate TLS certificates. In this sample, the gateway endpoint address is the address
+of a peer, which provides the Fabric Gateway service.
 
-.. note:: It is important to note that enrolling the admin and registering the app user are interactions that
-          take place between the application and the Certificate Authority, not between the application and the chaincode.
-          If you examine the chaincode in ``asset-transfer-basic/chaincode-javascript/lib`` you will find that the chaincode
-          does not contain any functionality that supports enrolling the admin or registering the user.
+.. note:: There is significant overhead associated with establishing gRPC connections, so this connection should be
+          retained by the application and used for all interactions with the Fabric Gateway.
 
-In the sample application code below, you will see that after getting reference to the
-common connection profile path, making sure the connection profile exists, and specifying where to create the wallet,
-``enrollAdmin()`` is executed and the admin credentials are generated from the Certificate Authority.
+.. warning:: In order to maintain security of any private data used in transactions, the application should connect to
+             a Fabric Gateway belonging to the same organization as the client identity. If the client identity's
+             organization does not host any gateways, then a trusted gateway in another organization should be used.
 
-.. code:: bash
+The TypeScript application creates a gRPC connection using the TLS certificate of the signing certificate authority so
+that the authenticity of the gateway's TLS certificate can be verified.
 
-  async function main() {
-    try {
-      // build an in memory object with the network configuration (also known as a connection profile)
-      const ccp = buildCCP();
+For a TLS connection to be successfully established, the endpoint address used by the client must match the address in
+the gateway's TLS certificate. Since the client accesses the gateway's Docker container at a ``localhost`` address, a
+gRPC option is specified to force this endpoint address to be interpreted as the gateway's configured hostname.
 
-      // build an instance of the fabric ca services client based on
-      // the information in the network configuration
-      const caClient = buildCAClient(FabricCAServices, ccp);
+.. code-block:: TypeScript
 
-      // setup the wallet to hold the credentials of the application user
-      const wallet = await buildWallet(Wallets, walletPath);
+  const peerEndpoint = 'localhost:7051';
 
-      // in a real application this would be done on an administrative flow, and only once
-      await enrollAdmin(caClient, wallet);
-
-This command stores the CA administrator's credentials in the ``wallet`` directory.
-You can find administrator's certificate and private key in the ``wallet/admin.id``
-file.
-
-.. note:: If you decide to start over by taking down the network and bringing it back up again, you will
-          have to delete the ``wallet`` folder and its identities prior to re-running the javascript application
-          or you will get an error. This happens because the Certificate Authority and its database are taken down
-          when the test-network is taken down but the original wallet still remains in the application-javascript directory
-          so it must be deleted.  When you re-run the sample javascript application, a new wallet and credentials will
-          be generated.
-
-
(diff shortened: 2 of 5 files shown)


docs/source/getting_started.rst
```diff
@@ -21,7 +21,7 @@ The Fabric application stack has five layers:
 * :doc:`Prerequisite software <prereqs>`: the base layer needed to run the software, for example, Docker.
 * :doc:`Fabric and Fabric samples <install>`: the Fabric executables to run a Fabric network along with sample code.
 * :doc:`Contract APIs <sdk_chaincode>`: to develop smart contracts executed on a Fabric Network.
-* :doc:`Application SDKs <sdk_chaincode>`: to develop your blockchain application.
+* :doc:`Application APIs <sdk_chaincode>`: to develop your blockchain application.
 * The Application: your blockchain application will utilize the Application SDKs to call smart contracts running on a Fabric network.
```
joshhus, line 25: hmmm, technically, doesn't the Fabric Gateway running on the peers "call smart contracts"? (not the SDKs?) ... (albeit the gateway SDKs are used to develop the app - which specifies which chaincode to call.

docs/source/getting_started.rst
```diff
@@ -21,7 +21,7 @@ The Fabric application stack has five layers:
 * :doc:`Prerequisite software <prereqs>`: the base layer needed to run the software, for example, Docker.
 * :doc:`Fabric and Fabric samples <install>`: the Fabric executables to run a Fabric network along with sample code.
 * :doc:`Contract APIs <sdk_chaincode>`: to develop smart contracts executed on a Fabric Network.
-* :doc:`Application SDKs <sdk_chaincode>`: to develop your blockchain application.
+* :doc:`Application APIs <sdk_chaincode>`: to develop your blockchain application.
```
joshhus, line 24: hmmm, technically, doesn't the Fabric Gateway running on the peers "call smart contracts"? (not the SDKs?) ... (albeit the gateway SDKs are used to develop the app - which specifies which chaincode to call.

docs/source/getting_started.rst
joshhus, line 25: hmmm, technically, doesn't the Fabric Gateway running on the peers "call smart contracts"? (not the SDKs?) ... (albeit the gateway SDKs are used to develop the app - which specifies which chaincode to call.

docs/source/write_first_app.rst
```diff
@@ -101,889 +97,485 @@ bringing up a new one.
...
-
-Second, the application registers and enrolls an application user
------------------------------------------------------------------
+First, the application creates a gRPC connection to the Gateway peer
+--------------------------------------------------------------------
 
-Now that we have the administrator's credentials in a wallet, the application uses the ``admin``
-user to register and enroll an app user which will be used
-to interact with the blockchain network. The section of the application code is shown below.
+The client application establishes a `gRPC <https://grpc.io/>`__ connection to the Gateway peer that it will use to
```
joshhus: so in markdown you can publish an active link like this: 
[gRPC}(https://grpc.io/)

joshhus: so for capitalization, terminology, and punctuation we are using any of the following: 
"Fabric Gateway"
"the Fabric Gateway service"
"the gateway" 
"peer" 
"target peer" 
We are not using "gateway peer" because all peers are gateway peers in 2.4
So in line 214 e.g., delete Gateway. And here the peer that gets the connection is the "target peer". ... so "connection to the target peer"

docs/source/write_first_app.rst
```diff
@@ -101,889 +97,485 @@ bringing up a new one.
...
+transact with the network. To do this it only requires the endpoint address of the Gateway peer and, if the Gateway is
```
joshhus: do you want to say "with the channeL" here instead of "with the network"? technically it's the channel right, which may not be the entire network of channels.

joshhus: and also "target peer and, if the gateway is ..."

docs/source/write_first_app.rst
```diff
@@ -101,889 +97,485 @@ bringing up a new one.
...
+configured to use TLS, appropriate TLS certificates to establish a secure connection. There is significant overhead
+associated with establishing gRPC connections, so this connection should be retained by the application and shared by
+all Gateway connections to the Gateway peer.
+
+.. note:: In order to maintain security of any private data used in transactions, the application should connect to a
```
joshhus: suggest changing "note:" to "Attention:"

docs/source/write_first_app.rst
```diff
@@ -101,889 +97,485 @@ bringing up a new one.
...
+          Gateway peer belonging to the same organization as the client identity. If the client identity's organization
+          does not host any peers then a Gateway peer hosted by an upstream organization through which the client's
```
joshhus: peers, 

(add comma)

joshhus: dyk if "upstream" is used in docs? I can check.
Wondering if this is correct/works: ... " ... does not host any peers, then a trusted peer in another organization can be used."

docs/source/write_first_app.rst
```diff
@@ -101,889 +97,485 @@ bringing up a new one.
...
+          organization is authorized to transact should be used.
+
+The TypeScript application creates a gRPC connection using the TLS certificate of the Certificate Authority for the
```
joshhus: "The TypeScript application creates a gRPC connection using the TLS certificate from the target peer organization's Certificate Authority to verify the authenticity of the peer's credentials."
(@bestbeforetoday this may need work but you get the idea to try to condense it.)

docs/source/write_first_app.rst
```diff
@@ -101,889 +97,485 @@ bringing up a new one.
...
+Gateway peer's organization, so that the authenticity of the peer's credentials can be verified. Since the Docker
+container hosting the Gateway peer is reached at a ``localhost`` address, a gRPC option is specified to force this
```
joshhus: Let's try to rephrase / simplify this long sentence. Maybe use 2 sentences or a semicolon;

docs/source/write_first_app.rst
```diff
@@ -101,889 +97,485 @@ bringing up a new one.
...
+  async function newGrpcConnection(): Promise<grpc.Client> {
+      const tlsRootCert = await fs.readFile(tlsCertPath);
+      const tlsCredentials = grpc.credentials.createSsl(tlsRootCert);
+      return new grpc.Client(peerEndpoint, tlsCredentials, {
+          'grpc.ssl_target_name_override': 'peer0.org1.example.com',
+      });
+  }
 
-.. code:: bash
+Second, the application creates a Gateway connection
```
joshhus: Is this step 2 / section heading maybe.

docs/source/write_first_app.rst
```diff
@@ -101,889 +97,485 @@ bringing up a new one.
...
+----------------------------------------------------
 
-  // in a real application this would be done only when a new user was required to be added
-  // and would be part of an administrative flow
-  await registerAndEnrollUser(caClient, wallet, mspOrg1, org1UserId, 'org1.department1');
+The application then creates a Gateway connection, which the application uses to access any of the networks
+(analogous to channels) accessible to the Gateway peer, and smart contracts deployed to those networks. A Gateway
```
joshhus: I think of a channel as more of a "sub-network" in the single overall network. 
Is this connection different than the gRPC connection above? 
Use "Fabric Gateway connection" first instance in a section, and then "gateway connection" following.

docs/source/write_first_app.rst
```diff
@@ -101,889 +97,485 @@ bringing up a new one.
...
+connection requires three things:
```
joshhus: connection has three requirements:

docs/source/write_first_app.rst
```diff
@@ -101,889 +97,485 @@ bringing up a new one.
...
-store its credentials alongside those of ``admin`` in the wallet. We now have
-identities for two separate users --- ``admin`` and ``appUser`` --- that can be
-used by our application.
+  1. gRPC connection to the Gateway peer.
+  2. Client identity used to transact with the network.
+  3. Signing implementation used to generate digital signatures for the client identity.
 
-Scrolling further down in your terminal output, you should see confirmation of the app user registration
-similar to this:
+In addition to the above, a hash implementation can also be specified. The default implementation will typically be
```
joshhus: delete "to the above"

docs/source/write_first_app.rst
```diff
@@ -101,889 +97,485 @@ bringing up a new one.
...
+sufficient.
 
-.. code:: bash
+The sample application uses an X.509 certificate for the Org1 user identity, and a signing implementation based on the
```
joshhus: is this "user" identity for the application?

docs/source/write_first_app.rst
```diff
@@ -101,889 +97,485 @@ bringing up a new one.
...
+corresponding private key. These credentials are created using the Org1 Certificate Authority when the test network
```
joshhus: why "test" ?

docs/source/write_first_app.rst
```diff
@@ -101,889 +97,485 @@ bringing up a new one.
...
+is started.
```
joshhus: So needs something like "The following code snippet ... " is what, or does what.

docs/source/write_first_app.rst
```diff
@@ -101,889 +97,485 @@ bringing up a new one.
...
-You will notice that in the following lines of application code, the application is getting reference
-to the Contract using the contract name and channel name via Gateway:
+  async function newSigner(): Promise<Signer> {
+      const privateKeyPem = await fs.readFile(keyPath);
+      const privateKey = crypto.createPrivateKey(privateKeyPem);
+      return signers.newPrivateKeySigner(privateKey);
+  }
 
-.. code:: bash
+Third, the application accesses the smart contract to be invoked
```
joshhus: Okay so this is the application flow right, for gateway. I think there are 3 phases in the other gateway pages. Are these related to your 3 steps?

docs/source/write_first_app.rst
```diff
@@ -101,889 +97,485 @@ bringing up a new one.
...
+----------------------------------------------------------------
 
-  // Create a new gateway instance for interacting with the fabric network.
-  // In a real application this would be done as the backend server session is setup for
-  // a user that has been verified.
-  const gateway = new Gateway();
+The sample application uses the Gateway connection to get a reference to the Network and then the default Contract
```
joshhus: so docs equate "chaincode" with "smart contract" elsewhere, but favor "chaincode". I guess you're saying the app references the smart contract within the chaincode, not the chaincode per se.

docs/source/write_first_app.rst
```diff
@@ -101,889 +97,485 @@ bringing up a new one.
...
+call you can specify both the name of the chaincode and a specific smart contract to target. For example:
 
-When a chaincode package includes multiple smart contracts, on the `getContract() API <https://hyperledger.github.io/fabric-sdk-node/release-2.2/module-fabric-network.Network.html#getContract>`__ you can specify both the name of the chaincode package and a specific smart contract to target. For example:
+.. code-block:: TypeScript
 
-.. code:: bash
+  const contract = network.getContract('chaincodeName', 'smartContractName');
 
-  const contract = await network.getContract('chaincodeName', 'smartContractName');
+Fourth, the application initializes the ledger with some sample assets
```
joshhus: ok, what does it mean to "initialize" the ledger? 
Are we making presumptions about the network that the user would / could run this on - if so let's be clear about what those are at the top / beginning. Maybe I missed it.

docs/source/write_first_app.rst
```diff
@@ -101,889 +97,485 @@ bringing up a new one.
...
+----------------------------------------------------------------------
 
-Fourth, the application initializes the ledger with some sample data
---------------------------------------------------------------------
+Now that we are at the point where we are actually invoking smart contract transactions, let’s
```
joshhus: How about ... 
"We are now at the point of invoking smart contract transactions. Let's look at them in sequence."  ... or similar.

docs/source/write_first_app.rst
```diff
@@ -101,889 +97,485 @@ bringing up a new one.
...
+go through them in sequence.
 
-Now that we are at the point where we are actually having the sample application submit transactions, let’s
-go through them in sequence. The application code snippets and invoked chaincode snippets
-are provided for each called function, as well as the terminal output.
+Immediately after initial deployment of the chaincode, the ledger is empty. The application uses ``submitTransaction()``
```
joshhus: Okay so this example "initiates" a brand new channel it seems. First transaction sent to the ledger right. Lets get some language at the top start that explains this full scenario and the e2e. If it's not there yet.

docs/source/write_first_app.rst
```diff
@@ -101,889 +97,485 @@ bringing up a new one.
...
+to invoke the chaincode ``InitLedger`` transaction function, which populates the ledger with some sample data. Behind
+the scenes, ``submitTransaction()`` will use the Gateway peer to:
 
-The submitTransaction() function is used to invoke the chaincode ``InitLedger`` function to populate the
-ledger with some sample data. Under the covers, the submitTransaction() function will use service discovery
-to find a set of required endorsing peers for the chaincode, invoke the chaincode
-on the required number of peers, gather the chaincode endorsed results from those peers,
-and finally submit the transaction to the ordering service.
+  1. Endorse the transaction proposal by successfully executing the transaction function on sufficient endorsing peers
```
joshhus: Okay so these steps we have in the peers page I think. Let's make sure they match / are in sync.

docs/source/write_first_app.rst
```diff
@@ -101,889 +97,485 @@ bringing up a new one.
...
-
-        for (const asset of assets) {
-            asset.docType = 'asset';
-            await ctx.stub.putState(asset.ID, Buffer.from(JSON.stringify(asset)));
-            console.info(`Asset ${asset.ID} initialized`);
-        }
-    }
 
-The terminal output entry should look similar to below:
+An initialization function to set up initial ledger state would typically be called only once after a smart contract
```
joshhus: meaning any smart contract is called, the first time this happens, for any smart contract on the channel. ... as opposed to after each s.c. is called for its first invocation.

docs/source/write_first_app.rst
```diff
@@ -101,889 +97,485 @@ bringing up a new one.
...
+is deployed for the first time. A new version of the smart contract would likely not require the initialization to be
```
joshhus: oh deployed. that's different than invoked right.

docs/source/write_first_app.rst
```diff
@@ -101,889 +97,485 @@ bringing up a new one.
...
+performed again. In this case it is good practice for the transaction function to check and perhaps also record some
```
joshhus: tx function being the smart contract code, within the chaincode right.

docs/source/write_first_app.rst
```diff
@@ -101,889 +97,485 @@ bringing up a new one.
...
---------------------------------------------------------------
+Fifth, the application invokes transaction functions to query and modify assets
+-------------------------------------------------------------------------------
 
 First, a word about querying the ledger.
 
-Each peer in a blockchain network hosts a copy of the `ledger <./ledger/ledger.html>`_. An application
-program can view the most recent data from the ledger using read-only invocations of
-a smart contract running on your peers called a query.
+Each peer in a blockchain network hosts a copy of the `ledger <./ledger/ledger.html>`_. An application can query the
```
joshhus: "channel," if we agree to not equate the entire network with a channel. ...

docs/source/write_first_app.rst
```diff
@@ -101,889 +97,485 @@ bringing up a new one.
...
-The most common queries involve the current values of data in the ledger -- its
-`world state <./ledger/ledger.html#world-state>`_. The world state is
-represented as a set of key-value pairs, and applications can query data for a
-single key or multiple keys. Moreover, you can use complex queries to read the
-data on the ledger when you use CouchDB as your state database and model your data in JSON.
-This can be very helpful when looking for all assets that match certain keywords
-with particular values; all assets with a particular owner, for example.
+The application uses ``evaluateTransaction()`` to query the ledger by performing a read-only transaction function
+invocation. Behind the scenes, ``evaluateTransaction()`` will use the Gateway peer to execute the transaction function
+on a suitable peer and return its result. No transaction is sent to the ordering service and no ledger update occurs.
```
joshhus: what makes a peer suitable, we could state that instead.

docs/source/write_first_app.rst
```diff
@@ -101,889 +97,485 @@ bringing up a new one.
...
+      assetId,
+      'yellow',
+      '5',
+      'Tom',
+      '1300',
+  );
 
 .. note:: In the application and chaincode snippets above, it is important
-          to note that the sample application submits the ``'CreateAsset'``
+          to note that the sample application submits the ``CreateAsset``
```
joshhus: oh so you have the special .md syntax for a note? cool.

docs/source/write_first_app.rst
```diff
@@ -101,889 +97,485 @@ bringing up a new one.
...
-Chaincode ``'ReadAsset'`` function
-
-  .. code:: bash
+          is expecting, and in the correct sequence. In this case the correctly sequenced arguments are:
+          
+          .. code-block:: text
+          
+            assetId, "yellow", "5", "Tom", "1300"
+          
+          because the corresponding chaincode ``CreateAsset`` is expecting the following sequence of arguments that
```
joshhus: ok we'll make this a complete sentence. Can just delete "because" and use "The corresponding ... "

docs/source/write_first_app.rst
```diff
@@ -101,889 +97,485 @@ bringing up a new one.
...
+          define the asset object:
+          
+          .. code-block:: text
+
+            ID, Color, Size, Owner, AppraisedValue
+          
+          Similar to transaction function return values, any data type can be converted to bytes and passed as
```
joshhus: Section headings might help us - at this point I've read a lot of text since the beginning and am not sure where I am.

docs/source/write_first_app.rst
```diff
@@ -101,889 +97,485 @@ bringing up a new one.
...
+          transaction arguments, provided those arguments match what is expected by the transaction function. Since
+          strings are commonly used, the application API accepts UTF-8 strings as transaction arguments and converts
+          them to bytes internally.
+
+Next, the sample application submits a transaction to transfer ownership of the newly created asset. This time
+the transaction is invoked using ``submitAsync()``, which returns after successfully submitting the endorsed
```
joshhus, line 426: note first time "endorsed" mentioned; may need to add it earlier.

docs/source/write_first_app.rst
```diff
@@ -101,889 +97,485 @@ bringing up a new one.
...
+transaction to the ordering service instead of blocking until the transaction is committed to the ledger. This allows
```
joshhus: ok, what is blocking. packaging in a block perhaps.

docs/source/write_first_app.rst
```diff
@@ -101,889 +97,485 @@ bringing up a new one.
...
 
+.. code-block:: text
 
-In the next part of the sequence, the sample application evaluates to see if
-``asset1`` exists, which will return a boolean value of true, because we populated
-the ledger with ``asset1`` when we initialized the ledger with assets. You may recall
-that the original appraised value of ``asset1`` was ``300``.
-The application then submits a transaction to update ``asset1`` with a new appraised value, and
-then immediately evaluates to read ``asset1`` from the ledger to show the new appraised value of ``350``.
+  *** Successfully submitted transaction to transfer ownership from Tom to Saptha
```
joshhus, line 453: hmmm where are these names / IDs defined.

docs/source/write_first_app.rst
```diff
@@ -101,889 +97,485 @@ bringing up a new one.
...
+      Color: 'yellow',
+      ID: 'asset1639084597466',
+      Owner: 'Saptha',
+      Size: 5
   }
 
-Terminal Output:
-
-.. code:: bash
+The final part of the sequence demonstrates an error submitting a transaction. In this example, the application attempts
```
joshhus: Should we mention "the gateway service" here as a player.

docs/source/write_first_app.rst
```diff
@@ -101,889 +97,485 @@ bringing up a new one.
...
+to submit an ``UpdateAsset`` transaction but specifies an asset ID that does not exist. The smart contract transaction
+function returns an error response, and the ``submitTransaction()`` call fails.
 
-  Evaluate Transaction: AssetExists, function returns "true" if an asset with given assetID exist
-  Result: true
+A ``submitTransaction()`` failure may generate several different types of error, indicating the point in the submit flow
+that the error occurred, and containing additional information to enable the application to respond appropriately.
+Consult the `API documentation <https://hyperledger.github.io/fabric-gateway/main/api/node/interfaces/Contract.html#submitTransaction>`_
```
joshhus, line 492: [link](link url) should work

docs/source/write_first_app.rst
```diff
@@ -1006,8 +598,7 @@ Additional resources
 
 As we said in the introduction, we have a whole section on
```
joshhus: I'll want to rewrite this one. A lot of words saying less.

docs/source/write_first_app.rst
```diff
@@ -997,7 +589,7 @@ Summary
 Now that we’ve seen how the sample application and chaincode are written and how they interact with each other, you should have a pretty
 good sense of how applications interact with a blockchain network using a smart
 contract to query or update the ledger. You’ve seen the basics of the roles
-smart contracts, APIs, and the SDK play in queries and updates and you should
+smart contracts, contract APIs, and application APIs play in queries and updates, and you should
```
joshhus: hmmm "contract" APIs immediately after smart contracts.

docs/source/write_first_app.rst
```diff
@@ -997,7 +589,7 @@ Summary
...
 have a feel for how different kinds of applications could be used to perform
```
joshhus: "can" not "could"

docs/source/write_first_app.rst
bestbeforetoday: This isn't a markdown file.

All peers are Gateway peers by default, but the Gateway service can be disabled on any peer. **But** I'm happy referring to Fabric Gateway.

docs/source/write_first_app.rst
bestbeforetoday: This is not a markdown file.

docs/source/write_first_app.rst
bestbeforetoday: This is all within a **note**. I'm not sure if headings within a note work.

docs/source/write_first_app.rst
bestbeforetoday: https://stackoverflow.com/questions/2407589/what-does-the-term-blocking-mean-in-programming

docs/source/write_first_app.rst
bestbeforetoday: See the code snippet

docs/source/write_first_app.rst
bestbeforetoday: This is not a markdown document.

docs/source/write_first_app.rst
bestbeforetoday: They are not the same thing, like golf and golf clubs both contain "golf" but are not the same thing. The original text used "API" to mean "contract API" and "SDK" to mean "application API". They are both APIs. I'm not keen on the section in general though, so perhaps I can rework to make us both happier.

docs/source/write_first_app.rst
bestbeforetoday: We really shouldn't. That's an implementation detail that is intended to be opaque to the client and may change over time. Instead perhaps I'll just list the key steps with less detail, since the detail is covered elsewhere.

docs/source/write_first_app.rst
bestbeforetoday: I don't see how "channel" can be correct here.

joshhus: You can check with others (@denyeart), but my view is that one blockchain "network" can contain multiple channels within it (essentially, sub-networks). In which case, each channel would host a copy of the ledger for that channel, but would not host a copy of the ledger for the whole network, as such a single thing doesn't exist. So it's just a semantics decision.

bestbeforetoday: I agree with the first sentence, although the terminology does not match that of the APIs, which this tutorial is describing. The rest makes no sense to me. The channel is not a physical thing so it can’t hold a copy of any data. Peers are physical things, and each one has a complete copy of the ledger stored on physical storage for each of the channels they belong to.

Please note that This text was already in the tutorial before my PR, and my PR has since removed this text.

docs/source/getting_started.rst
bestbeforetoday: Tackle any changes you want there in a separate PR.

docs/source/write_first_app.rst
bestbeforetoday: No, it doesn't restrict the client to interacting with a single channel.

docs/source/write_first_app.rst
bestbeforetoday: Changed to warning since that is the only option other than note.

docs/source/write_first_app.rst
bestbeforetoday: It is different to a gRPC connection. I've reworded to (hopefully) make it clearer.

docs/source/write_first_app.rst
bestbeforetoday: Yes, re-worded to clarify

docs/source/write_first_app.rst
bestbeforetoday: Because the Fabric sample network is called "test network"

docs/source/write_first_app.rst
bestbeforetoday: No idea. These are the steps from the application perspective when using the client API.

docs/source/write_first_app.rst
bestbeforetoday: There was a big discussion about this several years ago and a push to move the terminology away from the Fabric implementation-centric language to something that make more sense for end users. So the terminology of the Fabric programming model, exposed in the client APIs since v1.4, is Network (corresponding to channel) and Smart Contract. A chaincode package is what you deploy, and that package may contain multiple smart contracts. Often chaincodes only contain a single, default smart contract, in which case there is a one-to-one relationship between chaincode and smart contract. Partly because of this and partly for historic reasons, chaincode and smart contract often get used interchangeably. The world state is namespaced at the chaincode level, so all smart contracts within a given chaincode share the same world state.

docs/source/write_first_app.rst
bestbeforetoday: It doesn't mean anything beyond what is described for this transaction invocation. I've re-worded to try to make it clearer.

docs/source/write_first_app.rst
bestbeforetoday: Removed.

docs/source/write_first_app.rst
bestbeforetoday: No, absolutely not. Your question makes it sound like you have a fundamental mis-understanding about what a channel is.

docs/source/write_first_app.rst
bestbeforetoday: These describe the steps from the application perspective. Simplified description to try to avoid confusion.

docs/source/write_first_app.rst
bestbeforetoday: Deployed and invoked are not the same thing.

docs/source/write_first_app.rst
bestbeforetoday: Removed the second piece of information to avoid confusion.

docs/source/write_first_app.rst
bestbeforetoday: Changed the wording to make it clearer for non-programmers.

docs/source/write_first_app.rst
bestbeforetoday: Removed.

docs/source/write_first_app.rst
```diff
@@ -1,1014 +1,572 @@
...
-          smart contracts, check out our
-          :doc:`developapps/developing_applications` section or the
-          :doc:`tutorial/commercial_paper`.
+          smart contracts, check out our :doc:`developapps/developing_applications` section.
 
 This tutorial provides an introduction to how Fabric applications interact
 with deployed blockchain networks. The tutorial uses sample programs built using the
-Fabric SDKs -- described in detail in the :doc:`/developapps/application` topic --
-to invoke a smart contract which queries and updates the ledger with the smart
+Fabric Gateway application API -- described in detail in the :doc:`/developapps/application` topic --
```
denyeart: The Gateway API is not described in detail in Developing Applications (at this point), so I'd remove these links. I think we can just keep the link in the Note at the top of the topic.

docs/source/write_first_app.rst
```diff
@@ -1,1014 +1,572 @@
...
-We will also use our sample programs and a deployed Certificate Authority to generate
-the X.509 certificates that an application needs to interact with a permissioned
-blockchain.
 
 **About Asset Transfer**
 
-This Asset Transfer (basic) sample demonstrates how to initialize a ledger with assets, query those assets, create
-a new asset, query a single asset based on an asset ID, update an existing asset, and transfer an asset to a new owner.
-It involves the following two components:
+The Asset Transfer (basic) sample demonstrates how to initialize a ledger with assets, then query, create and update
```
denyeart:
```suggestion
The Asset Transfer (basic) sample demonstrates how to create, update, and query assets.
```

docs/source/write_first_app.rst
```diff
@@ -1,1014 +1,572 @@
...
+  contract for our application.
 
   .. image:: images/AppConceptsOverview.png
 
-  **2. Explore a sample smart contract.**
-  We’ll inspect the sample assetTransfer (javascript) smart contract to learn about the transactions within it,
-  and how they are used by an application to query and update the ledger.
+  2. **Run the sample application to interact with the smart contract.**
+  Our application will use the assetTransfer smart contract to create, query, and update assets on the ledger. We will
+  step through the code of the application and the transactions it invokes, including initializing the ledger with
```
denyeart:
```suggestion
  step through the code of the application and the transactions it invokes, including creating some initial
```

docs/source/write_first_app.rst
```diff
@@ -1,1014 +1,572 @@
...
 
-  .. code:: bash
+  .. code-block:: bash
 
     npm install --global windows-build-tools
 
-- If you are on Linux, you need to install `Python v2.7 <https://www.python.org/download/releases/2.7/>`_, `make <https://www.gnu.org/software/make/>`_, and a C/C++ compiler toolchain such as `GCC <https://gcc.gnu.org/>`_. You can run the following command to install the other tools:
+- If you are on Linux, you need to install `Python v2.7 <https://www.python.org/download/releases/2.7/>`_, `make <https://www.gnu.org/software/make/>`_,
+  and a C/C++ compiler toolchain such as `GCC <https://gcc.gnu.org/>`_. You can run the following command to install
+  the other tools:
```
denyeart:
```suggestion
  these tools:
```

docs/source/write_first_app.rst
```diff
@@ -1,1014 +1,572 @@
...
-ledger of a blockchain network.
 
 Before you begin
-----------------
-
-In addition to the standard :doc:`prereqs` for Fabric, this tutorial leverages the Hyperledger Fabric SDK for Node.js. See the Node.js SDK `README <https://github.com/hyperledger/fabric-sdk-node#build-and-test>`__ for a up to date list of prerequisites.
-
-- If you are using macOS, complete the following steps:
+================
+In addition to the standard :doc:`prereqs` for Fabric, this tutorial leverages the Hyperledger Fabric client API for Node.
```
denyeart: Should we call it "Fabric Gateway application API" consistently?

docs/source/write_first_app.rst
```diff
@@ -1,1014 +1,572 @@
...
-If you scroll back up to the beginning of the output in your terminal, it should be similar to below:
+  npm start
 
-.. code:: bash
 
-  Wallet path: /Users/<your_username>/fabric-samples/asset-transfer-basic/application-javascript/wallet
-  Successfully enrolled admin user and imported it into the wallet
+First, establish a gRPC connection to the Gateway
+-------------------------------------------------
+The client application establishes a `gRPC <https://grpc.io/>`_ connection to the Fabric Gateway that it will use to
```
denyeart:
```suggestion
The client application establishes a `gRPC <https://grpc.io/>`_ connection to the Fabric Gateway service running on a peer that it will use to
```

docs/source/write_first_app.rst
```diff
@@ -1,1014 +1,572 @@
...
+.. code-block:: bash
 
-If you scroll back up to the beginning of the output in your terminal, it should be similar to below:
+  npm start
 
-.. code:: bash
 
-  Wallet path: /Users/<your_username>/fabric-samples/asset-transfer-basic/application-javascript/wallet
-  Successfully enrolled admin user and imported it into the wallet
+First, establish a gRPC connection to the Gateway
```
denyeart, line 199:
```suggestion
First, establish a gRPC connection to the Gateway service on a peer
```

I think we need to make it clear that Fabric Gateway runs on a peer. After the first couple mentions, we can simply call it "Gateway" in later references. We did the same thing for the peer topic.

docs/source/write_first_app.rst
```diff
@@ -1,1014 +1,572 @@
...
+connection has three requirements:
 
-Scrolling further down in your terminal output, you should see confirmation of the app user registration
-similar to this:
+  1. gRPC connection to the Fabric Gateway.
+  2. Client identity used to transact with the network.
+  3. Signing implementation used to generate digital signatures for the client identity.
 
-.. code:: bash
+A hash implementation can also be specified. The default implementation will typically be sufficient.
```
denyeart: I don't know what this means. Either remove it or explain it.

docs/source/write_first_app.rst
```diff
@@ -1,1014 +1,572 @@
...
+Immediately after initial deployment of the chaincode package, the ledger is empty. The application uses
+``submitTransaction()`` to invoke the ``InitLedger`` transaction function, which populates the ledger with some sample
+data. ``submitTransaction()`` will use the Gateway to:
 
-The submitTransaction() function is used to invoke the chaincode ``InitLedger`` function to populate the
-ledger with some sample data. Under the covers, the submitTransaction() function will use service discovery
-to find a set of required endorsing peers for the chaincode, invoke the chaincode
-on the required number of peers, gather the chaincode endorsed results from those peers,
-and finally submit the transaction to the ordering service.
+  1. Endorse the transaction proposal.
```
denyeart, line 296: I think it makes sense to explain this at a high level (one paragraph), even though it is described in other key concepts topics in more detail.
e.g. something at this level of detail "submitTransaction() will request the targeted Gateway peer to execute the chaincode and gather endorsements from the required number of peers to meet the endorsement policy (in our case the Org1 gateway peer as well as the Org2 peer), then will sign the transaction."

docs/source/write_first_app.rst
```diff
@@ -1,1014 +1,572 @@
...
---------------------------------------------------------------------
 
-Now that we are at the point where we are actually having the sample application submit transactions, let’s
-go through them in sequence. The application code snippets and invoked chaincode snippets
-are provided for each called function, as well as the terminal output.
+Fourth, populate the ledger with sample assets
+----------------------------------------------
+Immediately after initial deployment of the chaincode package, the ledger is empty. The application uses
+``submitTransaction()`` to invoke the ``InitLedger`` transaction function, which populates the ledger with some sample
+data. ``submitTransaction()`` will use the Gateway to:
```
denyeart:
```suggestion
assets. ``submitTransaction()`` will use the Gateway to:
```

docs/source/write_first_app.rst
```diff
@@ -1,1014 +1,572 @@
...
-First, a word about querying the ledger.
 
-Each peer in a blockchain network hosts a copy of the `ledger <./ledger/ledger.html>`_. An application
-program can view the most recent data from the ledger using read-only invocations of
-a smart contract running on your peers called a query.
 
-Here is a simplified representation of how a query works:
+Fifth, invoke transaction functions to read and write assets
+------------------------------------------------------------
+Now the application is ready to execute business logic that queries, creates and modifies assets on the ledger by
```
denyeart:
```suggestion
Now the application is ready to execute business logic that queries, creates additional assets, and modifies assets on the ledger by
```

docs/source/write_first_app.rst
```diff
@@ -1,1014 +1,572 @@
...
+invoking transactions functions on the smart contract.
 
-.. image:: tutorial/write_first_app.diagram.1.png
+Query all assets
+~~~~~~~~~~~~~~~~
+The application uses ``evaluateTransaction()`` to query the ledger by performing a read-only transaction invocation.
+``evaluateTransaction()`` will use the Gateway to execute the transaction function and return its result. The
```
denyeart: "execute the transaction" sounds like we are submitting a transaction. Should we instead say something like "invoke the chaincode to query the ledger"?

docs/source/write_first_app.rst
bestbeforetoday: Talking about peers raised comments before so I'm avoiding it wherever possible.

docs/source/write_first_app.rst
bestbeforetoday: As before, discussing peers was confusing so it's avoided.

docs/source/write_first_app.rst
bestbeforetoday: It was described at a high level previously but was deemed confusing, so avoided.

docs/source/write_first_app.rst
bestbeforetoday: That doesn't match the API documentation, which consistently uses "client API", but we can consistently call it "application API" in the docs to better distinguish it from the contract API.

docs/source/write_first_app.rst
bestbeforetoday: Changing "execute" to "invoke". The previous sentence describes that this is a read-only transaction invocation to query the ledger. The APIs and this tutorial use smart contracts, not chaincode.

docs/source/write_first_app.rst
denyeart: Actually it's this reference I suggested removing since the Gateway API is not "described in detail" in Developing Applications.

docs/source/write_first_app.rst
denyeart: @joshhus What is the concern? I think it is more confusing to avoid using the term peer. For the peer topic, the resolution was to use language like "Gateway service running on a peer". If we don't do some kind of clarification like this, people will get the impression that the Gateway is its own runtime component. Or worse, might think we're talking about the Gateway within the client SDK since that is what it used to be called.

docs/source/write_first_app.rst
denyeart: Same

docs/source/write_first_app.rst
denyeart: @joshhus what was the concern? A sentence or two would definitely help here. Feel free to propose an edit to what I suggested.

joshhus: The discussion here was that it initially read as if an "endorsement" by one peer / org then leads to a commitment to the ledger. Without a full understanding of endorsement and endorsement policies, that sounded potentially misleading to me. We discussed this when reviewing the other topics also - endorsement by one org does not mean the Tx passes endorsement (policies). So generally a semantics issue of being imprecise with the word "endorsed" in different contexts.

joshhus:
The discussion around "gateway peer" - I don't see the comment here any more - was only that "gateway" to modify peer was deemed redundant in this context (in our prior reviews), because all peers are now "gateway" peers. So not to avoid using "peer" but to avoid using "gateway peer". ... So in prior docs we use "peer" when that is important - such as in getting a connection from the client app. ... When it's the "Fabric Gateway service" specifically (running on the peer) that is doing things, then we call it that, as opposed to a "peer" doing things.

docs/source/write_first_app.rst
bestbeforetoday: Actually the comments before were on whether the brief description of executing the transaction on peers matched the description of the endorsement process we already have in other pages.

My intent with this list of steps is to signpost the following application flow, which is what gets executed when you call `submitTransaction()`, not to describe the transaction flow at the server side:
- [endorse()](https://hyperledger.github.io/fabric-gateway/main/api/node/interfaces/Proposal.html#endorse)
- [submit()](https://hyperledger.github.io/fabric-gateway/main/api/node/interfaces/Transaction.html#submit)
- [getStatus()](https://hyperledger.github.io/fabric-gateway/main/api/node/interfaces/SubmittedTransaction.html#getStatus) to wait for the transaction commit

Understanding the application API flow is important for interpreting error types that may be generated by that flow, touched on in the last step.

docs/source/write_first_app.rst
bestbeforetoday: My bad, removed this reference too.

bestbeforetoday:
> The discussion around "gateway peer" - I don't see the comment here any more - was only that "gateway" to modify peer was deemed redundant in this context (in our prior reviews), because all peers are now "gateway" peers. So not to avoid using "peer" but to avoid using "gateway peer". ... So in prior docs we use "peer" when that is important - such as in getting a connection from the client app. ... When it's the "Fabric Gateway service" specifically (running on the peer) that is doing things, then we call it that, as opposed to a "peer" doing things.

The client app only interacts with the Gateway service. Whether that Gateway service runs on a peer is irrelevant to the client application. There was a point in time where the Gateway service was implemented in a separate component, not in a peer, and this could be the case again in the future. It really makes no difference from the client application perspective discussed in this tutorial.

joshhus:
Maybe 3 sections would have been clearer. To consider another time without a deadline -- 
A. Here's how to do the tutorial, only. (task)
B. Here is the Tx flow and other concepts that are revealed in the tutorial. (conceptual). If you want to read about that.
C. Here are the overall Fabric Gateway topics (links) and fabric topics (links) that apply to the tutorial (reference). If you want to go read that. ... (Note reference could also be presented as prereq, before the task.)

denyeart:
> 



> > The discussion around "gateway peer" - I don't see the comment here any more - was only that "gateway" to modify peer was deemed redundant in this context (in our prior reviews), because all peers are now "gateway" peers. So not to avoid using "peer" but to avoid using "gateway peer". ... So in prior docs we use "peer" when that is important - such as in getting a connection from the client app. ... When it's the "Fabric Gateway service" specifically (running on the peer) that is doing things, then we call it that, as opposed to a "peer" doing things.
> 
> The client app only interacts with the Gateway service. Whether that Gateway service runs on a peer is irrelevant to the client application. There was a point in time where the Gateway service was implemented in a separate component, not in a peer, and this could be the case again in the future. It really makes no difference from the client application perspective discussed in this tutorial.

I beg to differ... the client needs to connect to a certain endpoint, and that endpoint is generally known as the peer endpoint throughout all the other doc topics. I'm just suggesting to mention that the gateway service runs in a peer so that readers know exactly which endpoint they are connecting to. That seems like a fundamental concept that client application developers need to know.  Yes, the Gateway _could_ in theory run anywhere, and that's exactly why the reader may be in doubt. But we chose to run it in the peer and should communicate that. Just one mention early on would be sufficient to clear up any doubt.

bestbeforetoday:
> I beg to differ... the client needs to connect to a certain endpoint, and that endpoint is generally known as the peer endpoint throughout all the other doc topics. I'm just suggesting to mention that the gateway service runs in a peer so that readers know exactly which endpoint they are connecting to. That seems like a fundamental concept that client application developers need to know. Yes, the Gateway _could_ in theory run anywhere, and that's exactly why the reader may be in doubt. But we chose to run it in the peer and should communicate that. Just one mention early on would be sufficient to clear up any doubt.

The code snippets and some of the supporting text already explicitly state that it's the peer endpoint. I'll add a little more wording to reinforce that though.

denyeart (approved):

denyeart:

PR #3119, opened 2021-12-14
Reduce CPU&memory cost of collecting endorsements

andrew-coleman:
Based on performance analysis of the gateway, the following changes are made in this commit:
1. Instead of collecting and holding copies of all proposal responses in order to reuse the protoutil.CreateTx() function, just store a single response payload and all of the signed endorsement objects.  Check that all payloads are bitwise idential and endorsers are unique at collection time.
2. Eliminate the duplicate unmarshalling that was occurring as a result of reusing protoutil.CreateTx

### DIFF
--- a/internal/pkg/gateway/api.go
+++ b/internal/pkg/gateway/api.go
@@ -29,7 +29,7 @@ import (
 )
 
 type endorserResponse struct {
-	endorsementSet []*peer.ProposalResponse
+	action         *peer.ChaincodeEndorsedAction
 	err            *gp.ErrorDetail
 	timeoutExpired bool
 }
@@ -134,12 +134,28 @@ func (gs *Server) Endorse(ctx context.Context, request *gp.EndorseRequest) (*gp.
 	}
 	proposal, err := protoutil.UnmarshalProposal(signedProposal.ProposalBytes)
 	if err != nil {
-		return nil, status.Errorf(codes.InvalidArgument, "failed to unpack transaction proposal: %s", err)
+		return nil, status.Error(codes.InvalidArgument, err.Error())
 	}
-	channel, chaincodeID, hasTransientData, err := getChannelAndChaincodeFromSignedProposal(signedProposal)
+	header, err := protoutil.UnmarshalHeader(proposal.Header)
 	if err != nil {
-		return nil, status.Errorf(codes.InvalidArgument, "failed to unpack transaction proposal: %s", err)
+		return nil, status.Error(codes.InvalidArgument, err.Error())
 	}
+	channelHeader, err := protoutil.UnmarshalChannelHeader(header.ChannelHeader)
+	if err != nil {
+		return nil, status.Error(codes.InvalidArgument, err.Error())
+	}
+	payload, err := protoutil.UnmarshalChaincodeProposalPayload(proposal.Payload)
+	if err != nil {
+		return nil, status.Error(codes.InvalidArgument, err.Error())
+	}
+	spec, err := protoutil.UnmarshalChaincodeInvocationSpec(payload.Input)
+	if err != nil {
+		return nil, status.Error(codes.InvalidArgument, err.Error())
+	}
+
+	channel := channelHeader.ChannelId
+	chaincodeID := spec.GetChaincodeSpec().GetChaincodeId().GetName()
+	hasTransientData := len(payload.GetTransientMap()) > 0
 
 	defaultInterest := &peer.ChaincodeInterest{
 		Chaincodes: []*peer.ChaincodeCall{{
@@ -148,12 +164,12 @@ func (gs *Server) Endorse(ctx context.Context, request *gp.EndorseRequest) (*gp.
 	}
 
 	var plan *plan
-	var endorsements []*peer.ProposalResponse
+	var action *peer.ChaincodeEndorsedAction
 	if len(request.EndorsingOrganizations) > 0 {
 		// The client is specifying the endorsing orgs and taking responsibility for ensuring it meets the signature policy
 		plan, err = gs.registry.planForOrgs(channel, chaincodeID, request.EndorsingOrganizations)
 		if err != nil {
-			return nil, status.Errorf(codes.Unavailable, "%s", err)
+			return nil, status.Error(codes.Unavailable, err.Error())
 		}
 	} else {
 		// The client is delegating choice of endorsers to the gateway.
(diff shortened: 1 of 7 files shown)


internal/pkg/gateway/endorsement.go
```diff
@@ -85,8 +92,8 @@ func (p *plan) endorsers() []*endorser {
 
 // Invoke update when an endorsement has been successfully received for the given endorser.
 // All layouts containing the group that contains this endorser are updated with the endorsement.
-// Returns array of endorsements, if at least one layout in the plan has been satisfied, otherwise nil.
-func (p *plan) update(endorser *endorser, endorsement *peer.ProposalResponse) []*peer.ProposalResponse {
+// Returns a ChaincodeEndorsedAction, if at least one layout in the plan has been satisfied, otherwise nil.
+func (p *plan) update(endorser *endorser, response *peer.ProposalResponse) (*peer.ChaincodeEndorsedAction, error) {
```
denyeart: update() function is not introduced in this PR, but I've always thought the function name could be more descriptive, something like processEndorsement().

denyeart (commented):
LGTM, but we probably want @yacovm to take a look.

internal/pkg/gateway/api.go
```diff
@@ -134,26 +134,42 @@ func (gs *Server) Endorse(ctx context.Context, request *gp.EndorseRequest) (*gp.
 	}
 	proposal, err := protoutil.UnmarshalProposal(signedProposal.ProposalBytes)
 	if err != nil {
-		return nil, status.Errorf(codes.InvalidArgument, "failed to unpack transaction proposal: %s", err)
+		return nil, status.Error(codes.InvalidArgument, err.Error())
```
yacovm, line 137: Unrelated to this PR, but... this Endorse function is over 200 lines long. 

Shouldn't it be shorter and easier to parse? you have down the road a goroutine which spawns another goroutine inside a loop, this is very brittle and easy to introduce bugs with time.

internal/pkg/gateway/endorsement.go
```diff
@@ -102,25 +109,41 @@ func (p *plan) update(endorser *endorser, endorsement *peer.ProposalResponse) []
 		}
 	}
 
+	// check the proposal responses are the same
```
yacovm, line 108: Unrelated to this PR, but - in line 100 you get back from a peer an endorsement but its PKI-ID that you reference, was not computed from that peer's certificate, but it was computed based on what discovery told you. 

However, you may actually get a different certificate from what discovery told you, due to either the peer being malicious, having an endorsement plugin, or the peer being behind a load balancer and then this `group` will be the empty string, and endorsers a nil slice. 

I suggest that instead of looking at the endorser which you computed with discovery, you will lookup using the endorsement which represents the reality and not the intent.

internal/pkg/gateway/endorsement.go
```diff
@@ -102,25 +109,41 @@ func (p *plan) update(endorser *endorser, endorsement *peer.ProposalResponse) []
...
+		}
+	}
+
 	for i := p.nextLayout; i < len(p.layouts); i++ {
 		layout := p.layouts[i]
 		if layout == nil {
 			continue
 		}
 		if quantity, ok := layout.required[group]; ok {
+			// ensure endorser uniqueness
```
yacovm: How do you know that the certificate matches the layout you are currently trying? What if the certificate of the endorser actually matches a different layout?

internal/pkg/gateway/endorsement.go
```diff
@@ -8,14 +8,17 @@ package gateway
...
 
 	"github.com/hyperledger/fabric-protos-go/peer"
 )
 
 type layout struct {
-	required     map[string]int // group -> quantity
-	endorsements []*peer.ProposalResponse
+	required        map[string]int // group -> quantity
+	endorsements    []*peer.Endorsement
+	uniqueEndorsers map[string]struct{}
```
yacovm: I suggest that we de-duplicate certificates afterwards and not during the collection, as it will make everything easier to read.

internal/pkg/gateway/api.go
```diff
@@ -134,26 +134,42 @@ func (gs *Server) Endorse(ctx context.Context, request *gp.EndorseRequest) (*gp.
...
 	}
-	channel, chaincodeID, hasTransientData, err := getChannelAndChaincodeFromSignedProposal(signedProposal)
```
yacovm, line 139: so you remove the call to `getChannelAndChaincodeFromSignedProposal` and now you have the header, chaincode proposal payload at hand, to pass into `prepareTransaction`. 

So instead of doing unmarshaling of the proposal at CreateSignedTx and here, you just do it once.
This addresses (2) and I can understand why this is more efficien. 

I don't understand why:
> Instead of collecting and holding copies of all proposal responses in order to reuse the protoutil.CreateTx() function, just store a single response payload and all of the signed endorsement objects. 

Is worth it. Once you obtain the *ProposalResponse you have a pointer which you then pass "by reference" to a function. You don't copy the memory of the payload, as it is []byte and it's also passed "by reference" (you pass a [pointer](https://go.dev/blog/slices-intro) to an array when you do the function call, and not copy the array itself).
Where do you get the efficiency improvement from?

internal/pkg/gateway/api.go
andrew-coleman: Yes, good point.  I've raised a separate issue (https://github.com/hyperledger/fabric/issues/3122) to track this.

internal/pkg/gateway/api.go
andrew-coleman: I understand it's passed by reference.  The idea of this change was to let go of any references to structures that are no longer needed (and hence available for GC) while it potentially continues to collect more endorsements.  This might allow more concurrent clients to make calls to Endorse() before running out of memory.

yacovm: OK so you're saying that the memory corresponding to the payload can be GCed early, I see.

internal/pkg/gateway/endorsement.go
andrew-coleman: Ok, that makes sense.  It's only being used here to cross reference which endorsing peers have responded so they can be removed from the layout(s).  However, given what you've said, it looks like I need to change the registry to use an ID that's not PKI-ID.  I'll raise a separate issue for this.

internal/pkg/gateway/endorsement.go
andrew-coleman: OK

internal/pkg/gateway/endorsement.go
andrew-coleman: Moved the de-duplication code, as suggested below

denyeart (approved):

denyeart:

PR #3132, opened 2021-12-17
- Fix failure to generate all possible combinations

timtim-git:
#### Type of change

- Bug fix

#### Description

Bug happens when the endorsement policy is set to k-out-of-n where (n - k) > 1.
For example, when there are 6 organizations and the endorsement policy is set
to majority (i.e. 4-out-of-6). The combinations calculated by chooseKoutOfN()
in original code are as below.
	[0 1 2 5], [0 1 2 5], [0 1 2 5], [0 1 3 5],
	[0 1 3 5], [0 1 4 5], [0 2 3 5], [0 2 3 5],
	[0 2 4 5], [0 3 4 5], [1 2 3 5], [1 2 3 5],
	[1 2 4 5], [1 3 4 5], [2 3 4 5]

Obviously, the function fails to find out all the possible combinations. Some of
 the combinations are overrided by child's call. This bug would trigger below
error sometimes when the alive peers are not within the above combinations.

	[discovery] chaincodeQuery -> ERRO 13a Failed constructing descriptor for
	chaincode chaincodes:<name:"XXXX" > ,: no peer combination can satisfy the
	endorsement policy

To fix it, the last line of choose() should be changed to pass a copy of
"currentSubGroup" to the recursive call instead, i.e.

	choose(n, targetAmount, i+1, append(make([]int, 0), currentSubGroup...), subGroups)

After applying the fix, the resulting combinations generated should be corrected
as below.
	[0 1 2 3], [0 1 2 4], [0 1 2 5], [0 1 3 4],
	[0 1 3 5], [0 1 4 5], [0 2 3 4], [0 2 3 5],
	[0 2 4 5], [0 3 4 5], [1 2 3 4], [1 2 3 5],
	[1 2 4 5], [1 3 4 5], [2 3 4 5]

### DIFF
--- a/common/graph/choose.go
+++ b/common/graph/choose.go
@@ -67,9 +67,6 @@ func choose(n int, targetAmount int, i int, currentSubGroup []int, subGroups *or
 func concatInts(a []int, elements ...int) []int {
 	var res []int
 	res = append(res, a...)
-	for _, e := range elements {
-		res = append(res, e)
-	}
-
+	res = append(res, elements...)
 	return res
 }
--- a/common/graph/choose.go
+++ b/common/graph/choose.go
@@ -61,5 +61,5 @@ func choose(n int, targetAmount int, i int, currentSubGroup []int, subGroups *or
 	// We either pick the current element
 	choose(n, targetAmount, i+1, append(currentSubGroup, i), subGroups)
 	// Or don't pick it
-	choose(n, targetAmount, i+1, currentSubGroup, subGroups)
+	choose(n, targetAmount, i+1, append(make([]int, 0), currentSubGroup...), subGroups)
 }
--- a/common/graph/choose.go
+++ b/common/graph/choose.go
@@ -59,7 +59,17 @@ func choose(n int, targetAmount int, i int, currentSubGroup []int, subGroups *or
 		return
 	}
 	// We either pick the current element
-	choose(n, targetAmount, i+1, append(currentSubGroup, i), subGroups)
+	choose(n, targetAmount, i+1, concatInts(currentSubGroup, i), subGroups)
 	// Or don't pick it
-	choose(n, targetAmount, i+1, append(make([]int, 0), currentSubGroup...), subGroups)
+	choose(n, targetAmount, i+1, currentSubGroup, subGroups)
+}
+
+func concatInts(a []int, elements ...int) []int {
+	var res []int
+	res = append(res, a...)
+	for _, e := range elements {
+		res = append(res, e)
+	}
+
+	return res
 }
--- a/common/graph/choose_test.go
+++ b/common/graph/choose_test.go
@@ -7,7 +7,7 @@ SPDX-License-Identifier: Apache-2.0
 package graph
 
 import (
-	"reflect"
+	"fmt"
 	"testing"
 
 	"github.com/stretchr/testify/require"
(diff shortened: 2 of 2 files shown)


denyeart:
Thank you for the contribution! Please remember to sign the commit so that DCO check passes, instructions:
https://github.com/hyperledger/fabric/pull/3132/checks?check_run_id=4558455955

andrew-coleman:
@timtim-git, please could you add a unit test to `choose_test.go` that passes with your fix and fails without?
Many thanks!

common/graph/choose.go
```diff
@@ -61,5 +61,5 @@ func choose(n int, targetAmount int, i int, currentSubGroup []int, subGroups *or
 	// We either pick the current element
 	choose(n, targetAmount, i+1, append(currentSubGroup, i), subGroups)
 	// Or don't pick it
-	choose(n, targetAmount, i+1, currentSubGroup, subGroups)
```
yacovm, line 64: Nice catch and thanks for the fix.

I think it reads better if we clone when we pick the element and not when we don't pick it:

```
	// We either pick the current element
	choose(n, targetAmount, i+1, concatInts(currentSubGroup, i), subGroups)
	// Or don't pick it
	choose(n, targetAmount, i+1, currentSubGroup, subGroups)
```

and let's also move the make([]int, 0) to a separate function so it will read better, like:

```
func concatInts(a []int, elements  ... int) []int {
	var res []int
	res = append(res, a ...)
	for _, e := range elements {
		res = append(res, e)
	}

	return res
}
```

yacovm:
@denyeart when this is merged we would need to backport it everywhere...

common/graph/choose_test.go
```diff
@@ -24,3 +25,34 @@ func TestCombinationsExceed(t *testing.T) {
...
+		&indiceSet{[]int{0, 2, 3, 5}},
+		&indiceSet{[]int{0, 2, 4, 5}},
+		&indiceSet{[]int{0, 3, 4, 5}},
+		&indiceSet{[]int{1, 2, 3, 4}},
+		&indiceSet{[]int{1, 2, 3, 5}},
+		&indiceSet{[]int{1, 2, 4, 5}},
+		&indiceSet{[]int{1, 3, 4, 5}},
+		&indiceSet{[]int{2, 3, 4, 5}},
+	}
+	for _, expected := range expectedSets {
```
yacovm: Since the function `chooseKoutOfN` is deterministic, we do not have to check for set equality but we can just convert the indiceSets to a string array:

```
func indiceSetsToStrings(sets indiceSets) []string {
	var res []string
	for _, set := range sets {
		res = append(res, fmt.Sprintf("%v", set.indices))
	}
	return res
}
```

and then it will be more friendly to read the test:

```
require.Equal(t, indiceSetsToStrings(expectedSets), indiceSetsToStrings(results))
```

yacovm:
```
Checking with staticcheck
The following staticcheck issues were flagged
common/graph/choose.go:70:2: should replace loop with res = append(res, elements...) (S1011)
make: *** [Makefile:184: linter] Error 1

```

This will fix your build error:


```
func concatInts(a []int, elements ...int) []int {
	var res []int
	res = append(res, a...)
	res = append(res, elements...)
	return res
}

```

yacovm (approved):
LGTM

yacovm:
@denyeart let's merge this everywhere

yacovm:

yacovm:

yacovm:

yacovm:

yacovm:

PR #3134, opened 2021-12-18
sampleconfig - change address in core.yaml to avoid collision with orderer.yaml

rajan-31:
#### Type of change

- Improvement to code

#### Description

If "/fabric/sampleconfig" is used to create a network,

in "core.yaml"

```
operations:
# host and port for the operations server
listenAddress: 127.0.0.1:9443
```

in "orderer.yaml"

```
Admin:
# host and port for the admin server
ListenAddress: 127.0.0.1:9443
```

The port in both addresses is the same which is causing

```
Error: failed to initialize operations subsystem: listen tcp 127.0.0.1:9443: bind: address already in use
```

This happened while following [Running chaincode in development mode Tutorial](https://hyperledger-fabric.readthedocs.io/en/latest/peer-chaincode-devmode.html).

So, changing the address in the core.yaml solves the problem and it is not overlapping with any other config.

#### Additional details

Tested `127.0.0.1:9444` and it's not not colliding with any other config.


rajan-31:
@denyeart Please take a look into this to make necessary changes, or I can close this PR.

What am I missing?

sampleconfig/core.yaml
```diff
@@ -731,7 +731,7 @@ ledger:
 ###############################################################################
 operations:
     # host and port for the operations server
-    listenAddress: 127.0.0.1:9443
+    listenAddress: 127.0.0.1:9444
```
denyeart, line 734: The orderer operations port is 8443, so I'd suggest to change the peer operations port to 8444 instead (the pattern of incrementing by one between orderer and peer is done for the grpc port, so this would be consistent).

The old peer value of 9443 is mentioned several times in the /docs/source files. So please grep that directory recursively for any mentions and change to 8444 as well (don't change any orderer admin 9443 references, just the peer operations 9443 references.

denyeart (changes requested):

sampleconfig/core.yaml
denyeart: There is a long history of peer operations port being 9443, so I wouldn't change it as production users may use it.

Typically orderer and peer run in their own container so there is no such conflict. Perhaps it would be better to update the dev mode instructions and mention to set the operations listenAddress to something else, using env variable `CORE_OPERATIONS_LISTENADDRESS`?

denyeart (changes requested):

sampleconfig/core.yaml
rajan-31: I have done the changes in the peer dev mode instructions, please a look.

denyeart (approved):

denyeart:
Thank you!

PR #3165, opened 2022-01-10
Update fabric go module to v2

### LINKED ISSUE #2929: Go's indexing servers didn't pick up real version number for fabric
I use this script to retrieve all Go packages:
```
$ cat go-retrieve-all-packages.sh 
#!/bin/sh

SINCE=$1 # like 2021-08-01T00:00:00Z

while [ -n "${SINCE}" ]; do

	# download
	echo "downloading SINCE=${SINCE}"
	wget "https://index.golang.org/index?limit=2000&since=${SINCE}" -q -O output-${SINCE}.json

	# get last since
	SINCE_NEW=`tail -1 output-${SINCE}.json | sed -e 's|.*"Timestamp":"||; s|".*||'`

	# EOF?
	if [ "${SINCE_NEW}" = "${SINCE}" ] || [ -z "${SINCE_NEW}" ]; then
		echo "=== End of project list reached @ ${SINCE} ==="
		exit 0
	else
		SINCE=${SINCE_NEW}
	fi
done
```

For example, ```go-retrieve-all-packages.sh 2021-09-10T00:00:00Z```

fabric is registered with incorrect versions like ```v0.0.0-20210916200841-1243e99c1ff2```, and ```2.3.3``` doesn't appear there.

denyeart:
Although Fabric is not intended to be used as a go module that other
projects import, some projects use Go tooling to download repositories
and this does not work for Fabric v2.x since Fabric's go module is
not defined as a v2 module. This change marks Fabric's go module
as v2 so that Fabric v2.x versions can be downloaded with Go tooling.

Fixes #2929

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,4 +1,4 @@
-module github.com/hyperledger/fabric
+module github.com/hyperledger/fabric/v2
 
 go 1.14
 


denyeart:
Actually, this will require updating all the imports in all the go files, will evaluate further...

PR #3166, opened 2022-01-10
Update fabric go module to v2

### LINKED ISSUE #2929: Go's indexing servers didn't pick up real version number for fabric

denyeart:
Although Fabric is not intended to be used as a go module that other
projects import, some projects use Go tooling to download repositories
and this does not work for Fabric v2.x since Fabric's go module is
not defined as a v2 module. This change marks Fabric's go module
as v2 so that Fabric v2.x versions can be downloaded with Go tooling.

Fixes #2929


yacovm (changes requested):
> some projects use Go tooling to download repositories
> and this does not work for Fabric v2.x since Fabric's go module is
> not defined as a v2 module. 

But it's the wrong approach since Fabric was not meant to be consumed this way in the first place. If you want to download the repo, you can use git. 

Regardless, why do we need to put v2 in everywhere's import path? Can't you just update go.mod ? This way it will look v2 in the URL but the imports will remain the same. 

In [SmartBFT](https://github.com/SmartBFT-Go/consensus) we have modules and the latest version is v2 but none of the import paths have v2 in them.

denyeart:
@yacovm Yeah, the further I go the more disruptive it is looking. At first I tried to just update the module path in go.mod to include /v2, but then all the imports in all the go files resolved to old v1 fabric dependencies since there was no /v2 in the imports. Adding /v2 to the imports fixed that and it compiles now, but there are several more issues in script files that will need to be resolved. So I'm thinking of abandoning the idea for the reasons you mention. I've asked the issue opener for an example of why it is needed.

BTW the SmartBFT module is not tagged as /v2 at https://github.com/SmartBFT-Go/consensus/blob/v2/go.mod#L1, so I don't think consumers will be able to specify SmartBFT version v2.x. They would get error like
"version "v2.x.x" invalid: should be v0 or v1, not v2"
since the go index servers don't know about any /v2 module.

See https://go.dev/blog/v2-go-modules.

yacovm:
> Yeah, the further I go the more disruptive it is looking. At first I tried to just update the module path in go.mod to include /v2, but then all the imports in all the go files resolved to old v1 fabric dependencies since there was no /v2 in the imports.

Have you tried using a replace directive in go.mod to force Go to look at the local folder for github.com/hyperledger/fabric ? 


> BTW the SmartBFT module is not tagged as /v2 at https://github.com/SmartBFT-Go/consensus/blob/v2/go.mod#L1, so I don't think consumers will be able to specify SmartBFT version v2.x. They would get error like
> "version "v2.x.x" invalid: should be v0 or v1, not v2"
> since the go index servers don't know about any /v2 module.


They can just specify the shortened commit hash which is published in every release instead of using the tag "name".

denyeart:
The change to v2 appears to be too disruptive, closing.

Issue #3027, opened 2021-11-08
Update reconciliation process


manish-sethi:
Update reconciliation code to consider PurgeMarker and deal with partial write-sets

Issue #3023, opened 2021-11-08
Changes during private data commit and deletion caused by BlockToLive


manish-sethi:
Create HashedIndex entries when adding new private data and Delete HashedIndex entries when deleting private data caused by BlockToLive policy.

PR #3167, opened 2022-01-12
Maintain Pvt data hashed index

manish-sethi:
This commit adds code for creating hashed index entries during block commit and deleting the hashed index entries during purge triggered by _Block to live_ based expiry

#### Type of change
- New feature

#### Additional details
The indexes needs also to be maintained when the private data is committed via reconciliation. However, this will be covered in a separate story (issue #3027) because, during reconciliation we need to take care of trimming the already purged data, if any.

#### Related issues
issue #3023


denyeart (approved):
Private data store was already getting complex, now it is even more complex, took me a long time to connect all the dots between the data store entries to be able to review the PR. I don't see an alternative for making it less complex given the requirements though. Maybe I will attempt a PR to rename some variables to make things like "dataKeys" and "dataEntries" more expressive so that the code is more understandable in the future without having to constantly reference the definitions and data structure.

PR #3171, opened 2022-01-13
Refactor gateway Endorse() method

### LINKED ISSUE #3122: Refactor Endorser() function in gateway.
Improve the readability of the gateway's Endorse() function which has grown too long over time.  
Currently the retry logic is coded so that a goroutine spawns another goroutine.  This should be refactored to ease future maintenance.

andrew-coleman:
To improve readability and ease of maintenance

Resolves #3122 

### DIFF
--- a/internal/pkg/gateway/api.go
+++ b/internal/pkg/gateway/api.go
@@ -12,13 +12,13 @@ import (
 	"io"
 	"math/rand"
 	"strings"
-	"sync"
 
 	"github.com/golang/protobuf/proto"
 	"github.com/hyperledger/fabric-protos-go/common"
 	gp "github.com/hyperledger/fabric-protos-go/gateway"
 	ab "github.com/hyperledger/fabric-protos-go/orderer"
 	"github.com/hyperledger/fabric-protos-go/peer"
+	"github.com/hyperledger/fabric/common/flogging"
 	"github.com/hyperledger/fabric/common/ledger"
 	"github.com/hyperledger/fabric/core/aclmgmt/resources"
 	"github.com/hyperledger/fabric/core/chaincode"
@@ -77,7 +77,7 @@ func (gs *Server) Evaluate(ctx context.Context, request *gp.EvaluateRequest) (*g
 			ctx, cancel := context.WithTimeout(ctx, gs.options.EndorsementTimeout)
 			defer cancel()
 			pr, err := endorser.client.ProcessProposal(ctx, signedProposal)
-			code, message, retry, remove := gs.responseStatus(pr, err)
+			code, message, retry, remove := responseStatus(pr, err)
 			if code == codes.OK {
 				response = pr.Response
 				// Prefer result from proposal response as Response.Payload is not required to be transaction result
@@ -157,11 +157,7 @@ func (gs *Server) Endorse(ctx context.Context, request *gp.EndorseRequest) (*gp.
 	chaincodeID := spec.GetChaincodeSpec().GetChaincodeId().GetName()
 	hasTransientData := len(payload.GetTransientMap()) > 0
 
-	defaultInterest := &peer.ChaincodeInterest{
-		Chaincodes: []*peer.ChaincodeCall{{
-			Name: chaincodeID,
-		}},
-	}
+	logger := gs.logger.With("channel", channel, "chaincode", chaincodeID, "txID", request.TransactionId)
 
 	var plan *plan
 	var action *peer.ChaincodeEndorsedAction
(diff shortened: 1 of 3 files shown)


internal/pkg/gateway/api.go
```diff
@@ -173,183 +169,190 @@ func (gs *Server) Endorse(ctx context.Context, request *gp.EndorseRequest) (*gp.
...
-					logger.Warnw("Endorse call to endorser failed", "channel", request.ChannelId, "txID", request.TransactionId, "endorserAddress", firstEndorser.endpointConfig.address, "endorserMspid", firstEndorser.endpointConfig.mspid, "error", message)
-					errDetails = append(errDetails, errorDetail(firstEndorser.endpointConfig, message))
-					if remove {
-						gs.registry.removeEndorser(firstEndorser)
-					}
-					firstEndorser = plan.nextPeerInGroup(firstEndorser)
-					firstResponse = nil
+	}
+
+	for plan.completed == nil {
```
denyeart: It is not immediately obvious what the for condition is... from the name it seems like 'completed' would be  boolean, but once you navigate over to the plan struct it is a layout. Maybe just changing it to 'completedLayout' would help.

internal/pkg/gateway/api.go
```diff
@@ -173,183 +169,190 @@ func (gs *Server) Endorse(ctx context.Context, request *gp.EndorseRequest) (*gp.
...
+			// no more layouts
+			break
+		}
+		// send to all the endorsers
+		waitCh := make(chan bool, len(endorsers))
+		for _, e := range endorsers {
+			go func(e *endorser) {
+				success := false
+				for !success && e != nil {
+					e, success = gs.processProposal(ctx, plan, e, signedProposal, logger)
```
denyeart: I'm not following the intent of this inner loop... are these retries? To the same or different endorser? Haven't the set of endorsers already been selected in the outer endorsers loop?

Looking further, I guess within processProposal() the nextPeerInGroup() will increment to the next endorser in the group. But the intended logic is not clear when looking at this code. From looking at this code it seems like the endorser has already been chosen and we're calling processProposal() on it only. I think the logic would be more clear if the nextPeerInGroup() call were moved here and not embedded in processProposal().

internal/pkg/gateway/endorsement.go
```diff
@@ -89,7 +91,7 @@ func (p *plan) endorsers() []*endorser {
 // Invoke processEndorsement when an endorsement has been successfully received for the given endorser.
 // All layouts containing the group that contains this endorser are updated with the endorsement.
 // Returns a ChaincodeEndorsedAction, if at least one layout in the plan has been satisfied, otherwise nil.
-func (p *plan) processEndorsement(endorser *endorser, response *peer.ProposalResponse) (*peer.ChaincodeEndorsedAction, error) {
+func (p *plan) processEndorsement(endorser *endorser, response *peer.ProposalResponse) bool {
```
denyeart, line 95: The return comment needs to be updated.

internal/pkg/gateway/endorsement.go
```diff
@@ -128,12 +131,14 @@ func (p *plan) processEndorsement(endorser *endorser, response *peer.ProposalRes
 				delete(layout.required, group)
 				if len(layout.required) == 0 {
 					// no groups left - this layout is now satisfied
-					return &peer.ChaincodeEndorsedAction{ProposalResponsePayload: p.responsePayload, Endorsements: uniqueEndorsements(layout.endorsements)}, nil
+					p.completed = layout
+					// return &peer.ChaincodeEndorsedAction{ProposalResponsePayload: p.responsePayload, Endorsements: uniqueEndorsements(layout.endorsements)}, nil
```
denyeart: Remove commented return.

internal/pkg/gateway/api.go
```diff
@@ -173,183 +169,190 @@ func (gs *Server) Endorse(ctx context.Context, request *gp.EndorseRequest) (*gp.
...
+
+	return &gp.EndorseResponse{PreparedTransaction: preparedTransaction}, nil
+}
+
+type ppResponse struct {
+	response *peer.ProposalResponse
+	err      error
+}
+
+func (gs *Server) processProposal(ctx context.Context, plan *plan, endorser *endorser, signedProposal *peer.SignedProposal, logger *flogging.FabricLogger) (*endorser, bool) {
```
denyeart: For these non-trivial functions, it helps to have a function comment explaining what the function does.

internal/pkg/gateway/api.go
```diff
@@ -173,183 +169,190 @@ func (gs *Server) Endorse(ctx context.Context, request *gp.EndorseRequest) (*gp.
...
+}
 
-	var errorDetails []proto.Message
-	for action == nil {
-		// loop through the layouts until one gets satisfied
-		endorsers := plan.endorsers()
-		if endorsers == nil {
-			// no more layouts
-			break
+func (gs *Server) planFromFirstEndorser(ctx context.Context, channel string, chaincodeID string, hasTransientData bool, signedProposal *peer.SignedProposal, logger *flogging.FabricLogger) (*plan, error) {
```
denyeart, line 273: For these non-trivial functions, it helps to have a function comment explaining what the function does.

denyeart (commented):
This does indeed help with Endorse() readability. I think a few more things could be done to help readability though.

After that, I think we should have @yacovm take a look since he knows the discovery behavior best wrt plans, layouts, etc.

internal/pkg/gateway/api.go
andrew-coleman: Yes, they're retries on other endorsers in the same group.  I've moved the nextPeerInGroup() call as you suggested.  I agree it's clearer to do that.

denyeart (approved):
Looks good now, much easier to read. I'll go ahead and merge and backport so that the branches are consistent.

denyeart:

PR #3174, opened 2022-01-15
replace arm64 with amd64 for installation scripts

RohanShrothrium:
Signed-off-by: Rohan Shrothrium <shrothriumrohan@gmail.com>

While using the new mac books which has an arm64 infrastructure, the installation and bootstrap script fails as there is no specific binaries for arm64. I downloaded the amd64 binaries and fabric images and they work as expected on the new M1 architecture.

#### Type of change

#### Description
This PR incorporates changes to download the amd64 binaries and images for an arm64 architecture.

### DIFF
--- a/scripts/bootstrap.sh
+++ b/scripts/bootstrap.sh
@@ -9,7 +9,7 @@
 VERSION=2.4.1
 # if ca version not passed in, default to latest released version
 CA_VERSION=1.5.2
-ARCH=$(echo "$(uname -s|tr '[:upper:]' '[:lower:]'|sed 's/mingw64_nt.*/windows/')-$(uname -m | sed 's/x86_64/amd64/g')")
+ARCH=$(echo "$(uname -s|tr '[:upper:]' '[:lower:]'|sed 's/mingw64_nt.*/windows/')-$(uname -m | uname -m | sed -E -e 's/x86_64|arm64/amd64/g')")
 MARCH=$(uname -m)
 
 printHelp() {
--- a/scripts/bootstrap.sh
+++ b/scripts/bootstrap.sh
@@ -9,7 +9,7 @@
 VERSION=2.4.1
 # if ca version not passed in, default to latest released version
 CA_VERSION=1.5.2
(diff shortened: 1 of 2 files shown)


ale-linux:
@RohanShrothrium thanks for your contribution!

@denyeart if we go with this change, should we clarify somewhere that we don't do any testing on arm64 and it's a "use at your own risk" scenario?

davidkel:
This change seems to be too generic and will download amd64 binaries onto other arm based systems such as raspberry pi or other arm based environments that exist running linux distributions. If this change is accepted then it would be good if it could detect darwin and download the amd64 versions and reports an error for other platforms.

RohanShrothrium:
@davidkel, Yeah I could do that.
Update the arm64 to amd 64 only when the ARCH is darwin.
I'll update the PR and push the changes.

RohanShrothrium:
Hi @davidkel!
I have made the requested changes.
Take a look at it!
Thanks.

davidkel:
@ale-linux true that arm64 is not tested, but in this case what happens is that the amd64 binaries are downloaded and rosetta2 on a Mac provides whatever is required to execute amd64 on an arm64 architecture, so technically the images are tested, but the only way to test this scenario is by someone using an arm64 Mac. Given the uptake on arm64 Macs, I think this is a reasonable change to allow these mac users to be able to run fabric on their development laptops until such time as fabric decide to release native arm64 binaries

ale-linux:
@davidkel don't get me wrong, I don't debate whether we should accept this patch, but whether we should add to it a disclaimer that we don't consider this platform to be production ready (yet).

davidkel:
@ale-linux As far as I know the only platform we declare as production ready is linux, Mac and Windows are not production ready platforms. Sorry that was kind of what I was alluding to in my previous comment

RohanShrothrium:
@davidkel and @ale-linux , I updated the scripts, downloaded the images and binaries and ran the fabric-samples examples locally on my M1 mac and they work as expected.
This change will help in easy installation for the mac users when they follow the installation scripts as mentioned in the fabric documentation.

davidkel (approved):
This looks good to me from a windows and linux point of view, I'm not a maintainer but this is my opinion

denyeart (approved):
Thanks for the contribution, good update!

PR #3191, opened 2022-01-25
Update deploy_chaincode.md `package` command docs

curldapps2:
Define the default behavior of `peer lifecycle chaincode package`, which does *not* include node_modules in the resulting tar.gz package for chaincode. The existing step of `npm install` and quote "If the command is successful, the JavaScript packages will be installed inside a `node_modules` folder. Now that we that have our dependencies, we can create the chaincode package." would lead a reader to believe that `node_modules` are included when they are not. 

Also added more detail about what chaincode services allow `node_modules` when packaged manually.

#### Type of change

- Bug fix
- New feature
- Test update
- Documentation update

### DIFF
--- a/docs/source/deploy_chaincode.md
+++ b/docs/source/deploy_chaincode.md
@@ -247,7 +247,7 @@ To install the smart contract dependencies, run the following command from the `
 npm install
 ```
 
-If the command is successful, the JavaScript packages will be installed inside a `node_modules` folder.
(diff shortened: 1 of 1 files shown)


denyeart:
Thank you for the contribution! Please sign the commit using the instructions at https://github.com/hyperledger/fabric/pull/3191/checks?check_run_id=4941497881

denyeart:
@mbwhite Since we haven't received a signed commit yet, could you maybe move this one forward in a new PR? Take the inspiration from this PR and add any other clarifications that you think are needed...

ale-linux (approved):

ale-linux:
@curldapps2 thanks for your contribution, could you please take care of the DCO so that we can merge?

C0rWin:
this commit is stale for quite long period of time, closing it.

PR #3214, opened 2022-02-07
Encode Hashed Data To Hexadecimal

jcastrence:
#### Type of change

- Bug fix

#### Description

- Hashed data (snapshot record hashed keys and values) was not properly converting to string due to JSON encoding not supporting certain special characters, resulting in data loss. Hashed data is now converted to hexadecimal string to maintain data integrity.
- JSON output now contains new boolean field, hashed, indicating whether diffRecord has hashed data (key and value).

### DIFF
--- a/internal/ledgerutil/compare.go
+++ b/internal/ledgerutil/compare.go
@@ -10,6 +10,7 @@ import (
 	"bufio"
 	"bytes"
 	"container/heap"
+	"encoding/hex"
 	"encoding/json"
 	"fmt"
 	"io/ioutil"
@@ -84,7 +85,7 @@ func Compare(snapshotDir1 string, snapshotDir2 string, outputDirLoc string, firs
 		if err != nil {
 			return 0, "", err
 		}
-		outputPubFileWriter, err := findAndWriteDifferences(outputDirPath, AllPubDiffsByKey, channelName, snapshotPubReader1, snapshotPubReader2, firstDiffs, firstRecords)
+		outputPubFileWriter, err := findAndWriteDifferences(outputDirPath, AllPubDiffsByKey, channelName, false, snapshotPubReader1, snapshotPubReader2, firstDiffs, firstRecords)
 		if err != nil {
 			return 0, "", err
 		}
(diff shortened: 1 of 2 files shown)


internal/ledgerutil/compare.go
```diff
@@ -428,19 +432,29 @@ func earlierSSRecord(r1 *snapshotRecord, r2 *snapshotRecord) *snapshotRecord {
...
 	return &snapshotRecord{
-		Value:    string(record.Value),
+		Value:    bytesToString(record.Value, hashed),
 		BlockNum: blockNum,
 		TxNum:    txNum,
 	}, nil
 }
 
+// Converts byte slice to string, respects hashed data integrity
+// If data is hashsed, converts byte slice to hexadecimal string encoding
```
yacovm: typo

internal/ledgerutil/compare.go
```diff
@@ -428,19 +432,29 @@ func earlierSSRecord(r1 *snapshotRecord, r2 *snapshotRecord) *snapshotRecord {
...
+func bytesToString(v []byte, h bool) string {
```
yacovm, line 450: isn't it easier to just use `string(v)` explicitly instead of calling a function and passing it false?

jcastrence: It would still need to be determined whether hashed is false before using `string(v)`, resulting in the same comparison that `bytesToString` aims to isolate

yacovm: ok and here is another question/proposal: 

Why not use: [utf8.Valid](https://pkg.go.dev/unicode/utf8#Valid) to determine if it's a valid UTF8 encoding and then [print as hex based on that](https://go.dev/play/p/p6Ke0RgpeTb)? Then you won't need to "drag" the hashed boolean all the way down.

```
package main

import (
	"encoding/hex"
	"fmt"
	"unicode/utf8"
)

func asReadableString(in []byte) string {
	if utf8.Valid(in) {
		return string(in)
	}
	return fmt.Sprintf("0x%s", hex.EncodeToString(in))
}

func main() {
	realString := []byte("Hello, 世界")
	someGarbage := []byte{0xff, 0xfe, 0xfd}
	someGarbageAsHex := []byte(hex.EncodeToString(someGarbage))

	fmt.Println(asReadableString(realString))       // prints Hello, 世界
	fmt.Println(asReadableString(someGarbage))      // prints 0xfffefd
	fmt.Println(asReadableString(someGarbageAsHex)) // prints fffefd
}

```

jcastrence: This is also an option. I thought it'd be a little more uniform and less confusing to simply convert all hashed data to hex since converting it to UTF-8 string, even if it has a valid UTF-8 encoding, doesn't really provide any extra context to the user; it's still hashed data. If we use the convention of converting hashed data to hex and public data to regular UTF-8 string we at least give the user another way to differentiate between public and private data.

yacovm (approved):

PR #3218, opened 2022-02-08
ledgerutil exits with a proper code

shimos:
This patch makes ledgerutil exit with a proper code when it fails.

#### Type of change

#### Description

The ledgerutil command exits with 0 (success) whether or not it succeeds in the comparison; it exits with 1 (failure) only when the parsing of the command line arguments are unsuccessful.

This patch makes it exit with 1 additionally when
- it fails in the comparison of snapshots
- it finds that the snapshots are different after the comparison

### DIFF
--- a/cmd/ledgerutil/main.go
+++ b/cmd/ledgerutil/main.go
@@ -53,14 +53,14 @@ func main() {
 			*outputDir, err = os.Getwd()
 			if err != nil {
 				fmt.Printf("%s%s\n", compareErrorMessage, err)
-				return
+				os.Exit(1)
 			}
 		}
 
 		count, outputDirPath, err := ledgerutil.Compare(*snapshotPath1, *snapshotPath2, *outputDir, *firstDiffs)
 		if err != nil {
 			fmt.Printf("%s%s\n", compareErrorMessage, err)
-			return
+			os.Exit(1)
 		}
 
 		fmt.Print("\nSuccessfully compared snapshots. ")
@@ -68,6 +68,7 @@ func main() {
 			fmt.Println("Both snapshot public state and private state hashes were the same. No results were generated.")
 		} else {
 			fmt.Printf("Results saved to %s. Total differences found: %d\n", outputDirPath, count)
+			os.Exit(2)
 		}
 	}
 }
--- a/cmd/ledgerutil/main_test.go
+++ b/cmd/ledgerutil/main_test.go
@@ -38,7 +38,11 @@ func TestArguments(t *testing.T) {
 		},
 		"one-snapshot": {
 			exitCode: 1,
-			args:     []string{"compare, snapshotDir1"},
+			args:     []string{"compare", "snapshotDir1"},
+		},
+		"invalid-snapshot-dirs": {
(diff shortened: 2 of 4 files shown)


cmd/ledgerutil/main.go
```diff
@@ -53,21 +53,22 @@ func main() {
...
-			return
+			os.Exit(1)
 		}
 
 		fmt.Print("\nSuccessfully compared snapshots. ")
 		if count == -1 {
 			fmt.Println("Both snapshot public state and private state hashes were the same. No results were generated.")
 		} else {
 			fmt.Printf("Results saved to %s. Total differences found: %d\n", outputDirPath, count)
+			os.Exit(1)
```
manish-sethi: This is not an error scenario for the tool. The tool is meant to find the differences.

manish-sethi (approved):

manish-sethi (commented):
Sorry, hit the approve button earlier mistakenly.

manish-sethi (changes requested):
See above comment

cmd/ledgerutil/main.go
shimos: @manish-sethi Thank you for your review. While I agree that this is not an error case, I think it is more convenient for users to know if there is any difference by the exit code as they can know with diff(1). Does returning a different code, say 2, make sense?

manish-sethi: Sure, not a big deal if you want it to be distinguishable outcome. In that case, yes, better to use a different exit code.
If you want to go ahead with this change, please update the documentation as well.

BTW, do you have a scenario in mind where you would want to include this util as a part of some larger workflow where you would like to decide on the next step based on whether the snapshot are same or different?

manish-sethi: Also, in that case, I would suggest make this exit code as `0` and the one above it (`if count == -1`) as `1` as the one above is an exit point because of wrong input and this one is a default intended exit point.

shimos:
@manish-sethi Thank you for the comment. I have updated the patch to make ledgerutil return a different exit code when the snapshots are different as well as the documentation for ledgerutil.

As the command name `compare` suggests, I think the case that the snapshots are identical would not be an error case, either. Therefore, the question would be "`0` should be assigned for the 'identical' case or the 'different' case".

As a use case, I have in my mind some kind of a maintenance script that is executed automatically and periodically. In this case, snapshots being different is a scenario that needs attention of administrators. Considering some other common examples such as diff(1) and strcmp(3), I'd suggest the utility return `0` when the snapshots are identical.

manish-sethi:
> As a use case, I have in my mind some kind of a maintenance script that is executed automatically and periodically. In this case, snapshots being different is a scenario that needs attention of administrators. Considering some other common examples such as diff(1) and strcmp(3), I'd suggest the utility return `0` when the snapshots are identical.

Though the original intention of this utility was to use when actually the snapshots are different as for the same snapshots, it's easy to detect just by looking at their metadata file. But I fine with this new proposed meaning - thanks for your contribution.

manish-sethi (approved):

PR #3229, opened 2022-02-14
Fix inconsistent behavior of GetBookmarkAndClose method in different statedb

Manphil:
#### Type of change

- Bug fix

#### Description

When there are no more results in paginated range query, statecouchdb return endKey
as bookmark in GetBookmarkAndClose method, but stateleveldb return empty string as bookmark.
This inconsistent behavior may cause different code logic when use bookmark in smart contract
with different statedb.

### DIFF
--- a/core/ledger/kvledger/txmgmt/statedb/statecouchdb/statecouchdb.go
+++ b/core/ledger/kvledger/txmgmt/statedb/statecouchdb/statecouchdb.go
@@ -1098,6 +1098,9 @@ func (scanner *queryScanner) GetBookmarkAndClose() string {
 		retval = scanner.paginationInfo.bookmark
 	} else {
 		retval = scanner.queryDefinition.startKey
+		if retval == scanner.queryDefinition.endKey {
+			retval = ""
+		}
 	}
 	scanner.Close()
 	return retval
--- a/core/ledger/kvledger/txmgmt/statedb/commontests/test_common.go
+++ b/core/ledger/kvledger/txmgmt/statedb/commontests/test_common.go
@@ -913,6 +913,17 @@ func TestPaginatedRangeQuery(t *testing.T, dbProvider statedb.VersionedDBProvide
 	returnKeys = []string{"key1", "key10", "key11", "key12"}
 	_, err = executeRangeQuery(t, db, "ns1", "key1", "key15", int32(4), returnKeys)
 	require.NoError(t, err)
+
+	// Test the returned bookmark of range query when there are no more results in that range
+	returnKeys = []string{"key1", "key10", "key11"}
+	nextStartKey, err = executeRangeQuery(t, db, "ns1", "key1", "key15", int32(3), returnKeys)
+	require.NoError(t, err)
+
+	// NextStartKey is now passed in as startKey, and return the nextStartKey which should be empty string
+	returnKeys = []string{"key12", "key13", "key14"}
+	nextStartKey, err = executeRangeQuery(t, db, "ns1", nextStartKey, "key15", int32(3), returnKeys)
+	require.NoError(t, err)
+	require.Equal(t, "", nextStartKey)
 }
 
 // TestRangeQuerySpecialCharacters tests range queries for keys with special characters and/or non-English characters
--- a/core/ledger/kvledger/txmgmt/statedb/statecouchdb/statecouchdb_test.go
+++ b/core/ledger/kvledger/txmgmt/statedb/statecouchdb/statecouchdb_test.go
@@ -1690,7 +1690,7 @@ func testRangeQueryWithPageSize(
 			continue
 		}
 		nextStartKey := itr.GetBookmarkAndClose()
-		if nextStartKey == endKey {
+		if nextStartKey == "" {
 			break
 		}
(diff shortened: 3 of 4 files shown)


denyeart:
Thank you @Manphil , good find.

I looked into this and the problem actually appears even worse. It seems when using leveldb with a paginated range query, the bookmark is always returned as an empty string, even in cases where there are additional results beyond the requested page. This makes it impossible for a client to page through range query results as they won't know the next startKey to use.

If you can confirm the same, I think this is the root problem that should be fixed.

Manphil:
@denyeart

Manphil:
@denyeart 
Actually, I found this issue when using [GetStateByRangeWithPagination](https://github.com/hyperledger/fabric-chaincode-go/blob/1476cf1d3206f620db7eea12312c98669d39fa22/shim/interfaces.go#L136) function in chaincode. This function will return the result and metadata containing bookmark. When there are still more results, the bookmark returned is expected as normal. But when there are no more results, the bookmark returned is different in different statedb (i.e. the empty string of leveldb and endKey of couchdb).

So I think there should be a unified string to be returned (e.g. empty string or the endKey) to mark the end of paginated range query. And this end flag should be described in document of chaincode shim api(i.e. GetStateByRangeWithPagination), in this way, the chaincode developer will know when the paginated range query will stop according to the bookmark returned.

For example, Assuming that all of the keys stored in statedb are "1"  "2"  "3"  "4"  "5"  "6".

When using leveldb, the first function call of `GetStateByRangeWithPagination("1", "7", 3, "")` will return "4" as bookmark, and the second call `GetStateByRangeWithPagination("1", "7", 3, "4")` will return "" as bookmark. If the empty string is the end flag of paginated range query described in document, the chaincode developer will know the paginated range query is over.

When using couchdb, the first function call of `GetStateByRangeWithPagination("1", "7", 3, "")` will return "4" as bookmark, which is same as leveldb, and the second call `GetStateByRangeWithPagination("1", "7", 3, "4")` will return "7" as bookmark.  In this case, the chaincode developer will not know the paginated range query is over if the empty string is the end flag of query, which is the problem I mentioned. 

This pull request modifies the returned bookmark of couchdb, making it consistent with leveldb.

core/ledger/kvledger/txmgmt/statedb/statecouchdb/statecouchdb.go
```diff
@@ -1098,6 +1098,9 @@ func (scanner *queryScanner) GetBookmarkAndClose() string {
 		retval = scanner.paginationInfo.bookmark
 	} else {
 		retval = scanner.queryDefinition.startKey
+		if retval == scanner.queryDefinition.endKey {
+			retval = ""
```
ale-linux, line 1102: Sorry, maybe I'm missing something but: we assign retval in line 1100 and then in line 1101 we test retval. How can the if ever be executed?

ale-linux (changes requested):

core/ledger/kvledger/txmgmt/statedb/statecouchdb/statecouchdb.go
Manphil: The if will be executed when the scanner is exhausted, refer to the code [statecouchdb.go#L603](https://github.com/hyperledger/fabric/blob/1b887760ae6a0111fcf6bfdc18dd4e06173035d5/core/ledger/kvledger/txmgmt/statedb/statecouchdb/statecouchdb.go#L603) and [statecouchdb.go#L606](https://github.com/hyperledger/fabric/blob/1b887760ae6a0111fcf6bfdc18dd4e06173035d5/core/ledger/kvledger/txmgmt/statedb/statecouchdb/statecouchdb.go#L606); The nextStartKey is assigned to endKey when there is no more results, refer to [couchdb.go#L910](https://github.com/hyperledger/fabric/blob/1b887760ae6a0111fcf6bfdc18dd4e06173035d5/core/ledger/kvledger/txmgmt/statedb/statecouchdb/couchdb.go#L910).

The purpose of the if statement is to update the retval to match the behavior of leveldb when the scanner is exhausted.

ale-linux: I see, thanks for the explanation. I would personally find the code a lot easier to read if you tested
```
scanner.queryDefinition.startKey == scanner.queryDefinition.endKey
```
but that's just a matter of taste

manish-sethi:
Thanks @Manphil for your submission and sorry for a delayed review. 

While this is good to have consistency across leveldb and couchdb but this change now would break backward compatibility and will require update of chaincodes/applications that would have been using this feature. So, suggested approach for now would be to update the documentation highlighting the current behavior. Please feel free if you want to submit this PR.

Regarding the change, we can discuss to update this behavior in 3.0 version if there is an agreement between maintainers - however, personally, I don't think this to be significant enough to break the backward compatibility.

denyeart:
I looked into this in some more detail and was able to reproduce, but only when an endKey is defined in the query. My opinion is that the lack of consistency across the databases is a very rare annoyance. Typically applications are tested for a certain state database. I agree with Manish that a fix would be more likely to break existing applications than help new users. Therefore I agree that the best approach is to document the difference. I've opened #3463 to update the docs. Let's close this PR in favor of the doc PR.

PR #3235, opened 2022-02-17
Gateway support for mutual TLS networks

### LINKED ISSUE #3234: Gateway support for client authentication (mutual TLS)
Currently the gateway doesn't pass the host peer's client certificate when making connections to other nodes in the network.
This needs to be added in order for the gateway to work when mutual TLS has been configured.

andrew-coleman:
To support client authentication (mTLS), the gateway needs to pass the host peer’s client certificate when making connections to the other nodes in the network.
If client authentication is not enabled, then these certs will be ignored.

Resolves #3234 

### DIFF
--- a/internal/peer/node/start.go
+++ b/internal/peer/node/start.go
@@ -825,6 +825,7 @@ func serve(args []string) error {
 				serverEndorser,
 				discoveryService,
 				peerInstance,
+				&serverConfig.SecOpts,
 				aclProvider,
 				coreConfig.LocalMSPID,
 				coreConfig.GatewayOptions,
--- a/internal/pkg/gateway/gateway.go
+++ b/internal/pkg/gateway/gateway.go
@@ -12,7 +12,8 @@ import (
 	"github.com/hyperledger/fabric/common/flogging"
 	"github.com/hyperledger/fabric/common/ledger"
 	"github.com/hyperledger/fabric/core/peer"
-	"github.com/hyperledger/fabric/gossip/common"
+	gdiscovery "github.com/hyperledger/fabric/gossip/discovery"
+	"github.com/hyperledger/fabric/internal/pkg/comm"
 	"github.com/hyperledger/fabric/internal/pkg/gateway/commit"
 	"github.com/hyperledger/fabric/internal/pkg/gateway/config"
 	"google.golang.org/grpc"
@@ -51,7 +52,7 @@ type LedgerProvider interface {
 }
 
 // CreateServer creates an embedded instance of the Gateway.
-func CreateServer(localEndorser peerproto.EndorserServer, discovery Discovery, peerInstance *peer.Peer, policy ACLChecker, localMSPID string, options config.Options) *Server {
+func CreateServer(localEndorser peerproto.EndorserServer, discovery Discovery, peerInstance *peer.Peer, secureOptions *comm.SecureOptions, policy ACLChecker, localMSPID string, options config.Options) *Server {
 	adapter := &peerAdapter{
 		Peer: peerInstance,
 	}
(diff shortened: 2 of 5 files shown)


internal/pkg/comm/config.go
```diff
@@ -164,6 +164,8 @@ type SecureOptions struct {
 	Certificate []byte
 	// PEM-encoded private key to be used for TLS communication
 	Key []byte
+	// The client certificate which, if specified, will be used instead of the PEM-encoded Certificate/Key pair
```
yacovm: But why would you do this? Why have two different ways to construct the SecureOptions? 
Until this PR, the entire code base used the SecureOptions as-is. Do we really have to expand this configuration structure and add another field that can be constructed from the existing fields?

integration/gateway/gateway_mtls_test.go
```diff
@@ -0,0 +1,200 @@
...
+	"github.com/hyperledger/fabric-protos-go/gateway"
+	"github.com/hyperledger/fabric-protos-go/peer"
+	"github.com/hyperledger/fabric/integration/nwo"
+	"github.com/hyperledger/fabric/protoutil"
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+	"github.com/tedsuo/ifrit"
+)
+
+var _ = Describe("GatewayService with mutual TLS network", func() {
```
yacovm: Do we really have to spin up an entire network, (with three Raft nodes....), to check that the gateway sends its TLS certificate to a peer or an orderer?

internal/pkg/gateway/gateway.go
```diff
@@ -76,13 +77,13 @@ func CreateServer(localEndorser peerproto.EndorserServer, discovery Discovery, p
 	return server
 }
 
-func newServer(localEndorser peerproto.EndorserClient, discovery Discovery, finder CommitFinder, policy ACLChecker, ledgerProvider LedgerProvider, localPKIID common.PKIidType, localEndpoint, localMSPID string, options config.Options) *Server {
+func newServer(localEndorser peerproto.EndorserClient, discovery Discovery, finder CommitFinder, policy ACLChecker, ledgerProvider LedgerProvider, localInfo gdiscovery.NetworkMember, localMSPID string, clientCert tls.Certificate, options config.Options) *Server {
```
yacovm: Can we not test this with a unit test? We have plenty of TLS unit tests scattered around that test mutual TLS. They are much cheaper than to spin up an entire Fabric network.

internal/pkg/comm/config.go
andrew-coleman: Fair point

integration/gateway/gateway_mtls_test.go
andrew-coleman: It was probably overkill, but it gave me confidence I'd really fixed the problem.  I've replaced it with a minimal unit test that spins up a grpc server with client auth enabled, and checks the gateway's endpointFactory can connect to it.

yacovm (approved):

internal/peer/node/start.go
```diff
@@ -825,6 +825,7 @@ func serve(args []string) error {
 				serverEndorser,
 				discoveryService,
 				peerInstance,
+				&serverConfig.SecOpts,
```
yacovm, line 828: Just one small thing - here you passed the server config, not the client config. 
This would mean that you do not honor the `tls.clientKey` and `tls.clientCert`. 

We probably need to check if the client TLS config is defined, and if so, then use it instead and only otherwise fallback on the server TLS config.

denyeart:

PR #3237, opened 2022-02-18
Fix firewall warnings for integration tests on Mac

andrew-coleman:
When running integration tests locally on MacOS, each test causes a number of popup boxes containing the text
‘Do you want the application “peer” to accept incoming network connections?”.
Although this can be ignored, each popup takes the (cursor) focus away from the current window, and so is annoying when trying to do anything else.
This is caused because each test is launching one or more peer instances which get compiled on demand and appears to the firewall as a new executable.
The fix for this is to ensure that all listen addresses are explicitly set as localhost.  In this case, the chaincodeListenAddress host can be set to 127.0.0.1 rather than 0.0.0.0


denyeart:
@andrew-coleman Do you want to take this one out of Draft?

denyeart (approved):

Issue #3238, opened 2022-02-18
No TLSHANDSHAKETIMESHIFT for intermediate Ca Certificates


afrancoc2000:
Hy everyone,

I'm having a problem with expired certificates, I'm running a cluster with several channels and I updated the certificates on time for most of them, but I forgot about a channel that was created and used only during the first initialization of the cluster.
 
So today in my test environment the orderers got restarted and now they don't start, my production environment has the same problem but as they haven't been restarted they are running as if nothing has changed, for now.
 
So, I did as it is said, in this [issue](https://jira.hyperledger.org/browse/FAB-18384) and changed these environment variables:
 
- ORDERER_GENERAL_AUTHENTICATION_NOEXPIRATIONCHECKS: "true"
- ORDERER_GENERAL_CLUSTER_TLSHANDSHAKETIMESHIFT: 96h
- ORDERER_GENERAL_TLS_TLSHANDSHAKETIMESHIFT: 96h
- ORDERER_OPERATIONS_TLS_TLSHANDSHAKETIMESHIFT: 96h
- ORDERER_ADMIN_TLS_TLSHANDSHAKETIMESHIFT: 96h
 
The only handshake time shift still missing is the one from kafta that I don't use, I'm using Raft, but still, I'm getting this error trying to run the orderers:
 
PANI 00d Error creating ledger resources: error creating channelconfig bundle: initializing channelconfig failed: could not create channel Orderer sub-group config: setting up the MSP manager failed: CA Certificate is not valid, (SN: 512917139581241201287378615618194486727802882): could not obtain certification chain: the supplied identity is not valid: x509: certificate has expired or is not yet valid: current time 2022-02-17T14:28:20Z is after 2022-02-16T16:41:01Z
 
I published the issue in the message channel and with the help of @yacovm found that there's no TLS Handshake time shift for the CA certificates.

Here's the stack trace:

    Thu, Feb 17 2022 9:28:20 am	2022-02-17 14:28:20.681 UTC [localconfig] completeInitialization -> WARN 001 General.GenesisFile should be replaced by General.BootstrapFile
    Thu, Feb 17 2022 9:28:20 am	2022-02-17 14:28:20.681 UTC [localconfig] completeInitialization -> INFO 002 Kafka.Version unset, setting to 0.10.2.0
    Thu, Feb 17 2022 9:28:20 am	2022-02-17 14:28:20.681 UTC [orderer.common.server] prettyPrintStruct -> INFO 003 Orderer config values:
[...]

PR #3242, opened 2022-02-20
Ignore expired CA/TLS CA certs on msp init (#3238)

afrancoc2000:
Signed-off-by: Ana Maria Franco <afrancoc2000@gmail.com>

Adding this line guarantees that expired CA certificates wont prevent orderer initialization as happened in [Issue #3238](https://github.com/hyperledger/fabric/issues/3238)

#### Type of change

- Bug fix

#### Description

Adding this line guarantees that expired CA certificates wont prevent orderer initialization as happened in [Issue #3238](https://github.com/hyperledger/fabric/issues/3238)

#### Related issues

[Issue #3238](https://github.com/hyperledger/fabric/issues/3238)

### DIFF
--- a/msp/mspimplsetup.go
+++ b/msp/mspimplsetup.go
@@ -493,6 +493,7 @@ func (msp *bccspmsp) setupTLSCAs(conf *m.FabricMSPConfig) error {
 			return errors.WithMessagef(err, "CA Certificate problem with Subject Key Identifier extension, (SN: %x)", cert.SerialNumber)
 		}
 
+		opts.CurrentTime = cert.NotBefore.Add(time.Second)
 		if err := msp.validateTLSCAIdentity(cert, opts); err != nil {
 			return errors.WithMessagef(err, "CA Certificate is not valid, (SN: %s)", cert.SerialNumber)
 		}
--- a/msp/mspimplsetup_test.go
+++ b/msp/mspimplsetup_test.go
@@ -65,6 +65,21 @@ vpx58pjg15TqKgrZF2h+TJ5jFa48O1wBvtMhP8WL6/6O+NjOEP56UnXPGie/3HLC
 yvhEkMILRkzGUfd091cpuNxd+aGA37mZbwc+8UBpYbZFhq3NORL8zSxUQLzm1NcV
 U98sznvJPRCkRiwYp5L9C5Xq72CHG/3M6cmoN0Cl0xjZicfpfnZSA/ix
 -----END CERTIFICATE-----`
+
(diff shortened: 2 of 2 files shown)


msp/mspimplsetup.go
```diff
@@ -493,6 +493,7 @@ func (msp *bccspmsp) setupTLSCAs(conf *m.FabricMSPConfig) error {
 			return errors.WithMessagef(err, "CA Certificate problem with Subject Key Identifier extension, (SN: %x)", cert.SerialNumber)
 		}
 
+		opts.CurrentTime = cert.NotBefore.Add(time.Second)
```
yacovm, line 496: what about line 155? Can you also add a similar option there?

yacovm:
Please `git commit -s --amend` to add your signed-off-by

yacovm:
@adecaro

msp/mspimplsetup.go
afrancoc2000: Sure let me give it a look

afrancoc2000: The thing with line 155 is that there's no validation of the expiration date after defining the msp.opts there, the validation is on the finalizeSetupCAs method, on line 260, and when I tested finalizeSetupCAs with an expired certificate, using unit tests, they passed. So I'm not sure the current time override is needed for the CA, as the unit test for finalizeSetupCAs didn't failed, or if there's an extra validation missing for the CA, where that validation should be happening.

yacovm: Oh you're right. I didn't notice we're not verifying there at all. 

It doesn't fail your second test because `finalizeSetupCAs`  passes through `getValidityOptsForCert` which has the time override.

msp/mspimplsetup_test.go
```diff
@@ -121,6 +147,20 @@ func TestCAValidation(t *testing.T) {
 		gt.Expect(err).NotTo(gomega.HaveOccurred())
 	})
 
+	t.Run("ExpiredCert", func(t *testing.T) {
```
yacovm, line 150: Why do we need this test? It's irrelevant to the code change.

yacovm:
One more thing - can you please make the pull request be against the `main` branch and not `release-2.4` branch? 
We merge to `main` and then backport to earlier branches, not the other way around.

msp/mspimplsetup_test.go
afrancoc2000: Ok, I will delete it, it was my way of checking that there were no issues with the CA

afrancoc2000:
Closing it to create just one

PR #3283, opened 2022-03-13
Refactor of ChaincodeEvents service implementation to support resume

bestbeforetoday:
Enables the client to (optionally) specify an AfterTransactionId property in addition to a start block number when requesting chaincode events, which causes chaincode events up to that transaction ID (inclusive) to be ignored and not returned to the client. This supports resume of chaincode event listening on client reconnect without duplicating or missing any events.


andrew-coleman:
As discussed with @bestbeforetoday if the TxID is supplied, then the block number is not required, and can be ignored.  The server can lookup which block contains that transaction.  This makes no difference to the client code, so agreed this can be done in a followup PR.

andrew-coleman (approved):

Issue #2980, opened 2021-10-08
Ledger Utility Identifytxs - Main Base Functionality 2


jcastrence:
Extend functionality of tool to:
- Read list of diffRecords as json file (output from compare tool)
- Write transaction data to json of matched input keys
- Generate the output data in a file

Issue #2979, opened 2021-10-08
Ledger Utility Identifytxs - Main Base Functionality 1


jcastrence:
Command line parsing for new tool:
- Establish new command (identifytxs) with one required arg (ledgerid) and one optional arg (filesysPath)

Tool should be able to:
- Establish instance of peer's block store and iterate through list of blocks
- Iterate through transactions in each block and print relevant data
- Iterate through the write set of each transaction

Issue #2788, opened 2021-07-26
Ledger utility to identify transactions associated with in doubt state


denyeart:
GOAL
Identify transactions that have caused divergent state.

INPUT
Set of keys in question (JSON output from state comparison utility)

APPROACH
Traverse the block store on each peer (separately) to identify and extract transactions that wrote mismatched keys

OUTPUT
Set of transactions that wrote the keys (in human readable format)
All tx validation codes from the respective blocks
Block commit hashes from the respective blocks (available on channels created in v1.4.2 and later)

jcastrence:
In bold are assumptions and questions on the implementation of the tool

GOAL 
Identify potentially divergent transactions

INPUT 
Set of keys in question (from file utility or otherwise)

**Format: json file holding one diffRecord list. Only the JSON elements "namespace" and key" will be parsed. Therefore it will also be possible to hand craft a JSON array of "namespace" and "key" pairs that the utility will use as input.**

APPROACH 
Traverse the block store on each peer (separately) to identify and extract transactions that wrote mismatched keys

**Block store: list of blocks for peer, can be accessed from peer file system (no db access required)
Block: list of transactions and metadata on a block
Functions / files relevant to accessing the block store?
Found so far: blkstorage package, blockstore_provider.go (BlockStoreProvider type), leveldbhelper.Provider, fileledger package, fileledger/factory.go
What is meant by separately? It means that the utility would be run against peer1's block store, and then peer2's block store, so that the user can inspect and compare the outputs. The assumption is that you can only access one peer's file system at a time.
Does this mean the tool is ran twice, once for each peer? Or that the tool is ran once and checks both peers sequentially?
Functions / files relevant to transaction structure?
How are txs ordered?**

OUTPUT 
Set of transactions that wrote the keys (in human readable format) All tx validation codes from the respective blocks Block commit hashes from the respective blocks (available on channels created in v1.4.2 and later)

**File types for the different outputs?
JSON, log back to user, or both like compare utility? In success case, output to JSON. In error case, log back to user.
All output in the same file or different file for each (txs, tx codes, hashes)? I think we could output to a single file, with JSON in the following structure:
namespace
..key
....transaction id
......transaction number (in block)
......transaction validation status
......transaction content (in JSON form, not sure if we have a preferred JSON tx structure already?)
......block number
......block commit hash
[...]

PR #3285, opened 2022-03-15
Ledgerutil identifytxs

jcastrence:
#### Type of change

- New feature

#### Description
- The Identify Transactions tool (identifytxs) is a new addition to the Fabric ledger utilities that adds another layer of troubleshooting to the first Fabric ledger utilities tool, Compare Snapshots. After a user has obtained the list of different records between two snapshots using Compare Snapshots, the user can then pass this output to the Identify Transactions tool to obtain a list of relevant transactions that could potentially be the cause of the initial ledger divergence.
- This PR builds on top of a previous commit by @shimos and extends their work into a complete Fabric ledger utilities tool.

#### Related issues

- #2788
- #2979 
- #2980


internal/ledgerutil/identifytxs.go
```diff
@@ -0,0 +1,479 @@
...
+				return 0, 0, err
+			}
+			ch, err := protoutil.UnmarshalChannelHeader(pl.Header.ChannelHeader)
+			if err != nil {
+				return 0, 0, err
+			}
+			txID := ch.GetTxId()
+
+			// Extract write set from transaction then iterate through transaction write set
+			res, err := protoutil.GetActionFromEnvelope(envBytes)
```
denyeart: This line fails for any config block, it returns error:
`Ledger Identify Transactions Error: error unmarshalling ProposalResponsePayload: proto: can't skip unknown wire type 7`

Before doing this you need to make sure the transaction is an endorser transaction. See an example of this here:
https://github.com/hyperledger/fabric/blob/main/core/ledger/kvledger/txmgmt/validation/batch_preparer.go#L230

internal/ledgerutil/identifytxs.go
```diff
@@ -0,0 +1,479 @@
...
+							KeyWrite:  string(write.GetValue()),
+						}
+						// Capture new transaction entry
+						ckMapEmpty, err := captureTx(inputKeyMap, ck, entry)
+						if err != nil {
+							return 0, 0, err
+						}
+						// No more keys left
+						if ckMapEmpty {
+							return minBlockHeight, blockIndex, nil
```
denyeart, line 387: In my trial I had several transactions for a key. But only one transaction gets written to txlist1.json. It appears that ckMapEmpty gets set to true after the first 'hit' and it returns here. Is this the intended design? Why return after the first hit? Don't we want to see all the transactions for the key?

internal/ledgerutil/identifytxs.go
jcastrence: Would this tool only be concerned with endorser transactions? So if it encounters a non endorser transaction, it should skip it?

denyeart: Right.

internal/ledgerutil/identifytxs/identifytxs.go
```diff
@@ -0,0 +1,484 @@
...
+		}
+		// New entry, add entry and height
+		blockNum, txNum := r.GetLaterHeight()
+		// Only namespace and key were provided, record doesn't contain height. Set to "infinite" height to iterate entire block store.
+		if blockNum == 0 && txNum == 0 {
+			blockNum = math.MaxUint64
+			txNum = math.MaxUint64
+		}
+		// Check for max height
+		if blockNum >= maxBlock && txNum >= maxTx {
```
denyeart: I don't think you can check the block and tx heights independently.
e.g. If your prior max height was maxBlock=3, maxTx=3, and the current is blockNum=4, txNum=1, we have a new max height but this if statement would not identify it.
You need to check the blockNum first, and if they are equal THEN check the txNum.

internal/ledgerutil/identifytxs/identifytxs.go
```diff
@@ -0,0 +1,484 @@
...
+	}
+
+	// Set up block store and block iterator in preparation for traversal
+	blockStoreProvider, err := getBlockStoreProvider(fsPath)
+	if err != nil {
+		return 0, 0, err
+	}
+	defer blockStoreProvider.Close()
+
+	blockStore, err := blockStoreProvider.Open(records.Ledgerid)
```
denyeart, line 83: Before you Open() you should call blockStoreProvider.Exists() to make sure it exists. Otherwise if user indicates a non-existing channel in the JSON the Open() call will actually create it in the ledger.

cmd/ledgerutil/main.go
```diff
@@ -70,5 +82,29 @@ func main() {
...
+
+		firstBlock, lastBlock, err := identifytxs.IdentifyTxs(*snapshotDiffsPath, *blockStorePath, *outputDirId)
+		if err != nil {
+			fmt.Printf("%s%s\n", identifytxsErrorMessage, err)
+			os.Exit(1)
+		}
+		fmt.Print("\nSuccessfully ran identify transactions tool. ")
+		if firstBlock <= lastBlock {
+			fmt.Printf("Transactions were checked between blocks %d and %d.", firstBlock, lastBlock)
+			os.Exit(2)
```
denyeart: No need for Exit(2) upon success.

internal/ledgerutil/identifytxs/identifytxs.go
jcastrence: Good catch

cmd/ledgerutil/main.go
jcastrence: This is to match the form of the compare tool. Should I then Exit(2) for both?

denyeart: Successful execution should result in exit of 0, and no need for explicit Exit().
I would do that here.

I guess the Compare returns Exit(2) at line 83 in order to differentiate with no differences found which returns 0. That is non-standard but I guess justified and therefore can stay as-is.

Finally, at line 107 I would Exit(1) since it can be considered an error case.

denyeart (approved):

PR #3292, opened 2022-03-19
Change membership documentation about public keys

etlow:
#### Type of change
- Documentation update

Identity can be proven just by signing the transaction and verifying with the public key on the certificate, proving identity is not where the MSP comes in, this is more related to how CAs issue the key-pair.

> The private key is used to produce a signature on a transaction that only the corresponding public key, that is part of an MSP, can match.

The sentence does not say anything about whether this public key matches invalid signatures, it only says that given that particular signature, only that particular public key can match. It should go along the lines of "only a valid signature will work". The sentence has been removed as I have adapted the paragraph.

I do not have experience with Fabric; these changes are based off my reading of the documentation. Am I right to assume that the member's public keys do not need to be added to the organization's MSP in the current version? Please correct me if I am wrong. Thank you.

### DIFF
--- a/docs/source/membership/membership.md
+++ b/docs/source/membership/membership.md
@@ -7,7 +7,7 @@ Note: this topic describes a network that does not use a "system channel", a cha
 Because Fabric is a permissioned network, blockchain participants need a way to prove their identity to the rest of the network in order to transact on the network. If you've read through the documentation on [Identity](../identity/identity.html)
 you've seen how a Public Key Infrastructure (PKI) can provide verifiable identities through a chain of trust. How is that chain of trust used by the blockchain network?
 
-Certificate Authorities issue identities by generating a public and private key which forms a key-pair that can be used to prove identity. Because a private key can never be shared publicly, a mechanism is required to enable that proof which is where the MSP comes in. For example, a peer uses its private key to digitally sign, or endorse, a transaction.  The MSP on the ordering service contains the peer's public key which is then used to verify that the signature attached to the transaction is valid. The private key is used to produce a signature on a transaction that only the corresponding public key, that is part of an MSP, can match. Thus, the MSP is the mechanism that allows that identity to be trusted and recognized by the rest of the network without ever revealing the member’s private key.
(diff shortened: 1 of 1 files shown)


docs/source/membership/membership.md
```diff
@@ -7,7 +7,7 @@ Note: this topic describes a network that does not use a "system channel", a cha
...
 you've seen how a Public Key Infrastructure (PKI) can provide verifiable identities through a chain of trust. How is that chain of trust used by the blockchain network?
 
-Certificate Authorities issue identities by generating a public and private key which forms a key-pair that can be used to prove identity. Because a private key can never be shared publicly, a mechanism is required to enable that proof which is where the MSP comes in. For example, a peer uses its private key to digitally sign, or endorse, a transaction.  The MSP on the ordering service contains the peer's public key which is then used to verify that the signature attached to the transaction is valid. The private key is used to produce a signature on a transaction that only the corresponding public key, that is part of an MSP, can match. Thus, the MSP is the mechanism that allows that identity to be trusted and recognized by the rest of the network without ever revealing the member’s private key.
```
yacovm, line 10: > The MSP on the ordering service contains the peer's public key which is then used to verify that the signature attached to the transaction is valid

This is completely and utterly false :) 

> The private key is used to produce a signature on a transaction that only the corresponding public key, that is part of an MSP, can match

While this may be true empirically, it has absolutely no importance as that's not a security requirement.
Instead, what we need is that no one can forge a signature on a message for a specific public key, but not that there can not be another public key that the signature verifies under. For instance, if there is a scheme with a "bad" public key that any string is a valid signature under that public key, the general signature scheme may still be safe if that "bad" public key is avoided with overwhelming probability.

docs/source/membership/membership.md
```diff
@@ -7,7 +7,7 @@ Note: this topic describes a network that does not use a "system channel", a cha
...
+Certificate Authorities issue identities by generating a public and private key which forms a key-pair that can be used to prove identity. This identity needs a way to be recognized by the network, which is where the MSP comes in. For example, a peer uses its private key to digitally sign, or endorse, a transaction. The MSP on the ordering service is used to check that the peer is allowed to endorse the transaction. The public key from the peer's certificate is then used to verify that the signature attached to the transaction is valid. Thus, the MSP is the mechanism that allows that identity to be trusted and recognized by the rest of the network.
```
yacovm: > The MSP on the ordering service is used to check that the peer is allowed to endorse the transaction.

This is false, the orderer does not pry into the transaction to that extent. 
Can you just remove that sentence?

etlow: Thanks for sharing your insight. If I remove that sentence, there is no link from identity to MSP. Would it be better if I changed it to "MSP on the channel" or just "MSP"?

yacovm: Yeah, good idea.

yacovm (approved):

Issue #3278, opened 2022-03-09
Graphics and diagrams updates - docs


joshhus:
For example, this one is pending / working: 

https://hyperledger-fabric.readthedocs.io/en/release-2.4/getting_started.html

![image](https://user-images.githubusercontent.com/16542757/157545894-b070fee2-75b7-4296-9c02-2244311268c2.png)

Others can be added and worked in this issue.

joshhus:
https://github.ibm.com/WCE-ID/Content-Design/issues/386

joshhus:
https://github.com/hyperledger/fabric/issues/3278

AliakbarETH:
Hi @joshhus I have seen issue number #3278 opened whereas at the bottom the issue is closed.

Is the work done or needs some improvements?

Let me know.

Regards

denyeart:
It is resolved now, closing.

PR #3301, opened 2022-03-25
new stack image

joshhus:

Existing software stack image needs updates for both legibility and text.

#### Type of change

- Documentation update

#### Description

Just a new graphic image for legibility and updated text.

#### Related issues

https://github.com/hyperledger/fabric/issues/3278


denyeart:
@joshhus The image to replace is https://github.com/hyperledger/fabric/blob/main/docs/source/getting_started_image2.png

joshhus:
> @joshhus The image to replace is https://github.com/hyperledger/fabric/blob/main/docs/source/getting_started_image2.png

Yes I renamed it but somehow the .rst file didn't get included. Adding now.

denyeart:
@joshhus Please also delete the old image, we don't want to build up extra cruft.

Also, look at the new rendering - https://hyperledger-fabric--3301.org.readthedocs.build/en/3301/getting_started.html
It looks like a mistake because the 5 bullets and the 5 stack items are all next to each other, but the order is reversed between them. If we are going to have the stack directly next to the bullets, we should reverse the order of the bullets so that they line up well. Typically in these architecture stacks the application is on top and foundation on the bottom, so I agree with the diagram, just need to re-order the bullets to match.

joshhus:
> @joshhus Please also delete the old image, we don't want to build up extra cruft.
> 
> Also, look at the new rendering - https://hyperledger-fabric--3301.org.readthedocs.build/en/3301/getting_started.html It looks like a mistake because the 5 bullets and the 5 stack items are all next to each other, but the order is reversed between them. If we are going to have the stack directly next to the bullets, we should reverse the order of the bullets so that they line up well. Typically in these architecture stacks the application is on top and foundation on the bottom, so I agree with the diagram, just need to re-order the bullets to match.

Agreed, I hadn't gone back to check the built version yet, will do.

joshhus:
branch fully out of sync with upstream now. easier to just stash and start over with new PR.

joshhus:
@denyeart Okay image looks decent now. If I edited down the text too much, LMK. Previously it was too verbose IMO.

PR #3316, opened 2022-03-31
Update private data tests to use raft orderer

jt-nti:
#### Type of change

- Test update

#### Description

Update private data tests to use raft orderer since the solo orderer has been deprecated

### DIFF
--- a/integration/nwo/standard_networks.go
+++ b/integration/nwo/standard_networks.go
@@ -227,6 +227,24 @@ func MinimalRaft() *Config {
 	return config
 }
 
+// Like FullSolo, FullEtcdRaft is a configuration with two organizations and two peers per org.
+func FullEtcdRaft() *Config {
+	config := FullSolo()
+
+	config.Consensus.Type = "etcdraft"
+	config.Profiles = []*Profile{{
+		Name:     "SampleDevModeEtcdRaft",
+		Orderers: []string{"orderer"},
+	}, {
+		Name:          "TwoOrgsChannel",
+		Consortium:    "SampleConsortium",
+		Organizations: []string{"Org1", "Org2"},
+	}}
+	config.SystemChannel.Profile = "SampleDevModeEtcdRaft"
+
+	return config
+}
+
 func ThreeOrgRaft() *Config {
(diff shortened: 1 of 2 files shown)


integration/nwo/standard_networks.go
```diff
@@ -227,6 +227,24 @@ func MinimalRaft() *Config {
 	return config
 }
 
+// Like FullSolo, FullEtcdRaft is a configuration with two organizations and two peers per org.
+func FullEtcdRaft() *Config {
+	config := FullSolo()
```
yacovm, line 232: Isn't it a bit awkward to still call the solo constructor? 
Why not just call `MinimalRaft()` and change it accordingly?

integration/pvtdata/pvtdata_test.go
```diff
@@ -946,7 +946,7 @@ func initThreeOrgsSetup(removePeer1 bool) *nwo.Network {
 	client, err := docker.NewClientFromEnv()
 	Expect(err).NotTo(HaveOccurred())
 
-	config := nwo.FullSolo()
+	config := nwo.FullEtcdRaft()
```
yacovm, line 949: it seems there is mutation of the config afterwards, so why not just use the `MinimalRaft()` ?

jt-nti: Thanks for the review @yacovm, it's definitely a bit awkward but I was trying to minimise the changes- since it gets mutated in the private data tests anyway, do you think it would be better to leave the shared networks alone and just handle the private data specific network config entirely in `pvtdata_test.go`?

yacovm: Honestly I think that since these are ginkgo integration tests, and they're not used by consumers, it's easiest to just leave them be, and when we remove Solo, we'll clean up this place as well since we won't have a choice.

jt-nti:
Closing based on feedback from maintainers. (I only opened this PR while attempting to get familiar with the private data tests however I still think it would be better to gradually move tests off a deprecated orderer before the solo orderer is finally removed.)

PR #3324, opened 2022-04-09
Use actual address of the deadMembers2Expire slice variable

C0rWin:
Instead of passing the address of the for-loop variable, pass address of the 
actual slice element.

### DIFF
--- a/gossip/discovery/discovery_impl.go
+++ b/gossip/discovery/discovery_impl.go
@@ -735,9 +735,9 @@ func (d *gossipDiscoveryImpl) expireDeadMembers(dead []common.PKIidType) {
 
 	d.lock.Unlock()
 
-	for _, member2Expire := range deadMembers2Expire {
-		d.logger.Warning("Closing connection to", member2Expire)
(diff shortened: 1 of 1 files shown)


yacovm:
Why do you think it's a bug? While it's true the pointer is the same pointer at each iteration of the for loop, we never store it anywhere and always just take the inner value of the PKI-ID. 

Look at [this](https://go.dev/play/p/bhvcLw2fcEs) code sample where we print 1 to 5 

And if it's a bug, then can we have a test that demonstrates it?

C0rWin:
> Why do you think it's a bug? While it's true the pointer is the same pointer at each iteration of the for loop, we never store it anywhere and always just take the inner value of the PKI-ID.
> 
> Look at [this](https://go.dev/play/p/bhvcLw2fcEs) code sample where we print 1 to 5
> 
> And if it's a bug, then can we have a test that demonstrates it?

well, I was not saying this is a bug, while in this case, this might not be severe though still considered a bad practice, hence suggested PR.

yacovm (approved):

Issue #3337, opened 2022-04-18
Architecture difference between Step 13 "RUN make ccaasbuilder" and Step 21


Aniket21mathur:
Step 13('RUN make ccaasbuilder') of the installation process (`make dist-clean all`) is producing binaries with path having a `linux-arm64` architecture.

```bash
Step 13/23 : RUN make ccaasbuilder
 ---> Running in 12bebcc48412
rm -rf release/linux-arm64/bin/ccaas_builder
cd ccaas_builder && go test -v ./cmd/detect && go build -o ../release/linux-arm64/bin/ccaas_builder/bin/ ./cmd/detect/
go: downloading github.com/pkg/errors v0.9.1
?       github.com/hyperledger/fabric/ccaas_builder/cmd/detect    [no test files]
cd ccaas_builder && go test -v ./cmd/build && go build -o ../release/linux-arm64/bin/ccaas_builder/bin/ ./cmd/build/
go: downloading github.com/otiai10/copy v1.6.0
?       github.com/hyperledger/fabric/ccaas_builder/cmd/build    [no test files]
cd ccaas_builder && go test -v ./cmd/release && go build -o ../release/linux-arm64/bin/ccaas_builder/bin/ ./cmd/release/
?       github.com/hyperledger/fabric/ccaas_builder/cmd/release    [no test files]
Removing intermediate container 12bebcc48412
 ---> b86df7e4d9a9
```

While the Dockerfile involved in Step21 is reading from a hardcoded `linux-amd64` architecture path and hence running into an error->

```
Step 21/23 : COPY --from=peer /go/src/github.com/hyperledger/fabric/release/linux-amd64/bin/ccaas_builder/bin/ /opt/hyperledger/ccaas_builder/bin/
COPY failed: stat go/src/github.com/hyperledger/fabric/release/linux-amd64/bin/ccaas_builder/bin/: file does not exist
make: *** [build/images/peer/.dummy-arm64-2.5.0-snapshot-5df3d8568] Error 1
```

The architecture here(https://github.com/hyperledger/fabric/blob/main/images/peer/Dockerfile#L38) should change depending on the building platform.

Param-S:
Issue discussed in discord : https://discord.com/channels/905194001349627914/905205541326975046/965569733955035136

denyeart:
@jkneubuh Is this resolved with the new image and binary build approach?

PR #3315, opened 2022-03-31
CCAAS Builders

mbwhite:
- adjust the makefile to have targets for linux-amd64/darwin-amd64/windows-amd64
- adjust the packaging as suggested in issue 3286

Signed-off-by: Matthew B White <whitemat@uk.ibm.com>

#### Type of change

- Bug fix
- New feature
- Improvement (improvement to code, performance, etc)
- Test update
- Documentation update

PR #3338, opened 2022-04-18
Updated dockerfile with new build arg

Param-S:
#### Type of change

- Bug fix

#### Description

Updated multistage peer dockerfile with ccbuilder binaries location. It becomes 2 step process to copy the binaries Step 1. copy the files into temporary location, Step 2. copy the files into correction location into docker image

#### Related issues

#3337 

### DIFF
--- a/images/peer/Dockerfile
+++ b/images/peer/Dockerfile
@@ -27,6 +27,7 @@ FROM golang as peer
 ARG GO_TAGS
 RUN make peer GO_TAGS=${GO_TAGS}
 RUN make ccaasbuilder
+RUN export MARCH=$(go env GOOS)-$(go env GOARCH) && mkdir -p /tmp/ccaas_builder/bin/ && cp /go/src/github.com/hyperledger/fabric/release/${MARCH}/bin/ccaas_builder/bin/* /tmp/ccaas_builder/bin/
 
 FROM peer-base
 ENV FABRIC_CFG_PATH /etc/hyperledger/fabric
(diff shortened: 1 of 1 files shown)


images/peer/Dockerfile
```diff
@@ -29,12 +29,13 @@ RUN make peer GO_TAGS=${GO_TAGS}
...
 FROM peer-base
+ARG MARCH
 ENV FABRIC_CFG_PATH /etc/hyperledger/fabric
 VOLUME /etc/hyperledger/fabric
 VOLUME /var/hyperledger
 COPY --from=peer /go/src/github.com/hyperledger/fabric/build/bin /usr/local/bin
 COPY --from=peer /go/src/github.com/hyperledger/fabric/sampleconfig/msp ${FABRIC_CFG_PATH}/msp
 COPY --from=peer /go/src/github.com/hyperledger/fabric/sampleconfig/core.yaml ${FABRIC_CFG_PATH}/core.yaml
-COPY --from=peer /go/src/github.com/hyperledger/fabric/release/linux-amd64/bin/ccaas_builder/bin/ /opt/hyperledger/ccaas_builder/bin/
+COPY --from=peer /go/src/github.com/hyperledger/fabric/release/${MARCH}/bin/ccaas_builder/bin/ /opt/hyperledger/ccaas_builder/bin/
```
C0rWin: I do not think that will work since line 29 is running inside the staged docker image, and $MARCH will point to host architecture rather than the correct one.

Param-S: Agree, It should be set to stage docker image specific os-arch.  i think, the value should be read from the previous stage where ccaasbuilder is compiled.

denyeart:
@mbwhite Can you help with this one?

mbwhite:
I *think* that maybe this would be superseded by https://github.com/hyperledger/fabric/pull/3315

mbwhite:
@Param-S - I think PR https://github.com/hyperledger/fabric/pull/3315  should resolve problems here as well.. 
Can you say what issue you where hitting?

C0rWin:
> @Param-S - I think PR #3315 should resolve problems here as well.. Can you say what issue you where hitting?

The problem here is that #3315 doesn't really help for M1 and arm64-based architecture, because currently, the copy path is hardcoded into Dockerfile:

```COPY --from=peer /go/src/github.com/hyperledger/fabric/release/linux-amd64/builders/ccaas/bin/ /opt/hyperledger/ccaas_builder/bin/```

So on M1 and arm64, the folder would be `linux-arm64,` but the script doesn't capture it.

C0rWin:
Given that #3557  merged, we can close this one. Thanks, @Param-S.

PR #3339, opened 2022-04-18
Log warning for a non existent block delivery failure

### LINKED ISSUE #3320: Improve error message when deliver service is asked for a block that doesn't yet exist on the node
Improve error message when deliver service is asked for a block that doesn't yet exist on the node.

This is all we get for now:
`[common.deliver] deliverBlocks -> ERRO 18d59 [channel: channel1] Error reading from channel, cause was: NOT_FOUND`

It is enough for Fabric developers to know what happened, but not for users or support personnel.

Aniket21mathur:
#### Type of change

#### Description

Improve error message when delivery service is asked for a block that doesn't yet exist on the node to make it more clear for users or support personnel.

#### Related issues
Fixes https://github.com/hyperledger/fabric/issues/3320

### DIFF
--- a/common/deliver/deliver.go
+++ b/common/deliver/deliver.go
@@ -286,6 +286,7 @@ func (h *Handler) deliverBlocks(ctx context.Context, srv *Server, envelope *cb.E
 	for {
 		if seekInfo.Behavior == ab.SeekInfo_FAIL_IF_NOT_READY {
 			if number > chain.Reader().Height()-1 {
+				logger.Warningf("[channel: %s] Block %d not found, block number greater than chain length bounds", chdr.ChannelId, number)
 				return cb.Status_NOT_FOUND, nil
 			}
 		}
@@ -313,7 +314,7 @@ func (h *Handler) deliverBlocks(ctx context.Context, srv *Server, envelope *cb.E
 		}
 
 		if status != cb.Status_SUCCESS {
-			logger.Errorf("[channel: %s] Error reading from channel, cause was: %v", chdr.ChannelId, status)
+			logger.Warningf("[channel: %s] Error reading from channel, cause was: %v", chdr.ChannelId, status)
 			return status, nil
 		}
 


common/deliver/deliver.go
```diff
@@ -313,7 +313,11 @@ func (h *Handler) deliverBlocks(ctx context.Context, srv *Server, envelope *cb.E
 		}
 
 		if status != cb.Status_SUCCESS {
-			logger.Errorf("[channel: %s] Error reading from channel, cause was: %v", chdr.ChannelId, status)
+			if status == cb.Status_NOT_FOUND {
+				logger.Errorf("[channel: %s] Error reading from channel, block [%d] not found.", chdr.ChannelId, block.Header.Number)
```
yacovm: these log entries should be warnings, not errors.

Aniket21mathur: Updated the commit.

common/deliver/deliver.go
```diff
@@ -313,7 +313,11 @@ func (h *Handler) deliverBlocks(ctx context.Context, srv *Server, envelope *cb.E
 		}
 
 		if status != cb.Status_SUCCESS {
-			logger.Errorf("[channel: %s] Error reading from channel, cause was: %v", chdr.ChannelId, status)
+			if status == cb.Status_NOT_FOUND {
+				logger.Warningf("[channel: %s] Error reading from channel, block [%d] not found.", chdr.ChannelId, block.Header.Number)
```
yacovm: Isn't `block`  nil here? If you cannot find the block then how do you print its number? :)

Aniket21mathur: Ohh yes thanks for pointing out, my fault. I missed that the `iterator` will return a `nil` block value. Should we change the message simply to-

```go
logger.Warningf("[channel: %s] Error reading from channel, block not found.", chdr.ChannelId)
```

Also can you please help me in understanding how to reproduce this warning locally? I am new to fabric and this will help in testing other changes in the future apart from the unit tests.

Thanks!

yacovm: You don't need to reproduce it, just make a unit test.

yacovm: Look at the the other unit tests in the same package

yacovm: I think that maybe what we really wanted to change is the behavior in line 287.

Aniket21mathur: > I think that maybe what we really wanted to change is the behavior in line 287.

Makes sense. But I am not able to understand one thing if we are already checking in the conditional at 287 that the current block number is within the chain length, then how did we end up in #3320 ? The code flow should have not reached the `if` statement at 315 in case the current block `number` is more than the chain length.

I can think of 2 possibilities as of now->

-  `seekInfo.Behavior` might not always be equal to `ab.SeekInfo_FAIL_IF_NOT_READY` in cases where the current block `number` is out of range and we end up with the current warning.

```go
		if seekInfo.Behavior == ab.SeekInfo_FAIL_IF_NOT_READY {  // might not be true always in case number > chain.Reader().Height()-1
			if number > chain.Reader().Height()-1 {
				return cb.Status_NOT_FOUND, nil
			}
		}
```

- `cursor.Next()` fails to retrieve the next block for some reason even though it is available on the chain.
```go
		go func() {
			block, status = cursor.Next()
			close(iterCh)
		}()
```

Aniket21mathur: > Look at the the other unit tests in the same package

Sure, I will have a look

yacovm: So you have two behaviors of the block iterator:
- **SeekInfo_BLOCK_UNTIL_READY**: Makes you wait until the block exists
- **SeekInfo_FAIL_IF_NOT_READY**: Returns an error if the block doesn't exist

So I think the place where you added code - handles the case for the block not being found **and** the behavior is **SeekInfo_BLOCK_UNTIL_READY** because if the behavior is **SeekInfo_FAIL_IF_NOT_READY**, we will not get to that line and will abort early in line 289.

Aniket21mathur: Cool, so should we keep the current changes and add a warning to the code block at line 289 as well?

Also, I am not able to understand how the `deliverBlocks` function handles the two behaviors of the iterator? As for both the cases `SeekInfo_BLOCK_UNTIL_READY` and `SeekInfo_FAIL_IF_NOT_READY` it is returning the same response if the block is non-existent ->

```go
return cb.Status_NOT_FOUND, nil
```

How is it making the user wait in case of `SeekInfo_BLOCK_UNTIL_READY` behavior? Is this logic implemented in any other function up in the call tree?

yacovm: > Cool, so should we keep the current changes and add a warning to the code block at line 289 as well? 

no, I think we should just change this:

```
for {
		if seekInfo.Behavior == ab.SeekInfo_FAIL_IF_NOT_READY {
			if number > chain.Reader().Height()-1 {
                               // put here a warning and log number as well.
				return cb.Status_NOT_FOUND, nil
			}
}
```
The reason is that empirically I think that we always use block until ready when we ask for a sequence of blocks and we always use fail if not ready when we ask for a single block

> How is it making the user wait in case of SeekInfo_BLOCK_UNTIL_READY behavior? Is this logic implemented in any other function up in the call tree?

It's making the user wait by having the `Next` method wait:

> block, status = cursor.Next()

Aniket21mathur: I have updated the commit. Please have a look!

yacovm: looks good, what about the test?

Aniket21mathur: I checked and found that we already have a test for this case -> https://github.com/hyperledger/fabric/blob/main/common/deliver/deliver_test.go#L975

I ran the package tests, all passed successfully with the new warning getting logged ->
```bash
•2022-04-20 22:48:54.260 IST 0012 WARN [common.deliver] deliverBlocks -> [channel: chain-id] Block 1000 not found, block number greater than chain length bounds
•2022-04-20 22:48:54.260 IST 0013 WARN [common.deliver] deliverBlocks -> [channel: chain-id] Error reading from channel, cause was: UNKNOWN
•Ran 52 of 52 Specs in 0.262 seconds
SUCCESS! -- 52 Passed | 0 Failed | 0 Pending | 0 Skipped
--- PASS: TestDeliver (0.28s)
PASS
ok  	github.com/hyperledger/fabric/common/deliver	1.528s
```
 

Since we have only added a warning, do we need a new test?

yacovm (approved):

PR #3379, opened 2022-05-04
Properly handle concurrent building of chaincode packages

yacovm:
The chaincode building has a mechanism that detects parallel building of the same chaincode package,
and enforces mutual exclusion to prevent that from happening.

Unfortunately, the mechanism had a bug because it relies on locks but returned a copy of the lock and not
the real lock, thus each goroutine was locking its own lock and not a shared lock.

Another issue was that the mapping that held the locks per package ID was growing indefinitely and never shrinking.

Change-Id: I3d33c369487e0c81d8c64ce6f0bc833b3f2a01bc


C0rWin (approved):

denyeart:

denyeart:

PR #2495, opened 2021-03-19
Update "master" branch references to "main".

denyeart:
Update "master" branch references to "main".

Signed-off-by: David Enyeart <enyeart@us.ibm.com>

PR #3394, opened 2022-05-09
Update "master" branch references to "main". (backport #2495)

WaleedMortaja:
A backport of pull request #2495 ~with some extra edits in a separate commit.~


denyeart:
@WaleedMortaja Thanks for backporting to release-2.2! It looks good to me... are you still working on additional edits?

WaleedMortaja:
> @WaleedMortaja Thanks for backporting to release-2.2! It looks good to me... are you still working on additional edits?

Yes. I would like to customize the backport for branch `release-2.2` instead of `main`.
Since you are here, would suggest the additional edits be on a separate PR?

denyeart:
I'm actually not seeing a need to reference release-2.2 everywhere. Shifting from master to main was the main objective and it looks like that is done here. If you think a subset of the references should be to release-2.2, then yes open a separate PR for that and we can discuss there.

WaleedMortaja:
Thank you so much.
Then I think this PR is ready to merge.

denyeart (approved):

PR #3397, opened 2022-05-09
Update "main" branch references to "release-2.2"

WaleedMortaja:
Signed-off-by: Waleed Mortaja <waleedmortaja@protonmail.com>

### DIFF
--- a/docs/source/github/github.rst
+++ b/docs/source/github/github.rst
@@ -267,7 +267,7 @@ You can use the GitHub cherry-pick command, or an easier option is to paste the 
 
 .. code::
 
-   @Mergifyio backport release-2.0
+   @Mergifyio backport release-2.2
 
 Replace ``2.0`` with the branch that you want to backport to. If there are no merge conflicts,
 a new PR is automatically generated in that branch that still requires the normal approval process to be merged.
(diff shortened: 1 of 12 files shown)


WaleedMortaja:
What about [line 86 and 88 in `docs/source/conf.py` ](https://github.com/hyperledger/fabric/blob/2f4eb7f293938c8104fcb99c0823032acd92b1c8/docs/source/conf.py#L86)? Should they be changed to `release-2.2`?

WaleedMortaja:
[This line](https://github.com/hyperledger/fabric/blob/ed67dbe8c52de97cd03ca8a31f50a4d56563d558/docs/source/policies/policies.md?plain=1#L252) has a link with value `${BRANCH}` in banch `release-2.2`, while it has a value of `main` in [`main` branch](https://github.com/hyperledger/fabric/blob/f3a0d07459ea4542c98e7b4e4ae3347de4900906/docs/source/policies/policies.md?plain=1#L193). Is this acceptable for both branches?

denyeart:
I appreciate the help but we have intentionally been keeping the links as `main` in order to reduce the need to provide maintenance and updates in every single back release. Typically linking to main will provide the latest and greatest information and still be compatible. If there are a couple links that you don't think make sense going to main, you could provide a PR to update them on a case by case basis, but I don't think we want to do blanket updates on back releases.

WaleedMortaja:
Got it!
I still think the README build badge should be changed.

docs/source/github/github.rst
```diff
@@ -267,26 +267,26 @@ You can use the GitHub cherry-pick command, or an easier option is to paste the
...
+   @Mergifyio backport release-2.2
 
 Replace ``2.0`` with the branch that you want to backport to. If there are no merge conflicts,
 a new PR is automatically generated in that branch that still requires the normal approval process to be merged.
 Remember to add a comment to the original PR for each branch that you want to backport to.
 
 If there are merge conflicts, use the GitHub ``cherry-pick`` command instead, by providing the ``SHA`` from the commit in the main branch.
 
-- The following example shows how to cherry-pick a commit from the main branch into the release-2.0 branch:
+- The following example shows how to cherry-pick a commit from the main branch into the release-2.2 branch:
```
WaleedMortaja, line 278: @denyeart What do you think about updating these instructions to `2.2`?

WaleedMortaja:
> Got it! I still think the README build badge should be changed.

Made a new PR #3406

docs/source/github/github.rst
denyeart: It is just an example, it is not something we want to maintain every release.

denyeart:
Ok, #3406 merged, I'll close this one.

docs/source/github/github.rst
WaleedMortaja: Very well

PR #3404, opened 2022-05-10
Bump Go to 1.18.2

denyeart:
Bump Go to 1.18.2

Update staticcheck to a version that works with Go 1.18.
Fix code issues with the new staticcheck.
Fix unit test for new TLS messages and assertion changes in Go 1.18.


internal/pkg/comm/creds_test.go
```diff
@@ -94,7 +94,7 @@ func TestCreds(t *testing.T) {
 		MaxVersion: tls.VersionTLS10,
 	})
 	wg.Wait()
-	require.Contains(t, err.Error(), "protocol version not supported")
+	require.Contains(t, err.Error(), "tls: no supported versions satisfy MinVersion and MaxVersion")
```
C0rWin, line 97: you have same issue at `internal/pkg/comm/creds_test.go:97`

C0rWin:
Was trying to run tests locally there is a failure, not sure it's related to the change, but worth checking

```
• Failure [0.002 seconds]                                                                                                                                     Fields
/Users/c0rwin/golang/src/github.com/hyperledger/fabric/common/grpclogging/fields_test.go:22
  Error
  /Users/c0rwin/golang/src/github.com/hyperledger/fabric/common/grpclogging/fields_test.go:84
    creates an error field for zap [It]
    /Users/c0rwin/golang/src/github.com/hyperledger/fabric/common/grpclogging/fields_test.go:85

    Expected
        <struct { error }>: {
            error: {
                msg: "error",
                stack: [0x17eaa17, 0x17408f6, 0x173fff8, 0x173f235, 0x1744c9a, 0x17444cc, 0x177755b, 0x1777156, 0x17763b7, 0x1779796, 0x178c153, 0x178bee8, 0x
17eb9c5, 0x11abdf4, 0x10ae041],
            },
        }
    to equal
        <struct { error }>: {
            error: {
                msg: "error",
                stack: [0x17eaa17, 0x17408f6, 0x173fff8, 0x173f235, 0x1744c9a, 0x17444cc, 0x177755b, 0x1777156, 0x17763b7, 0x1779796, 0x178c153, 0x178bee8, 0x
17eb9c5, 0x11abdf4, 0x10ae041],
            },
        }

    /Users/c0rwin/golang/src/github.com/hyperledger/fabric/common/grpclogging/fields_test.go:90
------------------------------
••

Summarizing 1 Failure:

[Fail] Fields Error [It] creates an error field for zap
```

denyeart:
Only one failure remaining... the one that @C0rWin mentioned. I don't yet understand why the error interface compare is not working on Go 1.18.

C0rWin:
> Only one failure remaining... the one that @C0rWin mentioned. I don't yet understand why the error interface compare is not working on Go 1.18.

Dave, it doesn't work on Go 1.17 either, IMO this test was broken many weeks before.

denyeart:
@C0rWin At least in my trials in both CI and local, this test always passes when using Go 1.17 and always fails when using Go 1.18.

C0rWin:
> @C0rWin At least in my trials in both CI and local, this test always passes when using Go 1.17 and always fails when using Go 1.18.

Yes, my fault, reran it with 1.17, and it works, but I also discovered that now since 1.18 the following struct

```
package p1

type Test struct {
       A int
}
```

and 

```
package p2

type Test struct {
       A int
}
```

despite structural equivalence with

```
a := 42
A := TTT{A: a}
B := pfi.TTT{A: a}

assert.True(t, reflect.DeepEqual(A, B))
```

fails, i.e. no longer equivalent seems due to being placed in different packages.

C0rWin:
@denyeart see https://github.com/golang/go/issues/52856. We need to update the test because, with 1.18, the structural embedding works differently, and you no longer can do the assertion. I guess the intent of the test is not to ensure that error strips stack trace info after type coercion, probably need to find a different way of asserting this.

denyeart:
@C0rWin Ok, got it this time. Thanks for your help!

C0rWin (approved):

PR #3420, opened 2022-05-16
Log proposal response differences

andrew-coleman:
In the gateway, if proposal responses from different endorsing peers don’t match, then it’s very difficult for the end user to diagnose why.  This commit adds extra logging to help identify the cause of the mismatch.

If the log level is set to ‘debug’ for the ‘gateway’, then the proposal response payloads are unmarshalled and analysed to log specific differences between individual key/values in the read/write sets, SBE policy write differences, event differences and chaincode response differences.  For private collections, only the hashes are logged if they are different.

### DIFF
--- a/internal/pkg/gateway/api_test.go
+++ b/internal/pkg/gateway/api_test.go
@@ -10,6 +10,7 @@ import (
 	"context"
 	"fmt"
 	"io"
+	"strings"
 	"testing"
 	"time"
 
@@ -23,6 +24,8 @@ import (
 	ab "github.com/hyperledger/fabric-protos-go/orderer"
 	"github.com/hyperledger/fabric-protos-go/peer"
 	"github.com/hyperledger/fabric/common/crypto/tlsgen"
+	"github.com/hyperledger/fabric/common/flogging"
+	"github.com/hyperledger/fabric/common/flogging/mock"
 	"github.com/hyperledger/fabric/common/ledger"
 	"github.com/hyperledger/fabric/gossip/api"
 	"github.com/hyperledger/fabric/gossip/common"
@@ -37,6 +40,7 @@ import (
 	"github.com/pkg/errors"
 	"github.com/spf13/viper"
 	"github.com/stretchr/testify/require"
+	"go.uber.org/zap/zapcore"
 	"google.golang.org/grpc"
 	"google.golang.org/grpc/codes"
 	"google.golang.org/grpc/status"
@@ -159,6 +163,8 @@ type preparedTest struct {
 	ledgerProvider *ledgermocks.Provider
 	ledger         *ledgermocks.Ledger
 	blockIterator  *mocks.ResultsIterator
+	logLevel       string
+	logFields      []string
 }
 
 type contextKey string
@@ -762,6 +768,26 @@ func TestEndorse(t *testing.T) {
 					Message: "ProposalResponsePayloads do not match",
 				},
 			},
+			postSetup: func(t *testing.T, def *preparedTest) {
+				def.logLevel = flogging.LoggerLevel("gateway")
+				flogging.ActivateSpec("debug")
+				logObserver := &mock.Observer{}
+				logObserver.WriteEntryStub = func(entry zapcore.Entry, fields []zapcore.Field) {
+					if strings.HasPrefix(entry.Message, "Proposal response mismatch") {
+						for _, field := range fields {
+							def.logFields = append(def.logFields, field.String)
+						}
+					}
+				}
+				flogging.SetObserver(logObserver)
+			},
+			postTest: func(t *testing.T, def *preparedTest) {
+				require.Equal(t, "chaincode response mismatch", def.logFields[0])
+				require.Equal(t, "status: 200, message: , payload: different_response", def.logFields[1])
+				require.Equal(t, "status: 200, message: , payload: mock_response", def.logFields[2])
+				flogging.ActivateSpec(def.logLevel)
+				flogging.SetObserver(nil)
+			},
 		},
 		{
 			name: "discovery fails",
@@ -955,6 +981,9 @@ func TestEndorse(t *testing.T) {
 
 			if checkError(t, &tt, err) {
 				require.Nil(t, response, "response on error")
+				if tt.postTest != nil {
+					tt.postTest(t, test)
+				}
 				return
 			}
 
@@ -971,6 +1000,10 @@ func TestEndorse(t *testing.T) {
 
 			// check the correct endorsers (mocks) were called with the right parameters
 			checkEndorsers(t, tt.expectedEndorsers, test)
+
+			if tt.postTest != nil {
+				tt.postTest(t, test)
+			}
 		})
 	}
 }
--- a/internal/pkg/gateway/apiutils.go
+++ b/internal/pkg/gateway/apiutils.go
@@ -7,6 +7,8 @@ SPDX-License-Identifier: Apache-2.0
 package gateway
 
 import (
+	"bytes"
+	"encoding/hex"
 	"fmt"
 
 	"github.com/golang/protobuf/proto"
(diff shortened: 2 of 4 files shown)


internal/pkg/gateway/api_test.go
```diff
@@ -762,6 +767,25 @@ func TestEndorse(t *testing.T) {
...
+						}
+					}
+				}
+				flogging.SetObserver(logObserver)
+			},
+			postTest: func(t *testing.T, def *preparedTest) {
+				require.Equal(t, "chaincode response mismatch", def.logFields[0])
+				require.Equal(t, "status: 200, message: , payload: different_response", def.logFields[1])
+				require.Equal(t, "status: 200, message: , payload: mock_response", def.logFields[2])
+				flogging.ActivateSpec("info")
```
yacovm: nit, but - shouldn't you sample what was the spec before and save it rather then defaulting to what you think the spec was before?

yacovm (approved):

internal/pkg/gateway/endorsement.go
```diff
@@ -114,6 +115,17 @@ func (p *plan) processEndorsement(endorser *endorser, response *peer.ProposalRes
 	} else {
 		if !bytes.Equal(p.responsePayload, response.GetPayload()) {
 			logger.Warnw("ProposalResponsePayloads do not match (base64)", "payload1", b64.StdEncoding.EncodeToString(p.responsePayload), "payload2", b64.StdEncoding.EncodeToString(response.GetPayload()))
+			if logger.IsEnabledFor(zapcore.DebugLevel) {
+				diff, err := payloadDifference(p.responsePayload, response.GetPayload())
+				if err != nil {
+					logger.Debugf("Failed to analyse response mismatch: %s", err)
+				} else {
+					logger.Debugw("Compared to the initial endorser's response, the following debug log entries detail the differences in this endorser's response.", "address", endorser.address, "mspid", endorser.mspid)
```
denyeart, line 123: It's good that you report endorser address and mspid that doesn't match.
Would it be difficult to report the initial endorser's address and/or mspid? It will often be the current peer but not always, right?

internal/pkg/gateway/apiutils.go
```diff
@@ -125,3 +127,440 @@ func prepareTransaction(header *common.Header, payload *peer.ChaincodeProposalPa
...
+		description = "extraneous write"
+	} else if wd.actual == nil {
+		description = "missing write"
+	}
+	return []interface{}{
+		"type", description,
+		"namespace", wd.namespace,
+		"key", wd.key,
+		"expected value", string(wd.expected),
+		"actual value", string(wd.actual),
```
denyeart, line 475: The terms "expected" and "actual" in the log message are misleading. All we know is that two endorsers got different values. We don't know which one is 'right' or 'expected'.

internal/pkg/gateway/endorsement.go
```diff
@@ -114,6 +115,17 @@ func (p *plan) processEndorsement(endorser *endorser, response *peer.ProposalRes
 	} else {
 		if !bytes.Equal(p.responsePayload, response.GetPayload()) {
 			logger.Warnw("ProposalResponsePayloads do not match (base64)", "payload1", b64.StdEncoding.EncodeToString(p.responsePayload), "payload2", b64.StdEncoding.EncodeToString(response.GetPayload()))
+			if logger.IsEnabledFor(zapcore.DebugLevel) {
```
denyeart, line 118: In the Warnw message above, users will receive the base64 difference but will typically want to know the actual difference. In the Warnw message, could you advise the user to enable "debug" for "gateway" component to get the actual differences?

Actually, I still think we should report the actual difference by default, but have an option for the user to suppress the difference checking if it is too much for them. Having a different logger such as "gateway.resultschecker" would be an option. We could then log the differences at Warn level and user would have the option to switch "gateway.resultschecker" to Error level.

denyeart (commented):
The PR looks good overall and will be a great help for users and SREs. I tried it out and here is how the log statements look:

```
2022-05-19 11:05:16.533 UTC 07e5 WARN [gateway] processEndorsement -> ProposalResponsePayloads do not match (base64) payload1="CiBptArFMIP4q3VTbDedbKaC/UgRP21Y66nYAP70fdS64xLxAQrWARI2CgpfbGlmZWN5Y2xlEigKJgogbmFtZXNwYWNlcy9maWVsZHMvYmFzaWMvU2VxdWVuY2USAggFEpsBCgViYXNpYxKRAQoYChJhc3NldDE2NTI5NTgzMTIxNTcSAggHGnUKEmFzc2V0MTY1Mjk1ODMxMjE1NxpfeyJBcHByYWlzZWRWYWx1ZSI6MTMwMCwiQ29sb3IiOiJ5ZWxsb3ciLCJJRCI6ImFzc2V0MTY1Mjk1ODMxMjE1NyIsIk93bmVyIjoiU2FwdGhhNTkxIiwiU2l6ZSI6NX0aCAjIARoDVG9tIgwSBWJhc2ljGgMxLjA=" payload2="CiBptArFMIP4q3VTbDedbKaC/UgRP21Y66nYAP70fdS64xLxAQrWARI2CgpfbGlmZWN5Y2xlEigKJgogbmFtZXNwYWNlcy9maWVsZHMvYmFzaWMvU2VxdWVuY2USAggFEpsBCgViYXNpYxKRAQoYChJhc3NldDE2NTI5NTgzMTIxNTcSAggHGnUKEmFzc2V0MTY1Mjk1ODMxMjE1NxpfeyJBcHByYWlzZWRWYWx1ZSI6MTMwMCwiQ29sb3IiOiJ5ZWxsb3ciLCJJRCI6ImFzc2V0MTY1Mjk1ODMxMjE1NyIsIk93bmVyIjoiU2FwdGhhMTgwIiwiU2l6ZSI6NX0aCAjIARoDVG9tIgwSBWJhc2ljGgMxLjA="
2022-05-19 11:05:16.534 UTC 07e6 DEBU [gateway] processEndorsement -> Compared to the initial endorser's response, the following debug log entries detail the differences in this endorser's response. address=peer0.org2.example.com:9051 mspid=Org2MSP
2022-05-19 11:05:16.535 UTC 07e7 DEBU [gateway] processEndorsement -> Proposal response mismatch: type="write value mismatch" namespace=basic key=asset1652958312157 expectedvalue="{\"AppraisedValue\":1300,\"Color\":\"yellow\",\"ID\":\"asset1652958312157\",\"Owner\":\"Saptha591\",\"Size\":5}" actualvalue="{\"AppraisedValue\":1300,\"Color\":\"yellow\",\"ID\":\"asset1652958312157\",\"Owner\":\"Saptha180\",\"Size\":5}"
2022-05-19 11:05:16.535 UTC 07e8 INFO [comm.grpc.server] 1 -> unary call completed grpc.service=gateway.Gateway grpc.method=Endorse grpc.request_deadline=2022-05-19T11:05:31.512Z grpc.peer_address=172.21.0.1:62532 error="rpc error: code = Aborted desc = failed to collect enough transaction endorsements, see attached details for more info" grpc.code=Aborted grpc.call_duration=22.5227ms
```

In the review comments I've added a few suggestions to improve the log statements.

internal/pkg/gateway/endorsement.go
andrew-coleman: Yes, that would be useful - I'll do this in a followup PR

internal/pkg/gateway/apiutils.go
andrew-coleman: True, in fact they could both be wrong.  I'll change the words to something like 'initial value' and 'this value' while at the same time logging the address/mspid of 'initial endorser' and 'this endorser'.  Hopefully that will be clearer.

internal/pkg/gateway/endorsement.go
andrew-coleman: Ok, I'll do that.  Since the details will be logged by default, I'll remove the base64 encoded responses from the log.  These are pretty useless from a usability standpoint anyway.

denyeart:

Issue #3413, opened 2022-05-14
BFT: Implement New Block Attestation Service


Param-S:
For BFT block replication module, new RPC service for block attestation is required. RFC captures more details on this new service: [Block Attestation](https://github.com/hyperledger/fabric-rfcs/blob/main/text/0010-bft-signatures.md#attestation-for-block-existence)

PR #3426, opened 2022-05-17
New Block Attestation Orderer GRPC service

Param-S:
#### Type of change
- New feature

#### Description
New RPC service for block attestation is added. It will be used by BFT block replication

#### Related issues
#3413 


orderer/common/server/main.go
```diff
@@ -278,6 +278,15 @@ func Main() {
 		conf.General.Authentication.NoExpirationChecks,
 	)
 
+	attestationService := NewAttestationService(
```
yacovm: Let's put it optional for now, and disable it by default?

yacovm: Or we can not attach it for now

Param-S: Updated. The service is now not enabled.  

But, still the dead/unused code still be part of the other future releases of 2.x.  can we move this BFT changes into separate branch.

yacovm:
```
The following files contain gofumpt errors
orderer/common/server/attestationserver_test.go
The gofumpt command 'gofumpt -l -s -w' must be run for these files

```

orderer/common/server/main.go
yacovm: if it's not imported from the main, it is not part of the binary. I don't think we need to put it in a separate branch.

yacovm (approved):

PR #3439, opened 2022-05-23
Delete historical versions of private data marked for purging (#3024)

### LINKED ISSUE #3024: Process PurgeMarker in background
Process PurgeMarker in private data store to delete the private data historical versions in the background.

manish-sethi:

#### Type of change
- New feature

#### Description
This PR includes
  - Code for processing the purge marker in the background that deletes
  the historical versions of the private data and corresponding
  hashed-indexed-keys that qualifies for deletion
  - Modifies the transaction number match in the previously committed
   code for filtering the 'to be purged' data so as to have the same
   behavior as for the actually purged data.

#### Related issues
Closes #3024

### DIFF
--- a/core/ledger/pvtdatastorage/kv_encoding.go
+++ b/core/ledger/pvtdatastorage/kv_encoding.go
@@ -302,6 +302,18 @@ func encodePurgeMarkerKey(k *purgeMarkerKey) []byte {
 	return encKey
 }
 
+func rangeScanKeysForPurgeMarkers() ([]byte, []byte) {
+	return purgeMarkerKeyPrefix, []byte{purgeMarkerKeyPrefix[0] + 1}
+}
+
+// driveHashedIndexKeyRangeFromPurgeMarker returns the scan range for hashedIndexKeys for a key specified by the `purgeMarkerKey`.
+// The range covers all the hashedIndexKeys between block 0 and the height specified in the `purgeMarkerVal`
+func driveHashedIndexKeyRangeFromPurgeMarker(purgeMarkerKey, purgeMarkerVal []byte) ([]byte, []byte) {
+	startKey := append(hashedIndexKeyPrefix, purgeMarkerKey[1:]...)
+	endKey := append(startKey, purgeMarkerVal...)
+	return startKey, endKey
+}
+
 func encodePurgeMarkerVal(v *purgeMarkerVal) []byte {
 	return version.NewHeight(v.blkNum, v.txNum).ToBytes()
 }
--- a/core/ledger/pvtdatastorage/store.go
+++ b/core/ledger/pvtdatastorage/store.go
@@ -607,7 +607,7 @@ func (s *Store) removePurgedDataFromCollPvtRWset(k *dataKey, v *rwset.Collection
 			return err
 		}
 
-		if keyHt.Compare(purgeMarkerHt) > 0 {
+		if keyHt.Compare(purgeMarkerHt) >= 0 {
 			filterInKVWrites = append(filterInKVWrites, w)
 			continue
 		}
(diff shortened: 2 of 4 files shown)


core/ledger/pvtdatastorage/store.go
```diff
@@ -1053,3 +1087,83 @@ func (c *collElgProcSync) done() {
...
+
+	pvtWrites map[string]*rwsetutil.CollPvtRwSet
+	batch     *leveldbhelper.UpdateBatch
+
+	currentSize int
+}
+
+func newPurgeUpdatesProcessor(db *leveldbhelper.DBHandle, maxBatchSize int) *purgeUpdatesProcessor {
+	return &purgeUpdatesProcessor{
+		db:        db,
```
denyeart: Shouldn't maxBatchSize get set here?

core/ledger/pvtdatastorage/kv_encoding.go
```diff
@@ -302,6 +302,16 @@ func encodePurgeMarkerKey(k *purgeMarkerKey) []byte {
 	return encKey
 }
 
+func rangeScanKeysForPurgeMarkers() ([]byte, []byte) {
+	return purgeMarkerKeyPrefix, []byte{purgeMarkerKeyPrefix[0] + 1}
+}
+
+func driveHashedIndexKeyRangeFromPurgeMarker(purgeMarkerKey, purgeMarkerVal []byte) ([]byte, []byte) {
```
denyeart, line 311: Add a function comment explaining that "purgeMarkerVal" represents the purged height, and this function will return a startKey and endKey that will serve to process and remove all HashedIndex entries from genesis height (startKey) up until the purged height (endKey).

core/ledger/pvtdatastorage/store.go
```diff
@@ -831,6 +835,36 @@ func (s *Store) purgeExpiredData(minBlkNum, maxBlkNum uint64) error {
 	return nil
 }
 
+func (s *Store) deleteDataMarkedForPurge() error {
+	maxBatchSize := 4 * 1024 * 1024 // 4Mb
+	p := newPurgeUpdatesProcessor(s.db, maxBatchSize)
+	pStart, pEnd := rangeScanKeysForPurgeMarkers()
```
denyeart, line 841: Since PurgeMarkers are never deleted, does this imply that we'll be iterating over all previously purged keys as well as the newly purged keys every time? With the implication that it will take longer and longer as time goes on.

core/ledger/pvtdatastorage/store.go
```diff
@@ -1053,3 +1087,83 @@ func (c *collElgProcSync) done() {
...
+
+func newPurgeUpdatesProcessor(db *leveldbhelper.DBHandle, maxBatchSize int) *purgeUpdatesProcessor {
+	return &purgeUpdatesProcessor{
+		db:        db,
+		pvtWrites: map[string]*rwsetutil.CollPvtRwSet{},
+		batch:     db.NewUpdateBatch(),
+	}
+}
+
+func (p *purgeUpdatesProcessor) process(hashedIndexKey, hashedIndexVal []byte) error {
```
denyeart, line 1118: Add a function comment explaining what the hashedIndexKey (purge height for a key) and hashedIndexVal represent, and what exactly this function will do. That is, it will find the referenced private data store height entry, remove the private key write from that height entry, and then re-save the height entry, and then also delete the HashedIndex entry itself.

denyeart (commented):
Given that nobody in the world will understand all the key structures, I think we need function level comments indicating what each key and value represents, and how they will be used to perform the purging. I've highlighted where and what comments I think are needed for people to understand it.

core/ledger/pvtdatastorage/kv_encoding.go
manish-sethi: Done.

core/ledger/pvtdatastorage/store.go
manish-sethi: There will be at most one purge marker for a particular key. So, if a new purge operation is issued, the existing purge marker is over-written. So, if with time the purging gets issued for new keys, yes the scanning would be longer. But since, the hashed index keys would be purged, the processing would stop at scanning this list.

There are two ways to avoid this scanning as well - 1) Maintain another copy of the purge marker for this purpose and keep deleting from there. Or some complex garbage collection for the purge markers if and when there is no corresponding missing data. As of now, I am not sure if we would want to go to that level of optimization for this, as this is a background operation and performs a range scan. But certainly these options are there if we need them in future.

core/ledger/pvtdatastorage/store.go
manish-sethi: done

core/ledger/pvtdatastorage/store.go
manish-sethi: Good Catch

core/ledger/pvtdatastorage/store.go
denyeart: So if 1 million keys have ever been purged, every time we attempt to purge we will iterate through all million records, and for every one of them query for corresponding HashedIndex keys (one million range queries), even if there is no work to be done?

We should at least open an issue to identity this eventual need for optimization.

denyeart (approved):

Issue #3328, opened 2022-04-13
Unable to invoke transaction that interacts with a PDC using fabric-gateway


davidkel:
I have a simple chaincode setup which writes to a single PDC. The collections config is defined as follows
```json
[
    {
      "name": "CollectionOne",
      "policy": "OR('Org1MSP.member')",
      "requiredPeerCount": 0,
      "maxPeerCount": 1,
      "blockToLive": 0,
      "memberOnlyRead": false
    }
]
```

And the chaincode method is
```javascript
    async createPrivateAsset(ctx, uuid) {
        const collection = "CollectionOne"
        const privateAsset = {};
        const transientData = ctx.stub.getTransient();
        privateAsset.content = transientData.get('content').toString('utf8');
        await ctx.stub.putPrivateData(collection, uuid, Buffer.from(JSON.stringify(privateAsset)));
    }
```

using the 2.2 node sdk I can submit a transaction and it get's committed without issue. However when I perform the same using the fabric-gateway it performs the initial endorse without any problem but then returns the following error message
```
EndorseError: 9 FAILED_PRECONDITION: no combination of peers can be derived which satisfy the endorsement policy: no peer combination can satisfy the endorsement policy
```

turning on gateway debug doesn't provide any further detail in the peer logs
```
2022-04-13 10:07:25.361 UTC 009f DEBU [gateway] func1 -> Found local endorser pkiid=3a3ced39e536baae3c8f68dff67499073d918894a63ee36f8d2f18475976064c
2022-04-13 10:07:25.361 UTC 00a0 DEBU [gateway] planFromFirstEndorser -> Sending to first endorser: MSPID=Org1MSP endpoint=peer0.org1.example.com:7051
2022-04-13 10:07:25.368 UTC 00a1 INFO [endorser] callChaincode -> finished chaincode: fixed-asset duration: 16ms channel=mychannel txID=22c60fa0
[...]

PR #3445, opened 2022-05-24
Improved gateway error for transient data failure

andrew-coleman:
When transient data is passed to the gateway in a transaction submit, the gateway is more cautious about which orgs it will select to endorse.  It alters the ChaincodeInterest to only allow orgs that are in the distribution policy for the PDCs written to by the chaincode.
This commit doesn’t change the behaviour, but gives a more helpful error message if this layer of protection caused the submit to fail due to the lack of endorsing orgs.
This will help to diagnose issues such as #3328 which took a long time to resolve because of the generic error message.


denyeart (approved):

denyeart:

PR #3446, opened 2022-05-24
Use any peer to evaluate system chaincode transactions

### LINKED ISSUE #434: can't access to qscc
I use gateway API v1.0.1 (cluster:Fabric:v2.4.3),  try make an access to qscc GetTransactionByID

         network := gateway.GetNetwork("mychannel")                                                                                                                                                                                                         
         contract := network.GetContract("qscc")                                                                                                                                                                                                            
                                                                                                                                                                                                                                                            
         txId := "1d298fc9e709731439726d76794bb8e409dc13e5daf6f944e547d33da3c11e09"                                                                                                                                                                                                                                                                                                                                              
         body, err := contract.EvaluateTransaction("GetTransactionByID", "mychannel", txId)                                                                                                                                                                       
         if err != nil {                                                                                                                                                                                                                                    
                 panic(err)                                                                                                                                                                                                                                 
         }                                                                                                                                                                                                                                                  
         log.Println(string(body)) 


It give an error: 
rpc error: code = FailedPrecondition desc = no peers available to evaluate chaincode qscc in channel mychannel.

I'm sure the specified peer has joined the channel and is anchor peer，and it can access to mycc‘s function.
And, i make another try in mycc (user chaincode)

                channelID := stub.GetChannelID()
                args := [][]byte{[]byte("GetTransactionByID:"), []byte(channelID), []byte(txId)}                                                                                      
                rsp := stub.InvokeChaincode("qscc", args, channelID)                     
                ...                                                                                                
             
It give an error: 
Rejecting invoke of QSCC from another chaincode because of potential for deadlocks, original invocation for 'mycc'

What‘s the right way to access qscc‘s GetTransactionByID by code？

bestbeforetoday:
System chaincodes are not included in the installed chaincodes returned by service discovery. The Gateway service was relying on discovery results to find peers on which to evaluate transactions and so failed to evaluate system chaincode transaction functions. Now the Gateway service uses any network peer to evaluate transactions invoked on known built-in system chaincodes.

Also added a system test to confirm that the Gateway service can endorse system chaincode transaction functions.

Closes hyperledger/fabric-gateway#434

### DIFF
--- a/internal/pkg/gateway/registry.go
+++ b/internal/pkg/gateway/registry.go
@@ -20,9 +20,11 @@ import (
 	"github.com/hyperledger/fabric-protos-go/peer"
 	"github.com/hyperledger/fabric/common/channelconfig"
 	"github.com/hyperledger/fabric/common/flogging"
+	"github.com/hyperledger/fabric/core/scc"
 	gossipapi "github.com/hyperledger/fabric/gossip/api"
 	gossipcommon "github.com/hyperledger/fabric/gossip/common"
 	gossipdiscovery "github.com/hyperledger/fabric/gossip/discovery"
+	"github.com/pkg/errors"
 )
 
 type Discovery interface {
@@ -42,6 +44,7 @@ type registry struct {
 	channelInitialized map[string]bool
 	configLock         sync.RWMutex
 	channelOrderers    sync.Map // channel (string) -> orderer addresses (endpointConfig)
+	systemChaincodes   scc.BuiltinSCCs
 }
 
 type endorserState struct {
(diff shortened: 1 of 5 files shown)


internal/pkg/gateway/registry.go
```diff
@@ -465,3 +467,17 @@ func (reg *registry) closeStaleOrdererConnections(channel string, channelOrderer
 		}
 	}
 }
+
+func (reg *registry) hasChaincode(member gossipdiscovery.NetworkMember, chaincodeName string) bool {
+	if reg.systemChaincodes.IsSysCC(chaincodeName) {
+		return true
+	}
+
```
yacovm: Probably better to reverse the order. Application chaincode calls are much more frequent than SCC ones

ale-linux: Why do we still have system chaincode?

C0rWin: I do not think we have an alternative for [CSCC](https://github.com/hyperledger/fabric/blob/main/core/scc/cscc/configure.go#L73) to join peer channel, etc...

andrew-coleman (approved):

PR #3448, opened 2022-05-25
Support PurgePrivateData from the Chaincode

mbwhite:
- added handler support for the purgePrivateData
- empty message added to the TxSimulator

DRAFT PR

#### Type of change

- Bug fix
- New feature
- Test update
- Documentation update

### DIFF
--- a/core/chaincode/handler.go
+++ b/core/chaincode/handler.go
@@ -207,6 +207,8 @@ func (h *Handler) handleMessageReadyState(msg *pb.ChaincodeMessage) error {
 		go h.HandleTransaction(msg, h.HandleGetStateMetadata)
 	case pb.ChaincodeMessage_PUT_STATE_METADATA:
 		go h.HandleTransaction(msg, h.HandlePutStateMetadata)
+	case pb.ChaincodeMessage_PURGE_PRIVATE_DATA:
+		go h.HandleTransaction(msg, h.HandlePurgePrivateData)
 	default:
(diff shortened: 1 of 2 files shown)


core/chaincode/handler.go
```diff
@@ -1077,6 +1079,36 @@ func (h *Handler) HandleDelState(msg *pb.ChaincodeMessage, txContext *Transactio
...
+			return nil, errors.New("private data APIs are not allowed in chaincode Init()")
+		}
+		if err := errorIfCreatorHasNoWritePermission(namespaceID, collection, txContext); err != nil {
+			return nil, err
+		}
+		err = nil
+		// TODO  will call code from issue #3205
+		// txContext.TXSimulator.PurgePrivateData(namespaceID, collection, delState.Key)
+	} else {
+		return nil, errors.New("only applicable to for private data")
```
denyeart: In idiomatic Go we try to remove as many if/else branches as possible.
In this case you can first return the error if isCollectionSet() returns false, and THEN process the above logic in the main path.  Said another way, since this function is only intended for private data, better to validate that condition first and then there is no need for the "if isCollection()" statement.
Another reason this will flow better is that the following err check will be immediately after the corresponding logic, rather than having this else statement between the logic and the final error check.

mbwhite: I've re-ordered the functions as suggested....

denyeart (approved):

PR #3468, opened 2022-06-09
Update test_network.md

AliakbarETH:
Signed-off-by: Ali Akbar <48174866+BCdeveloper98@users.noreply.github.com>

#### Type of change

- Bug fix
- New feature
- Test update
- Documentation update

### DIFF
--- a/docs/source/test_network.md
+++ b/docs/source/test_network.md
@@ -25,7 +25,7 @@ current main branch, it is possible that you will encounter errors.
 Before you can run the test network, you need to clone the `fabric-samples`
 repository and download the Fabric images.
 
-**Important:** This tutorial is compatible with the Fabric test network sample v2.2.x. After you have installed the [prerequisites](../getting_started.html),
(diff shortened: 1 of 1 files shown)


docs/source/test_network.md
```diff
@@ -25,7 +25,7 @@ current main branch, it is possible that you will encounter errors.
 Before you can run the test network, you need to clone the `fabric-samples`
 repository and download the Fabric images.
 
-**Important:** This tutorial is compatible with the Fabric test network sample v2.2.x. After you have installed the [prerequisites](../getting_started.html),
+**Important:** This tutorial is compatible with the Fabric test network sample v2.2.x. After you have installed the [prerequisites]((https://hyperledger-fabric.readthedocs.io/en/release-2.2/getting_started.html)),
```
denyeart, line 28: Good catch. But we try to keep the URL and release out of the source docs so that it works across branches. In this case you can simply use `getting_started.html`. See the example above that works for `deployment_guide_overview.html`.

denyeart:
Haven't heard back from PR creator... fixed in #3615 instead. Closing...

Issue #3025, opened 2021-11-08
RWSet preparation during transaction simulation


manish-sethi:
Prepare RWSet to capture the purge operation during transaction simulation.

PR #3477, opened 2022-06-13
RWSet Preparation

jcastrence:
#### Type of change

- New feature

#### Description

- Prepare RWSet to capture the purge operation during transaction simulation

#### Related issues

#3025

### DIFF
--- a/core/ledger/kvledger/txmgmt/rwsetutil/rwset_builder.go
+++ b/core/ledger/kvledger/txmgmt/rwsetutil/rwset_builder.go
@@ -116,6 +116,12 @@ func (b *RWSetBuilder) AddToPvtAndHashedWriteSet(ns string, coll string, key str
 	b.getOrCreateCollHashedRwBuilder(ns, coll).writeMap[key] = kvWriteHash
 }
 
+// AddToHashedWriteSetPurge adds a purge key to the hashed write-set
+func (b *RWSetBuilder) AddToHashedWriteSetPurge(ns string, coll string, key string) {
+	kvWriteHashPurge := newKVWriteHashPurge(key)
+	b.getOrCreateCollHashedRwBuilder(ns, coll).writeMap[key] = kvWriteHashPurge
+}
+
 // AddToHashedMetadataWriteSet adds a metadata to a key in the hashed write-set
 func (b *RWSetBuilder) AddToHashedMetadataWriteSet(ns, coll, key string, metadata map[string][]byte) {
 	// pvt write set just need the key; not the entire metadata. The metadata is stored only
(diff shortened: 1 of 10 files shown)


core/ledger/kvledger/txmgmt/rwsetutil/rwset_builder.go
```diff
@@ -116,6 +116,12 @@ func (b *RWSetBuilder) AddToPvtAndHashedWriteSet(ns string, coll string, key str
 	b.getOrCreateCollHashedRwBuilder(ns, coll).writeMap[key] = kvWriteHash
 }
 
+// AddToHashedWriteSetPurge adds a purge key to the hashed write-set
+func (b *RWSetBuilder) AddToHashedWriteSetPurge(ns string, coll string, key string) {
+	kvWriteHashPurge := newKVWriteHashPurge(key)
+	b.getOrCreateCollHashedRwBuilder(ns, coll).writeMap[key] = kvWriteHashPurge
```
denyeart, line 122: I assume you've already discussed this with Manish, but why does the purge not have call to getOrCreateCollPvtRwBuilder() like a regular delete does?
Just trying to see if we can combine DeletePrivateData() and PurgePrivateData() into a single function that takes a purge flag. It looks like the extra getOrCreateCollPvtRwBuilder() call is the only difference.

core/ledger/kvledger/txmgmt/txmgr/tx_simulator.go
```diff
@@ -118,6 +118,19 @@ func (s *txSimulator) DeletePrivateData(ns, coll, key string) error {
...
+func (s *txSimulator) PurgePrivateData(ns, coll, key string) error {
+	if err := s.queryExecutor.validateCollName(ns, coll); err != nil {
+		return err
+	}
+	if err := s.checkWritePrecondition(key, nil); err != nil {
+		return err
+	}
+	s.writePerformed = true
+	s.rwsetBuilder.AddToHashedWriteSetPurge(ns, coll, key)
+	return nil
```
denyeart, line 131: Don't you need to call checkPrivateStateMetadata(), similar to how DeletePrivateData() / SetPrivateData() does it?

core/ledger/kvledger/txmgmt/rwsetutil/rwset_builder.go
jcastrence: My understanding is that since the purge key is only being written to the hashed read write set that it would not require a call to getOrCreateCollPvtRwBuilder()

core/ledger/kvledger/txmgmt/txmgr/tx_simulator.go
jcastrence: I don't think this call is necessary as the purge key is only written to the hashed read write set

denyeart (approved):
Ok, I'll merge so that we get the baseline function in that others can build on top of. When @manish-sethi returns he can help with any subsequent tweaks.

PR #3489, opened 2022-06-17
test: use `T.TempDir` to create temporary test directory

Juneezee:
#### Type of change

- Test update

#### Description

A testing cleanup. 

This pull request replaces `ioutil.TempDir` with `t.TempDir`. We can use the `T.TempDir` function from the `testing` package to create temporary directory. The directory created by `T.TempDir` is automatically removed when the test and all its subtests complete. 

This saves us at least 2 lines (error check, and cleanup) on every instance, or in some cases adds cleanup that we forgot.

Reference: https://pkg.go.dev/testing#T.TempDir

```go
func TestFoo(t *testing.T) {
	// before
	tmpDir, err := ioutil.TempDir("", "")
	require.NoError(t, err)
	defer os.RemoveAll(tmpDir)

	// now
	tmpDir := t.TempDir()
}
```

### DIFF
--- a/core/committer/txvalidator/v14/validator_test.go
+++ b/core/committer/txvalidator/v14/validator_test.go
@@ -9,7 +9,6 @@ package txvalidator_test
 import (
 	"errors"
 	"fmt"
-	"io/ioutil"
 	"os"
 	"strconv"
 	"testing"
@@ -117,7 +116,7 @@ func setupLedgerAndValidatorWithCapabilities(t *testing.T, c *tmocks.Application
 }
 
 func setupLedgerAndValidatorExplicitWithMSP(t *testing.T, cpb *tmocks.ApplicationCapabilities, plugin validation.Plugin, mspMgr msp.MSPManager) (ledger.PeerLedger, txvalidator.Validator, func()) {
-	ledgerMgr, cleanup := constructLedgerMgrWithTestDefaults(t, "txvalidator")
+	ledgerMgr, cleanup := constructLedgerMgrWithTestDefaults(t)
 	gb, err := ctxt.MakeGenesisBlock("TestLedger")
 	require.NoError(t, err)
 	theLedger, err := ledgerMgr.CreateLedger("TestLedger", gb)
@@ -1631,7 +1630,7 @@ func (exec *mockQueryExecutor) GetPrivateDataMetadata(namespace, collection, key
 }
 
 func createCustomSupportAndLedger(t *testing.T) (*mocktxvalidator.Support, ledger.PeerLedger, func()) {
-	ledgerMgr, cleanup := constructLedgerMgrWithTestDefaults(t, "txvalidator")
+	ledgerMgr, cleanup := constructLedgerMgrWithTestDefaults(t)
 	gb, err := ctxt.MakeGenesisBlock("TestLedger")
 	require.NoError(t, err)
 	l, err := ledgerMgr.CreateLedger("TestLedger", gb)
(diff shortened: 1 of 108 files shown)


yacovm (dismissed):
Thanks for the contribution. 

I suggest we hold off merging this PR in the meantime until we cut Fabric 2.5 branch and transition to work on Fabric 3.0, and then we can merge it there. 

We don't want to create unnecessary merge conflicts, and this PR touches a lot of files.

Juneezee:
Rebased onto `main`.

bestbeforetoday:
I particularly like this change since the ioutil package has been deprecated since Go 1.16. 👍

bestbeforetoday:
@yacov are you thinking this change should only go into **main** and not the **release-2.5** branch? If that is the case, it seems like a wide-ranging change to unit test like this could cause issues, as all changes made for v2.5 will also need to be applied to main.

yacovm:
> @yacov are you thinking this change should only go into **main** and not the **release-2.5** branch? If that is the case, it seems like a wide-ranging change to unit test like this could cause issues, as all changes made for v2.5 will also need to be applied to main.

I guess we could put it in release-2.5 so backporting bug fixes would be easier. I don't see any reason other than that.

Juneezee:
I'm not sure what's causing `TestEtcdraft` to fail. I could not reproduce the error on my local machine.

CI test logs: https://github.com/hyperledger/fabric/actions/runs/3030930972/jobs/4896676060

image

yacovm:
> I'm not sure what's causing `TestEtcdraft` to fail. I could not reproduce the error on my local machine.
> 

Welcome to Fabric :-)

Juneezee:
This is really weird. The same CI passed without errors in my fork https://github.com/Juneezee/fabric/actions/runs/3042353726/jobs/4900486546

orderer/consensus/etcdraft/storage_test.go
```diff
@@ -36,7 +37,8 @@ var (
 func setup(t *testing.T) {
 	logger = flogging.NewFabricLogger(zap.NewExample())
 	ram = raft.NewMemoryStorage()
-	dataDir = t.TempDir()
+	dataDir, err = ioutil.TempDir("", "etcdraft-")
```
yacovm: why the revert?

Juneezee: I couldn't find the reason why this is causing the CI unit test to fail. The CI passes when I revert back to `ioutil.TempDir`. 

Should we just continue with `t.TempDir`? What do you think?

yacovm: I'm fine with how it is right now

Juneezee: It seems like the CI unit test _can_ pass with `t.TempDir`... https://github.com/Juneezee/fabric/actions/runs/3046565807/jobs/4909614997

bestbeforetoday: I suspect that some of those unit tests are just a bit flakey in the build environment, and that your change is not the cause of any failure. @yacovm will know better than me though, and should be able to re-trigger failed builds if they are likely to be caused by intermitted test failures.

yacovm: I don't recall storage test to be flaky. chain test is flaky, but that's about it

yacovm (dismissed):
Looks OK, let's get someone else to look at it, in case I missed something

Juneezee:
Rebased onto `main` and resolved conflicts in `orderer/common/server/main_test.go`.

core/committer/txvalidator/v14/validator_test.go
```diff
@@ -1631,7 +1630,7 @@ func (exec *mockQueryExecutor) GetPrivateDataMetadata(namespace, collection, key
 }
 
 func createCustomSupportAndLedger(t *testing.T) (*mocktxvalidator.Support, ledger.PeerLedger, func()) {
-	ledgerMgr, cleanup := constructLedgerMgrWithTestDefaults(t, "txvalidator")
```
ale-linux, line 1634: sorry, possibly misguided Q: why remove the second argument?

Juneezee: See the change at line 1951 below. `constructLedgerMgrWithTestDefaults` now uses the name from `t.TempDir()` as `testDir`.

```diff
- func constructLedgerMgrWithTestDefaults(t *testing.T, testDir string) (*ledgermgmt.LedgerMgr, func()) {
+ func constructLedgerMgrWithTestDefaults(t *testing.T) (*ledgermgmt.LedgerMgr, func()) {
```

ale-linux (approved):

PR #3506, opened 2022-07-05
fix some typos

cuishuang:


ccaas_builder/cmd/detect/main.go
```diff
@@ -69,7 +69,7 @@ func run() error {
 		return fmt.Errorf("chaincode type not supported: %s", metadata.Type)
 	}
 
-	// returning nil indicates to the peer a successfull detection
+	// returning nil indicates to the peer a successfully detection
```
ale-linux: The correct word here would be successful, right?

ale-linux (changes requested):

ale-linux (approved):

PR #3519, opened 2022-07-09
fix(security): Path Traversal Bug

bhaskarvilles:
Unsanitized input from open tar file flows into os.Open, where it is used as a path. This may result in a Path Traversal vulnerability and allow an attacker to open arbitrary files.

primary changes :
	1.updates to archive.go

- Bug fix

#### Description

Unsanitized input from open tar file flows into os.Open, where it is used as a path. This may result in a Path Traversal vulnerability and allow an attacker to open arbitrary files.

## Data flow:
22 steps in 1 file
vendor/github.com/docker/docker/pkg/archive/diff.go


yacovm:
This is a dependency issue, not a Fabric bug. 

1. Do you know if the latest docker package has been patched?
2. Can you show that Fabric actually depends on this code path that you attempt to fix?

bhaskarvilles:
So, can i close my PR?

yacovm:
I suggest to open an issue and close the PR afterwards

PR #3524, opened 2022-07-14
Fix(security):Cross Site Scripting Attack(XSS)

bhaskarvilles:
Unsanitized input from the request URL flows into fmt.Fprintf, where it is used to render an HTML page returned to the user. This may result in a Reflected Cross-Site Scripting attack (XSS).

#### Type of change

- Bug fix

#### Description

Mitigate Cross Site Scripting Attack


yacovm:
This is an integration test framework. The input is coming from the Fabric system that is running

bhaskarvilles:
@yacovm So this bug didn't affect this program?

yacovm:
No, this is in a test. This is not production code.

PR #3525, opened 2022-07-15
Extra checks for invalid args in gateway api

andrew-coleman:
Additional argument validation and consistently use getters to safely traverse proto structures.

### DIFF
--- a/internal/pkg/gateway/api.go
+++ b/internal/pkg/gateway/api.go
@@ -63,7 +63,7 @@ func (gs *Server) Evaluate(ctx context.Context, request *gp.EvaluateRequest) (*g
 	var response *peer.Response
 	var errDetails []proto.Message
 	for response == nil {
-		gs.logger.Debugw("Sending to peer:", "channel", channel, "chaincode", chaincodeID, "txID", request.TransactionId, "MSPID", endorser.mspid, "endpoint", endorser.address)
+		gs.logger.Debugw("Sending to peer:", "channel", channel, "chaincode", chaincodeID, "txID", request.GetTransactionId(), "MSPID", endorser.mspid, "endpoint", endorser.address)
 
 		done := make(chan error)
 		go func() {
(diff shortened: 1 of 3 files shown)


internal/pkg/gateway/api.go
```diff
@@ -63,7 +63,7 @@ func (gs *Server) Evaluate(ctx context.Context, request *gp.EvaluateRequest) (*g
 	var response *peer.Response
 	var errDetails []proto.Message
 	for response == nil {
-		gs.logger.Debugw("Sending to peer:", "channel", channel, "chaincode", chaincodeID, "txID", request.TransactionId, "MSPID", endorser.mspid, "endpoint", endorser.address)
+		gs.logger.Debugw("Sending to peer:", "channel", channel, "chaincode", chaincodeID, "txID", request.GetTransactionId(), "MSPID", endorser.mspid, "endpoint", endorser.address)
```
yacovm, line 66: while I understand the approach (the methods check that the "self" / "this" object is nil and if so, return the zero value), from a high level perspective I am not sure this is a good approach: 

Consider you have a nested structure:
```
   {
      mandatory_field1: {
            mandatory_field2: {
                   .... 
                   mandatory_fieldN: { 
                         signature: ...
                   }
            }
       }
   }
```
with this approach, an improper empty request will be passed all the way and will then say "invalid signature", while the real problem is that the mandatory fields along the way were not initialized. 

That being said, I suppose we can merge this now, and address this issue later.

denyeart: Many other places in Fabric follow similar proto access patterns as those fixed in this PR. So I guess we need to consider fixing all of them? If that is the case I'd rather do a 'permanent' fix that can set the precedent than an 'interim' fix that we need to re-visit later.

Note that I'm not necessarily against the approach in this PR though, given the rarity that the error is hit. If there is no other way to keep the code fairly compact and lightweight it may be a good enough compromise. I'm just saying it would be better to finalize the approach now rather than create debt for later.

yacovm: @ale-linux what do you think?

bestbeforetoday: Well behaved clients (for example, the Fabric SDKs) never send invalid request messages. These code changes are just to protect the peer against malicious clients. How much does it matter about the accuracy of error messages returned to malicious clients?

denyeart: Agreed, the error message just has to be in the ballpark, not necessarily exact.

ale-linux: What is the gateway threat model? Can we assume clients are always well-behaved?

bestbeforetoday: Of course we can’t assume that malicious clients are well behaved. This PR is plugging some holes that could be exploited by these badly behaved clients to cause failures in the peer.

What I am suggesting is that non-malicious clients already have helpful error messages for cases where they might get things wrong, and there is little value but definitely some overhead in crafting additional error messages to be helpful only to deliberately malicious clients. How does the threat model relate to that?

denyeart:
I'm good with the approach here.

@C0rWin actually opened a similar PR, let's see if he wants to add his additional checks as a second commit on top of Andy's.

C0rWin:
> I'm good with the approach here.
> 
> @C0rWin actually opened a similar PR, let's see if he wants to add his additional checks as a second commit on top of Andy's.

Just rebased and pushed a new one #3572

denyeart:
This commit plus Artem's additional checks have been merged in https://github.com/hyperledger/fabric/pull/3572.

PR #3530, opened 2022-07-17
Insecure TLS Configuration

bhaskarvilles:
Insecure TLS Configuration

#### Type of change

Insecure TLS configuration is found to be in use. MaxVersion is set to a deprecated SSL/TLS version

### DIFF
--- a/common/fabhttp/server_test.go
+++ b/common/fabhttp/server_test.go
@@ -71,7 +71,7 @@ var _ = Describe("Server", func() {
 	When("trying to connect with an old TLS version", func() {
 		BeforeEach(func() {
 			tlsOpts := []func(config *tls.Config){func(config *tls.Config) {
-				config.MaxVersion = tls.VersionTLS11
+				config.MaxVersion = tls.VersionTLS12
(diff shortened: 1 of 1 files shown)


common/fabhttp/server_test.go
```diff
@@ -71,7 +71,7 @@ var _ = Describe("Server", func() {
 	When("trying to connect with an old TLS version", func() {
 		BeforeEach(func() {
 			tlsOpts := []func(config *tls.Config){func(config *tls.Config) {
-				config.MaxVersion = tls.VersionTLS11
+				config.MaxVersion = tls.VersionTLS12
```
yacovm, line 74: it should be min version, not max

yacovm, line 74: this is a test that forces to use TLS 1.1. It is called "trying to connect with an old TLS version" :-)

bhaskarvilles: So it doesn't make sense here ?

denyeart: Doesn't make sense here, the intent of the test is to use the old TLS version. Closing...

PR #3552, opened 2022-07-28
Support bootstrap using non docker container CLI

lehors:
This change allows specifying a different CLI so the following is
possible for instance:

CONTAINER_CLI=podman ./bootsrap.sh -s

Signed-off-by: Arnaud J Le Hors <lehors@us.ibm.com>

#### Type of change

- Bug fix
- New feature
- Test update
- Documentation update


yacovm (commented):
> CONTAINER_CLI=podman ./bootsrap.sh -s

If you need to specify manually the container CLI then you might as well alias docker to podman:

```
alias docker='podman'
```

and then you can run the bootstrap as usual, no?

lehors:
I could but that's not the choice that was made in fabric-samples because sometimes we need to be able to differentiate which cli is being used, such as in:
https://github.com/hyperledger/fabric-samples/blob/main/test-network/network.sh#L333

So it makes sense to be consistent.

yacovm (approved):

PR #3557, opened 2022-07-31
Enable compilation on linux/arm64 platforms

C0rWin:
Enable compilation on linux/arm64 platforms

    While building the peer docker image, the location of the ccaasbuilder
    binaries were hard-coded into the Dockerfile, though the output directory
    is platform-dependent. Therefore this commit suggests two changes:

    1. Prepare binary locations based on the architecture and copy them into
       the temporal location hence it will become possible to generalize
       file location inside Dockerfile

    2. Suggest static binaries compilation to include all dependent
       libraries in compile time, it turns out there are the different sets of
       libraries between alpine:${ALPINE_VER} and
       golang:${GO_VER}-alpine${ALPINE_VER} docker images which cause peer
       to segfault on ARM64-based platforms.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -357,8 +357,8 @@ ccaasbuilder/%: ccaasbuilder-clean
 	$(eval GOOS = $(word 1,$(subst -, ,$(platform))))
 	$(eval GOARCH = $(word 2,$(subst -, ,$(platform))))
 	@mkdir -p ../release/$(strip $(platform))/builders/ccaas/bin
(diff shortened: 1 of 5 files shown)


mbwhite (approved):
LGTM thanks @C0rWin

yacovm (approved):

bestbeforetoday:
Note this change seems to have caused some issues if trying to build docker images with PKCS11 support: #3645

PR #3565, opened 2022-08-04
fix timeout ginkgo

pfi79:
In the old ginkgo, the timeout by default is 24h

In the new ginkgo (v2) the timeout by default is 1h

1 hour is not enough to pass the integration tests

I suggest that you explicitly pass the timeout parameter equal to 24h, so that there is enough time to pass the integration tests


denyeart (approved):

denyeart:

PR #3569, opened 2022-08-05
Github Actions Verify Build

jcastrence:
#### Type of change

- New feature

#### Description

- Set of workflows in preparation to migrate Fabric CI/CD from Azure Pipelines to Github Actions
- Updates integration tests script to allow for test name arguments


mbwhite (dismissed):

.github/workflows/verify-build.yml
```diff
@@ -0,0 +1,32 @@
...
+# SPDX-License-Identifier: Apache-2.0
+
+name: Verify Build
+on:
+  push:
+    branches: ["main"]
+  pull_request:
+    branches: ["main"]
+  schedule:
+    - cron: "0 0 * * *"
```
denyeart: Remove the schedule.

.github/workflows/verify-build.yml
```diff
@@ -0,0 +1,32 @@
+# Copyright the Hyperledger Fabric contributors. All rights reserved.
+#
+# SPDX-License-Identifier: Apache-2.0
+
+name: Verify Build
+on:
+  push:
+    branches: ["main"]
```
denyeart: We'd want all branches right?

.github/workflows/verify-build.yml
```diff
@@ -0,0 +1,32 @@
...
+  pull_request:
+    branches: ["main"]
```
denyeart: Same.

.github/workflows/integration-tests-1.yml
```diff
@@ -0,0 +1,34 @@
+# Copyright the Hyperledger Fabric contributors. All rights reserved.
+#
+# SPDX-License-Identifier: Apache-2.0
+
+name: Integration Tests 1
+on:
+  workflow_run:
+    workflows: ["Verify Build"]
+    branches: ["main"]
```
denyeart: We'd want all branches right?

.github/workflows/integration-tests-1.yml
```diff
@@ -0,0 +1,34 @@
...
+      - completed
+  workflow_dispatch:
+
+env:
+  GOPATH: /opt/go
+  PATH: /opt/go/bin:/bin:/usr/bin:/sbin:/usr/sbin:/usr/local/bin:/usr/local/sbin
+  GO_VER: 1.18.2
+
+jobs:
+  unit-tests:
```
denyeart: integration-tests

Makefile
```diff
@@ -147,6 +147,26 @@ check-go-version:
 integration-test: integration-test-prereqs
 	./scripts/run-integration-tests.sh
 
+.PHONY: integration-test-suite-1
+integration-test-suite-1: integration-test-prereqs
+	./scripts/run-integration-tests.sh configtx gossip raft
```
denyeart: Instead of hardcoding five suites in the Makefile, how about passing the integration test directories to a single Make target?
e.g. https://chriswiegman.com/2021/08/using-parameters-in-a-make-target/

.github/workflows/integration-tests-1.yml
```diff
@@ -0,0 +1,34 @@
...
+    types:
+      - completed
+  workflow_dispatch:
+
+env:
+  GOPATH: /opt/go
+  PATH: /opt/go/bin:/bin:/usr/bin:/sbin:/usr/sbin:/usr/local/bin:/usr/local/sbin
+  GO_VER: 1.18.2
+
+jobs:
```
denyeart: Is it possible to have a single integration test workflow that runs 5 jobs? It appears that you can now re-trigger individual jobs that have failed in a workflow - https://docs.github.com/en/actions/managing-workflow-runs/re-running-workflows-and-jobs#re-running-a-specific-job-in-a-workflow

.github/workflows/verify-build.yml
denyeart: I think we can simply remove the `branches` filter, and backport to all release branches, so that it runs on all branches.

.github/workflows/integration-tests-1.yml
jcastrence: The one advantage to having each integration test in their own workflow is the ability to manually trigger any one of them at any time through the Actions tab. If we only care about re-triggering workflows when they have failed, then yes we can just have one workflow that runs 5 jobs.

denyeart: My opinion is that there is not much benefit to trigger them individually from the Actions tab. Typically if you want to trigger the integration tests from UI you'd want to run them all.

The split into 5 sets is somewhat arbitrary, so it doesn't feel right to make them 5 first class workflows that need to be maintained individually. I'd prefer one workflow to trigger them all while still benefitting from running them in parallel and ability to re-run any one that failed. So for me, that implies one workflow with 5 parallel jobs.

denyeart (approved):
Looks like a good place to start. I'll merge and we can assess the experience when PRs are opened.
Thanks @jcastrence

denyeart:

denyeart:

denyeart:

denyeart:

PR #3570, opened 2022-08-05
  Prevent peer from failure on malformed proposal

C0rWin:
- Bug fix

While the new peer gateway service tries to extract channel and chaincode information from the signed proposal, it doesn't check the proposal fields for validity. Therefore malformed proposal might end up crashing peer service. This commit adds validation to the parameters extracted from the proposal.


C0rWin:
@denyeart fyi

denyeart:
@C0rWin There is a similar PR already being discussed here:
https://github.com/hyperledger/fabric/pull/3525

It has a few more checks in general, but yours has a few more checks later in apiutils.go. Perhaps we should add those to Andy's PR...

C0rWin:
> @C0rWin There is a similar PR already being discussed here: #3525
> 
> It has a few more checks in general, but yours has a few more checks later in apiutils.go. Perhaps we should add those to Andy's PR...

Yes, since these, I think the object of the discussion. Anyway, I will close that one.

C0rWin:
I have noted a few checks that are still missing, based on @denyeart suggestion, rebased.

PR #3587, opened 2022-08-11
unittest update, fix unittest failed case for TestEcdsaPublicKey

edwardramsey:
#### Type of change
- Test update

#### Description
modify file: bccsp/sw/ecdsa_test.go
reason: with go version 1.19, `x509.MarshalPKIXPublicKey` -> `marshalPublicKey` -> `elliptic.Marshal` will panic if point not on curve because of function new function `panicIfNotOnCurve(curve, x, y)`

```go
func Marshal(curve Curve, x, y *big.Int) []byte {
	panicIfNotOnCurve(curve, x, y)
        // ...
}
```

### DIFF
--- a/bccsp/sw/ecdsa_test.go
+++ b/bccsp/sw/ecdsa_test.go
@@ -22,8 +22,8 @@ import (
 	"crypto/rand"
 	"crypto/sha256"
 	"crypto/x509"
-	"fmt"
 	"math/big"
+	"runtime"
 	"testing"
 
 	"github.com/hyperledger/fabric/bccsp/utils"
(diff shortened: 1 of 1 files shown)


bccsp/sw/ecdsa_test.go
```diff
@@ -176,6 +177,13 @@ func TestEcdsaPublicKey(t *testing.T) {
 	invalidCurve := &elliptic.CurveParams{Name: "P-Invalid"}
 	invalidCurve.BitSize = 1024
 	k.pubKey = &ecdsa.PublicKey{Curve: invalidCurve, X: big.NewInt(1), Y: big.NewInt(1)}
+	defer func() {
+		if p := recover(); p != nil {
```
yacovm: there is a function [require.Panics](https://pkg.go.dev/github.com/stretchr/testify/require#Panics), let's use it instead.

yacovm (changes requested):

bccsp/sw/ecdsa_test.go
edwardramsey: you means like this ?
```go
require.Panics(t, func() {
    _, err = k.Bytes()
})
```

but it would be failed when it is not version 1.19. If now we need support both 1.17 and 1.19 ?

yacovm: Then let's use `runtime.Version`

yacovm (approved):

PR #3592, opened 2022-08-16
BFT block verifier

andrew-coleman:
Implemented BFT block verifier in orderer and peer gossip


yacovm (approved):
Can we export `blockSignatureVerifier` to proto_utils after this PR? We could then just reference it in the block fetcher.

PR #3606, opened 2022-08-25
Orderer new communication service with auth support

### LINKED ISSUE #3474: BFT: Replace Mutual TLS with TLSBinding Authentication
As it is described in the RFC https://github.com/hyperledger/fabric-rfcs/pull/50/files#diff-8ba0214c161394f52cdac8a5414042fa8e99926620fb37556f8349e73cec1321R262 mutual TLS with pinning will be replaced with the TLSBinding based authentication. 

**Authentication Method:**
The connection between the orderers is secured using the orderer level authentication by making use of the EKM([Export Keying Materials](https://www.rfc-editor.org/rfc/rfc5705.html)) for TLS. 

**gRPC method update**
A new request(Auth Request) will be added to cluster service API `Step RPC`.  It will be called implicitly as part of the Submit or Consensus Request before the actual request sent to the other node

Param-S:
#### Type of change

- New feature

#### Description

Implementation of orderer v3 communication framework based on the RFC [Orderer v3](https://github.com/hyperledger/fabric-rfcs/blob/main/text/orderer-v3.md#orderer-to-orderer-authentication-and-communication)

#### Related issues
Closes #3474 


orderer/common/cluster/comm.go
```diff
@@ -55,6 +56,8 @@ type Comm struct {
 	Chan2Members                     MembersByChannel
 	Metrics                          *Metrics
 	CompareCertificate               CertificateComparator
+	NodeID                           uint64
+	Signer                           identity.Signer
```
yacovm, line 60: can we simply define the interface in `interfaces.go` ?

yacovm: can we simply define this interface `identity.Signer` in `interfaces.go` ?

orderer/common/cluster/stream.go
```diff
@@ -25,19 +25,19 @@ type Stream struct {
...
+	commShutdown       chan struct{}
+	abortReason        *atomic.Value
+	metrics            *Metrics
+	ID                 uint64
+	Channel            string
+	NodeName           string
+	Endpoint           string
+	Logger             *flogging.FabricLogger
+	Timeout            time.Duration
+	Cluster_StepClient StepClientStream
```
yacovm: why the renaming?

orderer/common/cluster/clusterservice.go
```diff
@@ -0,0 +1,277 @@
...
+	authReq, err := s.VerifyAuthRequest(stream, request)
+	if err != nil {
+		s.Lock.RUnlock()
+		s.Logger.Warnf("service authentication of %s failed with error: %v", addr, err)
+		return status.Errorf(codes.Unauthenticated, "access denied")
+	}
+
+	defer s.Logger.Debugf("Closing connection from %s(%s)", commonName, addr)
+
+	streamID := atomic.AddUint64(&s.MembershipByChannel[authReq.Channel].nextStreamID, 1)
```
yacovm, line 97: after obtaining the stream identifier, you need to ensure the stream id is garbage collected when the client disconnects:

```
	defer func() {
		s.MembershipByChannel[authReq.Channel].AuthorizedStreams.Delete(streamID)
	}()
```

orderer/common/cluster/clusterservice.go
```diff
@@ -0,0 +1,277 @@
...
+
+	if !bytes.Equal(tlsBinding, authReq.SessionBinding) {
+		return nil, errors.New("session binding mismatch")
+	}
+
+	msg, err := asn1.Marshal(AuthRequestSignature{
+		Version:        int64(authReq.Version),
+		Timestamp:      authReq.Timestamp.String(),
+		FromId:         strconv.FormatUint(authReq.FromId, 10),
+		ToId:           strconv.FormatUint(authReq.ToId, 10),
```
yacovm, line 142: Let's add a check that `ToId` is us. To do that, we can add an identity field and then lookup our own identity within the channel mapping and locate our identifier.

orderer/common/cluster/clusterservice.go
```diff
@@ -0,0 +1,277 @@
...
+	if membership == nil {
+		return nil, errors.Errorf("channel %s not found in config", authReq.Channel)
+	}
+
+	identity := membership.MemberMapping[authReq.FromId]
+	if identity == nil {
+		return nil, errors.Errorf("node %d is not member of channel %s", authReq.FromId, authReq.Channel)
+	}
+
+	err = VerifySignature(identity, msg, authReq.Signature)
```
yacovm: `msg` is the ASN1 encoding of `AuthRequestSignature` but `VerifySignature` expects a hash.

orderer/common/cluster/connectionsmgr.go
```diff
@@ -0,0 +1,84 @@
...
+	defer c.lock.Unlock()
+
+	conn, alreadyConnected := c.Connections.Lookup(endpoint)
+	if alreadyConnected {
+		return conn, nil
+	}
+
+	c.dialer.SecOpts.ServerRootCAs = [][]byte{serverRootCACert}
+
+	conn, err := c.dialer.Dial(endpoint)
```
yacovm: It's better not to `Dial()` under a lock.

orderer/common/cluster/connectionsmgr.go
```diff
@@ -0,0 +1,84 @@
...
+	}
+	c.lock.Lock()
+	defer c.lock.Unlock()
+
+	conn, alreadyConnected := c.Connections.Lookup(endpoint)
+	if alreadyConnected {
+		return conn, nil
+	}
+
+	c.dialer.SecOpts.ServerRootCAs = [][]byte{serverRootCACert}
```
yacovm: I am not sure this is good practice, to change the configuration of a shared field that is accessed via different goroutines. 

Can't we clone it first and then override the server Root CA certs? 
Other goroutines may use this reference later without a lock, we should not assume anything about how the field `ServerRootCAs` is used later on.

orderer/common/cluster/commauth.go
```diff
@@ -0,0 +1,339 @@
...
+
+	Lock           sync.RWMutex
+	shutdown       bool
+	shutdownSignal chan struct{}
+
+	Chan2Members MembersByChannel
+	Connections  *ConnectionsMgr
+
+	SendBufferSize int
+	NodeID         uint64
```
yacovm: unfortunately, we are not guaranteed we have the same identifier across all channels. we need to replace this with an identity and then do a lookup each time for each channel.

yacovm (changes requested):

orderer/common/cluster/comm.go
Param-S: `identity.Signer` is preexisting interface which is defined here `github.com/hyperledger/fabric/internal/pkg/identity`. Do we want to redefine this interface inside the `cluster` package? I am trying to understand the reason for defining it inside this package again? I see the `internal/pkg` referenced in other `cluster` package types
image

orderer/common/cluster/stream.go
Param-S: It is not renaming, the name is same. But, the type is updated here:
image

yacovm: I mean why is the name of the field `Cluster_StepClient` so odd?

orderer/common/cluster/comm.go
yacovm: oh, interesting. then nevermind.

orderer/common/cluster/stream.go
Param-S: I retained the name as is before. Its good to change it now.  I will change it to `StepClient`

orderer/common/cluster/clusterservice.go
Param-S: Yes, I missed out hashing even while Signing. Now. I will update both the places.

orderer/common/cluster/connectionsmgr.go
Param-S: agree. i will update it

Param-S:
In the latest commit, updates based on review.   One pending changes - Identity based lookup, i will work on that.

orderer/common/cluster/clusterservice.go
```diff
@@ -93,12 +93,15 @@ func (s *ClusterService) Step(stream orderer.ClusterNodeService_StepServer) erro
...
 
-	defer s.Logger.Debugf("Closing connection from %s(%s)", commonName, addr)
-
 	streamID := atomic.AddUint64(&s.MembershipByChannel[authReq.Channel].nextStreamID, 1)
 	s.MembershipByChannel[authReq.Channel].AuthorizedStreams.Store(streamID, authReq.FromId)
 	s.Lock.RUnlock()
 
+	defer s.Logger.Debugf("Closing connection from %s(%s)", commonName, addr)
+	defer func() {
+		s.MembershipByChannel[authReq.Channel].AuthorizedStreams.Delete(streamID)
```
yacovm, line 104: you need to RLock, no?

orderer/common/cluster/clusterservice.go
Param-S: If I understand correctly, we need to add a new field(identity) into Auth message which will be used by remote node to cross verify/confirm whether it is a right target node. 
Is it similar to mutual tls, but at the application layer?

Param-S: Updated in the latest commit(3)

orderer/common/cluster/commauth.go
Param-S: Updated

orderer/common/cluster/clusterservice.go
```diff
@@ -149,12 +154,29 @@ func (s *ClusterService) VerifyAuthRequest(stream orderer.ClusterNodeService_Ste
...
+
+	identityPEMBlock, _ := pem.Decode(s.NodeIdentity)
+	if identityPEMBlock == nil {
+		return nil, errors.Errorf("cert decoding failed")
+	}
+	toIdentityPEMblock, _ := pem.Decode(toIdentity)
+	if toIdentityPEMblock == nil {
+		return nil, errors.Errorf("cert decoding failed")
+	}
+	if err := crypto.CertificatesWithSamePublicKey(identityPEMBlock.Bytes, toIdentityPEMblock.Bytes); err != nil {
```
yacovm: Let's just compare bit-wise between `membership.MemberMapping[authReq.ToId]` and our identity. This PK based comparison is too much.

orderer/common/cluster/commauth.go
```diff
@@ -182,10 +184,20 @@ func (ac *AuthCommMgr) createRemoteContext(stub *Stub, channel string) func() (*
 			if err != nil {
 				return nil, err
 			}
+
+			membersMapping, exists := ac.Chan2Members[channel]
+			if !exists {
+				return nil, errors.Errorf("channel members not initialized")
+			}
+			nodeStub := membersMapping.LookupByIdentity(ac.NodeIdentity)
```
yacovm, line 190: I don't think we should be using PK based comparison. Let's do a bitwise one.

Unlike TLS, for which we made the PK based comparison, here we also have an MSP ID. When you compare only by PK, you don't look at the MSP ID.

yacovm (approved):

PR #3616, opened 2022-08-25
Reflectd Cross site Scripting (XSS)

bhaskarvilles:
Unsanitized input from the request body `r.Body` flows into `fmt.Fprintln`. This may result in a Reflected Cross-Site Scripting attack (XSS).

Signed-off-by: Bhaskar Ram <bhaskarvilles@duck.com>


denyeart:
Unsanitized input does not get returned. If you think that is the case, please include a unit test that proves it.

Also, please don't keep issues open that have failed to build. Closing...

PR #3631, opened 2022-09-07
Update prereqs.rst

AliakbarETH:
I have proposed some explination of the commands.

We are typing in the commands but we don't know the pros and cons.

If we write pros and cons with important, rather all commands it will be quite helpful as the support of HLF is quite less and I have seen people quitting it because of the same very reason.

Signed-off-by: Ali Akbar <48174866+AliakbarETH@users.noreply.github.com>

#### Type of change

- Bug fix
- New feature
- Test update
- Documentation update


denyeart:
Thank you for the input. Pull Requests are for actual changes to the repository. It appears rather than making changes, you are discussing what should be done. In this case an Issue would be more appropriate. I'll close this PR but please do open an Issue so that the proper resolution can be discussed.

Issue #3514, opened 2022-07-07
Orderer v3: remove solo support


tock-ibm:
In v3 we will remove support for the solo consensus protocol.

- [x] Disable solo as a consensus protocol
- [x] Change integration tests that use solo to single node raft
- [x] Change integration tests that do consensus type migration to check limited functionality (maintenance mode)
- [x] Remove integration tests that use solo but do not add value
- [x] Remove solo from integration tests infrastructure
- [x] orderer server main unit test uses a solo-based process 
- [x] Remove solo code from orderer / consensus
- [x] Clean up the code for kafka to raft migration, but keep some features of it that may be useful for migration from protocol to protocol (e.g. raft -> bft)
- [ ] Some unit tests use a solo config tx (genesis usually), change them to use single node raft
- [ ] Remove solo from local config objects and methods
- [ ] Remove solo from (shared) channel config objects and methods
- [ ] Remove solo kafka from configtxgen
- [ ] #4496
- [ ] Remove/deprecate solo from documentation

In Epic #3511

SamYuan1990:
will we remove solo support in fabric v3 ? hence for local development env, will use raft with a single node?

rasc0:
Documentation needs to be updated to reflect deprecation of solo consensus protocol. See [this page](https://hyperledger-fabric.readthedocs.io/en/release-2.5/peer-chaincode-devmode.html) for an example of a tutorial page still referencing solo orderer

Issue #3511, opened 2022-07-07
Orderer v3: remove deprecated components


tock-ibm:
The first part of the "Orderer V3 RFC" calls for the removal of deprecated components:
- Kafka
- Solo
- System channel support

See RFC: 
- https://hyperledger.github.io/fabric-rfcs/text/orderer-v3.html
- https://github.com/hyperledger/fabric-rfcs/pull/50

In addition, since block gossip is deprecated since v2.5, change the defaults accordingly, e.g. see #3961 , making it disabled by default, even if respective parameters are missing.

PR #3637, opened 2022-09-12
Ordrer v3: disable solo

tock-ibm:
Change-Id: I00846aa130d5b9fb486ace5f84898a0b0affddf5

#### Description

- Disable the solo consenter in the orderer main
- fix unit tests that fail
- fix integration tests that fail

#### Related issues

Issue: #3514 
Epic: #3511

### DIFF
--- a/orderer/common/server/main.go
+++ b/orderer/common/server/main.go
@@ -49,7 +49,6 @@ import (
 	"github.com/hyperledger/fabric/orderer/common/onboarding"
 	"github.com/hyperledger/fabric/orderer/consensus"
 	"github.com/hyperledger/fabric/orderer/consensus/etcdraft"
-	"github.com/hyperledger/fabric/orderer/consensus/solo"
 	"github.com/hyperledger/fabric/protoutil"
 	"go.uber.org/zap/zapcore"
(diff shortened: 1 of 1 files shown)


denyeart:
@tock-ibm Is there a need to disable the Solo option? Last time the team discussed this topic the conclusion was that we would keep the Solo option as a simple baseline option that works across all Fabric releases for simple trials, since it hasn't required any support or maintenance in many years. But if there is a valid reason to remove it I'm fine with that...

tock-ibm:
> @tock-ibm Is there a need to disable the Solo option? Last time the team discussed this topic the conclusion was that we would keep the Solo option as a simple baseline option that works across all Fabric releases for simple trials, since it hasn't required any support or maintenance in many years. But if there is a valid reason to remove it I'm fine with that...

The main reason is that solo does not support working with the channel participation API, nor without a system channel. So going forward, when we remove support for a system channel completely, solo will require some treatment. It is certainly doable to provide support for that, but I think it is easier and cleaner to just replace it with single node etcdraft everywhere.

yacovm (approved):

Issue #3629, opened 2022-09-06
Unit test flake in etcdraft


tock-ibm:
This use case in the `etcdraft` suite fails intermittently: 

------------------------------
• [FAILED] [10.153 seconds]
Chain
/home/runner/work/fabric/fabric/orderer/consensus/etcdraft/chain_test.go:101
  3-node Raft cluster
  /home/runner/work/fabric/fabric/orderer/consensus/etcdraft/chain_test.go:1565
    reconfiguring raft cluster
    /home/runner/work/fabric/fabric/orderer/consensus/etcdraft/chain_test.go:1715
      reconfiguration
      /home/runner/work/fabric/fabric/orderer/consensus/etcdraft/chain_test.go:1823
        [It] rotates leader certificate and triggers leadership transfer
        /home/runner/work/fabric/fabric/orderer/consensus/etcdraft/chain_test.go:1864

  Begin Captured GinkgoWriter Output >>
    Send artificial MsgTimeoutNow to elect node 1
    STEP: Submitting first tx to cut the block 09/05/22 14:17:19.904
    STEP: creating new configuration with removed node and new one 09/05/22 14:17:19.921
    STEP: sending config transaction 09/05/22 14:17:19.923
  << End Captured GinkgoWriter Output

  Timed out after 10.001s.
  Expected
      <int>: 1
  to equal
      <int>: 2
  In [It] at: /home/runner/work/fabric/fabric/orderer/consensus/etcdraft/chain_test.go:1901
------------------------------

tock-ibm:
[logs_6857.zip](https://github.com/hyperledger/fabric/files/9495557/logs_6857.zip)
Logs

Param-S:
Another use case failing in the same package with the following stack: 

**Logs:** https://github.com/hyperledger/fabric/actions/runs/3030930972/jobs/4896676060

```
Chain
/home/runner/work/fabric/fabric/orderer/consensus/etcdraft/chain_test.go:101
  3-node Raft cluster
  /home/runner/work/fabric/fabric/orderer/consensus/etcdraft/chain_test.go:1565
    reconfiguring raft cluster
    /home/runner/work/fabric/fabric/orderer/consensus/etcdraft/chain_test.go:1715
      reconfiguration
      /home/runner/work/fabric/fabric/orderer/consensus/etcdraft/chain_test.go:1823
        [It] removes leader from replica set
[...]

PR #3640, opened 2022-09-18
unit test flake in leadership transfer

Param-S:
#### Type of change
Bug fix

#### Description
When the leadership transfer not picked up the next leader, send out artificial MsgTimeOut msg to induce election.

#### Related issues
#3629 

### DIFF
--- a/orderer/consensus/etcdraft/chain_test.go
+++ b/orderer/consensus/etcdraft/chain_test.go
@@ -2385,11 +2385,21 @@ var _ = Describe("Chain", func() {
 					var newLeader, remainingFollower *chain
 					var c2state raft.SoftState
 					var c3state raft.SoftState
+					retry := 1
 					for newLeader == nil || remainingFollower == nil {
 						select {
 						case c2state = <-c2.observe:
 						case c3state = <-c3.observe:
 						case <-time.After(LongEventualTimeout):
+							// abdicateleader might fail to transfer the leadership when the next candidate
+							// busy with applying committed entries; in that case,
+							// send an artificial MsgTimeoutNow to node to pick next leader
+							if retry > 0 {
+								retry -= 1
+								By("leadership transfer not complete, hence retrying")
+								c2.Consensus(&orderer.ConsensusRequest{Payload: protoutil.MarshalOrPanic(&raftpb.Message{Type: raftpb.MsgTimeoutNow, To: 2})}, 0)
+								continue
+							}
 							Fail("Expected a new leader to present")
 						}
 


orderer/consensus/etcdraft/chain_test.go
```diff
@@ -2385,11 +2385,20 @@ var _ = Describe("Chain", func() {
 					var newLeader, remainingFollower *chain
 					var c2state raft.SoftState
 					var c3state raft.SoftState
+					retry := 1
```
yacovm, line 2388: I might be missing something, but, isn't `retry` always 1?

orderer/consensus/etcdraft/chain_test.go
```diff
@@ -2385,11 +2385,20 @@ var _ = Describe("Chain", func() {
...
 						select {
 						case c2state = <-c2.observe:
 						case c3state = <-c3.observe:
 						case <-time.After(LongEventualTimeout):
+							// abdicateleader might fail to transfer the leadership when the next candidate
+							// busy with applying committed entries; in that case,
+							// send an artificial MsgTimeoutNow to node to pick next leader
+							if retry > 0 {
+								By("leadership transfer not complete, hence retrying")
+								c2.Consensus(&orderer.ConsensusRequest{Payload: protoutil.MarshalOrPanic(&raftpb.Message{Type: raftpb.MsgTimeoutNow, To: 2})}, 0)
```
yacovm, line 2400: I'm not sure we should be sending artificial protocol messages in the test. This makes the test deviate from the protocol we are testing, no?

orderer/consensus/etcdraft/chain_test.go
Param-S: Its mistake, suppose to decrement the value inside the if block. corrected it now.

orderer/consensus/etcdraft/chain_test.go
Param-S: As per the design, if the leadership transfer is not completed then it is defined to complete in the next election timeout(ELECTION_TICK). Since we don't manipulate the election timeout, there are places in the unit test, we send out artificial MsgTimeOut to force the election(for only testing purpose).  It is documented/comment here: 
https://github.com/hyperledger/fabric/blob/2ef46bbc718de220bbd33ee4e1ce65194b963358/orderer/consensus/etcdraft/chain_test.go#L50

yacovm: OK and why are we sending to (2) and not to (3) also?

Param-S: we can send it to any node.  if we send it to multiple nodes at the same time, there is probable chance for split votes.

yacovm (approved):

PR #3642, opened 2022-09-20
verify leadership transfer status before waiting

Param-S:
#### Type of change
- Bug fix

#### Description
The thread which was waiting on the notifyc channel was not notified as the other thread well passed the execution before the channel gets established/initialized. Its a race condition between these 2 threads lead to the test case failure. Added additional check to check the status before waiting for notification from other go routine.

#### Related issues
#3629 

### DIFF
--- a/orderer/consensus/etcdraft/node.go
+++ b/orderer/consensus/etcdraft/node.go
@@ -276,6 +276,16 @@ func (n *node) abdicateLeader(currentLead uint64) {
 	timer := n.clock.NewTimer(time.Duration(n.config.ElectionTick) * n.tickInterval)
 	defer timer.Stop() // prevent timer leak
 
+	// Revalidate the status before waiting on notifyc channel as Ready go routine
+	// well passed the channel notification before establishing the channel
+	if status.ID != currentLead {
+		status = n.Status()
+		if status.Lead != raft.None && status.Lead != currentLead {
+			n.logger.Infof("Leadership transfer completed from %d to %d", currentLead, status.Lead)
+			return
+		}
+	}
+
 	select {
 	case <-timer.C():
 		n.logger.Warn("Leader transfer timeout")


orderer/consensus/etcdraft/node.go
```diff
@@ -276,6 +276,16 @@ func (n *node) abdicateLeader(currentLead uint64) {
 	timer := n.clock.NewTimer(time.Duration(n.config.ElectionTick) * n.tickInterval)
 	defer timer.Stop() // prevent timer leak
 
+	// Revalidate the status before waiting on notifyc channel as Ready go routine
+	// well passed the channel notification before establishing the channel
+	if status.ID != currentLead {
```
yacovm, line 281: what if we are a follower and we called `abdicateLeader()` but in the meanwhile there were two leadership changes that resulted in changing from node `i` to `j` and then back to `i` ? Then we won't notice the leadership change. 

Why not instead of looking at who is the leader, we will look at the `Term` from the `status` ? 
After all, each term has its own leader...

orderer/consensus/etcdraft/node.go
```diff
@@ -276,6 +276,16 @@ func (n *node) abdicateLeader(currentLead uint64) {
 	timer := n.clock.NewTimer(time.Duration(n.config.ElectionTick) * n.tickInterval)
 	defer timer.Stop() // prevent timer leak
 
+	// Revalidate the status before waiting on notifyc channel as Ready go routine
```
yacovm, line 279: One thing I do not understand is, why do we even wait for the leadership to be transferred before configuring the communication framework, if we're a follower? 

The one that is doing the leadership transfer is the leader. The followers have no part in this.

yacovm: Oh, I see - we want the followers to be able to absorb the abdication, so we delay their communication re-configuration

yacovm:
But even without this abdication, if the leader is removed, the other two nodes should be able to select a new leader. The test, tests this property. so it should succeed even if at the time of abdication, the follower didn't hear from the previous leader.

Param-S:
> But even without this abdication, if the leader is removed, the other two nodes should be able to select a new leader. The test, tests this property. so it should succeed even if at the time of abdication, the follower didn't hear from the previous leader.

the test-case tests the property of whether **_all the nodes_** is able to apply/commit the configuration change while rotating the current leader certificates.  https://github.com/hyperledger/fabric/blob/d18be4662d74bc3abef1b95fc703c62b4c80bfd5/orderer/consensus/etcdraft/chain_test.go#L1901 In the case of failure, one of the node was blocked on the leadership change notification condition, the node did not proceed further to make the configuration update which made the test to time out. 

If we increase the test timeout value or decrease the election tick value, the blocked thread would timeout and hence the testcase would pass. This is another option. But, it does not solve the real race condition between the node Ready loop thread and the abdicate leader thread.  As the current logic depends on the election_tick timeout, the node is not latest wrt to the configuration changes till that time as compared to other nodes in the quorum.

Param-S:
leadership change is redesigned in this PR #3651. This change is no more relevant.

PR #3648, opened 2022-09-26
GitHub Workflows security hardening

sashashura:
This PR adds explicit [permissions section](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#permissions) to workflows. This is a security best practice because by default workflows run with [extended set of permissions](https://docs.github.com/en/actions/security-guides/automatic-token-authentication#permissions-for-the-github_token) (except from `on: pull_request` [from external forks](https://securitylab.github.com/research/github-actions-preventing-pwn-requests/)). By specifying any permission explicitly all others are set to none. By using the principle of least privilege the damage a compromised workflow can do (because of an [injection](https://securitylab.github.com/research/github-actions-untrusted-input/) or compromised third party tool or action) is restricted.
It is recommended to have [most strict permissions on the top level](https://github.com/ossf/scorecard/blob/main/docs/checks.md#token-permissions) and grant write permissions on [job level](https://docs.github.com/en/actions/using-jobs/assigning-permissions-to-jobs) case by case.

### DIFF
--- a/.github/workflows/verify-build.yml
+++ b/.github/workflows/verify-build.yml
@@ -15,6 +15,9 @@ env:
   PATH: /opt/go/bin:/bin:/usr/bin:/sbin:/usr/sbin:/usr/local/bin:/usr/local/sbin
   GO_VER: 1.18.2
 
+permissions:
+  contents: read # to fetch code (actions/checkout)
+
 jobs:
   basic-checks:
     name: Basic Checks
--- a/.github/workflows/slash-commands.yml
+++ b/.github/workflows/slash-commands.yml
@@ -8,8 +8,12 @@ on:
     types:
       - created
       - edited
+permissions: {}
 jobs:
   notify:
+    permissions:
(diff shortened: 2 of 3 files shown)


.github/workflows/verify-build.yml
```diff
@@ -15,6 +15,9 @@ env:
   PATH: /opt/go/bin:/bin:/usr/bin:/sbin:/usr/sbin:/usr/local/bin:/usr/local/sbin
   GO_VER: 1.18.2
 
+permissions:
```
ale-linux, line 18: don't we miss the
```
permissions: {}
```
directive here?

sashashura: {} means none. It is is also possible, but then we need to add read in every single workflow that does check out.

sashashura:
Do you have any questions?

denyeart (approved):

denyeart:
Thanks @sashashura , good improvement!

PR #3651, opened 2022-09-28
Transfer leadership before submitting to raft library

Param-S:
#### Type of change
- Bug fix

#### Description

If transaction request rotates the current leader cert or removes it then transfer the leadership to another node before proposing the transaction for orderering. It would help to avoid the leadership transfer while committing the block

#### Related issues
#3629 

### DIFF
--- a/orderer/consensus/etcdraft/chain.go
+++ b/orderer/consensus/etcdraft/chain.go
@@ -10,7 +10,6 @@ import (
 	"context"
 	"encoding/pem"
 	"fmt"
-	"runtime/debug"
 	"sync"
 	"sync/atomic"
 	"time"
@@ -396,40 +395,6 @@ func (c *Chain) Order(env *common.Envelope, configSeq uint64) error {
 // Configure submits config type transactions for ordering.
 func (c *Chain) Configure(env *common.Envelope, configSeq uint64) error {
 	c.Metrics.ConfigProposalsReceived.Add(1)
-
-	payload, err := protoutil.UnmarshalPayload(env.Payload)
-	if err != nil {
-		return errors.Wrap(err, "failed to extract payload from config envelope")
-	}
-
-	// get config update
-	configUpdate, err := configtx.UnmarshalConfigUpdateFromPayload(payload)
-	if err != nil {
-		return errors.Wrap(err, "could not read config update")
-	}
-
-	configMeta, err := MetadataFromConfigUpdate(configUpdate)
-	if err != nil {
-		return errors.Wrap(err, "could not read config metadata")
-	}
-
-	if configMeta != nil {
-		membershipUpdates, err := ComputeMembershipChanges(c.opts.BlockMetadata, c.opts.Consenters, configMeta.Consenters)
-		if err != nil {
-			c.logger.Panicf("illegal configuration change detected: %s", err)
-		}
-
-		if membershipUpdates.Rotated() {
-			lead := atomic.LoadUint64(&c.lastKnownLeader)
-			fmt.Printf("My raftID: %d Leader raftID: %d rotatedNode raftID: %d\n", c.raftID, lead, membershipUpdates.RotatedNode)
-			if membershipUpdates.RotatedNode == lead && lead == c.raftID {
-				c.logger.Infof("Certificate of Raft leader is being rotated, attempt leader transfer before reconfiguring communication")
-				debug.PrintStack()
-				c.Node.abdicateLeader(lead)
-			}
-		}
-	}
-
 	return c.Submit(&orderer.SubmitRequest{LastValidationSeq: configSeq, Payload: env, Channel: c.channelID}, 0)
 }
 
@@ -744,11 +709,16 @@ func (c *Chain) run() {
 				continue
 			}
 
-			batches, pending, err := c.ordered(s.req, bc)
+			batches, pending, err := c.ordered(s.req)
 			if err != nil {
 				c.logger.Errorf("Failed to order message: %s", err)
 				continue
 			}
+
+			if !pending && len(batches) == 0 {
+				continue
+			}
+
 			if pending {
 				startTimer() // no-op if timer is already started
 			} else {
@@ -924,11 +894,13 @@ func (c *Chain) writeBlock(block *common.Block, index uint64) {
 
 // Orders the envelope in the `msg` content. SubmitRequest.
 // Returns
-//   -- batches [][]*common.Envelope; the batches cut,
-//   -- pending bool; if there are envelopes pending to be ordered,
-//   -- err error; the error encountered, if any.
+//
+//	-- batches [][]*common.Envelope; the batches cut,
+//	-- pending bool; if there are envelopes pending to be ordered,
+//	-- err error; the error encountered, if any.
+//
 // It takes care of config messages as well as the revalidation of messages if the config sequence has advanced.
-func (c *Chain) ordered(msg *orderer.SubmitRequest, bc *blockCreator) (batches [][]*common.Envelope, pending bool, err error) {
+func (c *Chain) ordered(msg *orderer.SubmitRequest) (batches [][]*common.Envelope, pending bool, err error) {
 	seq := c.support.Sequence()
 
 	isconfig, err := c.isConfig(msg.Payload)
@@ -947,13 +919,20 @@ func (c *Chain) ordered(msg *orderer.SubmitRequest, bc *blockCreator) (batches [
 			}
 		}
 
+		if c.checkForEvictionNCertRotation(msg.Payload) {
+			go func() {
+				c.Node.abdicateLeader(c.raftID)
+				c.Submit(msg, 0)
+			}()
+			return nil, false, nil
+		}
+
 		batch := c.support.BlockCutter().Cut()
 		batches = [][]*common.Envelope{}
 		if len(batch) != 0 {
 			batches = append(batches, batch)
 		}
 		batches = append(batches, []*common.Envelope{msg.Payload})
-
 		return batches, false, nil
 	}
 	// it is a normal message
@@ -1125,6 +1104,10 @@ func (c *Chain) apply(ents []raftpb.Entry) {
 				continue
 			}
 
+			if err := c.Node.storage.Sync(); err != nil {
+				c.logger.Debugf("Failed to sync raft log, error: %s", err)
+			}
+
 			c.confState = *c.Node.ApplyConfChange(cc)
 
 			switch cc.Type {
@@ -1150,17 +1133,9 @@ func (c *Chain) apply(ents []raftpb.Entry) {
 				c.Metrics.ClusterSize.Set(float64(len(c.opts.BlockMetadata.ConsenterIds)))
 			}
 
-			lead := atomic.LoadUint64(&c.lastKnownLeader)
-			removeLeader := cc.Type == raftpb.ConfChangeRemoveNode && cc.NodeID == lead
 			shouldHalt := cc.Type == raftpb.ConfChangeRemoveNode && cc.NodeID == c.raftID
-
 			// unblock `run` go routine so it can still consume Raft messages
 			go func() {
-				if removeLeader {
-					c.logger.Infof("Current leader is being removed from channel, attempt leadership transfer")
-					c.Node.abdicateLeader(lead)
-				}
-
 				if configureComm && !shouldHalt { // no need to configure comm if this node is going to halt
 					if err := c.configureComm(); err != nil {
 						c.logger.Panicf("Failed to configure communication: %s", err)
@@ -1333,22 +1308,10 @@ func (c *Chain) writeConfigBlock(block *common.Block, index uint64) {
 			}
 
 			c.configInflight = true
-		} else if configMembership.Rotated() {
-			/* 			lead := atomic.LoadUint64(&c.lastKnownLeader)
-			   			if configMembership.RotatedNode == lead {
-			   				c.logger.Infof("Certificate of Raft leader is being rotated, attempt leader transfer before reconfiguring communication")
-			   				debug.PrintStack()
-			   				go func() {
-			   					c.Node.abdicateLeader(lead)
-			   					if err := c.configureComm(); err != nil {
-			   						c.logger.Panicf("Failed to configure communication: %s", err)
-			   					}
-			   				}()
-			   			} else { */
+		} else {
 			if err := c.configureComm(); err != nil {
 				c.logger.Panicf("Failed to configure communication: %s", err)
 			}
-			//}
 		}
 
 	case common.HeaderType_ORDERER_TRANSACTION:
@@ -1526,4 +1489,46 @@ func (c *Chain) triggerCatchup(sn *raftpb.Snapshot) {
 	case c.snapC <- sn:
 	case <-c.doneC:
 	}
+}
+
+// checkForEvictionNCertRotation checks for node eviction and
+// certificate rotation, return true if request includes it
+// otherwise returns false
+func (c *Chain) checkForEvictionNCertRotation(env *common.Envelope) bool {
+	payload, err := protoutil.UnmarshalPayload(env.Payload)
+	if err != nil {
+		c.logger.Warnf("failed to extract payload from config envelope: %s", err)
+		return false
+	}
+
+	configUpdate, err := configtx.UnmarshalConfigUpdateFromPayload(payload)
+	if err != nil {
+		c.logger.Warnf("could not read config update: %s", err)
+		return false
+	}
+
+	configMeta, err := MetadataFromConfigUpdate(configUpdate)
+	if err != nil || configMeta == nil {
+		c.logger.Warnf("could not read config metadata: %s", err)
+		return false
+	}
+
+	membershipUpdates, err := ComputeMembershipChanges(c.opts.BlockMetadata, c.opts.Consenters, configMeta.Consenters)
+	if err != nil {
+		c.logger.Warnf("illegal configuration change detected: %s", err)
+		return false
+	}
+
+	if membershipUpdates.RotatedNode == c.raftID {
+		c.logger.Infof("Detected certificate rotation of our node")
+		return true
+	}
+
+	if _, exists := membershipUpdates.NewConsenters[c.raftID]; !exists {
+		c.logger.Infof("Detected eviction of ourselves from the configuration")
+		return true
+	}
+
+	c.logger.Debugf("Node %d is still part of the consenters set", c.raftID)
+	return false
 }
--- a/orderer/consensus/etcdraft/chain.go
+++ b/orderer/consensus/etcdraft/chain.go
@@ -204,7 +204,9 @@ type Chain struct {
 	status            types.Status
 
 	// BCCSP instance
-	CryptoProvider bccsp.BCCSP
+	CryptoProvider              bccsp.BCCSP
+	leadershipTransferCandidate uint64
+	leaderTransferFailed        uint32
 }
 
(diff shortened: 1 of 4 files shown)


orderer/consensus/etcdraft/chain.go
```diff
@@ -394,6 +396,37 @@ func (c *Chain) Order(env *common.Envelope, configSeq uint64) error {
...
+		c.logger.Panicf("illegal configuration change detected: %s", err)
+	}
+
+	if membershipUpdates.Rotated() {
+		lead := atomic.LoadUint64(&c.lastKnownLeader)
+		fmt.Printf("My raftID: %d Leader raftID: %d rotatedNode raftID: %d\n", c.raftID, lead, membershipUpdates.RotatedNode)
+		if membershipUpdates.RotatedNode == lead {
+			c.logger.Infof("Certificate of Raft leader is being rotated, attempt leader transfer before reconfiguring communication")
+			debug.PrintStack()
+			c.Node.abdicateLeader(lead)
```
yacovm: 1. You also need to forward the request to the next leader. 
2. You need to abdicate only if the removed node is you, and you're the leader.

Param-S: > You need to abdicate only if the removed node is you, and you're the leader.

Updated the condition:
`if membershipUpdates.RotatedNode == lead && lead == c.raftID {`

Param-S: > You also need to forward the request to the next leader.

It is taken care already.

orderer/consensus/etcdraft/node.go
```diff
@@ -266,23 +273,27 @@ func (n *node) abdicateLeader(currentLead uint64) {
...
-	case l := <-notifyc:
-		n.logger.Infof("Leader has been transferred from %d to %d", currentLead, l)
-	case <-n.chain.doneC:
+		select {
+		case <-timer.C():
+			n.logger.Warn("Leader transfer timeout")
+			return transferee, LEADERTRANSFERTIMEOUT
+		case l := <-notifyc:
+			n.logger.Infof("Leader has been transferred from %d to %d", currentLead, l)
+		case <-n.chain.doneC:
```
yacovm, line 295: shouldn't we put `return raft.None, nil` here?

orderer/consensus/etcdraft/node.go
```diff
@@ -19,10 +19,17 @@ import (
...
 	"github.com/hyperledger/fabric/common/flogging"
 	"github.com/hyperledger/fabric/protoutil"
+	"github.com/pkg/errors"
 	"go.etcd.io/etcd/raft/v3"
 	"go.etcd.io/etcd/raft/v3/raftpb"
 )
 
+var (
+	LEADERTRANSFERNOTRANSFEREE    = errors.New("leadership transfer failed to identify transferee")
+	LEADERTRANSFERTIMEOUT         = errors.New("leadership transfer timedout")
```
yacovm: "timed out", in two words.

orderer/consensus/etcdraft/chain.go
```diff
@@ -911,6 +921,41 @@ func (c *Chain) ordered(msg *orderer.SubmitRequest) (batches [][]*common.Envelop
...
 
+		if !c.leaderTransferFailed {
+			if c.checkForEvictionNCertRotation(msg.Payload) {
+				go func() {
+					if transferee, err := c.Node.abdicateLeader(c.raftID); err != nil {
+						if err.Error() == LEADERTRANSFERTIMEOUT.Error() && transferee != raft.None {
+							if transferee != c.leadershipTransferCandidate {
+								c.leadershipTransferCandidate = transferee
+							} else {
+								c.leaderTransferFailed = true
```
yacovm: you're writing to a variable `leaderTransferFailed` in a goroutine without but reading it from the goroutine spawned. This needs to be synchronized.

orderer/consensus/etcdraft/chain.go
```diff
@@ -911,6 +921,41 @@ func (c *Chain) ordered(msg *orderer.SubmitRequest) (batches [][]*common.Envelop
 			}
 		}
 
+		if !c.leaderTransferFailed {
```
yacovm: if the transaction has nothing to do with removal of the leader, you still set the transferFailed to false, why?

orderer/consensus/etcdraft/chain.go
```diff
@@ -911,6 +921,41 @@ func (c *Chain) ordered(msg *orderer.SubmitRequest) (batches [][]*common.Envelop
...
+			if c.checkForEvictionNCertRotation(msg.Payload) {
+				go func() {
+					if transferee, err := c.Node.abdicateLeader(c.raftID); err != nil {
+						if err.Error() == LEADERTRANSFERTIMEOUT.Error() && transferee != raft.None {
+							if transferee != c.leadershipTransferCandidate {
```
yacovm: I don't understand why we need this transfer candidate and transfer failed fields? 

Can't we simply this?

```
		if c.checkForEvictionNCertRotation(msg.Payload) {
			if ! atomic.CompareAndSwapUint32(&c.leadershipTransferInProgress, 0, 1) {
				return nil, false, errors.Errorf("leadership transfer is currently in progress")
			}
			go func() {
				
				defer func() {
					atomic.StoreUint32(&c.leadershipTransferInProgress, 0)	
				}()
				
				if transferee, err := c.Node.abdicateLeader(c.raftID); err != nil {
					... 

``

File: orderer/consensus/etcdraft/chain.go (hunk 4 above)
Comment (Param-S) [reply orderer/consensus/etcdraft/chain.go]: Updated with atomic constructs now.

File: orderer/consensus/etcdraft/chain.go (hunk 5 above)
Comment (Param-S) [reply orderer/consensus/etcdraft/chain.go]: now the `else block` resets only if leaderTransferFailed is set in previous iteration.

File: orderer/consensus/etcdraft/chain.go (hunk 6 above)
Comment (Param-S) [reply orderer/consensus/etcdraft/chain.go]: These 2 variables need to control the flow wrt the leadership timeout and retries. 
it would limit the leadership transfer attempt for a given node not more than 2 times.

**Comment by yacovm** [comment]
@Param-S please check my latest update to your PR.

@tock-ibm can you please review the PR?

File: orderer/consensus/etcdraft/chain.go [hunk 7]
```diff
@@ -921,37 +922,42 @@ func (c *Chain) ordered(msg *orderer.SubmitRequest) (batches [][]*common.Envelop
...
-							select {
-							case <-timer.C():
-							case <-c.Node.chain.doneC:
-							}
-						} else if err.Error() == LEADERTRNSFERCALLEDBYFOLLOWER.Error() {
-							c.logger.Debugf("Leadership transfer completed")
-						} else {
-							atomic.StoreUint32(&c.leaderTransferFailed, 1)
-						}
+		if c.checkForEvictionNCertRotation(msg.Payload) && atomic.CompareAndSwapUint32(&c.leadershipTransferInProgress, 0, 1) {
```
Param-S: If another config transaction arrives while the leadershipTransferInProgress, it would proceed further down to propose. 

1) I think, we need to handle it by rejecting it.

orderer/consensus/etcdraft/chain.go
```diff
@@ -921,37 +922,42 @@ func (c *Chain) ordered(msg *orderer.SubmitRequest) (batches [][]*common.Envelop
...
+			go func() {
+				if err := c.Node.abdicateLeader(c.raftID); err != nil {
+					// If there is no leader, abort and do not retry.
+
+					// Return early to prevent re-submission of the transaction
+					if err == ErrNoLeader || err == ErrChainHalting {
+						atomic.StoreUint32(&c.consecutiveAbdicationFailures, 0)
```
Param-S: I think, we also need to reset the `leadershipTransferInProgress`?

orderer/consensus/etcdraft/chain.go
```diff
@@ -921,37 +922,42 @@ func (c *Chain) ordered(msg *orderer.SubmitRequest) (batches [][]*common.Envelop
...
+						return
 					}
-					if err := c.Submit(msg, 0); err != nil {
-						c.logger.Errorf("Transaction proposal failed with error: %v", err)
+
+					if atomic.LoadUint32(&c.consecutiveAbdicationFailures) == 2 {
+						c.logger.Warnw("Abdication failed too many times consecutively, aborting retries")
+						// Reset failure counter as we do not retry anymore.
+						atomic.StoreUint32(&c.consecutiveAbdicationFailures, 0)
```
Param-S: here too, we need to reset `leadershipTransferInProgress`

orderer/consensus/etcdraft/chain.go
```diff
@@ -921,37 +922,42 @@ func (c *Chain) ordered(msg *orderer.SubmitRequest) (batches [][]*common.Envelop
...
+
+					// Else, it's one of the errors above, so we increment the consecutive failure counter and retry.
+					atomic.AddUint32(&c.consecutiveAbdicationFailures, 1)
+				} else {
+					atomic.StoreUint32(&c.consecutiveAbdicationFailures, 0)
+				}
+
+				atomic.StoreUint32(&c.leadershipTransferInProgress, 0)
+
+				if err := c.Submit(msg, 0); err != nil {
```
Param-S: instead of following the submit workflow, can we use the simple loop within the go routine to retry abdicateLeader with a threadhold

```
     if c.checkForEvictionNCertRotation(msg.Payload)  {
                if atomic.CompareAndSwapUint32(&c.leadershipTransferInProgress, 0, 1) {
        		go func() {
                                retryCount = 0
        			for {
                                    if err := c.Node.abdicateLeader(c.raftID); err != nil {
                                          retryCount += 1
                                          if err == ErrNoLeader || err == ErrChainHalting || retryCount == 2 {
                                             atomic.StoreUint32(&c.leadershipTransferInProgress, 0)
                                              return
                                          }
                                    } else {
                                          atomic.StoreUint32(&c.leadershipTransferInProgress, 0)
                                          return
                                    }
                               }
                           }
                  } else {
                           // reject the submit request as already config transaction inprogress
                  }
```

orderer/consensus/etcdraft/chain.go
yacovm: why?

Param-S: If other transaction is also removing the current leader, it would propose the transaction before transferring the leadership right?

yacovm:
Thanks, addressed comments.

orderer/consensus/etcdraft/chain.go
yacovm: oh right

orderer/consensus/etcdraft/chain.go
```diff
@@ -921,37 +925,43 @@ func (c *Chain) ordered(msg *orderer.SubmitRequest) (batches [][]*common.Envelop
...
-							}
-						} else if err.Error() == LEADERTRNSFERCALLEDBYFOLLOWER.Error() {
-							c.logger.Debugf("Leadership transfer completed")
-						} else {
-							atomic.StoreUint32(&c.leaderTransferFailed, 1)
+		if c.checkForEvictionNCertRotation(msg.Payload) {
+
+			if !atomic.CompareAndSwapUint32(&c.leadershipTransferInProgress, 0, 1) {
+				c.logger.Warnf("A reconfiguration transaction is already in progress, ignoring a subsequent transaction")
+				return
```
Param-S, line 932: this return statement would return 3 values(batches [][]*common.Envelope, pending bool, err error). do we need to be explicitly return values here?

yacovm: it will return zero values, and then in line 724 it will continue, no?

Param-S: Yes, It will continue. 

Do we need to be explicitly quote the return value like line no. https://github.com/hyperledger/fabric/blob/588ab20a3d00fd88d5b10792b9984a0096d8b4d3/orderer/consensus/etcdraft/chain.go#L964 for code readability

yacovm: I don't think so...

orderer/consensus/etcdraft/node.go
```diff
@@ -228,60 +237,81 @@ func (n *node) send(msgs []raftpb.Message) {
...
 
-// If this is called on leader, it picks a node that is
-// recently active, and attempt to transfer leadership to it.
-// If this is called on follower, it simply waits for a
-// leader change till timeout (ElectionTimeout).
-func (n *node) abdicateLeader(currentLead uint64) {
+// abdicateLeader picks a node that is recently active, and attempts to transfer leadership to it.
+// Receives the current leader (our id) as input.
+// Returns error upon failure.
+func (n *node) abdicateLeader(selfID uint64) error {
```
tock-ibm: The self ID is accessible from the `node`: `n.config.ID`, hence you don't need to pass it in, the node knows its self ID.
Or `status.ID`, same info.

tock-ibm: Nit: probably should be called `abdicateLeadership()`, as this only applies to a node which is the leader, no?

orderer/consensus/etcdraft/node.go
```diff
@@ -228,60 +237,81 @@ func (n *node) send(msgs []raftpb.Message) {
...
+
 	status := n.Status()
 
-	if status.Lead != raft.None && status.Lead != currentLead {
+	if status.Lead == raft.None {
+		n.logger.Warn("No leader, cannot transfer leadership")
+		return ErrNoLeader
+	}
+
+	if status.Lead != raft.None && status.Lead != selfID {
```
tock-ibm: above you are checking if `status.Lead == raft.None` and returning. So there is no need to check it here again...
This condition `status.Lead != raft.None` will always be true...

orderer/consensus/etcdraft/chain.go
```diff
@@ -911,6 +925,45 @@ func (c *Chain) ordered(msg *orderer.SubmitRequest) (batches [][]*common.Envelop
...
+			go func() {
+				defer atomic.StoreUint32(&c.leadershipTransferInProgress, 0)
+
+				for attempt := 1; attempt <= AbdicationMaxAttempts; attempt++ {
+					if err := c.Node.abdicateLeader(c.raftID); err != nil {
+						// If there is no leader, abort and do not retry.
+						// Return early to prevent re-submission of the transaction
+						if err == ErrNoLeader || err == ErrChainHalting {
+							return
+						}
```
tock-ibm, line 944: why aren't you returning if `ErrNotALeader`? I mean, a node needs to abdicate only if it was the leader, and saw a removal or cert rotation for itself, right? so if it is no longer the leader, why continue and not return?

orderer/consensus/etcdraft/node.go
```diff
@@ -228,60 +237,81 @@ func (n *node) send(msgs []raftpb.Message) {
...
+		select {
+		case <-timer.C():
+			n.logger.Warn("Leader transfer timed out")
+			return ErrTimedOutLeaderTransfer
+		case l := <-notifyc:
+			if l != selfID {
+				n.logger.Infof("Leader has been transferred from %d to %d", selfID, l)
+				return nil
+			}
+			n.logger.Infof("Notified about leader being %d but we expect someone else to be the leader", l)
```
tock-ibm: if the node was notified via this channel `notifyc`, then it will not be notified again, as L166 sets nil to the channel in got from `n.subscriberC`. You have to subscribe again or it will be a timeout or a halt, right?

But, subscribing again is creates a window where the event will not be delivered at all, as it might just happen when `notifyLeaderChangeC` is still `nil`... however, it reduces the chance of a TO if the event does happen eventually.

orderer/consensus/etcdraft/chain.go
```diff
@@ -1489,3 +1527,45 @@ func (c *Chain) triggerCatchup(sn *raftpb.Snapshot) {
...
+	membershipUpdates, err := ComputeMembershipChanges(c.opts.BlockMetadata, c.opts.Consenters, configMeta.Consenters)
+	if err != nil {
+		c.logger.Warnf("illegal configuration change detected: %s", err)
+		return false
+	}
+
+	if membershipUpdates.RotatedNode == c.raftID {
+		c.logger.Infof("Detected certificate rotation of our node")
+		return true
+	}
```
tock-ibm, line 1562: Is it possible to rotate the cert of a 1-node cluster? what happens then when the node tries to abdicate?

tock-ibm:
@Param-S @yacovm I have a possibly unrelated question:
Lets assume all goes well, the leader abdicates and forwards the config block, the new leader `A` commits the block, and then just before in starts the goroutine that does the `ProposeConfChange` to raft, it fails. What mechanism is there so that the new leader `A` after proposes the   `ProposeConfChange` to raft?

yacovm:
> @Param-S @yacovm I have a possibly unrelated question: Lets assume all goes well, the leader abdicates and forwards the config block, the new leader `A` commits the block, and then just before in starts the goroutine that does the `ProposeConfChange` to raft, it fails. What mechanism is there so that the new leader `A` after proposes the `ProposeConfChange` to raft?

[This](https://github.com/hyperledger/fabric/blob/main/orderer/consensus/etcdraft/chain.go#L646-L656) code section.

orderer/consensus/etcdraft/chain.go
yacovm: Good question. I think it will fail. But in any case, a single node Raft cluster is not something fit for production... I think we can leave this as is.

orderer/consensus/etcdraft/node.go
yacovm: Correct, good observation. 

I changed the mechanism to accommodate the fact that the only goroutine that writes into `n.subscriberC` is this one, and only one instance of this goroutine can run at a given time, as the chain's `leadershipTransferInProgress` guards from activating more than once at once.

Let me know if you think it's better.

orderer/consensus/etcdraft/chain.go
yacovm: You have a point, but returning would not be a good decision, as it would result in the transaction being dropped. 
I changed the code and instead of returning `ErrNotALeader` we now return `nil`, and this causes the forwarding of the transaction to the leader.

orderer/consensus/etcdraft/node.go
yacovm: Changed accordingly.

orderer/consensus/etcdraft/node.go
yacovm: Changed accordingly.

tock-ibm:
> > @Param-S @yacovm I have a possibly unrelated question: Lets assume all goes well, the leader abdicates and forwards the config block, the new leader `A` commits the block, and then just before in starts the goroutine that does the `ProposeConfChange` to raft, it fails. What mechanism is there so that the new leader `A` after proposes the `ProposeConfChange` to raft?
> 
> [This](https://github.com/hyperledger/fabric/blob/main/orderer/consensus/etcdraft/chain.go#L646-L656) code section.

OK, that answers it.

tock-ibm (approved):

orderer/consensus/etcdraft/chain.go
tock-ibm: Sounds plausible.

orderer/consensus/etcdraft/node.go
tock-ibm: Looks OK to me

orderer/consensus/etcdraft/chain.go
tock-ibm: Good solution

andrew-coleman (approved):

denyeart:

PR #3681, opened 2022-10-11
Use env variables to initialize the PKCS11 BCCSP

jpatchell:
The environment variables to configure the PKCS11 BCCSP are ignored for the peer node, such as CORE_PEER_BCCSP_PKCS11_LIBRARY. This change allows specifying the BCCSP Default and to configure all of the PKCS11 BCCSP settings using environment variables.

Signed-off-by: Jonathan Patchell <Jonathan.Patchell@thalesgroup.com>

#### Type of change

- Bug fix

#### Description

When integrating the peer node with an HSM, it isn't possible to specify environment variables to configure the PKCS11 BCCSP. With these changes the environment variables are no longer ignored allowing configuration from environment variables for the  PKCS11 BCCSP.

#### Related issues

#1900

### DIFF
--- a/internal/peer/common/common.go
+++ b/internal/peer/common/common.go
@@ -145,6 +145,16 @@ func InitConfig(cmdRoot string) error {
 	return nil
 }
 
+// InitBCCSPConfig initializes BCCSP config
+func InitBCCSPConfig(bccspConfig *factory.FactoryOpts) error {
+	SetBCCSPKeystorePath()
+	SetBCCSPPKCS11Keys()
+	if err := viper.UnmarshalKey("peer.BCCSP", &bccspConfig); err != nil {
+		return errors.WithMessage(err, "could not decode peer BCCSP configuration")
+	}
+	return nil
+}
+
 // InitCrypto initializes crypto for this peer
 func InitCrypto(mspMgrConfigDir, localMSPID, localMSPType string) error {
 	// Check whether msp folder exists
@@ -160,10 +170,10 @@ func InitCrypto(mspMgrConfigDir, localMSPID, localMSPType string) error {
 	}
 
 	// Init the BCCSP
-	SetBCCSPKeystorePath()
 	bccspConfig := factory.GetDefaultOpts()
-	if err := viper.UnmarshalKey("peer.BCCSP", &bccspConfig); err != nil {
-		return errors.WithMessage(err, "could not decode peer BCCSP configuration")
+	err = InitBCCSPConfig(bccspConfig)
+	if err != nil {
+		return err
 	}
 
 	conf, err := msp.GetLocalMspConfigWithType(mspMgrConfigDir, bccspConfig, localMSPID, localMSPType)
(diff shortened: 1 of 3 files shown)


ale-linux:
Thanks @jpatchell ! @adecaro would you mind taking a look pls?

adecaro:
@jpatchell @ale-linux , it looks good to me, but should we document this somewhere? @denyeart ?

internal/peer/common/commonpkcs11_test.go
```diff
@@ -0,0 +1,49 @@
...
+	"strings"
+	"testing"
+
+	"github.com/hyperledger/fabric/bccsp/factory"
+	"github.com/hyperledger/fabric/internal/peer/common"
+	"github.com/spf13/viper"
+	"github.com/stretchr/testify/require"
+)
+
+func TestBCCSPPKCS11EnvVars(t *testing.T) {
```
pfi79, line 23: this test does not occur in the unit test pipelines
image

C0rWin:
@jpatchell do you know why it works for all other config parameters but not for BCCSP?

internal/peer/common/commonpkcs11_test.go
jpatchell: @pfi79 I added github.com/hyperledger/fabric/internal/peer/common to the pkcs11 unit tests.

jpatchell:
@C0rWin I'm not entirely sure but I believe the difference is that for the other config options there are calls to viper.GetString etc to populate the Config in core/peer/config.go which will fetch the environment variables (I haven't confirmed this). The BCCSP config is getting populated using viper.UnmarshalKey which doesn't query the environment variables recursively, but will get the environment variable for CORE_PEER_BCCSP only. This leaves out all of the PKCS11 BCCSP settings as well as the BCCSP Default value. This change follows suit for how the BCCSP keystore calls SetBCCSPKeystorePath config.GetPath and viper.Set() as this causes the viper.UnmarshalKey to effectively use the environment variables for all of the PKCS11 BCCSP settings.

internal/peer/common/common.go
```diff
@@ -187,6 +197,15 @@ func SetBCCSPKeystorePath() {
 	}
 }
 
+// SetBCCSPPKCS11Keys sets all of the keys for the PKCS11 BCCSP provider
+func SetBCCSPPKCS11Keys() {
+	keys := []string{"peer.BCCSP.Default", "peer.BCCSP.PKCS11.Hash", "peer.BCCSP.PKCS11.Security", "peer.BCCSP.PKCS11.Library", "peer.BCCSP.PKCS11.Label", "peer.BCCSP.PKCS11.Pin"}
```
pfi79: not enough peer.BCCSP.PKCS11.Altid

and not just his

see [bccsp/pkcs11/conf.go](https://github.com/hyperledger/fabric/blob/main/bccsp/pkcs11/conf.go)

jpatchell: I added support for peer.BCCSP.PKCS11 SoftwareVerify, Immutable, AltID, and KeyIDs. KeyIDs requires the use of specifying a json string for the CORE_PEER_BCCSP_PKCS11 environment variable.

pfi79:
Sorry. But I offered my own solution to the problem raised here. https://github.com/hyperledger/fabric/pull/3697

jpatchell:
@pfi79 No worries! Looking forward to having a solution.

denyeart:
Ok, https://github.com/hyperledger/fabric/pull/3697 was merged as an alternative solution, closing this one. Thanks for pushing on this!

PR #3697, opened 2022-10-15
Use env variables to initialize the PKCS11 BCCSP

pfi79:
The environment variables to configure the PKCS11 BCCSP are ignored for the peer node, such as CORE_PEER_BCCSP_PKCS11_LIBRARY. This change allows specifying the BCCSP Default and to configure all of the PKCS11 BCCSP settings using environment variables.

This is an alternative solution to the problem proposed in https://github.com/hyperledger/fabric/pull/3681

Signed-off-by: Fred Partanskiy <pfi79@mail.ru>

### DIFF
--- a/internal/peer/common/common.go
+++ b/internal/peer/common/common.go
@@ -12,7 +12,6 @@ import (
 	"fmt"
 	"io/ioutil"
 	"os"
-	"reflect"
 	"strings"
 	"time"
 
@@ -21,7 +20,6 @@ import (
 	pb "github.com/hyperledger/fabric-protos-go/peer"
 	"github.com/hyperledger/fabric/bccsp"
 	"github.com/hyperledger/fabric/bccsp/factory"
-	"github.com/hyperledger/fabric/bccsp/pkcs11"
 	"github.com/hyperledger/fabric/common/channelconfig"
 	"github.com/hyperledger/fabric/common/flogging"
 	"github.com/hyperledger/fabric/core/config"
@@ -165,7 +163,7 @@ func InitBCCSPConfig(bccspConfig *factory.FactoryOpts) error {
 	opts := viper.DecodeHook(mapstructure.ComposeDecodeHookFunc(
 		mapstructure.StringToTimeDurationHookFunc(),
 		mapstructure.StringToSliceHookFunc(","),
-		StringToKeyIds(),
+		factory.StringToKeyIds(),
 	))
 
 	if err := subv.Unmarshal(&bccspConfig, opts); err != nil {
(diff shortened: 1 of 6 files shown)


internal/peer/common/commonpkcs11_test.go
```diff
@@ -0,0 +1,79 @@
...
+
+	"github.com/hyperledger/fabric/bccsp/factory"
+	"github.com/hyperledger/fabric/core/config/configtest"
+	"github.com/hyperledger/fabric/internal/peer/common"
+	"github.com/mitchellh/mapstructure"
+	"github.com/spf13/viper"
+	"github.com/stretchr/testify/require"
+)
+
+func TestBCCSPPKCS11EnvVars(t *testing.T) {
```
C0rWin, line 24: Having a test it's always nice, however, the goal to have `InitCrypto` tested rather than the approach you added to it because the test has to make sure that if someone touches `InitCrypto` and it doesn't conform with the test, has to fail. 

Could you please change the test to make sure it verifies `InitCrypto`?

pfi79: fix

internal/peer/common/common.go
```diff
@@ -20,6 +21,7 @@ import (
 	pb "github.com/hyperledger/fabric-protos-go/peer"
 	"github.com/hyperledger/fabric/bccsp"
 	"github.com/hyperledger/fabric/bccsp/factory"
+	"github.com/hyperledger/fabric/bccsp/pkcs11"
```
denyeart: This is a common file and not gated by the "pkcs11" build tag.
I believe this solution suffers from the same problem that caused the original solution in #1900 to be reverted. For details see  https://github.com/hyperledger/fabric/pull/1900#issuecomment-713057065.

denyeart (commented):
Do you think the solution offered in #3681 is not workable?

pfi79:
> Do you think the solution offered in #3681 is not workable?

#3681 it works

but my solution is more versatile

I discovered why this function `viper.UnmarshalKey("peer.BCCSP", &bccspConfig)` does not work

internal/peer/common/common.go
pfi79: do you want me to remove the `"github.com/hyperledger/fabric/bccsp/pkcs11"` import here ?

pfi79: @denyeart I've fixed it.

C0rWin:
> Do you think the solution offered in #3681 is not workable?

@denyeart, the IMO solution suggested here generalizes the one provided at #3681, the current one looks good to me if no objections, I would merge.

C0rWin (approved):

denyeart:
Thanks @C0rWin and @pfi79 , I'm glad this is finally resolved!

C0rWin:

C0rWin:

PR #3705, opened 2022-10-19
Commit path changes for private data purge

### LINKED ISSUE #3026: Process purge operation during block commit
Delete from hashed state atomically and create a PurgeMarker in private data store

manish-sethi:
This PR enables processing of purge operation in a transaction during block commit

#### Type of change
- New feature
#### Related issues
Closes #3026

### DIFF
--- a/core/ledger/kvledger/txmgmt/txmgr/lockbased_txmgr.go
+++ b/core/ledger/kvledger/txmgmt/txmgr/lockbased_txmgr.go
@@ -42,7 +42,7 @@ type LockBasedTxMgr struct {
 	ccInfoProvider      ledger.DeployedChaincodeInfoProvider
 	commitRWLock        sync.RWMutex
 	oldBlockCommit      sync.Mutex
-	current             *current
+	currentUpdates      *currentUpdates
 	hashFunc            rwsetutil.HashFunc
 }
 
@@ -53,18 +53,24 @@ type pvtdataPurgeMgr struct {
 	usedOnce bool
 }
 
-type current struct {
+type currentUpdates struct {
 	block     *common.Block
 	batch     *privacyenabledstate.UpdateBatch
 	listeners []ledger.StateListener
 }
 
-func (c *current) blockNum() uint64 {
+func (c *currentUpdates) blockNum() uint64 {
 	return c.block.Header.Number
 }
 
-func (c *current) maxTxNumber() uint64 {
+func (c *currentUpdates) maxTxNumber() uint64 {
 	return uint64(len(c.block.Data.Data)) - 1
+}
+
+func (c *currentUpdates) purgeAppInitiatedPvtKeys(keys map[privacyenabledstate.PvtdataCompositeKey]*version.Height) {
+	for ck, ver := range keys {
+		c.batch.PvtUpdates.Delete(ck.Namespace, ck.CollectionName, ck.Key, ver)
+	}
 }
 
 // Initializer captures the dependencies for tx manager
@@ -154,7 +160,7 @@ func (txmgr *LockBasedTxMgr) NewTxSimulator(txid string) (ledger.TxSimulator, er
 
 // ValidateAndPrepare implements method in interface `txmgmt.TxMgr`
 func (txmgr *LockBasedTxMgr) ValidateAndPrepare(blockAndPvtdata *ledger.BlockAndPvtData, doMVCCValidation bool) (
-	[]*validation.TxStatInfo, []byte, error,
+	[]*validation.AppInitiatedPurgeUpdate, []*validation.TxStatInfo, []byte, error,
 ) {
 	// Among ValidateAndPrepare(), PrepareForExpiringKeys(), and
 	// RemoveStaleAndCommitPvtDataOfOldBlocks(), we can allow only one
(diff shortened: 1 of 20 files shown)


core/ledger/kvledger/txmgmt/validation/validator.go
```diff
@@ -286,3 +291,58 @@ func (v *validator) validateKVReadHash(ns, coll string, kvReadHash *kvrwset.KVRe
...
+					}
+					continue
+				}
+
+				existingUpdate, ok := p.m[ck]
+				if !ok {
+					continue
+				}
+				// the hash of a key got overwritten by another transaction
+				existingUpdate.DeletePrivateKeyFromState = false
```
denyeart: This means it was purged by a prior transaction but re-written in a later transaction? Could you clarify the comment?

core/ledger/kvledger/txmgmt/txmgr/lockbased_txmgr.go
```diff
@@ -483,6 +489,12 @@ func (txmgr *LockBasedTxMgr) Shutdown() {
 	txmgr.db.Close()
 }
 
+// UpdateBatchWithAppInitiatedPvtKeysToPurge adds delete markers in the state update batch for the private data keys that
+// should be deleted from the state as a result of app initiated purge operations.
+func (txmgr *LockBasedTxMgr) UpdateBatchWithAppInitiatedPvtKeysToPurge(keys map[privacyenabledstate.PvtdataCompositeKey]*version.Height) {
+	txmgr.currentUpdates.purgeAppInitiatedPvtKeys(keys)
```
denyeart, line 495: The RFC says that purged private keys will be deleted from private state in the background. But it looks like you are deleting from private state atomically with the commit. Did you change your mind or have I got it wrong?

core/ledger/pvtdatastorage/helper.go
```diff
@@ -181,6 +189,41 @@ func prepareHashedIndexEntries(dataEntires []*dataEntry) ([]*hashedIndexEntry, e
...
+			&purgeMarkerEntry{
+				key: &purgeMarkerKey{
+					ns:         m.Ns,
+					coll:       m.Coll,
+					pvtkeyHash: m.PvtkeyHash,
+				},
+				value: &purgeMarkerVal{
+					blkNum: blkNum,
+					txNum:  m.TxNum,
+				},
```
denyeart, line 207: The RFC says the blkNum and txNum will be part of the PurgeMarker key. I agree it should be in the 'value' not in the 'key'. Could you update the RFC accordingly?

core/ledger/pvtdatastorage/helper.go
```diff
@@ -181,6 +189,41 @@ func prepareHashedIndexEntries(dataEntires []*dataEntry) ([]*hashedIndexEntry, e
...
+			&purgeMarkerCollEntry{
+				key: &purgeMarkerCollKey{
+					ns:   m.Ns,
+					coll: m.Coll,
+				},
+				value: &purgeMarkerVal{
+					blkNum: blkNum,
+					txNum:  m.TxNum,
+				},
+			},
```
denyeart, line 221: The RFC mentions the regular PurgeMarker, but not the PurgeMarker collection entry. Could you remind me what the PurgeMarker collection entry will be used for, and update the RFC accordingly?

There could be many purged keys in the same collection, resulting in many duplicates of the PurgeMarker collection entry. I guess when the leveldb batch is updated each occurrence will overwrite the same record and therefore only one record will eventually be written per collection. Is this the design you intended here?

core/ledger/kvledger/txmgmt/validation/validator.go
```diff
@@ -78,38 +78,43 @@ func (v *validator) preLoadCommittedVersionOfRSet(blk *block) error {
...
 		if validationCode == peer.TxValidationCode_VALID {
 			logger.Debugf("Block [%d] Transaction index [%d] TxId [%s] marked as valid by state validator. ContainsPostOrderWrites [%t]", blk.num, tx.indexInBlock, tx.id, tx.containsPostOrderWrites)
+
 			committingTxHeight := version.NewHeight(blk.num, uint64(tx.indexInBlock))
 			if err := updates.applyWriteSet(tx.rwset, committingTxHeight, v.db, tx.containsPostOrderWrites); err != nil {
-				return nil, err
+				return nil, nil, err
 			}
+
+			purgeTracker.update(tx.rwset, committingTxHeight)
```
denyeart, line 111: applyWriteSet() already iterates through all the write sets in the prior line.
purgeTracker.update() will iterate through all the write sets again.
It seems inefficient to iterate through write sets twice. But you probably did it this way to keep the code cleaner?

core/ledger/kvledger/txmgmt/txmgr/lockbased_txmgr.go
manish-sethi: Yes, your observation is correct. During the implementation, I felt that processing this in the background would have needed some changes that would make it a bit more complex. So, I picked the simplicity with the assumption that perhaps the purge is not going to be a very frequent transaction. We can always change it if we find it otherwise.

core/ledger/kvledger/txmgmt/validation/validator.go
manish-sethi: Yes, that is correct. Also, we have seen that most of the transactions contain a handful of writes (typically one or two) so in practice, I will be surprised if we notice any measurable difference.

core/ledger/kvledger/txmgmt/validation/validator.go
manish-sethi: Yes, that's correct. I'll expand the comment if you did not find it clear enough.

core/ledger/pvtdatastorage/helper.go
manish-sethi: Sorry for the lengthy reply below but since its more of a practice question so writing my opinion in detail.

Yes, there has been some changes during the implementation, as it would always be the case. I am not sure whether we should go back to RFC each time for these kind of changes or rather better have a separate design doc that covers the details of a component in full. IMO, RFCs are mostly feature specific and cuts across components, and are useful as a initial review tools. In practice, a more usable artifact would be component design doc (that too, perhaps detailed code comments). For instance, if someone is looking at some code and wants to understand the overall design, it will be hard for him to know even which RFC to go for.

Having said that, I am not against updating the RFC. I'll do that at the end. Also, will update such that we don't have to treat the RFC as a detailed description of code. For instance, relating back to earlier comment, whether to delete in the background v/s during commit, we should be able to change it without going back to RFC.

core/ledger/pvtdatastorage/helper.go
manish-sethi: This change is not introduced in this PR. The PurgeMarker collection entry was introduced in the PR that filters and deletes in the background. This is used [here](https://github.com/hyperledger/fabric/blob/811a36d8ab82e8a37640cbb521f1db641e5ac321/core/ledger/pvtdatastorage/store.go#L553) to reduce the overheads for the existing deployments until they start using this feature.

Yes, there would be a single entry per collection that we keep overwriting.

denyeart (approved):
Reviewed code and also tested end-to-end. Only issue I see has been opened as https://github.com/hyperledger/fabric/issues/3761, I'll go ahead and merge since that issue is not related to this PR.

denyeart:

PR #3706, opened 2022-10-20
Release Workflow

jcastrence:
#### Type of change

- New feature

#### Description

- Release workflow automates new release creation process similar the current Azure Pipelines release job
- Generates binaries for linux, darwin, and windows targets, pushes Docker images to Hyperledger DockerHub, and creates a new tag and release with artifacts for given version number based off of main branch latest commit
- Workflow must be manually triggered from Actions tab and requires input for the version number and two digit version number similar to Azure Pipelines
- GitHub secrets for DOCKERHUB_PASSWORD and DOCKERHUB_USERNAME must be added to repository matching Hyperledger DockerHub credentials before workflow can be run successfully

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,0 +1,98 @@
+# Copyright the Hyperledger Fabric contributors. All rights reserved.
+#
+# SPDX-License-Identifier: Apache-2.0
+
+name: Release
+on:
+  workflow_dispatch:
+    inputs:
+      release:
+        description: 'Fabric Release'
+        required: true
+        type: string
+      two_digit_release:
+        description: 'Fabric Two Digit Release'
+        required: true
+        type: string
+
+env:
+  GO_VER: 1.18.7
+
+permissions:
+  contents: read
+
+jobs:
+  build-binaries:
+    name: Build Fabric Binaries
+    strategy:
+      matrix:
+        include:
+        - image: ubuntu-22.04
+          target: linux
+          arch: amd64
+        - image: macos-11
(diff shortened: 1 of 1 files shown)


jcastrence:
[Successful workflow run](https://github.com/jcastrence/fabric/actions/runs/3291875607)
[Successful release creation](https://github.com/jcastrence/fabric/releases)

.github/workflows/release.yml
```diff
@@ -0,0 +1,98 @@
...
+        description: 'Fabric Release'
+        required: true
+        type: string
+      two_digit_release:
+        description: 'Fabric Two Digit Release'
+        required: true
+        type: string
+
+env:
+  GO_VER: 1.18
```
denyeart: For the actual release process, I think we'll want to continue to prescribe exactly which Go version is used, for now that would be 1.18.7.
This way, we can look at the release tag and always know what version of Go was actually used for the release artifacts. Also, a later Go release could potentially break us, rare, but possible.

For the [verify-build.yml](https://github.com/hyperledger/fabric/blob/main/.github/workflows/verify-build.yml) workflow, I think it is ok to use "1.18" so that CI always runs with the latest patch release (currently 1.18.7). This way, we'll get some notice in CI if a later Go patch release breaks us.

Please update the fabric-ca release job as well, I didn't catch this over there.

denyeart (approved):

denyeart:

denyeart:

denyeart:

denyeart:

PR #3765, opened 2022-11-03
Add private write-set for purge operation

### LINKED ISSUE #3761: Purge private data - inconsistent transient store behavior
In Fabric main development branch, when private data is purged during transaction simulation, the purged key is not written to transient store (unlike traditional 'delete' private data where key is written to transient store).

However, when the peer attempts to commit a block with purged private data, it will attempt to fetch the purged key from the transient store (or another peer's transient store). Since the purged key is not available in any peer's transient store, the commit will stall and keep re-trying the fetch until CORE_PEER_GOSSIP_PVTDATA_PULLRETRYTHRESHOLD times out, by default 60 seconds. At that point the peer will commit the block with missing private data, and then enter the missing private data reconciliation loop.

We need to either write the purged key to transient store at endorsement time, or not look for the purged key at commit time.

manish-sethi:
This commit adds a private write-set for a purge operation. The private write-set contains the delete operation on the key getting purged.

#### Type of change
- New feature

#### Related issues
Closes #3761

### DIFF
--- a/core/ledger/kvledger/txmgmt/rwsetutil/rwset_builder.go
+++ b/core/ledger/kvledger/txmgmt/rwsetutil/rwset_builder.go
@@ -116,10 +116,12 @@ func (b *RWSetBuilder) AddToPvtAndHashedWriteSet(ns string, coll string, key str
 	b.getOrCreateCollHashedRwBuilder(ns, coll).writeMap[key] = kvWriteHash
 }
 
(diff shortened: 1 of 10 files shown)


denyeart (approved):

denyeart:
I've confirmed that this PR resolves the issue in an actual trial, looks good!

denyeart:

PR #3768, opened 2022-11-08
Update write_first_app.rst

tekintian:
should run  npm run build  to build the dist/app.js, then run npm start

Signed-off-by: Tekin Tian <tekintian@gmail.com>

missing the npm run build

### DIFF
--- a/docs/source/write_first_app.rst
+++ b/docs/source/write_first_app.rst
@@ -179,7 +179,9 @@ Let's run the application and then step through each of the interactions with th
 ``asset-transfer-basic/application-gateway-typescript`` directory, run the following command:
 
 .. code-block:: bash
-
+  
+  npm run build
+  
   npm start
 
 


docs/source/write_first_app.rst
```diff
@@ -179,7 +179,9 @@ Let's run the application and then step through each of the interactions with th
 ``asset-transfer-basic/application-gateway-typescript`` directory, run the following command:
```
andrew-coleman, line 179:
```suggestion
``asset-transfer-basic/application-gateway-typescript`` directory, run the following:
```

andrew-coleman: Looks good, thanks.  Can you just make this suggested change to the wording since it is now referring to more than one command?

andrew-coleman (changes requested):

docs/source/write_first_app.rst
```diff
@@ -179,7 +179,9 @@ Let's run the application and then step through each of the interactions with th
...
 
 .. code-block:: bash
-
+  
+  npm run build
```
bestbeforetoday, line 183: The `npm install` command in the previous step already runs the TypeScript build. There is no need to explicitly run the build again, unless you are modifying the application code, in which case you can simply re-run `npm install`

bestbeforetoday (changes requested):
If the intent is to make the tutorial flow work, this change is unnecessary. If the intent is to make it easier for people to modify and re-run the sample, the instructions should probably be updated to describe the best way to do that.

denyeart:
No response from contributor, closing.

PR #3774, opened 2022-11-09
Fix cert sanitation after moving to go v1.19

### LINKED ISSUE #3661: Update to Go 1.19
Update to Go 1.19.

Note that Go 1.19 has x509 verification changes that will require MSP updates.

C0rWin:
Resolves #3661 


yacovm:
@adecaro @ale-linux can you guys also take a look?

adecaro:
@C0rWin , please, can you explain the point of this PR?

C0rWin:
> @C0rWin , please, can you explain the point of this PR?

@adecaro the main point is to make Fabric work with go 1.19 :)

I guess you are asking for more details. Let me try to explain. First, of it started from the recent changes in std of the go 1.19, where they did optimization on the way your check certificate. Bottom line if previously, there was a bitwise comparison, and after cert sanitation, the code to build the certs chain worked, now, after we changed to LowS the code no longer works as we expected. Below you can see a more detailed explanation.

**TLDR;**

After moving to the go v1.19, validating certificates signed with ECDSA signature after sanitation to LowS form is no longer possible.  The code which manifests the problem is the following:

```
certpool := x509.NewCertPool()

caFileBytes, err := os.ReadFile(rootCAFile)
if err != nil {
   panic(err)
}

caPemBlock, _ := pem.Decode(caFileBytes)
caCert, err := x509.ParseCertificate(caPemBlock.Bytes)
if err != nil {
   panic(err)
}

if isECDSASignedCert(caCert) {
   caCert, err = utils.SanitizeCert(caCert)
   if err != nil {
      panic(err)
   }
}

certpool.AddCert(caCert)
opts := x509.VerifyOptions{
   Roots: certpool,
}

certBytes, err := os.ReadFile(certFile)
if err != nil {
   panic(err)
}

certPemBlock, _ := pem.Decode(certBytes)
cert, err := x509.ParseCertificate(certPemBlock.Bytes)
if err != nil {
   panic(err)
}

_, err = cert.Verify(opts)
if err != nil {
   panic(err)
}

fmt.Println("SUCCESS")
```

Having following certificates 

```
rootCAFile := `-----BEGIN CERTIFICATE-----
MIICPzCCAeSgAwIBAgIRAONi5v8ImyejqCrCatbAW1QwCgYIKoZIzj0EAwIwaTEL
MAkGA1UEBhMCVVMxEzARBgNVBAgTCkNhbGlmb3JuaWExFjAUBgNVBAcTDVNhbiBG
cmFuY2lzY28xFDASBgNVBAoTC2V4YW1wbGUuY29tMRcwFQYDVQQDEw5jYS5leGFt
cGxlLmNvbTAeFw0xOTA3MDQxNjI3MDBaFw0yOTA3MDExNjI3MDBaMGkxCzAJBgNV
BAYTAlVTMRMwEQYDVQQIEwpDYWxpZm9ybmlhMRYwFAYDVQQHEw1TYW4gRnJhbmNp
c2NvMRQwEgYDVQQKEwtleGFtcGxlLmNvbTEXMBUGA1UEAxMOY2EuZXhhbXBsZS5j
b20wWTATBgcqhkjOPQIBBggqhkjOPQMBBwNCAAQAfjOlLCdB/6SsdPlbDHUsdK+b
gRuEN38QOFZ0Ws3aFAsER8ImqV3UIlsbKi5JnDs+OQnzrr3hrKA8downRRy/o20w
azAOBgNVHQ8BAf8EBAMCAaYwHQYDVR0lBBYwFAYIKwYBBQUHAwIGCCsGAQUFBwMB
MA8GA1UdEwEB/wQFMAMBAf8wKQYDVR0OBCIEIN/hgC0LnghmpjfwtsqvEvhE/2LI
wJbaCUtoY5cFF8WfMAoGCCqGSM49BAMCA0kAMEYCIQDhhgAHx0l7V5uAG2hATgCs
bvsbHiJpHUtiK7f1Qfxf2AIhANeukSgRU+AeGSzyVmAOKhIUS+grsPyspksUwVvB
ehXv
-----END CERTIFICATE-----`
```

and 

```
certFile := `-----BEGIN CERTIFICATE-----
MIICPzCCAeSgAwIBAgIRAONi5v8ImyejqCrCatbAW1QwCgYIKoZIzj0EAwIwaTEL
MAkGA1UEBhMCVVMxEzARBgNVBAgTCkNhbGlmb3JuaWExFjAUBgNVBAcTDVNhbiBG
cmFuY2lzY28xFDASBgNVBAoTC2V4YW1wbGUuY29tMRcwFQYDVQQDEw5jYS5leGFt
cGxlLmNvbTAeFw0xOTA3MDQxNjI3MDBaFw0yOTA3MDExNjI3MDBaMGkxCzAJBgNV
BAYTAlVTMRMwEQYDVQQIEwpDYWxpZm9ybmlhMRYwFAYDVQQHEw1TYW4gRnJhbmNp
c2NvMRQwEgYDVQQKEwtleGFtcGxlLmNvbTEXMBUGA1UEAxMOY2EuZXhhbXBsZS5j
b20wWTATBgcqhkjOPQIBBggqhkjOPQMBBwNCAAQAfjOlLCdB/6SsdPlbDHUsdK+b
gRuEN38QOFZ0Ws3aFAsER8ImqV3UIlsbKi5JnDs+OQnzrr3hrKA8downRRy/o20w
azAOBgNVHQ8BAf8EBAMCAaYwHQYDVR0lBBYwFAYIKwYBBQUHAwIGCCsGAQUFBwMB
MA8GA1UdEwEB/wQFMAMBAf8wKQYDVR0OBCIEIN/hgC0LnghmpjfwtsqvEvhE/2LI
wJbaCUtoY5cFF8WfMAoGCCqGSM49BAMCA0kAMEYCIQDhhgAHx0l7V5uAG2hATgCs
bvsbHiJpHUtiK7f1Qfxf2AIhANeukSgRU+AeGSzyVmAOKhIUS+grsPyspksUwVvB
ehXv
-----END CERTIFICATE-----`
```

The code mentioned above running with go 1.18 will result in `SUCCESS`; however, it will fail on go 1.19 with panic, because it will fail to build a verification path. The code implementing sanitation is following:

```
type validity struct {
   NotBefore, NotAfter time.Time
}

type publicKeyInfo struct {
   Raw       asn1.RawContent
   Algorithm pkix.AlgorithmIdentifier
   PublicKey asn1.BitString
}

type certificate struct {
   Raw                asn1.RawContent
   TBSCertificate     tbsCertificate
   SignatureAlgorithm pkix.AlgorithmIdentifier
   SignatureValue     asn1.BitString
}

type tbsCertificate struct {
   Raw                asn1.RawContent
   Version            int `asn1:"optional,explicit,default:0,tag:0"`
   SerialNumber       *big.Int
   SignatureAlgorithm pkix.AlgorithmIdentifier
   Issuer             asn1.RawValue
   Validity           validity
   Subject            asn1.RawValue
   PublicKey          publicKeyInfo
   UniqueId           asn1.BitString   `asn1:"optional,tag:1"`
   SubjectUniqueId    asn1.BitString   `asn1:"optional,tag:2"`
   Extensions         []pkix.Extension `asn1:"optional,explicit,tag:3"`
}

type ECDSASignature struct {
   R, S *big.Int
}

// curveHalfOrders contains the precomputed curve group orders halved.
// It is used to ensure that signature' S value is lower or equal to the
// curve group order halved. We accept only low-S signatures.
// They are precomputed for efficiency reasons.
var curveHalfOrders = map[elliptic.Curve]*big.Int{
   elliptic.P224(): new(big.Int).Rsh(elliptic.P224().Params().N, 1),
   elliptic.P256(): new(big.Int).Rsh(elliptic.P256().Params().N, 1),
   elliptic.P384(): new(big.Int).Rsh(elliptic.P384().Params().N, 1),
   elliptic.P521(): new(big.Int).Rsh(elliptic.P521().Params().N, 1),
}

func GetCurveHalfOrdersAt(c elliptic.Curve) *big.Int {
   return big.NewInt(0).Set(curveHalfOrders[c])
}

func MarshalECDSASignature(r, s *big.Int) ([]byte, error) {
   return asn1.Marshal(ECDSASignature{r, s})
}

func UnmarshalECDSASignature(raw []byte) (*big.Int, *big.Int, error) {
   // Unmarshal
   sig := new(ECDSASignature)
   _, err := asn1.Unmarshal(raw, sig)
   if err != nil {
      return nil, nil, fmt.Errorf("failed unmashalling signature [%s]", err)
   }

   // Validate sig
   if sig.R == nil {
      return nil, nil, errors.New("invalid signature, R must be different from nil")
   }
   if sig.S == nil {
      return nil, nil, errors.New("invalid signature, S must be different from nil")
   }

   if sig.R.Sign() != 1 {
      return nil, nil, errors.New("invalid signature, R must be larger than zero")
   }
   if sig.S.Sign() != 1 {
      return nil, nil, errors.New("invalid signature, S must be larger than zero")
   }

   return sig.R, sig.S, nil
}

func SignatureToLowS(k *ecdsa.PublicKey, signature []byte) ([]byte, error) {
   r, s, err := UnmarshalECDSASignature(signature)
   if err != nil {
      return nil, err
   }

   s, err = ToLowS(k, s)
   if err != nil {
      return nil, err
   }

   return MarshalECDSASignature(r, s)
}

// IsLow checks that s is a low-S
func IsLowS(k *ecdsa.PublicKey, s *big.Int) (bool, error) {
   halfOrder, ok := curveHalfOrders[k.Curve]
   if !ok {
      return false, fmt.Errorf("curve not recognized [%s]", k.Curve)
   }

   return s.Cmp(halfOrder) != 1, nil
}

func ToLowS(k *ecdsa.PublicKey, s *big.Int) (*big.Int, error) {
   lowS, err := IsLowS(k, s)
   if err != nil {
      return nil, err
   }

   if !lowS {
      // Set s to N - s that will be then in the lower part of signature space
      // less or equal to half order
      s.Sub(k.Params().N, s)

      return s, nil
   }

   return s, nil
}

func SanitizeCert(cert *x509.Certificate) (*x509.Certificate, error) {
   expectedSig, err := SignatureToLowS(cert.PublicKey.(*ecdsa.PublicKey), cert.Signature)
   if err != nil {
      return nil, err
   }

   if bytes.Equal(cert.Signature, expectedSig) {
      return cert, nil
   }

   var newCert certificate
   _, err = asn1.Unmarshal(cert.Raw, &newCert)
   if err != nil {
      panic(err)
   }

   newCert.SignatureValue = asn1.BitString{Bytes: expectedSig, BitLength: 8 * len(expectedSig)}
   newCert.Raw = nil

   newRaw, err := asn1.Marshal(newCert)
   if err != nil {
      return nil, err
   }

   return x509.ParseCertificate(newRaw)
}
```

Now, in go 1.18, the certificate verification compares certs bitwise (`crypto/x509/verify.go:829`):

```
considerCandidate := func(certType int, candidate *Certificate) {
	for _, cert := range currentChain {
		if cert.Equal(candidate) {
			return
		}
	}

	if sigChecks == nil {
		sigChecks = new(int)
	}
	*sigChecks++
	if *sigChecks > maxChainSignatureChecks {
		err = errors.New("x509: signature check attempts limit reached while verifying certificate chain")
		return
	}

	if err := c.CheckSignatureFrom(candidate); err != nil {
		if hintErr == nil {
			hintErr = err
			hintCert = candidate
		}
		return
	}

	err = candidate.isValid(certType, currentChain, opts)
	if err != nil {
		return
	}

	switch certType {
	case rootCertificate:
		chains = append(chains, appendToFreshChain(currentChain, candidate))
	case intermediateCertificate:
		if cache == nil {
			cache = make(map[*Certificate][][]*Certificate)
		}
		childChains, ok := cache[candidate]
		if !ok {
			childChains, err = candidate.buildChains(cache, appendToFreshChain(currentChain, candidate), sigChecks, opts)
			cache[candidate] = childChains
		}
		chains = append(chains, childChains...)
	}
}
```

The code above has been refactored and optimized to:

```
considerCandidate := func(certType int, candidate *Certificate) {
	if alreadyInChain(candidate, currentChain) {
		return
	}

	if sigChecks == nil {
		sigChecks = new(int)
	}
	*sigChecks++
	if *sigChecks > maxChainSignatureChecks {
		err = errors.New("x509: signature check attempts limit reached while verifying certificate chain")
		return
	}

	if err := c.CheckSignatureFrom(candidate); err != nil {
		if hintErr == nil {
			hintErr = err
			hintCert = candidate
		}
		return
	}

	err = candidate.isValid(certType, currentChain, opts)
	if err != nil {
		return
	}

	switch certType {
	case rootCertificate:
		chains = append(chains, appendToFreshChain(currentChain, candidate))
	case intermediateCertificate:
		var childChains [][]*Certificate
		childChains, err = candidate.buildChains(appendToFreshChain(currentChain, candidate), sigChecks, opts)
		chains = append(chains, childChains...)
	}
}
```

the `alreadyInChain` replaces the loop, which compares certs by checking Subject, PublicKey, and SAN.

```
func alreadyInChain(candidate *Certificate, chain []*Certificate) bool {
	type pubKeyEqual interface {
		Equal(crypto.PublicKey) bool
	}

	var candidateSAN *pkix.Extension
	for _, ext := range candidate.Extensions {
		if ext.Id.Equal(oidExtensionSubjectAltName) {
			candidateSAN = &ext
			break
		}
	}

	for _, cert := range chain {
		if !bytes.Equal(candidate.RawSubject, cert.RawSubject) {
			continue
		}
		if !candidate.PublicKey.(pubKeyEqual).Equal(cert.PublicKey) {
			continue
		}
		var certSAN *pkix.Extension
		for _, ext := range cert.Extensions {
			if ext.Id.Equal(oidExtensionSubjectAltName) {
				certSAN = &ext
				break
			}
		}
		if candidateSAN == nil && certSAN == nil {
			return true
		} else if candidateSAN == nil || certSAN == nil {
			return false
		}
		if bytes.Equal(candidateSAN.Value, certSAN.Value) {
			return true
		}
	}
	return false
}
```


Compares Subject, PublicKey, and SAN is equal cause we verify sanitized certs. However, the signature is different, and the function call in the example above should have resulted in false. Therefore, my question is whether the `alreadyInChain` should be extended with an additional check for signature equality? 



The code below demonstrates the issue with the test - https://go.dev/play/p/Vaephj6DLOl.

msp/mspimpl.go
```diff
@@ -890,9 +890,15 @@ func (msp *bccspmsp) getCertificationChainIdentifierFromChain(chain []*x509.Cert
 // is regenerated to have a Low-S signature.
 func (msp *bccspmsp) sanitizeCert(cert *x509.Certificate) (*x509.Certificate, error) {
 	if isECDSASignedCert(cert) {
+		validityOpts := msp.getValidityOptsForCert(cert)
+		if bytes.Equal(cert.RawSubject, cert.RawIssuer) && cert.IsCA {
+			validityOpts.Roots = nil
```
yacovm: why do you set these `validityOpts` field to `nil` if you return right after in line 897?

msp/mspimpl.go
```diff
@@ -890,9 +890,15 @@ func (msp *bccspmsp) getCertificationChainIdentifierFromChain(chain []*x509.Cert
 // is regenerated to have a Low-S signature.
 func (msp *bccspmsp) sanitizeCert(cert *x509.Certificate) (*x509.Certificate, error) {
 	if isECDSASignedCert(cert) {
+		validityOpts := msp.getValidityOptsForCert(cert)
+		if bytes.Equal(cert.RawSubject, cert.RawIssuer) && cert.IsCA {
```
yacovm: this only affects self signed root CAs, no? What about intermediate certs that have parents that are intermediate?

C0rWin: Yes, there is no problem with intermediates

msp/mspimpl.go
C0rWin: It's a leftover, I will remove it.

adecaro:
@C0rWin , can you remove the unrelated changes, please? They are puzzling me.

Apart that, will this change generate any fork? Will a certificate be valid without these changes and not valid with the changes, and vice versa?

C0rWin:
> @C0rWin , can you remove the unrelated changes, please? They are puzzling me.

@adecaro can you please point to the changes which are puzzling you?

> Apart that, will this change generate any fork? Will a certificate be valid without these changes and not valid with the changes, and vice versa?

Without this change, you simply won't be able to run Fabric on go v1.9+ because of the sanitation you have for certificates and the way it is now implemented since go v1.9. What happens now, in the code you have

```
		chain, err := msp.getUniqueValidationChain(cert, validityOpts)
```

which in turn calls

```
	validationChains, err := cert.Verify(opts)
```

which goes to the 

```
     cert.buildChains(....)
```

and `buildChains` impl looks as following:

```
func (c *Certificate) buildChains(currentChain []*Certificate, sigChecks *int, opts *VerifyOptions) (chains [][]*Certificate, err error) {
	var (
		hintErr  error
		hintCert *Certificate
	)

	considerCandidate := func(certType int, candidate *Certificate) {
		if alreadyInChain(candidate, currentChain) {
			return
		}

		if sigChecks == nil {
			sigChecks = new(int)
		}
		*sigChecks++
		if *sigChecks > maxChainSignatureChecks {
			err = errors.New("x509: signature check attempts limit reached while verifying certificate chain")
			return
		}

		if err := c.CheckSignatureFrom(candidate); err != nil {
			if hintErr == nil {
				hintErr = err
				hintCert = candidate
			}
			return
		}

		err = candidate.isValid(certType, currentChain, opts)
		if err != nil {
			return
		}

		switch certType {
		case rootCertificate:
			chains = append(chains, appendToFreshChain(currentChain, candidate))
		case intermediateCertificate:
			var childChains [][]*Certificate
			childChains, err = candidate.buildChains(appendToFreshChain(currentChain, candidate), sigChecks, opts)
			chains = append(chains, childChains...)
		}
	}

	for _, root := range opts.Roots.findPotentialParents(c) {
		considerCandidate(rootCertificate, root)
	}
	for _, intermediate := range opts.Intermediates.findPotentialParents(c) {
		considerCandidate(intermediateCertificate, intermediate)
	}

	if len(chains) > 0 {
		err = nil
	}
	if len(chains) == 0 && err == nil {
		err = UnknownAuthorityError{c, hintErr, hintCert}
	}

	return
}
```

the whole point in the change introduced at golong v1.9 was in the function

```
		if alreadyInChain(candidate, currentChain) {
			return
		}

```

where

```
// alreadyInChain checks whether a candidate certificate is present in a chain.
// Rather than doing a direct byte for byte equivalency check, we check if the
// subject, public key, and SAN, if present, are equal. This prevents loops that
// are created by mutual cross-signatures, or other cross-signature bridge
// oddities.
func alreadyInChain(candidate *Certificate, chain []*Certificate) bool {
	type pubKeyEqual interface {
		Equal(crypto.PublicKey) bool
	}

	var candidateSAN *pkix.Extension
	for _, ext := range candidate.Extensions {
		if ext.Id.Equal(oidExtensionSubjectAltName) {
			candidateSAN = &ext
			break
		}
	}

	for _, cert := range chain {
		if !bytes.Equal(candidate.RawSubject, cert.RawSubject) {
			continue
		}
		if !candidate.PublicKey.(pubKeyEqual).Equal(cert.PublicKey) {
			continue
		}
		var certSAN *pkix.Extension
		for _, ext := range cert.Extensions {
			if ext.Id.Equal(oidExtensionSubjectAltName) {
				certSAN = &ext
				break
			}
		}
		if candidateSAN == nil && certSAN == nil {
			return true
		} else if candidateSAN == nil || certSAN == nil {
			return false
		}
		if bytes.Equal(candidateSAN.Value, certSAN.Value) {
			return true
		}
	}
	return false
}
```


So, before it was bitwise comparison and already in the chain returned false, cause the key was sanitized, now it checks only 

```
subject, public key, and SAN
```

Hence after sanitation, it returns true, thinking the certificate was included and therefore doesn't append it to the chain, eventually leading to the following code:

```
	if len(chains) == 0 && err == nil {
		err = UnknownAuthorityError{c, hintErr, hintCert}
	}
``` 

Hence the error you will encounter switching to go above v1.9, hopefully, now it makes it clear @adecaro .

denyeart:
Closed and re-opened to reset checks.

denyeart:
@C0rWin Could you rebase to pick up your prior goimports change, and then take this PR out of Draft so that we can do final reviews.

Also, I think Angelo's question was whether this change risks introducing a fork between old peers on Go 1.18 and updated peers on Go 1.19, due to differences in certificate handling. So I understand changes are required to be able run with Go 1.19, the question remains is the new logic functionally equivalent with the old to ensure no forks?

C0rWin:
> Also, I think Angelo's question was whether this change risks introducing a fork between old peers on Go 1.18 and updated peers on Go 1.19, due to differences in certificate handling. So I understand changes are required to be able run with Go 1.19, the question remains is the new logic functionally equivalent with the old to ensure no forks?

@denyeart this commit actually makes sure certificate validation will end up with similar results for go 1.18 and go 1.19, note I made sure it passes integration and unit-tests.

denyeart:
This change looks good to me now, but would like @adecaro or @ale-linux to do a final review/approval before we merge.

msp/mspimpl.go
```diff
@@ -890,9 +890,13 @@ func (msp *bccspmsp) getCertificationChainIdentifierFromChain(chain []*x509.Cert
...
```
ale-linux: A couple of points:

1. why isn't the `cert.IsCa` check sufficient?
2. could we please add some comments here? Linking the issue with the explanation
3. can you initialise `parentCert` to `cert`, negate the condition of the if so that we end up calling `sanitizeECDSASignedCert` only once?
4. iiuc, `if cert.IsCA && len(chain) == 1 {` will never be hit again as the code is structured today - is that correct? If so I'd remove this

msp/mspimplsetup.go
```diff
@@ -562,6 +562,7 @@ func (msp *bccspmsp) preSetupV1(conf *m.FabricMSPConfig) error {
 
 func (msp *bccspmsp) preSetupV142(conf *m.FabricMSPConfig) error {
 	// setup crypto config
+
```
ale-linux: nit - sorry.. :) pls remove this newline

ale-linux (changes requested):
Thx - and sorry for looking at it just now. I left a couple of Qs and a nit

ale-linux:
...and on second thought I'm not sure I'm still fully convinced that the code change won't fork the chain. In the old code we used to always call `getUniqueValidationChain`, which has a complex error model, even for root certs. Now what we're doing is just skipping `getUniqueValidationChain` (and it's complex error model) for CA certs with issuer identical to subject. What if this function is ever fed with a certificate such that

- the `bytes.Equal(cert.RawSubject, cert.RawIssuer) && cert.IsCA` passes, but
- `getUniqueValidationChain` would fail

If such a certificate exists, the old code would return an error, the new code would succeed - thus forking the chain.

Wdyt? (@yacovm tagging you here because you're always good at spotting these things)

denyeart:
> ...and on second thought I'm not sure I'm still fully convinced that the code change won't fork the chain. In the old code we used to always call `getUniqueValidationChain`, which has a complex error model, even for root certs. Now what we're doing is just skipping `getUniqueValidationChain` (and it's complex error model) for CA certs with issuer identical to subject. What if this function is ever fed with a certificate such that
> 
> * the `bytes.Equal(cert.RawSubject, cert.RawIssuer) && cert.IsCA` passes, but
> * `getUniqueValidationChain` would fail
> 
> If such a certificate exists, the old code would return an error, the new code would succeed - thus forking the chain.
> 
> Wdyt? (@yacovm tagging you here because you're always good at spotting these things)

I share the concern... getUniqueValidationChain() is what calls cert.Verify(opts), so how can we just skip the getUniqueValidationChain() call?

msp/mspimpl.go
C0rWin: > A couple of points:
> 
>     1. why isn't the `cert.IsCa` check sufficient?

There is is actually described in my previous answer to @adecaro, in short I need to identify the attempt to sanitize root CA certificates only, cause after the changes introduced in v1.19 they moved away from bitwise comparison. Therefore it fails to build the chain for root CA, but there is no need to build the chain by calling cert.Verify in that particular case, since it's self signed cert. But you are right IMO the better way would be to change it to the `cert.CheckSignatureFrom(cert) == nil`. I will fix and push the code a little bit later, would like to check it first.

> 
>     2. could we please add some comments here? Linking the issue with the explanation

Good point, will do it.

> 
>     3. can you initialise `parentCert` to `cert`, negate the condition of the if so that we end up calling `sanitizeECDSASignedCert` only once?

Will do.

> 
>     4. iiuc, `if cert.IsCA && len(chain) == 1 {` will never be hit again as the code is structured today - is that correct? If so I'd remove this

This still covers intermediate certificates.

C0rWin:
> ...and on second thought I'm not sure I'm still fully convinced that the code change won't fork the chain. In the old code we used to always call `getUniqueValidationChain`, which has a complex error model, even for root certs. Now what we're doing is just skipping `getUniqueValidationChain` (and it's complex error model) for CA certs with issuer identical to subject. What if this function is ever fed with a certificate such that
> 
>     * the `bytes.Equal(cert.RawSubject, cert.RawIssuer) && cert.IsCA` passes, but
> 
>     * `getUniqueValidationChain` would fail
> 
> 
> If such a certificate exists, the old code would return an error, the new code would succeed - thus forking the chain.
> 
> Wdyt? (@yacovm tagging you here because you're always good at spotting these things)

Well, @ale-linux, the check for `bytes.Equal(cert.RawSubject, cert.RawIssuer) && cert.IsCA` is to ensure we took care of the root CA certificates (I agree the check has to be replaced with `cert.CheckSignatureFrom(cert) == nil`). Now, this is to avoid calling the cert.Verify which is called from `getUniqueValidationChain` and where the optimization introduced in v1.19 makes a difference, considering two certs before and after sanitation to be equal and therefore failing to build the chain. This commit ensures that the results of `sanitizeCert` will be the same for both versions, v1.18 and v1.19.

C0rWin:
> > ...and on second thought I'm not sure I'm still fully convinced that the code change won't fork the chain. In the old code we used to always call `getUniqueValidationChain`, which has a complex error model, even for root certs. Now what we're doing is just skipping `getUniqueValidationChain` (and it's complex error model) for CA certs with issuer identical to subject. What if this function is ever fed with a certificate such that
> > 
> > * the `bytes.Equal(cert.RawSubject, cert.RawIssuer) && cert.IsCA` passes, but
> > * `getUniqueValidationChain` would fail
> > 
> > If such a certificate exists, the old code would return an error, the new code would succeed - thus forking the chain.
> > Wdyt? (@yacovm tagging you here because you're always good at spotting these things)
> 
> I share the concern... getUniqueValidationChain() is what calls cert.Verify(opts), so how can we just skip the getUniqueValidationChain() call?

@denyeart Dave, this is simple. We are skipping it only for the very specific case of root CA certificate. The reason and the explanation I have provided above, the key reason, is the assumption we used to have before v1.19 and the optimization which was introduced in v1.19, moving away from bitwise certificate comparison. Note that in the case of Root CA cert, there is no need to call for `getUniqueValidationChain` since the chain is rather simple and obvious. This PR is refactoring the code to mitigate the assumption that now the code used to work in v1.18 and ensures the results of the `sanitizeCert` function in both versions will result in the same.

C0rWin:
@ale-linux, @denyeart, can you guys please retake a look? I have taken care of your comments.

adecaro:
Hi @C0rWin, 

Let's clarify the following line: 

```
if cert.CheckSignatureFrom(cert) != nil || !cert.IsCA {
```

if the cert is a leaf (not a CA), we go into the if.
What happens if cert is an intermediate CA? A path to a root certificate exists in this case. 
`!cert.IsCA` is false and `cert.CheckSignatureFrom(cert) != nil` is false as well
isn’t this a problem? We should check the chain in this case, no?

C0rWin:
> Hi @C0rWin,
> 
> Let's clarify the following line:
> 
> ```
> if cert.CheckSignatureFrom(cert) != nil || !cert.IsCA {
> ```
> 
> if the cert is a leaf (not a CA), we go into the if. What happens if cert is an intermediate CA? A path to a root certificate exists in this case. `!cert.IsCA` is false and `cert.CheckSignatureFrom(cert) != nil` is false as well isn’t this a problem? We should check the chain in this case, no?

In case this is an intermediate CA, you will also go into the if (since it's not a self-signed certificate) and will check the chain. As I explained briefly a few comments above (a more detailed explanation is also provided), the issue is only with the root CA.

adecaro:
ahhh, `cert.CheckSignatureFrom(cert) != nil` checks if it is self-signed and therefore a root CA. Any other certificate should pass this check only with negligible probability. 
All right. Then, I think we are good here. @ale-linux , can you confirm as well?

C0rWin:
> ahhh, `cert.CheckSignatureFrom(cert) != nil` checks if it is self-signed and therefore a root CA. Any other certificate should pass this check only with negligible probability. All right. Then, I think we are good here. @ale-linux , can you confirm as well?

thanks @adecaro

C0rWin:
@denyeart @ale-linux, guys do you have any other concerns? questions?

msp/mspimpl.go
```diff
@@ -890,27 +890,40 @@ func (msp *bccspmsp) getCertificationChainIdentifierFromChain(chain []*x509.Cert
...
-		if cert.IsCA && len(chain) == 1 {
-			// cert is a root CA certificate
-			parentCert = cert
-		} else {
-			parentCert = chain[1]
+			// at this point, cert might be a root CA certificate
+			// or an intermediate CA certificate
+			if cert.IsCA && len(chain) == 1 {
+				// cert is a root CA certificate
+				parentCert = cert
```
ale-linux: sorry, I'm still a little puzzled by this: if the length of the chain is 1, it's a root CA - at which point iiuc we would not be in this block anyway. What am I missing?

(and in case I'm wrong, we should add some tests because we don't seem to test this if case

ale-linux (changes requested):
Thx @C0rWin. One more comment and a sanity check:

basically we're claiming that no self-signed root CA could ever fail a pre 1.19 invocation of `getUniqueValidationChain`. Do you confirm?

C0rWin:
> Thx @C0rWin. One more comment and a sanity check:
> 
> basically we're claiming that no self-signed root CA could ever fail a pre 1.19 invocation of `getUniqueValidationChain`. Do you confirm?

indeed, and we had a validation of it, you can comment out my change and run the test with v1.18, then run it with v1.19, next uncomment, and repeat.

ale-linux:
Well, I asked a different Q - whether *NO* self-signed root CA cert could ever fail a pre 1.19 invocation of `getUniqueValidationChain`..

(and pls also have a look at https://github.com/hyperledger/fabric/pull/3774#discussion_r1113221488 when you can, thx!)

C0rWin:
> Well, I asked a different Q - whether _NO_ self-signed root CA cert could ever fail a pre 1.19 invocation of `getUniqueValidationChain`..
> 
> (and pls also have a look at [#3774 (comment)](https://github.com/hyperledger/fabric/pull/3774#discussion_r1113221488) when you can, thx!)

Em, I am not sure whenever I understand your question, there are a lot of different cases why it might have failed:


```
  716 func (msp *bccspmsp) getUniqueValidationChain(cert *x509.Certificate, opts x509.VerifyOptions) ([]*x509.Certificate, error) {
  717     // ask golang to validate the cert for us based on the options that we've built at setup time
  718     if msp.opts == nil {
  719         return nil, errors.New("the supplied identity has no verify options")
  720     }
  721     validationChains, err := cert.Verify(opts)
  722     if err != nil {
  723         return nil, errors.WithMessage(err, "the supplied identity is not valid")
  724     }
  725
  726     // we only support a single validation chain;
  727     // if there's more than one then there might
  728     // be unclarity about who owns the identity
  729     if len(validationChains) != 1 {
  730         return nil, errors.Errorf("this MSP only supports a single validation chain, got %d", len(validationChains))
  731     }
  732
  733     // Make the additional verification checks that were done in Go 1.14.
  734     err = verifyLegacyNameConstraints(validationChains[0])
  735     if err != nil {
  736         return nil, errors.WithMessage(err, "the supplied identity is not valid")
  737     }
  738
  739     return validationChains[0], nil
  740 }
```

There are set of validation rules which might lead to failure. I'm also not clear about NO self-seigned root CA, do you mean intermediate certs?

ale-linux:
yes, exactly: `getUniqueValidationChain` has a complex set of validation rules, and we must make sure that all self-signed root CA can never lead to a failure when evaluating those rules: before this patch, they used to undergo those validation rules, and so we must make sure that after the patch, those validation rules somehow still apply - and so that any self-signed root CA that failed those checks before the patch (because we used to call `getUniqueValidationChain`) still fails them now, right?

In other words: 

 - assume that there is a self-signed root CA certificate *C* that used to fail one of the checks in `getUniqueValidationChain`. *C* would then lead to a failure in `sanitizeCert` (with all the necessary consequences on the system.
 - now, after this patch, the invocation of `sanitizeCert` on *C* would likely succeed right?

Now, either we're able to prove that *C* cannot exist with overwhelming probability, or we need to show that either

 -  `sanitizeCert` still behaves identically and fails when it used to fail before (which I don't see yet by inspecting the code; or
 - that the different failure model of this function doesn't lead to a fork.

Does this make more sense?

adecaro:
Very interesting point, @ale-linux. I missed it.
Indeed, we now don't run `cert.Verify(opts)` on the root certificate and this is a problem. 
At this point, we need another way to avoid the issue introduced with go1.19

ale-linux:
@C0rWin - wdyt? Any chance I have missed something? Or is Angelo right and we need to rethink the approach?

C0rWin:
> @C0rWin - wdyt? Any chance I have missed something? Or is Angelo right and we need to rethink the approach?

Let me read thru and I will write back.

C0rWin:
@ale-linux @adecaro thanks guys, I think you are right and have a point. The `cert.Verify` in this case has two functions which is convoluted together:

 - certificate validation/verification 
 - the extraction of the parent  cert to get curve parameters so we can sanitize the certificate
 
 so I do agree with @adecaro and need to find a different way to solve the problem.

adecaro:
what about introducing a new capability that introduces a new MSP version for go1.19 that deviates from the previous behaviour?

ale-linux:
I would try to stay away from the capability if possible. The simplest approach seems to be to just clone the old validation logic - wdyt? How much code are we talking about?

C0rWin:
> I would try to stay away from the capability if possible. The simplest approach seems to be to just clone the old validation logic - wdyt? How much code are we talking about?

I do not think that memic go legacy code is a good idea. I would like to try a different approach. IMO the key issue here is that validation logic is convoluted here with the logic to extract the curve's parameters for sanitization.

C0rWin:
> what about introducing a new capability that introduces a new MSP version for go1.19 that deviates from the previous behaviour?

I agree with @ale-linux we should try to avoid going into new capability

ale-linux:
yeah but the options seem to be either cloning the old logic or introducing a new capability - do you see any other viable alternative?

pfi79:
it seems to me that the problem is that there are places where certificates are added without sanitizeECDSASignedCert
https://github.com/hyperledger/fabric/blob/main/msp/mspimplsetup.go#L466
https://github.com/hyperledger/fabric/blob/main/msp/mspimplsetup.go#L480

C0rWin:
> it seems to me that the problem is that there are places where certificates are added without sanitizeECDSASignedCert https://github.com/hyperledger/fabric/blob/main/msp/mspimplsetup.go#L466 https://github.com/hyperledger/fabric/blob/main/msp/mspimplsetup.go#L480

these are the places I'm looking right now to fix

denyeart:
@C0rWin Thanks for pushing an update, is this ready for a final review? Or is it still work in progress? Difficult to tell when the PR is in Draft since that typically implies work in progress...

C0rWin:
> @C0rWin Thanks for pushing an update, is this ready for a final review? Or is it still work in progress? Difficult to tell when the PR is in Draft since that typically implies work in progress...

I need to fix a few tests + rethinking approach based on comments from @ale-linux and @adecaro

ale-linux:
Thanks @C0rWin - I like the approach now - basically, by inverting the order of calls, you make sure that all cert types will call both functions. This at least ensures that the failure model remains identical - even though the actual error messages might differ. However I don't see this as an issue.

I'm happy to merge it now - but before that

1. I'll leave you a couple of minor comments
2. I'd be grateful to @adecaro and @yacovm if they could take one final look

.github/workflows/verify-build.yml
```diff
@@ -13,7 +13,7 @@ on:
 env:
   GOPATH: /opt/go
   PATH: /opt/go/bin:/bin:/usr/bin:/sbin:/usr/sbin:/usr/local/bin:/usr/local/sbin
-  GO_VER: 1.18.10
+  GO_VER: 1.19
```
ale-linux: Why not `1.19.2` as in the Makefile?

cmd/osnadmin/main_test.go
```diff
@@ -708,6 +708,7 @@ var _ = Describe("osnadmin", func() {
 					"--client-key", clientKey,
 				}
 				output, exit, err := executeForArgs(args)
+				By(output)
```
ale-linux: Why is this needed?

msp/mspimpl.go
```diff
@@ -889,28 +889,42 @@ func (msp *bccspmsp) getCertificationChainIdentifierFromChain(chain []*x509.Cert
...
+			}
+			isRootCACert = true
+			validityOpts.Roots = x509.NewCertPool()
+			validityOpts.Roots.AddCert(cert)
+		}
 		// Lookup for a parent certificate to perform the sanitization
-		var parentCert *x509.Certificate
-		chain, err := msp.getUniqueValidationChain(cert, msp.getValidityOptsForCert(cert))
+		// run cert validation at any rate, if this is a root CA
+		// we will validated already sanitized cert
```
ale-linux: > we will validated already sanitized cert
we will validate an already sanitized cert

msp/mspimpl.go
```diff
@@ -889,28 +889,42 @@ func (msp *bccspmsp) getCertificationChainIdentifierFromChain(chain []*x509.Cert
...
-			parentCert = chain[1]
+		// once we finish validation and this is already
+		// sanitized certificate, there is no need to
+		// sanitize it once again hence we can just return it
+		if isRootCACert {
+			return cert, nil
 		}
 
+		// ok, this is no a root CA cert, and now we
+		// then we have chain of certs we can get parent
```
ale-linux: `we` appears twice

msp/mspimplsetup_test.go
```diff
@@ -67,18 +67,15 @@ U98sznvJPRCkRiwYp5L9C5Xq72CHG/3M6cmoN0Cl0xjZicfpfnZSA/ix
...
-MjE1MjA1NzQ5WhcNMjIwMjE2MjA1NzQ5WjBYMQswCQYDVQQGEwJVUzETMBEGA1UE
-CAwKQ2FsaWZvcm5pYTEWMBQGA1UEBwwNU2FuIEZyYW5jaXNjbzENMAsGA1UECgwE
-T3JnMjENMAsGA1UEAwwET3JnMjBZMBMGByqGSM49AgEGCCqGSM49AwEHA0IABD9x
-9DArA8shjxhqajd9OjTThUoJAHMCKXEORYaN8p/sofXJYYBJvg9y2zEOuevB7++p
-PxhMmNISxt0U5IGlOlSjgYYwgYMwHQYDVR0OBBYEFLqzZtVcEWu2pw4IkpClBg9f
-S4EEMB8GA1UdIwQYMBaAFLqzZtVcEWu2pw4IkpClBg9fS4EEMA8GA1UdEwEB/wQF
-MAMBAf8wCwYDVR0PBAQDAgGmMA8GA1UdEQQIMAaCBE9yZzIwEgYDVR0TAQH/BAgw
-BgEB/wIBADAKBggqhkjOPQQDAgNHADBEAiAccYeHn6h6Q1AA2fZc88sYgReSDSGY
-MsALS92an024EQIgcFMjj0D0j2NhcjULCu0L7aGKac1q8XuCcvzfUdfbsdM=
+MIIBmTCCAUCgAwIBAgIRAKso9vIBAvgOF6UGPP8vGDowCgYIKoZIzj0EAwIwFjEU
```
ale-linux, line 70: can you note how this certificate was generated?

ale-linux (changes requested):
Thanks once more @C0rWin, this was a tough one! Thanks for your patience and perseverance!

cmd/osnadmin/main_test.go
C0rWin: note, this is draft and I wanted to see output of the command :)

.github/workflows/verify-build.yml
C0rWin: I will update to 1.20 once stabilized

msp/mspimplsetup_test.go
C0rWin: I can try, though I am not entirely sure what was the way I did it here. I understand why you are asking this, but is this crucial for this PR to be accepted?

cmd/osnadmin/main_test.go
ale-linux: ack - it can disappear in the final PR then, thx!

msp/mspimplsetup_test.go
ale-linux: no no, but you can see how if we don't document this, we'll have a tough time again soon...

ale-linux: at least let's open an issue if you can't find out how

C0rWin: Ok. I will try to recall, I think I just used some test to generate it.

Makefile
```diff
@@ -78,7 +78,7 @@ METADATA_VAR += CommitSHA=$(EXTRA_VERSION)
 METADATA_VAR += BaseDockerLabel=$(BASE_DOCKER_LABEL)
 METADATA_VAR += DockerNamespace=$(DOCKER_NS)
 
-GO_VER = 1.18.10
+GO_VER = 1.19.2
```
denyeart: At this point, we need 1.19.7 or 1.20.2.

Please update all references to Go version, you can look at https://github.com/hyperledger/fabric/pull/3959 for the complete set of references.

denyeart (commented):
This looks good to me now. Both root and non-root certs go through sanitization and validation now, using the same logic as before, but avoiding the Go 1.19 errors.

Nice job Artem, thanks!

@C0rWin Please update to latest Go version and finalize the PR so that we can merge and backport to release-2.5 and get v2.5.0 released.

pfi79:
maybe backport to 2.4?

C0rWin:
Thanks, @denyeart and @ale-linux, I've addressed all your comments, pushed the last update, and removed the draft.

ale-linux:
LGTM - thanks @C0rWin. @yacovm, @adecaro could you please take one last look as well?

C0rWin:
@denyeart shall we go ahead and merge?

msp/mspimpl.go
```diff
@@ -889,28 +889,42 @@ func (msp *bccspmsp) getCertificationChainIdentifierFromChain(chain []*x509.Cert
 // do have signatures in Low-S. If this is not the case, the certificate
 // is regenerated to have a Low-S signature.
 func (msp *bccspmsp) sanitizeCert(cert *x509.Certificate) (*x509.Certificate, error) {
+	var err error
 	if isECDSASignedCert(cert) {
+		isRootCACert := false
+		validityOpts := msp.getValidityOptsForCert(cert)
+		if cert.CheckSignatureFrom(cert) == nil && cert.IsCA {
```
yacovm: Here we anyway verify a signature even if it's not a CA certificate.

It makes more sense to switch the order of conditions of either side of the && so that if it's not a CA we won't bother verifying the signature.

yacovm:
I'm good with the PR modulo my comment.

msp/mspimpl.go
C0rWin: Pushed fix, thanks.

yacovm (approved):

ale-linux (approved):
Thanks @C0rWin !

denyeart:

PR #3775, opened 2022-11-10
Remove Purge-Marker to avoid unnecessary scanning

### LINKED ISSUE #3444: Delete purgemarker when purging data in background
In purge private data, in the initial design we considered keeping one purge marker that would be used for both a) deleting the data in the background and b) reconciliation. However, this adds inefficiency in the background goroutine as we need to keep the purge marker for reconciliation _forever_.
So, the proposal is to keep two copies of the purge markers for both the above purposes.

manish-sethi:
This PR removes the Purge-Marker after purging all the associated keys and hashed indexes. In addition, for reconciliation, it introduces a separate marker

#### Type of change
- New feature

#### Related issues
Closes #3444


denyeart (approved):

denyeart:

PR #3782, opened 2022-11-14
Update purge private data integration tests

jt-nti:
Test for error when the purge capability is not enabled

### DIFF
--- a/integration/pvtdata/data_purge_test.go
+++ b/integration/pvtdata/data_purge_test.go
@@ -8,12 +8,15 @@ package pvtdata
 
 import (
 	"context"
+	"encoding/base64"
+	"fmt"
 	"io/ioutil"
 	"os"
 	"path/filepath"
 	"syscall"
 
 	"github.com/hyperledger/fabric/integration/nwo"
+	"github.com/hyperledger/fabric/integration/nwo/commands"
 	"github.com/hyperledger/fabric/integration/pvtdata/marblechaincodeutil"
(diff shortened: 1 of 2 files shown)


integration/pvtdata/data_purge_test.go
```diff
@@ -8,151 +8,248 @@ package pvtdata
...
-		process = ifrit.Invoke(networkRunner)
-		Eventually(process.Ready(), network.EventuallyTimeout).Should(BeClosed())
-
-		orderer = network.Orderer("orderer")
-		network.CreateAndJoinChannel(orderer, "testchannel")
-		network.UpdateChannelAnchors(orderer, "testchannel")
-		network.VerifyMembership(
-			network.PeersWithChannel("testchannel"),
-			"testchannel",
+	When("the purge private data capability is not enabled", func() {
```
manish-sethi: This `When` could be placed after `BeforeEach` and `AfterEach`. So, in some sense, the common code for defining variables, network setup, and tear down should be defined once before any `When`. This would remove the duplicate code below as well.

jt-nti: The reason for the `When`s was to have separate `BeforeEach`s since the network is very slightly different (purge is not enabled in one, and enabled in the other. From memory Matt Sykes expected more code duplication in test code so I didn't worry about it but happy to investigate alternatives for setting up the network differently with more common code.

manish-sethi (approved):

manish-sethi:

PR #3800, opened 2022-11-21
Correction of the documentation

sufuy45:
Signed-off-by: sufuy45 <yusuf.cagabey@gmail.com>

#### Type of change

- Documentation update

#### Description

Change in line 326-327: There are no transaction results which could be forwarded to the client, because the transaction is not commited yet.

Change in line 331ff: Even though the gateway service forwards the transaction proposal to the required endorsing peers according to the endorsement policies, the actual tests of the endorsement policies will be take place in phase 3.

### DIFF
--- a/docs/source/peers/peers.md
+++ b/docs/source/peers/peers.md
@@ -323,14 +323,15 @@ a request to the Fabric Gateway service to evaluate a transaction proposal.
 
 The target peer, selected by the client application, executes the transaction
 by invoking chaincode --- this step can be described as simulating the transaction,
-because it runs the transaction without any effect on the ledger. The peer then
-returns its transaction result to the client.  
+because it runs the transaction without any effect on the ledger. The target peer then
+returns its proposal response to the gateway service.
 
 The gateway service also forwards the transaction proposal to the required
 *endorsing peers* (based on the endorsement policies), which also execute the
-transaction and return their results to the peer. The gateway service collects
-all responses, and if they collectively satisfy the endorsement policies, forwards
-the transaction to the ordering service.
+transaction and return their results to the gateway service. The gateway service collects
+all responses and creates a transaction envelope — which it returns to the client (SDK)
+for signing. After the client sent the signed envelope back to the gateway 
+service, the gateway service forwards the envelope to the ordering service.
 
 If the proposed transaction would write to a private data collection, (as *transient* data)
 the client application must explicitly specify the organizations required for endorsement.


docs/source/peers/peers.md
```diff
@@ -323,14 +323,15 @@ a request to the Fabric Gateway service to evaluate a transaction proposal.
...
+because it runs the transaction without any effect on the ledger. The target peer then
+returns its proposal response to the gateway service.
 
 The gateway service also forwards the transaction proposal to the required
 *endorsing peers* (based on the endorsement policies), which also execute the
-transaction and return their results to the peer. The gateway service collects
-all responses, and if they collectively satisfy the endorsement policies, forwards
-the transaction to the ordering service.
+transaction and return their results to the gateway service. The gateway service collects
+all responses and creates a transaction envelope — which it returns to the client (SDK)
```
denyeart, line 332: @andrew-coleman After collecting endorsements, does the gateway always return the proposal response to the client for signing? Just want to make sure the update here is entirely correct.

docs/source/peers/peers.md
```diff
@@ -323,14 +323,15 @@ a request to the Fabric Gateway service to evaluate a transaction proposal.
...
+for signing. After the client sent the signed envelope back to the gateway
```
denyeart, line 333: @sufuy45 
```suggestion
for signing. After the client sends the signed envelope back to the gateway 
```

Also, please keep the part about "The gateway collects all the responses, and if they collectively satisfy the endorsement policies".

denyeart:
The failed unit test is a flake failure and can be ignored.

docs/source/peers/peers.md
```diff
@@ -323,14 +323,15 @@ a request to the Fabric Gateway service to evaluate a transaction proposal.
 
 The target peer, selected by the client application, executes the transaction
 by invoking chaincode --- this step can be described as simulating the transaction,
-because it runs the transaction without any effect on the ledger. The peer then
-returns its transaction result to the client.  
+because it runs the transaction without any effect on the ledger. The target peer then
+returns its proposal response to the gateway service.
```
C0rWin, line 327: this is if you are using the gateway, which is a client in a sense, but you don't have to, hence:

```suggestion
returns its proposal response to the client.
```

docs/source/peers/peers.md
bestbeforetoday: Yes the Gateway peer does return the prepared transaction envelope to the client. The Gateway does not hold the client signing credentials so the client needs to sign the transaction before it is submitted to the Gateway, which in turn forwards to orderer node(s).

docs/source/peers/peers.md
andrew-coleman: This whole page has been rewritten in v2.4 to reflect that all transaction invocations would be done via the gateway service (the old SDKs will be deprecated soon).  So, from that point of view, the proposal response is returned to the gateway service.  However, this whole paragraph doesn't read very well in the context of what precedes it and what follows it.  I'd suggest rewording this paragraph as follows:
```
The target (gateway) peer, selected by the client application, executes the transaction
by invoking chaincode (this can be described as simulating the transaction,
because it runs the transaction without any effect on the ledger).   
The gateway service uses this initial invocation to analyse the actions of the 
chaincode and derive an endorsement policy for the transaction.
```
The next paragraph then goes on to say that it uses this endorsement policy to contact other peers and collect their responses.

PR #3803, opened 2022-11-21
Add the domain parameter to the external service chain code parameter

jinjianfeng-chn:
#### Type of change

- New feature

#### Description

When the chain code is deployed as an external service, the address may be an ip address rather than a domain.
So we need to add a domain field so that it can pass the TLS authentication.


mbwhite:
Hi @jinjianfeng-chn thanks for taking the time for a PR

Out of interest if the domain is known why dos the address need to be an IP address? If IP addresses were preferred would they not be in the CSR as well - so the certificate for TLS is correct?

core/container/externalbuilder/instance.go
```diff
@@ -108,6 +109,7 @@ func (c *ChaincodeServerUserData) ChaincodeServerInfo(cryptoDir string) (*ccintf
 		connInfo.ClientConfig.SecOpts.RequireClientCert = true
 		connInfo.ClientConfig.SecOpts.Certificate = []byte(c.ClientCert)
 		connInfo.ClientConfig.SecOpts.Key = []byte(c.ClientKey)
+		connInfo.ClientConfig.SecOpts.ServerNameOverride = c.Domain
```
C0rWin, line 112: > When the chain code is deployed as an external service, the address may be an ip address rather than a domain. So we need to add a domain field so that it can pass the TLS authentication.

What exactly are you trying to solve here? How is it related to the IP address declaration and the fact you are just trying to override the server name with a different domain name? Why can you not generate a proper certificate? 

You should be able to put IP into SAN within your certificate, and if the domain is known in advance, why would you need to override it?

jinjianfeng-chn:
> Hi @jinjianfeng-chn thanks for taking the time for a PR
> 
> Out of interest if the domain is known why dos the address need to be an IP address? If IP addresses were preferred would they not be in the CSR as well - so the certificate for TLS is correct?



> Hi @jinjianfeng-chn thanks for taking the time for a PR
> 
> Out of interest if the domain is known why dos the address need to be an IP address? If IP addresses were preferred would they not be in the CSR as well - so the certificate for TLS is correct?

The certificate used by the smart contract is used in the internal network. Using IP is the most convenient way to access smart contract services on an internal network. If I use the domain, I have to set up a DNS server.

jinjianfeng-chn:
> Hi @jinjianfeng-chn thanks for taking the time for a PR
> 
> Out of interest if the domain is known why dos the address need to be an IP address? If IP addresses were preferred would they not be in the CSR as well - so the certificate for TLS is correct?

The certificate used by the smart contract is used in the internal network. Using IP is the most convenient way to access smart contract services on an internal network. If I use the domain, I have to set up a DNS server.

core/container/externalbuilder/instance.go
jinjianfeng-chn: Using the service name is the most convenient way to access the smart contract service in the cluster. Using registry services or K8S services, we can easily convert a service name to an IP. But resolving domain into IP addresses can be difficult.
Sometimes, we don't know the Service name when the certificate is generated. Another possibility is that we change the service name, which is more likely to happen. Generating a new certificate is not an easy task.

jinjianfeng-chn: Using the service name is the most convenient way to access the smart contract service in the cluster. Using registry center or K8S services, we can easily convert a service name to an IP. But resolving domain into IP addresses can be difficult.
Sometimes, we don't know the Service name when the certificate is generated. Another possibility is that we change the service name, which is more likely to happen. Generating a new certificate is not an easy task.

jinjianfeng-chn: Sometimes I want to use a single certificate to manage multiple smart contract services. The new smart contract service will be deployed after the certificate is generated. These new smart contract services can't use this certificate.

C0rWin: Ok, so essentially, you just wanted to override the hostname while doing the TLS handshake?

jinjianfeng-chn: yes

ale-linux: Have we considered the security implications of doing this? What is the source of `c.Domain`?

denyeart: I guess "domain" would get sourced from the connection.json as described here?
https://hyperledger-fabric.readthedocs.io/en/latest/cc_service.html

Assuming that is the case, please also update the doc in the same PR at
https://github.com/hyperledger/fabric/blob/main/docs/source/cc_service.md

jinjianfeng-chn: > Assuming that is the case, please also update the doc in the same PR at
yes, domain is defined in connection.json.
I'm gonna go update the doc.

docs/source/cc_service.md
```diff
@@ -155,6 +155,7 @@ For example:
 ```json
 {
   "address": "your.chaincode.host.com:9999",
+  "domain": "your.chaincode.host.com",
```
denyeart, line 159: Please add in definition above also (line 145).

denyeart (commented):
I think the new enhancement is fine.

Please also add relevant tests in instance_test.go following the precedent for the other properties.

docs/source/cc_service.md
jinjianfeng-chn: I'm sorry. I'll go right away.

denyeart:
Thanks for adding the doc updates.

Please also add relevant tests in instance_test.go following the precedent for the other properties.

jinjianfeng-chn:
> Thanks for adding the doc updates.
> 
> Please also add relevant tests in instance_test.go following the precedent for the other properties.



> Thanks for adding the doc updates.
> 
> Please also add relevant tests in instance_test.go following the precedent for the other properties.

OK

denyeart (approved):

Issue #3515, opened 2022-07-07
Orderer v3: remove support for the system channel


tock-ibm:
In v3 we will remove support for the system channel.

- [x] remove the usage of system channel in integration tests
- [x] prevent boot with a system channel
- [x] prevent creating a system channel with channel participation
- [x] cleanup the bootstrap package
- [x] #4184
- [x] clean up the flow for creating a new channel
- [x] #4189
- [x] #4204
- [x] #4185
- [x] clean up config tx templating
- [x] #4203
- [x] #4205
- [x] #4211
- [x] clean up the migration path between system-channel usage and no-sys-channel channel participation api
- [x] #4194
- [x] #4223
- [x] #4202
- [x] #4206

In Epic: #3511

PR #3804, opened 2022-11-22
Orderer v3: remove sys-chan from e2e integration test

tock-ibm:
Change-Id: Ib2a13c26ef48fedae5a69e5ee4c9bad95dfd56d3

#### Type of change

#### Description
Removing system channel usage from integration tests

#### Related issues
Epic: #3511 
Issue: #3515


ale-linux:
@tock-ibm , sounds like you forgot the licence somewhere
```
The following files are missing SPDX-License-Identifier headers:
integration/channelparticipation/appchannel.go

Please replace the Apache license header comment text with:

Checking committed files for traditional Apache License headers ...
The following files are missing traditional Apache 2.0 headers:
integration/channelparticipation/appchannel.go
```

tock-ibm:
> @tock-ibm , sounds like you forgot the licence somewhere
> 
> ```
> The following files are missing SPDX-License-Identifier headers:
> integration/channelparticipation/appchannel.go
> 
> Please replace the Apache license header comment text with:
> SPDX-License-Identifier: Apache-2.0
> 
> Checking committed files for traditional Apache License headers ...
> The following files are missing traditional Apache 2.0 headers:
> integration/channelparticipation/appchannel.go
> ```

Thanks, addressed it. File was removed.

integration/e2e/instantiation_policy_test.go
```diff
@@ -51,27 +53,41 @@ var _ = Describe("InstantiationPolicy", func() {
...
 			network.GenerateConfigTree()
 			network.Bootstrap()
 
-			networkRunner := network.NetworkGroupRunner()
-			process = ifrit.Invoke(networkRunner)
-			Eventually(process.Ready(), network.EventuallyTimeout).Should(BeClosed())
+			// Start all the fabric processes
+			ordererRunner, ordererProcess, peerProcess = startNetwork(network, "orderer")
+
+			//ordererRunner = network.OrdererRunner(network.Orderer("orderer"))
```
yacovm: Let's try to avoid merging commented out code

integration/channelparticipation/channel_participation.go
```diff
@@ -158,3 +162,26 @@ func channelInfoShortMatcher(channel string) types.GomegaMatcher {
 		"URL":  Equal(fmt.Sprintf("/participation/v1/channels/%s", channel)),
 	})
 }
+
+// JoinOrdererJoinPeersAppChannel Joins an orderer to a channel for which the genesis block was created by the network
+// bootstrap. It assumes a channel with one orderer. It waits for a leader (single orderer, always node=1), and then
+// joins all the peers to the channel.
+func JoinOrdererJoinPeersAppChannel(network *nwo.Network, channelID string, orderer *nwo.Orderer, ordererRunner *ginkgomon.Runner) {
```
yacovm, line 169: This is a general thought, and less tied to this specific function, but - given we're removing the system channel, is there a point in having joining utility functions such as this, in the channel participation package of integration? Channel participation is all that is left, so shouldn't we extend nwo.Network to have these utilities instead, or at least have joining capabilities in a different package, as this is now essential and going to be widely used everywhere in the tests?

integration/nwo/network.go
```diff
@@ -773,27 +776,44 @@ func (n *Network) Bootstrap() {
...
+				BaseProfile:           c.BaseProfile,
+				ConfigPath:            n.RootDir,
+				OutputCreateChannelTx: n.CreateChannelTxPath(c.Name),
+			})
+			Expect(err).NotTo(HaveOccurred())
+			Eventually(sess, n.EventuallyTimeout).Should(gexec.Exit(0))
+		}
+	}
+
+	if n.SystemChannel == nil {
```
yacovm: isn't it better to simply check if it's nil, in which case to do the below and add also concatenation of TLS certs, and return?

integration/nwo/standard_networks.go
```diff
@@ -183,6 +183,107 @@ func BasicEtcdRaft() *Config {
 	return config
 }
 
+func BasicEtcdRaftNoSysChan() *Config {
```
yacovm, line 261: can't we simply use other functions like BasicEtcdRaft() and then just nil out the system channel?  Same for the multi channel function below.

integration/channelparticipation/channel_participation.go
tock-ibm: Absolutely correct. 

The end game is to have all this in `nwo` and replace the `nwo.Create/Join Channel` etc. 
However, right now the integration/channelparticipation depends on nwo, I suppose whoever wrote  it in the first place had a cyclic dependency issue. So my plan is to extend this while  we still have tests that use the system channel, and once we finish that clean up, do some refactoring and tidy up the package structure so that all those utility functions are in `nwo`.

integration/e2e/instantiation_policy_test.go
tock-ibm: done

integration/nwo/standard_networks.go
tock-ibm: We're using `BasicConfig()` as the starting point.

Once we're done with removing the system channel, `BasicEtcdRaft` will become unused (and will be eventually removed), and replaced everywhere by `BasicEtcdRaftNoSysChan`. Same goes for all other standard networks.

This is just for the transition period when we have both, this way it is easy to detect which tests where converted and which are still using the system channel.

integration/nwo/network.go
tock-ibm: I am not sure what you are suggesting.

This part of the code uses `configtxgen` to create the genesis blocks of application channels, instead of the channel creation transaction to the system-channel. When the tests use the system channel, creating a channel means loading a pre-made tx. When it is not used, creating a channel means loading a pre-made genesis block. Eventually, when the system channel is removed, we will not need to prepare channel creation transactions at all; but now we still have to support both, for different tests.

Note that this is done in order to avoid using the fabric-config repo as is done in the `integration/channelpaticipation` package, in order to generate genesis blocks.

What do you mean by "also concatenation of TLS certs"?

yacovm: I just meant, to move lines 804-815 to line 789 keeping the `if n.SystemChannel == nil` and inside that `if`, also add `n.ConcatenateTLSCACertificates()` as in line 817 and then `return`. 

This allows you to remove the `if n.SystemChannel != nil` in lines 779 and 790.

integration/nwo/standard_networks.go
yacovm: I understand the intent. 

What I'm trying to understand is, why do you need to create a no-system channel version of any network config constructor, instead of calling the constructor and then just overriding selectively the bootstrap method, channel participation, and system channel?

For example, in `MultiChannelEtcdRaftNoSysChan()` you have:

```
config.Consensus.Type = "etcdraft"
	config.Profiles = []*Profile{
		{
			Name:          "TwoOrgsAppChannelEtcdRaft",
			Consortium:    "SampleConsortium",
			Orderers:      []string{"orderer"},
			Organizations: []string{"Org1", "Org2"},
		},
	}
	config.Channels = []*Channel{
		{Name: "testchannel", Profile: "TwoOrgsAppChannelEtcdRaft"},
		{Name: "testchannel2", Profile: "TwoOrgsAppChannelEtcdRaft"},
	}

	for _, peer := range config.Peers {
		peer.Channels = []*PeerChannel{
			{Name: "testchannel", Anchor: true},
			{Name: "testchannel2", Anchor: true},
		}
	}
```

why?  Is it not possible to call `MultiChannelEtcdRaft` and override only what's needed? 
In general, for a `Foo()` is it not possible to simply create a `FooNoSysChannel()` that calls `Foo()` and then just overrides system channel, bootstrap method and channel participation? This way it's more clear and the cleanup is also more clear - just fold the two functions into one.

tock-ibm: Because the `Foo()` are doing it wrong, and then I have to correct it for the `FooNoSysChan()`, when eventually `Foo()` will disappear. Then I'll have to correct `FooNoSysChan()` again.... There is no point in reusing the `Foo` inside the `FooNoSyschan()` in my opinion. And, in a couple of PRs into the future this point will be moot anyway because all the old `Foo`s will be gone.

I refactored a bit to make my plan more clear, and also removed `MinimalEtcdRaft` which was no longer used.

integration/nwo/network.go
tock-ibm: Ah... got it. 

I did what you suggested but instead kept the `if n.SystemChannel != nil`, and removed the `if n.SystemChannel == nil`. This way when we finish removing the system channel we simply remove the whole  `if n.SystemChannel != nil` block.

yacovm (approved):

PR #3808, opened 2022-11-24
Update purge private data integration tests

jt-nti:
Test whether a new peer is able to reconcile from a purged peer

### DIFF
--- a/integration/pvtdata/data_purge_test.go
+++ b/integration/pvtdata/data_purge_test.go
@@ -15,6 +15,7 @@ import (
 	"path/filepath"
 	"sync"
 	"syscall"
+	"time"
 
 	"github.com/golang/protobuf/proto"
 	cb "github.com/hyperledger/fabric-protos-go/common"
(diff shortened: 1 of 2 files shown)


jt-nti:
Waiting for reconciliation implementation

integration/pvtdata/data_purge_test.go
```diff
@@ -314,12 +336,71 @@ var _ = Describe("Pvtdata purge", func() {
...
+			runner := peerRunners[org3Peer1.ID()]
+			Eventually(runner.Err(), network.EventuallyTimeout).Should(gbytes.Say("Could not fetch \\(or mark to reconcile later\\) 2 eligible collection private write sets for block \\[\\d+\\] \\(0 from local cache, 0 from transient store, 0 from other peers\\). Will commit block with missing private write sets:\\[txID: [0123456789abcdef]+, seq: \\d+, namespace: marblesp, collection: collectionMarblePrivateDetails"))
+
+			for _, peer := range stoppedPeers {
+				startPeer(network, processes, peerRunners, peer)
+			}
+			network.Peers = append(stoppedPeers, org3Peer1)
+
+			// TODO better way to wait for reconciliation on org3 peer1?!
+			time.Sleep(1 * time.Minute)
```
denyeart: It looks like the config already sets reconcileSleepInterval to 10s (down from default of 60s).

The other thing you can do is wait for the successful reconciliation log message similar to how you looked for the "Could not fetch" message above. Then you could wake up without waiting a full minute.

Here's an example of a successful reconciliation message:
`Reconciliation cycle finished successfully. reconciled 1 private data keys from blocks range [9 - 9] channel=mychannel`

In your scenario, it should reconcile everything in a single cycle.

jt-nti: I tried increasing `reconcileBatchSize` it's taking more than one cycle to reconcile everything. I've reduced the length of the sleep instead- hope that's ok.

denyeart (approved):

denyeart:

PR #3831, opened 2022-12-06
set Orderer SendBufferSize to 100 as default

davidkel:
10 has shown to be a bottleneck for Raft performance. 100 appears to be a better default, so proposing that this value is set to 100 by default

Signed-off-by: Dave Kelsey <d_kelsey@uk.ibm.com>

#### Type of change

#### Description

10 has shown to be a bottleneck for Raft performance. 100 appears to be a better default, so proposing that this value is set to 100 by default


yacovm (approved):

denyeart:

PR #3843, opened 2022-12-08
Publish release artifacts to conventional installation paths

jkneubuh:

#### Type of change

- Bug fix

#### Description

This PR updates the release pipeline to publish binary assets to the traditional paths within the GH Release.

In addition, this applies the `TWO_DIGIT_VERSION` release tag, without the leading 'v', to determine the coordinates of the embedded ccenv builder image.

This PR updates the target version of `install-fabric.sh` to 2.5.0 / 1.5.6, and includes an optional `FABRIC_DOCKER_REGISTRY` environment variable in order to override the default Docker Hub container registry.

### DIFF
--- a/scripts/install-fabric.sh
+++ b/scripts/install-fabric.sh
@@ -24,7 +24,8 @@ _arg_comp=('' )
 _arg_fabric_version="2.4.4"
 _arg_ca_version="1.5.5"
 
-ARCH=$(echo "$(uname -s|tr '[:upper:]' '[:lower:]'|sed 's/mingw64_nt.*/windows/')-$(uname -m |sed 's/x86_64/amd64/g')" |sed 's/darwin-arm64/darwin-amd64/g')
+REGISTRY=${FABRIC_DOCKER_REGISTRY:-docker.io/hyperledger}
+ARCH=$(echo "$(uname -s|tr '[:upper:]' '[:lower:]'|sed 's/mingw64_nt.*/windows/')-$(uname -m |sed 's/x86_64/amd64/g')")
 MARCH=$(uname -m)
 
 die()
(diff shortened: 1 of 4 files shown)


scripts/install-fabric.sh
```diff
@@ -21,10 +21,11 @@ _arg_comp=('' )
 
 # if version not passed in, default to latest released version
 # if ca version not passed in, default to latest released version
-_arg_fabric_version="2.4.4"
-_arg_ca_version="1.5.5"
+_arg_fabric_version="2.5.0"
```
denyeart: Since the install instructions have you get install-fabric.sh from main branch, we usually abandon updating this file in the release branches.
But still, it should be the latest published release (2.4.7) so that it actually works if people try it from the release branch.

jkneubuh: ok

scripts/install-fabric.sh
```diff
@@ -49,8 +50,8 @@ print_help()
 {
 	printf 'Usage: %s [-f|--fabric-version <arg>] [-c|--ca-version <arg>] <comp-1> [<comp-2>] ... [<comp-n>] ...\n' "$0"
 	printf '\t%s\n' "<comp> Component to install, one or more of  docker | binary | samples | podman  First letter of component also accepted; If none specified docker | binary | samples is assumed"
-	printf '\t%s\n' "-f, --fabric-version: FabricVersion (default: '2.4.4')"
-	printf '\t%s\n' "-c, --ca-version: Fabric CA Version (default: '1.5.5')"
+	printf '\t%s\n' "-f, --fabric-version: FabricVersion (default: '2.5.0')"
```
denyeart: Here too.

denyeart (approved):

PR #3846, opened 2022-12-12
Support linux-arm64 with install-fabric.sh

jkneubuh:

#### Type of change

- Bug fix

#### Description

When installing fabric CLI, the install-fabric.sh was overriding the platform arch incorrectly when installing on arm64 chipsets.

On linux, it was reading from `aarch64` 
On darwin, it was reading from `amd64`

Installer now downloads binaries from `${PLATFORM}`, defined as `${OS}-${ARCH}`, using the equivalent tokens as `go env GOOS` or `go env GOARCH`

### DIFF
--- a/scripts/install-fabric.sh
+++ b/scripts/install-fabric.sh
@@ -24,8 +24,15 @@ _arg_comp=('' )
 _arg_fabric_version="2.4.7"
 _arg_ca_version="1.5.5"
 
-ARCH=$(echo "$(uname -s|tr '[:upper:]' '[:lower:]'|sed 's/mingw64_nt.*/windows/')-$(uname -m |sed 's/x86_64/amd64/g')" |sed 's/darwin-arm64/darwin-amd64/g')
+REGISTRY=${FABRIC_DOCKER_REGISTRY:-docker.io/hyperledger}
+
+OS=$(uname -s|tr '[:upper:]' '[:lower:]'|sed 's/mingw64_nt.*/windows/')
+ARCH=$(uname -m | sed 's/x86_64/amd64/g' | sed 's/aarch64/arm64/g')
+PLATFORM=${OS}-${ARCH}
+
+# Fabric < 1.2 uses uname -m for architecture.
 MARCH=$(uname -m)
+
 
 die()
 {
@@ -42,7 +49,6 @@ begins_with_short_option()
 	first_option="${1:0:1}"
 	test "$all_short_options" = "${all_short_options/$first_option/}" && return 1 || return 0
 }
-
 
 
 print_help()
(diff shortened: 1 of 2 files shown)


denyeart:
Last year there was a change to pull the amd64 images and binaries even for Mac M1 chipsets:
https://github.com/hyperledger/fabric/pull/3174

I agree this should be undone for v2.5.0 given that we'll have arm64 images and binaries going forward.

I guess the open question is whether we'd want the existing behavior for images and binaries prior to v2.5.0, or is that getting too complex.

jkneubuh:
Apologies - there are two issues addressed by this PR: 

1.  The BUG is that when running on an arm64 on linux, the installer was looking for binaries at `aarch64`, not the "go way" of `arm64` 

2.  The new behavior is that when installing on arm64 for Darwin, we now download the new M1 arm binaries, rather than pulling the amd64 versions and letting the emulator sort it out.

IMO this is getting too complex.  If someone has a specific need to install an older rev of Fabric, they can pin the installer script to be read from a release branch URL, OR simply download the target binary archives.   The one case that breaks here is when an M1 mac runs the install-fabric.sh script, asking for older revs of the client binaries to be installed. 

I'm happy to put in some minor revision tweaking bits in here, but I think it's overkill.  The Mac / IoT / etc. users, in general, will be happy to have native Fabric binaries and docker images available.

jkneubuh:
Seems like an OK compromise with the latest update.   This will now install amd64 binaries for darwin for revs 2.[0-4].

If anyone _really_ wants to install Fabric 1 binaries on an M1 mac ... OK, they can still make this work, even without the convenience of install-fabric.sh.

scripts/install-fabric.sh
```diff
@@ -24,9 +24,14 @@ _arg_comp=('' )
 _arg_fabric_version="2.4.7"
 _arg_ca_version="1.5.5"
 
-ARCH=$(echo "$(uname -s|tr '[:upper:]' '[:lower:]'|sed 's/mingw64_nt.*/windows/')-$(uname -m |sed 's/x86_64/amd64/g')" |sed 's/darwin-arm64/darwin-amd64/g')
+OS=$(uname -s|tr '[:upper:]' '[:lower:]'|sed 's/mingw64_nt.*/windows/')
+ARCH=$(uname -m | sed 's/x86_64/amd64/g' | sed 's/aarch64/arm64/g')
+PLATFORM=${OS}-${ARCH}
+
+# Fabric < 1.0 uses uname -m for architecture.
```
denyeart: Fabric < 1.2 uses uname -m for architecture.

denyeart (commented):
Looks good now, just one nit on the comment.

Also the same needs to be applied to /scripts/bootstrap.sh
(bootstrap.sh is the old installer that we'd like to remove, but we've been keeping it around for a time for anybody that has been using it from a script... maybe v2.5 is a good time to remove...

scripts/install-fabric.sh
jkneubuh: If 2.5 is an LTS let's leave the bootstrap.sh.

denyeart (approved):

denyeart (approved):

PR #3863, opened 2022-12-15
change order of raft state change notification

Param-S:
#### Type of change
- Bug fix

#### Description
Change the order of notification about the raft state change,
first to chain loop and then abdicateleadership go rountine.

#### Additional details
Analysis and observation noted down in the issue: https://github.com/hyperledger/fabric/issues/3629#issuecomment-1353191438 

#### Related issues
#3629 

### DIFF
--- a/orderer/consensus/etcdraft/node.go
+++ b/orderer/consensus/etcdraft/node.go
@@ -151,15 +151,6 @@ func (n *node) run(campaign bool) {
 				n.chain.snapC <- &rd.Snapshot
 			}
 
-			lcs := n.leaderChangeSubscription.Load()
-
-			if lcs != nil && rd.SoftState != nil {
-				if l := atomic.LoadUint64(&rd.SoftState.Lead); l != raft.None {
-					subscription := lcs.(func(uint64))
-					subscription(l)
-				}
-			}
-
 			// skip empty apply
 			if len(rd.CommittedEntries) != 0 || rd.SoftState != nil {
 				n.chain.applyC <- apply{rd.CommittedEntries, rd.SoftState}
--- a/orderer/consensus/etcdraft/chain.go
+++ b/orderer/consensus/etcdraft/chain.go
@@ -792,6 +792,14 @@ func (c *Chain) run() {
 				case c.observeC <- soft:
 				default:
 				}
+
+				lcs := c.Node.leaderChangeSubscription.Load()
+				if lcs != nil {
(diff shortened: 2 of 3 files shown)


orderer/consensus/etcdraft/node.go
```diff
@@ -151,6 +151,11 @@ func (n *node) run(campaign bool) {
 				n.chain.snapC <- &rd.Snapshot
 			}
 
+			// skip empty apply
```
yacovm, line 154: > as per the logs, old leader failed to submit the config transaction post abdication as the node chain was not notified/updated on the raft state(leadership) changes.
> I suspect that the state change notification(order of notification) to different go routines from node ready loop causes this condition.

I agree with this analysis but not sure this is the right way to solve it: 

The chain's apply channel is processed by another goroutine, so it might take the apply from the channel, be preempted, and then this goroutine will reach line 168. So in a sense, there exists a scheduling in the new code form that is equivalent to the old code form, don't you agree?

I would say that what we probably really should have done in the first place is fire the subscriptions right after in `chain.go`:

`		790:		// notify external observer`

because this way we are sure that a subsequent `Submit()` by the abdication goroutine will detect the new leader.

Param-S: > The chain's apply channel is processed by another goroutine, so it might take the apply from the channel, be preempted, and then this goroutine will reach line 168. So in a sense, there exists a scheduling in the new code form that is equivalent to the old code form, don't you agree?

I completely agree.  It is not deterministic due to thread/go rountine scheduling/context switch.  

Since, now we are moving the subscription notification after the chain applied the state changes. it is now streamlined.

yacovm: But now we are firing the subscription twice, no?

In the chain, line 164, and in the node, line 800.

Param-S: sorry, it was mistake while adding the changes into the commit. I corrected it now.

yacovm: Looks good, let's get some more eyes on this. 

@tock-ibm would you please look at this change?

tock-ibm: nit: 
The comment line 1883 in the chain_test.go should be:
// remove first consenter, which is the leader

Otherwise, looks good to me.

yacovm (approved):

denyeart:

Issue #2994, opened 2021-10-24
Native Support For 64bit ARM in fabric


davidkel:
With the introduction of the M1 ARM chip for Apple Macbooks plus the increasing interest in ARM as an alternative to Intel it would be good if fabric supported and published native ARM binaries

M1 Mac users at the moment can make use of Rosetta 2 to use fabric (although the bootstrap script currently fails to pull down binaries and images as it looks for a non-existent ARM architecture in the image name or binary package)

On top of Mac book support it would also facilitate being able to run fabric nodes on 64bit raspberry pi 4 distros which seems to be quite popular for IOT projects

yuxuan-hou-x:
Guys, I find a solutions, refer to https://jira.hyperledger.org/browse/FAB-18389?page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel&showAll=true

denyeart:
Note that a workaround has been merged in bootstrap script to pull down amd64 when using M1 Mac (Rosetta approach that Dave K mentioned). Got confirmation that this approach works.

bestbeforetoday:
Confirmed that with current (v2.4) Fabric release and current fabric-samples, these things work:

- Fabric install using install script.
- Create test-network
- Deploy Go chaincode
- Run Go, Node and Java client applications

What doesn't work is deploying Node and Java chaincode using the `test-network/network.sh` script, with suitable nodeenv and javaenv Docker images not found:

```
Error: chaincode install failed with status: 500 - failed to invoke backing implementation of 'InstallChaincode': could not build chaincode: docker build failed: docker image build failed: docker build failed: Failed to pull hyperledger/fabric-nodeenv:2.4: no matching manifest for linux/arm64/v8 in the manifest list entries
```
```
Error: chaincode install failed with status: 500 - failed to invoke backing implementation of 'InstallChaincode': could not build chaincode: docker build failed: docker image build failed: docker build failed: Failed to pull hyperledger/fabric-javaenv:2.4: no matching manifest for linux/arm64/v8 in the manifest list entries
```

A workaround is to manually pull the x86 node/javaenv Docker images in advance so the chaincode deployment process doesn't try to pull the native arm64 images, which don't exist:

- `docker pull --platform amd64 hyperledger/fabric-nodeenv:2.4`
- `docker pull --platform amd64 hyperledger/fabric-javaenv:2.4`

1055373165:
I already solve this question on my mac m1.

[...]

PR #3882, opened 2022-12-21
Switch base docker image from golang-alpine to ubuntu:20.04

jkneubuh:
#### Type of change

- Bug fix

#### Description

This PR switches the Fabric base docker images from golang-alpine to ubuntu:20.04.  The upgrade is necessary as the golang-alpine libc runtimes (musl) are incompatible with the requirements on multi-arch runtimes. 

The motivations for this upgrade are described in detail in Discussion #3876 and comments in DRAFT PR #3877 

#### Additional details

- Discussion #3876 
- PR #3877  (DRAFT)

#### Related issues

- #2994

### DIFF
--- a/images/ccenv/Dockerfile
+++ b/images/ccenv/Dockerfile
@@ -1,18 +1,44 @@
-# Copyright IBM Corp. All Rights Reserved.
+#
+# Copyright contributors to the Hyperledger Fabric project
 #
 # SPDX-License-Identifier: Apache-2.0
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at:
+#
+# 	  http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
 
+ARG UBUNTU_VER
+FROM ubuntu:${UBUNTU_VER}
+
+ARG TARGETARCH
+ARG TARGETOS
 ARG GO_VER
-ARG ALPINE_VER
-FROM golang:${GO_VER}-alpine${ALPINE_VER}
-RUN apk add --no-cache \
-	binutils-gold \
-	g++ \
-	gcc \
-	git \
-	musl-dev
 
-RUN mkdir -p /chaincode/output /chaincode/input
-RUN addgroup -g 500 chaincode && adduser -u 500 -D -h /home/chaincode -G chaincode chaincode
-RUN chown -R chaincode:chaincode /chaincode
+RUN apt update && apt install -y \
+    binutils-gold \
+    curl \
+    g++ \
+    gcc \
+    git
+
+RUN curl -sL https://go.dev/dl/go${GO_VER}.${TARGETOS}-${TARGETARCH}.tar.gz | tar zxvf - -C /usr/local
+ENV PATH="/usr/local/go/bin:$PATH"
+
+RUN addgroup --gid 500 chaincode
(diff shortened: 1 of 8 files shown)


jkneubuh:
@denyeart @C0rWin @pfi79 @davidkel @lehors - final stake / compromise for arm64 support.

images/ccenv/Dockerfile
```diff
@@ -1,18 +1,44 @@
...
-RUN mkdir -p /chaincode/output /chaincode/input
-RUN addgroup -g 500 chaincode && adduser -u 500 -D -h /home/chaincode -G chaincode chaincode
-RUN chown -R chaincode:chaincode /chaincode
+
+RUN apt update
+RUN apt install -y binutils-gold
+RUN apt install -y curl
+RUN apt install -y g++
+RUN apt install -y gcc
+RUN apt install -y git
```
C0rWin: Why not to group with single RUN?

images/tools/Dockerfile
```diff
@@ -1,41 +1,76 @@
...
-VOLUME /etc/hyperledger/fabric
-COPY --from=tools /go/src/github.com/hyperledger/fabric/build/bin /usr/local/bin
-COPY --from=tools /go/src/github.com/hyperledger/fabric/sampleconfig ${FABRIC_CFG_PATH}
+
+RUN apt update
+RUN apt install -y bash
+RUN apt install -y curl
+RUN apt install -y jq
+RUN apt install -y tzdata
+
```
C0rWin: this also could be grouped together

images/tools/Dockerfile
```diff
@@ -1,41 +1,76 @@
...
+FROM ubuntu:${UBUNTU_VER}
+
+ARG TARGETOS
+ARG TARGETARCH
 ARG FABRIC_VER
-ENV FABRIC_VER ${FABRIC_VER}
-
-# Disable cgo when building in the container.  This will create a static binary, which can be
-# copied and run in the base alpine container without the libc/musl runtime.
-ENV CGO_ENABLED 0
```
C0rWin, line 27: Why did you remove static compilation?

jkneubuh: The static link in docker was only added to work around the SIGSEGVs in alpine.  Now with the build in Docker and with the matching libc runtime FROM ubuntu, the static build is no longer required.

jkneubuh:
Looks like ubuntu:20.04 has been on a tighter diet than debian:slim.  This is on the arm64 - also on amd64 the debian images are a few MB larger. 

```
jkneubuh@gala fabric % docker images 
debian    stable-slim    86f9b934c377   22 hours ago    74.3MB
ubuntu    20.04          43ed104e759f   12 days ago     65.7MB
alpine    latest         d3156fec8bcb   4 weeks ago     7.46MB
```

images/tools/Dockerfile
C0rWin: Compiling binary statically has another side effect of having all dependencies compiled into binary, ensuring the resulting executable is assembled and equipped with all dependencies it was compiled with, ensuring consistency.

jkneubuh: @C0rWin are you proposing that we change the `make` target to generate a static executable?  I thought we had settled on the use of Docker to establish consistency across runtimes.  Recall that the above update `ENV CGO_ENABLED 0` was _new behavior_ introduced in the release 2.5 alpha builds for alpine - the behavior in this PR reverts to the same mechanisms we've been using in Fabric since ~2017, with the primary difference that the base Docker image is now derived from ubuntu.  Running with the ubuntu / gnu libc seems to have resolved all of the errors encountered with alpine's musl runtime.

Also: 

- Formal release binaries are static, generated with the golang cross-compiler.  

- Changing the default `make` target to a static build will cause major issues for the pkcs11 runtimes.

denyeart (approved):
Looks like there is general agreement in the Discussion on the approach. Let's give it a try!

denyeart:
@Mergifyio backport main

PR #3895, opened 2023-01-04
Orderer v3: remove system-channel usage from integration tests: discovery

tock-ibm:
#### Type of change

#### Description

Remove system-channel usage from integration tests: discovery

#### Related issues

Epic: #3511
Issue: #3515


integration/discovery/discovery_test.go
```diff
@@ -114,580 +77,713 @@ var _ = Describe("DiscoveryService", func() {
...
+
+		It("discovers network configuration, endorsers, and peer membership", func() {
+			//
+			// bootstrapping a peer from snapshot
+			//
+			By("generating a snapshot at current block number on org1Peer0")
+			blockNum := nwo.GetLedgerHeight(network, org1Peer0, "testchannel") - 1
+			submitSnapshotRequest(network, "testchannel", 0, org1Peer0, "Snapshot request submitted successfully")
+
+			By("verifying snapshot completed on org1Peer0")
```
yacovm, line 235: Not sure this is really needed unless this somehow blocks until the previous snapshot ends, but I guess it's not worth another iteration to remove it

integration/discovery/discovery_test.go
```diff
@@ -114,580 +77,713 @@ var _ = Describe("DiscoveryService", func() {
...
+				CA:            &nwo.CA{Hostname: "ca"},
+			})
+			config.Consortiums[0].Organizations = append(config.Consortiums[0].Organizations, "Org3")
+			config.Profiles[0].Organizations = append(config.Profiles[0].Organizations, "Org3")
+			config.Peers = append(config.Peers,
+				&nwo.Peer{
+					Name:         "peer0",
+					Organization: "Org3",
+					Channels: []*nwo.PeerChannel{
+						{Name: "testchannel", Anchor: true},
```
yacovm, line 194: do we really have to put anchor: true here? we have the anchors for org1 and org2 and they should be sufficient no? I know it's like in the previous version, just thinking out loud. 

Maybe if there would be some comment that would require a revision, it might be good to change this to false.

integration/discovery/discovery_test.go
```diff
@@ -114,580 +77,713 @@ var _ = Describe("DiscoveryService", func() {
...
 
-		By("joining peer org3Peer0 to channel by a snapshot")
-		joinBySnapshot(network, orderer, org3Peer0, "testchannel", snapshotDir)
+			By("comparing configuration from org1Peer0 and org3Peer0")
+			Expect(proto.Equal(discoveredConfig, discoveredConfig2)).To(BeTrue())
 
-		//
-		// retrieving configuration and validating membership
-		//
+			By("validating the membership data")
```
yacovm, line 271: this is more configuration data than membership data. Maybe worth to address if we decide to make a revision

yacovm (approved):

PR #3897, opened 2023-01-04
Improve gateway retry logic sending to orderers

### LINKED ISSUE #517: Transient orderer error is not being retried
Gateway service fails to retry with orderers when it receives a response status == SERVICE_UNAVAILABLE

```
WARN [gateway] Submit -> Error sending transaction to orderer txID=<TXN_ID> endpoint=<ORDERER> err="rpc error: code = Aborted desc = received unsuccessful response from orderer: SERVICE_UNAVAILABLE"
```
 
As per this logic, when the status is != Status_SUCCESS, it return rpc error as `Aborted`…
https://github.com/hyperledger/fabric/blob/main/internal/pkg/gateway/api.go#L467-L469

Whereas the calling function will retry with the next orderer only if the code is `Unavailable`…
https://github.com/hyperledger/fabric/blob/main/internal/pkg/gateway/api.go#L444-L446

So if the intention is to retry with the next orderer, do we need to handle this status `Status_SERVICE_UNAVAILABLE` in this [condition](https://github.com/hyperledger/fabric/blob/main/internal/pkg/gateway/api.go#L467-L469)…

Ref:
Discussion thread in fabric-dev slack channel: https://ibm-blockchain.slack.com/archives/CL3L0REMD/p1671447903462259

andrew-coleman:
Status >= 500 errors returned by orderers should retried on other orderers

resolves https://github.com/hyperledger/fabric-gateway/issues/517

### DIFF
--- a/internal/pkg/gateway/api.go
+++ b/internal/pkg/gateway/api.go
@@ -435,13 +435,16 @@ func (gs *Server) Submit(ctx context.Context, request *gp.SubmitRequest) (*gp.Su
 		response, err := gs.broadcast(ctx, orderer, txn)
 		if err != nil {
 			errDetails = append(errDetails, errorDetail(orderer.endpointConfig, err.Error()))
+			logger.Warnw("Error sending transaction to orderer", "txID", request.TransactionId, "endpoint", orderer.address, "err", err)
+			continue
 		}
+
 		status := response.GetStatus()
 		if status == common.Status_SUCCESS {
 			return &gp.SubmitResponse{}, nil
 		}
 
-		logger.Warnw("Error sending transaction to orderer", "txID", request.TransactionId, "endpoint", orderer.address, "err", err, "status", status, "info", response.GetInfo())
+		logger.Warnw("Unsuccessful response sending transaction to orderer", "txID", request.TransactionId, "endpoint", orderer.address, "status", status, "info", response.GetInfo())
 
 		if status >= 400 && status < 500 {
 			// client error - don't retry
--- a/internal/pkg/gateway/api.go
+++ b/internal/pkg/gateway/api.go
@@ -432,43 +432,42 @@ func (gs *Server) Submit(ctx context.Context, request *gp.SubmitRequest) (*gp.Su
 	for _, index := range rand.Perm(len(orderers)) {
 		orderer := orderers[index]
 		logger.Infow("Sending transaction to orderer", "txID", request.TransactionId, "endpoint", orderer.address)
-		err := gs.broadcast(ctx, orderer, txn)
-		if err == nil {
+		response, err := gs.broadcast(ctx, orderer, txn)
+		if err != nil {
+			errDetails = append(errDetails, errorDetail(orderer.endpointConfig, err.Error()))
+		}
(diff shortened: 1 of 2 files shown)


internal/pkg/gateway/api.go
```diff
@@ -432,43 +432,42 @@ func (gs *Server) Submit(ctx context.Context, request *gp.SubmitRequest) (*gp.Su
...
-		err := gs.broadcast(ctx, orderer, txn)
-		if err == nil {
+		response, err := gs.broadcast(ctx, orderer, txn)
+		if err != nil {
+			errDetails = append(errDetails, errorDetail(orderer.endpointConfig, err.Error()))
+		}
+		status := response.GetStatus()
+		if status == common.Status_SUCCESS {
 			return &gp.SubmitResponse{}, nil
 		}
```
bestbeforetoday: `err` will be non-nil only on some gRPC connection/invocation error, and in this case `response` will be nil. GetStatus() for a nil `response` will return the zero value for the status so the code below works OK but the behaviour may not be immediately obvious. I wonder if, on a non-nil `err`, it would be clearer to have an explicit log and continue (or use an else-branch). The following code would then only run for a non-nil `response` and the behaviour (and log messages) may be more obvious. What do you think?

```suggestion
		response, err := gs.broadcast(ctx, orderer, txn)
		if err != nil {
			errDetails = append(errDetails, errorDetail(orderer.endpointConfig, err.Error()))
			logger.Warnw("Error sending transaction to orderer", "txID", request.TransactionId, "endpoint", orderer.address, "err", err)
			continue
		}

		status := response.GetStatus()
		if status == common.Status_SUCCESS {
			return &gp.SubmitResponse{}, nil
		}
		
		logger.Warnw("Unsuccessful response sending transaction to orderer", "txID", request.TransactionId, "endpoint", orderer.address, "status", status, "info", response.GetInfo())
```

bestbeforetoday (commented):
Unit tests look really good. One comment/query on the implementation code structure

internal/pkg/gateway/api.go
```diff
@@ -432,43 +432,47 @@ func (gs *Server) Submit(ctx context.Context, request *gp.SubmitRequest) (*gp.Su
...
 
-		logger.Warnw("Error sending transaction to orderer", "txID", request.TransactionId, "endpoint", orderer.address, "err", err)
-		errDetails = append(errDetails, errorDetail(orderer.endpointConfig, err.Error()))
+		logger.Warnw("Unsuccessful response sending transaction to orderer", "txID", request.TransactionId, "endpoint", orderer.address, "status", status, "info", response.GetInfo())
+
+		logger.Warnw("Error sending transaction to orderer", "txID", request.TransactionId, "endpoint", orderer.address, "err", err, "status", status, "info", response.GetInfo())
```
denyeart: Did you intend to log two similar warnings?

andrew-coleman: No I didn't - good spot

denyeart (approved):

denyeart:

denyeart:

PR #3901, opened 2023-01-06
Add PurgedKeyAuditLogging config option

denyeart:
Add peer config option to log purged private data keys. Useful to audit that private data has been purged from a peer, however some users may want to disable the audit logging to prevent private data keys from appearing in peer logs, while preserving the other pvtdatastorage log messages.

Sample log entry:
```
INFO [pvtdatastorage] process -> Purging private data from private data storage channel=mychannel chaincode=private collection=Org1MSPPrivateCollection key=assetp5 blockNum=12 tranNum=0
INFO [pvtdatastorage] process -> Purging private data from private data storage channel=mychannel chaincode=private collection=Org1MSPPrivateCollection key=assetp5 blockNum=13 tranNum=0
INFO [pvtdatastorage] process -> Purging private data from private data storage channel=mychannel chaincode=private collection=assetCollection key=assetp5 blockNum=12 tranNum=0
INFO [pvtdatastorage] process -> Purging private data from private data storage channel=mychannel chaincode=private collection=assetCollection key=assetp5 blockNum=13 tranNum=0

INFO [pvtdatastorage] deleteDataMarkedForPurge -> Purged private data from private data storage channel=mychannel numKeysPurged=2 numPrivateDataStoreRecordsPurged=4
```

### DIFF
--- a/core/ledger/pvtdatastorage/store.go
+++ b/core/ledger/pvtdatastorage/store.go
@@ -47,12 +47,13 @@ type PrivateDataConfig struct {
 
 // Store manages the permanent storage of private write sets for a ledger
 type Store struct {
-	db              *leveldbhelper.DBHandle
-	ledgerid        string
-	btlPolicy       pvtdatapolicy.BTLPolicy
-	batchesInterval int
-	maxBatchSize    int
-	purgeInterval   uint64
+	db                    *leveldbhelper.DBHandle
+	ledgerid              string
+	btlPolicy             pvtdatapolicy.BTLPolicy
+	batchesInterval       int
+	maxBatchSize          int
+	purgeInterval         uint64
+	purgedKeyAuditLogging bool
 
 	isEmpty            bool
 	lastCommittedBlock uint64
@@ -227,6 +228,7 @@ func (p *Provider) OpenStore(ledgerid string) (*Store, error) {
 		batchesInterval:                     p.pvtData.BatchesInterval,
 		maxBatchSize:                        p.pvtData.MaxBatchSize,
 		purgeInterval:                       uint64(p.pvtData.PurgeInterval),
+		purgedKeyAuditLogging:               p.pvtData.PurgedKeyAuditLogging,
 		deprioritizedDataReconcilerInterval: p.pvtData.DeprioritizedDataReconcilerInterval,
 		accessDeprioMissingDataAfter:        time.Now().Add(p.pvtData.DeprioritizedDataReconcilerInterval),
 		collElgProcSync: &collElgProcSync{
@@ -869,7 +871,7 @@ func (s *Store) deleteDataMarkedForPurge() error {
 	maxBatchSize := 4 * 1024 * 1024 // 4Mb
 	purgeMarkerCounter := 0
 	hashedIndexCounter := 0
-	p := newPurgeUpdatesProcessor(s.db, maxBatchSize)
+	p := newPurgeUpdatesProcessor(s.ledgerid, s.db, s.purgedKeyAuditLogging, maxBatchSize)
 	pStart, pEnd := rangeScanKeysForPurgeMarkers()
 
 	// get the purge markers that need to be processed at this block height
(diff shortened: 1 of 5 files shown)


core/ledger/pvtdatastorage/store.go
```diff
@@ -1185,7 +1189,7 @@ func newPurgeUpdatesProcessor(db *leveldbhelper.DBHandle, maxBatchSize int) *pur
 // writeset, hence in every invocation, we should not fetch the writeset from pvtdata store - this is required for correctness reasons not for the performance reasons.
 // Otherwise, previously performed delete operations on the same writeset will become void, as fetching from store will always give the full writeset (as the trimmed one is not yet committed).
 // The deletion of hashedIndexKey is also included in the same batch
-func (p *purgeUpdatesProcessor) process(hashedIndexKey, hashedIndexVal []byte) error {
+func (p *purgeUpdatesProcessor) process(ledgerid string, hashedIndexKey, hashedIndexVal []byte) error {
```
manish-sethi: Like `purgedKeyAuditLogging`, the `ledgerid` is better suited to pass during construction of this processor (`newPurgeUpdatesProcessor`) as it does not change across invocations to this `process` function.

core/ledger/pvtdatastorage/store.go
```diff
@@ -1218,6 +1222,16 @@ func (p *purgeUpdatesProcessor) process(hashedIndexKey, hashedIndexVal []byte) e
...
 	}
+
+	// If configured, log the purged key for audit purpose
+	if p.purgedKeyAuditLogging {
+		decodedDataKey, err := decodeDatakey(dataKey)
+		if err != nil {
+			return err
+		}
+		logger.Infow("Processed private data purge from private data storage", "channel", ledgerid, "chaincode", decodedDataKey.nsCollBlk.ns, "collection", decodedDataKey.nsCollBlk.coll, "key", string(hashedIndexVal), "blockNum", decodedDataKey.nsCollBlk.blkNum, "tranNum", decodedDataKey.txNum)
+	}
```
manish-sethi: As this is being highlighted a audit log, this requires some discussion to be aware of some scenario where it may report falsely.

The log at this place would be a premature. Here, the data for deletion is getting prepared in-memory (imagine a crash scenario). The alternate place for this log would be in the function after 
```
if err := p.db.WriteBatch(p.batch, true); err != nil {
		return err
}
``` 
in the funciton `commitPendingChanges` below. Even a better place could be at the end of function `deleteDataMarkedForPurge`.

In order to be able to do this in the above mentioned function, we would need to keep track of the keys getting deleted -- we can introduce a map in this struct for this purpose. Of course, this could create a reverse problem of not printing even after deleting -- I am not sure if we would want to care about that situation, but that can be handled by leaving this current log here and changing it's tone to "going to purge..."

Further, we can limit the number of logs per key (e.g., not to repeat the log for a key for each block but instead only for the highest block as user just wants to know what keys were purged in a cycle)

Also, about printing the private key in the logs, we can instead just print the hash of the key (in hex) so we don't create another problem for folks to purge from logs.

core/ledger/pvtdatastorage/store.go
denyeart: done

core/ledger/pvtdatastorage/store.go
denyeart: Makes sense... to keep the code most simple, I went with the approach of logging "Purging private data..." for each key as it is processed (if audit logging is enabled). Then conclude with "Purged private data..." with summary counts after the purges are all committed. This way the user sees the keys if they want to, and gets a final confirmation that they were are actually purged. This will also help us in support scenarios.

In terms of logging the actual key versus a hash, my logging philosophy has always been that we can safely log keys, but not values. If we want to scrub keys from log files, I think that would be a much wider change across many peer log statements. Users have two options with the current approach - they can exclude actual private data from the key, or they can disable audit logging. Let's see if that is sufficient.

manish-sethi (approved):

manish-sethi:

PR #3902, opened 2023-01-06
Fix approve for my org to not endorse committed transaction

C0rWin:
When a chaincode is approved and committed, it means that it has been accepted and added to the network. However, it is still possible to approve the exact chaincode definition with the same sequence after it has been committed. This can lead to confusing behavior because the transaction is endorsed even though it will never be considered valid.

This commit proposes a fix to prevent the endorsement of transactions that attempt to approve already committed chaincode definitions with the same sequence. By failing endorsement of these transactions, we can provide a better experience for end users. In addition, this will help to ensure that the chaincode approval process is more transparent and predictable for all parties involved.

### DIFF
--- a/core/chaincode/lifecycle/lifecycle.go
+++ b/core/chaincode/lifecycle/lifecycle.go
@@ -436,6 +436,8 @@ func (ef *ExternalFunctions) ApproveChaincodeDefinitionForOrg(chname, ccname str
 		return errors.WithMessagef(err, "could not set defaults for chaincode definition in channel %s", chname)
 	}
 
+	privateName := fmt.Sprintf("%s#%d", ccname, requestedSequence)
+
 	if requestedSequence == currentSequence {
 		metadata, ok, err := ef.Resources.Serializer.DeserializeMetadata(NamespacesName, ccname, publicState)
 		if err != nil {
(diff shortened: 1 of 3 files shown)


core/chaincode/lifecycle/lifecycle.go
```diff
@@ -436,6 +436,8 @@ func (ef *ExternalFunctions) ApproveChaincodeDefinitionForOrg(chname, ccname str
 		return errors.WithMessagef(err, "could not set defaults for chaincode definition in channel %s", chname)
 	}
 
+	privateName := fmt.Sprintf("%s#%d", ccname, requestedSequence)
```
denyeart, line 439: Did you forget to remove this?

denyeart (commented):
Thanks Artem, this does indeed seem to be more correct behavior.

Let's see if @ale-linux or @mbwhite have any other comments from prior knowledge of chaincode lifecycle design.

core/chaincode/lifecycle/lifecycle.go
C0rWin: Nope, it's being used in lines 462 and 475

mbwhite:
Yes this would be an improvement;  I think it's worth putting into main - and to be vigilant just incase there's edge case that we've fogottne.

C0rWin:
> and to be vigilant just incase there's edge case that we've fogottne.

I've tried to preserve the expected behavior and added some tests, including integration, but yes I think we need to keep an eye on the change. IMO, it's also worthwhile to backport to supported branches.

core/chaincode/lifecycle/lifecycle.go
```diff
@@ -515,6 +502,42 @@ func (ef *ExternalFunctions) ApproveChaincodeDefinitionForOrg(chname, ccname str
 	return nil
 }
 
+func (ef *ExternalFunctions) isAttemptToRedefine(ccname, privateName, packageID string, requestedSequence int64, cd *ChaincodeDefinition, orgState ReadWritableState) (bool, error) {
```
jt-nti: The error messages and code in this new function suggest that the definition is uncommitted however based on the "attempted to redefine the current committed sequence (%d) for namespace %s" message above it looks like the attempt to redefine can also occur for committed definitions which would be confusing. Is that right?

C0rWin: You have two cases:

1. You approved and committed and trying to redefine (requestedSequence == currentSequence)
2. You trying to approve something that was approved before but not committed yet, for instance, to change some of the parameters in chaincode definition (requestedSequence == currentSequence + 1)

In first case, you have an error message:

line 467: "attempted to redefine the current committed sequence (%d) for namespace %s"

In the second case, you have:

line 480: "attempted to redefine uncommitted sequence (%d) for namespace %s with unchanged content"

C0rWin: @jt-nti another case to support is when a new organization might join the channel, so the chaincode is already approved and committed by others organizations in the channel. So now, you need to approve this chaincode for the new organization so its peers can endorse it. Hence, it is necessary to look up from the org's private state whether it already has the chaincode definition approved. If it has, you are trying to change something that was approved already before; otherwise, it's fine since the state is empty, and you need to write your approval into the state.

jt-nti: Thanks for the examples @C0rWin - the first case "You approved and committed and trying to redefine" seems to confirm that the following should be updated to remove `uncommitted` in the new `isAttemptToRedefine` function: 
- `uncommittedMetadata`
- `"could not fetch uncommitted definition"`
- `"Attempting to redefine uncommitted definition at sequence %d"`
- `uncommittedParameters`

C0rWin: @jt-nti fixed as per your comment.

jt-nti: Fab, thanks

C0rWin: @jt-nti can you plz take a look, I have also added fixes for linters

C0rWin: @denyeart, @mbwhite can you guys please take a look on this PR?

yacovm (approved):

PR #3905, opened 2023-01-09
Orderer v3: Remove system channel usage from integration tests: configtx

tock-ibm:
#### Type of change

#### Description

Remove system channel usage from integration tests: configtx

#### Related issues

Epic:#3511
Issue: #3515

### DIFF
--- a/integration/nwo/standard_networks.go
+++ b/integration/nwo/standard_networks.go
@@ -230,31 +230,6 @@ func MultiNodeEtcdRaft() *Config {
 	return config
 }
 
-func MultiNodeBFT() *Config {
-	config := BasicConfig()
-
-	config.Consensus.Type = "BFT"
-	config.Orderers = []*Orderer{
-		{Name: "orderer1", Organization: "OrdererOrg"},
-		{Name: "orderer2", Organization: "OrdererOrg"},
-		{Name: "orderer3", Organization: "OrdererOrg"},
-	}
-	config.Profiles = []*Profile{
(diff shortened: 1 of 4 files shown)


yacovm:
> An the BFT case 

And in the?

integration/nwo/standard_networks.go
```diff
@@ -358,3 +333,28 @@ func FullEtcdRaftNoSysChan() *Config {
...
+	config.Profiles = []*Profile{
+		{
+			Name:          "TwoOrgsAppChannelBFT",
+			Consortium:    "SampleConsortium",
+			Organizations: []string{"Org1", "Org2"},
+			Orderers:      []string{"orderer1", "orderer2", "orderer3"},
+		},
+	}
+	config.SystemChannel = nil
+	config.Consensus.ChannelParticipationEnabled = true
```
yacovm, line 355: we should eventually remove `ChannelParticipationEnabled` option since it's the only way... both here and in the orderer.yaml

yacovm:
LGTM

integration/nwo/standard_networks.go
tock-ibm: That is correct. I added a comment about that to the issue #3515

tock-ibm:
> > An the BFT case
> 
> And in the?

In the BFT case

yacovm (approved):

tock-ibm:
included in #3916

Issue #3903, opened 2023-01-08
Data race in unit test around consensus etcd/raft


tock-ibm:
Package `orderer/common/server` fails unit test from time to time with a data race warning.
[5_Run Unit Tests.zip](https://github.com/hyperledger/fabric/files/10367868/5_Run.Unit.Tests.zip)

```
2023-01-05T11:07:34.9468133Z FAIL
2023-01-05T11:07:34.9468371Z coverage: 60.6% of statements
2023-01-05T11:07:34.9468727Z FAIL	github.com/hyperledger/fabric/orderer/common/server	25.492s
2023-01-05T11:07:34.9469148Z ?   	github.com/hyperledger/fabric/orderer/common/server/mocks	[no test files]
2023-01-05T11:07:34.9469618Z ok  	github.com/hyperledger/fabric/orderer/common/types	0.012s	coverage: [no statements]
2023-01-05T11:07:34.9470100Z ok  	github.com/hyperledger/fabric/orderer/consensus	0.395s	coverage: 50.0% of statements
2023-01-05T11:08:07.5638199Z ok  	github.com/hyperledger/fabric/orderer/consensus/etcdraft	29.101s	coverage: 86.6% of statements
```

https://github.com/hyperledger/fabric/actions/runs/3846044801/jobs/6550968715

```
2023-01-05T11:07:34.9122558Z ==================
2023-01-05T11:07:34.9122916Z WARNING: DATA RACE
2023-01-05T11:07:34.9123185Z Write at 0x00c0003cbbd0 by goroutine 59:
2023-01-05T11:07:34.9123602Z   internal/poll.(*FD).destroy()
2023-01-05T11:07:34.9124181Z       /opt/hostedtoolcache/go/1.18.9/x64/src/internal/poll/fd_unix.go:86 +0xc9
2023-01-05T11:07:34.9124550Z   internal/poll.(*FD).decref()
2023-01-05T11:07:34.9125068Z       /opt/hostedtoolcache/go/1.18.9/x64/src/internal/poll/fd_mutex.go:213 +0x3e
2023-01-05T11:07:34.9125425Z   internal/poll.(*FD).Close()
2023-01-05T11:07:34.9128115Z       /opt/hostedtoolcache/go/1.18.9/x64/src/internal/poll/fd_unix.go:107 +0x7c
2023-01-05T11:07:34.9128473Z   os.(*file).close()
[...]

PR #3919, opened 2023-01-12
synchronize access to wal storage from chain and node

Param-S:
#### Type of change
- Bug fix

#### Description
sychronized the access to raft/wal stroage from
node ready loop and chain apply go routine

#### Related issues
#3903 

### DIFF
--- a/orderer/consensus/etcdraft/chain.go
+++ b/orderer/consensus/etcdraft/chain.go
@@ -1149,12 +1149,10 @@ func (c *Chain) apply(ents []raftpb.Entry) {
 				continue
 			}
 
-			if err := c.Node.storage.Sync(); err != nil {
-				c.logger.Debugf("Failed to sync raft log, error: %s", err)
-			}
+			// persist the WAL entries into disk
+			c.Node.storage.WALSyncC <- struct{}{}
 
 			c.confState = *c.Node.ApplyConfChange(cc)
-
 			switch cc.Type {
 			case raftpb.ConfChangeAddNode:
 				c.logger.Infof("Applied config change to add node %d, current nodes in channel: %+v", cc.NodeID, c.confState.Voters)
--- a/orderer/consensus/etcdraft/node.go
+++ b/orderer/consensus/etcdraft/node.go
@@ -171,6 +171,11 @@ func (n *node) run(campaign bool) {
 			// to the followers and them writing to their disks. Check 10.2.1 in thesis
 			n.send(rd.Messages)
 
+		case <-n.storage.WALSyncC:
+			if err := n.storage.Sync(); err != nil {
+				n.logger.Warnf("Failed to sync raft log, error: %s", err)
+			}
+
 		case <-n.chain.haltC:
 			raftTicker.Stop()
 			n.Stop()
--- a/orderer/consensus/etcdraft/storage.go
+++ b/orderer/consensus/etcdraft/storage.go
@@ -59,6 +59,7 @@ type RaftStorage struct {
 
 	// a queue that keeps track of indices of snapshots on disk
 	snapshotIndex []uint64
+	WALSyncC      chan struct{}
 }
 
 // CreateStorage attempts to create a storage to persist etcd/raft data.
@@ -113,6 +114,7 @@ func CreateStorage(
 		walDir:        walDir,
 		snapDir:       snapDir,
 		snapshotIndex: ListSnapshots(lg, snapDir),
+		WALSyncC:      make(chan struct{}),
 	}, nil
 }
 


orderer/consensus/etcdraft/chain.go
```diff
@@ -1149,12 +1149,10 @@ func (c *Chain) apply(ents []raftpb.Entry) {
 				continue
 			}
 
-			if err := c.Node.storage.Sync(); err != nil {
-				c.logger.Debugf("Failed to sync raft log, error: %s", err)
-			}
+			// persist the WAL entries into disk
+			c.Node.storage.WALSyncC <- struct{}{}
```
yacovm, line 1153: You need to wait for the sync, no?

orderer/consensus/etcdraft/storage.go
```diff
@@ -441,6 +443,7 @@ func (rs *RaftStorage) ApplySnapshot(snap raftpb.Snapshot) {
 
 // Close closes storage
 func (rs *RaftStorage) Close() error {
+	close(rs.WALSyncC)
```
yacovm: why do you close this channel? you are writing to it in line 1153 in `chain.go` isn't that a problem?

orderer/consensus/etcdraft/node.go
```diff
@@ -171,6 +171,11 @@ func (n *node) run(campaign bool) {
 			// to the followers and them writing to their disks. Check 10.2.1 in thesis
 			n.send(rd.Messages)
 
+		case <-n.storage.WALSyncC:
+			if err := n.storage.Sync(); err != nil {
+				n.logger.Debugf("Failed to sync raft log, error: %s", err)
```
yacovm: why debug? we would want to know about it if we failed to sync

orderer/consensus/etcdraft/chain.go
Param-S: It need not to wait.  This is the reasoning for my statement, now the access to storage all moved to `node.go ready goroutine` and it is sequenced/synchronzied on select, hence, the `chain.go apply goroutine` need not to wait.

orderer/consensus/etcdraft/node.go
Param-S: Yes, We can make it warn level.

orderer/consensus/etcdraft/storage.go
Param-S: The channel is getting closed to release the resource when the chain is halted.

orderer/consensus/etcdraft/chain.go
yacovm: But if we don't wait for the sync, then the goroutine of the node can be too late at syncing and in the meanwhile the goroutine of the chain will continue without the sync happening, no? That's not equivalent to the previous code where the sync is synchronous and blocking.

orderer/consensus/etcdraft/storage.go
yacovm: To release what resource?

If the chain is still referenced by a goroutine then closing it won't do any good. If it's not referenced by any goroutine then the channel will also be garbage collected. 

But I am asking - what if the chain goroutine will write to the closed channel after it's closed?

orderer/consensus/etcdraft/chain.go
Param-S: as per my understanding, the sender is blocked until the receiver receives the value on the channel...
`Receivers always block until there is data to receive. If the channel is unbuffered, the sender blocks until the receiver has received the value.` [ref](https://go.dev/doc/effective_go#channels)

If this is the case, the node goroutine will get its chance to proceed further in concurrent with chain goroutine.
 
Yes, there is chance if the node goroutine is context switched between the channel receive and actual sync operation. Then. It is possible that the chain goroutine continues its operation. 

But, still chain goroutine execution does not conflict with the execution of the node goroutine even if the current config change evicts/halts the chain.    Please correct me if I am missing anything here.

orderer/consensus/etcdraft/storage.go
Param-S: Agree. I will remove it.

orderer/consensus/etcdraft/chain.go
yacovm: Fine, leave this as it is.

orderer/consensus/etcdraft/storage.go
yacovm: And seems like the unit test failed on the exact send:

```

==================
panic: send on closed channel

goroutine 2985 [running]:
github.com/hyperledger/fabric/orderer/consensus/etcdraft.(*Chain).apply(0xc007304300, {0xc0072b9830, 0x1, 0x0?})
	/home/runner/work/fabric/fabric/orderer/consensus/etcdraft/chain.go:1153 +0x52b
```

yacovm (approved):

yacovm:

PR #3923, opened 2023-01-14
Fix bug in purging private key from statedb

### LINKED ISSUE #3828: Purged data doesn't get removed from private state on ineligible peers
When purging private data from an ineligible peer, the key doesn't get removed from private state as expected.

The problem is that the purge process runs PRIOR to the state committer.
The purge process will remove the hashed index keys associated with the private key.
Then later when the state committer queries for hashed index keys to find the raw private key (in the hashed index key value), it will find no hashed index key records. The private key therefore won't get added to the set of private state deletions.

Additionally, the error handling for the scenario of no hashed index key records doesn't appear to be working. This is why the problem was silent in the peer logs.
It seems dbItr.Error() incorrectly returns nil at this line:
https://github.com/hyperledger/fabric/blob/main/core/ledger/pvtdatastorage/store.go#L1092

manish-sethi:

This PR
- Makes the read of private key via hashed-index from the private data store, prior to adding the purge-marker, so as to avoid the race condition, where the hashed-index entries could be deleted via the background process, before they are read in the commit path.

- Deletes the purged key from statedb even when the private key is overwritten by a succeeding transaction in the block, and the corresponding private data is missing

#### Type of change
- Bug fix

#### Related issues
closes #3828

### DIFF
--- a/core/ledger/kvledger/kv_ledger.go
+++ b/core/ledger/kvledger/kv_ledger.go
@@ -666,18 +666,11 @@ func (l *kvLedger) commit(pvtdataAndBlock *ledger.BlockAndPvtData, commitOpts *l
 			},
 		)
 	}
-	if err = l.commitToPvtAndBlockStore(pvtdataAndBlock, purgeMarkers); err != nil {
-		return err
-	}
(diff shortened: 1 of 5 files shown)


denyeart (approved):

denyeart:

Issue #3317, opened 2022-04-05
Enable better authentication mechanism for CouchDB


ajayatdlt:
Currently, Peer uses the username and password authentication for the CouchDB authentication.
We should be looking at improving the authentication mechanism. 
CouchDB supports the following types of authentication. 

- Basic
- Cookie 
- Proxy
- JWT 
https://docs.couchdb.org/en/3.2.0/api/server/authn.html
We should be looking for enhancing the authentication by having other mechanism apart from Basic.

aviralagrawalDLT:
Hello team, I am working on this issue. Requesting you to assign the same to me.

aviralagrawalDLT:
Hello team, in order to facilitate these changes, a PR has been created on version 2.2.2 as a direction for moving forward toward this feature support - https://github.com/dltglobal/fabric/tree/2.2.2-couchdb-jwt-support. Commit for the same is https://github.com/dltglobal/fabric/commit/8cfb6b6bcd4dabc0c45cfd1dc7118ddd8dbabc95

Any reviews on the same are welcome and appreciated.

PR #3929, opened 2023-01-16
Support for Couchdb JWT Authentation

aviralagrawalDLT:
Added support for couchdb JWT Authentatication via RS256 algorithm.

Signed-off-by: Aviral Agrawal <aviral.agrawal@dltlabs.io>

#### Type of change

- New feature

#### Description

Added support for token based JWT authentication supported in Couchdb by generating JWT token using RS256 algorithm. 

#### Related issues

https://github.com/hyperledger/fabric/issues/3317


ale-linux:
Thanks @aviralagrawalDLT . This isn't a minor change and so I was wondering whether there was any kind of design document I could read first to get familiar with the objectives, the threat model etc. Also, curious about what you mean by `RSA256` - 256 is an unusual bitlength for RSA?

aviralagrawalDLT:
Hello @ale-linux 

`RSA256 `seems to be a typo from my end(Updated the same in description). It is `RS256`, which is RSA signature algorithm using SHA-256 hashing algorithm.([Reference](https://www.rfc-editor.org/rfc/rfc7518#section-3.1)). As per couchdb, they support `RS256`, `RS384 `and `RS512`([Reference ](https://github.com/apache/couchdb/blob/main/src/jwtf/README.md)). `RS256` was chosen since it is the most commonly used and while development and testing, we consumed the same. There is no major security concern apart from the fact that longer hashing algorithms are generally more secure and `RS512` token generation is faster on 64bit machine while `RS256` token generation is faster on 32bit machine but not considerably([Reference](https://crypto.stackexchange.com/a/53840)). Any suggestions otherwise on what to use would be helpful and appreciated. 

I do not have any design document handy with me right now. Give me some time to whip something up on why this change0 is being suggested.

ale-linux:
> I do not have any design document handy with me right now. Give me some time to whip something up on why this change0 is being suggested.

Thanks!

cendhu (changes requested):
Please add unit-tests to ensure that the added feature works.

denyeart:
No activity for over a year, I'll go ahead and close for now. Please re-open if you intend to continue working on this PR.

PR #3936, opened 2023-01-20
Gateway should apply OrdererEndpointOverrides

### LINKED ISSUE #530: Apply endpoint overrides to orderer addresses in Gateway peer
**As a** Fabric deployer
**I want** orderer endpoint address overrides configured in the peer configuration to be used by the Gateway service
**So that** behaviour of the Gateway service is consistent with the peer delivery service

The relevant peer configuration is here:

https://github.com/hyperledger/fabric/blob/52b9f467531c442ce21d855b97f4fc2ac3fa50f3/sampleconfig/core.yaml#L386-L394

These overrides should be applied by the Gateway to orderer endpoint addresses obtained from service discovery.

andrew-coleman:
The peer local config allows the endpoint addresses and TLS root certs or ordering nodes to be overridden. The gateway should apply these overrides when connecting to orderer nodes for transaction submit.

Resolves https://github.com/hyperledger/fabric-gateway/issues/530

### DIFF
--- a/internal/pkg/gateway/endpoint.go
+++ b/internal/pkg/gateway/endpoint.go
@@ -15,6 +15,7 @@ import (
 	"github.com/hyperledger/fabric-protos-go/peer"
 	"github.com/hyperledger/fabric/gossip/common"
 	"github.com/hyperledger/fabric/internal/pkg/comm"
+	"github.com/hyperledger/fabric/internal/pkg/peer/orderers"
 	"google.golang.org/grpc"
 	"google.golang.org/grpc/connectivity"
 )
@@ -34,6 +35,7 @@ type orderer struct {
 type endpointConfig struct {
 	pkiid        common.PKIidType
 	address      string
+	logAddress   string
 	mspid        string
 	tlsRootCerts [][]byte
 }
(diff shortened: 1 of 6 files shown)


internal/pkg/gateway/endpoint.go
```diff
@@ -90,7 +101,7 @@ func (ef *endpointFactory) newOrderer(address, mspid string, tlsRootCerts [][]by
 	return &orderer{
 		client:          connectOrderer(conn),
 		closeConnection: conn.Close,
-		endpointConfig:  &endpointConfig{address: address, mspid: mspid, tlsRootCerts: tlsRootCerts},
+		endpointConfig:  &endpointConfig{address: address, logAddress: logAddess, mspid: mspid, tlsRootCerts: tlsRootCerts},
```
Param-S, line 104: I think,  endpointConfig: `address: address` should be holding the connAddress right? (if it is overridden then `address` is not the actual endpoint address)

andrew-coleman: No, the connAddress is only used to make the actual connection (on line 93 above).  The address in the endpointConfig structure must be the one that is returned by the discovery service because a future config update would need to match this address to determine whether the connection should be closed or not.

denyeart (approved):

denyeart:

denyeart:

PR #3942, opened 2023-01-25
BroadcastTimeout for Gateway

### LINKED ISSUE #529: Deadlines on all gRPC calls made by Gateway peer
**As a** Fabric deployer
**I want** to be able to configure deadlines / timeouts for all gRPC calls made by Gateway service
**So that** failures can be avoided due to slow or hanging network calls

A particular example of potential issues is in the Submit() implementation, where the Gateway makes broadcast calls to orderers. These calls are done using the Context passed in to Submit(). If the call to an orderer hangs for an extended period of time or indefinitely, the client will give up or the overall flow will be cancelled before another (potentially good) orderer is tried.

All calls made by the Gateway service should use a specific [Context with a timeout](https://pkg.go.dev/context#WithTimeout) created from the parent Context passed to the service call. The timeouts used should be configurable using peer configuration, with sensible default values.

More details on gRPC deadlines here:

https://grpc.io/blog/deadlines/

andrew-coleman:
The Gateway has retry logic for failing ordering nodes, however if an ordering node doesn’t respond, then the Submit call will timeout without allowing other ordering nodes to be tried. This commit implements a BroadcastTimeout that allows individual OSN’s to timeout before the overall Submit timeout expires. Similar logic already exists for calls to endorser peers.

Resolves https://github.com/hyperledger/fabric-gateway/issues/529

### DIFF
--- a/sampleconfig/core.yaml
+++ b/sampleconfig/core.yaml
@@ -53,6 +53,9 @@ peer:
         # endorsementTimeout is the duration the gateway waits for a response
         # from other endorsing peers before returning a timeout error to the client.
         endorsementTimeout: 30s
+        # broadcastTimeout is the duration the gateway waits for a response
+        # from ordering nodes before returning a timeout error to the client.
+        broadcastTimeout: 30s
         # dialTimeout is the duration the gateway waits for a connection
         # to other network nodes.
         dialTimeout: 2m
--- a/core/peer/config_test.go
+++ b/core/peer/config_test.go
@@ -383,6 +383,7 @@ func TestGlobalConfig(t *testing.T) {
 		GatewayOptions: config.Options{
 			Enabled:            true,
 			EndorsementTimeout: 10 * time.Second,
+			BroadcastTimeout:   10 * time.Second,
 			DialTimeout:        60 * time.Second,
 		},
 	}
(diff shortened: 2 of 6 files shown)


denyeart (approved):

denyeart:

denyeart:

denyeart:

denyeart:

sampleconfig/core.yaml
```diff
@@ -53,6 +53,9 @@ peer:
         # endorsementTimeout is the duration the gateway waits for a response
         # from other endorsing peers before returning a timeout error to the client.
         endorsementTimeout: 30s
+        # broadcastTimeout is the duration the gateway waits for a response
+        # from ordering nodes before returning a timeout error to the client.
```
bestbeforetoday, line 57: Is this description correct? Or is it a timeout before the Gateway gives up on a specific orderer and tries another one?

sampleconfig/core.yaml
```diff
@@ -53,6 +53,9 @@ peer:
...
+        broadcastTimeout: 30s
```
bestbeforetoday, line 58: This is a really long default timeout before failing over to another node. The legacy SDKs used 3 seconds. For a node deployed fairly locally on the network, even that is pretty long to wait for a remote call

PR #3945, opened 2023-01-25
Orderer v3: Remove system channel usage from integration tests: raft no.1

tock-ibm:
#### Type of change

#### Description

Orderer v3: Remove system channel usage from integration tests: raft

All of the "ConsensusTypeMigration" test cases from migration_test.go
The first few test cases from config_test.go, until and including  "a single node cluster has the tick interval overridden".

#### Related issues

Epic: #3511
Issue: #3515

### DIFF
--- a/integration/raft/config_test.go
+++ b/integration/raft/config_test.go
@@ -11,6 +11,7 @@ import (
 	"crypto/x509"
 	"encoding/pem"
 	"fmt"
+	"github.com/hyperledger/fabric/common/crypto/tlsgen"
 	"github.com/hyperledger/fabric/integration/channelparticipation"
 	"io/ioutil"
 	"net/http"
@@ -27,7 +28,6 @@ import (
 	"github.com/hyperledger/fabric-protos-go/msp"
 	protosorderer "github.com/hyperledger/fabric-protos-go/orderer"
 	"github.com/hyperledger/fabric-protos-go/orderer/etcdraft"
-	"github.com/hyperledger/fabric/common/crypto/tlsgen"
 	"github.com/hyperledger/fabric/integration/nwo"
 	"github.com/hyperledger/fabric/integration/nwo/commands"
 	"github.com/hyperledger/fabric/integration/ordererclient"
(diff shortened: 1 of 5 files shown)


integration/raft/config_test.go
```diff
@@ -124,8 +125,11 @@ var _ = Describe("EndToEnd reconfiguration and onboarding", func() {
 				Eventually(sess, network.EventuallyTimeout).Should(gexec.Exit(0))
 			}
 
-			By("Creating a new channel")
-			network.CreateChannel("testchannel", orderer1, peer)
+			By("Creating a new channel, joining all orderers")
+			for _, o := range network.Orderers {
+				channelparticipation.JoinOrdererAppChannelCluster(network, "testchannel", o)
```
Param-S: Everytime the call to JoinOrdererAppChannelCluster loads block from the disk & unmarshals the blocks, if we move out this step outside this function 

> func JoinOrdererAppChannelCluster(network *nwo.Network, channelID string, orderer *nwo.Orderer) {
>          appGenesisBlock := network.LoadAppChannelGenesisBlock(channelID)

& make function to cache or return the block, we can avoid the overhead of reading the block more than once in same iteration.

tock-ibm: Done. I pulled the loop into the function such that the block is loaded only once.

integration/raft/migration_test.go
```diff
@@ -82,191 +77,142 @@ var _ = Describe("ConsensusTypeMigration", func() {
...
 		})
 
-		// This test executes the "green path" migration config updates on an etcdraft based system with a system channel
-		// and a standard channel, and verifies that these config updates have the desired effect.
+		// This test executes the "green path" migration config updates on an etcdraft based system, on a standard
+		// channel, and verifies that these config updates have the desired effect.
 		//
 		// The green path is entering maintenance mode, and then changing the consensus type.
-		// In maintenance mode we check that channel creation is blocked and that normal transactions are blocked.
+		// In maintenance mode we check that normal transactions are blocked.
```
Param-S, line 108: channel creation is now allowed in maintenance mode, previously it was not allowed. Why it is allowed now, can you help me to understand it.

tock-ibm: Before, channel creation was done via a TX on the system channel. When the system channel was in maintenance mode, channel creation was not allowed. When there is no system channel, this check becomes moot.

Note that "maintenance mode" is the state of a channel, not an orderer. Thus, for example, when channel X on orderer o1 is in maintenance mode, this has no effect on any other channel that o1 is part of, and has no effect on the ability of o1 to join channel Y. It just blocks channel X to normal TXs and enforces certain config transition rules that facilitate consensus-type migration.

Param-S: Now. I am able to follow it. Thank you

Param-S (commented):
Reviewed PR. It looks fine to me.

yacovm (approved):

PR #3975, opened 2023-01-30
Add BFT support to gateway

andrew-coleman:
Extend the gateway Submit() logic to concurrently send transaction to all available orderers if the channel is configured to use BFT ordering service.


internal/pkg/gateway/api.go
```diff
@@ -427,47 +427,94 @@ func (gs *Server) Submit(ctx context.Context, request *gp.SubmitRequest) (*gp.Su
...
-			response, err = gs.broadcast(ctx, orderer, txn)
-		}()
-		select {
-		case <-done:
-			// Broadcast completed normally
-		case <-ctx.Done():
-			// Overall submit timeout expired
-			logger.Warn("Submit call timed out while broadcasting to ordering service")
-			return nil, newRpcError(codes.DeadlineExceeded, "submit timeout expired while broadcasting to ordering service")
+	if gs.getChannelConfig(request.ChannelId).ChannelConfig().Capabilities().ConsensusTypeBFT() {
```
yacovm: How is `gs.getChannelConfig(request.ChannelId)` implemented in practice? Are we going to the ledger? We shouldn't do I/O each time we send a transaction. 

If this requires I/O it makes sense to simply have a regular boolean field that will get loaded at initialization time for the time being.

internal/pkg/gateway/api.go
```diff
@@ -427,47 +427,94 @@ func (gs *Server) Submit(ctx context.Context, request *gp.SubmitRequest) (*gp.Su
...
+		// For BFT, need to send transaction to ALL orderers
```
yacovm: technically we don't have to send to everyone, but heuristically we do send to everyone.

internal/pkg/gateway/api.go
```diff
@@ -427,47 +427,94 @@ func (gs *Server) Submit(ctx context.Context, request *gp.SubmitRequest) (*gp.Su
...
+		atLeastOneSuccess := false
+	loop:
+		for i := 0; i < len(orderers); i++ {
+			select {
+			case osnErr := <-waitCh:
+				// Broadcast completed normally
+				if osnErr != nil {
+					errDetails = append(errDetails, osnErr)
+				} else {
+					atLeastOneSuccess = true
```
yacovm: at least one is not enough, we should have at least `2f+1`.

internal/pkg/gateway/api.go
```diff
@@ -427,47 +427,94 @@ func (gs *Server) Submit(ctx context.Context, request *gp.SubmitRequest) (*gp.Su
...
+			}(o)
 		}
-
-		if err != nil {
-			errDetails = append(errDetails, errorDetail(orderer.endpointConfig, err.Error()))
-			logger.Warnw("Error sending transaction to orderer", "txID", request.TransactionId, "endpoint", orderer.logAddress, "err", err)
-			continue
+		atLeastOneSuccess := false
+	loop:
+		for i := 0; i < len(orderers); i++ {
```
pfi79: You don't have to wait for all the nodes to respond.
You can wait for 2f+1 responses - nil and exit the function. 
The rest of the goroutines will timeout and the waitCh channel will be deleted

internal/pkg/gateway/api.go
pfi79: @yacovm there could be a variant like this:
- received less than 2f+1 positive responses (e.g., 1), sent an error to the client, but the orderer cluster executed the transaction and added it to the blockchain

yacovm: What you describe is what will happen with our without our control as a gateway. Once a transaction has been sent to an orderer, there is no way to stop it from being included in a block. 

That's why I said that you need to have at least 2f+1 positive responses. 

In practice, if you cannot get 2f+1 positive responses then probably something is wrong with the orderer cluster.

internal/pkg/gateway/api.go
yacovm: Good point. He can return early once he collected 2f+1 positive responses.

internal/pkg/gateway/api.go
pfi79: Not quite so.
It could be a failure between the gateway and several orderer nodes.
But at the same time there is complete consistency between the orderers.

pfi79: If we return an error, the gateway will not wait for the block to be released and will return an error to the client.

If we return "no error" then the gateway will wait for the block to be released. The block may be released or it may end with a timeout error. 

The timeout error is the worst error. What happened after the timeout error is a separate epic

pfi79: Perhaps an integration test is needed to decide how to proceed?

internal/pkg/gateway/api.go
andrew-coleman: The config is cached in memory and only updated in response to config update

internal/pkg/gateway/api.go
andrew-coleman: I'll add integration tests once the BFT orderer PR has been merged

internal/pkg/gateway/api.go
andrew-coleman: Agreed.  And conversely it can fail fast if too many failures are received to attain a quorum

pfi79: the other way around does not work
in the current implementation, one order available to the client may be sufficient

andrew-coleman: After discussing with Yacov, I changed the implementation to require 2f+1 successful responses

internal/pkg/gateway/api.go
```diff
@@ -427,51 +428,111 @@ func (gs *Server) Submit(ctx context.Context, request *gp.SubmitRequest) (*gp.Su
...
+			select {
+			case osnErr := <-waitCh:
+				// Broadcast completed normally
+				if osnErr != nil {
+					errDetails = append(errDetails, osnErr)
+					failures++
+					if failures > n-quorum {
+						break loop
```
pfi79: I guess I don't quite understand.
If you come out here, what happens next?
What's the next scenario?

andrew-coleman: If it finds that it's received too many failures to every attain 2f+1 successes, then bail out and return error to the client.  In this case the error is the gRPC code.Unavailable which hints to the client application that they could retry.  The client app might choose to re-submit the transaction with the same txID (which can be done via the finer-grained API).  If it turns out that the original transaction was processed successfully, then the one that was resent will be rejected as a duplicate.  I'm not sure the gateway can do any better than that.
Having said all that, I think the condition should be `failures > len(orderers) - quorum`.  I'll make that change.

pfi79: Can you tell me where I can see the "finer-grained API"?

andrew-coleman: If you look at the code for `SubmitTransaction()`, you'll see that it's just syntactic sugar for a sequence of method calls at the level of `NewProposal()` ... `Endorse()` ... `Submit()` ... `Status()`.
So you could write your code to use these instead, and put in extra logic between each step depending on your requirements.  In this case, you might want to check the error returned by `Submit()`, and then resend to the orderer by calling `Submit()` again on the same transaction object (which will have the same txID).

https://github.com/hyperledger/fabric-gateway/blob/main/pkg/client/contract.go#L97
https://github.com/hyperledger/fabric-gateway/blob/main/pkg/client/contract.go#L169

pfi79: @andrew-coleman thanks

pfi79: I still think it will increase the number of false positives to return an error.

If a raft client sends a transaction to hlf+smartbft via sdk, there is a 95% chance it will be accepted and executed. This is equivalent to the fact that the client, after sending a request to all orders, will get only one positive response.

What does a positive or negative response from the orderer tell us? Only that the orderer has put the transaction into RequestPool, and not executed it. The execution is still a long way off.

In a distributed system, network failures are normal. And from the fact that there are failures between the client and the cluster nodes, it is premature to conclude that there are the same failures inside the cluster.

internal/pkg/gateway/api.go
```diff
@@ -427,51 +428,111 @@ func (gs *Server) Submit(ctx context.Context, request *gp.SubmitRequest) (*gp.Su
...
+		ordererConfig, ok := config.OrdererConfig()
+		if ok {
+			n = len(ordererConfig.Consenters())
 		}
-
-		status := response.GetStatus()
-		if status == common.Status_SUCCESS {
-			return &gp.SubmitResponse{}, nil
+		f := (float64(n) - 1) / 3
+		quorum := int(math.Ceil(2*f + 1))
```
yacovm: this is not how you compute the quorum :( 

I know I told you 2f+1, but, I always use this term because it's I usually imagine that `n` is congruent to 1 modulo 3. 

However, if we have, for example - `n=9` then `f=2` because we can only support 2 failures (to support 3 failures we need 10 nodes but we have 9). 

In that case, `2f+1` is 5, which means that if the leader is Byzantine, if you only get 5 success acknowledgements and 4 failures, the leader is free to ignore the transaction because 5 nodes cannot lobby for a view change in such a setting. 
However, if you require `n-f` which is 7, then the leader cannot ignore the request because it will eventually be overthrown by the 7 nodes (of course, assuming you actually succeeded sending to 7 honest nodes)

Please look below how to compute the quorum:

https://github.com/SmartBFT-Go/consensus/blob/v2/internal/bft/util.go#L169-L182

pfi79: I agree with Yakov, he's right.
Here's a chart you can use to write a test
| n  |  f | (n+f+1)/2 |
|---|---|---|
| 1...3  | error  | error  |
| 4  | 1  | 3  |
| 5  | 1  | 4  |
| 6  | 1  | 4  |
| 7  | 2  | 5  |
| 8  | 2  | 6  |
| 9  | 2  | 6  |
| 10  | 3  | 7  |

internal/pkg/gateway/api.go
```diff
@@ -427,51 +427,109 @@ func (gs *Server) Submit(ctx context.Context, request *gp.SubmitRequest) (*gp.Su
...
-			// Overall submit timeout expired
-			logger.Warn("Submit call timed out while broadcasting to ordering service")
-			return nil, newRpcError(codes.DeadlineExceeded, "submit timeout expired while broadcasting to ordering service")
+	config := gs.getChannelConfig(request.ChannelId)
+	if config.ChannelConfig().Capabilities().ConsensusTypeBFT() {
+		// For BFT, we send transaction to ALL orderers
+		waitCh := make(chan *gp.ErrorDetail, len(orderers))
+		for _, o := range orderers {
+			go func(ord *orderer) {
+				logger.Infow("Sending transaction to orderer", "txID", request.TransactionId, "endpoint", ord.logAddress)
```
yacovm: I think the way the code is structured now, we have a potential resource accumulation problem.

Imagine you have an orderer that is byzantine and is not responding but you can connect to it. 

You're going to be stuck on: `response, err := broadcast.Recv()` for ever. Because the context's timeout is only on the establishment of the stream, but not on receiving the message of the status back. 

What we need, is to cancel the stream's context when the function that *spawned* the goroutine has ended. Otherwise, this goroutine will live for ever and will never free resources and memory. 

To do that, you need a parent context, that inherits from the client's context, and then the goroutine's context with timeout inherits from the parent context. 

When you exit the function, you cancel the parent context and this way all streams that are "stuck" receiving will be cancelled.

internal/pkg/gateway/api.go
```diff
@@ -427,51 +427,109 @@ func (gs *Server) Submit(ctx context.Context, request *gp.SubmitRequest) (*gp.Su
...
-		}()
-		select {
-		case <-done:
-			// Broadcast completed normally
-		case <-ctx.Done():
-			// Overall submit timeout expired
-			logger.Warn("Submit call timed out while broadcasting to ordering service")
-			return nil, newRpcError(codes.DeadlineExceeded, "submit timeout expired while broadcasting to ordering service")
+	config := gs.getChannelConfig(request.ChannelId)
+	if config.ChannelConfig().Capabilities().ConsensusTypeBFT() {
```
bestbeforetoday, line 435: This function is getting really complex. At the very least could we break the BFT and non-BFT branches out into separate functions?

Maybe it's possible to go further and have a BFT and non-BFT [strategy](https://en.wikipedia.org/wiki/Strategy_pattern) associated with the Server that is updated only in response to relevant config updates, but that is likely beyond the scope of this PR.

internal/pkg/gateway/api.go
```diff
@@ -427,11 +428,81 @@ func (gs *Server) Submit(ctx context.Context, request *gp.SubmitRequest) (*gp.Su
...
+		return gs.submitBFT(ctx, orderers, txn, config, logger)
+	} else {
+		return gs.submitNonBFT(ctx, orderers, txn, logger)
+	}
+}
+
+func (gs *Server) submitBFT(ctx context.Context, orderers []*orderer, txn *common.Envelope, config channelconfig.Resources, logger *flogging.FabricLogger) (*gp.SubmitResponse, error) {
+	// For BFT, we send transaction to ALL orderers
+	ctx, cancel := context.WithCancel(ctx)
```
yacovm: The problem now is that you may finish sending to the quorum before you send to the leader, and the function will return, cancelling the context which will cancel sending to the leader. 

While the transaction will be scheduled eventually, it will be only after a timeout, and this is bad for user experience. 

Let's do something like this:


```
diff --git a/internal/pkg/gateway/api.go b/internal/pkg/gateway/api.go
index d2eebcabe..665fac1cb 100644
--- a/internal/pkg/gateway/api.go
+++ b/internal/pkg/gateway/api.go
@@ -12,6 +12,8 @@ import (
        "io"
        "math/rand"
        "strings"
+       "sync/atomic"
+       "time"
 
        "github.com/hyperledger/fabric-protos-go/common"
@@ -438,6 +440,10 @@ func (gs *Server) Submit(ctx context.Context, request *gp.SubmitRequest) (*gp.Su
 }
 
 func (gs *Server) submitBFT(ctx context.Context, orderers []*orderer, txn *common.Envelope, config channelconfig.Resources, logger *flogging.FabricLogger) (*gp.SubmitResponse, error) {
+       var numFinishedSend uint32
+
+       earliestFinishDeadline := time.Now().Add(time.Millisecond * 500) // TODO: Probably should be configurable
+
        // For BFT, we send transaction to ALL orderers
        ctx, cancel := context.WithCancel(ctx)
        defer cancel()
@@ -448,6 +454,7 @@ func (gs *Server) submitBFT(ctx context.Context, orderers []*orderer, txn *commo
                        ctx, cancel := context.WithTimeout(ctx, gs.options.BroadcastTimeout)
                        defer cancel()
                        response, err := gs.broadcast(ctx, ord, txn)
+                       atomic.AddUint32(&numFinishedSend, 1)
 
                        if err != nil {
                                logger.Warnw("Error sending transaction to orderer", "endpoint", ord.logAddress, "err", err)
@@ -483,6 +490,9 @@ loop:
                        } else {
                                successes++
                                if successes >= quorum {
+                                       if atomic.LoadUint32(&numFinishedSend) < uint32(n) {
+                                               time.Sleep(time.Until(earliestFinishDeadline))
+                                       }
                                        return &gp.SubmitResponse{}, nil
                                }
                        }

```

This ensures we will always linger enough time for the leader to receive the transaction if it is not overloaded and not faulty. 

We probably should export this to a config parameter.

internal/pkg/gateway/api.go
```diff
@@ -427,11 +428,81 @@ func (gs *Server) Submit(ctx context.Context, request *gp.SubmitRequest) (*gp.Su
...
+				waitCh <- nil
+			} else {
+				logger.Warnw("Unsuccessful response sending transaction to orderer", "endpoint", ord.logAddress, "status", status, "info", response.GetInfo())
+				waitCh <- errorDetail(ord.endpointConfig, fmt.Sprintf("received unsuccessful response from orderer: %s", common.Status_name[int32(status)]))
+			}
+		}(o)
+	}
+	n := len(orderers)
+	ordererConfig, ok := config.OrdererConfig()
+	if ok {
```
yacovm, line 449: what if not ok?

internal/pkg/gateway/api.go
```diff
@@ -427,11 +428,81 @@ func (gs *Server) Submit(ctx context.Context, request *gp.SubmitRequest) (*gp.Su
...
+loop:
+	for i, total := 0, len(orderers); i < total; i++ {
+		select {
+		case osnErr := <-waitCh:
+			// Broadcast completed normally
+			if osnErr != nil {
+				errDetails = append(errDetails, osnErr)
+				failures++
+				if failures > total-quorum {
+					break loop
```
yacovm, line 464: I am not sure about "giving up" sending if we succeeded sending to less than a quorum.

I think we should still try to send to as many possible even if we failed to send to more than `f`. 

Probably it's best to do:
```
diff --git a/internal/pkg/gateway/api.go b/internal/pkg/gateway/api.go
index d2eebcabe..874c3a411 100644
--- a/internal/pkg/gateway/api.go
+++ b/internal/pkg/gateway/api.go
@@ -12,6 +12,8 @@ import (
        "io"
        "math/rand"
        "strings"
+       "sync/atomic"
+       "time"
 
        "github.com/hyperledger/fabric-protos-go/common"
@@ -438,6 +440,18 @@ func (gs *Server) Submit(ctx context.Context, request *gp.SubmitRequest) (*gp.Su
 }
 
 func (gs *Server) submitBFT(ctx context.Context, orderers []*orderer, txn *common.Envelope, config channelconfig.Resources, logger *flogging.FabricLogger) (*gp.SubmitResponse, error) {
+       var numFinishedSend uint32
+
+       earliestFinishDeadline := time.Now().Add(time.Millisecond * 500) // TODO: Probably should be configurable
+
+       n := len(orderers)
+
+       defer func() {
+               if atomic.LoadUint32(&numFinishedSend) < uint32(n) {
+                       time.Sleep(time.Until(earliestFinishDeadline))
+               }
+       }()
+
        // For BFT, we send transaction to ALL orderers
        ctx, cancel := context.WithCancel(ctx)
        defer cancel()
@@ -448,6 +462,7 @@ func (gs *Server) submitBFT(ctx context.Context, orderers []*orderer, txn *commo
                        ctx, cancel := context.WithTimeout(ctx, gs.options.BroadcastTimeout)
                        defer cancel()
                        response, err := gs.broadcast(ctx, ord, txn)
+                       atomic.AddUint32(&numFinishedSend, 1)
 
                        if err != nil {
                                logger.Warnw("Error sending transaction to orderer", "endpoint", ord.logAddress, "err", err)
@@ -461,7 +476,6 @@ func (gs *Server) submitBFT(ctx context.Context, orderers []*orderer, txn *commo
                        }
                }(o)
        }
-       n := len(orderers)
        ordererConfig, ok := config.OrdererConfig()
        if ok {
                n = len(ordererConfig.Consenters())

```

internal/pkg/gateway/api.go
```diff
@@ -427,11 +428,81 @@ func (gs *Server) Submit(ctx context.Context, request *gp.SubmitRequest) (*gp.Su
...
+func (gs *Server) submitBFT(ctx context.Context, orderers []*orderer, txn *common.Envelope, config channelconfig.Resources, logger *flogging.FabricLogger) (*gp.SubmitResponse, error) {
+	// For BFT, we send transaction to ALL orderers
+	ctx, cancel := context.WithCancel(ctx)
+	defer cancel()
+	waitCh := make(chan *gp.ErrorDetail, len(orderers))
+	for _, o := range orderers {
+		go func(ord *orderer) {
+			logger.Infow("Sending transaction to orderer", "endpoint", ord.logAddress)
+			ctx, cancel := context.WithTimeout(ctx, gs.options.BroadcastTimeout)
+			defer cancel()
```
yacovm: (Placing the comment here randomly)

Can we perhaps abstract this method into a function, and actually make a unit test for this? This method seems complex enough to deep a unit test of its own.

yacovm:
Actually, the below might be a better way of minimizing the time the method is active:

```
diff --git a/internal/pkg/gateway/api.go b/internal/pkg/gateway/api.go
index d2eebcabe..574229f3d 100644
--- a/internal/pkg/gateway/api.go
+++ b/internal/pkg/gateway/api.go
@@ -12,6 +12,8 @@ import (
        "io"
        "math/rand"
        "strings"
+       "sync/atomic"
+       "time"
 
        "github.com/hyperledger/fabric-protos-go/common"
@@ -438,7 +440,24 @@ func (gs *Server) Submit(ctx context.Context, request *gp.SubmitRequest) (*gp.Su
 }
 
 func (gs *Server) submitBFT(ctx context.Context, orderers []*orderer, txn *common.Envelope, config channelconfig.Resources, logger *flogging.FabricLogger) (*gp.SubmitResponse, error) {
+       earliestFinishDeadline := time.Now().Add(time.Millisecond * 500)
+       everyoneSubmitted := make(chan struct{})
+       var numFinishedSend uint32
+
+       defer func() {
+               t1 := time.NewTimer(time.Until(earliestFinishDeadline))
+               defer t1.Stop()
+               select {
+               case <-t1.C:
+                       return
+               case <-everyoneSubmitted:
+                       return
+               }
+       }()
+
        // For BFT, we send transaction to ALL orderers
+       n := len(orderers)
+
        ctx, cancel := context.WithCancel(ctx)
        defer cancel()
        waitCh := make(chan *gp.ErrorDetail, len(orderers))
@@ -448,6 +467,10 @@ func (gs *Server) submitBFT(ctx context.Context, orderers []*orderer, txn *commo
                        ctx, cancel := context.WithTimeout(ctx, gs.options.BroadcastTimeout)
                        defer cancel()
                        response, err := gs.broadcast(ctx, ord, txn)
+                       // If I'm the last to submit, notify this
+                       if atomic.AddUint32(&numFinishedSend, 1) == uint32(n) {
+                               close(everyoneSubmitted)
+                       }
 
                        if err != nil {
                                logger.Warnw("Error sending transaction to orderer", "endpoint", ord.logAddress, "err", err)
@@ -461,7 +484,6 @@ func (gs *Server) submitBFT(ctx context.Context, orderers []*orderer, txn *commo
                        }
                }(o)
        }
-       n := len(orderers)
        ordererConfig, ok := config.OrdererConfig()
        if ok {
                n = len(ordererConfig.Consenters())

```

internal/pkg/gateway/api.go
andrew-coleman: If it's not ok, then it defaults to the number of (channel) orderers that the gateway is connected to (& cached in the registry).  This should be the same, since the registry gets the connection information from the same OrdererConfig.  However, if the gateway was unable to connect to an orderer, then these numbers could be different.  It depends whether we want to calculate the consensus from the number of connected orderers, or the number of orderers defined in the channel config.  Either way, we can only attempt to broadcast to the connected ones.

internal/pkg/gateway/api.go
andrew-coleman: I didn't intent to give up sending the transaction to all orderers, but rather to give up waiting for responses from orderers once consensus has been achieved.  But I can see how the code I had written could have that effect.  I've re-coded that bit.

andrew-coleman:
I've amended the commit with a variation on this suggestion such that it will wait the full BroadcastTimeout for sending the transaction to all ordering nodes without holding up the gateway Submit() method beyond the time need to collect 'consensus' responses.  This is done by invoking a broadcastToAll() function in a separate go routine with a new (non derived) context, which will return when all orderers have received the transaction, or timeout has expired.

internal/pkg/gateway/api.go
```diff
@@ -427,11 +430,101 @@ func (gs *Server) Submit(ctx context.Context, request *gp.SubmitRequest) (*gp.Su
...
+		return gs.submitNonBFT(ctx, orderers, txn, logger)
+	}
+}
+
+func (gs *Server) submitBFT(ctx context.Context, orderers []*orderer, txn *common.Envelope, config channelconfig.Resources, logger *flogging.FabricLogger) (*gp.SubmitResponse, error) {
+	// For BFT, we send transaction to ALL orderers
+	waitCh := make(chan *gp.ErrorDetail, len(orderers))
+	go gs.broadcastToAll(orderers, txn, waitCh, logger)
+
+	n := len(orderers)
```
yacovm, line 447: if you had a config block committed during this method that changes `n`, then you will send to `n` and then analyze for `n' != n`. Let's cache `n` and the orderers and use both of them for both sending and analyzing.

yacovm (approved):
we can address the comment in a separate PR

PR #4010, opened 2023-02-09
Return orderer error text to client

andrew-coleman:
If a call from the gateway to an orderer node returns an error in the BroadcastResponse message, then it is appended to the details field of the rpc error that gets returned to the client application.  However, the Info field of that message is not getting passed on, so the client could be lacking valuable information on the underlying cause of the error.  This commit fixes that.


yacovm (approved):

andrew-coleman:

andrew-coleman:

PR #4015, opened 2023-02-10
Close connection to unavailable orderers

andrew-coleman:
Currently, if an orderer goes down, the stale connection to it still remains cached in the gateway registry.  This commit implements similar logic as used by connections to external endorsing peers whereby a stale connection is closed and flushed from the cache allowing future invocations to attempt to reconnect.

### DIFF
--- a/internal/pkg/gateway/api.go
+++ b/internal/pkg/gateway/api.go
@@ -490,8 +490,12 @@ func (gs *Server) broadcastToAll(orderers []*orderer, txn *common.Envelope, wait
 				close(everyoneSubmitted)
 			}
 			if err != nil {
-				logger.Warnw("Error sending transaction to orderer", "endpoint", ord.logAddress, "err", err)
+				rpcErr := status.Convert(err)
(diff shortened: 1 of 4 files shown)


yacovm:
I don't think this is a good idea. Why do we want to do book-keeping of the connections ourselves? 
gRPC has a mode where it reconnects to the remote node on its own when it is unreachable, via exponential backoff. 
You can activate it by using the `AsyncConnect` flag in the `ClientConfig`.  

I think the right approach would be to let gRPC do its probing for us, and we just probe its connectivity state similarly to what's done in the orderer cluster package. 

Suppose an orderer is down for a day. Every time a client tries to do a broadcast  - are you going to try and use this orderer and connect to it synchronously? This would slow down the client, no?

Besides, looking at the current code (not the code you have in the PR) don't we have a problem that if an orderer is down and the gateway is contacted by a lot of clients concurrently (suppose, by 20 clients), then you might connect to the orderer 20 times and then close 19 connections and use the first one?

PR #4018, opened 2023-02-13
ledgerutil: Add "verify" command

shimos:
#### Type of change

- New feature

#### Description

We would like to propose a new command in Ledger Utility that checks the integrity of a single ledger. The existing commands in Ledger Utility aim at checking the ledger when any difference is found by comparing two ledgers. The purpose of the command proposed is to complement these commands by finding any defect in ledger files in a single peer that could be caused by potential software bugs or disk errors.

This patch adds a new command, “verify,” to Ledger Utility, which performs integrity checks for a specified ledger. Currently, the checks for the hash values in the headers of the blocks are implemented.

A use case assumed is to run the command locally in a peer periodically, which can be utilized as one of the health checks for the peer.

#### Additional details

This patch contains a sample ledger that is tweaked to have a hash value error (the hash value for Block 0 is modified not to match the block payload). Using the ledger, the command proposed can detect the error as follows:

```
$ ./build/bin/ledgerutil verify ./internal/ledgerutil/testdata/sample_bad_ledger -o /tmp/ledgerutil-verify-result-bad
...
Successfully executed verify tool. Some error(s) are found.
$ cat /tmp/ledgerutil-verify-result-bad/mychannel_verification_result/blocks.json
[
{"number":0,"valid":false,"errors":["DataHash mismatch"]}
,
{"number":1,"valid":false,"errors":["PreviousHash mismatch"]}
]
```


internal/ledgerutil/verify/verify.go
```diff
@@ -0,0 +1,224 @@
...
+
+const (
+	ledgersDataDirName      = "ledgersData"
+	blockResultJsonFilename = "blocks.json"
+
+	errorDataHashMismatch = "DataHash mismatch"
+	errorPrevHashMismatch = "PreviousHash mismatch"
+)
+
+// Verify - Verifies the integrity of ledgers in a single peer
```
denyeart:
```suggestion
// VerifyLedger verifies the integrity of ledgers in a single peer
```
Exported function comment should start with the exact function name.

Also, please add more comment details about what is actually verified.

internal/ledgerutil/verify/verify.go
```diff
@@ -0,0 +1,224 @@
...
+func VerifyLedger(blockStorePath string, outputDir string) (bool, error) {
+	// Save whether any error is detected
+	noLedgerError := true
```
denyeart: It is good practice to use Go zero values, for a boolean it would be 'false'.

Then the variable could be called something like ledgerErrorFound.... it would remain false and then turn true if an error is found.

internal/ledgerutil/verify/verify.go
```diff
@@ -0,0 +1,224 @@
...
+
+	return blockStoreProvider, nil
+}
+
+// blockResultOutput - stores the number of valid blocks and those of erroneous blocks
+// and the JSON writer object
+type blockResultOutput struct {
+	ValidBlocks     uint64
+	ErroneousBlocks uint64
+	Writer          *jsonrw.JSONFileWriter
```
denyeart: These can start with lower case to signal unexported.

internal/ledgerutil/verify/verify.go
```diff
@@ -0,0 +1,224 @@
...
+}
+
+// blockCheckResult - stores the result of the checks for a block
+type blockCheckResult struct {
+	Number uint64   `json:"number"`
+	Valid  bool     `json:"valid"`
+	Errors []string `json:"errors,omitempty"`
+}
+
+// Perform checks on a block
```
denyeart:
```suggestion
// checkBlock checks a block's hash
```

Not as big a deal for unexported functions, but still the function comment should start with the function name.

denyeart (commented):
This is a great addition to ledgerutil, thanks!

shimos:
@denyeart Thank you for the review! I have fixed the patch according to your suggestions.

denyeart:
@manish-sethi Could you review as well? Anything else you think should be in a ledger 'verify' utility?

internal/ledgerutil/verify/verify.go
```diff
@@ -0,0 +1,228 @@
...
+			return false, err
+		}
+		defer store.Shutdown()
+
+		info, err := store.GetBlockchainInfo()
+		if err != nil {
+			return false, err
+		}
+
+		iterator, err := store.RetrieveBlocks(0)
```
manish-sethi: If a channel has been bootstapped by a snapshot, the starting block would be a non-zero. You can use the `info` struct in line 75 above to determine this, and get the first available block number.

manish-sethi (commented):
Not to discourage, but personally, I do not see much value (usefulness) in this command. Still, I won't object from merging this if any other maintainer wants (and @denyeart likes this) as it is in the offline utility space. Left a comment below.

denyeart:
@shimos One of the new unit tests failed.

shimos:
@manish-sethi Thank you for the comment and review. I have added a logic to determine the first block available and a unit test using a ledger bootstrapped from a snapshot.

@denyeart Thank you for letting me know. I have fixed the failed unit test.

internal/ledgerutil/verify/verify.go
```diff
@@ -0,0 +1,246 @@
...
+	return &blockResultOutput{
+		validBlocks:     0,
+		erroneousBlocks: 0,
+		writer:          writer,
+	}, nil
+}
+
+// blockCheckResult - Stores the result of the checks for a block
+type blockCheckResult struct {
+	Number uint64   `json:"number"`
```
denyeart: The casual user may not know what "number" is.
The other utility uses the json property "blockNum" for this, would be good to be consistent.

denyeart (commented):
Sorry it took me a while... I wanted to try out the user experience before merging.

The user experience is good. I've just highlighted one suggestion and then I think it will be ready for merge.

We'll also need to add the documentation, do you want to do that in this PR or a next one?

See the existing doc at:
https://hyperledger-fabric.readthedocs.io/en/latest/commands/ledgerutil.html

There are instructions for how to add CLI doc here:
https://github.com/hyperledger/fabric/blob/main/docs/source/docs_guide.md#adding-a-new-cli-command

shimos:
@denyeart Thank you for trying out and the suggestion. I've addressed your point.

> We'll also need to add the documentation, do you want to do that in this PR or a next one?

Yes, I'm happy to work on the documentation and submit another PR.

denyeart (approved):

denyeart:
@shimos Ok, I'll merge this PR and you can work on the doc in a next PR. Thanks!

denyeart:

PR #4026, opened 2023-02-14
Remainder of Performance considerations edit

joshhus:
Edited starting with "Orderer considerations" continuing through end: 
- Note in places we say collecting "enough" signatures - are there some cases where a specific signature is required in addition to number. 
- It may be useful to indicate how much better Go is than Node for performance, if they are close or not close.

#### Type of change

- Documentation update

#### Description

new doc, per request


docs/source/performance.md
```diff
@@ -126,84 +124,82 @@ Orderer: &OrdererDefaults
...
+CouchDB as a state database has other limitations which reduce overall network performance and require additional hardware resource (and cost). For one, JSON queries are not re-executed at validation time, so CouchDB does not offer built-in protection from phantom reads. Such protection is provided, however, by range queries. (Your application must be designed to tolerate phantom reads when using JSON queries, such as when records are added to state between the times of chaincode execution and block validation). You should therefore consider using range queries based on additional keys for high-throughput applications, rather than using CouchDB JSON with queries.
```
davidkel: ```so CouchDB does not offer built-in protection from phantom reads```
It's not CouchDB that offers this protection, it's fabric that offers this and can't do it for JSON queries. The above reads that it's more a CouchDB limitation

```rather than using CouchDB JSON with queries.``` I think should be ```rather than using CouchDB with JSON queries.```

File: docs/source/performance.md [hunk 2]
```diff
@@ -213,74 +209,74 @@ Suffice to say large payload sizes are an anti-pattern in any blockchain solutio
...
+Node is an asynchronous runtime implementation that utilizes only a single thread to execute code. Node does run background threads for activities such as garbage collection, but it is against best practices to allocate more than two vCPUs (generally equivalent to the number of concurrent threads which can be executed) to a Node chaincode (for example on Kubernetes which is limited to available resources). It is worth monitoring performance of Node chaincode to see how much vCPU it uses, though you cannot assign resource restrictions to Node chaincode as it is self-limiting.
```
davidkel: ```but it is against best practices to allocate more than two vCPUs``` I think this is too harsh a statement. It's not against best practices really and it may be that you can get slightly more performance if you do allocate more than 2 vCPUs because node can run other threads concurrently but it might not be worth the cost to do so for the little gain there is. So I would still prefer to emphasise that node is generally self-limiting (so not limiting the number of cpus will not cause the node chaincode to use cpu resource unbounded) and that there may not be much value allocating more than 2 vCPUs but monitoring resource utilisation is paramount

```though you cannot assign resource restrictions to Node chaincode as it is self-limiting```. You absolutely can assign resource restrictions to node chaincode processes. But if you don't node chaincode is self-limiting so it won't be unbounded

File: docs/source/performance.md [hunk 3]
```diff
@@ -213,74 +209,74 @@ Suffice to say large payload sizes are an anti-pattern in any blockchain solutio
...
 
-Optimise your queries, complex queries will take more time even with indexing. Ensure your queries result in a bounded set of data. Remember Fabric may also limit the total number of results returned.
+Optimize your queries, because complex queries will take more time even with indexing. Ensure that your queries result in a bounded set of data. Hyperledger Fabric may also limit the total number of results returned.
 
-You can check peer and CouchDB logs to see how long queries are taking and also whether a query was unable to use an index from warning log entries stating the query "should be indexed".
+You can check the peer and CouchDB logs to see how long queries are taking, and whether any query was unable to use an index from a warning log entry stating the query "should be indexed".
 
-If queries are taking a long time, you could try to increase CPU/Memory available to CouchDB or use faster storage.
+If queries are taking a long time, you can increase the CPU/Memory available to CouchDB or use faster storage.
```
davidkel: This implies it will definitely help, but we don't know for sure depending on setup/scenario etc. Which is why it originally was worded as `you can try`

docs/source/performance.md
```diff
@@ -126,84 +124,82 @@ Orderer: &OrdererDefaults
...
 
-### Use the new peer gateway service
+### Use the Peer Gateway Service
 
-The peer Gateway Service and the new Fabric-Gateway client SDKs are a substantial improvement over the legacy Go, Java and Node SDKs. Not only do they provide much improved throughput, they also provide better capability reducing the complexity of a client application. For example the Gateway SDKs will automatically collect enough endorsements to satisfy not only the chaincode endorsement policy but also any state-based endorsement policies that get included when the transaction is simulated, something that was not possible with the legacy SDKs.
+The Peer Gateway Service (new in v2.4) and Fabric-Gateway client SDKs offer substantial enhancements over the legacy Go, Java, and Node SDKs. The latest Gateway SDKs  provide much improved throughput, and more capabilities and reduced complexity to client applications. For example, the Gateway SDKs automatically (attempt to) collect the endorsements required to satisfy not only the chaincode endorsement policy, but also any state-based endorsement policies included with transaction simulation (which is not possible using the legacy SDKs).
```
davidkel: ```The latest Gateway SDKs provide much improved throughput, and more capabilities and reduced complexity to client applications. For example, the Gateway SDKs automatically (attempt to) collect the endorsements required to satisfy not only the chaincode endorsement policy, but also any state-based endorsement policies included with transaction simulation (which is not possible using the legacy SDKs).```

All of this statement implies it's the gateway SDKs the provide the capability. In fact the SDKs do very little and it's the Peer Gateway service that does all the work. Client implementations using the new SDKs do not have reduced complexity from the application developer, but the overall application will have a vastly reduced codepath (as the new SDKs are a lot smaller than the legacy sdks). This combination of new SDK and Peer Gateway service provides the improved throughput and new capabilities not available in the legacy SDKs

docs/source/performance.md
```diff
@@ -126,84 +124,82 @@ Orderer: &OrdererDefaults
...
 
-It also reduces the number of network connections a client needs to maintain in order for a client to submit a transaction. Previously, clients may need to connect to multiple peer and orderer nodes across organizations. The peer Gateway Service service enables an application to target a single trusted peer, then the peer Gateway Service connects to other peer and orderer nodes to gather endorsements and submit the transaction on behalf of the client application. Of course, you may want to target multiple trusted peers for high concurrency and redundancy.
+The Peer Gateway Service also reduces the number of network connections a client needs to maintain to submit a transaction. Using the legacy SDKs, clients may need to connect to multiple peer and orderer nodes across organizations. By contrast, the Peer Gateway Service service enables an application to connect to a single trusted peer, which then collects endorsements from other peer and orderer nodes and submits the transaction on behalf of the client application. However, you can still choose to have your application target multiple trusted peers, for high concurrency and redundancy.
```
davidkel: ```However, you can still choose to have your application target multiple trusted peers, for high concurrency and redundancy.```
This doesn't read right to me. Above we said you can connect to a single trusted peer. So the next sentence I think reads better if we way
```Your application can target multiple trusted peers for high concurrency and redundancy as required```

**Comment by davidkel** [comment]
- Note in places we say collecting "enough" signatures - are there some cases where a specific signature is required in addition to number.

There is a minimum number of signatures required by count as well as spanning the organisations that are required to endorse, however it isn't an issue if you provide more than is required, eg if you have a policy where either org1 must endose OR org2 must endorse, then you can provide endorsements (with signatures) from both Org1 and Org2.

- It may be useful to indicate how much better Go is than Node for performance, if they are close or not close

It will depend on the chaincode scenario but they are not close. Some old performance figures show a factor of 2x more throughput with Go chaincode compatred but this factor changes depending on things like asset size. I don't really want to quote a figure here given the factor changes based on conditions plus things will have changed in fabric v2.5, for example Go binaries could be more performant now and Node v16 could be less performant and the old performance figures I think can only give an indication of how the chaincodes compare now.

**Comment by davidkel** [comment]
@joshhus Many thanks for the updates Josh. I've made a few comments.

File: docs/source/performance.md (hunk 1 above)
Comment (joshhus) [reply docs/source/performance.md]: fixed both.

File: docs/source/performance.md (hunk 2 above)
Comment (joshhus) [reply docs/source/performance.md]: So the point is you don't need to assign resource restrictions bec. Node is self limiting? Is that the guidance.

Comment (joshhus) [reply docs/source/performance.md]: "Node chaincode is self-limiting and not unbounded, though you could also assign resource restrictions to Node chaincode processes."

File: docs/source/performance.md (hunk 3 above)
Comment (joshhus) [reply docs/source/performance.md]: sure, the trying applies to speeding up, not to the increasing memory.

Comment (joshhus) [reply docs/source/performance.md]: "If queries are taking a long time, increasing the CPU/Memory available to CouchDB or using faster storage may improve their performance."

File: docs/source/performance.md (hunk 5 above)
Comment (joshhus) [reply docs/source/performance.md]: sure. I changed "as required" to "when necessary" so it can't be read as a flat requirement.

**Review by joshhus — COMMENTED** [review]
ok

File: docs/source/performance.md (hunk 4 above)
Comment (joshhus) [reply docs/source/performance.md]: ok

**Review by joshhus — COMMENTED** [review]
ok

File: docs/source/performance.md [hunk 6]
```diff
@@ -213,74 +209,74 @@ Suffice to say large payload sizes are an anti-pattern in any blockchain solutio
...
+Node is an asynchronous runtime impluementation that utilizes only a single thread to execute code. Node does run background threads for activities such as garbage collection, but it may not improve performance to allocate more than two vCPUs (generally equivalent to the number of concurrent threads which can be executed) to a Node chaincode (for example on Kubernetes which is limited to available resources). It is worth monitoring performance of Node chaincode to see how much vCPU it uses. Node chaincode is self-limiting and not unbounded, though you could also assign resource restrictions to Node chaincode processes.
```
denyeart:
```suggestion
Node is an asynchronous runtime implementation that utilizes only a single thread to execute code. Node does run background threads for activities such as garbage collection, but it may not improve performance to allocate more than two vCPUs (generally equivalent to the number of concurrent threads which can be executed) to a Node chaincode (for example on Kubernetes which is limited to available resources). It is worth monitoring performance of Node chaincode to see how much vCPU it uses. Node chaincode is self-limiting and not unbounded, though you could also assign resource restrictions to Node chaincode processes. 
```

denyeart (approved):

PR #4028, opened 2023-02-16
Use local peer ledger height from ledger instead of discovery in Gateway

### LINKED ISSUE #558: Use Gateway peer local ledger height when selecting peers
**As an** application developer
**I want** the Gateway peer to use the ledger height obtained directly from its local ledger
**So that** the latest available information is used for evaluate / endorse peer selection

The Gateway peer currently uses only ledger heights obtained from service discovery to aid selection of the highest block height peer. Since there is a minimum time window between service discovery updates being published, the ledger height for the local peer recorded in service discovery can be slightly behind the actual point-in-time ledger height. Instead the Gateway peer should query the height of its local ledger directly instead of relying on the ledger height obtained from service discovery for itself.

bestbeforetoday:
Discovery ledger heights can be slightly behind the actual peer ledger height. Ensure the most up-to-date information is used for the local peer by querying the ledger height directly. This avoids the possibility of a remote peer with a lower ledger height being selected as an endorser.

Refactored Gateway service method implementations and associated tests into separate files as the size of the api files were getting unmanageable.

Closes hyperledger/fabric-gateway#558


yacovm (dismissed):
Can we please separate the refactoring part and using the ledger into different pull requests? 

I'm not sure merging one implies the other should be merged and vice versa.

bestbeforetoday:
> Can we please separate the refactoring part and using the ledger into different pull requests?

See #4032. I will rebase this change on top of that change.

internal/pkg/gateway/registry.go
```diff
@@ -200,13 +202,42 @@ func (reg *registry) endorsersByOrg(channel string, chaincode string) map[string
...
 }
 
 func (reg *registry) isLocalEndorserID(pkiID gossipcommon.PKIidType) bool {
 	return !pkiID.IsNotSameFilter(reg.localEndorser.pkiid)
 }
 
+func (reg *registry) localLedgerHeight(channel string) (height uint64, ok bool) {
+	ledger, err := reg.localProvider.Ledger(channel)
+	if err != nil {
+		reg.logger.Debugw("local endorser is not a member of channel", "channel", channel, "err", err)
```
yacovm: why is this a debug? You are receiving an abnormal request. It should be a warning no?

internal/pkg/gateway/registry.go
```diff
@@ -200,13 +202,42 @@ func (reg *registry) endorsersByOrg(channel string, chaincode string) map[string
...
+		return 0, false
+	}
+
+	info, err := ledger.GetBlockchainInfo()
+	if err != nil {
+		logger.Debugw("failed to get local ledger info", "err", err)
```
yacovm: This should be an error, no?

yacovm (commented):
LGTM modulo comments. Maybe Andrew can take a look at it too.

internal/pkg/gateway/registry.go
bestbeforetoday: Sure, I can change it to warning or error. It should never happen

internal/pkg/gateway/registry.go
bestbeforetoday: Pushed an update with the changes to log level

andrew-coleman (approved):

PR #4032, opened 2023-02-16
Refactor of Gateway code

bestbeforetoday:
Split some code into separate files. Particularly the gRPC service method implementation and tests were unreasonably large, making the codebase difficult to navigate and extend. Code that was previously in api.go is now located in: chaincodeevents.go, commitstatus.go, diff.go, endorse.go, evaluate.go, and submit.go. Unit tests specific to these functions are split out of api_test.go into corresponding _test.go files.

Simplification of the way chaincode events unit tests define the local peer ledger height. Previously a postSetup function for specific tests reconfigured the ledger mocks. Now only a localLedgerHeight property is specified in the test definition.


denyeart (approved):
Thanks for the refactor, much better!

PR #4060, opened 2023-02-22
Different Output in logs

vijayrajpanchall:
#### Type of change

- Documentation update

#### Description

after run the "./network.sh createChannel" command I got this output "Channel 'mychannel' joined"

image
image


denyeart:
Thanks for the update. CI failed since we are no longer maintaining/updating release-2.0 branch. It looks like this change is already in main branch the and the more recent release-2.x branches. I'll go ahead and close.

PR #4067, opened 2023-02-26
GetIdemixMspConfig should require signerBytes not nil

LiRiu:
GetIdemixMspConfig should require signerBytes not nil so that msp.Setup() do not panic

#### Type of change
- Improvement

#### Description

The GetIdemixMspConfig function ignores errors that the SignerConfig file is empty or does not exist. This causes a panic when the MSP is initialized later.


denyeart:
Files in the /vendor directory are go module dependencies that get pulled in via go.mod.

You'll need to make the contribution directly to https://github.com/IBM/idemix , and assuming it gets merged, then update fabric go.mod to pull in the updated dependency.

Also, every contribution commit must be signed, see details at https://github.com/hyperledger/fabric/pull/4067/checks?check_run_id=11606178938.

PR #4011, opened 2023-02-09
Orderer v3: fix IsChannelMember to match public keys

tock-ibm:
#### Type of change

- Bug fix
- Improvement (improvement to code, performance, etc)
- 
#### Description

Orderer v3: fix IsChannelMember to match public keys

- Remove system channel usage from:  `integration/raft/cft_test.go` use-case: "disregards certificate renewal if only the validity period changed"
- Fix a bug in IsChannelMember  that manifested in that test
- Addresses the bug reported in issue #3998

#### Related issues

Addresses: #3998 
Parent issue: #3515

PR #4070, opened 2023-02-27
Orderer v3: fix IsChannelMember to match public keys (backport #4011)

dviejokfs:
This is a backport of the pull request https://github.com/hyperledger/fabric/pull/4011.

### DIFF
--- a/orderer/consensus/etcdraft/consenter.go
+++ b/orderer/consensus/etcdraft/consenter.go
@@ -7,7 +7,6 @@ SPDX-License-Identifier: Apache-2.0
 package etcdraft
 
 import (
-	"bytes"
 	"path"
 	"reflect"
 	"time"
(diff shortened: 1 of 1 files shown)


denyeart:
@dviejokfs Is there a reason the updates to the two test files weren't backported? Even though the tests may not be identical between main and release-2.4 I believe it should be possible to backport the tests.

Also, usually the backport PRs for release-2.5 and release-2.4 are done at the same time so that we don't forget one of them.

dviejokfs:
@denyeart 

`MultiNodeEtcdRaftNoSysChan` is not defined in the release-2.4

There's a big difference between the code related to the Orderer in the `main` branch and the `release-2.4` branch

tock-ibm:
> @denyeart
> 
> `MultiNodeEtcdRaftNoSysChan` is not defined in the release-2.4
> 
> There's a big difference between the code related to the Orderer in the `main` branch and the `release-2.4` branch

@dviejokfs @denyeart 

The bug manifested itself when this testcase was converted to not use the system channel: integration/raft/cft_test.go use-case: "disregards certificate renewal if only the validity period changed"

Since in 2.4 & 2.5 we still support the system channel, what should be done is to add another version of the testcase, but without the system channel. Since 2.4 & 2.5 are lacking some of the support we already implemented into main, that would have to be done ad-hoc for these tests. I can help with this.

denyeart:
Ok, we could either add the test ad-hoc in the back releases, or we could accept the risk given that there are tests in main branch and the code change is the same across branches and fairly simple.

@tock-ibm As the maintainer with most knowledge here, I'll defer to your judgement...

tock-ibm:
@dviejokfs @denyeart 

see: backports with tests
#4071 to release-2.5
#4073 to release-2.4

denyeart:
Ok, closing this one in favor of #4073

PR #4079, opened 2023-03-07
Update default Fabric install version to latest release

bestbeforetoday:


denyeart:
Thanks @bestbeforetoday , I missed this after the last release... opened a separate PR to catch all the missed updates - https://github.com/hyperledger/fabric/pull/4081 , will close this one so that they're all together.

PR #4118, opened 2023-03-28
Log peer config upon startup

denyeart:
Log peer config upon startup.
Also log peer environment variables to help troubleshoot peer config.

PR #4122, opened 2023-03-29
Log peer config upon startup (#4118)

denyeart:
Log peer config upon startup.
Also log peer environment variables to help troubleshoot peer config.

### DIFF
--- a/internal/peer/node/start.go
+++ b/internal/peer/node/start.go
@@ -98,6 +98,7 @@ import (
 	"github.com/spf13/cobra"
 	"github.com/spf13/viper"
 	"google.golang.org/grpc"
+	"gopkg.in/yaml.v2"
 )
 
 const (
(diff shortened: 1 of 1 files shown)


internal/peer/node/start.go
```diff
@@ -185,6 +186,23 @@ func (c custodianLauncherAdapter) Stop(ccid string) error {
...
+	// Info logging for peer config, includes core.yaml settings and environment variable overrides
+	allSettings := viper.AllSettings()
+	settingsYaml, err := yaml.Marshal(allSettings)
+	if err != nil {
+		return err
+	}
+	logger.Infof("Peer config with combined core.yaml settings and environment variable overrides:\n%s", settingsYaml)
+
+	// Debug logging for peer environment variables
+	logger.Debugf("Environment variables:")
```
ale-linux, line 200: Sorry Dave, a nit: any specific reason why we use debug for this but info for the previous?

denyeart: I was debating this myself. In my opinion the exact peer config should be easily knowable by every user and support engineer hence Info level. This is consistent with what orderer node provides but for some reason peer never provided it.
The wider set of environment variables may be less important and potentially could leak sensitive information about the operating environment hence Debug level.
However, showing the environment variables will help a user or support engineer know whether a certain config setting came from core.yaml or from an environment variable override, so it is valuable information when troubleshooting a peer config.
I could go either way, WDYT?

celder628:
A couple of notes.

1.  I have to agree that INFO is the best way of dumping the variables in the log.  This prints unambiguously and is a big help to anyone who is reviewing logs.  Especially if they are not familiar with the variables being used (support).

2.  The precedence here is also the Ordering nodes which print all variable with INFO at startup.

denyeart:
I talked to celder628, he was talking about the effective config being printed at INFO level in the prior comment. He said that for the raw environment variables DEBUG is sufficient. So I think I'll keep the PR as-is.

manish-sethi (approved):

Issue #4127, opened 2023-03-30
PKCS11 in Fabric 2.2.x doesn't support MSPs with CA certificates that included RSA public keys


marianothiago:
When starting the peer on fabric 2.2.9 with HSM (PKCS11), the following error message is being displayed:

"Certificate's public key type not recognized. Supported keys: [ECDSA]

To fix this bug, a pull request was requested

https://github.com/hyperledger/fabric/pull/4128

denyeart:
Some history (also posted to the PR):

RSA removal commit - https://github.com/hyperledger/fabric/commit/80a20e3aaf5fcfa70e2e3cd8a3e10eb31b1d5901
Discussion about RSA removal and decision to restore - https://github.com/hyperledger/fabric/pull/2069
Restore RSA for SW implementation - https://github.com/hyperledger/fabric/pull/2121

It looks like it was only restored for SW implementation, so it makes sense to restore for PKCS11 implementation as well.

denyeart:
@marianothiago Thanks for fixing in release-2.2. Have you checked whether updates are needed in main and release-2.5? It looks like the code structure is different in those branches.

Depending on your answer, we could close this issue as complete or suggest updates in main and release-2.5. Note that most production users will be updating from v2.2.x to v2.5.x since it is the latest LTS release.

marianothiago:
@denyeart Thanks for help. I ran several tests on version 2.5.0, including organizations that have MSP certificates using RSA in their chain, and it worked great, as expected.

denyeart:
ok, closing.

PR #4128, opened 2023-03-30
bug fix 4127. Restore RSA support for x509 public key import on PKCS11

marianothiago:
#### Type of change

- Bug fix

#### Description

While RSA has never been supported by Fabric for transaction signing and
verification, prior to version 2.x, MSPs could include CA certificates
that included RSA public keys. This was broken with the removal of the
unsupported RSA code.

It was fixed in SW but the error is still in PKCS11

This change modifies the key import function used by the MSP on PKCS11 to convert
an RSA public key to a bccsp.Key that can then be used as part of an MSP
identity.

This code does *not* add support for RSA on transaction paths.
#### Related issues
https://github.com/hyperledger/fabric/issues/4127

### DIFF
--- a/bccsp/pkcs11/ecdsakey_test.go
+++ b/bccsp/pkcs11/ecdsakey_test.go
@@ -32,5 +32,5 @@ func TestX509PublicKeyImportOptsKeyImporter(t *testing.T) {
 	cert.PublicKey = "Hello world"
 	_, err = ki.KeyImport(cert, &bccsp.X509PublicKeyImportOpts{})
 	assert.Error(t, err)
-	assert.Contains(t, err.Error(), "Certificate's public key type not recognized. Supported keys: [ECDSA]")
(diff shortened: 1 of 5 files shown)


ale-linux:
Thanks for your PR. A couple of Qs/remarks:

1. please look at completing the DCO
2. you say that this PR restores code that was mistakenly remove - which commit removed RSA from pkcs? Could you pls point me to it?

denyeart:
@ale-linux 
RSA removal commit - https://github.com/hyperledger/fabric/commit/80a20e3aaf5fcfa70e2e3cd8a3e10eb31b1d5901
Discussion about RSA removal and decision to restore - https://github.com/hyperledger/fabric/pull/2069
Restore RSA for SW implementation - https://github.com/hyperledger/fabric/pull/2121

It looks like it was only restored for SW implementation, so it makes sense to restore for PKCS11 implementation as well.

ale-linux:
LGTM @marianothiago , if you follow the DCO and fix that issue, I think we're ready to merge!

ale-linux (approved):

PR #4141, opened 2023-04-03
use HasGet() instead of Has() and Get() on fastcache lib

cendhu:
#### Type of change

#### Description

This pull request addresses the issue of inefficiency when retrieving a stored value from Fastcache by consolidating the two API calls, `Has()` and `Get()`, into a single call using the `HasGet()` API. By doing so, the performance overhead associated with making two API calls is eliminated.


C0rWin (approved):

denyeart:

Issue #4143, opened 2023-04-03
Provide basic Issue Templates/Links


ghost:
In order to avoid questions on the issue-tracker, templates can be used. Provide a basic template/link.

(I can do this within the next days)              

---

> The project doesn't use the issues list for usage questions and getting help, but we often leave the issues open in case somebody can help.

If nobody can help, eventually the issue like the one referenced here is closed with a comment like the following:

> Please reserve Fabric issue list for code issues. 
> For getting community help, see 
> https://hyperledger-fabric.readthedocs.io/en/latest/CONTRIBUTING.html#communications-and-getting-help

_Originally posted by @denyeart in https://github.com/hyperledger/fabric/issues/4140#issuecomment-1494552050_

ghost:
> I think the two important issue categories are "Bug Report" which you have, and "Proposed Work Item". Could you add the latter? It could go to a blank entry.
https://github.com/hyperledger/fabric/pull/4144#issuecomment-1494828754

The quasi-standard on github is `bug` (labeled 'bug'), and `Feature Request` (labeled `Enhancement`).

For "Proposed Work Item", could you elaborate a little bit, so I can add an explanation. (who proposes this? if its the visitors, then an `Enhancement/Feature Request` would be more appropriate. If its for the Team, then ok.)

denyeart:
#4144 merged as a first pass on the issue template, thanks @abebeos .

Concerning a second button, I was envisioning it to be used by core contributors for "todos" and "improvements", as well as for public users for "feature requests", hence the generic name. But you're right, let's add two more buttons to make them clear, how about:

- "Feature Request" (`enhancement` label)

and

- "Other work item" (no label)

Both could use a blank template, so the only real difference would be the automatic `enhancement` label.

Thanks again for helping to improve the user community experience.

denyeart:
Also, I'm not seeing the "Support, Help, Questions, Chat" button. Maybe the `about` field is required under contact_links.
See the example at https://docs.github.com/en/communities/using-templates-to-encourage-useful-issues-and-pull-requests/configuring-issue-templates-for-your-repository#configuring-the-template-chooser.

ghost:
It should appear, did in other repos. Will take a look later.

ghost:
@denyeart I've create the templates, please feel free to change as you see fit.

[...]

PR #4144, opened 2023-04-03
add basic issue template/links

ghost:
#### Type of change

- Documentation update

re #4143

#### Description

Strangely, when I'm testing this on my work-fork, I go the bug-template and your vulnerability handling:

https://github.com/abebeos/fabric/issues/new/choose

No idea why, possibly once you merge to your repo this will behave as it should.

(nothing wrong to have the `Report vulnerability`, but it should not appear magically)

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug.yaml
+++ b/.github/ISSUE_TEMPLATE/bug.yaml
@@ -1,0 +1,22 @@
+# Copyright the Hyperledger Fabric contributors. All rights reserved.
+#
+# SPDX-License-Identifier: Apache-2.0
+
+name: Bug Report
+description: Let us know what went wrong
+labels: ["bug"]
+body:
+  - type: textarea
+    id: bug
+    attributes:
+      label: Description
+      description: What went wrong?
+    validations:
+      required: true
+  - type: textarea
+    id: repro
+    attributes:
+      label: Steps to reproduce
+      description: How to reproduce the issue (ideally, provide cli instructions)
+    validations:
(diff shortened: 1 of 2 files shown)


denyeart:
Thank you for the new issue template.

To get past the DCO failure, [please sign the commit](https://github.com/hyperledger/fabric/pull/4144/checks?check_run_id=12483890977).

To get past the Basic Checks, either add a license header at the top of the new files or add the .github directory to the [excluded_files list](https://github.com/hyperledger/fabric/pull/4144/checks?check_run_id=12483890977) (the latter probably makes more sense).

I think the two important issue categories are "Bug Report" which you have, and "Proposed Work Item". Could you add the latter? It could go to a blank entry.

ghost:
> I think the two important issue categories are "Bug Report" which you have, and "Proposed Work Item". Could you add the latter? It could go to a blank entry.

Please refer to https://github.com/hyperledger/fabric/issues/4143#issuecomment-1494887319

(I would prefer that we merge this here to see how it works, and it can be extended afterwards easily)

denyeart (approved):

denyeart:
Ok @abebeos I've merged it, I also see the "Report a security vulnerability" button.

It looks like github presents it automatically, see https://github.com/orgs/community/discussions/45567, although it may only be visible for the repo administrators. Do you see the option?

ghost:
Yes, though

> This submission will only be viewable to repository maintainers. 

So it seems all is good.

PR #4161, opened 2023-04-13
chore(ci): try new builers

ryjones:
Signed-off-by: Ry Jones <ry@linux.com>

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -24,10 +24,10 @@ jobs:
     strategy:
       matrix:
         include:
-        - image: ubuntu-20.04
+        - image: ubuntu-2004-paid
           target: linux
           arch: amd64
-        - image: ubuntu-20.04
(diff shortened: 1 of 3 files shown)


.github/workflows/release.yml
denyeart, line 1: Thanks @ryjones , I didn't realize this would also update to latest ubuntu. At least for the release jobs in this file, I think we need to keep them on ubuntu-20.04 to ensure the built binaries and images work on ubuntu-20.04 and higher.

For the other checks I think moving to latest will be fine, it will be interesting to see if our unit test and integration test flake failures get better on the faster runners (the problems got worse after shifting from azure pipelines to github actions, likely due to differences in run times).

ryjones: @denyeart I rolled back to a new 20-04 builder for the jobs that were using 20.04 on purpose

denyeart (approved):

PR #4169, opened 2023-04-18
Update dependencies (release-2.2)

denyeart:
Update dependencies to match the known good levels in release-2.5. This will simplify maintenance and support between the two LTS releases.


denyeart:
#4171 removes the last problematic commit.

PR #4176, opened 2023-04-19
Re-add Go to fabric-tools image

denyeart:
Since the images are no longer based on the golang alpine images, need to manually install go to the fabric-tools image. go is needed for users that use fabric-tools image to package go chaincodes.


andrew-coleman (approved):

denyeart:

Issue #4184, opened 2023-04-27
Orderer v3: remove system channel: cleanup the onboarding package


tock-ibm:
Without the system channel, the orderer/common/onboarding package is no longer needed.

The only method that is still used from the package is:

`func ValidateBootstrapBlock(block *common.Block, bccsp bccsp.BCCSP) error`

This method can be moved somewhere else (possibly `orderer/common/server/util.go`, or `main.go`) , and all the rest of the package deleted.

Some code still reference the `InactiveChainRegistry` / `InactiveChanReplicator`, However in the current code all these references are always nil, as the InactiveChanReplicator is never instantiated un the current code.

method 
```
func initializeEtcdraftConsenter(consenters map[string]consensus.Consenter, conf *localconfig.TopLevel, lf blockledger.Factory, clusterDialer *cluster.PredicateDialer, bootstrapBlock *cb.Block, ri *onboarding.ReplicationInitiator, srvConf comm.ServerConfig, srv *comm.GRPCServer, registrar *multichannel.Registrar, metricsProvider metrics.Provider, bccsp bccsp.BCCSP) 
```
in `main.go`  L632 can be removed, as it is used only in test, and the respective test  
```
func TestInitializeEtcdraftConsenter(t *testing.T)
```
in `main_test.go` L941 can be removed as well.

PR #4191, opened 2023-04-30
cleanup of the onboarding package + removed initializeEtcdraftConsenter

semil:
1. Cleaned dependencies on the onboarding package, so the package could be deleted.
2. Removed initializeEtcdraftConsenter

#### Type of change

- Test update

#### Description

As described in issue #4184.

#### Related issues

Issue: #4184.


orderer/common/server/util.go
```diff
@@ -27,3 +31,31 @@ func createLedgerFactory(conf *config.TopLevel, metricsProvider metrics.Provider
 	}
 	return lf, nil
 }
+
+// ValidateBootstrapBlock returns whether this block can be used as a bootstrap block.
+// A bootstrap block is a block of a system channel, and needs to have a ConsortiumsConfig.
+func ValidateBootstrapBlock(block *common.Block, bccsp bccsp.BCCSP) error {
```
yacovm: why is this still needed?

tock-ibm: To check that there is no system channel in chains and join blocks when the server starts... did that a few weeks ago.

semil: I just added a fix - that the code will be using the new ValidateBootstrapBlock instead of the one in the onboarding package.

yacovm: But do we still use the one in the onboarding package after this change? if no then we can remove it no?

semil: Yes we can, @tock-ibm correct me if I'm wrong. 
Do you want to do the removal of the package in the scope of this PR?

orderer/common/server/util_test.go
```diff
@@ -56,3 +66,90 @@ func TestCreateLedgerFactory(t *testing.T) {
...
+		"-outputBlock", genesisBlockPath,
+		"--configPath", tempDir,
+	)
+	configtxgenProcess, err := gexec.Start(cmd, nil, nil)
+	gt.Expect(err).NotTo(HaveOccurred())
+	gt.Eventually(configtxgenProcess, time.Minute).Should(gexec.Exit(0))
+	gt.Expect(configtxgenProcess.Err).To(gbytes.Say("Writing genesis block"))
+
+	return genesisBlockPath
+}
```
tock-ibm: This is one way to generate a genesis block, but we don't need it in this package, as you can use 
```
// produces a system channel genesis file to make sure the server detects it and refuses to start
func produceGenesisFileEtcdRaft(t *testing.T, channelID string, tmpDir string) (string, []byte)
```
which already exists in this file

orderer/common/server/main_test.go
```diff
@@ -55,11 +55,18 @@ type signerSerializer interface {
...
 
 func TestMain(m *testing.M) {
 	var err error
+
+	configtxgen, err = gexec.Build("github.com/hyperledger/fabric/cmd/configtxgen")
+	if err != nil {
+		fmt.Fprintf(os.Stderr, "configtxgen build failed: %v", err)
+		os.Exit(-1)
+	}
+
```
tock-ibm: We will not need this if use `func produceGenesisFileEtcdRaft(...)` instead of `generateBootstrapBlock`, which can be removed as well.

orderer/common/server/main_test.go
```diff
@@ -55,11 +55,18 @@ type signerSerializer interface {
 	identity.SignerSerializer
 }
 
-// the path to cryptogen, which can be used by tests to create certificates
-var cryptogen, tempDir string
+// The path to cryptogen, configtxgen which can be used by tests to create certificates
+var cryptogen, configtxgen string
```
tock-ibm: We can use an alternative to `configtxgen`, see below

orderer/common/server/main.go
```diff
@@ -252,7 +251,7 @@ func verifyNoSystemChannelJoinBlock(config *localconfig.TopLevel, cryptoProvider
 		if err != nil {
 			logger.Panicf("Failed unmarshalling join-block for channel '%s', error: %v", channelName, err)
 		}
-		if err = onboarding.ValidateBootstrapBlock(block, cryptoProvider); err == nil {
+		if err = ValidateBootstrapBlock(block, cryptoProvider); err == nil {
```
tock-ibm: you can change that method to package scope, i.e. ValidateBootstrapBlock -> validateBootstrapBlock

orderer/common/server/util.go
tock-ibm: Yes ;-)

orderer/common/server/util_test.go
```diff
@@ -7,13 +7,23 @@ SPDX-License-Identifier: Apache-2.0
...
+	"os/exec"
 	"path/filepath"
 	"testing"
+	"time"
+
+	"github.com/golang/protobuf/proto"
+	"github.com/hyperledger/fabric-protos-go/common"
+	"github.com/hyperledger/fabric/bccsp/sw"
+	"github.com/onsi/gomega/gbytes"
+	"github.com/onsi/gomega/gexec"
```
tock-ibm: remove all empty lines and do goimports again

orderer/common/server/util_test.go
tock-ibm: You can also use this one:
```
func produceGenesisFileEtcdRaftAppChannel(...)
```

tock-ibm (approved):

Issue #4199, opened 2023-05-05
Deadlock on etcd raft orderer chain apply change on orderer startup


vodsej:
### Description

This issue was found on orderer version 2.5.0 is apparently caused by pull request #3919
Deadlock can occur on orderer startup while applying raft configuration changes from previous blocks.

Deadlock occures as follows:

1. [node](https://github.com/hyperledger/fabric/blob/v2.5.0/orderer/consensus/etcdraft/node.go) run function receives `n.Ready()` signal and emits apply change signal (line 156 - `n.chain.applyC <- apply{rd.CommittedEntries, rd.SoftState}`)
2. [chain](https://github.com/hyperledger/fabric/blob/v2.5.0/orderer/consensus/etcdraft/chain.go) run function receives `c.applyC ` signal and start processing received changes with apply function (for loop in run function is blocked by this).
3. As the signal is received by chain, node run for loop is no longer blocked and can proceed and finish the loop iteration and wait for another signal.
4. chain apply function is processing ents and needs to sync WAL. After mentioned pull request it does so by emitting signal to node ( line 1153 - `c.Node.storage.WALSyncC <- struct{}{}` ).  Processing of ents is blocked until signal is received by node. This means that for loop in chain run function for receiving signals is blocked as well.
5. node run function uses select to receive waiting signals. But the selection is random, when multiple signals are waiting. In this situation there are `n.Ready()` `n.storage.WALSyncC` signals waiting for receive. In this scenario `n.Ready()` is selected.
6. Another apply change signal is then emitted  (line 156 - `n.chain.applyC <- apply{rd.CommittedEntries, rd.SoftState}`)
7. We are now in deadlock. node run for loop is blocked by waiting for `n.chain.applyC <- apply{rd.CommittedEntries, rd.SoftState}` receive and chain run for loop is blocked by not finnished apply function, which is waiting for receive of signal `c.Node.storage.WALSyncC <- struct{}{}`

The most simple fix would be to rollback mentioned pull request as it does not introduce new functionality and previous solution worked without any issue. I managed to verify it on local deployment. When v2.5.0 always produces this deadlock, rollbacking this pull request solved deadlock issue.

yacovm:
I think you're right, thanks for the report and the analysis.

[...]

PR #4201, opened 2023-05-06
Restrict WAL usage to node.go

yacovm:
See https://github.com/hyperledger/fabric/issues/4199

### DIFF
--- a/orderer/consensus/etcdraft/chain.go
+++ b/orderer/consensus/etcdraft/chain.go
@@ -1179,10 +1179,12 @@ func (c *Chain) apply(ents []raftpb.Entry) {
 				continue
 			}
 
-			// persist the WAL entries into disk
-			c.Node.storage.WALSyncC <- struct{}{}
+			if err := c.Node.storage.Sync(); err != nil {
+				c.logger.Debugf("Failed to sync raft log, error: %s", err)
+			}
 
 			c.confState = *c.Node.ApplyConfChange(cc)
+
 			switch cc.Type {
 			case raftpb.ConfChangeAddNode:
 				c.logger.Infof("Applied config change to add node %d, current nodes in channel: %+v", cc.NodeID, c.confState.Voters)
--- a/orderer/consensus/etcdraft/chain.go
+++ b/orderer/consensus/etcdraft/chain.go
@@ -1179,10 +1179,6 @@ func (c *Chain) apply(ents []raftpb.Entry) {
 				continue
 			}
 
-			if err := c.Node.storage.Sync(); err != nil {
-				c.logger.Debugf("Failed to sync raft log, error: %s", err)
-			}
-
 			c.confState = *c.Node.ApplyConfChange(cc)
 
 			switch cc.Type {
--- a/orderer/consensus/etcdraft/node.go
+++ b/orderer/consensus/etcdraft/node.go
@@ -171,11 +171,6 @@ func (n *node) run(campaign bool) {
 			// to the followers and them writing to their disks. Check 10.2.1 in thesis
 			n.send(rd.Messages)
 
-		case <-n.storage.WALSyncC:
-			if err := n.storage.Sync(); err != nil {
-				n.logger.Warnf("Failed to sync raft log, error: %s", err)
-			}
-
 		case <-n.chain.haltC:
 			raftTicker.Stop()
 			n.Stop()
--- a/orderer/consensus/etcdraft/node.go
+++ b/orderer/consensus/etcdraft/node.go
@@ -153,6 +153,7 @@ func (n *node) run(campaign bool) {
 
 			// skip empty apply
 			if len(rd.CommittedEntries) != 0 || rd.SoftState != nil {
+				n.maybeSyncWAL(rd.CommittedEntries)
 				n.chain.applyC <- apply{rd.CommittedEntries, rd.SoftState}
 			}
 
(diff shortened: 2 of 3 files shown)


denyeart:
@tock-ibm Could you review this one?  We'll want to backport it to release-2.5 as well.

tock-ibm:
> @tock-ibm Could you review this one? We'll want to backport it to release-2.5 as well.

I will do it

tock-ibm:
Is there a way to add a test case (unit, or integration) that will trigger the problem? 
Such that we'll know its failing in the existing code and fixed after this PR?

@yacovm @vodsej any suggestions?

yacovm:
> Is there a way to add a test case (unit, or integration) that will trigger the problem? Such that we'll know its failing in the existing code and fixed after this PR?
> 
> @yacovm @vodsej any suggestions?

You want to add a test that attempts to make the Raft chain deadlock? Usually unit tests are written to test a specific functional scenario such that if the implementation changes, the functional scenario still works as expected.

I'm not sure what's the point in writing a test that attempts to deadlock the two goroutines if the fix is to move the WAL sync to a single goroutine as was in the past.

I think it's possible to write such a test since the raft.Node is an interface and can be mocked, but do you not agree with the report that the deadlock may indeed occur?

tock-ibm:
> > Is there a way to add a test case (unit, or integration) that will trigger the problem? Such that we'll know its failing in the existing code and fixed after this PR?
> > @yacovm @vodsej any suggestions?
> 
> You want to add a test that attempts to make the Raft chain deadlock? Usually unit tests are written to test a specific functional scenario such that if the implementation changes, the functional scenario still works as expected.
> 
> I'm not sure what's the point in writing a test that attempts to deadlock the two goroutines if the fix is to move the WAL sync to a single goroutine as was in the past.
> 
> I think it's possible to write such a test since the raft.Node is an interface and can be mocked, but do you not agree with the report that the deadlock may indeed occur?

When there is a bug, it is good practice to reproduce the behavior in a test that fails on the existing code and passes on the fixed code. Some times it is very difficult, some times it is easy. The question is, how difficult is it in this case?
The reasoning behind the behavior is excellent, and it seems that the fix is doing its  job. However, we thought that the previous change was fine too, so some extra testing could add more confidence, in my opinion.

I'll leave it up to you to decide whether you can recreate the conditions in the report, and whether it is worth the effort.

vodsej:
I'm not so well versed in fabric codebase, but I would say it would take quite a lot of effort to write a test, that would truly verify this issue. Unfortunately I won't be any help here.

yacovm:
> > > Is there a way to add a test case (unit, or integration) that will trigger the problem? Such that we'll know its failing in the existing code and fixed after this PR?
> > > @yacovm @vodsej any suggestions?
> > 
> > 
> > You want to add a test that attempts to make the Raft chain deadlock? Usually unit tests are written to test a specific functional scenario such that if the implementation changes, the functional scenario still works as expected.
> > I'm not sure what's the point in writing a test that attempts to deadlock the two goroutines if the fix is to move the WAL sync to a single goroutine as was in the past.
> > I think it's possible to write such a test since the raft.Node is an interface and can be mocked, but do you not agree with the report that the deadlock may indeed occur?
> 
> When there is a bug, it is good practice to reproduce the behavior in a test that fails on the existing code and passes on the fixed code. Some times it is very difficult, some times it is easy. The question is, how difficult is it in this case? The reasoning behind the behavior is excellent, and it seems that the fix is doing its job. However, we thought that the previous change was fine too, so some extra testing could add more confidence, in my opinion.
> 
> I'll leave it up to you to decide whether you can recreate the conditions in the report, and whether it is worth the effort.

I think it's possible to try at least. 

@semil do you want to take a look at this after https://github.com/hyperledger/fabric/issues/4228 ?

tock-ibm:
@denyeart Do you think we should wait for the test before merging or merge the fix and add the test in a separate issue?

denyeart:
@tock-ibm Typically we would want a test with the PR. If the automated test is not practical in a short timeframe, I'd at least want to see that the reviewer (you in this case) has confidence in the fix and approves the PR. For example, when I review a PR that is difficult to test in CI, I will at least try to run the PR code and try to test the logic myself. Sometimes this requires modifying the code in my local environment to get the node into the appropriate state for the test and/or adding local log statements to prove out the fix. I often use the test-network-nano-bash sample for such trials, since it is simple to update code, recompile node, and spin it up.

Is that something that is feasible here?

tock-ibm:
> @tock-ibm Typically we would want a test with the PR. If the automated test is not practical in a short timeframe, I'd at least want to see that the reviewer (you in this case) has confidence in the fix and approves the PR. For example, when I review a PR that is difficult to test in CI, I will at least try to run the PR code and try to test the logic myself. Sometimes this requires modifying the code in my local environment to get the node into the appropriate state for the test and/or adding local log statements to prove out the fix. I often use the test-network-nano-bash sample for such trials, since it is simple to update code, recompile node, and spin it up.
> 
> Is that something that is feasible here?

@denyeart @yacovm 
I am working with @semil to define a unit test that cover the scenario described in the issue. This is the proposed flow:
- start a unit test with 3 nodes
- stop (not remove) 1 node
- add a 4th node
- order some messages 
- restart the stopped node

This has the potential to hit the problematic part. The restarted node will get the config change entry from the Ready channel and trigger the flow described in the issue. Whether a deadlock will happen is a probabilistic. It will probably be necessary to run this a few times to recreate the issue, and add some printouts at relevant places in the code.

semil:
Please look at #4262 for the test.

tock-ibm (approved):

tock-ibm:
@yacovm @denyeart @vodsej 

A unit test by @semil in #4262 recreates the bug on main, before this fix, and does not fail when applied on top of this fix. I am going to merge this one, and @semil will push the test in a separate PR.

denyeart:

PR #4207, opened 2023-05-07
system channel cleanup - remove replicator

semil:
Cleaned dependencies on system channel from the `orderer/common/cluster/replication.go`.

#### Type of change

- Improvement (improvement to code, performance, etc)
- Test update

#### Description

As described in issue #4202 .

#### Related issues

Issue: #4202 .

Issue #4202, opened 2023-05-07
Orderer v3: Remove system channel from orderer/common/cluster, remove replicator


tock-ibm:
The Replicator replicated chains by detecting them using the system channel.
```
// Replicator replicates chains
type Replicator struct {
	DoNotPanicIfClusterNotReachable bool
	Filter                          ChannelPredicate
	SystemChannel                   string
	ChannelLister                   ChannelLister
	Logger                          *flogging.FabricLogger
	Puller                          *BlockPuller
	BootBlock                       *common.Block
	AmIPartOfChannel                SelfMembershipPredicate
	LedgerFactory                   LedgerFactory
}
```
See `orderer/common/cluster/replication.go` L90.

It is no longer used. It should be removed. 
Note that the file has errors, interfaces, function definitions, as well as auxiliary functions that are used by other components. for example:
```
type SelfMembershipPredicate func(configBlock *common.Block) error
```
and 
```
func LatestHeightAndEndpoint(puller ChainPuller) (string, uint64, error)
```

So care must be taken to track the usage of each element and move it accordingly.

In addition, some interface are use to create mocks that are then used by testing code (with go:generate). So one needs to go over the existing mocks and check what is needed and what not, and avoid removing the generate statements.

PR #4210, opened 2023-05-08
system channel cleanup - remove replicator

semil:
Cleaned dependencies on system channel from the `orderer/common/cluster/replication.go`.

#### Type of change

- Test update

#### Description

As described in issue #4202 .

#### Related issues

Issue: #4202 .
Related PR: #4207 .


orderer/common/cluster/util.go
```diff
@@ -48,6 +48,17 @@ type ConnByCertMap map[string]*grpc.ClientConn
 // Lable used for TLS Export Keying Material call
 const KeyingMaterialLabel = "orderer v3 authentication label"
 
+//go:generate mockery --dir . --name LedgerWriter --case underscore --output mocks/
```
yacovm: I don't think they belong to util. These are definitions.

Let's put them in the file they're used at. which is probably deliver.go.

orderer/common/cluster/util.go
```diff
@@ -208,6 +219,17 @@ type Dialer interface {
 
 var errNotAConfig = errors.New("not a config block")
 
+// ErrForbidden denotes that an ordering node refuses sending blocks due to access control.
+var ErrForbidden = errors.New("forbidden pulling the channel")
```
yacovm: these belong to deliver.go, not in util

yacovm: these belong to deliver.go, not in util

orderer/common/cluster/util.go
semil: LedgerWriter is only used by LedgerInterceptor which is in the same file.
Am I missing something?

https://github.com/hyperledger/fabric/blob/c31a8a23ae414a1209c9019bf51120d03a21bded/orderer/common/cluster/util.go#L588

orderer/common/cluster/util.go
semil: moved all the error defs :)

orderer/common/cluster/util.go
yacovm: You're not, I didn't look at the code when reviewing. 

But now when I look at the code (at least, one version of it...)
Is this ledger interceptor in use after removing onboarding?

semil: You are correct, the LedgerWriter is used by LedgerInterceptor which is only by the test, removed both.

yacovm (approved):

Issue #4019, opened 2023-02-13
Orderer v3: prevent creating a system channel with channel participation


tock-ibm:
- [x] When channel participation receives a join block, it should check that this block does not define a system channel by verifying that the `Consortiums` field in nil.
- [x] Add an integration test to validate this scenario.

tock-ibm:
done

tock-ibm:
done

PR #4231, opened 2023-05-22
test that a join block does not define a system channel by verifying consortiums

MayRosenbaum:
#### Type of change

- Test update

#### Description

Addition of an integration test to validate that when channel participation receives a join block, this block does not define a system channel

#### Related issues

issue: #4019

### DIFF
--- a/integration/raft/channel_participation_test.go
+++ b/integration/raft/channel_participation_test.go
@@ -13,6 +13,7 @@ import (
 	"encoding/pem"
 	"fmt"
 	"io/ioutil"
+	"math"
 	"net/http"
 	"os"
 	"path/filepath"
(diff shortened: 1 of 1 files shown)


integration/raft/channel_participation_test.go
```diff
@@ -1109,3 +1115,48 @@ func multiNodeEtcdRaftTwoChannels() *nwo.Config {
...
+func createJoinBlockDefineSystemChannel(channelID string) *common.Block {
+	return &common.Block{
+		Data: &common.BlockData{
+			Data: [][]byte{
+				protoutil.MarshalOrPanic(&common.Envelope{
+					Payload: protoutil.MarshalOrPanic(&common.Payload{
+						Data: protoutil.MarshalOrPanic(&common.ConfigEnvelope{
+							Config: &common.Config{
+								ChannelGroup: &common.ConfigGroup{
+									Groups: map[string]*common.ConfigGroup{
```
yacovm, line 1128: do we really need all these values (hashing algorithm, orderer addresses, etc.) ?

yacovm (approved):

Issue #4233, opened 2023-05-23
Method AppRootCAsByChain() not thread safe


tock-ibm:
### Description

in `func (p *Peer) updateTrustedRoots(cm channelconfig.Resources)`, this part:
```
for _, roots := range p.CredentialSupport.AppRootCAsByChain() {
		trustedRoots = append(trustedRoots, roots...)
	}
```
is using  `p.CredentialSupport.AppRootCAsByChain()` in an un-thread safe manner, as that function takes a lock and returns a reference to the underlying protected map. 
The map is read, but might change at the same time.

### Solution

The serialization to an array (all trusted roots) should have been done within the lock. The AppRootCAsByChain() should not expose a reference to the internal protected object.

tock-ibm:
@MayRosenbaum can you take care of this?

tock-ibm:
Fixed

PR #4235, opened 2023-05-24
Method AppRootCAsByChain() not thread safe

MayRosenbaum:
#### Type of change

- Bug fix

#### Description

As described in issue #4233 

#### Related issues

issue: #4233

### DIFF
--- a/internal/pkg/comm/connection.go
+++ b/internal/pkg/comm/connection.go
@@ -75,10 +75,15 @@ func (cs *CredentialSupport) GetPeerCredentials() credentials.TransportCredentia
 	})
 }
 
-func (cs *CredentialSupport) AppRootCAsByChain() map[string][][]byte {
+func (cs *CredentialSupport) AppRootCAsByChain() [][]byte {
 	cs.mutex.RLock()
 	defer cs.mutex.RUnlock()
-	return cs.appRootCAsByChain
+	trustedRoots := make([][]byte, 0, len(cs.appRootCAsByChain))
+	// iterate over all roots for all app and orderer channels
+	for _, roots := range cs.appRootCAsByChain {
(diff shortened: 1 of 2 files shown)


internal/pkg/comm/connection.go
```diff
@@ -75,10 +75,18 @@ func (cs *CredentialSupport) GetPeerCredentials() credentials.TransportCredentia
...
 	defer cs.mutex.RUnlock()
-	return cs.appRootCAsByChain
+	var trustedRoots [][]byte
+	// iterate over all roots for all app and orderer channels
+	for _, roots := range cs.appRootCAsByChain {
+		trustedRoots = append(trustedRoots, roots...)
+	}
+
+	trustedRootsCopy := make([][]byte, len(trustedRoots))
+	copy(trustedRootsCopy, trustedRoots)
```
C0rWin:
```suggestion
	trustedRoots := make([][]byte, 0, len(cs.appRootCAsByChain))
	for _, roots := range cs.appRootCAsByChain {
		trustedRoots = append(trustedRoots, roots...)
	}

	return append([][]byte(nil), trustedRoots...)
}

Comment (C0rWin) [reply internal/pkg/comm/connection.go]: I think you can make it a little bit simpler.

Comment (C0rWin) [reply internal/pkg/comm/connection.go]: or even 

```
	trustedRoots := make([][]byte, 0, len(cs.appRootCAsByChain))
	for _, roots := range cs.appRootCAsByChain {
		trustedRoots = append(trustedRoots, roots...)
	}

	return trustedRoots
```

MayRosenbaum: Fixed to the second option

C0rWin (approved):

Issue #4211, opened 2023-05-09
Orderer v3: configtxgen - document system channel no longer supported


tock-ibm:
Document that the system channel is no longer supported.
Leave all capabilities and APIs as is.
Possibly - emit an error warning that system channel config block and channel creation TXs are not supported in v3.

tock-ibm:
done

PR #4238, opened 2023-05-28
system channel cleanup - configtxgen document system channel no longer supported

semil:
configtxgen - document system channel no longer supported

#### Type of change

- Test update
- Documentation update

#### Related issues

Issue: #4211 .

### DIFF
--- a/cmd/configtxgen/main.go
+++ b/cmd/configtxgen/main.go
@@ -41,7 +41,7 @@ func doOutputBlock(config *genesisconfig.Profile, channelID string, outputBlock 
 		return errors.New("refusing to generate block which is missing orderer section")
 	}
 	if config.Consortiums != nil {
-		logger.Info("Creating system channel genesis block")
+		logger.Error("Warning: 'Consortiums' should be nil since system channel is no longer supported in Fabric v3.x")
 	} else {
 		if config.Application == nil {
 			return errors.New("refusing to generate application channel block which is missing application section")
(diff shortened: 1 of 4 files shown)


docs/source/commands/configtxgen.md
```diff
@@ -20,7 +20,7 @@ Usage of configtxgen:
   -asOrg string
     	Performs the config generation as a particular organization (by name), only including values in the write set that org (likely) has privilege to set
   -channelCreateTxBaseProfile string
-    	Specifies a profile to consider as the orderer system channel current state to allow modification of non-application parameters during channel create tx generation. Only valid in conjunction with 'outputCreateChannelTx'.
+    	[DEPRECATED] Specifies a profile to consider as the orderer system channel current state to allow modification of non-application parameters during channel create tx generation. Only valid in conjunction with 'outputCreateChannelTx'.
```
yacovm, line 23: If this is deprecated why not remove it altogether?

sampleconfig/configtx.yaml
```diff
@@ -3,6 +3,16 @@
 # SPDX-License-Identifier: Apache-2.0
 #
 
+
```
yacovm, line 6: I don't understand why we need these comments in the `configtx.yaml`. 

Can you explain?

cmd/configtxgen/main.go
```diff
@@ -4,6 +4,14 @@ Copyright IBM Corp. 2017 All Rights Reserved.
 SPDX-License-Identifier: Apache-2.0
 */
 
+/**********************************************************************************
```
yacovm: Why put this comment in `main.go`?

cmd/configtxgen/main.go
```diff
@@ -220,7 +228,7 @@ func main() {
 	flag.StringVar(&outputBlock, "outputBlock", "", "The path to write the genesis block to (if set)")
 	flag.StringVar(&channelID, "channelID", "", "The channel ID to use in the configtx")
 	flag.StringVar(&outputChannelCreateTx, "outputCreateChannelTx", "", "The path to write a channel creation configtx to (if set)")
-	flag.StringVar(&channelCreateTxBaseProfile, "channelCreateTxBaseProfile", "", "Specifies a profile to consider as the orderer system channel current state to allow modification of non-application parameters during channel create tx generation. Only valid in conjunction with 'outputCreateChannelTx'.")
+	flag.StringVar(&channelCreateTxBaseProfile, "channelCreateTxBaseProfile", "", "[DEPRECATED] Specifies a profile to consider as the orderer system channel current state to allow modification of non-application parameters during channel create tx generation. Only valid in conjunction with 'outputCreateChannelTx'.")
```
yacovm, line 223: I think we can remove this flag and we also need to run the below script to update the docs:

https://github.com/hyperledger/fabric/blob/main/scripts/help_docs.sh

docs/source/commands/configtxgen.md
tock-ibm: I think we want to maintain backwards compatibility? i.e. be able to manage with the new version of the tool old deployments?

cmd/configtxgen/main.go
tock-ibm: I think we want to maintain backwards compatibility? i.e. be able to manage with the new version of the tool old deployments?

If this is not the case we can do a much more drastic pruning, but the idea initially was to be conservative and do a minimal effort change.

cmd/configtxgen/main.go
```diff
@@ -41,7 +49,7 @@ func doOutputBlock(config *genesisconfig.Profile, channelID string, outputBlock
 		return errors.New("refusing to generate block which is missing orderer section")
 	}
 	if config.Consortiums != nil {
-		logger.Info("Creating system channel genesis block")
+		logger.Error("'Consortiums' must be nil since system channel is no longer supported")
```
tock-ibm: The message should be 
"Warning: 'Consortiums' should be nil since system channel is no longer supported in Fabric v3.x"

sampleconfig/configtx.yaml
```diff
@@ -258,8 +268,9 @@ Application: &ApplicationDefaults
 Orderer: &OrdererDefaults
 
     # Orderer Type: The orderer implementation to start.
-    # Available types are "solo", "kafka" and "etcdraft".
-    OrdererType: solo
+    # Available types are "solo", "kafka", "etcdraft" and "BFT".
+    # Please note that "solo" and "kafka" are deprecated.
```
tock-ibm:
```
# Available types are "etcdraft" and "BFT".
# Please note that "solo" and "kafka" are no longer supported.
```

sampleconfig/configtx.yaml
```diff
@@ -3,6 +3,16 @@
 # SPDX-License-Identifier: Apache-2.0
 #
 
+
+################################################################################
+#
+#   NOTE
+#
+#   Since release V3.0 Solo, Kafka and the System Channel have been removed
+#   and no longer supported.longer supported.
```
tock-ibm: and are no longer supported.

cmd/configtxgen/main.go
tock-ibm: `outputCreateChannelTx` this flag is also deprecated, we no longer need channel creation TXs in v3

tock-ibm: also the inspect CreateChannelTx, and the respective entries in the *.md file below. everything about channel create TX is deprecated.

sampleconfig/configtx.yaml
semil: I added it to warn developers from using deprecated profiles (that use Solo, Kafka...)
Do you think it is not needed?

cmd/configtxgen/main.go
semil: You are correct, it is not needed.

sampleconfig/configtx.yaml
semil: Changed that, please note that there are profiles using Kafka and Solo in there.

tock-ibm: Do a text search on the fabric code base and find out which of those solo and kafka profiles are not in use, and delete them. I think there should not be any kafka profile in use. Solo might be found here and there in unit tests, it will take some time to clean that up (in another PR).

tock-ibm (approved):

Issue #4228, opened 2023-05-17
Add a check for V3 style signature - BFT consensus


arkadiPiven:
### Current Status

When running the test-network-nano-bash sample from the latest fabric binaries with BFT consensus I encountered a problem with configtx.yaml - V3 is not set to true explicitly in the configtx.yaml then the MSPId received by the peer from the orderer is empty. This is because the BFT orderer supports the V3 style signature which doesn't
 hold the certificate of the orderer that signed in the payload of the message, as a contrast to the V2.X style signature.
The current status is that Fabric doesn't have any check in the code that warns what kind of style is supported.
<img width="1042" alt="Screenshot 2023-05-17 at 16 23 25" src="https://github.com/hyperledger/fabric/assets/48690203/7247a5c8-2631-43f0-aca8-ab24d9d43b92">

### Expected

Add a check for when a channel is created,
If using V3 style signature then nothing printed,
If using V2.X style signature then print a warning.

### Solution

Couple of solution are valid,

1. If we're using a profile with orderer type BFT then we can print a warning log if the channel capabilities are set to V2.X when the first orderer joins the channel.
2. When running the command to create the genesis block we can print a warning log if the channel capabilities are set to V2.X in the configtx.yaml.

### Please let us know if you plan to work on this.

Maybe.

yacovm:
Possible solutions could be: 

1. Not allow a BFT orderer to join a channel without the V3 capabilities enabled - we need to ensure that we don't commit the genesis block before creating the chain instance.
2. Make configtxgen reject creating a BFT genesis block if V3 capabilities are not set

I think (2) is easiest.

yacovm:
@semil do you want to take a look at this?

semil:
@yacovm yes, on it :)

PR #4239, opened 2023-05-28
configtxgen - add a check for V3 style signature - BFT consensus

semil:
Add a check for V3 style signature - BFT consensus

#### Type of change

#### Related issues

Issue: #4228 .

### DIFF
--- a/internal/configtxgen/encoder/encoder.go
+++ b/internal/configtxgen/encoder/encoder.go
@@ -154,7 +154,7 @@ func NewChannelGroup(conf *genesisconfig.Profile) (*cb.ConfigGroup, error) {
 
 	var err error
 	if conf.Orderer != nil {
-		channelGroup.Groups[channelconfig.OrdererGroupKey], err = NewOrdererGroup(conf.Orderer)
+		channelGroup.Groups[channelconfig.OrdererGroupKey], err = NewOrdererGroup(conf.Orderer, conf.Capabilities)
 		if err != nil {
 			return nil, errors.Wrap(err, "could not create orderer group")
 		}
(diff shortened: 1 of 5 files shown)


C0rWin:
@semil can you add a test to validate that the check you have just added works as expected?

cmd/configtxgen/main.go
```diff
@@ -32,6 +32,13 @@ import (
 var logger = flogging.MustGetLogger("common.tools.configtxgen")
 
 func doOutputBlock(config *genesisconfig.Profile, channelID string, outputBlock string) error {
+	if config.Orderer != nil && config.Orderer.OrdererType == "BFT" {
```
yacovm: I don't think this is the right place to put it. 

It should be deeper into the stack. 

This is the `main.go`.  It should not contain validation logic. 

There is validation logic in `common/channelconfig/orderer.go`, maybe we can add it there?

```

func (oc *OrdererConfig) Validate() error {
	for _, validator := range []func() error{
		oc.validateBatchSize,
		oc.validateBatchTimeout,
	} {
		if err := validator(); err != nil {
			return err
		}
	}

	return nil
}
```

sampleconfig/configtx.yaml
```diff
@@ -664,6 +664,7 @@ Profiles:
     # that uses the etcd/raft-based orderer.
     SampleAppChannelSmartBft:
         <<: *ChannelDefaults
+        Consortium: SampleConsortium
```
yacovm, line 680: why did we add this to the sampleconfig?

yacovm (changes requested):
Please see comments.

internal/configtxgen/encoder/encoder.go
```diff
@@ -181,7 +181,13 @@ func NewChannelGroup(conf *genesisconfig.Profile) (*cb.ConfigGroup, error) {
 // NewOrdererGroup returns the orderer component of the channel configuration.  It defines parameters of the ordering service
 // about how large blocks should be, how frequently they should be emitted, etc. as well as the organizations of the ordering network.
 // It sets the mod_policy of all elements to "Admins".  This group is always present in any channel configuration.
-func NewOrdererGroup(conf *genesisconfig.Orderer) (*cb.ConfigGroup, error) {
+func NewOrdererGroup(conf *genesisconfig.Orderer, channelCapabilities map[string]bool) (*cb.ConfigGroup, error) {
+	if conf.OrdererType == "BFT" {
+		v3Enabled, v3Exists := channelCapabilities["V3_0"]
```
yacovm: `Capabilities` is of type `map[string]bool` so you can just do: 

```
if conf.OrdererType == "BFT" && ! channelCapabilities["V3_0"]` {
        return nil, errors.New("orderer type BFT must be used with V3_0 capability")
}

```

Because the "zero value" of bool is false, so if it's missing or false, `channelCapabilities["V3_0"]` evaluates to false.

semil: correct!
(I thought that if you check a missing key like that it will raise an error)

yacovm (dismissed):

yacovm:
```

2023-06-05T13:33:43.2006536Z [1685971554] [1mConfig Suite[0m - 2/2 specs [38;5;10m•[0m
2023-06-05T13:33:43.2007052Z [38;5;243m------------------------------[0m
2023-06-05T13:33:43.2007381Z [38;5;9m• [FAILED] [0.104 seconds][0m
2023-06-05T13:33:43.2008018Z [0mConfigTx ConsenterMapping [38;5;243m[38;5;9m[1m[It] generates an application channel genesis block and checks it contains the ConsenterMapping[0m[0m
2023-06-05T13:33:43.2010612Z [38;5;243m/home/runner/work/fabric/fabric/integration/configtx/consenter_mapping_test.go:22[0m
2023-06-05T13:33:43.2010989Z 
```

yacovm (approved):

PR #4256, opened 2023-06-02
Update CouchDB doc references to 'stable'

denyeart:
Instead of maintaining the exact version of CouchDB docs and risk getting outdated, it is safe to reference the 'stable' version of the CouchDB docs since the doc content in these topics has indeed been stable.


ale-linux (approved):

denyeart:

PR #4263, opened 2023-06-05
Raft WAL deadlock test

semil:
- There was a bug described at issue #4201. 
- The bug caused a deadlock on etcd raft orderer chain apply change on orderer startup.
- Issue #4201 fixed the issue. 
- We wanted to add a test can check this scenario with high probability (because of the multiple go routines, and how the code is built, this can't be 100% test.)
- We simulated the scenario of the deadlock using a test and debug logs (on the unfixed code) in #4262.
- This PR adds the same test but without the debug logs.

#### Type of change

- Bug fix
- Test update

#### Description

Please look at #4199, PR #4201 and PR #4262.

### DIFF
--- a/orderer/mocks/common/blockcutter/blockcutter.go
+++ b/orderer/mocks/common/blockcutter/blockcutter.go
@@ -17,17 +17,17 @@ var logger = flogging.MustGetLogger("orderer.mocks.common.blockcutter")
 
 // Receiver mocks the blockcutter.Receiver interface
 type Receiver struct {
-	// IsolatedTx causes Ordered returns [][]{curBatch, []{newTx}}, false when set to true
-	IsolatedTx bool
+	// isolatedTx causes Ordered returns [][]{curBatch, []{newTx}}, false when set to true
+	isolatedTx bool
 
-	// CutAncestors causes Ordered returns [][]{curBatch}, true when set to true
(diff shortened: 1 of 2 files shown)


orderer/mocks/common/blockcutter/blockcutter.go
```diff
@@ -103,3 +103,39 @@ func (mbc *Receiver) CurBatch() []*cb.Envelope {
 	defer mbc.mutex.Unlock()
 	return mbc.curBatch
 }
+
+// SetIsolatedTx is used to change IsolatedTx field in order to avoid data race in the tests
+// If the field is changed directly from the test, a data race is created between the thread using the block cutter
+// and the test thread setting the field's value.
```
tock-ibm: You don't need to explain why we did it in the comment, but in the issue / PR comments / commit message.

Here, just
// SetIsolatedTx is used to change the IsolatedTx field in a thread safe manner.

Moreover, since those fields should never be accessed from the outside, change their scope, i.e. `IsolatedTx  -> isolatedTx`
 
Same for all other fields.

semil: done!

tock-ibm (approved):

denyeart:

Issue #4264, opened 2023-06-06
Test flake in follower package


tock-ibm:
### Description

```
2023-06-06 06:07:57.555 UTC 0001 INFO [common.tools.configtxgen.localconfig] completeInitialization -> orderer type: etcdraft
2023-06-06 06:07:57.559 UTC 0002 INFO [common.tools.configtxgen.localconfig] Load -> Loaded configuration: /home/runner/work/fabric/fabric/sampleconfig/configtx.yaml
2023-06-06 06:07:57.659 UTC 0003 INFO [common.tools.configtxgen.localconfig] completeInitialization -> orderer type: etcdraft
2023-06-06 06:07:57.659 UTC 0004 INFO [common.tools.configtxgen.localconfig] Load -> Loaded configuration: /home/runner/work/fabric/fabric/sampleconfig/configtx.yaml
2023-06-06 06:07:57.807 UTC 0005 INFO [common.tools.configtxgen.localconfig] completeInitialization -> orderer type: etcdraft
2023-06-06 06:07:57.808 UTC 0006 INFO [common.tools.configtxgen.localconfig] Load -> Loaded configuration: /home/runner/work/fabric/fabric/sampleconfig/configtx.yaml
2023-06-06 06:07:57.949 UTC 0007 INFO [common.tools.configtxgen.localconfig] completeInitialization -> orderer type: etcdraft
2023-06-06 06:07:57.950 UTC 0008 INFO [common.tools.configtxgen.localconfig] Load -> Loaded configuration: /home/runner/work/fabric/fabric/sampleconfig/configtx.yaml
2023-06-06 06:07:57.987 UTC 0009 INFO [follower.test] NewChain -> Created with join-block number: 10, ledger height: 0 channel=my-channel
2023-06-06 06:07:57.987 UTC 000a INFO [follower.test] NewChain -> Created with join-block number: 10, ledger height: 0 channel=my-channel
2023-06-06 06:07:57.987 UTC 000b INFO [follower.test] Start -> Started channel=my-channel
2023-06-06 06:07:57.987 UTC 000c INFO [follower.test] halt -> Stopped channel=my-channel
2023-06-06 06:07:57.988 UTC 000d WARN [follower.test] run -> Pull failed, error: failed to pull up to join block: chain stopped channel=my-channel
2023-06-06 06:07:57.989 UTC 000e INFO [follower.test] NewChain -> Created with a nil join-block, ledger height: 5 channel=my-channel
[...]

PR #4265, opened 2023-06-06
Follower: check if join block different from fetched block

tock-ibm:
Change-Id: I7089df4f1bfc200a5f17582df0da7d4172f8f09b

#### Type of change

- Bug fix

#### Description

Follower: revert check of join block different from fetched block

#### Related issues

Addresses #4264

### DIFF
--- a/orderer/common/follower/follower_chain.go
+++ b/orderer/common/follower/follower_chain.go
@@ -12,7 +12,6 @@ import (
 	"time"
 
 	"github.com/golang/protobuf/proto"
-
 	"github.com/hyperledger/fabric-protos-go/common"
 	"github.com/hyperledger/fabric/bccsp"
 	"github.com/hyperledger/fabric/common/flogging"
@@ -297,7 +296,7 @@ func (c *Chain) run() {
 	}()
 
 	if err := c.pull(); err != nil {
-		c.logger.Warnf("Pull failed, error: %s", err)
+		c.logger.Warnf("Pull failed, follower chain stopped, error: %s", err)
 		// TODO set the status to StatusError (see FAB-18106)
 	}
 }
@@ -345,10 +344,6 @@ func (c *Chain) pull() error {
 			return errors.WithMessage(err, "failed to pull up to join block")
 		}
 		c.logger.Info("Onboarding finished successfully, pulled blocks up to join-block")
-	}
-
-	if c.joinBlock != nil && !proto.Equal(c.ledgerResources.Block(c.joinBlock.Header.Number).Data, c.joinBlock.Data) {
-		c.logger.Panicf("Join block (%d) we pulled mismatches block we joined with", c.joinBlock.Header.Number)
 	}
 
 	err = c.pullAfterJoin()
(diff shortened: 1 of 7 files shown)


yacovm (changes requested):
I don't think that simply removing the check without providing a countermeasure in the same fix is a good idea. 

With this change, the code allows an orderer to pull blocks while disregarding the join block, which obviously was retrieved from somewhere. 

The genesis block cannot be validated, and thus the current code completely puts the validity of the genesis block in the mercy of the orderer it is retrieved from. 

This check lets you know that you have diverged from the chain.

tock-ibm:
> I don't think that simply removing the check without providing a countermeasure in the same fix is a good idea.
> 
> With this change, the code allows an orderer to pull blocks while disregarding the join block, which obviously was retrieved from somewhere.
> 
> The genesis block cannot be validated, and thus the current code completely puts the validity of the genesis block in the mercy of the orderer it is retrieved from.
> 
> This check lets you know that you have diverged from the chain.

@yacovm I agree with what you said. I am changing the mechanism such that:
The follower checks if join block different from fetched block, both header and data, and
- it happens before fetched block is committed, otherwise if the orderer is restarted, the follower continues past that block
- don't panic, since it can affect other channels
- continue to retry, as it might have been an error in a single remote orderer

In the future, we may consider to stop the chain in events like this and reflect it in the status, as proposed in FAB-18106
https://jira.hyperledger.org/browse/FAB-18106

I also added a couple of unit tests for this scenario.

orderer/common/follower/follower_chain.go
```diff
@@ -527,11 +523,24 @@ func (c *Chain) pullUntilTarget(targetHeight uint64, updateEndpoints bool) (uint
...
+				return n, errors.Errorf("block header previous hash mismatch on sequence %d, expected %x, got %x",
 					nextBlock.Header.Number, actualPrevHash, reportedPrevHash)
 			}
+
+			if c.joinBlock != nil && c.joinBlock.Header.Number == nextBlock.Header.Number {
+				if !proto.Equal(nextBlock.Header, c.joinBlock.Header) {
+					c.logger.Errorf("Block header mismatch between the block we pulled and the block we joined with, sequence %d", c.joinBlock.Header.Number)
+					return n, errors.Errorf("block header mismatch between the block we pulled and the block we joined with, sequence %d", c.joinBlock.Header.Number)
+				}
+				if !proto.Equal(nextBlock.Data, c.joinBlock.Data) {
```
yacovm: why do we need the data check? The block puller assures us the data is the hash in the header, no?

yacovm: or are you trying to catch a case if we got a join block with a header hash that mismatches its data? if that's the reason then we should check it earlier and not here.

yacovm:
> > I don't think that simply removing the check without providing a countermeasure in the same fix is a good idea.
> > With this change, the code allows an orderer to pull blocks while disregarding the join block, which obviously was retrieved from somewhere.
> > The genesis block cannot be validated, and thus the current code completely puts the validity of the genesis block in the mercy of the orderer it is retrieved from.
> > This check lets you know that you have diverged from the chain.
> 
> @yacovm I agree with what you said. I am changing the mechanism such that: The follower checks if join block different from fetched block, both header and data, and
> 
>     * it happens before fetched block is committed, otherwise if the orderer is restarted, the follower continues past that block
> 
>     * don't panic, since it can affect other channels
> 
>     * continue to retry, as it might have been an error in a single remote orderer
> 
> 

> 
> I also added a couple of unit tests for this scenario.

As long as the follower doesn't evolve into a fully fledged consenter, we're good.

orderer/common/follower/follower_chain.go
tock-ibm: In the fetched block yes, but the join block can be anything... I am not sure we even verify the data hash equals the hash(data) when we accept it via the API... we should do it though

yacovm: That's what I'm saying  - we should verify the data matches the hash when we digest the block at the time of processing the API call, not **here**.

yacovm (dismissed):

yacovm (approved):

PR #4279, opened 2023-06-13
More verbose logs for etcdraft unit tests

tock-ibm:
Change-Id: Id6905e22f2826158150c8c36152dee1d9cf992dc

#### Type of change

#### Description

More verbose logs for etcdraft unit tests will make it easier to track down flakes


yacovm:
I ran it with your change, I am not sure what changed... can you explain in which way does it make it more verbose?

tock-ibm:
> I ran it with your change, I am not sure what changed... can you explain in which way does it make it more verbose?

When the tests pass, there is no change. When it fails, it is easier to find which test case failed. In the integration tests output we have it, but in the raft unit tests the ginkgo output is not divided into testcases by default.

yacovm (approved):

PR #4289, opened 2023-06-19
BFT config doc

HagarMeir:
Add BFT config documentation 

#### Type of change

- Documentation update

### DIFF
--- a/docs/source/bft_configuration.md
+++ b/docs/source/bft_configuration.md
@@ -1,0 +1,221 @@
+# Configuring and operating a BFT ordering service
+
+**Audience**: *BFT ordering service admins*
+
+## Conceptual overview
+
+For a high level overview of the concept of ordering and how the supported
+ordering service implementations (including BFT) work at a high level, check
+out our conceptual documentation on the [Ordering Service](./orderer/ordering_service.md).
+
+To learn about the process of setting up an ordering node, check out our
+documentation on [Planning for an ordering service](./deployorderer/ordererplan.html).
+
+## Configuration
+
+A BFT cluster is configured in two places:
+
+* **Local configuration**: Governs node specific aspects, such as TLS
+  communication, replication behavior, and file storage.
+
+* **Channel configuration**: Defines the membership of the  BFT cluster for the
+  corresponding channel, as well as protocol specific parameters such as timeouts.
+
+Unlike the [Raft ordering service](./raft_configuration.md), where nodes identify each other using TLS pinning, 
+BFT nodes identify each other using their enrollment certificate.
+
+Each channel has its own instance of a BFT protocol running. Thus, each
+BFT node must be referenced in the configuration of each channel it participates in
+by adding its enrollment certificate (in `PEM` format) and MSP ID to the channel
+config. 
+
+The following [section](https://github.com/hyperledger/fabric/blob/main/sampleconfig/configtx.yaml) from `configtx.yaml` shows four BFT nodes (also called
+“consenters”) in the channel:
+
+```
+    ConsenterMapping:
+    - ID: 1
+      Host: bft0.example.com
+      Port: 7050
+      MSPID: OrdererOrg1
+      Identity: /path/to/identity
+      ClientTLSCert: path/to/ClientTLSCert0
+      ServerTLSCert: path/to/ServerTLSCert0
+    - ID: 2
+      Host: bft1.example.com
+      Port: 7050
+      MSPID: OrdererOrg2
+      Identity: /path/to/identity
+      ClientTLSCert: path/to/ClientTLSCert1
+      ServerTLSCert: path/to/ServerTLSCert1
+    - ID: 3
+      Host: bft2.example.com
+      Port: 7050
+      MSPID: OrdererOrg3
+      Identity: /path/to/identity
+      ClientTLSCert: path/to/ClientTLSCert2
+      ServerTLSCert: path/to/ServerTLSCert2
+    - ID: 4
+      Host: bft3.example.com
+      Port: 7050
+      MSPID: OrdererOrg4
+      Identity: /path/to/identity
+      ClientTLSCert: path/to/ClientTLSCert3
+      ServerTLSCert: path/to/ServerTLSCert3
+```
+
+When the channel config block is created, the `configtxgen` tool reads the paths
+to the identities (enrollment certificates) and TLS certificates, and replaces the paths with the corresponding bytes of
+the identities and certificates.
+Note that the `Identity` refers to the path of the enrollment certificate, and not to the entire MSP directory.
+
+Note: it is possible to remove or add an ordering node to a channel dynamically, a process described in the reconfiguration section below and in more detail in the [reconfiguration tutorial](./create_channel/add_orderer.md).
+
+In addition, note that the [channel capabilities](./capabilities_concept.md) V3.0 flag must be set to true.
+
+### Local configuration
+
+The `orderer.yaml` has two configuration sections that are relevant for BFT
+orderers:
+
+**Cluster**, which determines the communication configuration, and
+**Consensus**, which determines BFT protocol related configuration.
+
+**Cluster parameters:**
+
+By default, the BFT service is running in the same gRPC server as the client
+facing gRPC API (which is used for transaction submission or block retrieval), but it can be
+configured to have a separate gRPC server with a separate port.
+
+This is useful for cases where you want TLS certificates issued by the
+organizational CAs, but used only by the cluster nodes to communicate among each
+other, and TLS certificates issued by a public TLS CA for the client facing API.
+
+* `ClientCertificate`, `ClientPrivateKey`: The file path of the client TLS certificate
+  and corresponding private key.
+* `ListenPort`: The port the cluster listens on.
+  It must be same as `consenters[i].Port` in Channel configuration.
+  If blank, the port is the same port as the orderer general port (`general.listenPort`)
+* `ListenAddress`: The address the cluster service is listening on.
+* `ServerCertificate`, `ServerPrivateKey`: The TLS server certificate key pair
+  which is used when the cluster service is running on a separate gRPC server
+  (different port).
+
+Note: `ListenPort`, `ListenAddress`, `ServerCertificate`, `ServerPrivateKey` must
+be either set together or unset together.
+If they are unset, they are inherited from the general TLS section,
+in example `general.tls.{privateKey, certificate}`.
+When general TLS is disabled:
+- Use a different `ListenPort` than the orderer general port
+- Properly configure TLS root CAs in the channel configuration.
+
+Currently, if the cluster communication uses a separate listener, then mutual TLS authentication is implicitly enforced,
+while if the cluster communication uses the same gRPC server as the client facing gRPC API, it is not implicitly enforced.
+
+There are also hidden configuration parameters for `general.cluster` which can be
+used to further fine tune the cluster communication or replication mechanisms:
+
+* `SendBufferSize`: Regulates the number of messages in the egress buffer.
+* `DialTimeout`, `RPCTimeout`: Specify the timeouts of creating connections and
+  establishing streams.
+* `ReplicationBufferSize`: the maximum number of bytes that can be allocated
+  for each in-memory buffer used for block replication from other cluster nodes.
+  Each channel has its own memory buffer. Defaults to `20971520` which is `20MB`.
+* `PullTimeout`: the maximum duration the ordering node will wait for a block
+  to be received before it aborts. Defaults to five seconds.
+* `ReplicationRetryTimeout`: The maximum duration the ordering node will wait
+  between two consecutive attempts. Defaults to five seconds.
+* `TLSHandshakeTimeShift`: If the TLS certificates of the ordering nodes
+  expire and are not replaced in time (see TLS certificate rotation below),
+  communication between them cannot be established, and it will be impossible
+  to send new transactions to the ordering service.
+  To recover from such a scenario, it is possible to make TLS handshakes
+  between ordering nodes consider the time to be shifted backwards a given
+  amount that is configured to `TLSHandshakeTimeShift`.
+  This setting only applies when a separate cluster listener is in use.  If
+  the cluster service is sharing the orderer's main gRPC server, then instead
+  specify `TLSHandshakeTimeShift` in the `General.TLS` section.
+
+**Consensus parameters:**
+
+* `WALDir`: the location at which Write Ahead Logs for BFT are stored.
+  Each channel will have its own subdirectory named after the channel ID.
+
+### Channel configuration
+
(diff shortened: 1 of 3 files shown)


docs/source/bft_configuration.md
```diff
@@ -0,0 +1,221 @@
...
+* `ViewChangeResendInterval`: Defines the interval after which the ViewChange message is resent.
+* `ViewChangeTimeout`: Is started when a node first receives a quorum of `ViewChange` messages, and defines the interval after which the node will try to initiate a view change with a higher view number.
+* `LeaderHeartbeatTimeout`: Is the interval after which, if nodes do not receive a "sign of life" from the leader, they complain about the current leader and try to initiate a view change. A sign of life is either a heartbeat or a message from the leader.
+* `LeaderHeartbeatCount`: Is the number of heartbeats per `LeaderHeartbeatTimeout` that the leader should emit. The heartbeat-interval is equal to: `LeaderHeartbeatTimeout/LeaderHeartbeatCount`.
+* `CollectTimeout`: Is the interval after which the node stops listening to `StateTransferResponse` messages, stops collecting information about view metadata from remote nodes.
+
+## Reconfiguration
+
+The Smart BFT orderer supports dynamic (meaning, while the channel is being serviced) addition and removal of nodes, and configuration changes.
+The only configuration parameter which cannot be changed is the `RequestPoolSize`.
```
yacovm: cannot be changed *dynamically*. We should be able to change it with a restart, no?

docs/source/bft_configuration.md
```diff
@@ -0,0 +1,221 @@
...
+
+Note that your cluster must be operational and able to achieve consensus before you attempt to reconfigure it.
+As a rule, you should never attempt any configuration changes to the Smart BFT consenters, such as adding or removing a consenter, or rotating a consenter's certificate, unless all consenters are online and healthy.
```
yacovm: unless it's a removal of a consenter that is known to be offline

docs/source/bft_configuration.md
```diff
@@ -0,0 +1,221 @@
...
+
+If you do decide to change these parameters, it is recommended to only attempt such a change during a maintenance cycle. 
+Problems are most likely to occur when a reconfiguration is attempted in clusters with only a few nodes while a node is down. 
+For example, if you have four nodes in your consenters set and one of them is down, it means you have three out of four nodes alive. 
+If you extend the cluster to five nodes while in this state, you will have only three out of five nodes alive, which is not a quorum.
+The fifth node won't be able to onboard because nodes can only onboard to functioning clusters (unless the total size of the cluster less than four).
```
yacovm: I don't think this is true for BFT. We can onboard even if the cluster has no consensus because of:
```
func (c *BFTChain) Errored() <-chan struct{} {
	// TODO: Implement Errored
	return nil
}
``` 

in `chain.go`

docs/source/bft_configuration.md
```diff
@@ -0,0 +1,221 @@
+# Configuring and operating a Smart BFT ordering service
+
+**Audience**: *Smart BFT ordering node admins*
```
yacovm: change smartBFT  to BFT. SmartBFT is only an implementation. throughout the document please.

docs/source/bft_configuration.md
HagarMeir: So what should I write instead?

docs/source/bft_configuration.md
HagarMeir: But Raft is also an implementation for example
And there are Smart BFT specific config here such as in the channel configuration section

docs/source/bft_configuration.md
```diff
@@ -0,0 +1,228 @@
...
+For a high level overview of the concept of ordering and how the supported
+ordering service implementations (including BFT) work at a high level, check
+out our conceptual documentation on the [Ordering Service](./orderer/ordering_service.md).
+
+To learn about the process of setting up an ordering node, check out our
+documentation on [Planning for an ordering service](./deployorderer/ordererplan.html).
+
+## Configuration
+
+A  BFT cluster is configured in two places:
```
denyeart:
```suggestion
A BFT cluster is configured in two places:
```

docs/source/bft_configuration.md
```diff
@@ -0,0 +1,228 @@
...
+
+* **Local configuration**: Governs node specific aspects, such as TLS
+  communication, replication behavior, and file storage.
+
+* **Channel configuration**: Defines the membership of the  BFT cluster for the
```
denyeart, line 21:
```suggestion
* **Channel configuration**: Defines the membership of the BFT cluster for the
```

docs/source/bft_configuration.md
```diff
@@ -0,0 +1,228 @@
+# Configuring and operating a BFT ordering service
```
denyeart, line 1: This file needs to be added to ops_guide.rst so that it shows up in the doc table of contents, just like raft_configuration.md is.

docs/source/bft_configuration.md
```diff
@@ -0,0 +1,228 @@
...
+A  BFT cluster is configured in two places:
+
+* **Local configuration**: Governs node specific aspects, such as TLS
+  communication, replication behavior, and file storage.
+
+* **Channel configuration**: Defines the membership of the  BFT cluster for the
+  corresponding channel, as well as protocol specific parameters such as timeouts.
+
+Unlike the [Raft ordering service](./raft_configuration.md), where nodes identify each other using TLS pinning, 
+BFT nodes identify each other using their enrollment certificate's private key.
```
denyeart: The private key? Doesn't the private key remain private... do you mean public key here?

docs/source/bft_configuration.md
```diff
@@ -0,0 +1,228 @@
...
+The following [section](https://github.com/hyperledger/fabric/blob/main/sampleconfig/configtx.yaml) from `configtx.yaml` shows four BFT nodes (also called
+“consenters”) in the channel:
+
+```
+    ConsenterMapping:
+    - ID: 1
+      Host: bft0.example.com
+      Port: 7050
+      MSPID: OrdererOrg1
+      Identity: /path/to/identity
```
denyeart, line 41: In some places Fabric config needs to refer to the node's /msp directory, in other places Fabric config needs to refer to the certificate that contains the public key. I believe in ConsenterMapping it is the latter. This should be clarified both in the sample configtx.yaml and this doc. This could be achieved with a yaml comment, e.g. specifying /msp/signcerts/orderer.example.com-cert.pem (to take the test-network example).

Really, this should be called `IdentityCert` or `EnrollmentCert` , to make it clear and to be consistent with the `ServerTLSCert` naming below.

docs/source/bft_configuration.md
```diff
@@ -0,0 +1,228 @@
...
+
+In addition, note that the [channel capabilities](./capabilities_concept.md) V3.0 flag must be set to true.
+
+### Local configuration
+
+The `orderer.yaml` has two configuration sections that are relevant for BFT
+orderers:
+
+**Cluster**, which determines the communication configuration, and
+**consensus**, which determines BFT protocol related configuration.
```
denyeart:
```suggestion
**Consensus**, which determines BFT protocol related configuration.
```

(this change should be made in raft_configuration.md as well)

docs/source/bft_configuration.md
```diff
@@ -0,0 +1,228 @@
...
+Note: `ListenPort`, `ListenAddress`, `ServerCertificate`, `ServerPrivateKey` must
+be either set together or unset together.
+If they are unset, they are inherited from the general TLS section,
+in example `general.tls.{privateKey, certificate}`.
+When general TLS is disabled:
+- Use a different `ListenPort` than the orderer general port
+- Properly configure TLS root CAs in the channel configuration.
+
+There are also hidden configuration parameters for `general.cluster` which can be
+used to further fine tune the cluster communication or replication mechanisms:
```
denyeart, line 116: Are these all really identical to the Raft config? If so, nice job making it consistent!

docs/source/bft_configuration.md
```diff
@@ -0,0 +1,228 @@
...
+  amount that is configured to `TLSHandshakeTimeShift`.
+  This setting only applies when a separate cluster listener is in use.  If
+  the cluster service is sharing the orderer's main gRPC server, then instead
+  specify `TLSHandshakeTimeShift` in the `General.TLS` section.
+
+**Consensus parameters:**
+
+* `WALDir`: the location at which Write Ahead Logs for BFT are stored.
+  Each channel will have its own subdirectory named after the channel ID.
+
```
denyeart, line 143: Just to confirm, the Raft SnapDir, EvictionSuspicion, and TickIntervalOverride don't apply to BFT?

docs/source/bft_configuration.md
```diff
@@ -0,0 +1,228 @@
...
+        # BlockValidation specifies what signatures must be included in the block
+        # from the orderer for the peer to validate it.
+        BlockValidation:
+            Type: ImplicitMeta
+            Rule: "ANY Writers"
+```
+
+In BFT, the [configtxgen](./commands/configtxgen.md) tool encodes a policy that is suitable for BFT.
+However, when adding or removing ordering service nodes from the channel,
+the policy should be adjusted accordingly.
```
denyeart: This seems important, but not enough information is provided.
What policy is "suitable for BFT"?
What adjustment should be made when adding ordering service nodes?
A conceptual description could be provided, and then you could point to the sample as an example - 
fabric-samples/test-network/scripts/add_new_orderer_to_config.py

docs/source/orderer/ordering_service.md
```diff
@@ -185,7 +185,17 @@ between ordering service nodes.
...
+* **BFT** 
+  New as of v3.0, the Byzantine Fault Tolerant (BFT) ordering service, as its name implies,
+  can withstand not only crash failures, but also a subset of nodes behaving maliciously.
+  An example of malicious behavior is a leader node sending different blocks with the same sequence number,
+  hence creating a fork in the blockchain. An ordering service that is only Crash Fault Tolerant (CFT) 
+  can not protect against such a behavior. 
+  A Byzantine Fault Tolerant ordering service can withstand up and not including to a third of its nodes misbehaving
+  (either crashing or deviating from the protocol).
+  More information about the BFT protocol can be found in its corresponding [paper](https://arxiv.org/abs/2107.06922). 
+
```
denyeart, line 197: This is a good intro. For Raft, this document also has a decent amount of conceptual information about the protocol. Are we intending to add similar conceptual information about SmartBFT?

docs/source/bft_configuration.md
yacovm: correct. SmartBFT has no snap dir, since it piggybacks snapshots on Fabric blocks, because in SmartBFT, the WAL may garbage collect itself every Fabric block.

docs/source/bft_configuration.md
yacovm: There is no change in the cluster configuration. The only change is in the authentication mechanism which is not configurable anyway.

docs/source/bft_configuration.md
yacovm: I don't get it, what's the difference? these two lines look the same to me.

docs/source/bft_configuration.md
yacovm: ok can we do it after it's merged?

docs/source/bft_configuration.md
yacovm: The enrollment certificate's private key, contains the public key. I guess it's better to say "using the enrollment certificate".

docs/source/bft_configuration.md
yacovm: Right now, this is an enrollment certificate. In the future there may be a different kind of identity that is not based on x509 certificates. The name selection is to be generic.

docs/source/orderer/ordering_service.md
yacovm: can  you be more specific?

docs/source/bft_configuration.md
yacovm: added

docs/source/bft_configuration.md
denyeart: The config section is capital C Consensus.

docs/source/orderer/ordering_service.md
denyeart: Take a look at https://hyperledger-fabric.readthedocs.io/en/latest/orderer/ordering_service.html#raft. The bottom half of that document provides a conceptual description of how Raft works in the Fabric ordering service. This is missing for SmartBFT.

docs/source/bft_configuration.md
denyeart: Seems a little silly to do a whole separate PR to add it, might as well add it when the content is added, but if you have a reason and want to add it later, fine by me.

docs/source/bft_configuration.md
denyeart: Ok, so we'll leave the name as-is, but we still need to clarify that this points to the actual pem certificate file, rather than the msp directory as is common in other Fabric configs. A comment would suffice.

docs/source/bft_configuration.md
```diff
@@ -0,0 +1,223 @@
...
+        # from the orderer for the peer to validate it.
+        BlockValidation:
+            Type: ImplicitMeta
+            Rule: "ANY Writers"
+```
+
+In BFT, the [configtxgen](./commands/configtxgen.md) tool encodes a policy that is suitable for BFT.
+It automatically derives the policy from the nodes configured in the `ConsenterMapping` section.
+However, when adding or removing ordering service nodes from the channel,
+the policy should be adjusted accordingly as described in the reconfiguration [tutorial](https://github.com/hyperledger/fabric/blob/main/docs/source/create_channel/add_orderer.md#3-add-the-orderer-to-the-policy-rules).
```
denyeart:
```suggestion
the policy should be adjusted accordingly as described in the reconfiguration [tutorial](./create_channel/add_orderer.html#add-the-orderer-to-the-policy-rules).
```

We use relative links so that it works regardless of the branch name, this should do it.

docs/source/orderer/ordering_service.md
yacovm: OK i will add this later :)

docs/source/bft_configuration.md
yacovm: done

docs/source/bft_configuration.md
yacovm: done

docs/source/bft_configuration.md
yacovm: done

docs/source/bft_configuration.md
```diff
@@ -0,0 +1,224 @@
+# Configuring and operating a BFT ordering service
+
+**Audience**: * BFT ordering service admins*
```
denyeart:
```suggestion
**Audience**: *BFT ordering service admins*
```

The extra space caused a formatting issue.

denyeart (approved):

PR #4292, opened 2023-06-21
BFT Block Puller: isolate common subcomponents

tock-ibm:
Change-Id: I274d8883add91fac08da5202199d1bac53e99adb

#### Type of change

- New feature

#### Description
- Isolate subcomponents from the CFT Deliverer that can be used in the BFT Deliverer
- A skeleton of the BFT Deliverer (no censurship monitor, no tests)

#### Related issues
#4293


yacovm:
why is this in internal though? This can be useful for other projects.

tock-ibm:
> why is this in internal though? This can be useful for other projects.

Correct. The plan is to put the reusable components in `common/deliverclient`, as `common/deliver` already contains the server components. This will be done in one of the future commits. See https://github.com/hyperledger/fabric/issues/4350

internal/pkg/peer/blocksprovider/bft_deliverer.go
```diff
@@ -0,0 +1,263 @@
...
+
+	"github.com/hyperledger/fabric-protos-go/orderer"
+	"github.com/hyperledger/fabric/common/flogging"
+	"github.com/hyperledger/fabric/internal/pkg/identity"
+	"github.com/hyperledger/fabric/internal/pkg/peer/orderers"
+	"github.com/pkg/errors"
+)
+
+// BFTDeliverer TODO this is a skeleton
+type BFTDeliverer struct { // TODO
```
yacovm, line 22: why are we coupling gossip stuff with block delivery? 

Let's create a self contained component that only fetches blocks and is not aware of what is gossip, yielding leadership, etc.

internal/pkg/peer/blocksprovider/block_receiver.go
```diff
@@ -0,0 +1,171 @@
...
+	}
+
+	// cancel the sending side and wait for the start goroutine to exit
+	br.cancelSendFunc()
+	<-br.recvC
+
+	return err
+}
+
+func (br *BlockReceiver) processMsg(msg *orderer.DeliverResponse) (uint64, error) {
```
yacovm, line 127: please let's not do it this way. 
We need a component we can re-use for orderer too. 

Here, we couple relaying the block to gossip with retrieving it from the orderer.

internal/pkg/peer/blocksprovider/util.go
```diff
@@ -0,0 +1,80 @@
...
+type errFatal struct {
+	message string
+}
+
+func (e *errFatal) Error() string {
+	return e.message
+}
+
+func backOffDuration(base float64, exponent uint, minDur, maxDur time.Duration) time.Duration {
+	if base < 1.0 {
```
yacovm, line 49: base is generic but you set it to a constant where you call it.
Why not just pass the constant all the way here and avoid this check? 

and why use nano second resolution? Why not just use milliseconds resolution?

internal/pkg/peer/blocksprovider/bft_deliverer.go
```diff
@@ -0,0 +1,263 @@
...
+	}
+
+	// clean up everything because we are closing
+	d.mutex.Lock()
+	defer d.mutex.Unlock()
+	d.blockReceiver.Stop()
+	// TODO stop the monitor
+}
+
+func (d *BFTDeliverer) refreshSources() {
```
yacovm, line 155: This is a question / proposition, not a request. 

As you know, when this thing is deployed on WAN, the remote node you pull blocks from, pays for egress traffic.

Would it not make sense to introduce a preference to pull blocks from your own organization? 

It would also make sense to do so censorship wise.

Why not for simplicity, start the endpoint when we initialize the structure to be the endpoint of your own org?

internal/pkg/peer/blocksprovider/bft_deliverer.go
```diff
@@ -0,0 +1,263 @@
...
+			}
+			// TODO can it be nil?
+		}
+	}
+
+	// clean up everything because we are closing
+	d.mutex.Lock()
+	defer d.mutex.Unlock()
+	d.blockReceiver.Stop()
+	// TODO stop the monitor
```
yacovm, line 152: why stop the monitor from here?

internal/pkg/peer/blocksprovider/block_receiver.go
```diff
@@ -0,0 +1,149 @@
...
+	switch t := msg.GetType().(type) {
+	case *orderer.DeliverResponse_Status:
+		if t.Status == common.Status_SUCCESS {
+			return 0, errors.Errorf("received success for a seek that should never complete")
+		}
+
+		return 0, errors.Errorf("received bad status %v from orderer", t.Status)
+	case *orderer.DeliverResponse_Block:
+		blockNum := t.Block.Header.Number
+		if err := br.blockVerifier.VerifyBlock(gossipcommon.ChannelID(br.channelID), blockNum, t.Block); err != nil {
```
yacovm, line 137: how do we deal a corner case where we ingest a config block and then verify the successive block before we have time to ingest the config block?

internal/pkg/peer/blocksprovider/deliverer.go
```diff
@@ -142,196 +165,74 @@ func (d *Deliverer) DeliverBlocks() {
...
-				break RecvLoop
-			}
+		d.mutex.Lock()
+		blockReceiver := &BlockReceiver{
+			channelID:      d.ChannelID,
+			blockHandler:   d.BlockHandler,
+			blockVerifier:  d.BlockVerifier,
+			deliverClient:  deliverClient,
+			cancelSendFunc: cancel,
+			recvC:          make(chan *orderer.DeliverResponse),
```
yacovm, line 195: this is a non buffered channel. so we ingest a block one by one. 

In gossip there is a buffer, but in the orderer there isn't one, so this is not efficient to pull blocks one by one and commit them one by one.

internal/pkg/peer/blocksprovider/bft_deliverer.go
tock-ibm: Done

internal/pkg/peer/blocksprovider/block_receiver.go
tock-ibm: Done

internal/pkg/peer/blocksprovider/bft_deliverer.go
tock-ibm: This could be an interesting idea to develop in the long run. However, right now, we don't have Org info in this particular data structure, and the peer org may be completely different from all ordering orgs, with no obvious ways of determining who is better to pull from. 
I prefer to keep a simple randomized + round-robin policy and evaluate later on whether an improvement is needed.

This idea could be revisited late when we deal with the "connection source" data structure: https://github.com/hyperledger/fabric/issues/4347

internal/pkg/peer/blocksprovider/block_receiver.go
tock-ibm: Right now, we don't. This will be the subject of a future PR where the deliverer hold its own copy of the most current config. See issue: https://github.com/hyperledger/fabric/issues/4346

internal/pkg/peer/blocksprovider/deliverer.go
tock-ibm: The plan is to have a buffer at the orderer as well, a buffer that gets all the blocks taken from this channel. The follower/chain goroutines then pull blocks from that buffer, not directly from the `recvC` channel shown here.

internal/pkg/peer/blocksprovider/util.go
tock-ibm: The base may be different in different places. For example in the old CFT code it is 1.2 and In the new BFT code it is 2.0. 
This function is flexible enough to cope with different bases.
Nanos-seconds is the underlying resolution of time.Duration (int64 in ns). I see no reason the decrease the resolution.

internal/pkg/peer/blocksprovider/bft_deliverer.go
tock-ibm: The BFTDeliverer is work in progress. It will be delt with here: https://github.com/hyperledger/fabric/issues/4348

yacovm (approved):

Issue #4261, opened 2023-06-04
BFT Block Puller: Remove redundant internal map in deliveryServiceImpl.blockproviders


tock-ibm:
The gossip service:
- It contains a map of channelID -> DeliverService (an interface).
- The DeliveryService is  implemented by a deliveryServiceImpl, which contains blockProviders a map of channelID -> Deliverer.
- The implementation shows that the external and internal maps are always keyed together, with the same channelID, which means that the internal map has always a single entry...

Remove the redundant internal map.

tock-ibm:
Done

tock-ibm:
There is a bug:

After the delivery service StopDeliveryForChannel is called, it should be possible to call StartDeliveryForChannel again.
After the delivery service Stop is called, it should not be possible to call <Start/Stop>DeliveryForChannel again.

This was introduced by https://github.com/hyperledger/fabric/pull/4260

PR #4297, opened 2023-06-22
BFT Block Puller: Stop deliver service correctly

tock-ibm:
Change-Id: Icd1725913e3be913b3628b3f37e252c8e13a0461

#### Type of change
- Bug fix

#### Description

After the delivery service StopDeliveryForChannel is called, it should be possible to call StartDeliveryForChannel  again.
After the delivery service Stop is called, it should not be possible to call <Start/Stop>DeliveryForChannel  again.

This was introduced by #4260

#### Related issues

#4261

### DIFF
--- a/core/deliverservice/deliveryclient.go
+++ b/core/deliverservice/deliveryclient.go
@@ -37,8 +37,12 @@ type DeliverService interface {
 	StartDeliverForChannel(chainID string, ledgerInfo blocksprovider.LedgerInfo, finalizer func()) error
 
 	// StopDeliverForChannel dynamically stops delivery of new blocks from ordering service
-	// to channel peers.
+	// to channel peers. StartDeliverForChannel can be called again, and delivery will resume.
 	StopDeliverForChannel() error
+
+	// Stop terminates delivery service and closes the connection. Marks the service as stopped, meaning that
+	// StartDeliverForChannel cannot be called again.
+	Stop()
 }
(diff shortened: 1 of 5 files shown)


core/deliverservice/deliveryclient.go
```diff
@@ -223,10 +227,13 @@ func (d *deliverServiceImpl) StopDeliverForChannel() error {
 		return errors.New(errMsg)
 	}
 
-	d.stopping = true
```
yacovm, line 226: did we remove this and not re-add it?

tock-ibm: In the previous PR I made an error and added this flag here, in the wrong place, into the `StopDeliverForChannel()`, and now I am removing it, keeping it only in the right place, which is the `Stop()`. 

Compare it to commit e8f817a418b6aef14a5592011b2ebf9d3427122d right before the removal of the redundant mapping. There, the flag if not there. Basically I re-instated the used of  `Stop` & `StopDeliverForChannel` to what it was at that commit.

I also added a test that verifies we can do `StartDeliverForChannel -> StopDeliverForChannel -> StartDeliverForChannel`.

tock-ibm: As you can see I also re-added the `Stop` into the interface.

yacovm (approved):

PR #4303, opened 2023-06-30
Ledger block cache

yacovm:
This commit introduces an in-memory cache for the block storage of the ledger. It caches new blocks that are committed and assumes blocks are committed in-order and with consecutive sequences.

The block iterators now attempt to retrieve the blocks from the cache if possible before going to the block storage.

The intent is twofold:

1) Speedup the block Deliver API by not doing disk I/O when clients (peers, orderers) fetch blocks. 

2) Reduce the impact of the deliver API on the performance of writing new blocks into the ledger.

### DIFF
--- a/common/ledger/blkstorage/blockfile_mgr.go
+++ b/common/ledger/blkstorage/blockfile_mgr.go
@@ -27,6 +27,7 @@ const (
 	blockfilePrefix                   = "blockfile_"
 	bootstrappingSnapshotInfoFile     = "bootstrappingSnapshot.info"
 	bootstrappingSnapshotInfoTempFile = "bootstrappingSnapshotTemp.info"
+	defaultBlockCacheSizeBytes        = 1024 * 1024 * 50
 )
 
 var blkMgrInfoKey = []byte("blkMgrInfo")
@@ -41,6 +42,7 @@ type blockfileMgr struct {
 	blkfilesInfoCond          *sync.Cond
 	currentFileWriter         *blockfileWriter
 	bcInfo                    atomic.Value
+	cache                     *cache
 }
 
 /*
@@ -96,7 +98,7 @@ func newBlockfileMgr(id string, conf *Conf, indexConfig *IndexConfig, indexStore
 	if err != nil {
 		panic(fmt.Sprintf("Error creating block storage root dir [%s]: %s", rootDir, err))
 	}
-	mgr := &blockfileMgr{rootDir: rootDir, conf: conf, db: indexStore}
+	mgr := &blockfileMgr{rootDir: rootDir, conf: conf, db: indexStore, cache: newCache(defaultMaxBlockfileSize)}
 
 	blockfilesInfo, err := mgr.loadBlkfilesInfo()
 	if err != nil {
@@ -327,6 +329,8 @@ func (mgr *blockfileMgr) addBlock(block *common.Block) error {
 		}
 		return errors.WithMessage(err, "error appending block to file")
 	}
+
+	defer mgr.cache.put(block, blockBytesLen)
 
 	// Update the blockfilesInfo with the results of adding the new block
 	currentBlkfilesInfo := mgr.blockfilesInfo
--- a/common/ledger/blkstorage/blockfile_mgr.go
+++ b/common/ledger/blkstorage/blockfile_mgr.go
@@ -98,7 +98,7 @@ func newBlockfileMgr(id string, conf *Conf, indexConfig *IndexConfig, indexStore
 	if err != nil {
 		panic(fmt.Sprintf("Error creating block storage root dir [%s]: %s", rootDir, err))
 	}
-	mgr := &blockfileMgr{rootDir: rootDir, conf: conf, db: indexStore, cache: newCache(defaultMaxBlockfileSize)}
+	mgr := &blockfileMgr{rootDir: rootDir, conf: conf, db: indexStore, cache: newCache(defaultBlockCacheSizeBytes)}
 
 	blockfilesInfo, err := mgr.loadBlkfilesInfo()
 	if err != nil {
--- a/common/ledger/blkstorage/blocks_itr.go
+++ b/common/ledger/blkstorage/blocks_itr.go
@@ -20,13 +20,12 @@ type blocksItr struct {
 	stream               *blockStream
 	closeMarker          bool
 	closeMarkerLock      *sync.Mutex
-	cachedPrev           bool
 }
 
 func newBlockItr(mgr *blockfileMgr, startBlockNum uint64) *blocksItr {
 	mgr.blkfilesInfoCond.L.Lock()
 	defer mgr.blkfilesInfoCond.L.Unlock()
-	return &blocksItr{mgr, mgr.blockfilesInfo.lastPersistedBlock, startBlockNum, nil, false, &sync.Mutex{}, false}
+	return &blocksItr{mgr, mgr.blockfilesInfo.lastPersistedBlock, startBlockNum, nil, false, &sync.Mutex{}}
 }
 
 func (itr *blocksItr) waitForBlock(blockNum uint64) uint64 {
@@ -72,18 +71,18 @@ func (itr *blocksItr) Next() (ledger.QueryResult, error) {
 
 	cachedBlock, existsInCache := itr.mgr.cache.get(itr.blockNumToRetrieve)
 	if existsInCache {
-		logger.Debugf("Retrieved block %d from ledger in-memory cache", itr.blockNumToRetrieve)
-		itr.cachedPrev = true
-		itr.blockNumToRetrieve++
-		return cachedBlock, nil
-	}
 
-	if itr.cachedPrev {
-		itr.cachedPrev = false
+		// Close the stream if applicable, because since we're fetching from
+		// the cache, the file stream is no longer in sync with the block number to retrieve.
+		// Hopefully the next iteration will also be a cache hit.
 		if itr.stream != nil {
 			itr.stream.close()
 			itr.stream = nil
 		}
+
+		logger.Debugf("Retrieved block %d from ledger in-memory cache", itr.blockNumToRetrieve)
+		itr.blockNumToRetrieve++
+		return cachedBlock, nil
 	}
 
 	if itr.stream == nil {
--- a/common/ledger/blkstorage/blocks_itr.go
+++ b/common/ledger/blkstorage/blocks_itr.go
@@ -20,12 +20,13 @@ type blocksItr struct {
 	stream               *blockStream
 	closeMarker          bool
 	closeMarkerLock      *sync.Mutex
+	cachedPrev           bool
 }
 
 func newBlockItr(mgr *blockfileMgr, startBlockNum uint64) *blocksItr {
 	mgr.blkfilesInfoCond.L.Lock()
 	defer mgr.blkfilesInfoCond.L.Unlock()
-	return &blocksItr{mgr, mgr.blockfilesInfo.lastPersistedBlock, startBlockNum, nil, false, &sync.Mutex{}}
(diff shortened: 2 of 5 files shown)


common/ledger/blkstorage/cache.go
```diff
@@ -0,0 +1,98 @@
...
+	"sync"
+
+	"github.com/hyperledger/fabric-protos-go/common"
+)
+
+const (
+	estimatedBlockSize = 512 * 1024
+)
+
+type Cache struct {
```
cendhu: unexport `Cache`

common/ledger/blkstorage/cache.go
```diff
@@ -0,0 +1,98 @@
...
+	isCacheDisabled := maxSizeBytes == 0
+
+	return &Cache{
+		disabled:     isCacheDisabled,
+		cache:        make(map[uint64]cachedBlock, maxSizeBytes/estimatedBlockSize),
+		maxSizeBytes: maxSizeBytes,
+	}
+}
+
+func (c *Cache) retrieveCachedBlock(seq uint64) (*common.Block, bool) {
```
cendhu: nit. `get(blockNum uint64)` can be more intuitive

common/ledger/blkstorage/cache.go
```diff
@@ -0,0 +1,98 @@
...
+	}
+
+	c.cacheLock.RLock()
+	defer c.cacheLock.RUnlock()
+
+	cachedBlock, exists := c.cache[seq]
+	return cachedBlock.block, exists
+}
+
+func (c *Cache) cacheBlock(block *common.Block, blockSize int) {
```
cendhu: nit. `put(blockNum uint64)` can be more intuitive

common/ledger/blkstorage/cache.go
```diff
@@ -0,0 +1,98 @@
...
+	}
+
+	seq := block.Header.Number
+
+	if c.maxSeq > seq {
+		return
+	}
+
+	// Insert the block to the cache
+	c.maxSeq = seq
```
cendhu: if `seq` can be lower than the `maxSeq`, it can be higher too. It may be good to check `if c.maxSeq + 1 == seq` too

common/ledger/blkstorage/blockfile_mgr.go
```diff
@@ -328,6 +330,8 @@ func (mgr *blockfileMgr) addBlock(block *common.Block) error {
 		return errors.WithMessage(err, "error appending block to file")
 	}
 
+	defer mgr.cache.cacheBlock(block, blockBytesLen)
```
cendhu: Please note that `blockBytesLen` refers to the length of the serialized bytes, rather than the memory occupied by `*common.Block`. However, in my view, the differences between these two might not be significant enough to cause concern.

common/ledger/blkstorage/cache.go
```diff
@@ -0,0 +1,98 @@
...
+import (
+	"fmt"
+	"sync"
+
+	"github.com/hyperledger/fabric-protos-go/common"
+)
+
+const (
+	estimatedBlockSize = 512 * 1024
+)
```
cendhu, line 18: Maybe we can get rid of this to keep it simple.

common/ledger/blkstorage/blocks_itr.go
```diff
@@ -68,12 +69,32 @@ func (itr *blocksItr) Next() (ledger.QueryResult, error) {
...
+		return cachedBlock, nil
+	}
+
+	if itr.cachedPrev {
+		itr.cachedPrev = false
+		if itr.stream != nil {
+			itr.stream.close()
+			itr.stream = nil
+		}
+	}
```
cendhu: There is a rare corner case where we might be opening and closing the stream too frequently. This depends on the blocks present in the cache and the requested blocks.

Example scenario: The cache currently contains blocks 100 to 199 (with a cache size of 100 blocks), and the iterator is reading from block 99 to 200 while new blocks are being added to the cache. In this situation, we may encounter cache hits for some blocks and cache misses for others, resulting in the opening and closing of the stream.

It would be beneficial if we could incorporate a `Seek(fileNum, offset)` function into the existing stream instead of closing it.

common/ledger/blkstorage/cache_test.go
```diff
@@ -0,0 +1,92 @@
...
+
+	block, exists := c.retrieveCachedBlock(9)
+	require.False(t, exists)
+	require.Nil(t, block)
+
+	for i := 10; i < 20; i++ {
+		block, exists := c.retrieveCachedBlock(uint64(i))
+		require.True(t, exists)
+		require.NotNil(t, block)
+		require.Equal(t, uint64(i), block.Header.Number)
```
cendhu: nit. We can introduce a function `assertCache(t *testing.T, blockNum uint64, exist bool)` to improve the readability.

common/ledger/blkstorage/cache.go
cendhu: I assume we are not using lock for this because at most one thread is going to call `cacheBlock()` as we process one block at a time and also `c.maxSeq` is not used by the `retrieveCachedBlock()`.

common/ledger/blkstorage/cache.go
```diff
@@ -0,0 +1,98 @@
...
+	maxSeq       uint64
+	maxSizeBytes int
+}
+
+type cachedBlock struct {
+	block     *common.Block
+	blockSize int
+}
+
+func NewCache(maxSizeBytes int) *Cache {
```
cendhu: unexport `NewCache()`

cendhu (commented):
This is a useful optimisation.

common/ledger/blkstorage/cache.go
```diff
@@ -0,0 +1,98 @@
...
+	// Given a series of k > 0 consecutive elements: {i, i+1, i+2, ... , i+k-1}
+	// If the max sequence is j then j=i+k-1, and then the lowest element i is j-k+1
+	evictedIndex := c.maxSeq - uint64(cachedItemCount) + 1
+	evictedBlock, exists := c.cache[evictedIndex]
+	if !exists {
+		panic(fmt.Sprintf("programming error: last stored block sequence is %d and cached block count"+
+			" is %d but index to be evicted %d was not found", c.maxSeq, cachedItemCount, evictedIndex))
+	}
+	delete(c.cache, evictedIndex) // Delete minimum entry
+	c.sizeBytes -= evictedBlock.blockSize
```
pfi79, line 104: When removing a single block from the cache, it is assumed that their sizes are comparable. 
If they are not? 
If one block is much larger than the others?
Perhaps more than one block needs to be removed?

yacovm: What makes you think it is assumed their size is comparable? We track the cumulative size of the blocks added. 

Had the code assumed blocks were similar in size, it would have only kept track of their amount.

common/ledger/blkstorage/blocks_itr.go
yacovm: The only way to open the stream and then close it, is if you retrieved a block from the cache and then you got a cache miss. 

If you retrieve a block from the cache and then get a cache miss, it means you were pulling the "tail" of the cache (since it is continuous) and then the ledger is appended with so many transactions per second that the client that pulls the transactions can't keep up with the speed of blocks being appended to the ledger.

If the client is the peer or something equivalent to the peer that processed transactions, and it can't keep up with the ordering service, then we have a much bigger problem here as the effective latency of transactions will explode.

common/ledger/blkstorage/cache.go
yacovm: If it's higher it means we have appended a block out of order. If we did that, we should panic here, no?

common/ledger/blkstorage/cache.go
pfi79: I decided so because you can only remove one block from the cache per block added to the cache.
And the number of blocks in the cache can only increase or stay the same.
If there are 10 blocks in the cache now, it will never become 5 if the chain is not interrupted.

pfi79: illustrate:
let's say max 100.
add 100 blocks with weight 1.
occupied 100

add a block with a weight of 95
total 100+95-1 = 194
We'll say it 100 times.
and with a maximum of 100, the sum of the weights would be 9,500.

it's all in your algorithm.

common/ledger/blkstorage/cache.go
yacovm: On second thought, we can't do this because the block cache is not initialized with the last block number. I prefer to keep it this way for simplicity.

common/ledger/blkstorage/cache.go
yacovm: I see your point. I don't think it's an issue unless the system increases its block size over time. 

Regardless, if we increase the block size we get to a situation where the cache's max size is not taken into consideration.

I changed accordingly, thanks for pointing this out.

common/ledger/blkstorage/cache.go
yacovm: I would rather not keep [magic numbers](https://pkg.go.dev/github.com/tommy-muehle/go-mnd) if possible :-)

common/ledger/blkstorage/blockfile_mgr.go
yacovm: I am aware, but it's the fastest way of doing it.

common/ledger/blkstorage/cache.go
yacovm: Please look at the new version I pushed. It protects against all cases except where the cache is not initialized. I did not want to initialize the cache with the latest block number for simplicity reasons.

common/ledger/blkstorage/blockfile_mgr.go
```diff
@@ -96,7 +98,7 @@ func newBlockfileMgr(id string, conf *Conf, indexConfig *IndexConfig, indexStore
 	if err != nil {
 		panic(fmt.Sprintf("Error creating block storage root dir [%s]: %s", rootDir, err))
 	}
-	mgr := &blockfileMgr{rootDir: rootDir, conf: conf, db: indexStore}
+	mgr := &blockfileMgr{rootDir: rootDir, conf: conf, db: indexStore, cache: newCache(defaultMaxBlockfileSize)}
```
manish-sethi: `defaultBlockCacheSizeBytes` not `defaultMaxBlockfileSize` :-)

common/ledger/blkstorage/blocks_itr.go
manish-sethi: Just for the sake of completeness - This could happen in the normal scenario as well as you first append the block to the file and then add it to the cache. Occasionally, the iterator may keep oscillating between the file and the cache. But, yes, that's not expected to be frequent at the regular transaction rates.

However, one note on code readability here. Why do you have this `if itr.cachedPrev {` block? Can't you simply close the stream inside the `if existsInCache {` block and get rid of `cachedPrev` concept altogether?

manish-sethi (commented):
I left a couple of comments; Otherwise LGTM.

Just curious, have you measured the improvement because of this change? I am raising this because, for the most recent blocks, the blocks mostly come from the filesystem cache. Moreover, the underlying iterator uses a buffered I/O (the default value is 4K, but can be increased). So, basically what this cache saves is one de-serialization; unlike what you mentioned in the commit message about disk IO.

common/ledger/blkstorage/blockfile_mgr.go
yacovm: good catch.

common/ledger/blkstorage/blocks_itr.go
yacovm: Done

manish-sethi (approved):

PR #4305, opened 2023-07-03
BFT Block Puller: censorship monitor

tock-ibm:
#### Type of change
- New feature

#### Description

BFT Block Puller: censorship monitor

BFT header receivers that pull headers from the orderers, and keep the time and number of the last one.
The header receivers verify each block as it arrives.
The header receivers will receive (or ask for) full config blocks - in a later commit.
The header receivers will maintain their own private block verifier (bundle) - in a later commit.

A BFT censorship monitor which periodically compares the progress of headers relative to blocks.
The monitor triggers an alert if a header is ahead of the block stream for more than a certain time.

The BFTDeliverer is only a skeleton.

#### Related issues
#4306


internal/pkg/peer/blocksprovider/bft_censorship_monitor.go
```diff
@@ -0,0 +1,277 @@
...
+	"time"
+
+	"github.com/hyperledger/fabric-protos-go/orderer"
+	"github.com/hyperledger/fabric/common/flogging"
+	"github.com/hyperledger/fabric/internal/pkg/peer/orderers"
+	"github.com/pkg/errors"
+)
+
+type BlockProgressReporter interface {
+	BlockProgress() (uint64, time.Time)
```
yacovm, line 30: which struct implements this?

internal/pkg/peer/blocksprovider/bft_header_receiver.go
```diff
@@ -0,0 +1,164 @@
...
+}
+
+func (hr *bftHeaderReceiver) setStarted() {
+	hr.mutex.Lock()
+	defer hr.mutex.Unlock()
+	hr.started = true
+}
+
+// CloseSend closes the client connection
+func (hr *bftHeaderReceiver) CloseSend() error {
```
yacovm: should we not wait until `DeliverHeaders()` exits?

internal/pkg/peer/blocksprovider/bft_censorship_monitor.go
```diff
@@ -0,0 +1,277 @@
...
+		select {
+		case <-m.stopCh:
+			m.logger.Debug("Stopping to monitor progress")
+			m.errorCh <- &errStopping{message: "received a stop signal"}
+			return
+		case <-time.After(bftBlockCensorshipTimeout / 100):
+			if m.detectBlockCensorship() {
+				m.errorCh <- &errCensorship{message: fmt.Sprintf("block censorship detected, endpoint: %s", m.fetchSources[m.blockSourceIndex])}
+				m.logger.Debug("Block censorship detected, stopping to monitor progress")
+				m.Stop()
```
yacovm, line 179: I don't understand. Who relaunches this function if we exit?

internal/pkg/peer/blocksprovider/bft_censorship_monitor.go
```diff
@@ -0,0 +1,277 @@
...
+				n: headerNum,
+			})
+		}
+	}
+
+	if len(ahead) == 0 {
+		return false
+	}
+
+	sort.Slice(ahead, func(i, j int) bool {
```
yacovm: why do you sort this? You can just iterate over it and find the max/min no? 

I know the computation here is negligible, but this is just awkward to sort in order to get the minimum / maximum.

internal/pkg/peer/blocksprovider/bft_header_receiver.go
```diff
@@ -0,0 +1,164 @@
...
+func (hr *bftHeaderReceiver) LastBlockNum() (uint64, time.Time, error) {
+	hr.mutex.Lock()
+	defer hr.mutex.Unlock()
+
+	if hr.lastHeader == nil {
+		return 0, time.Unix(0, 0), errors.New("Not found")
+	}
+
+	if !hr.lastHeaderVerified {
+		hr.lastHeaderVerified = true
```
yacovm: you are setting this to true prematurely, no? what if it's not verified in the end? don't you need to set it to false then?

internal/pkg/peer/blocksprovider/bft_header_receiver.go
```diff
@@ -0,0 +1,164 @@
...
+		hr.lastHeaderOK = true
+
+		err := hr.msgCryptoVerifier.VerifyBlockAttestation(hr.chainID, hr.lastHeader)
```
yacovm: Why do you lazily verify the attestation when you probe for it and not when you ingest it? What's the point of it? You anyway process through all block headers here, no?

internal/pkg/peer/blocksprovider/bft_censorship_monitor.go
```diff
@@ -0,0 +1,277 @@
...
+		if headerNum > lastBlockNumber {
+			m.logger.Debugf("[%s] header receiver: %s, is ahead of block receiver; header num=%d, time=%s; block num=%d, time=%s", m.chainID, ep, headerNum, headerTime, lastBlockNumber, lastBlockTime)
+			ahead = append(ahead, timeNumber{
+				t: headerTime,
+				n: headerNum,
+			})
+		}
+	}
+
+	if len(ahead) == 0 {
```
yacovm, line 263: I don't think this is correct. 

We can have a few headers ahead of us, can we not? 

We should simply not have headers ahead of us with timestamps that are too in the past. 
What we should really look for, is the timestamp of when we received our last block and when was the last block header received from the header pullers, don't you agree?

internal/pkg/peer/blocksprovider/bft_censorship_monitor.go
```diff
@@ -0,0 +1,277 @@
...
+// We track the progress of header receivers against the block reception progress.
+// If there is a header that is ahead of the last block, and a timeout had passed since that header was received, we
+// declare that censorship was detected.
+// When censorship is detected, errCensorship is sent to the errorCh which can be read by ErrorsChannel() method.
+type bftCensorshipMonitor struct {
+	chainID          string
+	headerVerifier   BlockVerifier
+	requester        *DeliveryRequester
+	fetchSources     []*orderers.Endpoint
+	blockSourceIndex int
```
yacovm, line 52: the source index and sources are never updated.

internal/pkg/peer/blocksprovider/bft_header_receiver.go
tock-ibm: It is lazy because we test the front of the header receivers array every T ms.

If T is larger compared to the block rate (say T=100ms, and rate=1k block/sec), we save a lot of unnecessary verifications (100x saving, resp).

If T is small compared to the block rate (say T=100ms, and rate = 1 block/sec), we use the verification caching feature, i.e. we only verify a header once and remember what the result was (10x saving, resp.).

Sig. verification is relatively costly, so we want to avoid unnecessary verifications.

internal/pkg/peer/blocksprovider/bft_censorship_monitor.go
tock-ibm: When we have a header ahead of us, we raise a suspicion and record the time. We then wait for a censorship-timeout before we declare the suspicion as a censorship event.

If a block arrives before the timeout expires we remove the suspicion.

internal/pkg/peer/blocksprovider/bft_censorship_monitor.go
tock-ibm: done.

internal/pkg/peer/blocksprovider/bft_censorship_monitor.go
tock-ibm: This method should be run using a dedicated goroutine.

internal/pkg/peer/blocksprovider/bft_censorship_monitor.go
tock-ibm: This will be implemented by the BFTDeliverer, which keeps track of block arrival.

internal/pkg/peer/blocksprovider/bft_header_receiver.go
tock-ibm: Not really. Actually `DeliverHeaders` calls this one as well when it exits. 
This is used to send a signal to the orderer to stop sending headers and close the connection.
`DeliverHeaders` will pickup the error from the Recv() and exit.

internal/pkg/peer/blocksprovider/bft_header_receiver.go
tock-ibm: when we receive a header we do 
```
// do not verify, just save for later, in case the block-receiver is suspected of censorship
			hr.logger.Debugf("[%s][%s] Saving block with header & metadata, blockNum = [%d], block = [%v]", hr.chainID, hr.endpoint, blockNum, t.Block)
			hr.mutex.Lock()
			hr.lastHeader = t.Block
			hr.lastHeaderTime = time.Now()
			hr.lastHeaderVerified = false
			hr.lastHeaderOK = false
			hr.mutex.Unlock()
```
L84

internal/pkg/peer/blocksprovider/bft_censorship_monitor.go
tock-ibm: True, it is only set in the constructor method. The current plan is to stop the monitor and recreate it when ever we 
- refresh endpoints
- detect censorship (i.e. change the fetch index)

Since these are rare events, I prefer simplicity.

Issue #4302, opened 2023-06-29
panic : runtime error : index out of range [1] with length 1


JungHyeokChoi:
### Description

Hello, I started the peer node with version 2.5.3 but got an error.

The error is follow:

```
2023-06-29 05:52:57.596 UTC 05a5 INFO [gossip.gossip] Stop -> Stopping gossip
2023-06-29 05:52:57.596 UTC 05a6 DEBU [gossip.gossip] handlePresumedDead -> Exiting
2023-06-29 05:52:57.596 UTC 05a7 DEBU [gossip.gossip] acceptMessages -> Exiting
2023-06-29 05:52:57.596 UTC 05a8 INFO [gossip.discovery] Stop -> Stopping
2023-06-29 05:52:57.596 UTC 05a9 INFO [gossip.discovery] Stop -> Stopped
2023-06-29 05:52:57.596 UTC 05aa INFO [gossip.comm] Stop -> Stopping
2023-06-29 05:52:57.596 UTC 05ab DEBU [gossip.comm] Stop -> Shut down connection store, connection count: 0
2023-06-29 05:52:57.596 UTC 05ac DEBU [gossip.comm] func1 -> Exiting Accept() loop
2023-06-29 05:52:57.596 UTC 05ad INFO [gossip.comm] Stop -> Stopped
panic: runtime error: index out of range [1] with length 1

goroutine 1 [running]:
http://github.com/hyperledger/fabric/msp.(*bccspmsp).sanitizeCert(0x0?, 0xc0000f7180)
	/msp/mspimpl.go:924 +0x245
http://github.com/hyperledger/fabric/msp.newIdentity(0xc00054e960?, {0x1d77100?, 0xc000015b18}, 0xc000413400)
	/msp/identities.go:61 +0x105
http://github.com/hyperledger/fabric/msp.(*bccspmsp).getIdentityFromConf(0xc000413400, {0xc00016b200?, 0x44b?, 0x480?})
	/msp/mspimpl.go:200 +0x9e
http://github.com/hyperledger/fabric/msp.(*bccspmsp).setupCAs(0xc000413400, 0xc0001e3680)
	/msp/mspimplsetup.go:135 +0x41f
http://github.com/hyperledger/fabric/msp.(*bccspmsp).preSetupV142(0xc000413400, 0xe9228e?)
	/msp/mspimplsetup.go:570 +0x3f
[...]

PR #4307, opened 2023-07-03
[Fix]: Provide a meaningful error for cert sanitization

C0rWin:
This commit handles the error where the certificate sanitization procedure fails to construct the certificate chain due to misconfiguration. Before this commit, the peer will simply fail with panic without a clear explanation of what exactly was wrong.

Addresses (#4302).

### DIFF
--- a/msp/mspimpl.go
+++ b/msp/mspimpl.go
@@ -13,6 +13,7 @@ import (
 	"encoding/asn1"
 	"encoding/hex"
 	"encoding/pem"
+	"fmt"
 	"strings"
 
 	"github.com/golang/protobuf/proto"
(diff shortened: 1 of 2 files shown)


msp/mspimpl.go
```diff
@@ -919,9 +920,14 @@ func (msp *bccspmsp) sanitizeCert(cert *x509.Certificate) (*x509.Certificate, er
 		}
 
 		// ok, this is no a root CA cert, and now we
-		// then we have chain of certs and can get parent
+		// have chain of certs and can extract parent
 		// to sanitize the cert whenever it's intermediate or leaf certificate
-		parentCert := chain[1]
+		var parentCert *x509.Certificate
```
yacovm, line 925: i think it's safe to merge since if it's not at least of size 1, it crashes... right?

C0rWin: yep.

yacovm (approved):

yacovm:
LGTM, maybe also @ale-linux can take a look

yacovm:
@denyeart this should probably be backported, and reviewed again by someone else

denyeart:

denyeart:
LGTM, backported to release-2.5.

PR #4309, opened 2023-07-07
feat(peer): enable empty env variable 

### LINKED ISSUE #4076: peer: unable to disable core.vm.endpoint from env var
core.vm.endpoint has default value set https://github.com/hyperledger/fabric/blob/bdb311b8c70d82b387b6bac2e8feaa7f8a486f71/sampleconfig/core.yaml#L516

I assume due to backward compatibility since users might relay on that default value?

I'm trying to run peer on kubernetes using default config file from  `/etc/hyperledger/fabric/core.yaml` but setting `CORE_VM_ENDPOINT` to empty string does not override default config, resulting in healthcheck failing on trying to check docker service. And looks like the only solution is to mirror whole core.yaml file to configmap and then override it using env vars
https://github.com/hyperledger/fabric-samples/blob/main/test-network-k8s/config/org1/core.yaml

I was able to track this to https://github.com/spf13/viper/blob/3970ad177ecc286c1e14b64c18449cb30f73d642/README.md?plain=1#L283 :
> By default empty environment variables are considered unset and will fall back to
the next configuration source. To treat empty environment variables as set, use
the `AllowEmptyEnv` method.

Would be nice to have it enabled here: https://github.com/hyperledger/fabric/blob/bdb311b8c70d82b387b6bac2e8feaa7f8a486f71/cmd/peer/main.go#L32 

Or perhaps just comment default value in example config file?

yuriechan:
#### Type of change

#### Description
Enable empty value to be set for environment variable, and not have it overwrite it with a default value. 

#### Additional details
- Pulled the viper configuration into a utility function 
- Added a viper configuration to allow empty env variables 
- Added a test that verifies empty env variables to take precedence over defaults

#### Related issues
None

Resolves #4076

### DIFF
--- a/cmd/peer/main.go
+++ b/cmd/peer/main.go
@@ -28,7 +28,7 @@ import (
 var mainCmd = &cobra.Command{Use: "peer"}
 
 func main() {
-	setEnvConfig()
+	setEnvConfig(viper.GetViper())
 
 	// Define command-line flags that are valid for all peer commands and
 	// subcommands.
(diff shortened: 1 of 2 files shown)


denyeart:
Thank you for the contribution.

A few things that need to be resolved:

- Please add a PR Description
- The Basic Checks failed because "goimports -l -w" needs to be run against the test file.
- The DCO check failed because you need to sign the commit, see https://github.com/hyperledger/fabric/pull/4309/checks?check_run_id=14869644772

Please amend and re-push.

yuriechan:
@denyeart 

Hi, I resolved your review comments. 
Would it be possible for you to review the PR?

yacovm (approved):

PR #4328, opened 2023-07-23
fix: missing tag for make docker

Haupc:
# Fix missing tag for `make docker`
#### Type of change

- Bug fix

#### Description

When `make docker` make file is miss the tag. So, `make docker-tag-latest` and `make docker-tag-stable` will not work

#### Additional details

Reproduce: run `make docker-tag-latest` or `make docker-tag-stable` after run `make docker`

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -148,7 +148,7 @@ integration-test: integration-test-prereqs
 	./scripts/run-integration-tests.sh $(INTEGRATION_TEST_SUITE)
 
 .PHONY: integration-test-prereqs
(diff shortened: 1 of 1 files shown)


denyeart:
@Haupc Thanks for the contribution, but I believe the use of DOCKER_TAG has been retired. I'd therefore suggest it be entirely removed from Makefile rather than adding it back. Additionally, I think we can remove the Make targets that referenced it, including docker-list, docker-tag-latest, and docker-tag-stable.

Issue #4348, opened 2023-08-03
BFT Block Puller: A BFTDeliverer in the peer


tock-ibm:

tock-ibm:
Merged

PR #4365, opened 2023-08-15
BFT Block Puller: BFTDeliverer

tock-ibm:
#### Type of change
- New feature

#### Description

- A BFTDeliverer that fetches blocks using a block receiver and maintains a BFTCensorshipMonitor.
- Abstract the creation of a BFTCensorshipMonitor via an abstract factory, so that we can use a mock for it in testing.
- Add a "ShuffledEndpoints" method to the connection source and test it. 
- Unit testing of BFTDeliverer.

#### Related issues

#4348


internal/pkg/peer/blocksprovider/bft_deliverer.go
```diff
@@ -85,76 +123,123 @@ func (d *BFTDeliverer) DeliverBlocks() {
...
+				d.blockReceiver.Stop()
+				d.mutex.Unlock()
+				d.fetchSourceIndex = (d.fetchSourceIndex + 1) % len(d.fetchSources)
+				d.incFetchFailureCounter()
+			case *ErrFatal:
+				d.Logger.Errorf("Failure in CensorshipMonitor, something is critically wrong: %s", errMonitor)
+				break FetchAndMonitorLoop
+			default:
+				d.Logger.Errorf("Unexpected error from CensorshipMonitor, something is critically wrong: %s", errMonitor)
+				break FetchAndMonitorLoop
```
yacovm: Can we please not use goto style flow control? 

This function is pretty big, we can extract the for loop to a separate function and just use `continue`.

internal/pkg/peer/blocksprovider/bft_deliverer.go
```diff
@@ -85,76 +123,123 @@ func (d *BFTDeliverer) DeliverBlocks() {
...
-	// TODO stop the monitor
+	if d.censorshipMonitor != nil {
+		d.censorshipMonitor.Stop()
+	}
 }
 
 func (d *BFTDeliverer) refreshSources() {
 	// select an initial source randomly
-	d.fetchSources = shuffle(d.Orderers.Endpoints())
+	d.fetchSources = d.Orderers.ShuffledEndpoints()
```
yacovm, line 279: why do we shuffle each time we refresh an endpoint?

internal/pkg/peer/blocksprovider/bft_deliverer.go
```diff
@@ -85,76 +123,123 @@ func (d *BFTDeliverer) DeliverBlocks() {
...
+				d.Logger.Errorf("Failure in CensorshipMonitor, something is critically wrong: %s", errMonitor)
+				break FetchAndMonitorLoop
+			default:
+				d.Logger.Errorf("Unexpected error from CensorshipMonitor, something is critically wrong: %s", errMonitor)
+				break FetchAndMonitorLoop
 			}
-			// TODO can it be nil?
 		}
+
+		d.censorshipMonitor.Stop()
```
yacovm, line 178: why do we start everything just to tear it down? Can't we just stop one censorship monitor?

internal/pkg/peer/blocksprovider/bft_deliverer.go
tock-ibm: When even a single endpoint is changed, all the endpoints are refreshed, i.e. respective channel is closed. In that case we have a new array of sources, possibly of different length. This only happens following a config TX that actually changes the endpoints, and the first time the bft deliverer is started. That is - it happens very rarely. Following a config TX the raw source slice will be the same in all peers. Therefore, in order to avoid all peers selecting the same orderer as their block source, we have to shuffle the raw sources slice and select from the shuffled slice.

internal/pkg/peer/blocksprovider/bft_deliverer.go
tock-ibm: the `d.censorshipMonitor.Stop()` at the end of the loop is only called when something "bad" happens, which is recoverable with a retry, that is: 

- block fetch error
- censorship event

These are hopefully rare events. Therefore, for simplicity, it is easier to close all the header receivers and block receiver, and reassign everything from scratch, rather than to complicate the implementation and 
- close only the block receiver and replace it with a header receive
- close a header receiver and replace it with a block receiver

Note that the censorship monitor is a collection of header receivers, yet failures in header receivers do not cause a restart of the monitor. Header receiver failures are handled within. This is because eventually, an incorrect orderer (malicious) will be assigned to a header receiver, which should be able to tolerate its behavior and not disrupt the reception of blocks.

internal/pkg/peer/blocksprovider/bft_deliverer.go
tock-ibm: Not sure what you mean by goto style, these are just `break` of the for loop. since they are called within a select, they must be labeled. Labeling break of loops is actually recommended.... 

However, you are correct, this loop is indeed too big, as it does not enter a single screen... so I will break it up and try and make it more readable and simplify the flow control.

yacovm (approved):

PR #4373, opened 2023-08-17
BFT Block Puller: delivery client integration

tock-ibm:
#### Type of change
- New feature

#### Description
Integrate the BFT block deliverer with the deliver client.
If the consensus-type is BFT, the BFTDeliverer is created.

#### Related issues

#4348


internal/pkg/peer/orderers/connection.go
```diff
@@ -81,6 +82,20 @@ func (cs *ConnectionSource) Endpoints() []*Endpoint {
 	return cs.allEndpoints
 }
 
+// ShuffledEndpoints returns a shuffled array of endpoints in a new slice.
```
yacovm, line 85: You added this method but now no one calls `Endpoints()` anymore. 
Maybe delete `Endpoints()` ?

yacovm, line 85: Do we still need the `Refreshed chan struct{}` field?

yacovm (approved):

PR #4378, opened 2023-08-17
fix warnings

pfi79:
- Leading whitespaces before Go directives in comments.
- Incorrect usages of fmt.Printf, fmt.Println, and similar formatting and printing functions.
- Redundant parentheses in expressions and types.
- Types in variable and constant declarations that can be omitted since they can be inferred by the compiler. Such types are redundant, omitting them may improve readability of the code.
- Type conversions that may be omitted.
- Aliases of imported packages that may be omitted.

### DIFF
--- a/bccsp/sw/keygen.go
+++ b/bccsp/sw/keygen.go
@@ -43,7 +43,7 @@ type aesKeyGenerator struct {
 }
 
 func (kg *aesKeyGenerator) KeyGen(opts bccsp.KeyGenOpts) (bccsp.Key, error) {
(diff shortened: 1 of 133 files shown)


denyeart:
Thanks for the cleanup, it all looks reasonable to me. Were the changes recommended by some utility?

pfi79:
> Thanks for the cleanup, it all looks reasonable to me. Were the changes recommended by some utility?

I don't know exactly. I'll show you how I do it. I use goland.
image
enter the inspection code

I get this result
image

Use with care.
I don't agree with all the recommendations.

C0rWin (approved):

PR #4392, opened 2023-08-22
delete ioutil pkg and redundant type casting

pfi79:
ioutil.Discard -> os.Discard
ioutil.NopCloser -> io.NopCloser
delete redundant type casting

### DIFF
--- a/gossip/service/gossip_service_test.go
+++ b/gossip/service/gossip_service_test.go
@@ -914,7 +914,7 @@ func TestChannelConfig(t *testing.T) {
 	require.NoError(t, err)
 
 	mockSignerSerializer := &mocks.SignerSerializer{}
-	mockSignerSerializer.SerializeReturns(api.PeerIdentityType(orgInChannelA), nil)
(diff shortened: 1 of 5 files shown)


gossip/service/gossip_service_test.go
```diff
@@ -914,7 +914,7 @@ func TestChannelConfig(t *testing.T) {
 	require.NoError(t, err)
 
 	mockSignerSerializer := &mocks.SignerSerializer{}
-	mockSignerSerializer.SerializeReturns(api.PeerIdentityType(orgInChannelA), nil)
+	mockSignerSerializer.SerializeReturns(orgInChannelA, nil)
```
denyeart, line 917: This change is not related to ioutil. I'll defer to @C0rWin since he merged the last change to this file.

pfi79: sorry forgot to add a description

C0rWin: indeed, has nothing to do with the declared changes in commit message.
though seems pretty much ok to me.

C0rWin (approved):

PR #4396, opened 2023-08-24
connecting smartbft metrics

pfi79:


yacovm (changes requested):
Please do one change at a time: 

1. One PR to update the library dependency
2. One PR to connect metrics. 

They should not be coupled together.

Apart from that, please consider the following two fixes:

https://github.com/fabric-research/consensus/commit/bd70a992f3320ae47c0e9cba115378d4db9ca2cd

https://github.com/fabric-research/consensus/commit/ddcb609a6e0ee0997d7a7dcdc24fdf33ad7b0f6b

vendor/github.com/SmartBFT-Go/consensus/pkg/types/types.go
```diff
@@ -111,9 +111,8 @@ type Reconfig struct {
 }
 
 type SyncResponse struct {
-	Latest     Decision
-	Reconfig   ReconfigSync
-	RequestDel []RequestInfo
```
yacovm: I never liked the `RequestDel` thing. It does not help since when you sync from another node, it only contains the requests in the last block and not in all blocks that were synced.

orderer/consensus/smartbft/metrics_provider_converter.go
```diff
@@ -0,0 +1,109 @@
...
+	o := metrics.CounterOpts{
+		Namespace:    opts.Namespace,
+		Subsystem:    opts.Subsystem,
+		Name:         opts.Name,
+		Help:         opts.Help,
+		LabelNames:   opts.LabelNames,
+		LabelHelp:    opts.LabelHelp,
+		StatsdFormat: opts.StatsdFormat,
+	}
+	// return m.metricsProvider.NewCounter(o)
```
yacovm: please remove commented out line

yacovm:
The library changes look fine, I think there were two commits, right?  One that removed the deletion of requests and the other one with the mutually exclusive deliver? 

As for the metrics - can we have metrics such that [Fabric metric script](https://github.com/hyperledger/fabric/blob/main/scripts/metrics_doc.sh) will generate the corresponding documentation?

vendor/github.com/SmartBFT-Go/consensus/pkg/types/types.go
pfi79: I didn't like it either. So when your solution appeared, I deleted it.
But, as I remember, it was accumulating requests from all blocks.

orderer/consensus/smartbft/metrics_provider_converter.go
pfi79: removed

yacovm:
```
This branch has conflicts that must be resolved
Use the [web editor](https://github.com/hyperledger/fabric/pull/4396/conflicts) or the to resolve conflicts.
Conflicting files
orderer/consensus/smartbft/chain.go
```

pfi79:
> Please do one change at a time:
> 
> 1. One PR to update the library dependency
> 2. One PR to connect metrics.
> 
> They should not be coupled together.
> 
> Apart from that, please consider the following two fixes:
> 
> [fabric-research/consensus@bd70a99](https://github.com/fabric-research/consensus/commit/bd70a992f3320ae47c0e9cba115378d4db9ca2cd)
> 
> [fabric-research/consensus@ddcb609](https://github.com/fabric-research/consensus/commit/ddcb609a6e0ee0997d7a7dcdc24fdf33ad7b0f6b)

split the changes

pfi79:
> The library changes look fine, I think there were two commits, right? One that removed the deletion of requests and the other one with the mutually exclusive deliver?
> 
> As for the metrics - can we have metrics such that [Fabric metric script](https://github.com/hyperledger/fabric/blob/main/scripts/metrics_doc.sh) will generate the corresponding documentation?

should be done, but if you don't mind I'd do it as a separate pr

yacovm (approved):

pfi79:
> The library changes look fine, I think there were two commits, right? One that removed the deletion of requests and the other one with the mutually exclusive deliver?
> 
> As for the metrics - can we have metrics such that [Fabric metric script](https://github.com/hyperledger/fabric/blob/main/scripts/metrics_doc.sh) will generate the corresponding documentation?

here's the promised change
https://github.com/hyperledger/fabric/pull/4434

yacovm:
> > The library changes look fine, I think there were two commits, right? One that removed the deletion of requests and the other one with the mutually exclusive deliver?
> > As for the metrics - can we have metrics such that [Fabric metric script](https://github.com/hyperledger/fabric/blob/main/scripts/metrics_doc.sh) will generate the corresponding documentation?
> 
> here's the promised change #4434

I saw it, but unfortunately I'm busy right now so I don't have time to review it properly

PR #4398, opened 2023-08-25
up smartbft library

pfi79:


vendor/github.com/SmartBFT-Go/consensus/pkg/consensus/consensus.go
```diff
@@ -222,8 +216,21 @@ func (c *Consensus) reconfig(reconfig types.Reconfig) {
 		c.Logger.Panicf("Configuration is invalid, error: %v", err)
 	}
 
+	tmp := c.nodes
+	var newNodes []uint64
 	c.setNodes(reconfig.CurrentNodes)
 
+OuterLoop:
```
yacovm, line 223: i know this is not related to this PR, but, it's better to collapse this loop into a function, in the library code.

yacovm (approved):

Issue #4406, opened 2023-08-28
Timestamp string conversion is not deterministic


semil:
### Description

Critical areas of the code like https://github.com/hyperledger/fabric/blob/c24c9b57439cfb1a9c2332d953286946ebfc3600/orderer/common/cluster/util.go#L812-L820
are using timestamp's `String()` method in order to calculate it's string representation.

The string method uses:
https://github.com/hyperledger/fabric/blob/c24c9b57439cfb1a9c2332d953286946ebfc3600/vendor/google.golang.org/protobuf/types/known/timestamppb/timestamp.pb.go#L271-L273

https://github.com/hyperledger/fabric/blob/c24c9b57439cfb1a9c2332d953286946ebfc3600/vendor/google.golang.org/protobuf/internal/impl/api_export.go#L173-L177

https://github.com/hyperledger/fabric/blob/c24c9b57439cfb1a9c2332d953286946ebfc3600/vendor/google.golang.org/protobuf/encoding/prototext/encode.go#L85-L98

Which clearly states "**Do not depend on the output being stable. It may change over time across different versions of the program.**".

Evidence to that is the way the timestamp `seconds:1693225332; nanos:471402000` is converted to string:
- `seconds:1693225332  nanos:471402000`
- `seconds:1693225332 nanos:471402000`
note the double vs single spacing.

This string is then converted to bytes array and it's hash is used to calculated different security components.

semil:
@arkadiPiven and I are working on a fix.

Priyansurout:
@semil i like to work on this issues

Priyansurout:
@semil can i work on this issues . i have more reliable approach is to serialize the timestame. we can use  ISO 8601 Format for formating the timestamp

denyeart:
I think we just forgot to close this issue when the fix https://github.com/hyperledger/fabric/pull/4407 was merged. Should we just close this issue @semil or do you want to consider different approaches?

PR #4407, opened 2023-08-28
timestamp encoding bug fix

semil:
See #4406

#### Type of change

- Bug fix

#### Description

#4406

### DIFF
--- a/orderer/common/cluster/clusterservice.go
+++ b/orderer/common/cluster/clusterservice.go
@@ -59,7 +59,7 @@ type ClusterService struct {
 
 type AuthRequestSignature struct {
 	Version        int64
-	Timestamp      string
+	Timestamp      []byte
 	FromId         string
 	ToId           string
 	Channel        string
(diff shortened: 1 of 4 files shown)


orderer/common/cluster/clusterservice.go
```diff
@@ -59,7 +59,7 @@ type ClusterService struct {
 
 type AuthRequestSignature struct {
 	Version        int64
-	Timestamp      string
+	Timestamp      []byte
```
yacovm, line 62: You can just put here int64, and then you don't need EncodeTimestamp at all

semil: IMO we will still need it, because:
1. We want to avoid bugs such as `timestamp.Nano` (taking the wrong time from the struct).
2. We have at least 2 places where we need to convert the timestamp to byte array.

yacovm: right now you're encoding the seconds and not the nano seconds: `binary.LittleEndian.PutUint64(b, uint64(t.Seconds))` so what's the difference from taking the wrong time from the struct? 

This struct is only used for signing and verifying a signature, I don't understand where you convert it to a byte array?

semil: Just saying that we still need the function (to convert timestamp/uint64 to a byte array).
We use the byte array to calculate the session binding hash on both ends.

orderer/common/cluster/util.go
```diff
@@ -942,3 +945,9 @@ func LatestHeightAndEndpoint(puller ChainPuller) (string, uint64, error) {
 	}
 	return mostUpToDateEndpoint, maxHeight, nil
 }
+
+func EncodeTimestamp(t *timestamp.Timestamp) []byte {
+	b := make([]byte, 8)
+	binary.LittleEndian.PutUint64(b, uint64(t.Seconds))
+	return b
```
tock-ibm, line 952: I think you have to add the Nanos as well, append the 4 bytes to the 8 bytes Seconds to make a 12 byte slice.

yacovm: why do we need the nanos?

yacovm (approved):

Issue #4346, opened 2023-08-03
BFT Block Puller: A block validator that is updated immediately after a config block is sent to gossip / orderer-pull-buffer


tock-ibm:
In the peer, the blocks and headers fetched from the orderer are validated using a component in gossip that is validated every time a config block is committed. We need to change that to a mechanism that keeps an earlier view of the config, that is, a config that is updated when a block is given to gossip, rather than committed. 

This kind of mechanism already exists in the orderer block puller, and when we unify the two implementation we will create a single instance that behaves in that manner.

tock-ibm:
Make sure to incorporate https://github.com/hyperledger/fabric/pull/4490 in this PR

PR #4408, opened 2023-08-29
 BFT Block Puller: a common block verifier

tock-ibm:
#### Type of change
- New feature

#### Description
 BFT Block Puller: a common block verifier to the peer the orderer.

The new implementation verifies the integrity and signatures of a block stream, while keeping a copy of the latest configuration.

Every time a config block arrives, it must first be verified using VerifyBlock and then used as an argument to the UpdateConfig method.  

The block stream could be composed of either:

- full blocks, which are verified using the VerifyBlock method, or
- block attestations (a header+metadata, with nil data) which are verified using the VerifyBlockAttestation method.

In both cases, config blocks must arrive in full.

This commit also moved `BlockVerifierAssembler` form `orderer/common/cluster/util.go` to `common/deliverclient/verifier_assembler.go`, as it is now used by both orderer and peer. In addition, `VerificationRegistry` was removed from  `orderer/common/cluster/util.go`  as it is no longer in use.

#### Related issues

#4346


common/deliverclient/block_verification.go
```diff
@@ -0,0 +1,350 @@
...
+	if block.Metadata == nil || len(block.Metadata.Metadata) == 0 {
+		return errors.Errorf("block with id [%d] on channel [%s] does not have metadata", block.Header.Number, a.channelID)
+	}
+
+	configIndex, err := protoutil.GetLastConfigIndexFromBlock(block)
+	if err != nil {
+		return errors.Wrapf(err, "failed getting config index from block number [%d] on channel [%s]", block.Header.Number, a.channelID)
+	}
+
+	if protoutil.IsConfigBlock(block) {
```
yacovm: why are we doing this? If the block is properly signed then it should be equal.

common/deliverclient/block_verification.go
```diff
@@ -0,0 +1,350 @@
...
+}
+
+// VerifyBlock checks block integrity and its relation to the chain, and verifies the signatures.
+func (a *BlockVerificationAssistant) VerifyBlock(block *common.Block) error {
+	if err := a.verifyHeader(block); err != nil {
+		return err
+	}
+
+	// Verify channel ID
+	channelID, err := protoutil.GetChannelIDFromBlock(block)
```
yacovm: why do this?

common/deliverclient/block_verification.go
```diff
@@ -0,0 +1,350 @@
...
+	if err != nil {
+		return errors.Wrapf(err, "failed getting channel id from block number [%d] on channel [%s]", block.Header.Number, a.channelID)
+	}
+
+	if channelID != a.channelID {
```
yacovm: why are we protecting against every possible CPU configuration? Are we afraid we will get blocks for the wrong channel? If we do, the signature check won't pass because the previous block's hash will be different.

common/deliverclient/block_verification.go
```diff
@@ -0,0 +1,350 @@
...
+		return errors.Errorf("invalid block's channel id. Expected [%s]. Given [%s]", a.channelID, channelID)
+	}
+
+	if err = a.verifyMetadata(block); err != nil {
+		return err
+	}
+
+	dataHash, err := protoutil.BlockDataHash(block.Data)
```
yacovm, line 252: this should be first, before we check the signature.

common/deliverclient/block_verification.go
tock-ibm: correct, it is redundant

common/deliverclient/block_verification.go
tock-ibm: correct, it is redundant

common/deliverclient/block_verification.go
tock-ibm: The signature is verified after this, on line 267:
```
err = a.sigVerifierFunc(block.Header, block.Metadata)
```

common/deliverclient/block_verification.go
tock-ibm: Correct, removed.

tock-ibm: done

tock-ibm:
merged

PR #4410, opened 2023-08-30
Add BFT to ordering service overview

yacovm:


docs/source/orderer/ordering_service.md
```diff
@@ -352,5 +359,51 @@ blocks `101` to `180`. Blocks `180` to `196` would then be replicated to `R1`
 through the normal Raft protocol.
 
 
+## BFT
```
denyeart, line 354: Awesome, this section is exactly what I was hoping to see before doing a v3.0.0-preview release.

docs/source/orderer/ordering_service.md
```diff
@@ -352,5 +359,51 @@ blocks `101` to `180`. Blocks `180` to `196` would then be replicated to `R1`
...
+by the rest of the ordering nodes. This is in contrast to Raft, which is not suitable to be deployed in such a harsh adversary model.
+
+Operating the BFT orderer is identicaly to how the Raft orderer is operated: New nodes can be added and removed from
+the channel dynamically and while the system is running, and it is described in the [reconfiguration guide](../create_channel/add_orderer.md).
+
+Similarly to Raft, the BFT leader sends periodical heartbeats to each follower, and if the latter does not hear
+from the leader within a period of time, it starts lobbying other followers to change the leader.
+
+A major difference from Raft, is that when a client submits a transaction to a BFT ordering service, it should send
+the transaction to all nodes instead of to a single node. Sending the transaction to everyone will not only
```
denyeart: It would be good to mention that when using the Fabric Gateway SDKs, the peer Gateway service (../gateway.html) automatically submits the transaction to all ordering service nodes (I think this work has been completed, right?). Otherwise, people will think they need to handle this in their own client applications.

Also, I'd change "everyone" to "all ordering service nodes".

docs/source/orderer/ordering_service.md
```diff
@@ -205,6 +205,13 @@ implementation the node will be used in), check out [our documentation on deploy
   up and manage than Kafka-based ordering services, and their design allows
   different organizations to contribute nodes to a distributed ordering service.
 
+* **BFT** (New!)
```
denyeart: The new BFT option was already introduced in line 188 above. Could you merge the section at line 188 with this section?
Also it is good to use language like "New as of v3.0" instead of "New!" so that we don't have to change the docs again in the next version.

yacovm: Done

docs/source/orderer/ordering_service.md
yacovm: Done

denyeart (approved):

Issue #4415, opened 2023-08-30
Malfunction gateway submit bft


pfi79:
### Description

I removed the "Resubmitting the same transaction" step in the test and the test still ran successfully. 
The error message from the submit function misleads the user.

### Steps to reproduce

https://github.com/hyperledger/fabric/pull/4414

pfi79:
In the standard bft client-server communication scheme, the client sends a query "2x2=?" and receives a calculation result of "4" from each node, collects a quorum of responses and concludes that 2x2=4.
In this variant, losing more nodes will result in an error.

In the fabric with the, the client sends a request "2x2=?" and receives a "accepted for processing" response from each node. 
And quorum is collected by peers using BFTDeliverer.
The client receives the result of calculations from the peer.

Therefore, the broadcastToAll function should be rewritten as follows:
- if the request has not reached anyone, an error is returned
- if nodes >= quorum are reached, then we can successfully exit without waiting for all nodes to be reached.
- if at least one node has been reached, then success is also returned

pfi79:
my option is to submit bft

https://github.com/hyperledger/fabric/pull/4417

yacovm:
> if at least one node has been reached, then success is also returned

why? if you reached a single node, including the transaction in a block is not guaranteed.

pfi79:
> why? if you reached a single node, including the transaction in a block is not guaranteed.

Right, no guarantees, but in most cases (temporary failures) the transaction will be executed and delivered.
This is what the test shows (https://github.com/hyperledger/fabric/pull/4414/files), I removed Resubmitting and after the nodes returned, the transaction was executed.

But in the fabric we don't expect results from orders, we expect results from peers.

pfi79:
Now I have one point, the point of "it's not clear what happened" where I can stop, figure it out, and fix it. It's the point of waiting for an event from the peer. Your variant adds another such point for me: the error is back, but check it may not be an error. 

Consequently, the client application to the fabric will have to become more complex. I would like to see such applications simplified.

yacovm:
If you send a transaction and it failed reaching a quorum of nodes, the right way to deal with it, is retrying and then it will hopefully reach a quorum of nodes.

In any case, the client application should be aware that once it submitted a transaction to the network, it may have been included in a block. That is the case for both CFT and BFT.

Sending the transaction only to 1 node will only delay its inclusion in a block and it's better to retry than to hope it will get in eventually.

pfi79:
[...]

PR #4417, opened 2023-08-31
change submit bft

pfi79:
https://github.com/hyperledger/fabric/issues/4415

### DIFF
--- a/internal/pkg/gateway/submit.go
+++ b/internal/pkg/gateway/submit.go
@@ -11,8 +11,6 @@ import (
 	"fmt"
 	"math"
 	"math/rand"
-	"sync/atomic"
-	"time"
 
 	"github.com/golang/protobuf/proto"
 	"github.com/hyperledger/fabric-protos-go/common"
@@ -58,24 +56,22 @@ func (gs *Server) Submit(ctx context.Context, request *gp.SubmitRequest) (*gp.Su
 }
 
 func (gs *Server) submitBFT(ctx context.Context, orderers []*orderer, txn *common.Envelope, clusterSize int, logger *flogging.FabricLogger) (*gp.SubmitResponse, error) {
+	ctx, cancel := context.WithCancel(ctx)
+	defer cancel()
 	// For BFT, we send transaction to ALL orderers
 	waitCh := make(chan *gp.ErrorDetail, len(orderers))
-	go gs.broadcastToAll(orderers, txn, waitCh, logger)
+	go gs.broadcastToAll(ctx, orderers, txn, waitCh, logger)
 
 	quorum, _ := computeBFTQuorum(uint64(clusterSize))
-	successes, failures := 0, 0
+	successes := 0
 	var errDetails []proto.Message
-loop:
+
 	for i, total := 0, len(orderers); i < total; i++ {
 		select {
 		case osnErr := <-waitCh:
 			// Broadcast completed normally
 			if osnErr != nil {
 				errDetails = append(errDetails, osnErr)
-				failures++
-				if failures > total-quorum {
-					break loop
-				}
 			} else {
 				successes++
 				if successes >= quorum {
@@ -88,26 +84,20 @@ loop:
 			return nil, newRpcError(codes.DeadlineExceeded, "submit timeout expired while broadcasting to ordering service")
 		}
 	}
-	logger.Warnw("Insufficient number of orderers could successfully process transaction to satisfy quorum requirement", "successes", successes, "quorum", quorum)
-	return nil, newRpcError(codes.Unavailable, "insufficient number of orderers could successfully process transaction to satisfy quorum requirement", errDetails...)
+	if successes == 0 {
+		logger.Warnw("Insufficient number of orderers could successfully process transaction to satisfy quorum requirement", "successes", successes, "quorum", quorum)
+		return nil, newRpcError(codes.Unavailable, "insufficient number of orderers could successfully process transaction to satisfy quorum requirement", errDetails...)
+	}
+	return &gp.SubmitResponse{}, nil
 }
 
-func (gs *Server) broadcastToAll(orderers []*orderer, txn *common.Envelope, waitCh chan<- *gp.ErrorDetail, logger *flogging.FabricLogger) {
-	everyoneSubmitted := make(chan struct{})
-	var numFinishedSend uint32
-
-	broadcastContext, broadcastCancel := context.WithCancel(context.Background())
-	defer broadcastCancel()
+func (gs *Server) broadcastToAll(ctx context.Context, orderers []*orderer, txn *common.Envelope, waitCh chan<- *gp.ErrorDetail, logger *flogging.FabricLogger) {
 	for _, o := range orderers {
 		go func(ord *orderer) {
 			logger.Infow("Sending transaction to orderer", "endpoint", ord.logAddress)
-			ctx, cancel := context.WithCancel(broadcastContext)
+			ctx, cancel := context.WithTimeout(ctx, gs.options.BroadcastTimeout)
 			defer cancel()
 			response, err := gs.broadcast(ctx, ord, txn)
-			// If I'm the last to submit, notify this
-			if atomic.AddUint32(&numFinishedSend, 1) == uint32(len(orderers)) {
-				close(everyoneSubmitted)
-			}
 			if err != nil {
 				logger.Warnw("Error sending transaction to orderer", "endpoint", ord.logAddress, "err", err)
 				waitCh <- errorDetail(ord.endpointConfig, err.Error())
@@ -124,15 +114,6 @@ func (gs *Server) broadcastToAll(orderers []*orderer, txn *common.Envelope, wait
 				}
 			}
 		}(o)
-	}
-
-	t1 := time.NewTimer(gs.options.BroadcastTimeout)
-	defer t1.Stop()
-	select {
-	case <-everyoneSubmitted:
-		return
-	case <-t1.C:
-		return
 	}
 }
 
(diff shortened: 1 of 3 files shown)


internal/pkg/gateway/submit.go
```diff
@@ -88,26 +84,20 @@ loop:
 			return nil, newRpcError(codes.DeadlineExceeded, "submit timeout expired while broadcasting to ordering service")
 		}
 	}
-	logger.Warnw("Insufficient number of orderers could successfully process transaction to satisfy quorum requirement", "successes", successes, "quorum", quorum)
-	return nil, newRpcError(codes.Unavailable, "insufficient number of orderers could successfully process transaction to satisfy quorum requirement", errDetails...)
+	if successes == 0 {
```
yacovm, line 87: if we submitted to less than a quorum, inclusion of our transaction is not guaranteed.

internal/pkg/gateway/submit_test.go
```diff
@@ -555,67 +555,6 @@ func TestSubmit(t *testing.T) {
 				orderer4Mock.client.(*mocks.ABClient).BroadcastReturns(nil, errors.New("orderer4 fails"))
 			},
 		},
-		{
```
yacovm, line 558: did you remove these tests because they fail with your production code changes?

yacovm (changes requested):
Please either link the issue you opened to this PR or explain why you wish to make these changes.

internal/pkg/gateway/submit.go
pfi79: we're not waiting for the orderers to give us a result
we expect results from peers

internal/pkg/gateway/submit_test.go
pfi79: yes

pfi79:
> Please either link the issue you opened to this PR or explain why you wish to make these changes.

I thought I tied it up and explained why. https://github.com/hyperledger/fabric/issues/4415

internal/pkg/gateway/submit.go
yacovm: No, in the gateway we wait for results from the orderers because we want to be certain we submitted to enough orderers so the transaction will be included even if the orderers are malicious.

yacovm:
> > Please either link the issue you opened to this PR or explain why you wish to make these changes.
> 
> I thought I tied it up and explained why. #4415

But issues are not part of the git commit history

internal/pkg/gateway/submit.go
pfi79: In theory we should expect a result, but we don't get a result here. The result comes at the peers and is checked there as well

pfi79: In the version that is now in main, I have no confidence in the error message. The error message will often appear falsely.

The client logic is as follows: I perform an operation, if the operation was not successful, then it is definitely unsuccessful. In your case, there is an error message, but the transaction was successful. False error return.

yacovm: The client listens to the transaction resultss through a different API, not though this one. 

This API (submit) is used to submit transactions to the orderer.

At the time of transaction submission you cannot know if the transaction will be in, or not. 

Therefore, it's classified as a failure because you are not 100% certain.

This is different from the semantics of Raft, where sending to the orderer may result in the transaction not being in the block.

internal/pkg/gateway/submit_test.go
yacovm: and should you not change the tests according to how you changed the production code?

internal/pkg/gateway/submit.go
pfi79: I'm sorry, but the client's pattern is always the same:
.........
- sent to the orderer(s)
- waiting for the result from the peers
- what happened? why is my transaction gone? are you sure it's gone? maybe I just timed out.
..........

internal/pkg/gateway/submit_test.go
pfi79: They're all there. There's enough of them. The only error is when there are no warrants.

internal/pkg/gateway/submit.go
yacovm: The Fabric core dictates how clients should act, and not the other way around. 

If sending to the orderers fails, the client should retry, but in any case it should be aware that the transaction might be included in a block in the future, even if the gateway returned a failure, and the application (client side and chaincode) should rely on persistent finalized data (blockchain, world state).

pfi79: I agree with you, but sending to the orderers succeeded

pfi79: Here's your [test](https://github.com/hyperledger/fabric/pull/4414). I took out Resubmitting. I didn't add anything, just removed it. And it works. Run it 1,000 times, it will execute 1,000 times.

pfi79: So Resubmitting is not necessary. Is there a test that would show that Resubmitting is necessary?

yacovm: You removed a line from a test and it still succeed. I don't understand how this means you need to change the production code?

internal/pkg/gateway/submit_test.go
yacovm: But, the production code in your change requires a single acknowledgement from an orderer, the previous code required a quorum. Does that not mean you need to change the tests accordingly?

internal/pkg/gateway/submit.go
pfi79: I removed the code that in the test showed the need for what you require from the client: Resubmitting the same transaction.

There is no test that shows the need for Resubmitting

https://github.com/hyperledger/fabric/pull/4414/files#diff-b76a22c911dc666e77b5c1181ecade1fd9e2eeef1402a1b861ad0d1d0efba264

internal/pkg/gateway/submit_test.go
pfi79: quorum will be checked at the BFTDeliverer stage

internal/pkg/gateway/submit.go
pfi79: Is there a test that would show the need for Resubmitting?

Sorry, but so far the current implementation shows unnecessary error and unnecessary Resubmitting.
This is what compelled me to start the corresponding discussion.

internal/pkg/gateway/submit_test.go
yacovm: The BFT deliverer has nothing to do with the gateway or the gateway unit tests.

internal/pkg/gateway/submit.go
yacovm: Why do you think there is a *need* for resubmitting? The test resubmits, and checks it doesn't fail. It does not mean you need to resubmit.

internal/pkg/gateway/submit_test.go
pfi79: The BFT-deliverer relates to the interaction with bft as a second part of the customer. The first part of the client is sending a request to the ordering, in this case it is the gateway.

[edcc12-modsmart.pdf](https://github.com/hyperledger/fabric/files/12489569/edcc12-modsmart.pdf)

```
From Byzantine Consensus to BFT State Machine Replication: A Latency-
Optimal Transformation
.........
B. Client Operation
.........
When a client wants to issue a request to the replicas, it sends a REQUEST message in the format specified
.........
The client waits for at least f + 1 matching replies from different replicas, for the same sequence number, and return the operation result.
```

The client that sends is the gateway or sdk. 
The client that waits the result is BFT-deliverer.

And the client expects computation results, not "accepted for computation" responses. 

Ok. You decided to confuse the client by expecting (n+f+1)/2 such answers. You have the right. But it has nothing to do with bft.

internal/pkg/gateway/submit.go
pfi79: That's not my opinion. It is the test that thinks so. 
And as soon as I remove this step, and the test is executed successfully and will always be executed, then the repeated resubmitting and returning of the error becomes meaningless. Because the error didn't happen.

internal/pkg/gateway/submit_test.go
yacovm: In the paper you cite, upon a timeout, a non faulty node forwards the requests **to all replicas**. 

In the SmartBFT library, the request is forwarded **only to the leader**. 

Therefore, it is not sufficient to send the request to f+1, because the non faulty node among them will only send to the leader, and not to everyone.

internal/pkg/gateway/submit.go
yacovm: A test might test a specific step does not fail, but it does not mean that this step is essential. It just tests that it does not fail if you do that. It's not the same as testing that it is required.

pfi79: By the way, just a reminder that this is a continuation of a past [discussion](https://github.com/hyperledger/fabric/pull/3975#discussion_r1091543576). I still do not agree with this decision.

For me, this step was the only one that could show the advantage of your solution. 
It turns out that there is no test that would show the correctness of the existing solution.

internal/pkg/gateway/submit_test.go
pfi79: The practice of building factory applications with bft has shown me that sending a transaction to a single node as well will lead to success 99% of the time. Yes, it is better to send to quorum or all (to hit the leader). But even in those rare cases where fewer nodes than the quorum receive the transaction, the transaction is accepted into the block.
This point affects the speed of the application. In my case, the requirement is to complete the transaction as quickly as possible and move on to the next ones. Handling unnecessary fake errors can slow things down.

Well, this is so, from personal cases.

PR #4422, opened 2023-09-03
up consensus library

pfi79:


vendor/github.com/SmartBFT-Go/consensus/internal/bft/controller.go
```diff
@@ -630,6 +623,10 @@ func (c *Controller) sync() (viewNum uint64, seq uint64, decisions uint64) {
 
 	latestSequence := c.latestSeq()
 
+	c.Logger.Debugf("Node %d is setting the checkpoint after sync to view %d and seq %d", c.ID, md.ViewId, md.LatestSequence)
```
yacovm: I'm not sure I agree with this [change](https://github.com/SmartBFT-Go/consensus/commit/154c86662788250a1e54b9c4ba303677399da24f). 

The unit test there uses a decision returned from a synchronization with view number 0, but the view proposal committed is the same sequence but with view 1. This should not be possible.
If view 0 committed sequence 1 then we cannot have view 1 committing this sequence again as done in the test.

When I run the test with view number 0 instead of 1, it crashes because negative view number. 

Can you perhaps fix the test to be more realistic? 

The only way to skip updating the checkpoint, is if we have advanced the view due to a view change, but the sync has returned a decision with an old view. 

Now, if we have moved to a new view, then either we have all blocks up until that new view, or we don't.

If we do, then we don't need to set the checkpoint, as we should have updated it in the past via sync or via direct consensus agreement.

If we don't, then there are some blocks that we have not committed but we did move to a new view regardless. 
However, this is not possible, because whenever we process we new view message, we check our latest decision is aligned with the latest decision in the new view message, so we will not actually move to the new view unless we have actually committed all blocks in the previous view.

yacovm (changes requested):
I agree to the changes up until and not including [this change](https://github.com/SmartBFT-Go/consensus/commit/154c86662788250a1e54b9c4ba303677399da24f).

So either update the dependency to a git commit below it, or please address my comment about that change.

vendor/github.com/SmartBFT-Go/consensus/internal/bft/controller.go
pfi79: I don't understand you.

Do you disagree with the solution to the problem?
or
You don't think there's a problem to be solved here at all?

yacovm: If there is a problem, the unit test in your commit doesn't show it.

pfi79: Why? It completely replicates what I saw in person. Here's the [test.log](https://github.com/hyperledger/fabric/files/12508192/test.log) file where it shows up. Would you like me to show you the lines where it happens?

pfi79: We need to observe node 4

- 1799: got the PrePrepare
- 2045: collected 4 prepares from [1 2 7 3]
- 2103: processing a new view message < NewView
- 2249, 2254,2284: got message commit from 1 3 and 2
- 2303-2313: Starting view with number 1, sequence 1, and decisions 0
- 3873: Sync deliver view 0 last seq 0
- 3897: Deliver view 0 last seq 1 prevSeq 0
- 3929: Synchronizer returned with view number 0 but the controller is at view number 1
- 4067: Delivering to app from deliverDecision the last decision proposal
- 4068: Deliver view 0 last seq 1 prevSeq 1
- 4069: Committed sequence 1 twice

pfi79: I can tell you how to replicate it so you'll see that it's possible

yacovm: Can you make a test that replicates this scenario you described? Because you made a unit test which synchronizes a **different block** than what was committed. 

Can you make a test in [this package](https://github.com/SmartBFT-Go/consensus/blob/v2/test/basic_test.go) that reproduces this? 

I am asking because you are claiming a scenario happens which involves view change and synchronization, yet you made a unit test for the controller.

yacovm: Can you point me to the test of TestViewChangeAfterTryingToFork3053337697 ?

pfi79: [TestViewChangeAfterTryingToFork](https://github.com/SmartBFT-Go/consensus/blob/v2/test/basic_test.go#L2463)

I was unable to replicate this case on the integration test so that it reproduced in 100%.

But here are the steps in which it definitely shows up. Check out a version of this [commit](https://github.com/SmartBFT-Go/consensus/commit/071648f76967ae272f4770ae8281c17a1832db9a) and execute this command:
```bash
for i in {1..200}; do go test -count 1 -race ./...  >> test.log 2>&1; done; 
```

Оn my computer, out of 200 cycles, 1 or 2 end up with exactly the same result "Committed sequence 1 twice".

pfi79: > Can you make a test that replicates this scenario you described? Because you made a unit test which synchronizes a **different block** than what was committed.
> 
> Can you make a test in [this package](https://github.com/SmartBFT-Go/consensus/blob/v2/test/basic_test.go) that reproduces this?
> 
> I am asking because you are claiming a scenario happens which involves view change and synchronization, yet you made a unit test for the controller.

The way I determined the error was this. During synchronization, a block was received, but it did not set checkpoint. Then the next delivery results in a Committed sequence twice.

After the above fix, there has never been a "Committed sequence twice" error

pfi79: I can crack the consensus
I know how to fix consensus.
I can't replicate the failure in a controlled manner.

yacovm: @HagarMeir could you please take a look at this?

HagarMeir: I understand you see a problem.
We need to be able to reproduce it.
The unit test in the controller you provided does not reproduce it, but creates a scenario that is not possible. 
I don't agree with your fix, if the sync returned with a lower view number then why do we set the checkpoint anyway? 

If there is a problem we need to be able to reproduce it and provide a better fix.

pfi79: > I understand you see a problem. We need to be able to reproduce it. The unit test in the controller you provided does not reproduce it, but creates a scenario that is not possible. I don't agree with your fix, if the sync returned with a lower view number then why do we set the checkpoint anyway?
> 
> If there is a problem we need to be able to reproduce it and provide a better fix.

1. I cannot reproduce this error in a controlled way. I have tried. On the integration test it appears 1 time out of 200.
2. Why do you say that the scenario is not possible? There is proof that it is possible. Above is the [log file](https://github.com/hyperledger/fabric/pull/4422#discussion_r1314336154). The unit test repeats exactly what is happening there, or rather the problem I saw. 
- The node fails to participate in the consensus.
- Increases the view.
- during synchronization we get a block that we don't have yet, but it has a smaller view number.
- did not set a checkpoint
3. Synchronization gave me back a block I don't have. Even if it's a block from a smaller view. But I don't have it. I set the checkpoint not the view, but the block numbers. 

I don't understand what's wrong.

HagarMeir: 1. I am now also trying to run the integration tests many times...will update if it was reproduce. 

2. As Yacov said: If view 0 committed sequence 1 then we cannot have view 1 committing this sequence again as done in the test.

3. Is it possible maybe that the sync returned something old? And we have already advanced? If it is possible then we shouldn't update the checkpoint at this stage.

pfi79: > 2. As Yacov said: If view 0 committed sequence 1 then we cannot have view 1 committing this sequence again as done in the test.

The block was received on the previous view at the same time as the view was changed. And it turns out that we don't have it. It came to us during synchronization (this process does not depend on smartbft). [Here](https://github.com/hyperledger/fabric/pull/4422#discussion_r1314341969) is a brief explanation of what is going on there.

> 3. Is it possible maybe that the sync returned something old? And we have already advanced? If it is possible then we shouldn't update the checkpoint at this stage.

No, it can't. Otherwise we would see a corresponding entry in the [log file](https://github.com/hyperledger/fabric/pull/4422#discussion_r1314336154)

pfi79: @yacovm @HagarMeir 
I think I know what you don't like. Yes, in the test, the second delivery of a block from the previous view is done from the viewChanger component of the deliverDecision function. Which in turn calls the Deliver component of MutuallyExclusiveDeliver.  In the unit test it was important for me to deliver a block from the previous view through the Deliver function of the MutuallyExclusiveDeliver component. I did it in the way available to me. 

It seems to me that it doesn't matter how the block from the previous view reaches the Deliver function of the MutuallyExclusiveDeliver component. The main thing is that it can do it, the proof of that is the log file with this case.

yacovm: @pfi79 why don't you make this PR upgrade to the SmartBFT commit before the controversial commit, and also take a look at [this fix](https://github.com/fabric-research/consensus/pull/11)?

pfi79: Let me check your decision. And if the error is corrected, I will easily and without hesitation replace my solution with yours.

yacovm: no no, the fix i am referencing has no connection to what we're discussing, it's something else.

pfi79: So your solution doesn't fix the twice-delivery error?

yacovm: Right, this is an unrelated fix.

pfi79: Okay. I will accept your change now.
But there is one thing that is bothering me: I know how to crash a smartbft orderer with unknown consequences (tested on the bench). I really want to deprive myself of this knowledge.
Why not accept my solution for now, until a better one comes along?

HagarMeir: BTW I am now trying to look at the log you attached with the failure and hopefully write a test that will reach the committed sequence twice problem

pfi79: Were you able to replicate the case with a large number of test runs?

HagarMeir: No...I ran it many times and it did not fail

pfi79: Okay. I'm gonna roll mine right now.

Do you need an up-to-date log file with this error? I can replicate it for you.

yacovm: > Okay. I will accept your change now. But there is one thing that is bothering me: I know how to crash a smartbft orderer with unknown consequences (tested on the bench). I really want to deprive myself of this knowledge. 

How did you reproduce it in a real Fabric environment? What did you do?


> Why not accept my solution for now, until a better one comes along?

Because in order to accept a change we need to understand how it fixes a problem, and for that we need to know exactly how the problem manifests and why.

pfi79: Did you run it that way?

```bash
for i in {1..200}; do go test -count 1 -race ./...  >> test.log 2>&1; done; 
```

When I just ran it without that command, it failed to replicate. Everything completed successfully.

pfi79: > How did you reproduce it in a real Fabric environment? What did you do?

During actual operation, terminated and after some time started 3 out of 7 nodes. Never touched the leader. That is, I emulated a failure, such that at some point the consensus was lost, but then recovered.

pfi79: > How did you reproduce it in a real Fabric environment? What did you do?

This [test](https://github.com/SmartBFT-Go/consensus/blob/v2/test/basic_test.go#L2463) is a representation of what we had going on in the booth. And how it led to bad situations.

- first node 6 removed itself from the consensus.
- then terminated 3 nodes including (4, 5, 6)
- sent the transaction
- wait for a message from nodes (0, 1, 2, 3) that they want to change the leader.
- include (4, 5, 6) nodes
- wait for the actual change of leader and view
- wait for the transaction delivery

But there are some interesting effects here, one of which is " twice delivery".

Most likely it turns out that the leader change and the transaction solution occur simultaneously

pfi79: @yacovm done

pfi79: Pipeline for raft is hung up. Can I ask you to restart pipelines for raft?

yacovm (approved):

vendor/github.com/SmartBFT-Go/consensus/internal/bft/controller.go
yacovm: Thanks

yacovm:
@pfi79 can you perhaps make an integration test that causes an orderer to commit the same block sequence twice?

pfi79:
> @pfi79 can you perhaps make an integration test that causes an orderer to commit the same block sequence twice?

Alas, if I could do that, I would not persuade you. I'd just show you the integration test. 

So far, I don't know how the stars should align for the error to manifest itself in a controlled way.

pfi79:
@yacovm @HagarMeir 
I don't know how to do the right integration test, I'm missing something. I'll give you my observations, maybe it will help you.

After the 3 nodes return, 2 processes start simultaneously: deciding the proposal and changing the view.

By the way, a little digression. Why do these processes occur at the same time? It seems that if a node decides to change the view, it should not participate in the solution of the proposal until the process is finished.

So, 
- our node doesn't get the last commit in time to deliver the proposal. it gets a view change.
- our node still has an undelivered proposal in InFlight. Hence, at all stages of view change, the proposal is not delivered.
- after the view change, during synchronization the last proposal is delivered to us, because other nodes have had time to deliver it
- but it is still in InFlight.
- and there comes a moment when viewChanger delivers to us the proposal that is in InFlight (at this moment double delivery has occurred)

Maybe my observations will help?

HagarMeir:
Here is what I observed with the log you (@pfi79) provided:

Running TestViewChangeAfterTryingToFork
The number of nodes is 7 (n=7; f=2; q=5) (edited)
We will look at node 4
Node 4 receives a view change to view 1 msg from node 7
(node 7 was disconnected for a while and hit heartbeat timeout)
Node 4 submits request alice 1
Node 4 receives prePrepare msg from node 1
Node 4 hits leader-forwarding timeout and complains about the leader (nodes 5,6, and 7 are out)
Node 4 sent view data msg, with next view 1, to the new leader 2
Node 4 got a view change timeout
Node 4 is calling sync because it got a view change timeout
Node 4 synchronizer returned with proposal metadata nil (edited)
Node 4 starting view with number 0, sequence 1, and decisions 0
Node 4 started view change, last view is 1 (after view change timeout)
(nodes 5,6, and 7 reconnected)
Node 4 processed proposal with seq 1
Node 4 collected 4 prepares from [1 2 7 3]
Node 4 got message from 2: < NewView with <ViewData signed by 2 with NextView: 1>, <ViewData signed by 1 with NextView: 1>, <ViewData signed by 4 with NextView: 1>, <ViewData signed by 3 with NextView: 1>, <ViewData signed by 6 with NextView: 1>>
Node 4 processed prepares for proposal with seq 1 (sending commit msg)
Node 4 ViewChanged, the new view is 1
Node 4 starting view with number 1, sequence 1, and decisions 0
(nodes 1, 2, 3, and 7 committed the proposal)
Node 4 got commit of seq 1 in view 0 from 7 while being in seq 1 in view 1
Node 4 is joining view change, last view is 1
Node 4 sent view data msg, with next view 2, to the new leader 3
Node 4 request {alice 1} timeout expired
Node 4 got a view change timeout
Node 4 is calling sync because it got a view change timeout
Node 4 started view change, last view is 2
Node 4 sync deliver view 0
Node 4 deliver view 0 last seq 1 prevSeq 0
Node 4 Synchronizer returned with view number 0 but the controller is at view number 1
Node 4 is joining view change, last view is 2
Node 4 sent view data msg, with next view 3, to the new leader 4
Node 4 got message from 1: <ViewData signed by 1 with NextView: 3> (edited)
Node 4 got message from 2: <ViewData signed by 2 with NextView: 3>
Node 4 got message from 3: <ViewData signed by 3 with NextView: 3>
Node 4 got <ViewData signed by 1 with NextView: 3> from 1, the last decision seq is 1 and this node's current sequence is 0
Node 4 delivering to app from deliverDecision the last decision proposal
Node 4 deliver view 0 last seq 1 prevSeq 1
Node 4 committed sequence 1 twice

HagarMeir:
So I think we should make sure the view and view change don't run at the same time
And also update the checkpoint when sync returns making sure it delivered something new 


I will first try to recreate this with a test

pfi79:
> So I think we should make sure the view and view change don't run at the same time And also update the checkpoint when sync returns making sure it delivered something new
> 
> I will first try to recreate this with a test

@HagarMeir , may I ask, were you able to recreate? 
Maybe there are some intermediate results?

HagarMeir:
@pfi79 Unfortunately I was not able to reproduce it yet

yacovm:
> > So I think we should make sure the view and view change don't run at the same time And also update the checkpoint when sync returns making sure it delivered something new
> > I will first try to recreate this with a test
> 
> @HagarMeir , may I ask, were you able to recreate? Maybe there are some intermediate results?

Hey @pfi79 , here is Hagar's fix https://github.com/fabric-research/consensus/pull/13

yacovm:
I think she did a good job in re-doing that function, I suggest you take a look and perhaps also align the `sync()` function to her PR.

PR #4429, opened 2023-09-07
Smart BFT GRPC forwarding invalid message test

semil:
#### Type of change

- Test update

#### Description

A test to check that if an orderer is forwarding to the leader an invalid TX, the operation should fail.

### DIFF
--- a/integration/nwo/network.go
+++ b/integration/nwo/network.go
@@ -9,6 +9,8 @@ package nwo
 import (
 	"bytes"
 	"context"
+	"crypto/tls"
+	"crypto/x509"
 	"fmt"
 	"io"
 	"net"
@@ -1408,9 +1410,11 @@ func (n *Network) PeerUserSession(p *Peer, user string, command Command) (*gexec
 // services. The client connection should be closed when the tests are done
 // using it.
 func (n *Network) PeerClientConn(p *Peer) *grpc.ClientConn {
-	return n.newClientConn(
+	return n.NewClientConn(
 		n.PeerAddress(p, ListenPort),
 		filepath.Join(n.PeerLocalTLSDir(p), "ca.crt"),
+		"",
+		"",
 	)
 }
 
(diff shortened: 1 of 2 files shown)


integration/nwo/network.go
```diff
@@ -1419,24 +1424,55 @@ func (n *Network) PeerClientConn(p *Peer) *grpc.ClientConn {
...
+		// read ca's cert
+		caCert, err := os.ReadFile(caCertPath)
+		if err != nil {
+			log.Fatal(caCert)
+		}
+
+		// create cert pool and append ca's cert
+		certPool := x509.NewCertPool()
+		if ok := certPool.AppendCertsFromPEM(caCert); !ok {
+			log.Fatal(err)
```
yacovm: `Expect(err).NotTo(HaveOccurred())` instead of `log.Fatal` everywhere

integration/smartbft/smartbft_test.go
```diff
@@ -1258,6 +1263,120 @@ var _ = Describe("EndToEnd Smart BFT configuration test", func() {
...
+			env2 := createEnvelope(
+				channel,
+				network.Organization(follower.Organization).MSPID,
+				cert,
+				[]byte("TEST_MESSAGE_2"),
+				[]byte("nonce_2"),
+				signer,
+			)
+
+			By("Should not be an error")
```
yacovm, line 1370: and a block should be created, no?

integration/smartbft/smartbft_test.go
```diff
@@ -1655,3 +1774,42 @@ func updateBatchSize(
 	}
 	nwo.UpdateOrdererConfig(network, orderer, channel, config, updatedConfig, peer, orderer)
 }
+
+// marshalOrPanic serializes a protobuf message and panics if this
+// operation fails
+func marshalOrPanic(pb proto.Message) []byte {
```
yacovm: we have a `MarshalOrPanic` function already in the code base.

integration/smartbft/smartbft_test.go
```diff
@@ -1655,3 +1774,42 @@ func updateBatchSize(
...
+	data, err := proto.Marshal(pb)
+	if err != nil {
+		panic(err)
+	}
+	return data
+}
+
+func createEnvelope(
```
yacovm: this is nice but can we please try to make a real transaction? Look at how this is done [here](https://github.com/hyperledger/fabric/blob/main/protoutil/txutils.go#L135).

yacovm (approved):

PR #4434, opened 2023-09-07
add docs for smartbft metrics

pfi79:


pfi79:
Can anyone reset the piplines for sbe nwo msp?

common/metrics/gendoc/options.go
```diff
@@ -72,7 +72,8 @@ func FileOptions(f *ast.File) ([]interface{}, error) {
 				if !ok {
 					continue
 				}
-				if imports[ident.Name] != "github.com/hyperledger/fabric/common/metrics" {
```
yeasy, line 75: This is not only doc change, but some code change too.

pfi79: yes, without this change, there would be no documentation

pfi79:
@yacovm, perhaps you might have time to review pr properly?

yacovm (approved):

PR #4438, opened 2023-09-11
add integration test in which the smartbft leader must freeze,

pfi79:
waiting for an answer.

There was a case where the leader froze up and the followers couldn't change him. 
There were two reasons for this:

1. a flaw in the behavior of the smartbft library
2. gateway was only for raft

Both reasons are fixed now.
But a test that showed this error would be nice to add.


yacovm (dismissed):
I don't understand what's the point of having this tested in integration tests? That was a pure library bug. 


Putting that aside, why do you use chaincode deployment as a test transaction? It's needlessly complex as it writes to a local collection.

pfi79:
> I don't understand what's the point of having this tested in integration tests? That was a pure library bug.
> 
> Putting that aside, why do you use chaincode deployment as a test transaction? It's needlessly complex as it writes to a local collection.

Because that's where our consensus was slipping. And we couldn't move it forward. The only thing that helped in this case was to restart the leader orderer with this channel.
Here's the case: 
- we are just starting to deploy the factory
- we try to execute ApproveForMyOrg through the peercli raft.
- the leader hangs up waiting for prepare responses
- other followers don't know anything about the transaction and can't change the leader
- From outside we can still repeat ApproveForMyOrg, but it will go to the leader again and will not be executed.

Only restart orderer. 

Until it's fixed.

pfi79:
It's a repeat of what we encountered in production.

yacovm:
I get what you're saying and I appreciate the initiative to write a test. But this is a corner case that its source is a bug in the library, so it should not be tested in Fabric. 

You've seen that the integration test passes, and I went through the test and it seems correct to me.
I just don't think we should merge the code because the code to test this case should not be in Fabric but in the library.

C0rWin:
> I get what you're saying and I appreciate the initiative to write a test. But this is a corner case that its source is a bug in the library, so it should not be tested in Fabric.

I think there is still a point to having such a test in Fabric, cause this is an integration test and not a unit test where the purpose is to make sure the integration path with the library works as expected. It demonstrates potential flow in the library, which might cause Fabric to manifest the error and hang the process. IMO, there is nothing really wrong with adding this test as an integration test.

pfi79:
> I get what you're saying and I appreciate the initiative to write a test. But this is a corner case that its source is a bug in the library, so it should not be tested in Fabric.
> 
> You've seen that the integration test passes, and I went through the test and it seems correct to me. I just don't think we should merge the code because the code to test this case should not be in Fabric but in the library.

@yacovm , I'm asking you to reconsider your decision. 
Yes, you're right, this is a test of the consensus bug. But it's only half of it.
This test, in the future, will protect fabric from consensus developers. From their ad hoc or random decisions.

yacovm:
> > I get what you're saying and I appreciate the initiative to write a test. But this is a corner case that its source is a bug in the library, so it should not be tested in Fabric.
> > You've seen that the integration test passes, and I went through the test and it seems correct to me. I just don't think we should merge the code because the code to test this case should not be in Fabric but in the library.
> 
> @yacovm , I'm asking you to reconsider your decision. Yes, you're right, this is a test of the consensus bug. But it's only half of it. This test, in the future, will protect fabric from consensus developers. From their ad hoc or random decisions.

Sure, since @C0rWin is seemingly strongly in favor of merging it, we can proceed with this.

integration/smartbft/smartbft_test.go
```diff
@@ -1662,6 +1662,159 @@ var _ = Describe("EndToEnd Smart BFT configuration test", func() {
...
+			Eventually(ordererRunners[2].Err(), network.EventuallyTimeout, time.Second).Should(gbytes.Say("Message from 1"))
+			Eventually(ordererRunners[3].Err(), network.EventuallyTimeout, time.Second).Should(gbytes.Say("Message from 1"))
+
+			// Stopped 2 out of 4 nodes
+			for i := 2; i <= 3; i++ {
+				orderer := network.Orderers[i]
+				By(fmt.Sprintf("Killing %s", orderer.Name))
+				ordererProcesses[i].Signal(syscall.SIGTERM)
+				Eventually(ordererProcesses[i].Wait(), network.EventuallyTimeout).Should(Receive())
+			}
```
C0rWin: Why do you need to start four nodes to shut down 2, send a command to the first one, and then start the other two back again? 

Can't you start only two? And could you do the same without restarting? My understanding of the case should be:

1. The network is being initialized, but not the entire cluster has started yet
2. Non-BFT client sends transaction only to the leader; the rest are not aware, and some nodes have not even started yet
3. The leader is waiting to get prepared responses, while the rest are not even aware he is waiting 
4. No one can wake up the leader from waiting; the rest of the cluster is not aware of the transaction being in flight

Therefore, I do not understand why you cannot simulate the same behavior in the test and need this extra step of starting, stoping and restarting again.

integration/smartbft/smartbft_test.go
```diff
@@ -1662,6 +1662,159 @@ var _ = Describe("EndToEnd Smart BFT configuration test", func() {
...
+
+			By("Joining peers to testchannel1")
+			network.JoinChannel(channel, network.Orderers[0], network.PeersWithChannel(channel)...)
+
+			By("Try deploying chaincode")
+			peers := network.PeersWithChannel(channel)
+			if len(peers) == 0 {
+				return
+			}
```
C0rWin: Is this expected here to leave the test and not indicate the error of not being able to find peers to send transactions to?

integration/smartbft/smartbft_test.go
pfi79: You're right, I'll redo it.

integration/smartbft/smartbft_test.go
```diff
@@ -1662,6 +1662,159 @@ var _ = Describe("EndToEnd Smart BFT configuration test", func() {
...
+				By(fmt.Sprintf("Launching %s", orderer.Name))
+				runner := network.OrdererRunner(orderer)
+				runner.Command.Env = append(runner.Command.Env, "FABRIC_LOGGING_SPEC=orderer.consensus.smartbft=debug:grpc=debug")
+				ordererRunners[i] = runner
+				proc := ifrit.Invoke(runner)
+				ordererProcesses[i] = proc
+				Eventually(proc.Ready(), network.EventuallyTimeout).Should(BeClosed())
+			}
+
+			// wait for a change in leader
```
C0rWin:
```suggestion
			// wait for the leader change
```

pfi79: fix

integration/smartbft/smartbft_test.go
pfi79: fix

integration/smartbft/smartbft_test.go
pfi79: rewritten

C0rWin (approved):

pfi79:
This pr conflicts with main. The reason is not clear. Created a new clean version.

PR #4445, opened 2023-09-18
Add orderer backoff

pfi79:
By default, the maximum time between attempts to connect one orderer to another via grpc is 2 minutes. And the waiting time for a heartbeat message from the leader in smartbft is 1 minute.
Therefore, it is possible to choose a time interval between the start of the leader and the follower, such that the follower will leave the consensus immediately.

This hotfix adds the ability to set the maximum time between connection attempts (ORDERER_GENERAL_BACKOFF_MAXDELAY).

Recommendations: if LeaderHeartbeatTimeout is 1m, then MaxDelay 20s.


orderer/common/server/main.go
```diff
@@ -442,6 +443,24 @@ func initializeClusterClientConfig(conf *localconfig.TopLevel) (comm.ClientConfi
 		UseTLS:            true,
 	}
 
+	if conf.General.Backoff.BaseDelay > 0 ||
```
yeasy, line 446: It seems that we do not need to check "conf.General.Backoff.BaseDelay > 0 || conf.General.Backoff.Multiplier > 0 || conf.General.Backoff.MaxDelay > 0", because we check each item's value later.

pfi79: It's a test, made only for another.
This is a trigger that shows "are we going to change the default mechanism at all or not"

integration/smartbft/smartbft_test.go
```diff
@@ -1577,6 +1572,87 @@ var _ = Describe("EndToEnd Smart BFT configuration test", func() {
...
+			By("invoking the chaincode")
+			invokeQuery(network, peer, network.Orderers[1], channel, 90)
+
+			numberKill := 3
+			orderer := network.Orderers[numberKill]
+			By(fmt.Sprintf("Killing %s", orderer.Name))
+			ordererProcesses[numberKill].Signal(syscall.SIGTERM)
+			Eventually(ordererProcesses[numberKill].Wait(), network.EventuallyTimeout).Should(Receive())
+
+			time.Sleep(time.Second * 240)
```
C0rWin: why are you sleeping here?

pfi79: This is the formula for calculating the time between attempts to connect to the orderer node
```go
        backoff, max := float64(bc.Config.BaseDelay), float64(bc.Config.MaxDelay)
	for backoff < max && retries > 0 {
		backoff *= bc.Config.Multiplier
		retries--
	}
	if backoff > max {
		backoff = max
	}
	// Randomize backoff delays so that if a cluster of requests start at
	// the same time, they won't operate in lockstep.
	backoff *= 1 + bc.Config.Jitter*(grpcrand.Float64()*2-1)
	if backoff < 0 {
		return 0
	}
```

Because of "Randomize backoff delays", it was difficult to select such a sleep time in the test that the error was guaranteed to occur. 
Experimentally, I found that the leader sleeps for 2 minutes between attempts to join the 3rd order.  And during these 2 minutes, the 3rd orderer manages to get up and not wait for a heartbeat from the leader for 1 minute

pfi79:
To make the test show a negative result, remove the line "ORDERER_GENERAL_BACKOFF_MAXDELAY=20s" from the test. Then it will work as it does now in main

integration/smartbft/smartbft_test.go
pfi79: If I remove or reduce the slip, the test will succeed in the current main.

Let's say my changes aren't there. Then this test will work like this:

| event | time | delta time | 
|---|---|---|
| Killing orderer4 ($\textcolor{green}{\textsf{START TIMEOUT}}$) | 0s | 0s |
| orderer1 disconnect from orderer4 | 0s | 0s |
| orderer1 attempt connect (first backoff) | 24s | 24s |
| orderer1 attempt connect | 31s | 7s |
| orderer1 attempt connect | 38s | 7s |
| orderer1 attempt connect | 45s | 7s |
| orderer1 attempt connect | 52s | 7s |
| orderer1 attempt connect | 64s | 12s |
| orderer1 attempt connect | 83s | 19s |
| orderer1 attempt connect | 115s | 32s |
| orderer1 attempt connect | 160s | 45s |
| orderer1 attempt connect | 240s | 80s |
| Launching orderer4 ($\textcolor{green}{\textsf{STOP TIMEOUT}}$) | 241s | 1s |
| orderer4 start | 242s | 1s |
| orderer4 heartbeat message waiting - begin | 246s | 1s |
| orderer4 heartbeat message waiting - end ($\textcolor{red}{\textsf{FAIL TEST}}$) | 306s | 60s |
| orderer1 attempt connect | 361s | 55s |

pfi79: @C0rWin Through customization, I reduced the sleep time to 60 seconds

integration/smartbft/smartbft_test.go
```diff
@@ -1577,6 +1572,91 @@ var _ = Describe("EndToEnd Smart BFT configuration test", func() {
...
+			By("invoking the chaincode")
+			invokeQuery(network, peer, network.Orderers[1], channel, 90)
+
+			numberKill := 3
+			orderer := network.Orderers[numberKill]
+			By(fmt.Sprintf("Killing %s", orderer.Name))
+			ordererProcesses[numberKill].Signal(syscall.SIGTERM)
+			Eventually(ordererProcesses[numberKill].Wait(), network.EventuallyTimeout).Should(Receive())
+
+			time.Sleep(time.Second * 60)
```
C0rWin: I guess the key reason to sleep here is because you are waiting for certain condition, why wouldn't you just change here to Eventually waiting for that condition? So at least it will be clear what are you doing here and why this sleep/wait is needed.

integration/smartbft/smartbft_test.go
C0rWin: > Experimentally, I found that the leader sleeps for 2 minutes between attempts to join the 3rd order. And during these 2 minutes, the 3rd orderer manages to get up and not wait for a heartbeat from the leader for 1 minute

Well, the problem I have with such a statement is that you are tested it locally, while CI runners might be affected differently, and we will end up adjusting this sleep interval over time.

pfi79: Yes you are right, but besides testing locally, I also tested in my fork in pipelines.
And it doesn't guarantee that this test will fail over time without my changes. Time is a delicate matter.
But right now this test shows exactly that you need to be able to manage backoff. 
And I also recommend that if LeaderHeartbeatTimeout is 1m, then MaxDelay backoff should be 20s.

integration/smartbft/smartbft_test.go
pfi79: Yes, I am waiting for the unmanaged backoff to reach the critical interval to trigger an orderer.
The whole point is that this event doesn't manifest itself in any way, anywhere. There are only indirect signs. Because we use default backoff values.

pfi79:
@C0rWin I took out the sleep. See if you like this option?
Also, in the [previous commit](https://github.com/hyperledger/fabric/actions/runs/6763011078/job/18379664027) I showed how the test will fail if the backoff is not controlled.
The current test in smartbft was successful.

Please restart unsuccessful Unit Tests.

integration/smartbft/smartbft_test.go
```diff
@@ -1577,6 +1572,94 @@ var _ = Describe("EndToEnd Smart BFT configuration test", func() {
...
+				"FABRIC_LOGGING_SPEC=orderer.consensus.smartbft=debug:grpc=debug",
+				"ORDERER_GENERAL_BACKOFF_MAXDELAY=3s",
+			)
+			ordererRunners[numberKill] = runner
+			proc := ifrit.Invoke(runner)
+			ordererProcesses[numberKill] = proc
+			Eventually(proc.Ready(), network.EventuallyTimeout).Should(BeClosed())
+
+			By("Waiting for followers to see the leader, again")
+			Eventually(ordererRunners[3].Err(), 15*time.Second, time.Second).Should(gbytes.Say("Message from 1 channel=testchannel1"))
```
C0rWin:
```suggestion
			Eventually(ordererRunners[numberKill].Err(), 15*time.Second, time.Second).Should(gbytes.Say("Message from 1 channel=testchannel1"))
```

integration/smartbft/smartbft_test.go
```diff
@@ -1577,6 +1572,94 @@ var _ = Describe("EndToEnd Smart BFT configuration test", func() {
...
+
+			By("invoking the chaincode, again")
```
C0rWin, line 1662: I guess this part is to make sure the node you have killed is participating in consensus again, while I am not sure whenever it actually is enough to state that node=3 is actually functioning

integration/smartbft/smartbft_test.go
```diff
@@ -1577,6 +1572,94 @@ var _ = Describe("EndToEnd Smart BFT configuration test", func() {
...
+			invokeQuery(network, peer, network.Orderers[2], channel, 80)
```
C0rWin:
```suggestion
			invokeQuery(network, peer, network.Orderers[numberKill], channel, 80)
```

Probably this will make sure the node is getting up to speed after being restarted.

integration/smartbft/smartbft_test.go
```diff
@@ -1577,6 +1572,94 @@ var _ = Describe("EndToEnd Smart BFT configuration test", func() {
...
+			By("invoking the chaincode")
+			invokeQuery(network, peer, network.Orderers[1], channel, 90)
+
+			numberKill := 3
+			orderer := network.Orderers[numberKill]
+			By(fmt.Sprintf("Killing %s", orderer.Name))
+			ordererProcesses[numberKill].Signal(syscall.SIGTERM)
+			Eventually(ordererProcesses[numberKill].Wait(), network.EventuallyTimeout).Should(Receive())
+
+			Eventually(ordererRunners[0].Err(), network.EventuallyTimeout, time.Second).Should(gbytes.Say("Closing: connection error: desc = \"error reading from server: EOF\""))
```
C0rWin, line 1639: Can you please add a comment? This is needed to ensure the leader doesn't see the node got disconnected, and then you wait to see attempts to reconnect to it. To make it clear?

integration/smartbft/smartbft_test.go
```diff
@@ -1577,6 +1572,94 @@ var _ = Describe("EndToEnd Smart BFT configuration test", func() {
...
+				Eventually(ordererRunners[0].Err(), network.EventuallyTimeout, time.Second).Should(gbytes.Say("pickfirstBalancer: UpdateSubConnState.*IDLE connection error"))
+			}
+
+			By(fmt.Sprintf("Launching %s", orderer.Name))
+			runner := network.OrdererRunner(orderer)
+			runner.Command.Env = append(
+				runner.Command.Env,
+				"FABRIC_LOGGING_SPEC=orderer.consensus.smartbft=debug:grpc=debug",
+				"ORDERER_GENERAL_BACKOFF_MAXDELAY=3s",
+			)
```
C0rWin: This is redundant, you can pass envs variables into `network.OrdererRunner(orderer, envs)`

C0rWin (changes requested):

integration/smartbft/smartbft_test.go
pfi79: It's more of a formal check that the system is working.

integration/smartbft/smartbft_test.go
pfi79: This will show that there is a connect from node 3 to node 0. But not the other way around. And in the test, we are trying to break the connectivity from node 0 to node 3.

But I agree with you, it's fairer to send the transaction to node 3.

integration/smartbft/smartbft_test.go
pfi79: Here I will also insert a check that the message "Heartbeat timeout expired" should not appear within 15 seconds

pfi79: fixed it

integration/smartbft/smartbft_test.go
pfi79: fixed it

integration/smartbft/smartbft_test.go
pfi79: fixed it

integration/smartbft/smartbft_test.go
pfi79: fixed it

C0rWin (approved):

PR #4449, opened 2023-09-26
Clean up duplicated check

yeasy:
Fix duplicated check of file existence and clean up.

Change-Id: If162b6fbfebffd493f345439d346881068916b74

#### Type of change

#### Description

The patchset removes a duplicated check of the existence of the metadata file, and directly uses the err, which is common following Go style.

#### Additional details

N/A

#### Related issues

N/A

#### Release Note

N/A

### DIFF
--- a/ccaas_builder/cmd/detect/main.go
+++ b/ccaas_builder/cmd/detect/main.go
@@ -28,7 +28,6 @@ func main() {
 	}
 
 	logger.Printf("::Type detected as ccaas")
-
 }
 
 type chaincodeMetadata struct {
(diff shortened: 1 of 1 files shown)


ccaas_builder/cmd/detect/main.go
```diff
@@ -42,26 +41,22 @@ func run() error {
...
+	// Check metadata file's existence
 	metadataFile := filepath.Join(chaincodeMetaData, "metadata.json")
-	_, err := os.Stat(metadataFile)
-	if err != nil {
-		return fmt.Errorf("%s not found ", metadataFile)
-	}
-
 	if _, err := os.Stat(metadataFile); err != nil {
 		return errors.WithMessagef(err, "%s not found ", metadataFile)
 	}
```
bestbeforetoday, line 48: Even this check seems pointless to me. It provides no guarantee that the file will still be there when it subsequently read. It would maybe be better to omit this check and just go ahead and read the file. You will get an error either way, and the (wrapped) error should contain some information on why the read failed.

C0rWin (approved):

ccaas_builder/cmd/detect/main.go
yeasy: Thanks for the comments. And keeping this check has no harm right now, hence let's leave it alone. There're other places needing optimized, I will propose more clean up later.

Issue #2814, opened 2021-08-09
Run dead link checker against Fabric docs


denyeart:
See history at https://jira.hyperledger.org/browse/FAB-18102

ggold7046:
Hi @denyeart , could you please guide me a little more about how to solve this issue as I'm new to hyperledger ?

denyeart:
@ggold7046 As a first pass, you could manually run a broken link checker against [Fabric ReadTheDocs latest version](https://hyperledger-fabric.readthedocs.io/en/latest/) and fix any broken links.

Ideally there would be a Github Action to run a broken link checker on a regular basis, for example https://github.com/marketplace/actions/my-broken-link-checker.

Any help on either approach would be appreciated.

yuriechan:
@denyeart 

Hi, I started by adding a github workflow for checking the broken links. 
I think introducing the process for continuously checking the broken links and fixing the broken links should be done in a separate effort. 
Would it be possible for you to review this PR when time allows? Thanks again. 
https://github.com/hyperledger/fabric/pull/4312

denyeart:
@yuriechan 
Thank you for adding the daily broken link checker.
We can now use the results to fix the broken links.

ggold7046:
> We can now use the results to fix the broken links.

@denyeart , could you tell me how these broken links are fixed ? Where would I find the broken links ?

denyeart:
@ggold7046 

Broken Link Checker runs nightly, see results here:
https://github.com/hyperledger/fabric/actions/workflows/broken-link-checker.yml

In the report, look for 404 not found errors. Here's an example:

```
https://hyperledger-fabric.readthedocs.io/en/latest/pluggable_endorsement_and_validation.html
	404	https://hyperledger.github.io/fabric-sdk-node/main/tutorial-chaincode-lifecycle.html
```

In this case, the problem is that Node SDK no longer supports chaincode deployment, and therefore the referenced doc page doesn't exist anymore. In this case the entire following sentence should be deleted:

```
If you are using the
Fabric SDK for Node.js, visit `How to install and start your chaincode <https://hyperledger.github.io/fabric-sdk-node/{BRANCH}/tutorial-chaincode-lifecycle.html>`__.
```

You could do a pull request for that one to get started.

Every occurrence of a broken link will have a different solution, I'd suggest to scan through the 404 errors and determine which ones you can fix quickly. Then the remaining broken links could be fixed by a maintainer who has more context.

[...]

PR #4451, opened 2023-09-29
Update pluggable_endorsement_and_validation.rst

ggold7046:
#2814

#### Type of change

- Documentation update

#### Description

Node SDK no longer supports chaincode deployment, and therefore the referenced doc page doesn't exist anymore. In this case the entire following sentence should be deleted.


denyeart (approved):
Perfect, thank you for resolving a dead link!

PR #4461, opened 2023-10-03
Improve error msg by using the error pkg

yeasy:
The patchset improves the error message.

Change-Id: I7d7bb084d0ba9d6d33a2b583a763752fca530b9b

#### Type of change

#### Description

The errors package has better features than the fmt package in terms of error processing. E.g., errors.Errorf records the stack trace at the point it is called.

On the other hand, the patchset eliminates the usage of fmt package.

#### Additional details

N/A

#### Related issues

N/A

#### Release Note

N/A

### DIFF
--- a/common/cauthdsl/policy.go
+++ b/common/cauthdsl/policy.go
@@ -7,8 +7,6 @@ SPDX-License-Identifier: Apache-2.0
 package cauthdsl
 
 import (
-	"fmt"
-
 	"github.com/golang/protobuf/proto"
 	cb "github.com/hyperledger/fabric-protos-go/common"
 	"github.com/hyperledger/fabric/common/policies"
@@ -32,11 +30,11 @@ func NewPolicyProvider(deserializer msp.IdentityDeserializer) policies.Provider 
 func (pr *provider) NewPolicy(data []byte) (policies.Policy, proto.Message, error) {
 	sigPolicy := &cb.SignaturePolicyEnvelope{}
 	if err := proto.Unmarshal(data, sigPolicy); err != nil {
-		return nil, nil, fmt.Errorf("Error unmarshalling to SignaturePolicy: %s", err)
+		return nil, nil, errors.Wrap(err, "error unmarshalling to SignaturePolicy")
 	}
 
 	if sigPolicy.Version != 0 {
-		return nil, nil, fmt.Errorf("This evaluator only understands messages of version 0, but version was %d", sigPolicy.Version)
+		return nil, nil, errors.Errorf("this evaluator only understands messages of version 0, but version was %d", sigPolicy.Version)
 	}
 
 	compiled, err := compile(sigPolicy.Rule, sigPolicy.Identities)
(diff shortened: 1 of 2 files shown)


ale-linux (changes requested):
The [errors](https://github.com/pkg/errors) package is no longer maintained and we should stop using it. Since a couple of releases, Go has the verb `%w` that can be used when supplying an error to produce a meaningful message that includes `Unwrap` information.

denyeart:
While I agree with @ale-linux in theory since Go added better error wrapping support in `errors` standard library as of  Go 1.13, in practice the Fabric codebase makes widespread use of `github.com/pkg/errors`, including the use of `%+v` formatting in higher level error logging to print the stack trace obtained from lower level `github.com/pkg/errors` generated errors. This feature is specific to `github.com/pkg/errors` and has been helpful when troubleshooting errors.

The Fabric doc also advises developers to use `github.com/pkg/errors`:
https://hyperledger-fabric.readthedocs.io/en/latest/error-handling.html
(although this guidance should really be in the [developer coding guidelines](https://hyperledger-fabric.readthedocs.io/en/latest/style-guides/go-style.html) rather than the user docs).

Maybe we should open a future issue to convert `github.com/pkg/errors` usage to `errors` throughout the code, and accept this PR in the meantime.

common/cauthdsl/policy.go
```diff
@@ -32,11 +30,11 @@ func NewPolicyProvider(deserializer msp.IdentityDeserializer) policies.Provider
 func (pr *provider) NewPolicy(data []byte) (policies.Policy, proto.Message, error) {
 	sigPolicy := &cb.SignaturePolicyEnvelope{}
 	if err := proto.Unmarshal(data, sigPolicy); err != nil {
-		return nil, nil, fmt.Errorf("Error unmarshalling to SignaturePolicy: %s", err)
+		return nil, nil, errors.Errorf("error unmarshalling to SignaturePolicy: %s", err)
```
denyeart: If we do stick with `github.com/pkg/errors` for now, I think this one should use errors.Wrap() rather than errors.Errorf() as described at https://hyperledger-fabric.readthedocs.io/en/latest/error-handling.html.

yeasy: Good point, let me update and clean up other places.

yeasy:
> While I agree with @ale-linux in theory since Go added better error wrapping support in `errors` standard library as of Go 1.13, in practice the Fabric codebase makes widespread use of `github.com/pkg/errors`, including the use of `%+v` formatting in higher level error logging to print the stack trace obtained from lower level `github.com/pkg/errors` generated errors. This feature is specific to `github.com/pkg/errors` and has been helpful when troubleshooting errors.
> 
> The Fabric doc also advises developers to use `github.com/pkg/errors`: https://hyperledger-fabric.readthedocs.io/en/latest/error-handling.html (although this guidance should really be in the [developer coding guidelines](https://hyperledger-fabric.readthedocs.io/en/latest/style-guides/go-style.html) rather than the user docs).
> 
> Maybe we should open a future issue to convert `github.com/pkg/errors` usage to `errors` throughout the code, and accept this PR in the meantime.

This will need lots of efforts. Do we need to wait till the `errors` package become more mature?

denyeart:
The `errors` package in standard library is mature, we just need to make a project decision on whether to transition to it. I've opened issue #4468 with more background so that we can discuss a plan in more detail.

My suggestion would be to not make these error handling changes until we've got some more opinions gathered in #4468.

Issue #4468, opened 2023-10-05
Transition away from archived package `github.com/pkg/errors`


denyeart:
### Current Status

The Fabric project intentionally adopted use of `github.com/pkg/errors` package for error handling and publicized the guidance at:
https://hyperledger-fabric.readthedocs.io/en/latest/error-handling.html

The rationale at the time was that `github.com/pkg/errors` injected stack trace context at the point of error in low level code, which could be easily logged at higher levels of the control flow with the `%+v` formatter verb, and layers of code inbetween could add message context using WithMessage(), or simply pass the existing stack trace context up with the familiar and simple error handling like this:

```
if err != nil {
   return err
}
```

Since that time, Go added better error wrapping support in `errors` standard library as of Go 1.13 as described in the [blog](https://go.dev/blog/go1.13-errors) and [wiki](https://github.com/golang/go/wiki/ErrorValueFAQ), and `github.com/pkg/errors` got archived as not needed anymore.

### Goal

The project should transition away from the archived package `github.com/pkg/errors`.

I see two basic approaches:

#### Approach 1 - Switch to `errors` package in standard library

It makes logical sense to use the standard `errors` package now, however without the stack trace context available anymore, each layer that passes an error up should add its own context, either by extending the error message or using the new wrappering approach available in standard library `errors` package.

The downside is that transitioning to the standard library `errors` package will require a large work effort to update all the code, as we really don't want a mismatch of error handling assumptions and techniques across a single code flow. The best time to make an exhaustive change like this would be at a major release boundary, for example between v2.x and v3.x.

#### Approach 2 - Keep overall approach by copying `github.com/pkg/errors` into a fabric `errors` package

If we want to preserve the stack trace and limit the code changes, there isn't much risk of keeping the overall `github.com/pkg/errors` approach. Although the project is archived, the project was simple and didn't pull in other dependencies. The utility code could easily be copied into a fabric specific `errors` package to eliminate the external dependency on `github.com/pkg/errors`, while keeping the existing overall approach and the associated benefits of having stack trace available.

This approach would be simpler in the short term, but would come at the cost of maintaining the (small amount) of utility code, and educating developers to use it.

The purpose of this issue is to finalize an approach and then update the developer guidance and code accordingly.

ale-linux:
[...]

PR #4470, opened 2023-10-05
Use the errors.Is to check error type

yeasy:
The patchset uses errors.Is to check error type.

Change-Id: Ieae0eaa2527d05fa67aa31afdda42d150fff0980

#### Type of change

#### Description

errors.Is is recommened since Go 1.13 to check error type. The error may be wrapped.

More details: https://github.com/golang/go/wiki/ErrorValueFAQ.

#### Additional details

N/A

#### Related issues

N/A

#### Release Note

N/A

### DIFF
--- a/bccsp/pkcs11/pkcs11.go
+++ b/bccsp/pkcs11/pkcs11.go
@@ -664,7 +664,7 @@ func (csp *Provider) verifyP11ECDSA(ski []byte, msg []byte, R, S *big.Int, byteS
 		return false, fmt.Errorf("PKCS11: Verify-initialize [%s]", err)
 	}
 	err = csp.ctx.Verify(session, msg, sig)
-	if err == pkcs11.Error(pkcs11.CKR_SIGNATURE_INVALID) {
+	if errors.Is(err, pkcs11.Error(pkcs11.CKR_SIGNATURE_INVALID)) {
 		return false, nil
 	}
 	if err != nil {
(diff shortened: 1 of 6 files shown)


denyeart:
In many of these files the `errors` package is actually the `github.com/pkg/errors` package which is archived and we are considering transitioning away from it.  I'm thinking we shouldn't increase the usage of the archived package at this time, let's see what comes out of the discussion in #4468.

yeasy:
> In many of these files the `errors` package is actually the `github.com/pkg/errors` package which is archived and we are considering transitioning away from it. I'm thinking we shouldn't increase the usage of the archived package at this time, let's see what comes out of the discussion in #4468.

Correct. 
However, the `Is()` method is supported by the official `errors` pkg if you look at https://github.com/golang/go/wiki/ErrorValueFAQ. This patchset does not change the pkg name (we can migrate `github.com/pkg/errors` to `errors` later without changing the `Is()`) now because of the discussions in #4468.

denyeart (approved):
Ok, I agree this is moving in the correct direction, merging.
But I still think we should give #4468 some more time before opening additional PRs.

yeasy:
> Ok, I agree this is moving in the correct direction, merging. But I still think we should give #4468 some more time before opening additional PRs.

Agree. Once we made the decision on #4468, I will continue the migration work.

PR #4477, opened 2023-10-17
prealloc slice in rwset builder

cendhu:
#### Type of change

#### Description

This commit defines capacity for slices used in
rwset building to avoid reallocation of memory, and copy of contents.


denyeart (approved):
Good improvement!

Issue #4428, opened 2023-09-07
Extend to lifecycle chaincode checkcommitreadiness to provide details of the discrepancies


satota2:
### Current Status

The `peer lifecycle chaincode checkcommitreadiness` command provides a way to check whether channel members have approved the same chaincode definition. Currently, this command only outputs a true/false result, indicating if the specified chaincode definition parameters match those approved by each organization.
Since the command output lacks detailed information about mismatches, Fabric administrators often find it time-consuming to identify the root cause of discrepancies.

Real-world issues, like the one discussed in https://github.com/hyperledger/fabric/issues/3330, show that Fabric administrators may face difficulties in troubleshooting discrepancies.

### Expected

It is expected that `peer lifecycle chaincode checkcommitreadiness` provides additional information to  diagnose why a chaincode definition's parameters and each organization's approval status are mismatched. This will facilitate root cause analysis and streamline inter-organizational coordination during operations.

#### Solution
A possible solution is to add a new `--inspect` option to `checkcommitreadiness` subcommand under `peer lifecycle chaincode`. This option outputs additional information to identify the cause when the approval from each organization is `false`.

*The example of executing checkcommitreadiness with inspection and JSON output format:*
```
$ peer lifecycle chaincode checkcommitreadiness --channelID [channel_name] --name [chaincode_name] --version 2.0 --init-required --sequence 2 --output json --inspect
{
   "Approvals": {
      "Org1MSP": {
        "result": true,
      },
      "Org2MSP": {
        "result": false,
        "mismatch": ["EndorsementInfo (Check the Version, InitRequired, EndorsementPlugin)", "ValidationInfo (Check the ValidationParameter, ValidationPlugin)",
        "Collections"]
      },
      "Org3MSP": {
        "result": false,
        "mismatch": ["Metadata (Check the Sequence, ChaincodeName)"]
      },
   }
}
```

*The example of executing checkcommitreadiness with inspection and Plain output format:*
```
$ peer lifecycle chaincode checkcommitreadiness --channelID [channel_name] --name [chaincode_name] --version 2.0 --init-required --sequence 2 --inspect
Chaincode definition for chaincode '[chaincode_name]', version '2.0', sequence '2' on channel
'[channel_name]' approval status by org:
Org1MSP: true
Org2MSP: false (mismatch: EndorsementInfo, ValidationInfo, Collections)
Org3MSP: false (mismatch: Metadata)
```

#### Detailed design
[...]

PR #4483, opened 2023-10-18
Extend integration tests to cover inspect option of checkcommitreadiness

satota2:
#### Type of change

- Test update

#### Description

This patch extends integration tests to cover inspect option of checkcommitreadiness.

#### Related issues

#4428

### DIFF
--- a/integration/nwo/deploy.go
+++ b/integration/nwo/deploy.go
@@ -73,8 +73,16 @@ func DeployChaincode(n *Network, channel string, orderer *Orderer, chaincode Cha
 	// approve for each org
 	ApproveChaincodeForMyOrg(n, channel, orderer, chaincode, peers...)
 
-	// commit definition
+	// wait for checkcommitreadiness returns ready status
 	CheckCommitReadinessUntilReady(n, channel, chaincode, n.PeerOrgs(), peers...)
+
+	// after the chaincode definition has been correctly approved for each org,
+	// demonstrate the capability to inspect the discrepancies in the chaincode definitions
+	// by executing checkcommitreadiness with inspect flag,
+	// with intentionally altered values for chaincode definition parameters
+	InspectChaincodeDiscrepancies(n, channel, chaincode, n.PeerOrgs(), peers...)
+
+	// commit definition
 	CommitChaincode(n, channel, orderer, chaincode, peers[0], peers...)
 
 	// init the chaincode, if required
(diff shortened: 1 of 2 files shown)


integration/nwo/deploy.go
```diff
@@ -73,8 +73,13 @@ func DeployChaincode(n *Network, channel string, orderer *Orderer, chaincode Cha
 	// approve for each org
 	ApproveChaincodeForMyOrg(n, channel, orderer, chaincode, peers...)
 
-	// commit definition
+	// wait for checkcommitreadiness returns ready status
 	CheckCommitReadinessUntilReady(n, channel, chaincode, n.PeerOrgs(), peers...)
+
+	// inspect discrepancies between a wrong chaincode definition and each org
+	InspectDiscrepancies(n, channel, chaincode, n.PeerOrgs(), peers...)
```
denyeart: I'm not following this... if the chaincode was installed and approved for each org, why would there be discrepancies at this step? Could you add some comments explaining it?

C0rWin: IMO, adding an actual test that will utilize this function will demonstrate the intent behind it.

satota2: @denyeart @C0rWin Thank you for your review and comment.

Aiming to emulate a more realistic case, I have updated the test scenario. As a step prior to making the correct approvals, the test approves a chaincode definition with some incorrect parameters intentionally. Following this, the test utilizes the checkcommitreadiness command with the inspect option to check and demonstrate the ability to identify these discrepancies.

Does this address your concerns?

integration/nwo/deploy.go
```diff
@@ -506,6 +514,69 @@ func checkCommitReadiness(n *Network, peer *Peer, channel string, chaincode Chai
...
+func ApproveIncorrectChaincodeForEachOrg(n *Network, channel string, orderer *Orderer, chaincode Chaincode, peers ...*Peer) {
+	// used to ensure we only approve once per org
+	approvedOrgs := map[string]bool{}
+	for _, p := range peers {
+		if _, ok := approvedOrgs[p.Organization]; !ok {
+			sess, err := n.PeerAdminSession(p, commands.ChaincodeApproveForMyOrg{
+				ChannelID:           channel,
+				Orderer:             n.OrdererAddress(orderer, ListenPort),
+				Name:                chaincode.Name,
+				Version:             "incorrect-version", // Intentionally set incorrect version
```
denyeart: Much improved, the test now shows the intent of the new inspect option.

I still have a question though... if all the orgs approve the same incorrect version, won't there be no discrepancies to detect? Maybe I'm missing something...

satota2: @denyeart In this case, I assume that the organization executing commitreadiness has also specified the incorrect parameters in its own approve  mistakenly.
Since the correct chaincode definition is specified in the argument to commitreadiness, all organizations will detect the mismatch.

Other scenario ideas are a case where the organization executing commitreadiness with inspect only specifies the correct version, or where only one organization specifies incorrect version. In these scenarios, the test code may become a bit more complex.

denyeart: Ah right, they all approve the same incorrect definition, but checkcommitreadiness checks for the correct definition. Looks good!

denyeart (dismissed):

denyeart (commented):
Ok, I see why other tests are failing. They use the nwo.DeployChaincode() and make assumptions, for example the number of blocks that were required to deploy the chaincode, which is now changed. We don't want to run these extra checks every time a test calls DeployChaincode().

So I'd suggest one of the following:
- Go back to the original PR approach. Now that I understand the discrepancy wasn't between the orgs, it was between the approved chaincode definition and the checkcommitreadiness (incorrect) inputs, I am fine with the approach. I'd just suggest adding some comments to clarify it so that others are not confused.
- Add a new test for the discrepancy instead of leveraging the existing DeployChaincode() function. The downside is that it will extend the overall test execution time since chaincode deployment is the most time consuming aspect across the tests.

satota2:
@denyeart 
Thank you for your feedback. I wasn't aware of the dependency between DeployChaincode() and other tests, and I apologize for the trouble. I have chosen your suggested approach 1 and updated the PR accordingly. I have also added some comments to clarify the scenario, to prevent any confusion for others in the future.I appreciate your review on the updated PR. Thank you.

denyeart (approved):
Thank you!

Issue #3983, opened 2023-02-01
Fabric v3: Remove legacy chaincode (LSCC) support


denyeart:
Fabric v2.x supports both the v1.x chaincode lifecycle and the v2.x chaincode lifecycle. The proposal here is to remove support for the v1.x chaincode lifecycle for Fabric v3.x.

There are many places in the code that have dual support for v1.x and v2.x chaincode lifecycle. Each of those locations will need to be assessed to remove the v1.x chaincode lifecycle support, and then the v1.x chaincode lifecycle itself will need to be removed.

Sayalikukkar:
Hi @denyeart, I would like to contribute on this. Could you please guide me?

denyeart:
@Sayalikukkar This will be a complex epic as the v1.x chaincode lifecycle is embedded in many places. The first thing to do would be to decompose the work into smaller issues under this epic.

@ale-linux led the work to add dual support in Fabric v2.x, so probably best to get his high level thoughts first.

C0rWin:
> @Sayalikukkar This will be a complex epic as the v1.x chaincode lifecycle is embedded in many places. The first thing to do is decompose the work into smaller issues under this epic.
> 
> @ale-linux led the work to add dual support in Fabric v2.x, so probably best to get his high level thoughts first.

Hi @denyeart, I would like to take the lead on this if you are fine. Several things have to be done to get it complete:

1. First, we need to remove it from the integration tests cause it's still being widely used there. We need to make sure to switch all places we have legacy lifecycle used.
2. Once integration tests are fixed, we can get rid of the legacy lifecycle from the NWO APIs
3. The next step would be to make sure to adhere to SDKs, to make sure it's no longer available 
4. Take care of fabric samples. Almost sure it's not used, but to make sure we won't break anything
5. Remove from peer cli, so no one will be able to use it any longer
6. Refactor system chaincodes to stop referring to legacy Lifecycle APIs
7. Remove legacy.

I can prepare a more detailed plan on how to do it in small portions.

denyeart:
Thanks @C0rWin , that would be great! Perhaps @Sayalikukkar could help on some of the tasks.

I mentioned @ale-linux because I remember there was some prior design thought around removing the old lifecycle... I've found the presentation where @ale-linux wrote this down in 2019:

- Long term we expect that LSCC will disappear
- For this reason, whenever a fabric component requires lifecycle data we have
  - Created appropriate interfaces (removing any direct access to LSCC)
[...]

PR #4488, opened 2023-10-22
Removing nwo.DeployChaincodeLegacy()

C0rWin:
This commit removes the DeployChaincodeLegacy function from the integration tests suite to address #3983 as part of eliminating the legacy chaincode lifecycle. Due to the changes in size, the work will be split into several PRs. This is the first one in a row.

### DIFF
--- a/integration/devmode/devmode_test.go
+++ b/integration/devmode/devmode_test.go
@@ -34,7 +34,6 @@ var _ = Describe("Devmode", func() {
 		ordererRunner               *ginkgomon.Runner
 		ordererProcess, peerProcess ifrit.Process
 		chaincode                   nwo.Chaincode
-		legacyChaincode             nwo.Chaincode
 		chaincodeRunner             *ginkgomon.Runner
 		chaincodeProcess            ifrit.Process
 		channelName                 string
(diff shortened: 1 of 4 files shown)


integration/devmode/devmode_test.go
```diff
@@ -83,54 +82,6 @@ var _ = Describe("Devmode", func() {
...
-		nwo.InstallChaincodeLegacy(network, legacyChaincode, org1peer0)
-
-		By("instantiating the chaincode")
-		nwo.InstantiateChaincodeLegacy(network, channelName, orderer, legacyChaincode, org1peer0, org1peer0)
-
-		By("querying and invoking the chaincode")
-		RunQueryInvokeQuery(network, orderer, org1peer0, channelName, 100)
-		Eventually(chaincodeRunner).Should(gbytes.Say("invoking in devmode"))
-	})
-
```
C0rWin, line 133: This one is no longer needed cause there same set of tests exists for the new lifecycle, and since we are about to deprecate the legacy lscc this part could be removed.

denyeart (approved):
Thanks @C0rWin , appreciate the progress on the legacy lifecycle removal!

PR #4489, opened 2023-10-24
Update docs on checkcommitreadiness with inspect flag

### LINKED ISSUE #4428: Extend to lifecycle chaincode checkcommitreadiness to provide details of the discrepancies

satota2:
#### Type of change

- Documentation update

#### Description

This patch updates documents on `checkcommitreadiness` with `inspect` flag.

#### Related issues

Resolve #4428

### DIFF
--- a/docs/source/commands/peerlifecycle.md
+++ b/docs/source/commands/peerlifecycle.md
@@ -705,6 +705,61 @@ definition can be committed to a channel.
     }
     ```
 
+  * You can also use the `--inspect` flag to output additional information to identify the cause when the approval from each organization is false. This will facilitate root cause analysis and streamline inter-organizational coordination during operations.
+
+    ```
+    peer lifecycle chaincode checkcommitreadiness --channelID mychannel --name basic --version 1.0 --sequence 1 --inspect
+    ```
+
+    If successful, the command will output additional information to identify the cause when the approval from each organization is false.
+    This example outputs that there is a mismatch with respect to the organization Org2MSP and Org3MSP, respectively,
+    compared to the chaincode definition specified in the arguments of checkcommitreadiness.
+
+    ```
+    Chaincode definition for chaincode 'basic', version '1.0', sequence '1' on channel 'mychannel' approval status by org:
+    Org1MSP: true
+    Org2MSP: false (mismatch: [EndorsementInfo, ValidationInfo, Collections])
+    Org3MSP: false (mismatch: [ChaincodeParameters])
+    ```
+
+    You can also use the `--inspect` flag and `--output` flag simultaneously.
+
+    ```
+    peer lifecycle chaincode checkcommitreadiness --channelID mychannel --name basic --version 1.0 --sequence 1 --inspect
+    ```
+
+    If successful, the command will return a JSON format with mismatches.
+
+    ```
+    {
+      "approvals": {
+        "Org1MSP": true,
+        "Org2MSP": false,
+        "Org3MSP": false
+      },
+      "mismatches": {
+        "Org2MSP": {
+          "items": [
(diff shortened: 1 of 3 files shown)


denyeart (approved):
Nice documentation, thank you!

PR #4490, opened 2023-10-25
Verify transactions in a block are well formed

yacovm:

### DIFF
--- a/protoutil/blockutils.go
+++ b/protoutil/blockutils.go
@@ -10,6 +10,7 @@ import (
 	"bytes"
 	"crypto/sha256"
 	"encoding/asn1"
+	"encoding/base64"
 	"fmt"
 	"math/big"
 
@@ -296,5 +297,41 @@ func searchConsenterIdentityByID(consenters []*cb.Consenter, identifier uint32) 
 			})
 		}
 	}
+	return nil
+}
+
+func VerifyTransactionsAreWellFormed(block *cb.Block) error {
+	if block == nil || block.Data == nil || len(block.Data.Data) == 0 {
+		return fmt.Errorf("empty block")
+	}
+
+	for i, rawTx := range block.Data.Data {
+		env := &cb.Envelope{}
+		if err := proto.Unmarshal(rawTx, env); err != nil {
+			return fmt.Errorf("transaction %d is invalid: %v", i, err)
+		}
+
+		if len(env.Payload) == 0 {
+			return fmt.Errorf("transaction %d has no payload", i)
+		}
+
+		if len(env.Signature) == 0 {
+			return fmt.Errorf("transaction %d has no signature", i)
+		}
+
+		expected, err := proto.Marshal(env)
+		if err != nil {
+			return fmt.Errorf("failed re-marshaling envelope: %v", err)
+		}
+
+		if len(expected) < len(rawTx) {
+			return fmt.Errorf("transaction %d has %d trailing bytes", i, len(rawTx)-len(expected))
+		}
+		if !bytes.Equal(expected, rawTx) {
+			return fmt.Errorf("transaction %d (%s) does not match its raw form (%s)", i,
+				base64.StdEncoding.EncodeToString(expected), base64.StdEncoding.EncodeToString(rawTx))
+		}
+	}
+
 	return nil
 }
--- a/integration/smartbft/smartbft_test.go
+++ b/integration/smartbft/smartbft_test.go
@@ -1081,7 +1081,7 @@ var _ = Describe("EndToEnd Smart BFT configuration test", func() {
 
 			By("Sending TX with corrupted signature")
 			env = ordererclient.CreateBroadcastEnvelope(network, leader, channel, []byte("MESSSAGE_2"))
-			env.Signature = []byte{}
+			env.Signature = []byte{1, 2, 3}
 			resp, err = ordererclient.Broadcast(network, leader, env)
 			Expect(err).NotTo(HaveOccurred())
 			Expect(resp.Status).To(Equal(common.Status_FORBIDDEN))
@@ -1383,7 +1383,7 @@ var _ = Describe("EndToEnd Smart BFT configuration test", func() {
 				0,
 			)
 			Expect(err).NotTo(HaveOccurred())
-			env.Signature = []byte{}
+			env.Signature = []byte{1, 2, 3}
 
 			By("Authenticate to the leader step service")
 			err = client.Auth()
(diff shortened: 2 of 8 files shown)


protoutil/blockutils.go
```diff
@@ -298,3 +299,35 @@ func searchConsenterIdentityByID(consenters []*cb.Consenter, identifier uint32)
 	}
 	return nil
 }
+
+func VerifyTransactionsAreWellFormed(block *cb.Block) error {
+	if block == nil || block.Data == nil || len(block.Data.Data) == 0 {
+		return nil
+	}
+
+	for i, rawTx := range block.Data.Data {
```
pfi79: If you do `range block.Data.Data`, then the check `len(block.Data.Data) == 0` is unnecessary

denyeart:
At least a small PR description would be nice.

yacovm:
> At least a small PR description would be nice.

Added a description

protoutil/blockutils.go
```diff
@@ -298,3 +299,35 @@ func searchConsenterIdentityByID(consenters []*cb.Consenter, identifier uint32)
 	}
 	return nil
 }
+
+func VerifyTransactionsAreWellFormed(block *cb.Block) error {
+	if block == nil || block.Data == nil {
+		return nil
+	}
```
tock-ibm: I think that this part should return an error, and also check against len(block.Data.Data) == 0.

In mcs.go, in the peer, the invocation of this function is preceded by a call to GetChannelIDFromBlock  that checks that. 

```
// GetChannelIDFromBlock returns channel ID in the block
func GetChannelIDFromBlock(block *cb.Block) (string, error) {
	if block == nil || block.Data == nil || block.Data.Data == nil || len(block.Data.Data) == 0 {
		return "", errors.New("failed to retrieve channel id - block is empty")
	}
...
}
```

However, in the orderer, in `orderer/common/cluster/util.go`

VerifyBlocksBFT -> verifyBlockSequence -> VerifyBlockHash -> VerifyTransactionsAreWellFormed

there is no such provision and a block with no data or no transactions passes to the next stage.
This causes a block with `block.Data = nil` to crash the orderer in VerifyBlockHash.

e.g. add this test case to `TestVerifyBlockHash`:
```
{
			name:          "block with nil data",
			errorContains: "sequences 12 and 666 were received consecutively",
			mutateBlockSequence: func(blockSequence []*common.Block) []*common.Block {
				blockSequence[len(blockSequence)/2].Data = nil
				return blockSequence
			},
		},
```
 
I think it is better to assume that in a block w/o TXs (nil Data etc) the TXs are NOT well formed.

protoutil/blockutils.go
```diff
@@ -298,3 +299,35 @@ func searchConsenterIdentityByID(consenters []*cb.Consenter, identifier uint32)
...
+
+		if len(env.Payload) == 0 {
+			return fmt.Errorf("transaction %d has no payload", i)
+		}
+
+		if len(env.Signature) == 0 {
+			return fmt.Errorf("transaction %d has no signature", i)
+		}
+
+		expected := MarshalOrPanic(env)
```
tock-ibm: The input to this loop is a rawTx supplied by a foreign process that may be malicious.
It is therefore not good form to throw panic in response to anything that comes from a foreign party.
Here the risk is low, as there are very few cases in proto3  in which proto.Marshal a message causes an error. This however may change in the future depending on the message and implementation. For example in proto2  a missing required field in the message would cause that (but we are proto3). 
In the code base MarshalOrPanic is used mainly in tests, or when the same process produced the message and now marshals it. 
This is a nit, but in my opinion, since it is so easy to fix, I would do it.

protoutil/blockutils.go
```diff
@@ -298,3 +299,39 @@ func searchConsenterIdentityByID(consenters []*cb.Consenter, identifier uint32)
 	}
 	return nil
 }
+
+func VerifyTransactionsAreWellFormed(block *cb.Block) error {
+	if block == nil || block.Data == nil || len(block.Data.Data) == 0 {
+		return fmt.Errorf("empty block")
```
ale-linux, line 305: while it's generally good to have such check, I'm thinking that a nil block, or a block with nil data wouldn't ever reach this function. Are we able to actually exercise this line in any test?

protoutil/blockutils.go
```diff
@@ -298,3 +299,39 @@ func searchConsenterIdentityByID(consenters []*cb.Consenter, identifier uint32)
...
+		if err != nil {
+			return fmt.Errorf("failed re-marshaling envelope: %v", err)
+		}
+
+		if len(expected) < len(rawTx) {
+			return fmt.Errorf("transaction %d has %d trailing bytes", i, len(rawTx)-len(expected))
+		}
+		if !bytes.Equal(expected, rawTx) {
+			return fmt.Errorf("transaction %d (%s) does not match its raw form (%s)", i,
+				base64.StdEncoding.EncodeToString(expected), base64.StdEncoding.EncodeToString(rawTx))
```
ale-linux, line 332: why not hex encode?

ale-linux (commented):
LGTM overall. The only thing that isn't clear yet is how we can convince ourselves that we're calling `VerifyTransactionsAreWellFormed` in all the right places to achieve complete mediation. I see that we call it in 3 places - how can we be sure these are the only three places?

tock-ibm (approved):

protoutil/blockutils.go
tock-ibm: It does... see my comment above. 
https://github.com/hyperledger/fabric/pull/4490#discussion_r1373235343

protoutil/blockutils.go
yacovm: It's less characters this way

protoutil/blockutils.go
yacovm: > while it's generally good to have such check, I'm thinking that a nil block, or a block with nil data wouldn't ever reach this function. Are we able to actually exercise this line in any test?

In a previous incarnation of this PR it would return nil. Yoav asked me to change to return error.

ale-linux: ah very good - I didn't see the comment because the conversation was marked as resolved. Thanks!

yacovm:
> LGTM overall. The only thing that isn't clear yet is how we can convince ourselves that we're calling `VerifyTransactionsAreWellFormed` in all the right places to achieve complete mediation. I see that we call it in 3 places - how can we be sure these are the only three places?

Not only that we can't, but there is a stack of PRs pending to be reviewed that will need to be re-evaluated that we call it in different places. 

The right approach would be to change the function that computes the hash of the block, to receive as parameter the entire block, not just the header.

tock-ibm:
> > LGTM overall. The only thing that isn't clear yet is how we can convince ourselves that we're calling `VerifyTransactionsAreWellFormed` in all the right places to achieve complete mediation. I see that we call it in 3 places - how can we be sure these are the only three places?
> 
> Not only that we can't, but there is a stack of PRs pending to be reviewed that will need to be re-evaluated that we call it in different places.
> 
> The right approach would be to change the function that computes the hash of the block, to receive as parameter the entire block, not just the header.

I'll take this commit and incorporate it into my stack of PRs, here: 
https://github.com/hyperledger/fabric/issues/4346
https://github.com/hyperledger/fabric/pull/4408

yacovm:
@denyeart we can backport it now

PR #4492, opened 2023-10-25
Use append to concatenate bytes

yeasy:
Change-Id: I09e552881dab2a02a0b2b40f67d9f597975d90b4

#### Type of change

#### Description

Improve the performance of `ConcatenateBytes`.

The patchset simplify the code of method `ConcatenateBytes`.

And the speed can increase by about 12%.

#### Additional details

N/A

#### Related issues

N/A

#### Release Note

N/A

### DIFF
--- a/common/util/utils.go
+++ b/common/util/utils.go
@@ -81,6 +81,7 @@ func ToChaincodeArgs(args ...string) [][]byte {
 
 // ConcatenateBytes is useful for combining multiple arrays of bytes, especially for
 // signatures or digests over multiple fields
+// This way is more efficient in speed
 func ConcatenateBytes(data ...[]byte) []byte {
 	finalLength := 0
 	for _, slice := range data {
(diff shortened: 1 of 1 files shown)


yacovm (changes requested):
It simplifies but now it's less efficient this way because there is more memory allocation.

common/util/utils.go
```diff
@@ -82,17 +82,9 @@ func ToChaincodeArgs(args ...string) [][]byte {
...
 // signatures or digests over multiple fields
 func ConcatenateBytes(data ...[]byte) []byte {
-	finalLength := 0
+	var result []byte
 	for _, slice := range data {
-		finalLength += len(slice)
-	}
-	result := make([]byte, finalLength)
-	last := 0
-	for _, slice := range data {
```
yacovm, line 91: if you want to simplify this implementation then you can use the `copy()` built-in directive instead of this for loop

yeasy: Use copy() in the new version

yacovm (approved):

PR #4497, opened 2023-10-26
fix integration raft test

pfi79:
I think I've found the cause of raft itegration tests freezing


tock-ibm:
> I think I've found the cause of raft itegration tests freezing

Can you please elaborate a bit on why you think this is the case?

pfi79:
> Can you please elaborate a bit on why you think this is the case?

I can see I'm already wrong.
I'll keep looking.

The symptoms are as follows:
- The tests are freezing. Most likely, the test is waiting for an event that never occurs. And this waiting has no time limit.
- The fact that there are no messages during the hang and when the test is interrupted most likely blocks the functions of the gomega package.
- having looked through all the hangs for the last 3 months, I know for sure that the cause of the hang is in the file github.com/hyperledger/fabric/integration/raft/config_test.go

pfi79:
> > I think I've found the cause of raft itegration tests freezing
> 
> Can you please elaborate a bit on why you think this is the case?

All variants of using orderer.Err() happen via the Eventually method and only here was working via Expect (without time limit). I don't know if there could be a bug here, but I decided to redo it uniformly.

I also found that channelparticipation.Remove() can hang on execution forever.

I haven't fixed the bugs. I'm just trying to make it so that the pipline doesn't hang for 6 hours.

Have you found all the places where tests can hang? I'm not sure.

tock-ibm:
> > > I think I've found the cause of raft itegration tests freezing
> > 
> > 
> > Can you please elaborate a bit on why you think this is the case?
> 
> All variants of using orderer.Err() happen via the Eventually method and only here was working via Expect (without time limit). I don't know if there could be a bug here, but I decided to redo it uniformly.
> 
> I also found that channelparticipation.Remove() can hang on execution forever.
> 
> I haven't fixed the bugs. I'm just trying to make it so that the pipline doesn't hang for 6 hours.
> 
> Have you found all the places where tests can hang? I'm not sure.

Thanks for looking for these hangs. They are indeed very annoying.

pfi79:
> > > > I think I've found the cause of raft itegration tests freezing
> > > 
> > > 
> > > Can you please elaborate a bit on why you think this is the case?
> > 
> > 
> > All variants of using orderer.Err() happen via the Eventually method and only here was working via Expect (without time limit). I don't know if there could be a bug here, but I decided to redo it uniformly.
> > I also found that channelparticipation.Remove() can hang on execution forever.
> > I haven't fixed the bugs. I'm just trying to make it so that the pipline doesn't hang for 6 hours.
> > Have you found all the places where tests can hang? I'm not sure.
> 
> Thanks for looking for these hangs. They are indeed very annoying.

Checked. 2 times raft tests ended with an error. And not once did they hang.

I consider fix ready

C0rWin:
> > > > > I think I've found the cause of raft itegration tests freezing
> > > > 
> > > > 
> > > > Can you please elaborate a bit on why you think this is the case?
> > > 
> > > 
> > > All variants of using orderer.Err() happen via the Eventually method and only here was working via Expect (without time limit). I don't know if there could be a bug here, but I decided to redo it uniformly.
> > > I also found that channelparticipation.Remove() can hang on execution forever.
> > > I haven't fixed the bugs. I'm just trying to make it so that the pipline doesn't hang for 6 hours.
> > > Have you found all the places where tests can hang? I'm not sure.
> > 
> > 
> > Thanks for looking for these hangs. They are indeed very annoying.
> 
> Checked. 2 times raft tests ended with an error. And not once did they hang.
> 
> I consider fix ready

Given the Raft integration tests are failed, I think you still need to take a look into it

pfi79:
> Given the Raft integration tests are failed, I think you still need to take a look into it

@C0rWin, I don't know how raft works and what errors there might be. I was annoyed by raft tests constantly freezing for 6 hours or 24 hours. I set myself a goal to find why they hang and solve it. I solved it. 
Out of the last 8 runs of raft tests, 3 failed and not a single raft test hangs.

See raft tests:

- https://github.com/hyperledger/fabric/actions/runs/6710811336 - SUCCESS
- https://github.com/hyperledger/fabric/actions/runs/6711015934 - SUCCESS
- https://github.com/hyperledger/fabric/actions/runs/6711503502 - SUCCESS
- https://github.com/hyperledger/fabric/actions/runs/6711643493 - FAIL
- https://github.com/hyperledger/fabric/actions/runs/6713410869/attempts/1 - SUCCESS
- https://github.com/hyperledger/fabric/actions/runs/6713410869/attempts/2 - FAIL
- https://github.com/hyperledger/fabric/actions/runs/6713410869/attempts/3 - FAIL
- https://github.com/hyperledger/fabric/actions/runs/6713410869 - SUCCESS

I will ask you to restart the last integration test for raft. It will most likely run successfully.

denyeart:
I re-ran the raft tests and sure enough they passed. Great progress avoiding the annoying hangs. I agree this can be merged without fixing all the other raft flakes, but let's have somebody more familiar with the raft tests do a final approval. @yacovm @tock-ibm @C0rWin

C0rWin:
> I re-ran the raft tests and sure enough they passed. Great progress avoiding the annoying hangs. I agree this can be merged without fixing all the other raft flakes, but let's have somebody more familiar with the raft tests do a final approval. @yacovm @tock-ibm @C0rWin

I do not think this fixes the root cause of the test being frozen hanging for hours, while definitely enough that having explicit timeouts at least will make it fail much faster. I'd go and merge it.

C0rWin (approved):
Thanks @pfi79

PR #4499, opened 2023-10-27
Verify transactions in a block are well formed (release-2.5)

denyeart:

Backport of #4490

### DIFF
--- a/protoutil/blockutils.go
+++ b/protoutil/blockutils.go
@@ -10,6 +10,8 @@ import (
 	"bytes"
 	"crypto/sha256"
 	"encoding/asn1"
+	"encoding/base64"
+	"fmt"
 	"math/big"
 
 	"github.com/golang/protobuf/proto"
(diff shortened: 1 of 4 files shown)


denyeart:
@Mergifyio rebase

protoutil/blockutils.go
```diff
@@ -218,3 +220,39 @@ func InitBlockMetadata(block *cb.Block) {
...
+			return fmt.Errorf("transaction %d is invalid: %v", i, err)
+		}
+
+		if len(env.Payload) == 0 {
+			return fmt.Errorf("transaction %d has no payload", i)
+		}
+
+		if len(env.Signature) == 0 {
+			return fmt.Errorf("transaction %d has no signature", i)
+		}
```
denyeart, line 241: @yacovm I think system channel unit tests are failing in prior release branches because configtxgen doesn't sign the config transaction in the genesis block. Would it be safe to remove this signature check in prior releases?

yacovm: > @yacovm I think system channel unit tests are failing in prior release branches because configtxgen doesn't sign the config transaction in the genesis block. Would it be safe to remove this signature check in prior releases?

For system channel we need a slightly different solution - one that ensures that only a single transaction exists in any block. You can't have 2 transactions in a system channel block, so the problem doesn't manifest vacuously.

PR #4503, opened 2023-10-29
Verify transactions in a block are well formed

yacovm:


denyeart:
I had refactored the pvtdata integration tests in main to eliminate these failures. I'll do the same for release-2.5. I think we can ignore the failures for now.

ale-linux (approved):

denyeart (approved):

PR #4505, opened 2023-10-29
Skip checking well formed-ness for config blocks

yacovm:
If the block is a config block, no point in checking its well formed-ness as it only can contain a single transaction.

### DIFF
--- a/protoutil/blockutils.go
+++ b/protoutil/blockutils.go
@@ -305,6 +305,12 @@ func VerifyTransactionsAreWellFormed(block *cb.Block) error {
 		return fmt.Errorf("empty block")
 	}
 
+	// If we have a single transaction, and the block is a config block, then no need to check
+	// well formed-ness, because there cannot be another transaction in the original block.
+	if IsConfigBlock(block) {
+		return nil
+	}
+
 	for i, rawTx := range block.Data.Data {
 		env := &cb.Envelope{}
 		if err := proto.Unmarshal(rawTx, env); err != nil {
--- a/protoutil/blockutils_test.go
+++ b/protoutil/blockutils_test.go
@@ -518,6 +518,23 @@ func TestVerifyTransactionsAreWellFormed(t *testing.T) {
 		block         *cb.Block
 	}{
 		{
+			name: "config block",
+			block: &cb.Block{Data: &cb.BlockData{
+				Data: [][]byte{
+					protoutil.MarshalOrPanic(
+						&cb.Envelope{
+							Payload: protoutil.MarshalOrPanic(&cb.Payload{
+								Header: &cb.Header{
+									ChannelHeader: protoutil.MarshalOrPanic(&cb.ChannelHeader{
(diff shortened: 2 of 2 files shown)


protoutil/blockutils.go
```diff
@@ -305,6 +305,12 @@ func VerifyTransactionsAreWellFormed(block *cb.Block) error {
 		return fmt.Errorf("empty block")
 	}
 
+	// If we have a single transaction, and the block is a config block, then no need to check
+	// well formed-ness, because there cannot be another transaction in the original block.
+	if len(block.Data.Data) == 1 && IsConfigBlock(block) {
```
ale-linux: you can actually drop the len check as `IsConfigBlock` already does it.

yacovm: Yes but it's a cheap check and I don't want to rely on the implementation of `IsConfigBlock`

ale-linux: But it's pointless to only check for a single tx anyway because it's not a unique trait of config blocks. And if we can't rely on the implementation of `IsConfigBlock` then we shouldn't use it.

yacovm: I disagree. IsConfigBlock may check if the block contains a config transaction as its first transaction. 
It does not have to check the number of transactions per block. In the past, [it did not](https://github.com/hyperledger/fabric/commit/9677344011f5910a044f9493fe1ddb6ee16efa21)! 
We check the number of transactions is 1 in this function, as an optimization, to avoid checking the first transaction in the block, which is inefficient.

ale-linux: Then it really sounds like we shouldn't use `IsConfigBlock` which doesn't seem to have been designed for our purposes. We can create our own version of that function where it's needed and have it behave the way we want to.

yacovm: IsConfigBlock does exactly what we need it to do. It checks if the block contains a config transaction.

ale-linux: either it does what we need it to do and we're comfortable using it, at which point we can remove the superfluous check, or we can't trust the stability of its implementation, at which point we should copy it and call it. Either way the duplicate len check should go imho

yacovm: I want the implementation of this function to be the same across all branches. 
In main, `IsConfigBlock` checks the size of the block. 
In earlier releases it does not. 

That's why I have this length check, to have the function be uniform across all branches.

ale-linux: > I want the implementation of this function to be the same across all branches.

iiuc that's exactly what I'm saying: create a new local version of that function doing what we want it to do as opposed to duplicating the check.

denyeart: How about putting the length check in `IsConfigBlock` across all releases. Then it will be consistent, easy to understand, and not duplicated in the calling function.

ale-linux: that also works!

ale-linux (approved):
LGTM, thanks Yacov

denyeart (approved):

Issue #4509, opened 2023-10-31
Configuration parameters CA and address in peer.deliveryclient.addressOverrides area ignored when connecting to the orderers


marianothiago:
### Description

Fabric reads the values stored in core.yaml in the peer.deliveryclient.addressOverrides section, but when connecting to orderers through the connect method in the blocksprovider.go file (internal/pkg/peer/blocksprovider/blocksprovider.go) it is using the RandomEndpoint function (endpoint, err := d.Orderers.RandomEndpoint()) located in the connection.go file (internal/pkg/peer/orderers/connection.go) to set up the endpoint that will make the connection. However the RandomEndpoint function does not use the peer.deliveryclient.addressOverrides information read by Fabric.

### Steps to reproduce

Create a fabric network with two organizations (each organization with a peer) and install the chaincode on the peers. Stop the peer from one of the organizations and change the Orderers TLS CA and the orderers TLS certificates. Start the organization's peer that was stopped. The peer willn't to connect to the orderers due to a TLS failure. Try using peer.deliveryclient.addressOverrides from core.yaml. Does not work.

PULL Rquest: https://github.com/hyperledger/fabric/pull/4510

tock-ibm:
Can you provide evidence that what doesn't work is indeed Fabric's fault? (e.g. logs, or a branch with a test?)

The override endpoints are first loaded from the core yaml when the peer starts. Then when ever
```
Update(globalAddrs []string, orgs map[string]OrdererOrg)
```
 is called, after a config TX, the source slice is rebuilt (this is the way the current config is applied as well after the peer starts).  The sources slice is rebuilt by comparing the overrides to the addresses that come from the config block - `globalAddrs`  or per org `orgs` .  If an orderer override has the same address as one of the endpoints in the   `globalAddrs`  or `orgs`, it will be used instead of that specific address. This is the place to look for a bug, if there is one, in my opinion. 

Moreover, using just one peer in an org is a bad idea, exactly because of situations like these. With more peers in an org you can restart the peer that has fallen behind with a snapshot from an up to date peer.

marianothiago:
[...]

PR #4510, opened 2023-10-31
bug fix 4509. Configuration parameters CA and address in peer.deliver…

marianothiago:
#### Type of change

- Bug fix

#### Description

Fabric reads the values stored in core.yaml in the peer.deliveryclient.addressOverrides section, but when connecting to orderers through the connect method in the blocksprovider.go file (internal/pkg/peer/blocksprovider/blocksprovider.go) it is using the RandomEndpoint function (endpoint, err := d.Orderers.RandomEndpoint()) located in the connection.go file (internal/pkg/peer/orderers/connection.go) to set up the endpoint that will make the connection. However the RandomEndpoint function does not use the peer.deliveryclient.addressOverrides information read by Fabric.

#### Related issues

https://github.com/hyperledger/fabric/issues/4509

### DIFF
--- a/internal/pkg/peer/orderers/connection.go
+++ b/internal/pkg/peer/orderers/connection.go
@@ -50,7 +50,14 @@ func (cs *ConnectionSource) RandomEndpoint() (*Endpoint, error) {
 	if len(cs.allEndpoints) == 0 {
 		return nil, errors.Errorf("no endpoints currently defined")
 	}
-	return cs.allEndpoints[rand.Intn(len(cs.allEndpoints))], nil
+	OrderersOverridekeys := make([]string, 0, len(cs.overrides))
+	for key := range cs.overrides {
+		OrderersOverridekeys = append(OrderersOverridekeys, key)
+	}
+	random := rand.Intn(len(OrderersOverridekeys))
+	endpoint := cs.overrides[OrderersOverridekeys[random]]
+	endpoint.Address = cs.allEndpoints[random].Address
+	return endpoint, nil
 }
 
 func (cs *ConnectionSource) Update(globalAddrs []string, orgs map[string]OrdererOrg) {


C0rWin:
Thanks for providing the fix. Would you also mind adding some unit tests to demonstrate the issue described and to make sure the fix indeed solves the problem? Moreover, to ensure we got it covered for potential future modifications?

denyeart:
Does the problem still occur in main branch? Typically fixes are made to main branch and then backported as desired to the release branches.

marianothiago:
> Thanks for providing the fix. Would you also mind adding some unit tests to demonstrate the issue described and to make sure the fix indeed solves the problem? Moreover, to ensure we got it covered for potential future modifications?

OK. I will do this.

marianothiago:
> Does the problem still occur in main branch? Typically fixes are made to main branch and then backported as desired to the release branches.

I tested branches 2.1, 2.2, 2.3, 2.4 and 2.5. I haven't tested it in main. I'll test it and let you know.

tock-ibm:
@marianothiago @C0rWin @denyeart 

I don't think this is the right approach. 
The override endpoints are first loaded from the core yaml when the peer starts. Then when ever
```
Update(globalAddrs []string, orgs map[string]OrdererOrg)
```
 is called, after a config TX, the source slice is rebuilt (this is the way the current config is applied as well after the peer starts).  The sources slice is rebuilt by comparing the overrides to the addresses that come from the config block - `globalAddrs`  or per org `orgs` .  If an orderer override has the same address as one of the endpoints in the   `globalAddrs`  or `orgs`, it will be used instead of that specific address. This is the place to look for a bug, if there is one, in my opinion. According to the current code, overrides are a mechanism to override the definition of an endpoint, not to add more endpoints, which is what you do here.

With respect to your use case: you describe a case where a peer falls behind the ordering cluster config, and cannot connect to it. While the overrides may give you a partial solution, I think the more "updated" way to approach this is by starting the peer from a snapshot taken from another peer that is up to date.

And finally, this object is about to change in V3, as a result of the work we do to implement a byzantine block puller.

tock-ibm:
> @marianothiago @C0rWin @denyeart
> 
> I don't think this is the right approach. The override endpoints are first loaded from the core yaml when the peer starts. Then when ever
> 
> ```
> Update(globalAddrs []string, orgs map[string]OrdererOrg)
> ```
> 
> is called, after a config TX, the source slice is rebuilt (this is the way the current config is applied as well after the peer starts). The sources slice is rebuilt by comparing the overrides to the addresses that come from the config block - `globalAddrs` or per org `orgs` . If an orderer override has the same address as one of the endpoints in the `globalAddrs` or `orgs`, it will be used instead of that specific address. This is the place to look for a bug, if there is one, in my opinion. According to the current code, overrides are a mechanism to override the definition of an endpoint, not to add more endpoints, which is what you do here.
> 
> With respect to your use case: you describe a case where a peer falls behind the ordering cluster config, and cannot connect to it. While the overrides may give you a partial solution, I think the more "updated" way to approach this is by starting the peer from a snapshot taken from another peer that is up to date.
> 
> And finally, this object is about to change in V3, as a result of the work we do to implement a byzantine block puller.

Actually, looking at the code, I see overrides are not applied when there are per org endpoints, only when there are global endpoints. This can explain the behavior you are describing @marianothiago 

@denyeart aren't global endpoints deprecated?

C0rWin:
@denyeart I'm following up on this PR after some time. I'm wondering if you've had a chance to review this change again or if there are still concerns with the approach?

If there are still issues with this implementation, or if you feel this isn't the right direction given potential V3 changes mentioned by @tock-ibm, would it be better to close this PR?

I'd appreciate your guidance on whether we should continue with this PR or consider closing it at this point.

denyeart:
@C0rWin 

Users have successfully used addressOverrides for the purpose of bringing up a new peer that connects to different orderers than those defined in block 0. It also looks like there is an integration test demonstrating that it works at https://github.com/hyperledger/fabric/blob/main/integration/e2e/e2e_test.go#L518. If there is a bug in this space we'll need to understand better why it works in some scenarios and not in others. 

Based on @tock-ibm comments, if there is a bug, the fix from this PR is not in the correct place.

My suggestion:

- keep issue #4509 open (I've added the same text to the issue), in case anybody can provide more information.
- close this PR since there has been no activity for over 1 year.

PR #4523, opened 2023-11-02
Make clear of osnadmin remove usage

yeasy:
Change-Id: Id90cea1c9152f2f6c2d0225488eb6e08a5808a68

#### Type of change

#### Description

The patchset explains more details about `osnadmin remove`.

#### Additional details

N/A

#### Related issues

N/A

#### Release Note

N/A

### DIFF
--- a/cmd/osnadmin/main.go
+++ b/cmd/osnadmin/main.go
@@ -54,7 +54,7 @@ func executeForArgs(args []string) (output string, exit int, err error) {
 	list := channel.Command("list", "List channel information for an Ordering Service Node (OSN). If the channelID flag is set, more detailed information will be provided for that channel.")
 	listChannelID := list.Flag("channelID", "Channel ID").Short('c').String()
 
-	remove := channel.Command("remove", "Remove an Ordering Service Node (OSN) from a channel.")
+	remove := channel.Command("remove", "Remove a channel from an Ordering Service Node (OSN).")
 	removeChannelID := remove.Flag("channelID", "Channel ID").Short('c').Required().String()
 
 	command, err := app.Parse(args)
--- a/docs/source/bft_configuration.md
+++ b/docs/source/bft_configuration.md
@@ -193,13 +193,13 @@ Extending the cluster to five nodes means until the fifth node finishes replicat
 To add a new node to the ordering service:
 
 1. **Ensure the orderer organization that owns the new node is one of the orderer organizations on the channel**. If the orderer organization is not an administrator, the node will be unable to pull blocks as a follower or be joined to the consenter set.
-2. **Start the new ordering node**. For information about how to deploy an ordering node, check out [Planning for an ordering service](./deployorderer/ordererdeploy.html). Note that when you use the `osnadmin` CLI to create and join a channel, you do not need to point to a configuration block when starting the node.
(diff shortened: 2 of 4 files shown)


cmd/osnadmin/main.go
```diff
@@ -54,7 +54,7 @@ func executeForArgs(args []string) (output string, exit int, err error) {
 	list := channel.Command("list", "List channel information for an Ordering Service Node (OSN). If the channelID flag is set, more detailed information will be provided for that channel.")
 	listChannelID := list.Flag("channelID", "Channel ID").Short('c').String()
 
-	remove := channel.Command("remove", "Remove an Ordering Service Node (OSN) from a channel.")
+	remove := channel.Command("remove", "Remove an Ordering Service Node (OSN)'s local joining block and ledger of a channel.")
```
denyeart: This isn't actually more clear to me... what specifically are you trying to solve? I believe the `remove` command will both remove the local channel ledger on the OSN and will stop the OSN from communicating in that channel, which can be summarized as "removing the OSN from the channel".

The doc [here](https://hyperledger-fabric.readthedocs.io/en/latest/create_channel/create_channel_participation.html#add-or-remove-orderers-from-existing-channels) sounds like the correct full description:

> Add or remove orderers from existing channels
> 
> You can continue to use the osnadmin channel join and osnadmin channel remove commands to add and remove orderers on each channel according to your business needs. Be aware that before you remove a channel from an orderer, it is recommended that you first remove the orderer from the channel’s consenter set by submitting a channel update request.

The doc [here](https://hyperledger-fabric.readthedocs.io/en/latest/raft_configuration.html) sounds incorrect:

> To remove an ordering node from the consenter set of a channel, use the osnadmin channel remove command to remove its endpoint and certificates from the channel.

@tock-ibm What do you think? Does that latest doc paragraph need to be fixed? The osnadmin command won't actually update the channel config as is implied here, it just removes the channel on the OSN.

yeasy: "Remove an OSN from channel" sounds to remove the OSN from the channel configuration.
Here the `osnadmin remove` is to remove the OSN's local ledger corresponding to a channel.

tock-ibm: @denyeart @yeasy 

Indeed the first doc is correct. Removing a channel from an OSN not only deletes the ledger, but stops the OSN from communicating with other OSNs that are members of the channel. 

The second doc is indeed incorrect.
It should read  "...use a config transaction to remove its endpoint and certificates from the channel."

There is also another typo there:
```
Note that **when** you use the osnadmin CLI to create and join a channel, you do not need to point to a configuration block when starting the node.
```
should be:
```
Note that **while** you use the osnadmin CLI to create and join a channel, you do not need to point to a configuration block when starting the node.
```

As to the proposed language change, I propose

"Remove a channel from an Ordering Service Node (OSN)."

which is what actually happens.

yeasy: Yes, that's exactly what i meant. The `osnadmin channel remove` cmd only removes the ledger data from an OSN locally, but the OSN is still a member of the channel, aka, it does not change the channel config.

And I can use the `"Remove a channel from an Ordering Service Node (OSN)."` as suggested.

Will update the pr. Thanks!

denyeart (approved):
Excellent, thank you!

PR #4550, opened 2023-11-20
BFT Block Puller: verifier in the monitor

tock-ibm:
#### Type of change
- New feature

#### Description

Integration of the stand alone verifier with the censorship monitor.

#### Related issues
#4353


common/deliverclient/block_verification.go
```diff
@@ -0,0 +1,350 @@
...
+		return errors.New("missing channel header")
+	}
+
+	chdr, err := protoutil.UnmarshalChannelHeader(payload.Header.ChannelHeader)
+	if err != nil {
+		return errors.WithMessage(err, "error unmarshalling channel header")
+	}
+
+	if chdr.GetChannelId() != a.channelID {
+		return errors.New("config block channel ID '%s' does not match expected: '%s'")
```
C0rWin:
```suggestion
		return errors.New("config block channel ID '%s' does not match expected: '%s'", chdr.GetChannelId(), a.ChannelID)
```

common/deliverclient/block_verification.go
```diff
@@ -0,0 +1,350 @@
...
+	"github.com/hyperledger/fabric/protoutil"
+	"github.com/pkg/errors"
+)
+
+type CloneableUpdatableBlockVerifier interface {
+	// VerifyBlock checks block integrity and its relation to the chain, and verifies the signatures.
+	VerifyBlock(block *common.Block) error
+
+	// VerifyBlockAttestation does the same as VerifyBlock, except it assumes block.Data = nil. It therefore does not
+	// compute the block.Data.Hash() and compare it to the block.Header.DataHash. This is used when the orderer
```
yacovm: s/compare/compares/

common/deliverclient/block_verification.go
```diff
@@ -0,0 +1,350 @@
...
+	// delivers a block with header & metadata only, as an attestation of block existence.
+	VerifyBlockAttestation(block *common.Block) error
```
yacovm, line 28: If it assumes the data is nil, why not pass only the header and metadata as parameters?

common/deliverclient/block_verification.go
```diff
@@ -0,0 +1,350 @@
...
+		return nil, errors.Errorf("config block is nil")
+	}
+	if configBlock.Header == nil {
+		return nil, errors.Errorf("config block header is nil")
+	}
+	configIndex, err := protoutil.GetLastConfigIndexFromBlock(configBlock)
+	if err != nil {
+		return nil, errors.WithMessage(err, "error getting config index from config block")
+	}
+	if configIndex != configBlock.Header.Number {
```
yacovm, line 86: why are we checking this specific aspect of the initialization parameter? 

I'm not saying we should necessarily remove this, but what is the benefit here? 
you did not check it's a config block, but only this aspect. 

what's so special about it?

common/deliverclient/block_verification.go
```diff
@@ -0,0 +1,350 @@
...
+		return nil, errors.Errorf("config block number [%d] is different than its own config index [%d]", configBlock.Header.Number, configIndex)
+	}
+
+	if lastBlock != nil {
```
yacovm: how can last block be nil? We always have the genesis block don't we?

common/deliverclient/block_verification.go
```diff
@@ -0,0 +1,350 @@
...
+// used as an argument to the UpdateConfig method.
+// The block stream could be composed of either:
+// - full blocks, which are verified using the VerifyBlock method, or
+// - block attestations (a header+metadata, with nil data) which are verified using the VerifyBlockAttestation method.
+// In both cases, config blocks must arrive in full.
+type BlockVerificationAssistant struct {
+	channelID string
+
+	// Creates the sigVerifierFunc whenever the configuration is set or updated.
+	verifierAssembler *BlockVerifierAssembler
```
yacovm, line 55: this is a concrete implementation, why not use an interface? then you don't need to pass the BCCSP anymore, any you just need to initialize the interface implementation with BCCSP.

common/deliverclient/block_verification.go
```diff
@@ -0,0 +1,350 @@
...
+	}
+
+	// Verify that Header.DataHash is equal to the hash of block.Data
+	// This is to ensure that the header is consistent with the data carried by this block
+	if !bytes.Equal(dataHash, block.Header.DataHash) {
+		return errors.Errorf("Header.DataHash is different from Hash(block.Data) for block with id [%d] on channel [%s]", block.Header.Number, a.channelID)
+	}
+
+	err = a.sigVerifierFunc(block.Header, block.Metadata)
+	if err == nil {
```
yacovm: please return early if err is not nil instead

internal/pkg/peer/blocksprovider/block_receiver.go
```diff
@@ -134,12 +136,23 @@ func (br *BlockReceiver) processMsg(msg *orderer.DeliverResponse) (uint64, error
...
 	case *orderer.DeliverResponse_Block:
 		blockNum := t.Block.Header.Number
-		if err := br.blockVerifier.VerifyBlock(gossipcommon.ChannelID(br.channelID), blockNum, t.Block); err != nil {
-			return 0, errors.WithMessage(err, "block from orderer could not be verified")
+
+		if err := br.updatableBlockVerifier.VerifyBlock(t.Block); err != nil {
+			return 0, errors.WithMessagef(err, "block [%d] from orderer [%s] could not be verified", blockNum, br.endpoint.String())
 		}
+
 		err := br.blockHandler.HandleBlock(br.channelID, t.Block)
```
yacovm, line 143: Is here in HandleBlock where a buffer would be built? Otherwise we're pulling them and committing them one by one

internal/pkg/peer/blocksprovider/bft_deliverer.go
```diff
@@ -46,17 +47,7 @@ type DurationExceededHandler interface {
 
 //go:generate counterfeiter -o fake/updatable_block_verifier.go --fake-name UpdatableBlockVerifier . UpdatableBlockVerifier
 type UpdatableBlockVerifier interface {
```
yacovm: why not `type UpdatableBlockVerifier deliverclient.CloneableUpdatableBlockVerifier` ?

common/deliverclient/block_verification.go
tock-ibm: done

common/deliverclient/block_verification.go
tock-ibm: just easier that way

common/deliverclient/block_verification.go
tock-ibm: correct, that can never happen

internal/pkg/peer/blocksprovider/block_receiver.go
tock-ibm: In the orderer, yes, HandleBlock will append it to a buffer. in the peer it hands it over to gossip which has buffering already.

common/deliverclient/block_verification.go
tock-ibm: correct, I added a check
```
	if !protoutil.IsConfigBlock(configBlock) {
		return nil, errors.New("config block parameter does not carry a config block")
	}
```

the reason I am checking the index is because I am verifying that the last block refers to the same config block we are starting with.

common/deliverclient/block_verification.go
tock-ibm: True, we can do that, but that does not buy us much, as building the assembler is local to the constructor, which gets the BCCSP anyway... it does not help testing either

internal/pkg/peer/blocksprovider/bft_deliverer.go
tock-ibm: that works too

common/deliverclient/block_verification.go
tock-ibm: done

yacovm (approved):

PR #4555, opened 2023-11-27
Remove chaincode deploy legacy from pvt data tests

C0rWin:
This commit removes usages of the chaincode legacy deploy code from private data integration tests.

### DIFF
--- a/integration/pvtdata/pvtdata_test.go
+++ b/integration/pvtdata/pvtdata_test.go
@@ -80,20 +80,23 @@ var _ = Describe("PrivateData", func() {
 		})
 
 		It("disseminates private data per collections_config1 (positive test) and collections_config8 (negative test)", func() {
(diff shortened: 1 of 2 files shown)


integration/pvtdata/pvtdata_test.go
```diff
@@ -498,14 +487,6 @@ var _ = Describe("PrivateData", func() {
 			}
 
 			When("chaincode is migrated from legacy to new lifecycle with same collection config", func() {
```
denyeart: Since we are keeping assertReconcileBehavior() in here but not the upgrade from legacy chaincode, change the description text in the When().

denyeart (commented):
Thanks Artem! Just one nit here.

integration/pvtdata/pvtdata_test.go
C0rWin: fixed

denyeart (approved):

PR #4559, opened 2023-11-29
Removes methods of legacy chaincode lifecycle from nwo

C0rWin:
This commit concludes removal of the legace lifecycle methods from integration tests nwo package.


denyeart (approved):
Thanks Artem!

Issue #3773, opened 2022-11-09
Migration from RAFT to BFT


andrew-coleman:
- [x] Allow migration from Raft to BFT - migration filter, chain, etc
- [x] Integration tests
- [ ] #4642
- [ ] #4643
- [x] Documentation

bh4rtp:
Which release will support BFT?

Param-S:
The current plan is to include BFT consensus in 3.0 release.

aruns05:
Hello @andrew-coleman  @Param-S 
Is this being worked on ?

PR #4561, opened 2023-11-30
Migration from Raft to BFT test

MayRosenbaum:
#### Type of change

- Test Improvement 
- Test update

#### Related Issue

#3773

### DIFF
--- a/integration/raft/migration_test.go
+++ b/integration/raft/migration_test.go
@@ -23,7 +23,6 @@ import (
 	"github.com/golang/protobuf/proto"
 	"github.com/hyperledger/fabric-protos-go/common"
 	protosorderer "github.com/hyperledger/fabric-protos-go/orderer"
-	protosraft "github.com/hyperledger/fabric-protos-go/orderer/etcdraft"
 	"github.com/hyperledger/fabric/integration/channelparticipation"
 	"github.com/hyperledger/fabric/integration/nwo"
 	"github.com/hyperledger/fabric/integration/nwo/commands"
@@ -75,6 +74,10 @@ var _ = Describe("ConsensusTypeMigration", func() {
 		_ = os.RemoveAll(testDir)
 	})
 
+	// This test executes the migration on an etcdraft based system.
+	// The migration includes change to maintenance mode, config updates needed for migration to a bft based system and
+	// exiting maintenance mode back to the normal state.
+	// This test restarts the orderers and ensures they operate under the new configuration.
 	Describe("Raft to BFT migration", func() {
 		It("migrates from Raft to BFT", func() {
 			networkConfig := nwo.MultiNodeEtcdRaft()
@@ -126,39 +129,25 @@ var _ = Describe("ConsensusTypeMigration", func() {
 			By("Change to maintenance mode")
 
 			peer := network.Peer("Org1", "peer0")
-			orderer := network.Orderer("orderer1")
-
-			addBFTInConfigTree(network, peer, orderer, "testchannel")
 
 			By("Config update on standard channel, State=MAINTENANCE, enter maintenance-mode")
-			config, updatedConfig := prepareTransition(network, peer, orderer, "testchannel",
+			config, updatedConfig := prepareTransition(network, peer, o1, "testchannel",
 				"etcdraft", protosorderer.ConsensusType_STATE_NORMAL,
 				"etcdraft", nil, protosorderer.ConsensusType_STATE_MAINTENANCE)
-			nwo.UpdateOrdererConfig(network, orderer, "testchannel", config, updatedConfig, peer, orderer)
+			nwo.UpdateOrdererConfig(network, o1, "testchannel", config, updatedConfig, peer, o1)
 
-			bftMetadata := protoutil.MarshalOrPanic(&smartbft.Options{
-				RequestBatchMaxCount:      types.DefaultConfig.RequestBatchMaxCount,
-				RequestBatchMaxBytes:      types.DefaultConfig.RequestBatchMaxBytes,
-				RequestBatchMaxInterval:   types.DefaultConfig.RequestBatchMaxInterval.String(),
-				IncomingMessageBufferSize: types.DefaultConfig.IncomingMessageBufferSize,
-				RequestPoolSize:           types.DefaultConfig.RequestPoolSize,
-				RequestForwardTimeout:     types.DefaultConfig.RequestForwardTimeout.String(),
-				RequestComplainTimeout:    types.DefaultConfig.RequestComplainTimeout.String(),
-				RequestAutoRemoveTimeout:  types.DefaultConfig.RequestAutoRemoveTimeout.String(),
-				ViewChangeResendInterval:  types.DefaultConfig.ViewChangeResendInterval.String(),
-				ViewChangeTimeout:         types.DefaultConfig.ViewChangeTimeout.String(),
-				LeaderHeartbeatTimeout:    types.DefaultConfig.LeaderHeartbeatTimeout.String(),
-				LeaderHeartbeatCount:      types.DefaultConfig.LeaderHeartbeatCount,
-				CollectTimeout:            types.DefaultConfig.CollectTimeout.String(),
-				SyncOnStart:               types.DefaultConfig.SyncOnStart,
-				SpeedUpViewChange:         types.DefaultConfig.SpeedUpViewChange,
-			})
-			config, updatedConfig = prepareTransition(network, peer, orderer, "testchannel",
+			By("Config updates: migration from Raft to BFT")
+
+			bftMetadata := protoutil.MarshalOrPanic(prepareBftMetadata())
+
+			config, updatedConfig = prepareTransition(network, peer, o1, "testchannel",
 				"etcdraft", protosorderer.ConsensusType_STATE_MAINTENANCE,
 				"BFT", bftMetadata, protosorderer.ConsensusType_STATE_MAINTENANCE)
-			nwo.UpdateOrdererConfig(network, orderer, "testchannel", config, updatedConfig, peer, orderer)
+			currentBlockNumber := nwo.CurrentConfigBlockNumber(network, peer, o1, "testchannel")
+			nwo.UpdateOrdererConfig(network, o1, "testchannel", config, updatedConfig, peer, o1)
 
-			time.Sleep(time.Second) // TODO: check block was committed in all orderers
+			// check block was committed in all orderers
+			assertBlockReceptionInAllOrderers(network.Orderers[1:], peer, network, "testchannel", currentBlockNumber)
 
 			for _, oProc := range []ifrit.Process{o1Proc, o2Proc, o3Proc, o4Proc} {
 				if oProc != nil {
@@ -169,18 +158,17 @@ var _ = Describe("ConsensusTypeMigration", func() {
 
 			runOrderers()
 
-			// TODO:
 			By("Waiting for followers to see the leader, again")
 			Eventually(o2Runner.Err(), network.EventuallyTimeout, time.Second).Should(gbytes.Say("Message from 1 channel=testchannel"))
 			Eventually(o3Runner.Err(), network.EventuallyTimeout, time.Second).Should(gbytes.Say("Message from 1 channel=testchannel"))
 			Eventually(o4Runner.Err(), network.EventuallyTimeout, time.Second).Should(gbytes.Say("Message from 1 channel=testchannel"))
 
 			// Exit maintenance mode
-
-			config, updatedConfig = prepareTransition(network, peer, orderer, "testchannel",
+			config, updatedConfig = prepareTransition(network, peer, o1, "testchannel",
 				"BFT", protosorderer.ConsensusType_STATE_MAINTENANCE,
 				"BFT", bftMetadata, protosorderer.ConsensusType_STATE_NORMAL)
-			nwo.UpdateOrdererConfig(network, orderer, "testchannel", config, updatedConfig, peer, orderer)
+			currentBlockNumber = nwo.CurrentConfigBlockNumber(network, peer, o1, "testchannel")
+			nwo.UpdateOrdererConfig(network, o1, "testchannel", config, updatedConfig, peer, o1)
 
 			// Now, run a transaction to ensure BFT works.
 			env = CreateBroadcastEnvelope(network, o1, "testchannel", []byte("foo"))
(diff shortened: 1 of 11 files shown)


orderer/consensus/etcdraft/chain.go
```diff
@@ -1115,12 +1115,17 @@ func (c *Chain) commitBlock(block *common.Block) {
 func (c *Chain) detectConfChange(block *common.Block) *MembershipChanges {
 	// If config is targeting THIS channel, inspect consenter set and
 	// propose raft ConfChange if it adds/removes node.
-	configMetadata := c.newConfigMetadata(block)
+	configMetadata, consensusType := c.newConfigMetadata(block)
```
tock-ibm, line 1121: It seems that this will never return configMetadata which is not nil and consensusType.Type == "BFT"; I don't see how this works...

orderer/consensus/etcdraft/util.go
```diff
@@ -84,22 +84,22 @@ func MetadataHasDuplication(md *etcdraft.ConfigMetadata) error {
...
 	consensusTypeValue := &orderer.ConsensusType{}
 	if err := proto.Unmarshal(configValue.Value, consensusTypeValue); err != nil {
-		return nil, errors.Wrap(err, "failed to unmarshal consensusType config update")
+		return nil, nil, errors.Wrap(err, "failed to unmarshal consensusType config update")
 	}
 
 	updatedMetadata := &etcdraft.ConfigMetadata{}
 	if err := proto.Unmarshal(consensusTypeValue.Metadata, updatedMetadata); err != nil {
-		return nil, errors.Wrap(err, "failed to unmarshal updated (new) etcdraft metadata configuration")
+		return nil, nil, errors.Wrap(err, "failed to unmarshal updated (new) etcdraft metadata configuration")
```
tock-ibm, line 100: if the new metadata contains smartbft options, as it should when migration happens, an error will occur

orderer/consensus/etcdraft/chain_test.go
tock-ibm, line 1: add a test case that emulates what happens when a raft chain receives the config block that changes from raft to BFT

orderer/consensus/etcdraft/util.go
```diff
@@ -84,22 +86,32 @@ func MetadataHasDuplication(md *etcdraft.ConfigMetadata) error {
...
-		return nil, errors.Wrap(err, "failed to unmarshal consensusType config update")
+		return nil, nil, errors.Wrap(err, "failed to unmarshal consensusType config update")
+	}
+
+	if consensusTypeValue.Type != "etcdraft" {
+		if consensusTypeValue.Type == "BFT" {
+			updatedBftMetadata := &smartbft.Options{}
+			if err := proto.Unmarshal(consensusTypeValue.Metadata, updatedBftMetadata); err != nil {
+				return nil, nil, errors.Wrap(err, "failed to unmarshal updated (new) bft metadata configuration")
+			}
```
tock-ibm: actually, I think lets not import smartbft package here... don't unmarshal to &smartbft.Options{}

orderer/consensus/etcdraft/util.go
```diff
@@ -84,22 +86,32 @@ func MetadataHasDuplication(md *etcdraft.ConfigMetadata) error {
 }
 
 // MetadataFromConfigValue reads and translates configuration updates from config value into raft metadata
-func MetadataFromConfigValue(configValue *common.ConfigValue) (*etcdraft.ConfigMetadata, error) {
+func MetadataFromConfigValue(configValue *common.ConfigValue) (*etcdraft.ConfigMetadata, *orderer.ConsensusType, error) {
```
tock-ibm, line 88: add a unit test

orderer/consensus/etcdraft/chain.go
```diff
@@ -1446,6 +1461,17 @@ func (c *Chain) ValidateConsensusMetadata(oldOrdererConfig, newOrdererConfig cha
 		return nil
 	}
 
+	if newOrdererConfig.ConsensusType() != "etcdraft" {
+		if newOrdererConfig.ConsensusType() == "BFT" {
+			// This is a migration, so we don't know how to validate this config change.
+			return nil
+		} else {
+			c.logger.Panicf("illegal consensus type detected during consensus metadata validation: %s", newOrdererConfig.ConsensusType())
+			panic("illegal consensus type detected during consensus metadata validation")
```
tock-ibm: why panic? return error

orderer/common/msgprocessor/maintenancefilter.go
```diff
@@ -46,9 +46,7 @@ func NewMaintenanceFilter(support MaintenanceFilterSupport, bccsp bccsp.BCCSP) *
 		bccsp:                         bccsp,
 	}
 	mf.permittedTargetConsensusTypes["etcdraft"] = true
-	// Until we have a BFT consensus type, we use this for integration testing of consensus-type migration.
-	// Caution: proposing a config block with this type will cause panic.
-	mf.permittedTargetConsensusTypes["testing-only"] = true
+	mf.permittedTargetConsensusTypes["BFT"] = true
```
tock-ibm, line 51: This is the permitted target type. So you have to remove "etcdraft", because we only allow etcdraft -> BFT, but the other way around.

tock-ibm: In the `inspect` method, line 136, you to check that if the target is BFT, you can unmarshall the metadata to &smartbft.Options{}. Make sure all the field are recognized ;-) in unit test

orderer/common/msgprocessor/maintenancefilter.go
```diff
@@ -133,10 +135,15 @@ func (mf *MaintenanceFilter) inspect(configEnvelope *cb.ConfigEnvelope, ordererC
...
-			updatedMetadata := &protoetcdraft.ConfigMetadata{}
+		if nextOrdererConfig.ConsensusType() == "BFT" {
+			updatedMetadata := &smartbft.Options{}
 			if err := proto.Unmarshal(nextOrdererConfig.ConsensusMetadata(), updatedMetadata); err != nil {
-				return errors.Wrap(err, "failed to unmarshal etcdraft metadata configuration")
+				return errors.Wrap(err, "failed to unmarshal BFT metadata configuration")
+			}
+
+			_, err := validateBFTMetadataOptions(1, updatedMetadata)
+			if updatedMetadata.XXX_unrecognized != nil || err != nil {
```
yacovm: what is this XXX field thing check? This is prorobuf implementation specific, isn't it? 

Once we upgrade the protobuf version, we might not have this field anymore.

orderer/consensus/etcdraft/chain.go
```diff
@@ -1115,14 +1115,28 @@ func (c *Chain) commitBlock(block *common.Block) {
...
 		return nil
 	}
 
 	if consensusType.Type != "etcdraft" {
-		c.logger.Infof("Detected migration to %s", consensusType.Type)
+		if consensusType.Type == "BFT" {
+			c.logger.Infof("Detected migration to %s", consensusType.Type)
+			return nil
+		} else {
+			c.logger.Panicf("illegal consensus type detected: %s", consensusType.Type)
```
yacovm, line 1134: this is post consensus. Do we have a validation check for this condition pre-consensus anywhere?

yacovm: otherwise, you know what will happen once we reach this point ;-)

orderer/common/msgprocessor/maintenancefilter.go
MayRosenbaum: That's right, we can skip this check, err check is enough.

orderer/consensus/etcdraft/chain.go
MayRosenbaum: We have one more check in the maintenance filter.

orderer/common/msgprocessor/maintenancefilter.go
tock-ibm: True, the XXX fields are not required to preserve backwards compatibility going forward, so we better not use them.

orderer/common/msgprocessor/maintenancefilter.go
```diff
@@ -199,3 +204,59 @@ func (mf *MaintenanceFilter) ensureConsensusTypeChangeOnly(configEnvelope *cb.Co
 
 	return nil
 }
+
+func validateBFTMetadataOptions(selfID uint64, options *smartbft.Options) (types.Configuration, error) {
```
tock-ibm: Why copy the method here? you can reuse the origin just by making it public.

orderer/common/msgprocessor/maintenancefilter.go
tock-ibm: verifying with `validateBFTMetadataOptions` should be enough, but see comment on verifying the consenters map as well.

orderer/common/msgprocessor/maintenancefilter.go
```diff
@@ -135,10 +135,15 @@ func (mf *MaintenanceFilter) inspect(configEnvelope *cb.ConfigEnvelope, ordererC
 				ordererConfig.ConsensusType(), nextOrdererConfig.ConsensusType())
 		}
 
-		if nextOrdererConfig.ConsensusType() == "etcdraft" {
-			updatedMetadata := &protoetcdraft.ConfigMetadata{}
+		if nextOrdererConfig.ConsensusType() == "BFT" {
```
tock-ibm, line 138: When this happens, i.e. a migration from etcdraft to BFT, validate also that there is a consenters map under the channel group, after you validate the `smartbft.Options`.

In my opinion, we should validate that the consenters map in the next config has the same consenters as raft has in the current config. That is, we should not try and add/remove nodes or change certs while at the same time while we're switching the consensus type.

integration/raft/migration_test.go
```diff
@@ -69,373 +74,349 @@ var _ = Describe("ConsensusTypeMigration", func() {
...
+			By("performing operation with orderer1")
+			env := CreateBroadcastEnvelope(network, o1, "testchannel", []byte("foo"))
+			resp, err := ordererclient.Broadcast(network, o1, env)
+			Expect(err).NotTo(HaveOccurred())
+			Expect(resp.Status).To(Equal(common.Status_SUCCESS))
+
+			block := FetchBlock(network, o1, 1, "testchannel")
+			Expect(block).NotTo(BeNil())
+
+			By("Change to maintenance mode")
```
tock-ibm: before each step of the migration, add a numbered comment like:
```
// === Step 1: Config update, change to MAINTENANCE ===
By("1) ...")
```
as is done in the tests below. this helps keep track of what is going on in the logs.
be consistent about how these look.

integration/raft/migration_test.go
```diff
@@ -69,373 +74,349 @@ var _ = Describe("ConsensusTypeMigration", func() {
...
+			resp, err = ordererclient.Broadcast(network, o1, env)
+			Expect(err).NotTo(HaveOccurred())
+			Expect(resp.Status).To(Equal(common.Status_SUCCESS))
+
+			// check block was successfully committed in all orderers
+			assertBlockReceptionInAllOrderers(network.Orderers[1:], peer, network, "testchannel", currentBlockNumber)
+		})
+	})
+
+	Describe("Enforcing config updates during Raft to BFT migration", func() {
```
tock-ibm, line 192: this suite tests the negative paths that may occur. what I am missing are test cases for:
- when the smartbft.Options are corrupt, does not pass validation, or missing
- when the consenters map is corrupt, does not pass validation, or missing

integration/raft/migration_test.go
```diff
@@ -69,373 +74,349 @@ var _ = Describe("ConsensusTypeMigration", func() {
...
+			// orderer3
+			config = nwo.GetConfig(network, peer, o3, "testchannel")
+			consensusTypeValue = extractOrdererConsensusType(config)
+			validateConsensusTypeValue(consensusTypeValue, "BFT", protosorderer.ConsensusType_STATE_MAINTENANCE)
 
-			// === Step 7: config update on standard channel2, State=MAINTENANCE, type=future-type ===
+			// orderer4
+			config = nwo.GetConfig(network, peer, o4, "testchannel")
+			consensusTypeValue = extractOrdererConsensusType(config)
+			validateConsensusTypeValue(consensusTypeValue, "BFT", protosorderer.ConsensusType_STATE_MAINTENANCE)
```
tock-ibm: you can put this in a for loop, e.g.

for _, o := range {o1, o2, o3, o4}

integration/raft/migration_test.go
```diff
@@ -69,373 +74,349 @@ var _ = Describe("ConsensusTypeMigration", func() {
...
 		})
 
 		// This test executes the "green path" migration config updates on an etcdraft based system, on a standard
 		// channel, and verifies that these config updates have the desired effect.
-		//
 		// The green path is entering maintenance mode, and then changing the consensus type.
 		// In maintenance mode we check that normal transactions are blocked.
-		//
 		// We also check that after entering maintenance mode, we can exit it without making any changes - the "abort path".
 		It("executes raft2future green path", func() {
```
tock-ibm: It("executes raft2bft green path with extra checks") 
Check the abort path (exit from maintenance w/o changes)
This checks that normal TXs are blocked when in maintenance mode
etc. 
i.e. - future -> bft

integration/raft/migration_test.go
```diff
@@ -69,373 +74,349 @@ var _ = Describe("ConsensusTypeMigration", func() {
...
 		// This test executes the migration flow and checks that forbidden transitions are rejected.
 		// These transitions are enforced by the maintenance filter:
 		// - Entry to & exit from maintenance mode can only change ConsensusType.State.
-		// - In maintenance mode one can only change ConsensusType.Type & ConsensusType.Metadata.
-		// - ConsensusType.Type can only change from "etcdraft" to "future-type" (to be replaced by future consensus
-		//   protocol), and only in maintenance mode.
+		// - In maintenance mode one can only change ConsensusType.Type & ConsensusType.Metadata & Orderers.ConsenterMapping.
+		// - ConsensusType.Type can only change from "etcdraft" to "BFT", and only in maintenance mode.
 		It("executes raft2future forbidden transitions", func() {
 			// === Step 1: ===
```
tock-ibm: raft2future -> raft2bft

integration/raft/migration_test.go
```diff
@@ -69,373 +74,349 @@ var _ = Describe("ConsensusTypeMigration", func() {
...
+				"BFT", nil, protosorderer.ConsensusType_STATE_NORMAL)
 			updateConfigWithBatchTimeout(updatedConfig)
-			updateOrdererConfigFailed(network, orderer, channel1, config, updatedConfig, peer, orderer)
-		})
-
-		// Note:
-		// Instead of booting to a future-type which does not exist yet, we change some other config value in
-		// maintenance mode, reboot, and exit maintenance mode. Once we have a future consensus-type that can be
-		// migrated to from raft is available, this test should be completed.
-		It("executes bootstrap to future-type - single node", func() {
```
tock-ibm: Keep this test but, rename it and comment on it that it checks changing a config value under maintenance (this is permitted!) that is unrelated to consensus type migration.

integration/raft/migration_test.go
```diff
@@ -526,6 +507,9 @@ func prepareTransition(
 	consensusTypeValue := extractOrdererConsensusType(current)
 	validateConsensusTypeValue(consensusTypeValue, fromConsensusType, fromMigState)
 	updateConfigWithConsensusType(toConsensusType, toConsensusMetadata, toMigState, updated, consensusTypeValue)
+	if toConsensusType == "BFT" {
+		updateBFTOrderersConfig(network, updated)
+	}
```
tock-ibm: I want to see a negative test that checks what happens when we move from etcdraft to BFT but the `config.ChannelGroup.Groups["Orderer"].Values["Orderers"]` entry is corrupt, does not pass validation, or missing

tock-ibm: in addition, a negative test that checks what happens when we move from etcdraft to BFT but the  `toConsensusMetadata []byte` which should contain the smartbft.Options is corrupt, does not pass validation, or missing

orderer/common/msgprocessor/maintenancefilter.go
```diff
@@ -8,11 +8,14 @@ package msgprocessor
 
 import (
 	"bytes"
+	"time"
+
+	"github.com/SmartBFT-Go/consensus/pkg/types"
```
tock-ibm: remove empty lines and rerun goimports

orderer/consensus/etcdraft/util.go
```diff
@@ -84,22 +84,29 @@ func MetadataHasDuplication(md *etcdraft.ConfigMetadata) error {
...
 	consensusTypeValue := &orderer.ConsensusType{}
 	if err := proto.Unmarshal(configValue.Value, consensusTypeValue); err != nil {
-		return nil, errors.Wrap(err, "failed to unmarshal consensusType config update")
+		return nil, nil, errors.Wrap(err, "failed to unmarshal consensusType config update")
+	}
+
+	if consensusTypeValue.Type != "etcdraft" {
+		if consensusTypeValue.Type == "BFT" {
+			return nil, consensusTypeValue, nil
+		}
```
tock-ibm: In case it is not etcdraft just do `return nil, consensusTypeValue, nil` if it is not BFT, the chain will panic

orderer/consensus/etcdraft/util_test.go
```diff
@@ -410,3 +412,18 @@ func TestVerifyConfigMetadata(t *testing.T) {
 		require.Nil(t, VerifyConfigMetadata(metadataWithExpiredConsenter, goodVerifyingOpts))
 	})
 }
+
+func TestMetadataFromConfigValue(t *testing.T) {
```
tock-ibm, line 418: update in line of comment on method

orderer/consensus/etcdraft/chain.go
tock-ibm: Yes, May is correct, we prevent it the maintenance filter.

orderer/consensus/etcdraft/chain.go
```diff
@@ -1115,7 +1115,26 @@ func (c *Chain) commitBlock(block *common.Block) {
...
 	// propose raft ConfChange if it adds/removes node.
-	configMetadata := c.newConfigMetadata(block)
+	c.logger.Infof("Detected configuration change")
+
+	// if the consensusType is bft, then configMetadata which represents the raft metadata should be nil
+	configMetadata, consensusType := c.newConfigMetadata(block)
+	c.logger.Infof("Detected configuration change: consensusType is: %s, configMetadata is: %v", consensusType, configMetadata)
+
+	if consensusType == nil {
+		c.logger.Infof("ConsensusType is %v", consensusType)
```
tock-ibm, line 1125: "ConsensusType is nil"

does this ever happen?

orderer/common/msgprocessor/maintenancefilter.go
MayRosenbaum: I did it since trying to reuse the function by making it public causes an import cycle which is not allowed. 
The cycle is between:
imports github.com/hyperledger/fabric/orderer/consensus/smartbft
imports github.com/hyperledger/fabric/orderer/common/msgprocessor

orderer/consensus/etcdraft/chain.go
MayRosenbaum: In the absence of this check, one of the chain tests fails.
Anyway, to be safe, it is better to check it

yacovm:
@tock-ibm what is left in your opinion so this can be merged?

orderer/common/msgprocessor/maintenancefilter.go
```diff
@@ -135,10 +135,20 @@ func (mf *MaintenanceFilter) inspect(configEnvelope *cb.ConfigEnvelope, ordererC
...
-		if nextOrdererConfig.ConsensusType() == "etcdraft" {
-			updatedMetadata := &protoetcdraft.ConfigMetadata{}
+		if nextOrdererConfig.ConsensusType() == "BFT" {
+			updatedMetadata := &smartbft.Options{}
 			if err := proto.Unmarshal(nextOrdererConfig.ConsensusMetadata(), updatedMetadata); err != nil {
-				return errors.Wrap(err, "failed to unmarshal etcdraft metadata configuration")
+				return errors.Wrap(err, "failed to unmarshal BFT metadata configuration")
+			}
+
+			_, err := validateBFTMetadataOptions(1, updatedMetadata)
```
tock-ibm: Either 
- remove the first return value, because it is not used here, and the id can be hard coded, or
- move the function `consensus/smartbft/configFromMetadataOptions` to a package `consensus/smartbft/util` and make it public. This would break the import cycle. I prefer this option, this way we don't duplicate code.

tock-ibm (approved):

Issue #4347, opened 2023-08-03
BFT Block Puller: A connection source that is updated immediately after a config block is sent to gossip 


tock-ibm:
Right now the connection source (set of orderers) is updated when a config block is committed.  We should change that so that the connection source is updated after a config block is validated and sent to gossip.

tock-ibm:
See also review comments: https://github.com/hyperledger/fabric/pull/4292#discussion_r1281540753

PR #4571, opened 2023-12-11
BFT Block Puller: updatable connection source

tock-ibm:
#### Type of change
- New feature

#### Description

Make the connection source update when a config block is processed by the block puller, not when it is committed.

#### Related issues

#4347


common/deliverclient/util.go
```diff
@@ -0,0 +1,55 @@
+/*
```
yacovm, line 1: don't we have any util package in our entire codebase we can move this function to?

tock-ibm: It got promoted from `orderer/common/cluster/util.go` to `common/deliverclient/util.go`, where I plan to put all the stuff that is common to both orderer and peer in the context of the block puller. Can't put it in package `common/util` nor `protoutil` because of import cycles. I am open to suggestions.

yacovm: I suppose the import cycle is because of `github.com/hyperledger/fabric/common/configtx`? Then can't you move it there?

yacovm: However if you think there are additional things to be placed here, I guess we can leave it as is.

tock-ibm: In terms of import cycles it can be placed in `common/configtx`, but I think it is not a good place to put code that parses blocks, as that package deals only with individual txs. I prefer to leave it where it is for the time being. I think that during the refactoring that is about to happen when we overhaul the orderer, several additional things will get moved into that util.go file.

yacovm (approved):

PR #4586, opened 2023-12-28
Remove chaincode legacy cli command

C0rWin:
This commit is to remove the peer cli chaincode command to get rid of the deprecated chaincode lifecycle.


satota2 (commented):
Hello @C0rWin 
My understanding is that the `peer chaincode invoke` and `query` commands are not deprecated and are currently still in use. So, I think they should not be removed yet. Could you please confirm this?

C0rWin:
> Hello @C0rWin My understanding is that the `peer chaincode invoke` and `query` commands are not deprecated and are currently still in use. So, I think they should not be removed yet. Could you please confirm this?

Yes, you are correct, will bring it back, thanks.

yeasy (approved):
Fine with the change. 
And should we move this part to somewhere as an individual package? in case someone want to check/read the code?

C0rWin:
> Fine with the change. And should we move this part to somewhere as an individual package? in case someone want to check/read the code?

wdym? it's already in separate package.

internal/peer/chaincode/chaincode.go
```diff
@@ -115,18 +97,8 @@ func resetFlags() {
 		"The channel on which this command should be executed")
 	flags.StringVarP(&policy, "policy", "P", common.UndefinedParamValue,
 		"The endorsement policy associated to this chaincode")
```
denyeart: I believe these flags can also be removed since they are not used for invoke or query:
- lang
- path
- version
- username
- policy

Also remove the corresponding variables above.

C0rWin: I do not think you can remove version, since it's part of the chaincode specification. I will remove the rest.

C0rWin: @denyeart `lang`, `version` and `path` are used in `ChaincodeSpec`, hence we cannot take them away, while I removed `policy` and `username`. 

See:

image

denyeart (approved):

satota2:
Hello @C0rWin @denyeart
Regarding this PR, I've noticed that the [documentation](https://hyperledger-fabric.readthedocs.io/en/latest/commands/peerchaincode.html) still includes examples of the sub-commands that were removed. My apologies if this is already known or scheduled to be addressed.
If necessary, I can create a patch to update the documentation. Would it be helpful if I did so?

denyeart:
> Hello @C0rWin @denyeart Regarding this PR, I've noticed that the [documentation](https://hyperledger-fabric.readthedocs.io/en/latest/commands/peerchaincode.html) still includes examples of the sub-commands that were removed. My apologies if this is already known or scheduled to be addressed. If necessary, I can create a patch to update the documentation. Would it be helpful if I did so?

My plan was to wait for @C0rWin to complete his series of removal PRs and then do a full assessment of any remaining references to be removed. I'll defer to @C0rWin whether he wants help with the docs or other areas.

C0rWin:
> > Hello @C0rWin @denyeart Regarding this PR, I've noticed that the [documentation](https://hyperledger-fabric.readthedocs.io/en/latest/commands/peerchaincode.html) still includes examples of the sub-commands that were removed. My apologies if this is already known or scheduled to be addressed. If necessary, I can create a patch to update the documentation. Would it be helpful if I did so?
> 
> My plan was to wait for @C0rWin to complete his series of removal PRs and then do a full assessment of any remaining references to be removed. I'll defer to @C0rWin whether he wants help with the docs or other areas.

I have the final mile to remove the actual code that implements these commands. Unfortunately, pieces of this code are convoluted with the new lifecycle and it's trickier given that we need to make sure not to break the validation part in case new peers will get connected with some legacy systems.

satota2:
@C0rWin @denyeart Thanks for your response. I understand the plan and the current situation now.
Please let me know if there is anything I can help with, including updating the documentation, at any time.

PR #4602, opened 2024-01-10
added time to the test to make it long enough

pfi79:


denyeart:
The gateway BFT test has been a common failure point, so hopefully the resolution is as simple as extending the timeout. I'll approve since there is no harm in trying, but let's see if developers more familiar with the test have any other comments before merging @andrew-coleman @tock-ibm

denyeart (approved):

andrew-coleman (approved):

PR #4612, opened 2024-01-17
Malicious Block Deliverer Test

arkadiPiven:
- Test update: Add a test that will checkout how the block deliverer acts with a malicious orderer in the network.

- Description: The test creates a network, creates blocks on the chain, replaces the orderers with mocks and communicates with the peer.

<arkadi.piven@ibm.com>

### DIFF
--- a/integration/ordererclient/envelope_creator.go
+++ b/integration/ordererclient/envelope_creator.go
@@ -14,7 +14,7 @@ import (
 	"github.com/hyperledger/fabric/integration/nwo"
 )
 
-func CreateBroadcastEnvelope(n *nwo.Network, entity interface{}, channel string, data []byte) *common.Envelope {
+func CreateBroadcastEnvelope(n *nwo.Network, entity interface{}, channel string, data []byte, optionalEnvelopeType ...common.HeaderType) *common.Envelope {
 	var signer *nwo.SigningIdentity
 	switch creator := entity.(type) {
 	case *nwo.Peer:
@@ -24,8 +24,14 @@ func CreateBroadcastEnvelope(n *nwo.Network, entity interface{}, channel string,
 	}
 	Expect(signer).NotTo(BeNil())
 
+	envelopeType := common.HeaderType_MESSAGE
+
+	if len(optionalEnvelopeType) > 0 {
+		envelopeType = optionalEnvelopeType[0]
+	}
+
 	env, err := protoutil.CreateSignedEnvelope(
-		common.HeaderType_MESSAGE,
+		envelopeType,
 		channel,
 		signer,
 		&common.Envelope{Payload: data},
--- a/integration/smartbft/mock_orderer.go
+++ b/integration/smartbft/mock_orderer.go
@@ -1,0 +1,246 @@
+/*
+Copyright IBM Corp All Rights Reserved.
+
+SPDX-License-Identifier: Apache-2.0
+*/
+
+package smartbft
+
+import (
+	"context"
+	"io"
+	"math"
+	"time"
+
+	"github.com/golang/protobuf/proto"
+	cb "github.com/hyperledger/fabric-protos-go/common"
+	ab "github.com/hyperledger/fabric-protos-go/orderer"
+	"github.com/hyperledger/fabric/common/flogging"
+	"github.com/hyperledger/fabric/common/util"
+	"github.com/hyperledger/fabric/internal/pkg/comm"
+	"github.com/hyperledger/fabric/protoutil"
+	"github.com/pkg/errors"
+)
+
+type MockOrderer struct {
+	address     string
+	ledgerArray []*cb.Block
+	logger      *flogging.FabricLogger
+	grpcServer  *comm.GRPCServer
+	channel     chan string
+}
+
+func (mo *MockOrderer) parseEnvelope(ctx context.Context, envelope *cb.Envelope) (*cb.Payload, *cb.ChannelHeader, *cb.SignatureHeader, error) {
+	payload, err := protoutil.UnmarshalPayload(envelope.Payload)
+	if err != nil {
+		return nil, nil, nil, err
+	}
+
+	if payload.Header == nil {
+		return nil, nil, nil, errors.New("envelope has no header")
+	}
+
+	chdr, err := protoutil.UnmarshalChannelHeader(payload.Header.ChannelHeader)
+	if err != nil {
+		return nil, nil, nil, err
+	}
+
+	shdr, err := protoutil.UnmarshalSignatureHeader(payload.Header.SignatureHeader)
+	if err != nil {
+		return nil, nil, nil, err
+	}
+
+	return payload, chdr, shdr, nil
+}
+
+func (mo *MockOrderer) Broadcast(server ab.AtomicBroadcast_BroadcastServer) error {
+	panic("implement me: Broadcast")
+}
+
+func (mo *MockOrderer) Deliver(server ab.AtomicBroadcast_DeliverServer) error {
+	ctx := server.Context()
+	addr := util.ExtractRemoteAddress(ctx)
+
+	/* Create DeliverServer here */
+
+	mo.logger.Infof("Attempting to read seek info message from %s", addr)
+	envelope, err := server.Recv()
+	if err == io.EOF {
+		mo.logger.Debugf("Received EOF from %s, hangup", addr)
+		return nil
+	}
+	if err != nil {
+		mo.logger.Warningf("Error reading from %s: %s", addr, err)
+		return err
+	}
+
+	status, err := mo.deliverBlocks(ctx, server, envelope)
+	if err != nil {
+		return err
+	}
+
+	if status != cb.Status_SUCCESS {
+		return err
+	}
+	if err != nil {
+		mo.logger.Warningf("Error sending to %s: %s", addr, err)
+		return err
+	}
+
+	mo.logger.Debugf("Waiting for new SeekInfo from %s", addr)
+
+	return nil
+}
+
+func (mo *MockOrderer) deliverBlocks(
+	ctx context.Context,
+	server ab.AtomicBroadcast_DeliverServer,
+	envelope *cb.Envelope,
+) (status cb.Status, err error) {
+	addr := util.ExtractRemoteAddress(ctx)
+	payload, chdr, _, err := mo.parseEnvelope(ctx, envelope)
+	if err != nil {
+		mo.logger.Warningf("error parsing envelope from %s: %s", addr, err)
+		return cb.Status_BAD_REQUEST, nil
+	}
+
+	if chdr.ChannelId != "testchannel1" {
+		panic("WHAT CHANNEL ARE YOU TALKING ABOUT?")
+	}
+
+	seekInfo := &ab.SeekInfo{}
+	if err = proto.Unmarshal(payload.Data, seekInfo); err != nil {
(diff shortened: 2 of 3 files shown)


integration/ordererclient/envelope_creator.go
```diff
@@ -14,6 +14,29 @@ import (
...
+		channel,
+		signer,
+		&common.Envelope{Payload: data},
+		0,
+		0,
+	)
+	Expect(err).NotTo(HaveOccurred())
+
+	return env
+}
```
tock-ibm: Instead of duplicating the code from below, make a common function that accepts as another parameter the type, i.e. common.HeaderType_ENDORSER_TRANSACTION or common.HeaderType_MESSAGE. This can be a package private function that is called by these two functions (CreateBroadcastEnvelope and) , and in this way other files will not change.

integration/smartbft/smartbft_block_deliverer_test.go
```diff
@@ -0,0 +1,205 @@
...
+	"time"
+
+	"github.com/hyperledger/fabric-protos-go/common"
+	ordererProtos "github.com/hyperledger/fabric-protos-go/orderer"
+	"github.com/hyperledger/fabric/integration/nwo"
+	"github.com/hyperledger/fabric/integration/ordererclient"
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+	"github.com/onsi/gomega/gbytes"
+	"github.com/tedsuo/ifrit"
```
tock-ibm: delete all the line breaks and call goimport

integration/smartbft/smartbft_test.go
```diff
@@ -1919,6 +1919,16 @@ var _ = Describe("EndToEnd Smart BFT configuration test", func() {
...
+
+		})
+		It("smartbft block deliverer test", func() {
+			smartBftBlockDelivererTest(
+				testDir,
+				client,
+				network,
+				networkProcess,
+				ordererProcesses,
+				peerProcesses)
```
tock-ibm: lets separate and upgrade this test case to its own suite. That is, make file smartbft_block_deliverer_test.go have its own test suite by including the respective parts at the beginning. This test is different enough from the rest to deserve its own suite. In addition, I think we will add test cases to it, which means it will likely grow.

```

var _ = Describe("Smart BFT Block Deliverer", func() {
	var (
		testDir          string
		client           *docker.Client
		network          *nwo.Network
		networkProcess   ifrit.Process
		ordererProcesses []ifrit.Process
		peerProcesses    ifrit.Process
	)

	BeforeEach(func() {
		...
	})

	AfterEach(func() {
		....
	})
...
```

integration/smartbft/smartbft_block_deliverer_test.go
```diff
@@ -0,0 +1,205 @@
...
+
+	Expect(err).NotTo(HaveOccurred())
+
+	/* Create networkConfig for multiple orderers */
+	networkConfig := nwo.MultiNodeSmartBFT()
+	networkConfig.Channels = nil
+
+	channel := "testchannel1"
+
+	fmt.Println(client.Endpoint())
```
tock-ibm: Please don't use fmt.Println in order to track the progress of the test. Use `By(fmt.Sprintf(""))` etc

integration/smartbft/smartbft_block_deliverer_test.go
```diff
@@ -0,0 +1,205 @@
...
+
+	/* Create a network from the networkConfig */
+	startPort := StartPort()
+	print(startPort)
```
tock-ibm: Use By()

integration/smartbft/smartbft_block_deliverer_test.go
```diff
@@ -0,0 +1,205 @@
...
+		BlockPath:  network.OutputBlockPath(channel),
+		ClientAuth: network.ClientAuthRequired,
+	})
+	Expect(err).NotTo(HaveOccurred())
+	/* Close the mocks */
+
+	time.Sleep(10 * time.Second)
+
+	fmt.Println(nwo.GetLedgerHeight(network, network.Peers[0], channel))
+	Expect(nwo.GetLedgerHeight(network, network.Peers[0], channel)).To(Equal(11))
```
tock-ibm: Use eventually, w/o sleep.

integration/smartbft/smartbft_block_deliverer_test.go
```diff
@@ -0,0 +1,205 @@
...
+
+	time.Sleep(10 * time.Second)
```
tock-ibm: why wait here?

integration/smartbft/smartbft_block_deliverer_test.go
```diff
@@ -0,0 +1,205 @@
...
+				finishSending = true
+			case <-time.After(time.Second):
+			}
+			if finishSending {
+				return
+			}
+		}
+	}()
+	Eventually(ordererRunners[1].Err(), 10*time.Second).Should(gbytes.Say("Delivering proposal, writing block 10"))
+	closeTxProducer <- true
```
tock-ibm: This may fill the orderer with way too many blocks. why not run this precisely the number of desired times "K"?
Moreover, since each orderer is committing asynch, you have to check that every orderer had arrive to K, or at least, check the correct runner (you check 1 but then grab the ledger from 0, below). I think it will be sfare to check all orderers had reached K.

integration/smartbft/smartbft_block_deliverer_test.go
```diff
@@ -0,0 +1,205 @@
...
+	var ledgerArray []*common.Block
+	for i := uint64(0); i < fileLedger.Height(); i++ {
+		block, err := fileLedger.RetrieveBlockByNumber(i)
+		Expect(err).NotTo(HaveOccurred())
+		ledgerArray = append(ledgerArray, block)
+	}
+	/* The ledger now saved and can be accessed at ledgerArray */
+
+	/* Create orderer mocks and make sure they can access the ledger */
+	var mocksArray [4]*MockOrderer
```
tock-ibm: make a slice with dynamic size. In later tests we may want to check how deliverer works against a bigger cluster.

integration/smartbft/smartbft_block_deliverer_test.go
```diff
@@ -0,0 +1,205 @@
...
+
+		secOpts := comm.SecureOptions{
+			UseTLS:      true,
+			Certificate: serverTLSCerts[tlsCertPath],
+			Key:         serverTLSKeys[tlsKeyPath],
+		}
+		mo, err := NewMockOrderer(network.OrdererAddress(orderer, nwo.ListenPort), ledgerArray, secOpts)
+		Expect(err).NotTo(HaveOccurred())
+		mo.logger.Infof("Mock orderer started at port %v", network.OrdererAddress(orderer, nwo.ListenPort))
+		mocksArray[i] = mo
```
tock-ibm: append to a slice

integration/smartbft/smartbft_block_deliverer_test.go
```diff
@@ -0,0 +1,205 @@
...
+		block, err := fileLedger.RetrieveBlockByNumber(i)
+		Expect(err).NotTo(HaveOccurred())
+		ledgerArray = append(ledgerArray, block)
+	}
+	/* The ledger now saved and can be accessed at ledgerArray */
+
+	/* Create orderer mocks and make sure they can access the ledger */
+	var mocksArray [4]*MockOrderer
+
+	for i, orderer := range network.Orderers {
```
tock-ibm: we can decouple the number of orderers and mocks. We can generate the ledegr from 4 orderers but then span the ledger across N mock.

integration/smartbft/smartbft_block_deliverer_test.go
```diff
@@ -0,0 +1,205 @@
...
+			Certificate: serverTLSCerts[tlsCertPath],
+			Key:         serverTLSKeys[tlsKeyPath],
+		}
+		mo, err := NewMockOrderer(network.OrdererAddress(orderer, nwo.ListenPort), ledgerArray, secOpts)
+		Expect(err).NotTo(HaveOccurred())
+		mo.logger.Infof("Mock orderer started at port %v", network.OrdererAddress(orderer, nwo.ListenPort))
+		mocksArray[i] = mo
+	}
+
+	/* Create peer */
```
tock-ibm: up until here it is the preparation of the mocks with ledgers. I think you can extract most of it to utility functions call this function in a "Before". 

From here on, it's the actual tests. Currently you have a mock that behaves like a correct orderer. This is test number 1 and it works fine. It should be in an `It("correct mock")`

In additional tests we will change the mock to exhibit byzantine behavior, each type of behavior in a separate "It". Let do that in separate PRs, lets finish this basic one first.

integration/smartbft/mock_orderer.go
```diff
@@ -0,0 +1,256 @@
...
+		if err == io.EOF {
+			mo.logger.Debugf("Received EOF from %s, hangup", addr)
+			return nil
+		}
+		if err != nil {
+			mo.logger.Warningf("Error reading from %s: %s", addr, err)
+			return err
+		}
+
+		status, err := mo.deliverBlocks(ctx, server, envelope)
```
tock-ibm: This delivers blocks like a correct orderer. This is good for the first test case.
In the next PR we'll add byzantine behavior here.

integration/smartbft/smartbft_block_deliverer_test.go
arkadiPiven: Done

integration/ordererclient/envelope_creator.go
arkadiPiven: Done.

integration/smartbft/smartbft_test.go
arkadiPiven: Done

integration/smartbft/smartbft_block_deliverer_test.go
arkadiPiven: Done

integration/smartbft/smartbft_block_deliverer_test.go
arkadiPiven: Done

integration/smartbft/smartbft_block_deliverer_test.go
arkadiPiven: Done

integration/smartbft/smartbft_block_deliverer_test.go
arkadiPiven: Done

integration/smartbft/smartbft_block_deliverer_test.go
arkadiPiven: Done

integration/smartbft/smartbft_block_deliverer_test.go
arkadiPiven: Done

integration/smartbft/smartbft_block_deliverer_test.go
arkadiPiven: Done

integration/smartbft/smartbft_block_deliverer_test.go
arkadiPiven: Done

integration/smartbft/smartbft_block_deliverer_test.go
arkadiPiven: Done

tock-ibm (dismissed):

tock-ibm (commented):
The first commit does not need to be there.... It is an artifact of the rebase maybe?

tock-ibm (approved):

PR #4633, opened 2024-01-22
Check the TLS.Config error immediately

yeasy:
The patchset checks the TLS.Config()'s potential return error immediately.

Change-Id: I16a39e0f0152179300d5626ffdffb5b2d16f5a6d

#### Type of change

#### Description

The patchset checks the TLS.Config()'s potential return error immediately after it is initialized. 

This is to follow Golang's style.

#### Additional details

N/A

#### Related issues

N/A

#### Release Note

N/A

### DIFF
--- a/common/fabhttp/server.go
+++ b/common/fabhttp/server.go
@@ -124,11 +124,10 @@ func (s *Server) Listen() (net.Listener, error) {
 	if err != nil {
 		return nil, err
 	}
-	tlsConfig, err := s.options.TLS.Config()
-	if err != nil {
+
+	if tlsConfig, err := s.options.TLS.Config(); err != nil {
 		return nil, err
-	}
-	if tlsConfig != nil {
+	} else if tlsConfig != nil {
 		listener = tls.NewListener(listener, tlsConfig)
 	}
 	return listener, nil


common/fabhttp/server.go
```diff
@@ -124,11 +124,10 @@ func (s *Server) Listen() (net.Listener, error) {
...
 		return nil, err
 	}
-	tlsConfig, err := s.options.TLS.Config()
-	if err != nil {
+
+	if tlsConfig, err := s.options.TLS.Config(); err != nil {
 		return nil, err
-	}
-	if tlsConfig != nil {
+	} else if tlsConfig != nil {
```
denyeart, line 130: In idiomatic Go the normal flow should not be indented within an 'else'.

See:
https://go.dev/wiki/CodeReviewComments#indent-error-flow

The original error can be handled inline like you've done, or it can be handled in a follow-on 'if' statement like it was previously in the code. Either way is ok, but most of the idiomatic Go examples and most of the Fabric code use the pattern as it was previously (separate 'if' statement). I'm not in favor of changing every occurrence through PRs, as it bloats the commits and unnecessarily creates a divergence between main branch and the prior release branches which makes backports more difficult. So honestly, I think it would be better to just close this one rather than fix the indenting.

yeasy: OK, close it

Issue #4566, opened 2023-12-06
BFT Block Puller: in orderer


tock-ibm:
Use BFT block deliverer relevant components in the orderer chain + follower

- Block receiver
- Censorship monitor
- Config updated dynamically
- Orderer source addresses  updated dynamically

- [x] Detect the self endpoint
- [x] Allow the orderer to choose a block-replication-policy: between a naive scheme (existing) and the BFT resistant scheme
- [x] Allow the BFTDeliverer to skip the self endpoint when pulling blocks
- [x] Implement the BFT resistant scheme into a new BFT synchronizer
- [x] Target height selection: 
      (1) simple (based on height by endpoints and F), 
      (2) optimistic (based on verifying the last block from each orderer, selecting max height from those that verify). Deferred to a separate issue #4698 .

PR #4637, opened 2024-01-28
BFT synchronizer in orderer

tock-ibm:
#### Type of change

- New feature

#### Description

Create a BFT synchronizer for the smartbft chain

#### Related issues

#4566

### DIFF
--- a/core/deliverservice/deliveryclient.go
+++ b/core/deliverservice/deliveryclient.go
@@ -280,7 +280,7 @@ func (d *deliverServiceImpl) createBlockDelivererBFT(chainID string, ledgerInfo 
 		dcBFT.TLSCertHash = util.ComputeSHA256(cert.Certificate[0])
 	}
 
-	dcBFT.Initialize(d.conf.ChannelConfig)
+	dcBFT.Initialize(d.conf.ChannelConfig, "")
 
 	return dcBFT, nil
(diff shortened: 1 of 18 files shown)


common/ledger/blockledger/fileledger/impl.go
```diff
@@ -110,6 +110,16 @@ func (fl *FileLedger) Height() uint64 {
 	return info.Height
 }
 
+// GetCurrentBlockHash returns the hash of the current block's header
```
yacovm: why do we need to expand the API of the ledger? You can just retrieve the current block, now it's cached.

tock-ibm: In the orderer we can, but in the peer we sometime can't, because of a snapshot. Hence, the BFTDeliverer expects this API, as it is common to the peer as well.

yacovm: Can you explain where in the production code this is used? I can't find it in the PR.

tock-ibm: After a second look, you are correct, this is not needed in the orderer. 
I removed the respective changes to the `fileledger`.

tock-ibm:
This was completed in another branch / PR

PR #4640, opened 2024-01-28
Throttle greedy clients in orderer Broadcast API

yacovm:
This commit implements a shared rate limiter for connections authenticated with mutual TLS. The throttling and rate limiting is applied at both the Authority Key Identifier and Subject Key Identifier level, which guarantees no single organization or client can bombard the orderer with endless transactions on the expense of others.

The effective rate is divided across all clients and orgs for a time period specified by InactivityTimeout.

By default this feature is turned off.


orderer/common/throttle/ratelimit.go
```diff
@@ -0,0 +1,184 @@
...
+	stop    chan struct{}
+	lock    sync.RWMutex
+	clients map[string]*clientState
+}
+
+type clientState struct {
+	lock       sync.Mutex
+	lastSeen   atomic.Value
+	budget     chan struct{}
+	budgetUsed uint32
```
pfi79, line 40: maybe make the variable budgetUsed atomic.Uint32

orderer/common/throttle/ratelimit.go
```diff
@@ -0,0 +1,184 @@
...
+}
+
+func (srl *SharedRateLimiter) LimitRate(client string) {
+	srl.once.Do(srl.init)
+
+	cs := srl.getOrCreateState(client)
+
+	srl.lock.RLock()
+	numClients := len(srl.clients)
+	srl.lock.RUnlock()
```
pfi79: Maybe make a separate getLen method to improve readability? Above there is a call to getOrCreateState method inside which srl.lock is also handled.

orderer/common/throttle/ratelimit.go
```diff
@@ -0,0 +1,184 @@
...
+
+func (srl *SharedRateLimiter) Stop() error {
+	select {
+	case <-srl.stop:
+		return fmt.Errorf("stop must be called only once")
+	default:
+		close(srl.stop)
+		return nil
+	}
+}
```
pfi79, line 134: Where is the call to this method? Can it be called from 2 different goroutines? If yes, there may be a 2nd channel closure.

orderer/common/throttle/ratelimit.go
```diff
@@ -0,0 +1,184 @@
...
+	srl.lock.RLock()
+	for _, cs := range srl.clients {
+		fillBudgetForClient(cs)
+	}
+	clientsToBePruned := srl.clientsInactiveForTooLong()
+	srl.lock.RUnlock()
+
+	// Next, prune the clients that were lastly seen.
+	// It matters not if that client had activity since unlocking the mutex and now.
+	srl.lock.Lock()
```
pfi79, line 166: Maybe check the length of clientsToBePruned beforehand to avoid calling Lock/Unlock unnecessarily

orderer/common/throttle/ratelimit.go
yacovm: why? it's just a wrapper for uint32, it doesn't add much.

orderer/common/throttle/ratelimit.go
yacovm: It's only called in the test. I added a comment about it not being thread safe.

orderer/common/throttle/ratelimit.go
pfi79: So that those who will later accompany the code, could not just do simple mathematical operations (+, -, ++, --, =).

orderer/common/throttle/ratelimit.go
yacovm: I see your point but if we have no clients connected then we are not delaying anyone, and locking is not computationally expensive. For code readability it's better to keep it simple.

orderer/common/throttle/ratelimit.go
yacovm: done

orderer/common/throttle/ratelimit.go
pfi79: The wrapper creates a new data type that will limit the opreations that can be produced.

But if you leave it as it is, that's fine with me.

yacovm: Thanks for your agreement

sampleconfig/orderer.yaml
```diff
@@ -66,6 +66,14 @@ General:
     MaxRecvMsgSize: 104857600
     # Max message size in bytes the GRPC server and client can send
     MaxSendMsgSize: 104857600
+    Throttling:
+        # Rate is the maximum rate (transactions per second) for all clients combined.
+        # A zero rate disables throttling.
```
denyeart, line 79: Since the orderer.yaml serves as the only documentation for this feature, can you expand a bit. e.g. mention that it is for transaction submission, that it applies at both the client and org level, and mention the behavior when threshold is exceeded.

yacovm: done

pfi79:
@yacovm 
Why didn't you use an off-the-shelf library like [golang.org/x/time/rate](https://pkg.go.dev/golang.org/x/time/rate)?

yacovm:
> @yacovm Why didn't you use an off-the-shelf library like [golang.org/x/time/rate](https://pkg.go.dev/golang.org/x/time/rate)?

I am not sure it fits our need. I need a dynamic rate limiter that limits on a per client basis and the budget is adjusted dynamically according to how many clients are connected.

pfi79:
> > @yacovm Why didn't you use an off-the-shelf library like [golang.org/x/time/rate](https://pkg.go.dev/golang.org/x/time/rate)?
> 
> I am not sure it fits our need. I need a dynamic rate limiter that limits on a per client basis and the budget is adjusted dynamically according to how many clients are connected.

I used it for dynamic loading. Conditionally: the TPS of the system had to be X and had to be divided by all nodes. Nodes could crash and restart. And the throughput of the system should not change.

C0rWin:
There are already rate limiters implemented for Endorsering, Deliver, and Gateway services; wouldn't it be better to extend its usage for ordering service as well? See here: https://github.com/hyperledger/fabric/pull/647 . I understand the existing approach probably doesn't provide the desired granularity, though still, even putting rate limits with the grpc interceptor is a little bit cleaner approach rather than extending the broadcast handler.

C0rWin:
@yacovm, btw, do you also intend to add metrics to expose limiter budget utilization so operators can trace down setup and alerts to notify whenever the limit is exhausted per client or organization? Now, where the model is to use gateway service, meaning peers are collecting endorsements and submitting transactions, this might be crucial.

yacovm:
> > > @yacovm Why didn't you use an off-the-shelf library like [golang.org/x/time/rate](https://pkg.go.dev/golang.org/x/time/rate)?
> > 
> > 
> > I am not sure it fits our need. I need a dynamic rate limiter that limits on a per client basis and the budget is adjusted dynamically according to how many clients are connected.
> 
> I used it for dynamic loading. Conditionally: the TPS of the system had to be X and had to be divided by all nodes. Nodes could crash and restart. And the throughput of the system should not change.

I think you're talking about sending transactions and not receiving them, am I mistaken?

yacovm:
> There are already rate limiters implemented for Endorsering, Deliver, and Gateway services; wouldn't it be better to extend its usage for ordering service as well? See here: #647 . I understand the existing approach probably doesn't provide the desired granularity, though still, even putting rate limits with the grpc interceptor is a little bit cleaner approach rather than extending the broadcast handler.

They are not rate limiters, they are simply locks that limit too many gRPC streams being opened. This does not address the problem solved here.

I think what I could have done alternatively was to make a proxy `Broadcast(srv ab.AtomicBroadcast_BroadcastServer)` wrapper which throttles right after calling `Recv()` and then forwards and returns the backend `Recv()` of the gRPC stream, and then the `broadcast.go` would have stayed intact. 

@C0rWin I made the above modification, now the broadcast service implementation remains as-is.

yacovm:
> @yacovm, btw, do you also intend to add metrics to expose limiter budget utilization so operators can trace down setup and alerts to notify whenever the limit is exhausted per client or organization? Now, where the model is to use gateway service, meaning peers are collecting endorsements and submitting transactions, this might be crucial.

The idea is cool but I have no intention of doing this, as this piece of work is not meant to be used in this manner.
This is only to prevent a client from spamming an orderer with too many transactions.
As long as you're an "average" client and you don't over saturate your budget on a timely manner, you should be good.

C0rWin (approved):

PR #4664, opened 2024-02-02
Bump github.com/containerd/containerd from 1.6.18 to 1.6.26

denyeart:
Bump github.com/containerd/containerd from 1.6.18 to 1.6.26
Only move golang.org/x/tools to v0.14.0 (later versions introduce an issue with ./scripts/metrics_doc.sh check)
Only move google.golang.org/protobuf to v1.31.0 (later versions introduce an issue with chaincode build)

Update tests to reflect message updates in latest grpc v1.61.0.

PR #4668, opened 2024-02-05
fix error "runtime error: index out of range [0] with length 0"

pfi79:
fix for this [error](https://github.com/hyperledger/fabric/actions/runs/7764041048/job/21228336098)

### DIFF
--- a/integration/gateway/gateway_bft_test.go
+++ b/integration/gateway/gateway_bft_test.go
@@ -167,9 +167,19 @@ var _ = Describe("GatewayService with BFT ordering service", func() {
 		rpcErr = status.Convert(err)
 		Expect(rpcErr.Message()).To(Equal("insufficient number of orderers could successfully process transaction to satisfy quorum requirement"))
 
-		peerLog := peerGinkgoRunner[0].Err().Contents()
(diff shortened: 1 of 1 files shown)


denyeart:
Thank you!  I'm hitting the error consistently while trying to update dependencies. Even with your suggested update I'm still hitting the error. I'll look into root cause...

pfi79:
> Thank you! I'm hitting the error consistently while trying to update dependencies. Even with your suggested update I'm still hitting the error. I'll look into root cause...

I see what this is about. I'll fix it. The grpc has been updated and the log text in it has changed.

hyperledger/fabric/integration/smartbft/smartbft_test.go 1646 line. need to be replaced
```
pickfirstBalancer: UpdateSubConnState.*IDLE connection error
```
on
```
Received SubConn state update:.*IDLE ConnectionError
```

pfi79:
hyperledger/fabric/integration/gateway/gateway_bft_test.go

line 236 need to be replaced
```
pickfirstBalancer: UpdateSubConnState.*%d
```
on
```
Received SubConn state update:
```

line 242 need to be replaced (don't forget the space at the end of the line)
```
pickfirstBalancer: UpdateSubConnState: 
```
on
```
Received SubConn state update: 
```

denyeart:
Thank you, I've updated https://github.com/hyperledger/fabric/pull/4664 accordingly. I'll close this one now.

PR #4673, opened 2024-02-08
Block Censorship Test

arkadiPiven:
* Test update: Add a test that will checkout how the block deliverer acts with a malicious orderer that censors blocks.

Description: The test creates a network, creates blocks on the chain, replaces the orderers with mocks and communicates with the peer. One of the orderers becomes malicious and after delivering a few data blocks he will stop. Afterwards the peer will ask for another orderer to deliver the data blocks, which should be successful, since he should be a honest orderer.

arkadi.piven@ibm.com

### DIFF
--- a/integration/smartbft/mock_orderer.go
+++ b/integration/smartbft/mock_orderer.go
@@ -122,14 +122,10 @@ func (mo *MockOrderer) deliverBlocks(
 		mo.logger.Warningf("[channel: %s] Received seekInfo message from %s with missing start or stop %v, %v", chdr.ChannelId, addr, seekInfo.Start, seekInfo.Stop)
 		return cb.Status_BAD_REQUEST, nil
 	}
-	mo.logger.Infof("[%s] Recieved seekInfo.Start %v", mo.address, seekInfo.Start)
-
-	mo.logger.Infof("[channel: %s] Received seekInfo (%p) %v from %s", chdr.ChannelId, seekInfo, seekInfo, addr)
 
 	ledgerIdx := seekInfo.Start.GetSpecified().Number
 	number := uint64(1)
 	ledgerLastIdx := uint64(len(mo.ledgerArray) - 1)
-	mo.logger.Infof("<%s> delivering blocks", mo.address)
 	var stopNum uint64
 
 	switch stop := seekInfo.Stop.Type.(type) {
@@ -189,9 +185,7 @@ func (mo *MockOrderer) deliverBlocks(
 			return status, nil
 		}
 
-		// increment block number to support FAIL_IF_NOT_READY deliver behavior
 		number++
-		//mo.logger.Infof("[channel: %s] Delivering block [%d] for (%p) for %s", chdr.ChannelId, block.Header.Number, seekInfo, addr)
 
 		block2send := &cb.Block{
 			Header:   block.Header,
@@ -199,12 +193,10 @@ func (mo *MockOrderer) deliverBlocks(
 			Data:     block.Data,
 		}
 
-		onlyHeaderType := true
 		if seekInfo.ContentType == ab.SeekInfo_HEADER_WITH_SIG && !protoutil.IsConfigBlock(block) {
 			mo.logger.Infof("ASKED FOR HEADER WITH SIG")
 			block2send.Data = nil
 		} else if mo.censorDataMode {
-			onlyHeaderType = false
 			if !mo.peerFirstTry {
 				mo.malicious = true
 			}
@@ -216,12 +208,6 @@ func (mo *MockOrderer) deliverBlocks(
 
 		blockResponse := &ab.DeliverResponse{
 			Type: &ab.DeliverResponse_Block{Block: block2send},
-		}
-
-		if !onlyHeaderType {
-			mo.logger.Infof("%s >>>> SENDING DATA BLOCK [%d]######", mo.address, block2send.Header.Number)
-		} else {
-			mo.logger.Infof("%s >>>> SENDING HEADER BLOCK [%d]######", mo.address, block2send.Header.Number)
 		}
 
 		err = server.Send(blockResponse)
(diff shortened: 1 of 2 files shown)


integration/smartbft/mock_orderer.go
```diff
@@ -184,19 +180,25 @@ func (mo *MockOrderer) deliverBlocks(
...
-		mo.logger.Infof("[channel: %s] Delivering block [%d] for (%p) for %s", chdr.ChannelId, block.Header.Number, seekInfo, addr)
 
 		block2send := &cb.Block{
 			Header:   block.Header,
 			Metadata: block.Metadata,
 			Data:     block.Data,
 		}
+
 		if seekInfo.ContentType == ab.SeekInfo_HEADER_WITH_SIG && !protoutil.IsConfigBlock(block) {
-			mo.logger.Infof("ASKED FOR HEADER WITH SIG")
```
tock-ibm, line 197: Don't remove these logs, convert them to debug, lower case, with details, like block number, etc

integration/smartbft/mock_orderer.go
```diff
@@ -119,12 +119,9 @@ func (mo *MockOrderer) deliverBlocks(
 		return cb.Status_BAD_REQUEST, nil
 	}
 
-	mo.logger.Infof("[channel: %s] Received seekInfo (%p) %v from %s", chdr.ChannelId, seekInfo, seekInfo, addr)
```
tock-ibm, line 122: Don't remove this log, we don't need the pointer, do `%+v` and make sure it prints the start number ("ledgerIdx").

integration/smartbft/smartbft_block_deliverer_test.go
```diff
@@ -199,7 +186,47 @@ var _ = Describe("Smart BFT Block Deliverer", func() {
...
+		peerRunner.Command.Env = append(peerRunner.Command.Env, "FABRIC_LOGGING_SPEC=debug")
+		peerProcesses = ifrit.Invoke(peerRunner)
+		Eventually(peerProcesses.Ready(), network.EventuallyTimeout).Should(BeClosed())
+
+		_, err = network.PeerAdminSession(p0, commands.ChannelJoin{
+			BlockPath:  network.OutputBlockPath(channel),
+			ClientAuth: network.ClientAuthRequired,
+		})
+		Expect(err).NotTo(HaveOccurred())
+
```
tock-ibm: In order to verify that the peer indeed detects a censorship event, we have to insert here a check that the peer's log includes a log message that arrives from this line:
`m.logger.Warningf("Block censorship detected, block source endpoint: %s", m.fetchSources[m.blockSourceIndex])`
internal/pkg/peer/blocksprovider/bft_censorship_monitor.go L177

for example, take a look at how it's done in integration/smartbft/smartbft_test.go L120
i.e.

`Eventually(ordererRunners[1].Err(), network.EventuallyTimeout, time.Second).Should(gbytes.Say("Message from 1"))`

change the runner and expected message.

integration/smartbft/mock_orderer.go
```diff
@@ -184,19 +180,25 @@ func (mo *MockOrderer) deliverBlocks(
...
 			block2send.Data = nil
+		} else if mo.censorDataMode {
+			if !mo.peerFirstTry {
+				mo.malicious = true
+			}
+			if mo.sentCount >= 5 && mo.malicious {
+				break
+			}
 		}
+		mo.peerFirstTry = true
```
tock-ibm: This style of controlling the mock will only cover one use case.
We want to achieve the same goal with a more flexible API on the mock, not by setting a single private field `mo.censorDataMode`.

I am thinking about the following:

on every mock you can have the following functions:

// will deliver either data/headers until this point "n" and will then wait for further instructions but NOT break the delivery loop. If a subsequent call DeliverUntil(n2 uint64) (n2>n) the mock will continue to deliver until n2.
DeliverUntil(number uint64)

// answers true if the mock was last requested to deliver data
IsDataDeliverer() bool

// answers true if the mock was last requested to deliver headers
IsHeaderDeliverer() bool

// stops the delivery of blocks but does not stop the mock server.
// If the mock was instructed to deliver until n=5, it will wait there, and then only when instructed to stop, it will exit the loop
StopDelivery()

With this, the "block censorship" test case can then be implemented by:

1. mark all mocks by DeliverUntil(5)
2. discover who IsDataDeliverer and who are IsHeaderDeliverer (with eventually)
3. let the header deliverers continue until the end by marking them DeliverUntil(11)
4. etc...

I hope you see how this style will allow us to test many more test cases with ease.

The "normal" case is just
1.  mark all mocks by DeliverUntil(11)
2. etc...

**PS** - this refactoring can be done in a separate PR. Let's finish the important stuff first (Block censorship detected, etc) and do this after this PR is merged.

integration/smartbft/mock_orderer.go
```diff
@@ -184,19 +180,25 @@ func (mo *MockOrderer) deliverBlocks(
...
+
 		if seekInfo.ContentType == ab.SeekInfo_HEADER_WITH_SIG && !protoutil.IsConfigBlock(block) {
-			mo.logger.Infof("ASKED FOR HEADER WITH SIG")
 			block2send.Data = nil
+		} else if mo.censorDataMode {
+			if !mo.peerFirstTry {
+				mo.malicious = true
+			}
+			if mo.sentCount >= 5 && mo.malicious {
+				break
```
tock-ibm: I don't think it is correct to break the loop here. You should not return from the DeliverBlocks loop but rather wait here until instructed to continue or exit the loop.
Exiting the loop here is like server failure, the BFTDeliverer will immediatlly detect this choose another mock. What we need in order to check the the peer detects detect censorship is for this server to remain silent but not break the connection.

integration/smartbft/mock_orderer.go
tock-ibm: https://github.com/hyperledger/fabric/issues/4724

tock-ibm (approved):

Issue #4008, opened 2023-02-08
Add unit test for SmartBFT orderer chain


Param-S:
- [x] Chain unit test infrastructure
- [x] Start a network of 4 node, detect leader, submit TXs
- [x] Restart network
- [x] Stop leader, detect new leader, submit TXs
- [x] Restart node when it's behind, check Sync()
- [x] Reconfigure: add/remove nodes
- [x] Reconfigure while node is down, restart, check Sync()

PR #4679, opened 2024-02-14
BFT chain unit tests: create a new chain without errors

MayRosenbaum:
#### Type of change

- Test update

#### Description

The purpose of this PR is to create BFT chain unit tests. 
The first step is to create an active chain using mocks. 

#### Related issues

issue #4008


orderer/consensus/smartbft/chain.go
```diff
@@ -64,20 +68,82 @@ type signerSerializer interface {
 	Serialize() ([]byte, error)
 }
 
+//go:generate mockery --dir . --name ConsenterSupport --case underscore --with-expecter=true --output mocks
+
+// ConsenterSupport provides the resources available to a Consenter implementation.
+type ConsenterSupport interface {
```
tock-ibm: this mock already exists here:
`orderer/consensus/mocks/mock_consenter_support.go`
This is a rare case where you don't need to recreate the mock.

orderer/consensus/smartbft/chain.go
```diff
@@ -64,20 +68,82 @@ type signerSerializer interface {
...
+	// of the given channel, or error if connection cannot be established, or
+	// the channel wasn't configured
+	Remote(channel string, id uint64) (*cluster.RemoteContext, error)
+	// Configure configures the communication to connect to all
+	// given members, and disconnect from any members not among the given
+	// members.
+	Configure(channel string, members []cluster.RemoteNode)
+	// Shutdown shuts down the communicator
+	Shutdown()
+}
```
tock-ibm: you don't have to copy the fields,  this is better:

```
type Communicator interface {
    cluster.Communicator
}
```

plus, you can do this in the test files, for example define a file `interfaces_test.go` where you have these definitions. This has the anvantace of not cluttering the implementation.
It makes the implementation cleaner, and you can keep `cluster.Communicator` in the chain.

tock-ibm (commented):
A couple of comments about the mocks.

tock-ibm:
completed in #4704

PR #4685, opened 2024-02-19
Detect your own endpoint in HeightsByEndpoints()

yacovm:
Currently, when an orderer node chooses a single node endpoint to pull blocks from (and the rest to pull headers from), it may pick itself, which is a problem because this endpoint will never bring it any blocks since it is itself.

This commit makes the function HeightsByEndpoint return the endpoint of yourself, such that the caller can filter it out.

The identification of your own endpoint is done by looking at the TLS certificate of the remote node and comparing it with your own.


orderer/common/cluster/deliver.go
```diff
@@ -115,15 +116,26 @@ func (p *BlockPuller) PullBlock(seq uint64) *common.Block {
...
 
 // HeightsByEndpoints returns the block heights by endpoints of orderers
-func (p *BlockPuller) HeightsByEndpoints() (map[string]uint64, error) {
+func (p *BlockPuller) HeightsByEndpoints() (map[string]uint64, string, error) {
 	endpointsInfo := p.probeEndpoints(0)
 	res := make(map[string]uint64)
+	var myEndpoint string
+
+	var myCert *x509.Certificate
+	if len(p.TLSCert) > 0 {
```
tock-ibm: This cannot fail? no need to handle the error?

orderer/common/cluster/deliver.go
```diff
@@ -115,15 +116,26 @@ func (p *BlockPuller) PullBlock(seq uint64) *common.Block {
...
+		myCert, _ = x509.ParseCertificate(p.TLSCert)
+	}
+
 	for endpoint, endpointInfo := range endpointsInfo.byEndpoints() {
+		if myCert != nil && reflect.DeepEqual(myCert, endpointInfo.certificate) {
```
tock-ibm: Just a nit: endpointInfo shadows the name of the struct

```
Declaration of 'endpointInfo' shadows declaration at deliver.go 
```
L531

orderer/common/cluster/deliver.go
```diff
@@ -115,15 +116,26 @@ func (p *BlockPuller) PullBlock(seq uint64) *common.Block {
...
+			myEndpoint = endpoint
+		}
 		endpointInfo.conn.Close()
 		res[endpoint] = endpointInfo.lastBlockSeq + 1
 	}
+
 	p.Logger.Info("Returning the heights of OSNs mapped by endpoints", res)
-	return res, endpointsInfo.err
+	return res, myEndpoint, endpointsInfo.err
```
tock-ibm, line 134: the log above should be `Infof()`,  and add `myEndpoint` as well

tock-ibm (commented):
Looks good, just a few nits

orderer/common/cluster/deliver.go
yacovm: why handle the error? if our TLS cert cannot be parsed we have bigger problems, no?

orderer/common/cluster/deliver.go
yacovm: i know, but this is irrelevant to this PR.

tock-ibm (approved):

PR #4686, opened 2024-02-19
Catch up installation procedures in main to v2.5.6

satota2:
#### Type of change

- Documentation update

#### Description

Since the install procedures in the documentation for each version refer to the install scripts in the main repository, the install scripts in the main repository should also be updated to catch up with v2.5.6.


denyeart (approved):
Thank you!

PR #4689, opened 2024-02-20
Allow error to be wrapped

yeasy:
Change-Id: I3620094d9aece9151bf5dd08005673c2610aa041

#### Type of change

#### Description

Change output fmt from %s to %w, which is supported by errors and fmt 
to wrap error.

#### Additional details

N/A

#### Related issues

N/A

#### Release Note

N/A

### DIFF
--- a/ccaas_builder/cmd/build/main.go
+++ b/ccaas_builder/cmd/build/main.go
@@ -88,12 +88,12 @@ func run() error {
 	}
 
 	if err := copy.Copy(metadataDir, outputDir); err != nil {
-		return fmt.Errorf("failed to copy build metadata folder: %s", err)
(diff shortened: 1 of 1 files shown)


denyeart:
My opinion is that we shouldn't be changing error handling as a series of one-off pull requests. In this case the caller of the function doesn't have any need to unwrap the error so I'm not understanding how wrapping adds value here.

If you think there is value in changing error handling practices in general, let's have that discussion in the [error handling issue](https://github.com/hyperledger/fabric/issues/4468) and come up with an overall plan for rolling the changes into the codebase. This should include discussion of how callers should be updated to take advantage of new proposed practices.

PR #4694, opened 2024-02-25
Excessive mutex locks were removed

fubss:
Signed-off-by fubss <ivanlaish@gmail.com>

# Removing excessive mutex locks increasing performance

#### Type of change

#### Description

As part of my research [Uncovering the Perfect Scalable Database for Hyperledger Fabric's Evolution](https://ieeexplore.ieee.org/document/10189331) it was discovered that the Hyperledger Fabric peer node does not use resources at 100% for some types of transactions. One of the reasons of it is the often mutexes locks in the source.

This change removes three excessive mutexes locks. Some of them were locking already atomic operations and some were locking nothing.

#### Additional details

The results of the peer node performance improvement were published in the paper provided above. The change has given better CPU and Disk usage and increased transaction per second rate.

Performance benchmarks of peer node were made with hyperledger-caliper. The results for write operation are:

  * for 100 byte transactions:

    * TPS rate increased to 5% 

    * CPU usage increased to 7%

    * Disk Write rate increased to 5%

  * for 4000 byte transactions:

    * TPS rate increased to 5% 

    * CPU usage increased to 6%

    * Disk Write rate increased to 4%

  * for 64000 byte transactions:

    * TPS rate increased to 2% 

    * CPU usage stayed the same

    * Disk Write rate increased to 1%

Full Caliper reports are availiable [here](https://drive.google.com/drive/folders/15eOiYOG7QeLv-qj5A5RNHWLNRwf2xduo?usp=sharing). For read operation no performance changes were detected.

The changes were sucessfully tested locally and with GitHub CI/CD.

### DIFF
--- a/core/ledger/kvledger/kv_ledger.go
+++ b/core/ledger/kvledger/kv_ledger.go
@@ -517,6 +517,8 @@ func (l *kvLedger) GetTransactionByID(txID string) (*peer.ProcessedTransaction, 
 
 // GetBlockchainInfo returns basic info about blockchain
 func (l *kvLedger) GetBlockchainInfo() (*common.BlockchainInfo, error) {
+	l.blockAPIsRWLock.RLock()
+	defer l.blockAPIsRWLock.RUnlock()
 	bcInfo, err := l.blockStore.GetBlockchainInfo()
 	return bcInfo, err
 }
@@ -543,6 +545,8 @@ func (l *kvLedger) GetBlocksIterator(startBlockNumber uint64) (commonledger.Resu
 
 // GetBlockByHash returns a block given it's hash
 func (l *kvLedger) GetBlockByHash(blockHash []byte) (*common.Block, error) {
+	l.blockAPIsRWLock.RLock()
+	defer l.blockAPIsRWLock.RUnlock()
 	block, err := l.blockStore.RetrieveBlockByHash(blockHash)
 	return block, err
 }
@@ -1020,6 +1024,8 @@ func (itr *blocksItr) Next() (commonledger.QueryResult, error) {
 	if err != nil {
 		return nil, err
 	}
+	itr.blockAPIsRWLock.RLock()
+	defer itr.blockAPIsRWLock.RUnlock()
 	return block, nil
 }
 
--- a/core/ledger/kvledger/kv_ledger.go
+++ b/core/ledger/kvledger/kv_ledger.go
@@ -517,8 +517,6 @@ func (l *kvLedger) GetTransactionByID(txID string) (*peer.ProcessedTransaction, 
 
 // GetBlockchainInfo returns basic info about blockchain
 func (l *kvLedger) GetBlockchainInfo() (*common.BlockchainInfo, error) {
-	l.blockAPIsRWLock.RLock()
-	defer l.blockAPIsRWLock.RUnlock()
 	bcInfo, err := l.blockStore.GetBlockchainInfo()
 	return bcInfo, err
 }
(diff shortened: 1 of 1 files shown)


yacovm:
https://github.com/hyperledger/fabric/commit/f17d1d934ca2f65740561f06277b90c7eee5fa89

fubss:
Removing such locks should not affect the state and blockchain integrity, because:


* for function `GetBlockchainInfo`:

  * there are no acquired locks in any other places of calling this method e.g.: [1](https://github.com/hyperledger/fabric/blob/95f28a3838d1a094a43315f0d5145f232d831df2/internal/ledgerutil/verify/verify.go#L75), [2](https://github.com/hyperledger/fabric/blob/e65d6bd4d5626e184e3d5e24cd203f827b93d22f/core/ledger/kvledger/kv_ledger.go#L293), [3](https://github.com/hyperledger/fabric/blob/e65d6bd4d5626e184e3d5e24cd203f827b93d22f/core/ledger/kvledger/kv_ledger.go#L367);

  * the [bcInfo](https://github.com/hyperledger/fabric/blob/f224d69ad123d732af2832466227462395a7c55c/common/ledger/blkstorage/blockfile_mgr.go#L490) value which is used under the hood is already [atomic value](https://github.com/hyperledger/fabric/blob/f224d69ad123d732af2832466227462395a7c55c/common/ledger/blkstorage/blockfile_mgr.go#L44);


* function `RetrieveBlockByHash`:

  * after my additional check the race condition is possible here; so I am going to retract [the change proposal for this place](https://github.com/hyperledger/fabric/blob/v3.0.0-preview/core/ledger/kvledger/kv_ledger.go#L548-L549);

* function `Next`:

  * in this method mutex lock was acquired for no purpose / no action;

  * the lock already exist in the [method Next implementation](https://github.com/hyperledger/fabric/blob/8a33e5f53e2f7ba6dd202d3d69926903542cda0a/common/ledger/blkstorage/blocks_itr.go#L66-L67)

yacovm:
@manish-sethi , @denyeart @cendhu  do you guys want to take a look at this?

manish-sethi:
Thanks @yacovm for giving the pointer to the relevant commit. 

@fubss - As mentioned in the commit message, this synchronization was added for maintaining an atomic view from a client perspective. By atomic I mean, committing block to block-storage and applying its effects on statedb in an atomic operation. The corresponding write lock is acquired [here](https://github.com/hyperledger/fabric/blob/b891ea794662b3186ee54ed9f868a967bd2abd64/core/ledger/kvledger/kv_ledger.go#L655C4-L655C19). 

If you remove this read lock, there is a window where a client can get a response from the API `GetBlockchainInfo()` returning the most recent in-process block which is committed to the block store, but corresponding state updates are still in progress. After having received this response, the client can query the state but still may find that his proposed state updates are not applied to the statedb. 

(In other words, this raises a question what does committing a block mean from clients' perspective?). I remember there was a significant Jira discussion on alternate designs and associated performance implications. However, not sure if the archive of Jira discussions is available.

fubss:
Thank you for the detailed answer!
Now it is clear to me why it works this way. Then the `defer` is necessary in the line https://github.com/hyperledger/fabric/blob/d0bad2e9c0ad1359a49a130c5ceb98d16bd8ed52/core/ledger/kvledger/kv_ledger.go#L1028

denyeart:
Thank you for the update with the 2nd commit. However, could I ask you to open a separate pull request? The original pull request description no longer applies. And having the original commit and the reversion merged would complicate the commit history. It will be cleaner to have a new pull request with a single commit for the additional defer statement.

manish-sethi:
> Thank you for the detailed answer! Now it is clear to me why it works this way. Then the `defer` is necessary in the line
> 
> https://github.com/hyperledger/fabric/blob/d0bad2e9c0ad1359a49a130c5ceb98d16bd8ed52/core/ledger/kvledger/kv_ledger.go#L1028

No, `defer` is not "necessary" here! This is the last line in this tiny (two-line) function. `defer` has associated cost, so it's better not to have it. If you still want to have it, as Dave suggested, you can do it in a separate PR. 

However, I believe rather a more useful PR would be to capture the discussion in this PR in the code comment so the purpose is clear upfront to a reader.

denyeart:
Comments added in #4827

PR #4695, opened 2024-02-25
BFT: A BFT synchronizer for smartbft

tock-ibm:
#### Type of change
- New feature

#### Description

A BFT synchronizer for smartbft

#### Related issues

Issue #4566


orderer/consensus/smartbft/synchronizer_bft.go
```diff
@@ -0,0 +1,239 @@
...
+	"time"
+
+	"github.com/SmartBFT-Go/consensus/pkg/types"
+	"github.com/SmartBFT-Go/consensus/smartbftprotos"
+	"github.com/hyperledger/fabric-lib-go/bccsp"
+	"github.com/hyperledger/fabric-lib-go/common/flogging"
+	"github.com/hyperledger/fabric-protos-go/common"
+	"github.com/hyperledger/fabric/common/deliverclient"
+	"github.com/hyperledger/fabric/internal/pkg/peer/blocksprovider"
+	"github.com/hyperledger/fabric/internal/pkg/peer/orderers"
```
yacovm, line 20: didn't we say we will move this package somewhere else?

orderer/consensus/smartbft/sync_buffer.go
```diff
@@ -0,0 +1,70 @@
...
+import (
+	"sync"
+
+	"github.com/hyperledger/fabric-protos-go/common"
+	"github.com/pkg/errors"
+)
+
+type SyncBuffer struct {
+	blockCh  chan *common.Block
+	stopCh   chan struct{}
```
yacovm, line 18: why do you need to stop the buffer? you don't have a goroutine here.

orderer/consensus/smartbft/synchronizer_bft.go
```diff
@@ -0,0 +1,239 @@
...
+			CurrentNodes:          s.lastReconfig.CurrentNodes,
+		},
+	}
+}
+
+func (s *BFTSynchronizer) synchronize() (*types.Decision, error) {
+	//=== We use the BlockPuller to probe all the endpoints and establish a target height, as well as detect
+	// the self endpoint.
+
+	// In BFT it is highly recommended that the channel/orderer-endpoints (for delivery & broadcast) map 1:1 to the
```
yacovm: while that is true, i don't see how it helps us having this comment written here, since the people that develop Fabric don't use it :-) 

Maybe instead move it to `configtx.yaml` ?

orderer/consensus/smartbft/synchronizer_bft.go
```diff
@@ -0,0 +1,239 @@
...
+	heightByEndpoint, myEndpoint, err := blockPuller.HeightsByEndpoints()
+	if err != nil {
+		return nil, errors.Wrap(err, "cannot get HeightsByEndpoints")
+	}
+
+	s.Logger.Infof("HeightsByEndpoints: %+v, my endpoint: %s", heightByEndpoint, myEndpoint)
+
+	var heights []uint64
+	for ep, value := range heightByEndpoint {
+		if ep == myEndpoint {
```
yacovm: instead of doing that, can't you just delete myEndpoint from heightByEndpoint prior to the iteration? It reads better, no?

orderer/consensus/smartbft/synchronizer_bft.go
```diff
@@ -0,0 +1,239 @@
...
+		Latest: *decision,
+		Reconfig: types.ReconfigSync{
+			InReplicatedDecisions: s.lastReconfig.InLatestDecision,
+			CurrentConfig:         s.lastReconfig.CurrentConfig,
+			CurrentNodes:          s.lastReconfig.CurrentNodes,
+		},
+	}
+}
+
+func (s *BFTSynchronizer) synchronize() (*types.Decision, error) {
```
yacovm, line 87: can we make this function smaller? perhaps extract everything from beginning to right before we start the actual BFT delivery into a function?

orderer/consensus/smartbft/synchronizer_bft.go
```diff
@@ -0,0 +1,239 @@
...
+	}
+
+	startSeq := startHeight
+	s.Logger.Infof("Finished synchronizing with cluster, fetched %d blocks, starting from block [%d], up until and including block [%d]",
+		blocksFetched, startSeq, lastPulledBlock.Header.Number)
+
+	viewMetadata, lastConfigSqn := s.getViewMetadataLastConfigSqnFromBlock(lastPulledBlock)
+
+	s.Logger.Infof("Returning view metadata of %v, lastConfigSeq %d", viewMetadata, lastConfigSqn)
+	return s.BlockToDecision(lastPulledBlock), nil
```
yacovm: why is `lastConfigSqn` interesting here?

orderer/consensus/smartbft/synchronizer_bft.go
tock-ibm: Yes. After I'll finish this one. It will then be easy to move everything in one go.

orderer/consensus/smartbft/sync_buffer.go
tock-ibm: To release external go routines that block on PullBlock or HandleBlock.

orderer/consensus/smartbft/synchronizer_bft.go
tock-ibm: Good suggestion.

orderer/consensus/smartbft/synchronizer_bft.go
tock-ibm: It is not...

orderer/consensus/smartbft/synchronizer_bft.go
tock-ibm: Good suggestion.

orderer/consensus/smartbft/sync_buffer.go
pfi79: Can you please tell me why you don't use `context` for such purposes? Like ```ctx, cancel := context.WithCancel(context.Background())```

yacovm: I would guess he wants to keep the same method signature

orderer/consensus/smartbft/synchronizer_bft.go
tock-ibm: Done

orderer/consensus/smartbft/sync_buffer.go
tock-ibm: > Can you please tell me why you don't use `context` for such purposes? Like `ctx, cancel := context.WithCancel(context.Background())`

It can be done with context as well, `Stop()` will simply call the `CancelFunc`.
I do want to keep the `Stop()`  function in the API.

orderer/consensus/smartbft/sync_buffer.go
```diff
@@ -0,0 +1,70 @@
...
+	stopOnce sync.Once
+}
+
+func NewSyncBuffer() *SyncBuffer {
+	return &SyncBuffer{
+		blockCh: make(chan *common.Block, 10),
```
yacovm, line 24: can we increase this buffer to 100 by default? and ideally this will be configurable. can we open an issue to make this configurable?

yacovm (approved):
merging but let's address the comment about the buffer size in another PR and issue.

orderer/consensus/smartbft/sync_buffer.go
tock-ibm: sure
https://github.com/hyperledger/fabric/issues/4715

PR #4720, opened 2024-03-05
BFT header receiver verifies config blocks in full

tock-ibm:
#### Type of change

#### Description

Once a header receiver discovers it has a config block rather then a block attestation, it will verify the block in full, i.e. including the data hash.

### DIFF
--- a/internal/pkg/peer/blocksprovider/bft_header_receiver.go
+++ b/internal/pkg/peer/blocksprovider/bft_header_receiver.go
@@ -21,8 +21,8 @@ import (
 // It keeps track of the last header it received, and the time it was received.
 // The header receivers verify each block as it arrives.
 //
-// TODO The header receiver will receive (or ask for) full config blocks - in a later commit.
-// TODO The header receiver will maintain its own private block verifier (bundle) - in a later commit.
+// The header receiver receives full config blocks.
+// The header receiver maintains its own private block verifier that gets updated on every config block.
 type BFTHeaderReceiver struct {
 	mutex                  sync.Mutex
 	chainID                string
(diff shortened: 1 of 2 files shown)


internal/pkg/peer/blocksprovider/bft_header_receiver.go
```diff
@@ -110,15 +110,21 @@ func (hr *BFTHeaderReceiver) DeliverHeaders() {
...
 		case *orderer.DeliverResponse_Block:
 			blockNum := t.Block.Header.Number
-			err := hr.updatableBlockVerifier.VerifyBlockAttestation(t.Block)
-			if err != nil {
-				hr.logger.Warningf("[%s][%s] Last block verification failed, blockNum [%d], err: %s", hr.chainID, hr.endpoint, blockNum, err)
-				return
-			}
 
-			if protoutil.IsConfigBlock(t.Block) { // blocks with block.Data==nil return false
+			if !protoutil.IsConfigBlock(t.Block) { // normal blocks with block.Data==nil
```
yacovm, line 114: why use `! IsConfigBlock` and not just check the data? It's cheaper to do the latter, no?

tock-ibm: `IsConfigBlock` will react fast (with false) for if data==nil, so it is as cheap as testing the data directly... We optimized for that in a previous PR.

Note that we must call IsConfigBlock in any case before we try to update the config. 
I use `if else` here in order to avoid invoking `VerifyBlockAttestation` and `VerifyBlock` when a config block does arrive.

yacovm (approved):

PR #4728, opened 2024-03-07
change flag for ginkgo exec

pfi79:
--slow-spec-threshold is deprecated --slow-spec-threshold has been deprecated and will be removed in a future version of Ginkgo.  This feature has proved to be more noisy than useful.  You can use --poll-progress-after, instead, to get more actionable feedback about potentially slow specs and understand where they might be getting stuck.


ale-linux (approved):

denyeart:

PR #4746, opened 2024-03-11
BFT chain unit tests: sync a node

MayRosenbaum:
#### Type of change

- Test update

#### Description

This PR aims to test whether a node can sync successfully when it's behind. 

#### Additional details

For the implementation, a synchronizer factory was needed.

#### Related issues

#4008


orderer/consensus/smartbft/consenter.go
```diff
@@ -216,6 +217,65 @@ func (c *Consenter) HandleChain(support consensus.ConsenterSupport, metadata *cb
 		}
 	}
 
+	synchronizerFactory := func(
```
tock-ibm: This too big to just drop it in here as a function. Please use the style that uses an interface, not a function.
place the implementation called `synchronizerCreator`, in a separate file `synchronizer_factory.go`, with the definition of the interface and respective mocks. For example, use the pattern here:

`orderer/consensus/smartbft/block_puller_factory.go`

orderer/consensus/smartbft/consenter.go
```diff
@@ -216,6 +217,65 @@ func (c *Consenter) HandleChain(support consensus.ConsenterSupport, metadata *cb
...
+		logger *flogging.FabricLogger,
+		localConfigCluster localconfig.Cluster,
+		rtc RuntimeConfig,
+		blockToDecision func(block *cb.Block) *types.Decision,
+		pruneCommittedRequests func(block *cb.Block),
+		updateRuntimeConfig func(block *cb.Block) types.Reconfig,
+		support consensus.ConsenterSupport,
+		bccsp bccsp.BCCSP,
+		clusterDialer *cluster.PredicateDialer,
+		clusterSize uint64,
```
tock-ibm: cluster size is no longer needed.

orderer/consensus/smartbft/interfaces.go
```diff
@@ -39,3 +40,8 @@ type policyManager interface {
 type policy interface {
 	policies.Policy
 }
+
+//go:generate mockery --dir . --name synchronizer --case underscore --with-expecter=true --exported --output mocks
+type synchronizer interface {
+	api.Synchronizer
+}
```
tock-ibm: see comment above

orderer/consensus/smartbft/util_network_test.go
```diff
@@ -92,6 +97,7 @@ func (ns *NetworkSetupInfo) SendTxToAllAvailableNodes(tx *cb.Envelope) error {
 			ns.t.Logf("Error occurred during sending tx to node %v: %v", idx, err)
 		}
 		errorsArr = append(errorsArr, err)
+		ns.t.Logf("Tx to node %v was sent successfully", idx)
```
tock-ibm: there might have been an error here. 
you also append nil errors.
why not do - if there is an error, append it and continue?

orderer/consensus/smartbft/util_network_test.go
```diff
@@ -186,6 +192,8 @@ func (n *Node) Restart() error {
 	}
 	n.Chain = newChain
 	n.Chain.Start()
+	n.IsConnectedToNetwork = true
+	n.IsStarted = true
```
tock-ibm, line 193: why are you calling `n.Chain.Halt()` when you restart?

orderer/consensus/smartbft/util_network_test.go
```diff
@@ -233,14 +242,14 @@ func (n *Node) sendRequest(sender uint64, target uint64, request []byte) error {
 }
 
 func (n *Node) receiveMessage(sender uint64, message *smartbftprotos.Message) {
-	n.lock.Lock()
-	defer n.lock.Unlock()
+	n.lock.RLock()
+	defer n.lock.RUnlock()
```
tock-ibm, line 243: why was this changed from Lock() to RLock()?

I am not sure it is related but:
It seems to me that the main problem here is that the routine that does `send` also invokes the `receive` and the `Handle`. In a real network that does not happen, there is a go-routine and the receiving end that reads incoming messages and invokes the "Handle".

orderer/consensus/smartbft/util_network_test.go
```diff
@@ -404,14 +427,64 @@ func createBFTChainUsingMocks(t *testing.T, node *Node) (*smartbft.BFTChain, err
...
+	synchronizerMock := smartBFTMocks.NewSynchronizer(t)
+	synchronizerMock.EXPECT().Sync().RunAndReturn(
+		func() types.SyncResponse {
+			t.Logf("Sync Called by node %v", node.NodeId)
+			// iterate over the ledger of the other nodes and find the highest ledger to sync with
+			max := 0
+			var ledgerToCopy []*cb.Block
+			for _, n := range node.nodesMap {
+				if n.NodeId != node.NodeId {
+					ledgerToCopy = n.State.GetLedgerArray()
```
tock-ibm: after the loop ends, the ledger to copy is not always the maximal one

orderer/consensus/smartbft/util_network_test.go
```diff
@@ -404,14 +427,64 @@ func createBFTChainUsingMocks(t *testing.T, node *Node) (*smartbft.BFTChain, err
...
+					len := len(ledgerToCopy)
```
tock-ibm: you have a function `GetLedgerHeight()`

orderer/consensus/smartbft/util_network_test.go
```diff
@@ -404,14 +427,64 @@ func createBFTChainUsingMocks(t *testing.T, node *Node) (*smartbft.BFTChain, err
...
+			}
+
+			// send response
+			// at this point the chain exists, so we can use its methods
+			return types.SyncResponse{
+				Latest: *node.Chain.BlockToDecision(ledgerToCopy[len(ledgerToCopy)-1]),
+				Reconfig: types.ReconfigSync{
+					InReplicatedDecisions: false,
+					CurrentNodes:          []uint64{1, 2, 3, 4},
+					CurrentConfig:         types.Configuration{SelfID: 1},
```
tock-ibm: it should be the ID of the node calling sync

orderer/consensus/smartbft/util_network_test.go
```diff
@@ -404,14 +427,64 @@ func createBFTChainUsingMocks(t *testing.T, node *Node) (*smartbft.BFTChain, err
...
+	localConfigCluster := localconfig.Cluster{ReplicationPolicy: "consensus"}
+	clusterDialer := &cluster.PredicateDialer{}
 
 	bftChain, err := smartbft.NewChain(
 		configValidatorMock,
 		nodeId,
 		config,
 		node.WorkingDir,
-		nil,
+		clusterDialer,
```
tock-ibm, line 478: not needed anymore, remove it from `smartbft.NewChain` as well

orderer/consensus/smartbft/util_network_test.go
```diff
@@ -404,14 +427,64 @@ func createBFTChainUsingMocks(t *testing.T, node *Node) (*smartbft.BFTChain, err
...
+				node.State.AddBlock(clonedBlock)
+			}
+
+			// send response
+			// at this point the chain exists, so we can use its methods
+			return types.SyncResponse{
+				Latest: *node.Chain.BlockToDecision(ledgerToCopy[len(ledgerToCopy)-1]),
+				Reconfig: types.ReconfigSync{
+					InReplicatedDecisions: false,
+					CurrentNodes:          []uint64{1, 2, 3, 4},
```
tock-ibm: this should be derived from the network definition no?

orderer/consensus/smartbft/util_network_test.go
MayRosenbaum: It seemed to me that this was necessary, but in fact it is not required.

orderer/consensus/smartbft/util_network_test.go
MayRosenbaum: Right, I fixed it

orderer/consensus/smartbft/util_network_test.go
MayRosenbaum: This is true. I already changed it in the next PR, but I will fix it in the current PR to be accurate.

orderer/consensus/smartbft/util_network_test.go
tock-ibm: Ignore

orderer/consensus/smartbft/consenter.go
MayRosenbaum: Done.

orderer/consensus/smartbft/synchronizer_factory.go
```diff
@@ -0,0 +1,114 @@
...
+	localConfigCluster localconfig.Cluster,
+	rtc RuntimeConfig,
+	blockToDecision func(block *cb.Block) *types.Decision,
+	pruneCommittedRequests func(block *cb.Block),
+	updateRuntimeConfig func(block *cb.Block) types.Reconfig,
+	support consensus.ConsenterSupport,
+	bccsp bccsp.BCCSP,
+	clusterDialer *cluster.PredicateDialer,
+) api.Synchronizer {
+	return newSynchronizer(logger, localConfigCluster, rtc, blockToDecision, pruneCommittedRequests, updateRuntimeConfig, support, bccsp, clusterDialer)
```
tock-ibm, line 54: We don't need another function that is only called once. Just implement it in the method here.

tock-ibm (approved):
A minor comment that we can take later.

PR #4747, opened 2024-03-12
Simplify code using direct check

yeasy:
Change-Id: Ic8f17b24373a1cb4d43f7b7344d8f987f00b48d6

#### Type of change

#### Description

The patchset follows Golang style to check method return directly, 
which can simplify the code too.

#### Additional details

N/A

#### Related issues

N/A

#### Release Note

N/A

### DIFF
--- a/common/configtx/update.go
+++ b/common/configtx/update.go
@@ -32,9 +32,8 @@ func (vi *ValidatorImpl) verifyReadSet(readSet map[string]comparable) error {
 func computeDeltaSet(readSet, writeSet map[string]comparable) map[string]comparable {
 	result := make(map[string]comparable)
 	for key, value := range writeSet {
-		readVal, ok := readSet[key]
 
-		if ok && readVal.version() == value.version() {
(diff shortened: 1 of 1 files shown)


common/configtx/update.go
```diff
@@ -32,9 +32,8 @@ func (vi *ValidatorImpl) verifyReadSet(readSet map[string]comparable) error {
 func computeDeltaSet(readSet, writeSet map[string]comparable) map[string]comparable {
 	result := make(map[string]comparable)
 	for key, value := range writeSet {
-		readVal, ok := readSet[key]
 
-		if ok && readVal.version() == value.version() {
+		if readVal, ok := readSet[key]; ok && readVal.version() == value.version() {
```
denyeart, line 36: Although it is less lines of code, my personal opinion is that the change makes it less readable. Regardless of whether it is a little better or a little worse, what is the motivation for making what is essentially a no-op change? If there is value then it's fine, I'm just concerned with polluting the commit history and diverging branches if there is not a clear benefit.

PR #4748, opened 2024-03-13
Add release notes for v3.0.0-beta

denyeart:
Add release notes for v3.0.0-beta.

### DIFF
--- a/release_notes/v3.0.0-beta.md
+++ b/release_notes/v3.0.0-beta.md
@@ -1,0 +1,122 @@
+v3.0.0-beta Release Notes - March 14, 2024
+==========================================
+
+The v3.0.0-beta release is an early release of Fabric v3.0, specifically to demonstrate and get feedback on the new Byzantine Fault Tolerant (BFT) ordering service.
+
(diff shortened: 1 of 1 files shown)


denyeart:
@yacovm @tock-ibm ,

Anything else you want mentioned in SmartBFT section of release notes, or in [What's New](https://hyperledger-fabric.readthedocs.io/en/latest/whatsnew.html#byzantine-fault-tolerant-bft-ordering-service), or in the announcement email?

yacovm:
maybe write that the delta from the preview version is the byzantine block pulling?

denyeart:
> maybe write that the delta from the preview version is the byzantine block pulling?

Agreed, go ahead and suggest exact text or make a 2nd commit for this and other details, because your wording will certainly be more accurate than mine!

yacovm:
I tried pushing the commit to your repo but I got access denied, so I added my signed off here: https://github.com/hyperledger/fabric/pull/4751

denyeart:
Ok, the DCO check is sensitive... we have ability to force pass it after confirming the commit is indeed signed, but I'll go ahead and close and we can use your PR.

PR #4759, opened 2024-03-20
add a suffix to the name of module v3

pfi79:
Wanted to bring the name of the module to the requirements [go](https://go.dev/ref/mod#major-version-suffixes)

It is not possible to break pr into parts.

It seems to me that such a change would help the community a lot


denyeart:
We talked about this for v2, but decided not to do it since the [Go docs](https://go.dev/doc/modules/release-workflow) talk about module versioning from the perspective of other developers importing your module. Fabric is a repository for the runtime binaries and doesn't have public APIs that other projects should import. The APIs for other projects to import should be in fabric-protos, or common code in fabric-lib-go. We thought that versioning the module would signal that the module could be imported and used by other Go modules, which is not intended or something that we want.

This being said, I am open to the discussion and other thoughts...

pfi79:
I know community members try to reuse fabric code all the time. As a rule, they end up copying whole packages into their code (very bad).

Personally, I have a task now to write integration tests for products for which fabric is the alpha and omega. I would like to reuse the ready code base of fabric integration tests. I wouldn't have to worry about how to get the fabric up and running. It seems impossible to put the integration test framework in a separate repository, for example because of the peer and orderer startup.

Having a suffix for fabric would simplify this and other tasks that constantly arise for those who use fabric. And of course they will do it, knowing the risks involved.

pfi79:
> ... the [Go docs](https://go.dev/doc/modules/release-workflow) talk about module versioning from the perspective of other developers importing your module. Fabric is a repository for the runtime binaries and ......

I reread the documentation, and I didn't find that executable modules should be versioned differently. Everywhere in the documentation the description is given for all modules.

@denyeart , could you point out the place in the golang documentation where it would be written that in certain cases, modules v2 and older may not have the suffix added to the name.

denyeart:
> > ... the [Go docs](https://go.dev/doc/modules/release-workflow) talk about module versioning from the perspective of other developers importing your module. Fabric is a repository for the runtime binaries and ......
> 
> I reread the documentation, and I didn't find that executable modules should be versioned differently. Everywhere in the documentation the description is given for all modules.
> 
> @denyeart , could you point out the place in the golang documentation where it would be written that in certain cases, modules v2 and older may not have the suffix added to the name.

I didn't say that the docs addressed executable modules one way or the other. I only said that the docs always talk about module versioning from the perspective of clients importing public APIs. Fabric doesn't have public APIs in the fabric repository. The project generally follows semantic versioning from the perspective of executable users. If people are importing and using internal APIs, it is unclear to me whether applying the executable semantic version tags to the internal APIs is beneficial. The internal APIs will always be unstable relative to the executable semantic versioning tags, and can therefore be thought of as v0.x (unversioned). Therefore it may be better for projects using internal APIs to pick specific commits to depend on rather than read anything into the Fabric versioning scheme.

pfi79:
> Therefore it may be better for projects using internal APIs to pick specific commits to depend on rather than read anything into the Fabric versioning scheme.

You're probably right. I just tried the command:
`go get -v github.com/hyperledger/fabric@08cd8a667128dc077` - last commit

my anticipation:
`github.com/hyperledger/fabric v3.0.0-beta1.0.20240319070946-08cd8a667128` - similar to this

got this:
`github.com/hyperledger/fabric v1.4.0-rc1.0.20240319070946-08cd8a667128`

weird, why is that? I don't understand.

pfi79:
@denyeart Thank you. I've found the solution to my problem. pr closed. But the behaviour go strange )).

denyeart:
I think v1.4.0-rc1 was the last tag before go modules, or something like that. v0.0.0-20240319070946-08cd8a667128 might also work?

PR #4770, opened 2024-03-28
Update channel_update_tutorial.rst to follow the latest fabric-samples

satota2:
#### Type of change

- Bug fix
- Documentation update

#### Description

Minor updates on  channel_update_tutorial.rst to follow the latest fabric-samples.

### DIFF
--- a/docs/source/channel_update_tutorial.rst
+++ b/docs/source/channel_update_tutorial.rst
@@ -579,7 +579,7 @@ The first step is to package the Basic chaincode:
 
 .. code:: bash
 
-    peer lifecycle chaincode package basic.tar.gz --path ../asset-transfer-basic/chaincode-go/ --lang golang --label basic_1
+    peer lifecycle chaincode package basic.tar.gz --path ../asset-transfer-basic/chaincode-go/ --lang golang --label basic_1.0
 
 This command will create a chaincode package named ``basic.tar.gz``, which we can
 install on the Org3 peer. Modify the command accordingly if the channel is running a
@@ -606,7 +606,7 @@ You should see output similar to the following:
 .. code:: bash
 
       Get installed chaincodes on peer:
-      Package ID: basic_1:5443b5b557efd3faece8723883d28d6f7026c0bf12245de109b89c5c4fe64887, Label: basic_1
+      Package ID: basic_1.0:c6a45e2d5563c883869149c3dbd941c22fbe27daa21f0552834f5a53fbb8058a, Label: basic_1.0
 
 We are going to need the package ID in a future command, so lets go ahead and
 save it as an environment variable. Paste the package ID returned by the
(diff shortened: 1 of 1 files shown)


yeasy (approved):
Simply update the version number from 1 to 1.0.1

satota2:
@hyperledger/fabric-core-maintainers
It's a very minor fix, but it's important to keep the documentation up to date.
I would appreciate it if someone could take the time to review and merge it.

docs/source/channel_update_tutorial.rst
```diff
@@ -579,7 +579,7 @@ The first step is to package the Basic chaincode:
 
 .. code:: bash
 
-    peer lifecycle chaincode package basic.tar.gz --path ../asset-transfer-basic/chaincode-go/ --lang golang --label basic_1
+    peer lifecycle chaincode package basic.tar.gz --path ../asset-transfer-basic/chaincode-go/ --lang golang --label basic_1.0.1
```
denyeart: Apologies for not reviewing sooner, I think I did not understand the intent of why use v1.0.1 instead of v1. Could you explain? Does test-network default to v1.0.1?

satota2: @denyeart

>Apologies for not reviewing sooner, I think I did not understand the intent of why use v1.0.1 instead of v1. Could you explain? Does test-network default to v1.0.1?

No problem at all, please don’t worry about it. Thank you for your review.

The latest test-network defaults to using version 1.0.1, as indicated in [this code](https://github.com/hyperledger/fabric-samples/blob/3826626d0049258abe256379be6a69bf4c687fee/test-network/network.config#L29C13-L29C18). My intention was to align with this update.

While addressing your comment, I noticed a discrepancy regarding the default chaincode version in the latest test-network (e.g., 1.0.1, [1.0.0](https://github.com/hyperledger/fabric-samples/blob/3826626d0049258abe256379be6a69bf4c687fee/test-network/scripts/packageCC.sh#L23), [1.0](https://github.com/hyperledger/fabric-samples/blob/3826626d0049258abe256379be6a69bf4c687fee/test-network/scripts/utils.sh#L75)), including an inconsistency in the command message description.

Would it be more appropriate to resolve this discrepancy by unifying the default value to '1' in the fabric-samples repository, rather than addressing it through this PR? I would appreciate your thoughts on the best approach.

denyeart: I think the value of `CC_VERSION="1.0.1"` was a mistake. Prior to this [commit](https://github.com/hyperledger/fabric-samples/commit/e73bb717db47185a8ccd1b701503b369cd2a73e1) it was `"1.0"`

And in the lifecycle ReadTheDoc tutorials it is "1.0".

I think "1.0" was chosen for samples to differentiate it from the sequence number which is an integer like "1".

I'd suggest to align the samples default and the tutorials to be "1.0".

Thank you for looking into these details!

satota2: @denyeart 
Thank you for your response.

I agree with your suggestion. To align with the tutorials, I will create and submit a patch to fix the version to “1.0” in the fabric-samples repository. Once that's done, I will update this PR accordingly.

Thank you again for your guidance!

satota2: @denyeart 
I updated the PR to align to "1.0".
Also, I submitted the following PR in fabric-samples. 
https://github.com/hyperledger/fabric-samples/pull/1245

Please confirm them.

denyeart (approved):
Thanks!

Issue #4731, opened 2024-03-07
BFT synchronizer: integration test against an orderer that does block censorship


tock-ibm:
- Run a 4 orderer smart bft cluster.
- Do 5 TXs.
- Shut down o4.
- Do 5 more TXs.
- Shutdown o1,o2,o3
- Prepare 3 orderer mocks with 10 TXs, with the addresses of o1,o2,o3
   - same mocks as used in the peer test
   -  emulate block censorship
- Start o4
- O4 should detect the block censorship attempt and get all 10 blocks to its ledger

tock-ibm:
Done

PR #4771, opened 2024-03-28
BFT synchronizer: integration test against an orderer that does block censorship

arkadiPiven:
TEST update

* Run a 4 orderer smart bft cluster.
* Do 10 TXs.
* Shut down o4.
* Do 10 more TXs.
* Shutdown o1,o2,o3
* Prepares 3 orderer mocks with 20 TXs, with the addresses of o1,o2,o3
* Emulates block censorship
* Start o4
* O4 should detects the block censorship attempt and get all 20 blocks to its ledger

#### Related issues
BFT synchronizer: integration test against an orderer that does block censorship #4731


tock-ibm:
finished in #4881

Issue #4763, opened 2024-03-24
Remove global level endpoints from the channel config


MayRosenbaum:
### Current Status

The channel config specifies both [global-level endpoints](https://github.com/hyperledger/fabric/blob/main/common/channelconfig/util.go#L101-L105) and [per-organization endpoints](https://github.com/hyperledger/fabric/blob/main/integration/nwo/fabricconfig/configtx.go#L22-L29). 

### Expected

In V3.0 we should use the per-organization endpoints and eliminate the option to specify global-level endpoints.
This means that the configtx validation unit will check if any global-level endpoints have been configured and refuse and return a suitable error if so.

### Please let us know if you plan to work on this.

Yes!

tock-ibm:
Finished with #4879

PR #4778, opened 2024-04-02
Remove global-level endpoints from channel config - WIP

MayRosenbaum:
#### Type of change

- config update

#### Description

This PR aims to block the option to specify global-level endpoints and enforce that endpoints per org are defined. 

#### Related issues

 issue #4763

### DIFF
--- a/integration/raft/migration_test.go
+++ b/integration/raft/migration_test.go
@@ -9,6 +9,7 @@ package raft
 import (
 	"errors"
 	"fmt"
+	"github.com/hyperledger/fabric/internal/configtxlator/update"
 	"os"
 	"path/filepath"
 	"strings"
@@ -99,11 +100,6 @@ var _ = Describe("ConsensusTypeMigration", func() {
 				o3Runner = network.OrdererRunner(o3)
 				o4Runner = network.OrdererRunner(o4)
 
-				o1Runner.Command.Env = append(o1Runner.Command.Env, "FABRIC_LOGGING_SPEC=orderer.common.cluster=debug:orderer.consensus.smartbft=debug:policies.ImplicitOrderer=debug")
-				o2Runner.Command.Env = append(o2Runner.Command.Env, "FABRIC_LOGGING_SPEC=orderer.common.cluster=debug:orderer.consensus.smartbft=debug:policies.ImplicitOrderer=debug")
-				o3Runner.Command.Env = append(o3Runner.Command.Env, "FABRIC_LOGGING_SPEC=orderer.common.cluster=debug:orderer.consensus.smartbft=debug:policies.ImplicitOrderer=debug")
-				o4Runner.Command.Env = append(o4Runner.Command.Env, "FABRIC_LOGGING_SPEC=orderer.common.cluster=debug:orderer.consensus.smartbft=debug:policies.ImplicitOrderer=debug")
-
 				o1Proc = ifrit.Invoke(o1Runner)
 				o2Proc = ifrit.Invoke(o2Runner)
 				o3Proc = ifrit.Invoke(o3Runner)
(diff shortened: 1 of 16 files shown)


integration/raft/migration_test.go
```diff
@@ -15,6 +15,8 @@ import (
 	"syscall"
 	"time"
 
+	"github.com/hyperledger/fabric/common/channelconfig"
+
```
pfi79, line 20: please delete the empty line and run the command 'goimports -l -w'

MayRosenbaum: Done.
I'm still working on this PR, the title is updated to be W.I.P.

internal/configtxgen/encoder/encoder.go
```diff
@@ -329,6 +329,8 @@ func NewOrdererOrgGroup(conf *genesisconfig.Organization) (*cb.ConfigGroup, erro
 
 	if len(conf.OrdererEndpoints) > 0 {
 		addValue(ordererOrgGroup, channelconfig.EndpointsValue(conf.OrdererEndpoints), channelconfig.AdminsPolicyKey)
+	} else {
+		return nil, errors.Errorf("orderer endpoints for organization %s are missing and must be cofigured", conf.Name)
```
denyeart, line 333: Do we really want to make OrdererEndpoints required? For example, could an OrdererOrg be added to the channel initially without OrdererEndpoints (without any orderer nodes), with the intent for that organization to later add their own nodes.

yacovm: Dave, this is `encoder.go`. It's only used when creating genesis blocks.

yacovm: When we add to a new channel we don't use that, but we download the existing config.

internal/configtxgen/encoder/encoder.go
```diff
@@ -142,7 +142,7 @@ func NewChannelGroup(conf *genesisconfig.Profile) (*cb.ConfigGroup, error) {
 	addValue(channelGroup, channelconfig.HashingAlgorithmValue(), channelconfig.AdminsPolicyKey)
 	addValue(channelGroup, channelconfig.BlockDataHashingStructureValue(), channelconfig.AdminsPolicyKey)
 	if conf.Orderer != nil && len(conf.Orderer.Addresses) > 0 {
-		addValue(channelGroup, channelconfig.OrdererAddressesValue(conf.Orderer.Addresses), ordererAdminsPolicyName)
+		return nil, errors.New("error adding global level endpoints to channel group, in V3.0 global level endpoints are not supported")
```
yacovm, line 145: this should not be here but outside the `if`

orderer/common/multichannel/chainsupport.go
```diff
@@ -161,6 +161,9 @@ func (cs *ChainSupport) ProposeConfigUpdate(configtx *cb.Envelope) (*cb.ConfigEn
 	}
 
 	if err = cs.ValidateConsensusMetadata(oldOrdererConfig, newOrdererConfig, false); err != nil {
+		if env.Config.ChannelGroup.Values["OrdererAddressesKey"] != nil && env.Config.ChannelGroup.Values["OrdererAddressesKey"].Value == nil {
+			return nil, errors.WithMessage(err, "consensus metadata update for channel config update is invalid since it includes global level endpoints which are not supported in V3.0")
```
yacovm: `errors.WithMessage` returns `nil` if the `err` is nil. It wraps an error if it's **not** nil. 
You need to use `errors.Errof` or something similar... otherwise we return here `nil, nil`

yacovm:
@MayRosenbaum which one is the relevant one? This one or https://github.com/hyperledger/fabric/pull/4800 ? Please close the other one.

Issue #4780, opened 2024-04-02
Add CouchDB index support to implicit private data collections


denyeart:
### Current Status

Defining a CouchDB index in a chaincode for an implicit private data collection such as:
`META-INF/statedb/couchdb/collections/_implicit_org_Org1MSP/indexes`

results in error:
```
[privacyenabledstate] HandleChaincodeDeploy -> ERRO 4e9 Error processing index for chaincode [marblesp]: cannot create an index for an undefined collection=[_implicit_org_Org1MSP]
```

### Expected

This story has two variants:
- In the chaincode definition, define a CouchDB index for a specific named organization's implicit private data collection
- In the chaincode definition, define a CouchDB index for all organization's implicit private data collections

 
Acceptance
- Index created for named organization upon chaincode deployment.
- Index created for each organization upon creation of their implicit private data collection state database (occurs upon first write to an org's collection).

solo-daemon:
@bkiran6398 are you working on it?

bkiran6398:
> @bkiran6398 are you working on it?

Hello @solo-daemon , Yes I am working on this feature.

PR #4794, opened 2024-04-04
CouchDB index support for implicit collections

bkiran6398:
Allow CouchDB index creation on implicit collection of organisations.

Indexes can be created for specific organisation's implicit collections at:
`META-INF/statedb/couchdb/collections/_implicit_org_<MSP_ID>` directory of the chaincode package.

Common indexes for all the organisations' implicit collection can be created at:
`META-INF/statedb/couchdb/collections/_implicit_org_*` directory of the chaincode package.
( `*` acts as wildcard for all organisation MSP ids. )

#### Type of change

- New feature

#### Description
Updated chaincode deployment handler to check if the indexes are for implicit collection if given collection directory name not explicitly defined in collection config. If the implicit collection indexes belong this handling peer's organisation or passed as global implicit collection indexes (using `*` notation), then indexes are created on corresponding implicit collection.

#### Additional details
- Please review the way of fetching `localMspId` directly using `viper` at `NewDB` constructor function.
- Added few more tar file entries to represent implicit collection indexes in unit tests. (However UT `TestHandleChainCodeDeployOnCouchDB` was already failing due to some other error)

#### Related issues
- #4780

### DIFF
--- a/core/ledger/kvledger/txmgmt/privacyenabledstate/db.go
+++ b/core/ledger/kvledger/txmgmt/privacyenabledstate/db.go
@@ -13,6 +13,7 @@ import (
 	"github.com/hyperledger/fabric-lib-go/common/flogging"
 	"github.com/hyperledger/fabric-lib-go/common/metrics"
 	"github.com/hyperledger/fabric-lib-go/healthz"
+	"github.com/hyperledger/fabric/core/chaincode/implicitcollection"
 	"github.com/hyperledger/fabric/core/common/ccprovider"
 	"github.com/hyperledger/fabric/core/ledger"
 	"github.com/hyperledger/fabric/core/ledger/cceventmgmt"
@@ -23,6 +24,7 @@ import (
 	"github.com/hyperledger/fabric/core/ledger/kvledger/txmgmt/statedb/stateleveldb"
 	"github.com/hyperledger/fabric/core/ledger/util"
 	"github.com/pkg/errors"
+	"github.com/spf13/viper"
 )
 
 var logger = flogging.MustGetLogger("privacyenabledstate")
@@ -31,6 +33,10 @@ const (
 	nsJoiner       = "$$"
 	pvtDataPrefix  = "p"
 	hashDataPrefix = "h"
+)
+
+const (
+	allImplicitCollectionNotation string = "*"
 )
 
 // StateDBConfig encapsulates the configuration for stateDB on the ledger.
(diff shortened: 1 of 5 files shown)


core/ledger/kvledger/txmgmt/privacyenabledstate/db.go
```diff
@@ -124,12 +130,13 @@ func (p *DBProvider) Drop(ledgerid string) error {
...
 	statedb.VersionedDB
 	metadataHint *metadataHint
+	localMspId   string
 }
 
 // NewDB wraps a VersionedDB instance. The public data is managed directly by the wrapped versionedDB.
 // For managing the hashed data and private data, this implementation creates separate namespaces in the wrapped db
 func NewDB(vdb statedb.VersionedDB, ledgerid string, metadataHint *metadataHint) (*DB, error) {
-	return &DB{vdb, metadataHint}, nil
+	return &DB{vdb, metadataHint, viper.GetString("peer.localMspId")}, nil
```
denyeart: Getting the config from within the implementation is considered a bad practice. We use the dependency injection pattern to pass down dependencies including peer configuration.
In this case, in kv_ledger_provider.go when calling privacyenabledstate.NewDBProvider() you can pass the existing 
 `p.initializer.MembershipInfoProvider.MyImplicitCollectionName()`.

denyeart (changes requested):
Sorry for the delay but I wanted to actually test it myself since there was no integration test here. This is an overall good approach, thank you for filling this gap!

In my test I found that the implicit collection wildcard approach causes an error at deploy time:

`Error: error getting chaincode bytes: walk failed: metadata file path or name is not supported: META-INF/statedb/couchdb/collections/_implicit_org_*/indexes/`

This is because of this validation that doesn't allow asterisk:

https://github.com/hyperledger/fabric/blob/main/internal/ccmetadata/validators.go#L31

denyeart:
We'll want an integration test for both the single org use case and all org use case.
It should be feasible to extend the existing CouchDB index integration test here:
https://github.com/hyperledger/fabric/blob/main/integration/ledger/couchdb_indexes_test.go

denyeart:
Not many people use Windows for Fabric, but still, asterisk is not allowed in Windows directory names. Maybe use `_implicit_all_orgs`?

C0rWin:
Not active for more than one year, closing it for the meantime.

PR #4803, opened 2024-04-09
Update gRPC dependency to address GO-2024-2687

bestbeforetoday:
- Use grpc.NewClient() instead of (soon to be) deprecated grpc.Dial().
- Update staticcheck version to avoid panic when running `make basic-checks` -- at least with Go 1.22.
- Regenerate documentation and remove trailing space that caused check failure.


denyeart:
Isn't GO-2024-2687 resolved by updating Go rather than dependencies?

@benjsmi is actually working on update to Go 1.21.9 today. We plan to update to Go 1.22 after the upcoming v2.5.7 release.

For the other updates, ideally they would be separate PRs or at least separate commits in this PR. But I realize you probably hit the issues after making updates.

Any idea why the doc trailing spaces became an issue for you but not for prior checks?

denyeart:
Ok, main and release-2.5 have been updated to Go 1.21.9 now.

Security vulnerability scan is still showing we need at least `golang.org/x/net@v0.23.0`, so this PR should indeed resolve the remaining issue.

bestbeforetoday:
[grpc-go v1.61.2](https://github.com/grpc/grpc-go/releases/tag/v1.61.2) (and v1.62.2) mention http2 library updates to address [CVE-2023-45288](https://www.kb.cert.org/vuls/id/421644). This PR would resolve that.

The additional changes seemed to be required to make my local checks and the automated build happy.

Unfortunately the gRPC update seems to have changed some behaviour around gRPC status and the detail protobuf messages we deliver using that mechanism, which breaks the gateway unit tests. Right now I don't know what needs to be done to resolve that.

denyeart:
@bestbeforetoday Ben was able to resolve the vulnerability by simply updating the golang.org/x/net to v0.23.0 in
https://github.com/hyperledger/fabric/pull/4811.

Updating grpc still a good idea but not urgent I think.

Also, is the Dial() versus NewClient() change required now, or soon, or are you just be a good citizen and cleaning up a deprecation?

bestbeforetoday:
> @bestbeforetoday Ben was able to resolve the vulnerability by simply updating the golang.org/x/net to v0.23.0 in #4811.
> 
> Updating grpc still a good idea but not urgent I think.

👍 

> Also, is the Dial() versus NewClient() change required now, or soon, or are you just be a good citizen and cleaning up a deprecation?

gRPC [v1.63.0](https://github.com/grpc/grpc-go/releases/tag/v1.63.0) deprecated Dial() and introduced NewClient() as its replacement. Rather than store up technical debt and (potentially) fail linting checks, I changed to use the new API. [v1.63.1](https://github.com/grpc/grpc-go/releases/tag/v1.63.1) then removed the deprecated tag but it is still scheduled for deprecation in v1.64.

I'll close this for now and somebody can figure out the unit test breakages with later gRPC versions another time.

PR #4825, opened 2024-04-24
adding external chaincode binaries to the base peer image WIP

rlfnb:
Simplify the use of external chaincodes

#### Type of change

- New feature

Might be a new feature or an improvement.

#### Description

Using external chaincode as a service needs to change the peer image providing three binaries (detect/build/release). To simplify the setup for use cases where docker (podman, kubernetes, openshift to name a few) is not available you are forced to use external chaincodes.

#### Additional details

As its basically my first PR for fabric, feedback regarding tools/guidelines would be appreciated.
Maybe, the functionality could be moved into the existing tools in ccaasbuilder, but I wont risk a bug there.

#### Release Note

The change wont impact existing installations

### DIFF
--- a/ccaas_builder/cmd/detect/main.go
+++ b/ccaas_builder/cmd/detect/main.go
@@ -26,8 +26,6 @@ func main() {
 		logger.Printf("::Error: %v\n", err)
 		os.Exit(1)
 	}
-
-	logger.Printf("::Type detected as ccaas")
 }
 
 type chaincodeMetadata struct {
@@ -59,9 +57,11 @@ func run() error {
 		return errors.WithMessage(err, "Unable to parse the metadata.json file")
 	}
 
-	if strings.ToLower(metadata.Type) != "ccaas" {
+	if strings.ToLower(metadata.Type) != "ccaas" && strings.ToLower(metadata.Type) != "remote" {
 		return fmt.Errorf("chaincode type not supported: %s", metadata.Type)
 	}
+
+	logger.Printf("::Type detected as: " + strings.ToLower(metadata.Type))
 
 	// returning nil indicates to the peer a successful detection
 	return nil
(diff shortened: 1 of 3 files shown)


denyeart:
When ccaasbuilder was added, the thought was that it would be the one and only 'generic' builder included in the open source version of the peer. Other use case specific builders would have to be built into your own peer image, or perhaps shared in the fabric-samples repository.

That being said, I'd like to understand your use case and why you can't use the existing ccaasbuilder.

It looks like the externalbuilder code is a copy of the ccaasbuilder code?

rlfnb:
totally agree, I just copied ccaasbuilder to make sure i dont interfere with existing functionality. If its ok for you, I would simply integrate the three cmds into existing ccaasbuilder commands. agree?
My intention is  to make external chaincode deployments simpler (as I'm using fabric in non-docker environments only). Personally, I can build my custom peer image as well and everything will work out for me, but my idea was to make it easier for others as well (e.g. kubernetes/openshift/podman users).

denyeart:
Ok, do you mean for discussion purposes you will update the externalbuilder code with your ideas? That would be fine for discussion. If your improvements are compatible enhancements with the existing ccaasbuilder, then you could just update ccaasbuilder. I'm still not sure the extent of changes you are proposing so I'll defer to your choice for the initial PR and discussion.

Also I'll mention that the name externalbuilder is not very clear, because that could imply a traditional external builder that actually builds chaincode, or a chaincode-as-a-service external 'builder' that points to an external service ('builder' in quotes because it doesn't actually build chaincode). I think you are talking about the latter.

rlfnb:
so, I did find some time and would like ask for a small review @denyeart . I added remote as a type to _detect_ (the ccaas binary), so we wont have additional commands. If its ok like its done for _detect_, I would proceed with _build_ and _release_ as well.

C0rWin:
No progress been done for a long time, PR seems stale, closing it for now, could be re-opened if needed.

Issue #4358, opened 2023-08-10
peer: error validating proposal: access denied: channel [] creator org unknown, creator is malformed


rlfnb:
### Description

https://stackoverflow.com/questions/76480170/validating-proposal-access-denied-channel-creator-org-unknown-creator-is-m

Seems to be a newly introduced thing, maybe an additional field in the channel config is missing? 
Basically, I followed the procedure here:

https://hyperledgendary.github.io/unstable-fabric-docs/create_channel/create_channel_test_net.html#step-two-create-the-application-channel

`
2023-08-10 07:00:27.002 UTC 0001 INFO [channelCmd] InitCmdFactory -> Endorser and orderer connections initialized
Error: proposal failed (err: rpc error: code = Unknown desc = error validating proposal: access denied: channel [] creator org unknown, creator is malformed)
`
The output indicates, setup of TLS and so on is ok?

### Steps to reproduce

$ podman run --rm --env-file=peercli.env -v $(pwd):/opt docker.io/hyperledger/fabric-tools peer channel join -o orderer.example.com:7050 -b /opt/channel-artifacts/blue.pb 
2023-08-10 07:00:27.002 UTC 0001 INFO [channelCmd] InitCmdFactory -> Endorser and orderer connections initialized
Error: proposal failed (err: rpc error: code = Unknown desc = error validating proposal: access denied: channel [] creator org unknown, creator is malformed)

rlfnb:
error description seems to be coming from:

https://github.com/hyperledger/fabric/blob/aa81271b8d4d7ade43cbbf2eb09b435896c12273/core/endorser/msgvalidation.go#L173

rlfnb:
I found the issue by myself, pretty annoying:

https://hyperledger-fabric.readthedocs.io/en/release-2.5/channel_update_tutorial.html?highlight=anchor%20peer#fetch-the-configuration

tells how to fetch the configuration:

`
# you can issue all of these commands at once
export PATH=${PWD}/../bin:$PATH
export FABRIC_CFG_PATH=${PWD}/../config/
export CORE_PEER_TLS_ENABLED=true
export CORE_PEER_LOCALMSPID="Org1MSP"
export CORE_PEER_TLS_ROOTCERT_FILE=${PWD}/organizations/peerOrganizations/org1.example.com/peers/peer0.org1.example.com/tls/ca.crt
[...]

PR #4841, opened 2024-05-06
Doc tutorials - Remove double quotes for env variables

denyeart:
Double quotes are problematic in some environments, see https://github.com/hyperledger/fabric/issues/4358.


satota2 (approved):
@denyeart Thank you for the update. LGTM!

PR #4843, opened 2024-05-07
Add GHA permission to broken link checker

denyeart:
Every github action should have top level permission set to read.


denyeart:
Broken Link Checker failure can be ignored.

PR #4849, opened 2024-05-09
Add logging spec for integration tests

denyeart:
Add logging spec for integration tests so that it can
be edited when troubleshooting specific tests.


satota2 (approved):
@denyeart LGTM!

Issue #4847, opened 2024-05-08
Allow the peer delivery client to select between the Deliverer or BFTDeliverer


tock-ibm:
### Current Status

the deliverclient 
core/deliverservice/deliveryclient.go

Uses either the Deliverer
common/deliverclient/blocksprovider/deliverer.go

or the BFTDeliverer
common/deliverclient/blocksprovider/bft_deliverer.go

depending on the consensus type - etcdraft/BFT.

### Goal

We want to add to the core.yaml a key "peer.deliveryclient.policy" = "simple" / "cluster" that would allow the use of "simple" even when the consensus type is BFT. When consensus type is etcdraft this key is ignored.

### Solution

We want to add to the core.yaml a key "peer.deliveryclient.policy" = "simple" / "cluster" that would allow the use of "simple" even when the consensus type is BFT. When consensus type is etcdraft this key is ignored.

tock-ibm:
@semil

pfi79:
@tock-ibm Why do you want to do this?
What problem in the fabric will it solve?

tock-ibm:
> @tock-ibm Why do you want to do this? What problem in the fabric will it solve?

It was suggested by @yacovm that in case the peer organization trusts the ordering service and wants to save bandwidth, it could select the "simple" policy that consumes blocks from a single orderer, rather than the "cluster" policy that consumes blocks from a single orderer and block headers from every other orderer.

tock-ibm:
Thinking about it, there is also another policy we could easily implement, in which the peer specifies the ordering-org it trusts and wants to pull blocks from. @pfi79 @yacovm @denyeart what do you think?

yacovm:
> Thinking about it, there is also another policy we could easily implement, in which the peer specifies the ordering-org it trusts and wants to pull blocks from. @pfi79 @yacovm @denyeart what do you think?

Yes that's also an option.

pfi79:
> Thinking about it, there is also another policy we could easily implement, in which the peer specifies the ordering-org it trusts and wants to pull blocks from. @pfi79 @yacovm @denyeart what do you think?

It can be done. But why?
I mean, the configuration and code will become more difficult. And what do we get?

This is not the only “simple” option for delivering blocks to peer.

yacovm:
> > Thinking about it, there is also another policy we could easily implement, in which the peer specifies the ordering-org it trusts and wants to pull blocks from. @pfi79 @yacovm @denyeart what do you think?
> 
[...]

PR #4856, opened 2024-05-12
Allow the peer delivery client to select policy

semil:
#### Type of change

#### Description

Issue: #4847

#### Related issues

#4847

### DIFF
--- a/core/deliverservice/config.go
+++ b/core/deliverservice/config.go
@@ -9,6 +9,7 @@ package deliverservice
 import (
 	"encoding/pem"
 	"os"
+	"slices"
 	"time"
 
 	"github.com/hyperledger/fabric/common/deliverclient/orderers"
@@ -25,6 +26,7 @@ const (
 	DefaultConnectionTimeout           = time.Second * 3
 	DefaultBlockCensorshipTimeoutKey   = time.Second * 30
 	DefaultMinimalReconnectInterval    = time.Millisecond * 100
+	DefaultPolicy                      = "cluster"
 )
 
 // DeliverServiceConfig is the struct that defines the deliverservice configuration.
@@ -52,6 +54,11 @@ type DeliverServiceConfig struct {
 	// OrdererEndpointOverrides is a map of orderer addresses which should be
 	// re-mapped to a different orderer endpoint.
 	OrdererEndpointOverrides map[string]*orderers.Endpoint
+	// Determines which delivery client will be used when consensus type is "BFT"
+	// (when consensus type is "etcdraft" this key is ignored).
+	// "simple" - use CFT deliverer
+	// "cluster" - use BFT deliverer
+	Policy string
 }
 
 type AddressOverride struct {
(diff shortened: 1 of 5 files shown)


core/deliverservice/config.go
```diff
@@ -52,6 +54,7 @@ type DeliverServiceConfig struct {
 	// OrdererEndpointOverrides is a map of orderer addresses which should be
 	// re-mapped to a different orderer endpoint.
 	OrdererEndpointOverrides map[string]*orderers.Endpoint
+	Policy                   string
```
tock-ibm: Add comment with an explanation, same as in the yaml file.

core/deliverservice/deliveryclient.go
```diff
@@ -128,14 +128,25 @@ func (d *deliverServiceImpl) StartDeliverForChannel(chainID string, ledgerInfo b
...
 
+	createCftDeliverer := false
 	switch ct := oc.ConsensusType(); ct {
 	case "etcdraft":
-		d.blockDeliverer, err = d.createBlockDelivererCFT(chainID, ledgerInfo)
+		createCftDeliverer = true
 	case "BFT":
-		d.blockDeliverer, err = d.createBlockDelivererBFT(chainID, ledgerInfo)
+		switch d.conf.DeliverServiceConfig.Policy {
+		case "cluster", "":
```
tock-ibm: at this stage the policy should not be "", as we det the default "cluster" much earlier.

core/deliverservice/deliveryclient.go
```diff
@@ -128,14 +128,25 @@ func (d *deliverServiceImpl) StartDeliverForChannel(chainID string, ledgerInfo b
...
+			createCftDeliverer = true
+		default:
+			err = errors.Errorf("unexpected delivey service policy: `%s`", d.conf.DeliverServiceConfig.Policy)
+		}
 	default:
 		err = errors.Errorf("unexpected consensus type: `%s`", ct)
 	}
+	if createCftDeliverer {
+		d.blockDeliverer, err = d.createBlockDelivererCFT(chainID, ledgerInfo)
+	}
```
tock-ibm: why not simply call this function from 2 places? it will save 4 line of code, no? (and would be more readable in my opinion).

tock-ibm (commented):
Looks good overall. Just a few minor comments.
We'll need to add an integration test that checks that this indeed works (another PR). We can use the same approach is in the orderer.ReplicationPolicy: copy a simple test from the smartbft test suite (that uses a peer) and simply change to Policy="simple".

core/deliverservice/deliveryclient.go
semil: it is needed for the unit tests, they do not go through the code that "patches" this case, so we need the code here in the delivery client itself.

core/deliverservice/deliveryclient.go
semil: correct :)

core/deliverservice/deliveryclient.go
tock-ibm: Yes, but it unit test you should check that if the policy is not set you get an error... IMO

semil: fixed :)

tock-ibm (approved):

PR #4870, opened 2024-05-21
Bump actions/download-artifact to v4

denyeart:
download-artifact needs to be v4 because
upload-artifact is already v4.
They must be at the same level for release pipeline to work correctly.


yacovm (approved):

denyeart:

PR #4879, opened 2024-06-02
Remove global endpoints

tock-ibm:
#### Type of change

- New feature

#### Description

Remove and prevent the use of global orderer endpoints on version V3 and above

#### Related issues

#4763


cmd/configtxgen/main.go
```diff
@@ -118,12 +119,26 @@ func doInspectChannelCreateTx(inspectChannelCreateTx string) error {
 func doPrintOrg(t *genesisconfig.TopLevel, printOrg string) error {
 	for _, org := range t.Organizations {
 		if org.Name == printOrg {
-			og, err := encoder.NewOrdererOrgGroup(org)
+			if len(org.OrdererEndpoints) > 0 {
+				// An Orderer OrgGroup
+				channelCapabilities := t.Capabilities["Channel"]
```
yacovm, line 124: can you explain why we need this?

common/channelconfig/orderer.go
```diff
@@ -228,6 +233,19 @@ func (oc *OrdererConfig) validateBatchTimeout() error {
 	return nil
 }
 
+func (oc *OrdererConfig) validateSomeOrgHasEndpoints() error {
+	var someOrgHasEndpoints bool
+	for _, org := range oc.Organizations() {
+		fmt.Printf(">>> org %v", org)
```
yacovm: please remove this >>>

integration/lifecycle/lifecycle_test.go
```diff
@@ -295,6 +295,9 @@ var _ = Describe("Lifecycle", func() {
 		By("updating the channel config to include org3")
 		// get the current channel config
 		currentConfig := nwo.GetConfig(network, testPeers[0], orderer, "testchannel")
+		//jsonObj, _ := json.MarshalIndent(currentConfig, "", "    ")
```
yacovm: please remove commented out code

integration/lifecycle/lifecycle_test.go
```diff
@@ -304,12 +307,19 @@ var _ = Describe("Lifecycle", func() {
 		})
 		Expect(err).NotTo(HaveOccurred())
 		Eventually(sess, network.EventuallyTimeout).Should(gexec.Exit(0))
-		org3Group := &ordererext.DynamicOrdererOrgGroup{ConfigGroup: &common.ConfigGroup{}}
+		org3Group := &peerext.DynamicApplicationOrgGroup{ConfigGroup: &common.ConfigGroup{}}
 		err = protolator.DeepUnmarshalJSON(bytes.NewBuffer(sess.Out.Contents()), org3Group)
 		Expect(err).NotTo(HaveOccurred())
+		// delete(org3Group.Values, "Endpoints")
```
yacovm: please remove commented out code

integration/raft/migration_test.go
```diff
@@ -128,10 +129,25 @@ var _ = Describe("ConsensusTypeMigration", func() {
...
+			peer2 := network.Peer("Org2", "peer0")
+
+			// config update that should fail
+			By("Config update with global level endpoints")
+			config := nwo.GetConfig(network, peer, o1, "testchannel")
+			updatedConfig := proto.Clone(config).(*common.Config)
+			addGlobalLevelEndpointsToConfig(updatedConfig)
+			updateOrdererEndpointsConfigFails(network, o1, "testchannel", config, updatedConfig, peer, peer, peer2)
+
+			// config update that succeeds but cause a failure during the migration step
```
yacovm: causes

integration/raft/migration_test.go
```diff
@@ -594,6 +626,58 @@ func updateOrdererConfigFailed(n *nwo.Network, orderer *nwo.Orderer, channel str
...
+
+	sess, err = n.PeerAdminSession(peer, commands.ChannelUpdate{
+		ChannelID:  channel,
+		Orderer:    n.OrdererAddress(orderer, nwo.ListenPort),
+		File:       updateFile,
+		ClientAuth: n.ClientAuthRequired,
+	})
+	Expect(err).NotTo(HaveOccurred())
+	Eventually(sess, n.EventuallyTimeout).Should(gexec.Exit(1))
+	Expect(sess.Err).NotTo(gbytes.Say("Successfully submitted channel update"))
```
yacovm: where do we see why it fails?

common/channelconfig/orderer.go
tock-ibm: done

integration/lifecycle/lifecycle_test.go
tock-ibm: done

integration/lifecycle/lifecycle_test.go
tock-ibm: done

integration/raft/migration_test.go
tock-ibm: done

cmd/configtxgen/main.go
tock-ibm: The `org` data structure here carries either a App org or an Orderer org.
We want to check whether to marshal the org into a `&ordererext.DynamicOrdererOrgGroup` of a `&peerext.DynamicApplicationOrgGroup`. 

Previously it was just a `&ordererext.DynamicOrdererOrgGroup`. Now when the encoder for the orderer group is strict about having endpoints (in v3), if you pass an app org it crashes. This was the cause of the failures in the lifecycle integration test.

In addition, I inserted the channelCapabilities into `encoder.NewOrdererOrgGroup` so that we can be strict about endpoints in v3 but not on earlier versions.

integration/raft/migration_test.go
tock-ibm: done

orderer/common/cluster/util_test.go
```diff
@@ -470,7 +539,32 @@ func TestEndpointconfigFromConfigBlockFailures(t *testing.T) {
 }
 
 func TestBlockValidationPolicyVerifier(t *testing.T) {
+	dir := filepath.Join(os.TempDir())
+	os.Mkdir(dir, 0o700)
+	defer os.RemoveAll(dir)
```
denyeart: Maybe this RemoveAll() is removing temp files that the other tests need?

tock-ibm: I think that is it.... it should be `t.TempDir()` ... 
Thanks!

orderer/common/multichannel/chainsupport.go
```diff
@@ -149,6 +150,23 @@ func (cs *ChainSupport) ProposeConfigUpdate(configtx *cb.Envelope) (*cb.ConfigEn
...
+	// if we move from V2 to V3 we need the check the org endpoints, if one of them is empty return an error
+	if !isOldV3 && isNewV3 {
+		orderer, err := bundle.OrdererConfig()
+		if err {
+			return nil, errors.New("new config is missing orderer group")
+		}
+
+		for _, org := range orderer.Organizations() {
+			if len(org.Endpoints()) == 0 {
+				return nil, errors.Errorf("illegal orderer config detected during consensus metadata validation: endpoints of org %s are missing", org.Name())
```
yacovm: I don't see this covered in any test, am I wrong?

yacovm:
Looks good modulo the comment above.

@denyeart maybe you can also review this one?

orderer/common/multichannel/chainsupport.go
tock-ibm: Correct. I'll add some.

tock-ibm: And... this piece of code is not in the right place. It should be in `func (b *Bundle) ValidateNew(nb Resources) error `

common/channelconfig/orderer.go
```diff
@@ -140,6 +144,13 @@ func NewOrdererConfig(ordererGroup *cb.ConfigGroup, mspConfig *MSPConfigHandler,
 			return nil, err
 		}
 	}
+
+	if channelCapabilities.ConsensusTypeBFT() {
+		if err := oc.validateSomeOrgHasEndpoints(); err != nil {
```
denyeart: Here we validate that _some_ (any) orderer org has endpoints.
Above in common/channelconfig/bundle.go we validate that every orderer org has endpoints.

Should we be consistent? Or is there a reason the validation is different?

internal/peer/common/common.go
```diff
@@ -280,7 +280,23 @@ func GetOrdererEndpointOfChain(chainID string, signer Signer, endorserClient pb.
 		return nil, errors.WithMessage(err, "error loading channel config")
 	}
 
-	return bundle.ChannelConfig().OrdererAddresses(), nil
+	ordererAddresses := bundle.ChannelConfig().OrdererAddresses()
+	if len(ordererAddresses) > 0 {
+		logger.Warningf("Deprecated global OrdererAddresses exist: %v; ignoring", ordererAddresses)
+	}
```
denyeart: Why ignore the old global OrdererAddresses in this location, but not in other locations, for example in:
common/deliverclient/blocksprovider

I agree that if capability V3_0 is enabled, there should be no old global OrdererAddresses. But I'm thinking if we honor the old global OrdererAddresses in some places without capability checks, we should honor it everywhere without capability checks.

I believe this reference is only used for the peer client (not the peer node). It should probably return a union of the global addresses and orderer org endpoints.

sampleconfig/configtx.yaml
```diff
@@ -658,6 +658,10 @@ Profiles:
                       Admins:
                           Type: Signature
                           Rule: "OR('SampleOrg.member')"
+                  OrdererEndpoints:
+                    - "127.0.0.1:7050"
+                    - "127.0.0.1:7051"
+                    - "127.0.0.1:7052"
```
denyeart, line 664: SampleOrg already has an OrdererEndpoint (line 76).

Note that the comment in line 316 of this file needs to be updated.

denyeart (commented):
I tested an upgrade from v2.5 to v3.0 with global addresses configured (without updating to V3_0 capability).
It appears to still work - peers pull blocks from orderers using the configured global addresses. This is good. I agree we should only enforce it when V3_0 capability is set.

A few minor comments...

sampleconfig/configtx.yaml
tock-ibm: True, I just override the SampleOrg definition for this particular profile, `SampleAppChannelEtcdRaft`, which is used in one of the unit tests I updated. I wanted to get multiple addresses and not just one. See for example `common/deliverclient/blocksprovider/deliverer_test.go` line 713 as well as `common_test.go` above.

I updated the comment.

common/channelconfig/orderer.go
tock-ibm: There was a discussion on whether to allow adding an org without endpoints. I think that is a remnant of that discussion. I think it should be strict, so I'll change it to be consistent with the rest.

internal/peer/common/common.go
tock-ibm: You are correct, there is a problem here.

Two lines above, the code 
```
	bundle, err := channelconfig.NewBundle(chainID, channelConfig, cryptoProvider)
	if err != nil {
		return nil, errors.WithMessage(err, "error loading channel config")
	}
```
will make sure that in V3, we have no global address and an endpoint for each org.

Here, I should probably replicate the peer policy that if both global and per-org exist (i.e. in V2), global are ignored, but if only global exist, they are taken (not a concatenation).

tock-ibm: I also updated `common_test.go` to reflect the above change.

common/channelconfig/channel.go
```diff
@@ -213,3 +221,10 @@ func (cc *ChannelConfig) validateOrdererAddresses() error {
 	}
 	return nil
 }
+
+func (cc *ChannelConfig) validateNoOrdererAddresses() error {
+	if len(cc.protos.OrdererAddresses.Addresses) > 0 {
+		return fmt.Errorf("global OrdererAddresses are not allowed in V3, use org specific addresses only")
```
denyeart, line 227:
```suggestion
		return fmt.Errorf("global OrdererAddresses are not allowed with V3_0 capability, use org specific addresses only")
```

common/channelconfig/bundle.go
```diff
@@ -93,6 +93,17 @@ func (b *Bundle) ValidateNew(nb Resources) error {
 			}
 		}
 
+		// When we move from V2 to V3 we insist on per Org endpoints for every org
```
denyeart, line 96:
```suggestion
		// When we move to capability V3_0 we insist on per Org endpoints for every org
```

denyeart (commented):
Looks good now, just a couple comment and message nits so that we don't get confused in the future.

denyeart (approved):
Since only a couple nits remain, let's go ahead and merge.

tock-ibm:
@denyeart Thanks! I saw you addressed the comments while I was away on vacation.

Issue #2983, opened 2021-10-14
Refactor Developing Applications documentation


denyeart:
"Developing Applications" [doc](https://hyperledger-fabric.readthedocs.io/en/latest/developapps/developing_applications.html) has gotten somewhat outdated and is based on the Commercial Paper chaincode that many people don't understand. It should be overhauled and potentially based on a simpler chaincode that new developers can more easily understand.

Additionally it needs to be refactored for the new embedded peer Gateway.

DARK-art108:
Can I work on this issue? @denyeart

denyeart:
@DARK-art108 Yes, please coordinate with @bestbeforetoday (aka Mark Lewis, aka bestbeforetoday) on the proposed approach. Mark is the application SDK lead and should be involved in Developing Applications changes.

jonmchan:
Offline signing seems like a significant feature or use case for the Fabric Gateway SDKs. I was unable to find it in the documentation anywhere except in the code and tests. Would this be something that should get a section?

bestbeforetoday:
> Offline signing seems like a significant feature or use case for the Fabric Gateway SDKs. I was unable to find it in the documentation anywhere except in the code and tests. Would this be something that should get a section?

There is some description and code examples in the API documentation for the Contract in each of the client languages:

- Go: https://pkg.go.dev/github.com/hyperledger/fabric-gateway/pkg/client#Contract
- Node: https://hyperledger.github.io/fabric-gateway/main/api/node/interfaces/Contract.html
- Java: https://hyperledger.github.io/fabric-gateway/main/api/java/org/hyperledger/fabric/client/Contract.html

Any additional documentation giving a language agnostic overview of the client API structure and usage patterns (perhaps a reworked developing applications section) should cover off-line signing too.

denyeart:
Developing Applications doc has been removed. Intent now is to add some of the main concepts to the Key Concepts and Tutorial topics.

joaormatos:
> Developing Applications doc has been removed. Intent now is to add some of the main concepts to the Key Concepts and Tutorial topics.

Meanwhile there are a few dangling links to the Developing Applications section in the current version of the documentation.

[A code search for _developapps_ easily reveals them.](https://github.com/search?q=repo%3Ahyperledger%2Ffabric+path%3A%2F%5Edocs%5C%2F%2F+developapps&type=code)

As a result, I couldn't find a full explanation of chaincode namespaces in the current version of the docs.
[...]

PR #4880, opened 2024-06-04
Fix: URL Path in smartcontract.html

parth5805:
fix: "Developing applications topic" URL path path

#### Type of change

- Documentation update


denyeart:
Actually, the Developing Applications content was removed because it was outdated. There is an issue to clean up the content and remove the dead links - https://github.com/hyperledger/fabric/issues/2983#issuecomment-2147540752

PR #4884, opened 2024-06-05
up go 1.22.4

pfi79:


go.mod
```diff
@@ -1,6 +1,6 @@
 module github.com/hyperledger/fabric
 
-go 1.22
+go 1.22.4
```
denyeart: Let's keep the 2 digit version here. Every other go.mod file I've seen uses the 2 digit version since new features are only introduced in major.minor releases.

pfi79: I can do that. 

But, I did so according to the changes in the documentation, since go version 1.21. At least that's how I understood the documentation.

https://go.dev/doc/go1.21

https://go.dev/doc/toolchain#version

denyeart: Ok, thanks for the links. I still think we should keep it at major.minor (e.g. 1.22). We have the _option_ to include the patch version here (e.g. 1.22.4), but my opinion is that we should do that only if the patch becomes important, that is, if Fabric depends on some behavior of the patch. Since Fabric continues to work with any 1.22.x version, I think 1.22 continues to be correct here.

denyeart (approved):

pfi79:
@denyeart please, run manual actions vulnerability-scan

denyeart:
> @denyeart please, run manual actions vulnerability-scan

scan run successfully.

Backported to release-2.5 in https://github.com/hyperledger/fabric/pull/4886

PR #4892, opened 2024-06-13
Implement 'osnadmin channel info' subcommand and update tests

### LINKED ISSUE #4890: Should we introduce a `osnadmin channel info` subcmd?
### Current Status

osnadmin support following cmds today: join, list and remove.
If we want to get detailed information on specific channel, we use the `list` cmd with specific channel id.
```
List channel information for an Ordering Service Node (OSN). If the channelID flag is set, more detailed information will be provided for that channel.
```

### Expected

Another better option is to let `list` only do list all channels work, and introduce a separate subcmd such as `info` to do the detailed info query on specific channel.

### Solution

Add a new subcmd of `info`.

Yes!

Priyansurout:
Implemented the `osnadmin channel info` subcommand to retrieve detailed information about a specific channel from an Ordering Service Node (OSN). This involved modifying the CLI commands and adding corresponding functionality to interact with the channel participation API.

#### Type of change

- New feature

Implemented the `osnadmin channel info` subcommand to retrieve detailed information about a specific channel from an Ordering Service Node (OSN). 

- Introduced `info` subcommand under `osnadmin channel` to support querying individual channel details.
- Updated existing tests and added new tests to cover the new feature and edge cases.

Closes #4890

### DIFF
--- a/cmd/osnadmin/main.go
+++ b/cmd/osnadmin/main.go
@@ -52,7 +52,10 @@ func executeForArgs(args []string) (output string, exit int, err error) {
 	configBlockPath := join.Flag("config-block", "Path to the file containing an up-to-date config block for the channel").Short('b').Required().String()
 
 	list := channel.Command("list", "List channel information for an Ordering Service Node (OSN). If the channelID flag is set, more detailed information will be provided for that channel.")
-	listChannelID := list.Flag("channelID", "Channel ID").Short('c').String()
+	// listChannelID := list.Flag("channelID", "Channel ID").Short('c').String()
+	// Define the new info subcommand
+	info := channel.Command("info", "Get detailed information about a specific channel.")
+	infoChannelID := info.Flag("channelID", "Channel ID").Short('c').Required().String()
 
 	remove := channel.Command("remove", "Remove a channel from an Ordering Service Node (OSN).")
 	removeChannelID := remove.Flag("channelID", "Channel ID").Short('c').Required().String()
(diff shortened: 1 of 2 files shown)


cmd/osnadmin/main.go
```diff
@@ -52,7 +52,8 @@ func executeForArgs(args []string) (output string, exit int, err error) {
 	configBlockPath := join.Flag("config-block", "Path to the file containing an up-to-date config block for the channel").Short('b').Required().String()
 
 	list := channel.Command("list", "List channel information for an Ordering Service Node (OSN). If the channelID flag is set, more detailed information will be provided for that channel.")
-	listChannelID := list.Flag("channelID", "Channel ID").Short('c').String()
+	info := channel.Command("info", "Get detailed information about a specific channel.")
+	infoChannelID := info.Flag("channelID", "Channel ID").Short('c').Required().String()
```
denyeart, line 56: I'm thinking we can do this in a backward compatible way.
Keep the support for passing channel id to list with this line:
`listChannelID := list.Flag("channelID", "Channel ID").Short('c').String()`

Then add an entirely new "info" command.

Change the function osnadmin.ListSingleChannel() to osnadmin.Info() and have both of them call it.

The help text doesn't need to mention that you can pass channel id to list command anymore, but keep the functionality there so that any existing users that depend on it don't break.

denyeart (changes requested):
You got a check failure because anytime you change commands, you need to regenerate the doc to match. 

See the instructions at
https://github.com/hyperledger/fabric/blob/main/docs/source/docs_guide.md#commands-reference-updates

Basically, the command help will be called when you regenerate the doc with each of the subcommands (including the new one which needs to be added in scripts/generateHelpDocs.sh), and this help text will be injected into the osnadminchannel.md doc topic which also needs to be included in the pull request.

You'll also need to add some text to the doc topic preamble and postscript.

To better understand, you can look at the commit that added the initial osnadmin help doc. This will show all the updates that are needed.
https://github.com/hyperledger/fabric/commit/1e90c8889448c79e402f06c84106c0ca89bbe731

PR #4912, opened 2024-06-26
find and fix error in tests

pfi79:
- For unit-test in smartbft I increased the waiting time for the desired altitude. There were cases when orders went to synchronization and there was simply not enough time and changed the heartbeat settings
- Fix the integration test in smartbft_block_deliverer_test.go

### DIFF
--- a/integration/smartbft/mock_orderer.go
+++ b/integration/smartbft/mock_orderer.go
@@ -182,7 +182,12 @@ func (mo *MockOrderer) deliverBlocks(
 		iterCh := make(chan struct{})
 		go func() {
 			if ledgerIdx > ledgerLastIdx {
-				time.Sleep(math.MaxInt64)
+				select {
+				case <-time.After(math.MaxInt64):
+				case <-ctx.Done():
+					close(iterCh)
+					return
+				}
 			}
 			block = mo.ledgerArray[ledgerIdx]
 			status = cb.Status_SUCCESS
--- a/integration/smartbft/smartbft_block_deliverer_test.go
+++ b/integration/smartbft/smartbft_block_deliverer_test.go
@@ -95,16 +95,14 @@ var _ = Describe("Smart BFT Block Deliverer", func() {
 
 		/* Create a network from the networkConfig */
 		network = nwo.New(networkConfig, testDir, client, StartPort(), components)
-		network.Profiles[0].SmartBFT = &nwo.SmartBFT{
-			LeaderHeartbeatTimeout: 10,
-			LeaderHeartbeatCount:   10,
-		}
 		network.GenerateConfigTree()
 		network.Bootstrap()
 
 		for _, orderer := range network.Orderers {
-			runner := network.OrdererRunner(orderer)
-			runner.Command.Env = append(runner.Command.Env, "FABRIC_LOGGING_SPEC=debug")
+			runner := network.OrdererRunner(orderer,
+				"FABRIC_LOGGING_SPEC=debug",
+				"ORDERER_GENERAL_BACKOFF_MAXDELAY=20s",
+			)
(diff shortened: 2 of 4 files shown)


integration/smartbft/mock_orderer.go
```diff
@@ -182,7 +182,12 @@ func (mo *MockOrderer) deliverBlocks(
 		iterCh := make(chan struct{})
 		go func() {
 			if ledgerIdx > ledgerLastIdx {
-				time.Sleep(math.MaxInt64)
```
yacovm, line 185: oh my god

integration/smartbft/smartbft_block_deliverer_test.go
```diff
@@ -95,16 +95,14 @@ var _ = Describe("Smart BFT Block Deliverer", func() {
...
-		}
 		network.GenerateConfigTree()
 		network.Bootstrap()
 
 		for _, orderer := range network.Orderers {
-			runner := network.OrdererRunner(orderer)
-			runner.Command.Env = append(runner.Command.Env, "FABRIC_LOGGING_SPEC=debug")
+			runner := network.OrdererRunner(orderer,
+				"FABRIC_LOGGING_SPEC=debug",
+				"ORDERER_GENERAL_BACKOFF_MAXDELAY=20s",
```
yacovm, line 104: why do we need the max delay?

orderer/consensus/smartbft/util_network_test.go
```diff
@@ -552,7 +554,7 @@ func createBFTChainUsingMocks(t *testing.T, node *Node, configInfo *ConfigInfo)
 				return
 			}
 			t.Logf("Node %d requested SendTransaction to node %d", node.NodeId, targetNodeId)
-			err := node.sendRequest(node.NodeId, targetNodeId, message)
+			err = node.sendRequest(node.NodeId, targetNodeId, message)
```
yacovm, line 557: why?

pfi79: It seems to me that using existing variables is better than declaring new ones and allocating memory for them.

orderer/consensus/smartbft/chain_test.go
```diff
@@ -361,6 +360,7 @@ func TestAddAndRemoveNodeWithoutStop(t *testing.T) {
 
 	// start a network
 	networkSetupInfo := NewNetworkSetupInfo(t, channelId, dir)
+	networkSetupInfo.configInfo.leaderHeartbeatTimeout = time.Minute
```
yacovm, line 363: isn't it the default?

integration/smartbft/smartbft_block_deliverer_test.go
pfi79: These are my recommendations for setting up grpc so that some individual orderer can't leave the consensus due to lack of heartbeat from the leader.

The author added the settings, most likely by copying them:
```
                network.Profiles[0].SmartBFT = &nwo.SmartBFT{
			LeaderHeartbeatTimeout: 10,
			LeaderHeartbeatCount:   10,
		}
```

These settings are not needed here. I've tweaked it as it should be.

yacovm (approved):

orderer/consensus/smartbft/chain_test.go
pfi79: Here's where I don't fully understand mock.
But I caught that nodes go to the state of view change due to heartbeat.

In this case for TestAddNodeWhileAnotherNodeIsDown test it takes 20 seconds, and for TestAddAndRemoveNodeWithoutStop test it takes 1 minute.

Otherwise they crash.

Perhaps there is something missing in the mock to function properly.

PR #4941, opened 2024-08-03
Remove slash-commands workflow

denyeart:
The workflow has not been used for the last several years. Additionally, we'd like to ensure that only authorized users can close issues.


bestbeforetoday (approved):
This looks like a good move to me

yacovm (approved):

PR #4942, opened 2024-08-06
added timestamp proposal check so that it does not go beyond timewindow

pfi79:
Added an auth filter that checks the timestamp of the proposal so that it doesn't differ too much from the peer time using the peer.authentication.timewindow parameter.

Also made it so that gateway calling Proposal locally does not pass auth filters

### DIFF
--- a/core/handlers/auth/filter/timewindow.go
+++ b/core/handlers/auth/filter/timewindow.go
@@ -1,0 +1,69 @@
+/*
+Copyright IBM Corp, SecureKey Technologies Inc. All Rights Reserved.
+
+SPDX-License-Identifier: Apache-2.0
+*/
+
+package filter
+
+import (
+	"context"
+	"time"
+
+	"github.com/hyperledger/fabric-protos-go/peer"
+	"github.com/hyperledger/fabric/core/handlers/auth"
+	"github.com/hyperledger/fabric/protoutil"
+	"github.com/pkg/errors"
+)
+
+// NewTimeWindowCheckFilter creates a new Filter that checks timewindow expiration
+func NewTimeWindowCheckFilter(timeWindow time.Duration) auth.Filter {
+	return &timewindowCheckFilter{
+		timeWindow: timeWindow,
+	}
+}
+
+type timewindowCheckFilter struct {
+	next       peer.EndorserServer
+	timeWindow time.Duration
+}
+
+// Init initializes the Filter with the next EndorserServer
+func (f *timewindowCheckFilter) Init(next peer.EndorserServer) {
+	f.next = next
+}
+
+func validateTimewindowProposal(signedProp *peer.SignedProposal, timeWindow time.Duration) error {
+	prop, err := protoutil.UnmarshalProposal(signedProp.ProposalBytes)
+	if err != nil {
+		return errors.Wrap(err, "failed parsing proposal")
+	}
+
+	hdr, err := protoutil.UnmarshalHeader(prop.Header)
+	if err != nil {
+		return errors.Wrap(err, "failed parsing header")
+	}
+
+	chdr, err := protoutil.UnmarshalChannelHeader(hdr.ChannelHeader)
+	if err != nil {
+		return errors.Wrap(err, "failed parsing channel header")
+	}
+
+	timeProposal := chdr.Timestamp.AsTime().UTC()
+	now := time.Now().UTC()
+
+	if timeProposal.Add(timeWindow).Before(now) || timeProposal.Add(-timeWindow).After(now) {
+		return errors.Errorf("request unauthorized due to incorrect timestamp %s, peer time %s, peer.authentication.timewindow %s",
+			timeProposal.Format(time.RFC3339), now.Format(time.RFC3339), timeWindow.String())
+	}
+
+	return nil
+}
+
+// ProcessProposal processes a signed proposal
+func (f *timewindowCheckFilter) ProcessProposal(ctx context.Context, signedProp *peer.SignedProposal) (*peer.ProposalResponse, error) {
+	if err := validateTimewindowProposal(signedProp, f.timeWindow); err != nil {
+		return nil, err
+	}
+	return f.next.ProcessProposal(ctx, signedProp)
+}
--- a/core/handlers/library/registry.go
+++ b/core/handlers/library/registry.go
@@ -10,6 +10,7 @@ import (
 	"fmt"
 	"reflect"
 	"sync"
+	"time"
 
 	"github.com/hyperledger/fabric-lib-go/common/flogging"
 	"github.com/hyperledger/fabric/core/handlers/auth"
@@ -47,10 +48,11 @@ const (
 )
 
 type registry struct {
-	filters    []auth.Filter
-	decorators []decoration.Decorator
-	endorsers  map[string]endorsement2.PluginFactory
-	validators map[string]validation.PluginFactory
+	filters                  []auth.Filter
+	decorators               []decoration.Decorator
+	endorsers                map[string]endorsement2.PluginFactory
+	validators               map[string]validation.PluginFactory
+	authenticationTimeWindow time.Duration
 }
 
 var (
@@ -63,8 +65,9 @@ var (
 func InitRegistry(c Config) Registry {
 	once.Do(func() {
 		reg = registry{
-			endorsers:  make(map[string]endorsement2.PluginFactory),
-			validators: make(map[string]validation.PluginFactory),
+			endorsers:                make(map[string]endorsement2.PluginFactory),
+			validators:               make(map[string]validation.PluginFactory),
+			authenticationTimeWindow: c.AuthenticationTimeWindow,
 		}
 		reg.loadHandlers(c)
 	})
@@ -100,7 +103,9 @@ func (r *registry) evaluateModeAndLoad(c *HandlerConfig, handlerType HandlerType
 
 // loadCompiled loads a statically compiled handler
 func (r *registry) loadCompiled(handlerFactory string, handlerType HandlerType, extraArgs ...string) {
-	registryMD := reflect.ValueOf(&HandlerLibrary{})
+	registryMD := reflect.ValueOf(&HandlerLibrary{
+		authenticationTimeWindow: r.authenticationTimeWindow,
+	})
 
 	o := registryMD.MethodByName(handlerFactory)
 	if !o.IsValid() {
--- a/sampleconfig/core.yaml
+++ b/sampleconfig/core.yaml
@@ -313,6 +313,8 @@ peer:
     authentication:
         # the acceptable difference between the current server time and the
         # client's time as specified in a client request message
+        # this value is used for delivery service and
+        # endorsement service (if the authFilter is enabled)
         timewindow: 15m
 
     # Path on the file system where peer will store data (eg ledger). This
@@ -454,13 +456,11 @@ peer:
     #     library: /etc/hyperledger/fabric/plugin/escc.so
     handlers:
         authFilters:
-          -
-            name: DefaultAuth
-          -
-            name: ExpirationCheck    # This filter checks identity x509 certificate expiration
+          - name: DefaultAuth
+          - name: ExpirationCheck # This filter checks identity x509 certificate expiration
+          - name: TimeWindowCheck # This filter checks the timestamp of an proposal request with the peer.authentication.timewindow parameter from core.yaml
         decorators:
-          -
-            name: DefaultDecorator
+          - name: DefaultDecorator
         endorsers:
           escc:
             name: DefaultEndorsement
--- a/core/deliverservice/testdata/core.yaml
+++ b/core/deliverservice/testdata/core.yaml
@@ -441,13 +441,11 @@ peer:
   #     library: /etc/hyperledger/fabric/plugin/escc.so
   handlers:
     authFilters:
-      -
-        name: DefaultAuth
-      -
-        name: ExpirationCheck    # This filter checks identity x509 certificate expiration
+      - name: DefaultAuth
+      - name: ExpirationCheck # This filter checks identity x509 certificate expiration
+      - name: TimeWindowCheck # This filter checks Timestamp in TimeWindow
     decorators:
-      -
-        name: DefaultDecorator
+      - name: DefaultDecorator
     endorsers:
       escc:
         name: DefaultEndorsement
--- a/core/handlers/auth/filter/timewindow_test.go
+++ b/core/handlers/auth/filter/timewindow_test.go
@@ -1,0 +1,64 @@
+/*
+Copyright IBM Corp. All Rights Reserved.
+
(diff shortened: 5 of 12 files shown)


C0rWin:
thanks @pfi79, could you please add integration tests also?

pfi79:
> thanks @pfi79, could you please add integration tests also?

Okay, I will.

sampleconfig/core.yaml
```diff
@@ -454,13 +454,11 @@ peer:
     #     library: /etc/hyperledger/fabric/plugin/escc.so
     handlers:
         authFilters:
-          -
-            name: DefaultAuth
-          -
-            name: ExpirationCheck    # This filter checks identity x509 certificate expiration
+          - name: DefaultAuth
```
yacovm, line 457: While I'm fine with implementing the filters, I don't think we should activate them by default. Let's keep them commented out in the sampleconfig.

yacovm:
Do clients (SDKs, gateway pseudo-SDK) actually set the timestamp in the channel header? Let's be sure we're not breaking all clients with this PR :-)

sampleconfig/core.yaml
```diff
@@ -454,13 +454,11 @@ peer:
     #     library: /etc/hyperledger/fabric/plugin/escc.so
     handlers:
         authFilters:
-          -
-            name: DefaultAuth
```
yacovm, line 458: Looking at this, I'm not even sure we should have activated the expiration check by default.
So, I am not sure we should have this new check activated by default.

Perhaps let it be commented out?

core/handlers/library/registry.go
```diff
@@ -73,6 +75,8 @@ func InitRegistry(c Config) Registry {
 
 // loadHandlers loads the configured handlers
 func (r *registry) loadHandlers(c Config) {
+	r.authenticationTimeWindow = c.AuthenticationTimeWindow
```
yacovm, line 78: we don't have this for expiration check so why is this needed? what am i missing?

yacovm: I think that `InitRegistry()` is the right place to initialize it. Not here.

pfi79:
> Do clients (SDKs, gateway pseudo-SDK) actually set the timestamp in the channel header? Let's be sure we're not breaking all clients with this PR :-)

This is the time that we then get inside the chaincode using stub.GetTxTimestamp()

yacovm:
> > Do clients (SDKs, gateway pseudo-SDK) actually set the timestamp in the channel header? Let's be sure we're not breaking all clients with this PR :-)
> 
> This is the time that we then get inside the chaincode using stub.GetTxTimestamp()

Yeah I'm asking whether all client implementations correctly set it. We should check this, otherwise we're breaking the clients.

pfi79:
> > > Do clients (SDKs, gateway pseudo-SDK) actually set the timestamp in the channel header? Let's be sure we're not breaking all clients with this PR :-)
> > 
> > 
> > This is the time that we then get inside the chaincode using stub.GetTxTimestamp()
> 
> Yeah I'm asking whether all client implementations correctly set it. We should check this, otherwise we're breaking the clients.

Then I have a counter question.
What do you call "properly installed" and "improperly installed"?

core/handlers/library/registry.go
pfi79: fix

sampleconfig/core.yaml
pfi79: 1. The DefaultAuth and ExpirationCheck filters were enabled by default. I don't know how long it's been on.

2. I think that enabling filters by default makes the fabric safer for newcomers who are just starting to use and familiarize themselves with it.

3. Confident professionals can comment these filters consciously, realizing what it can lead to.

yacovm:
> > > > Do clients (SDKs, gateway pseudo-SDK) actually set the timestamp in the channel header? Let's be sure we're not breaking all clients with this PR :-)
> > > 
> > > 
> > > This is the time that we then get inside the chaincode using stub.GetTxTimestamp()
> > 
> > 
> > Yeah I'm asking whether all client implementations correctly set it. We should check this, otherwise we're breaking the clients.
> 
> Then I have a counter question. What do you call "properly installed" and "improperly installed"?

well for instance, a client may not initialize the timestamp field at all.

pfi79:
> well for instance, a client may not initialize the timestamp field at all.

That's great, I've never seen one of those.

Then they don't use stub.GetTxTimestamp() in their chaincode. No way, especially for selecting the code execution path (if....else).  
They can comment out this filter in core.yaml.

yacovm:
> > well for instance, a client may not initialize the timestamp field at all.
> 
> That's great, I've never seen one of those.
> 
> Then they don't use stub.GetTxTimestamp() in their chaincode. No way, especially for selecting the code execution path (if....else). They can comment out this filter in core.yaml.

No I mean the code that the client SDK executes, the code that is found in the official Fabric client SDK, not the code that the user implements as part of the solution. It doesn't matter if they use the chaincode stub API, as the proposal won't reach that far with the new filter. And the user and the deployer of the peer may be different people, so we should verify the client SDK code in multiple implementations sets this field.

pfi79:
> > > well for instance, a client may not initialize the timestamp field at all.
> > 
> > 
> > That's great, I've never seen one of those.
> > Then they don't use stub.GetTxTimestamp() in their chaincode. No way, especially for selecting the code execution path (if....else). They can comment out this filter in core.yaml.
> 
> No I mean the code that the client SDK executes, the code that is found in the official Fabric client SDK, not the code that the user implements as part of the solution. It doesn't matter if they use the chaincode stub API, as the proposal won't reach that far with the new filter. And the user and the deployer of the peer may be different people, so we should verify the client SDK code in multiple implementations sets this field.

I'll check

pfi79:
> > > well for instance, a client may not initialize the timestamp field at all.
> > 
> > 
> > That's great, I've never seen one of those.
> > Then they don't use stub.GetTxTimestamp() in their chaincode. No way, especially for selecting the code execution path (if....else). They can comment out this filter in core.yaml.
> 
> No I mean the code that the client SDK executes, the code that is found in the official Fabric client SDK, not the code that the user implements as part of the solution. It doesn't matter if they use the chaincode stub API, as the proposal won't reach that far with the new filter. And the user and the deployer of the peer may be different people, so we should verify the client SDK code in multiple implementations sets this field.

checked:
- fabric-gateway
- fabric-sdk-java
- fabric-sdk-node
- fabric-sdk-go
- fabric-admin-sdk

everyone fills in the timestamp field

Forgetting someone?

yacovm:
> > > > well for instance, a client may not initialize the timestamp field at all.
> > > 
> > > 
> > > That's great, I've never seen one of those.
> > > Then they don't use stub.GetTxTimestamp() in their chaincode. No way, especially for selecting the code execution path (if....else). They can comment out this filter in core.yaml.
> > 
> > 
> > No I mean the code that the client SDK executes, the code that is found in the official Fabric client SDK, not the code that the user implements as part of the solution. It doesn't matter if they use the chaincode stub API, as the proposal won't reach that far with the new filter. And the user and the deployer of the peer may be different people, so we should verify the client SDK code in multiple implementations sets this field.
> 
> checked:
> 
>     * fabric-gateway
> 
>     * fabric-sdk-java
> 
>     * fabric-sdk-node
> 
>     * fabric-sdk-go
> 
>     * fabric-admin-sdk
> 
> 
> everyone fills in the timestamp field
> 
> Forgetting someone?

Can you please link to the corresponding lines, if that's not too much trouble?

pfi79:
> > > > > well for instance, a client may not initialize the timestamp field at all.
> > > > 
> > > > 
> > > > That's great, I've never seen one of those.
> > > > Then they don't use stub.GetTxTimestamp() in their chaincode. No way, especially for selecting the code execution path (if....else). They can comment out this filter in core.yaml.
> > > 
> > > 
> > > No I mean the code that the client SDK executes, the code that is found in the official Fabric client SDK, not the code that the user implements as part of the solution. It doesn't matter if they use the chaincode stub API, as the proposal won't reach that far with the new filter. And the user and the deployer of the peer may be different people, so we should verify the client SDK code in multiple implementations sets this field.
> > 
> > 
> > checked:
> > ```
> > * fabric-gateway
> > 
> > * fabric-sdk-java
> > 
> > * fabric-sdk-node
> > 
> > * fabric-sdk-go
> > 
> > * fabric-admin-sdk
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > everyone fills in the timestamp field
> > Forgetting someone?
> 
> Can you please link to the corresponding lines, if that's not too much trouble?

we are interested in ChannelHeader only during the formation of Proposal
- fabric-gateway
[go](https://github.com/hyperledger/fabric-gateway/blob/a4aef4da2aa565ba296a76a0fff55083c4db8ebd/pkg/client/proposalbuilder.go#L118) 
[java](https://github.com/hyperledger/fabric-gateway/blob/a4aef4da2aa565ba296a76a0fff55083c4db8ebd/java/src/main/java/org/hyperledger/fabric/client/ProposalBuilder.java#L130)
[node](https://github.com/hyperledger/fabric-gateway/blob/a4aef4da2aa565ba296a76a0fff55083c4db8ebd/node/src/proposalbuilder.ts#L96)
- fabric-sdk-java
[this](https://github.com/hyperledger/fabric-sdk-java/blob/5a5259d77cf27a91f99407a5d3807ca6e664d618/src/main/java/org/hyperledger/fabric/sdk/transaction/ProposalBuilder.java#L151) [this](https://github.com/hyperledger/fabric-sdk-java/blob/5a5259d77cf27a91f99407a5d3807ca6e664d618/src/main/java/org/hyperledger/fabric/sdk/transaction/ProposalBuilder.java#L129) [this](https://github.com/hyperledger/fabric-sdk-java/blob/5a5259d77cf27a91f99407a5d3807ca6e664d618/src/main/java/org/hyperledger/fabric/sdk/transaction/ProposalBuilder.java#L130) and [this](https://github.com/hyperledger/fabric-sdk-java/blob/main/src/main/java/org/hyperledger/fabric/sdk/transaction/ProtoUtils.java#L94)
- fabric-sdk-node
[this](https://github.com/hyperledger/fabric-sdk-node/blob/fa3df2b69a7323fffe127d9560491257f8e097fe/fabric-common/lib/Proposal.js#L292) [this](https://github.com/hyperledger/fabric-sdk-node/blob/fa3df2b69a7323fffe127d9560491257f8e097fe/fabric-common/lib/Channel.js#L506)
- fabric-sdk-go
[this](https://github.com/hyperledger/fabric-sdk-go/blob/5d6ca326e01b057f4f1b3233e70a36c7623fae6f/internal/github.com/hyperledger/fabric/protoutil/proputils.go#L98)
- fabric-admin-sdk
[node1](https://github.com/hyperledger/fabric-admin-sdk/blob/f896a39f3e3e20ee003cb9fbd712a742bd97c3a8/ts/formatter/lib/proto/proposal-builder.ts#L66) [node2](https://github.com/hyperledger/fabric-admin-sdk/blob/f896a39f3e3e20ee003cb9fbd712a742bd97c3a8/ts/formatter/lib/proto/channel-builder.ts#L32)
[go](https://github.com/hyperledger/fabric-admin-sdk/blob/f896a39f3e3e20ee003cb9fbd712a742bd97c3a8/pkg/internal/proposal/builder.go#L121)

sampleconfig/core.yaml
yacovm: In that case I'd like to hear other people's thoughts on this. 

@C0rWin , or @denyeart  wdyt ?

sampleconfig/core.yaml
```diff
@@ -454,13 +454,11 @@ peer:
...
-          -
-            name: ExpirationCheck    # This filter checks identity x509 certificate expiration
+          - name: DefaultAuth
+          - name: ExpirationCheck # This filter checks identity x509 certificate expiration
+          - name: TimeWindowCheck # This filter checks Timestamp in TimeWindow
```
denyeart: Please add a more descriptive comment, mentioning that the client provided timestamp will be checked against the peer's timestamp, using the timewindow configured above in `peer.authentication.timewindow`.

Also, in the `peer.authentication.timewindow`config comment above, indicate that the timewindow value is used for delivery service and endorsement service (if the authFilter is enabled).  I can update the comment in release-2.5 branch to indicate that the timewindow was only used on delivery path in that release.

core/handlers/auth/filter/timewindow.go
```diff
@@ -0,0 +1,68 @@
...
+	chdr, err := protoutil.UnmarshalChannelHeader(hdr.ChannelHeader)
+	if err != nil {
+		return errors.Wrap(err, "failed parsing channel header")
+	}
+
+	timeProposal := chdr.Timestamp.AsTime().UTC()
+	now := time.Now().UTC()
+
+	if timeProposal.Add(timeWindow).Before(now) || timeProposal.Add(-timeWindow).After(now) {
+		return errors.New("request unauthorized due to incorrect timestamp")
```
denyeart: Please add a more descriptive error message, including the client timestamp, peer timestamp, and configured timewindow. Mention peer.authorization.timewindow property by name so that users can tie it back to the config in core.yaml.

sampleconfig/core.yaml
denyeart: I think this is a great feature for v3, thank you @pfi79.
Since it is a major release, I think it is completely acceptable to add a new check, as long as we remember to mention it in the release notes.

pfi79:
> thanks @pfi79, could you please add integration tests also?

added an integration test

also found that Proposal was bypassing auth filters locally via gateway
fixed

denyeart:
@pfi79 All pull requests should have some description text, even if it is short and sweet.

sampleconfig/core.yaml
pfi79: added

core/handlers/auth/filter/timewindow.go
pfi79: added

pfi79:
> @pfi79 All pull requests should have some description text, even if it is short and sweet.

added

sampleconfig/core.yaml
C0rWin: I think I do agree with @denyeart. This is a good feature as it will eliminate the issues with newcomers, and given v3.0 already brings some breaking changes, IMO,  this should not be a problem enabling it.

pfi79:
@C0rWin @yacovm @denyeart 
I think I've corrected all the comments. 
Please take a look and if I forgot something, please indicate it.

By the way, since yesterday the pipelines stopped working. 
Now there are 4 frozen pipelines

denyeart (approved):
Looks good to me, Iet's see if there are any other final comments...

C0rWin (approved):

C0rWin:
> Looks good to me, Iet's see if there are any other final comments...

I think we need to take to properly document it to make sure there are nor further confusions. I am good with the change.

PR #4958, opened 2024-08-21
Documentation - Remove references to v1.x legacy lifecycle

denyeart:
v1.x legacy lifecycle is removed in Fabric v3.
This commit removes all references from the documentation.

The lifecycle upgrade docs will be removed in a separate pull request when the overall upgrade docs are refactored for v3.

### DIFF
--- a/docs/source/cc_launcher.md
+++ b/docs/source/cc_launcher.md
@@ -202,7 +202,7 @@ If you do not need to fallback to the legacy Docker build process for your chain
 
 ## Chaincode packages
 
-As part of the new lifecycle introduced with Fabric 2.0, the chaincode package format changed from serialized protocol buffer messages to a gzip compressed POSIX tape archive. Chaincode packages created with `peer lifecycle chaincode package` use this new format.
+A chaincode package contains chaincode and associated metadata in a compressed gzip file that can be installed to a peer. The peer command `peer lifecycle chaincode package` can be used to create a chaincode package. You can also create a chaincode package using third party tools.
 
 ### Lifecycle chaincode package contents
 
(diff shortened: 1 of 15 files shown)


satota2 (commented):
@denyeart 
Thank you for the documentation update. I was wondering if we should also consider removing the following two pages:

- https://hyperledger-fabric.readthedocs.io/en/latest/enable_cc_lifecycle.html
- https://hyperledger-fabric.readthedocs.io/en/latest/upgrade_to_newest_version.html#upgrading-from-v1-4-x-release-to-a-v2-x-release (only the "Upgrading from v1.4.x release to a v2.x release" section)

Other than that, everything looks good to me!

denyeart:
> @denyeart Thank you for the documentation update. I was wondering if we should also consider removing the following two pages:
> 
> * https://hyperledger-fabric.readthedocs.io/en/latest/enable_cc_lifecycle.html
> * https://hyperledger-fabric.readthedocs.io/en/latest/upgrade_to_newest_version.html#upgrading-from-v1-4-x-release-to-a-v2-x-release (only the "Upgrading from v1.4.x release to a v2.x release" section)
> 
> Other than that, everything looks good to me!

You're right, those topics need to be updated as well. I intend to refactor the upgrade docs including removal of the v2.x lifecycle enablement topic in a subsequent pull request. Those topics will remain in release-2.5 docs but will be removed from main branch (v3) docs with the assumption that users have already done it.

andrew-coleman (approved):

satota2:
@denyeart Thank you for the response. Understood that these will be removed in the next PR.

Issue #3650, opened 2022-09-27
Use new (non-deprecated) protocol buffer bindings


bestbeforetoday:
**As a** user of Fabric
**I want** Fabric v3 to use fabric-protos Go bindings based on protocol buffer APIv2
**So that** Fabric is using actively maintained and non-deprecated protobuf APIs

In early 2020, a new Go API for protocol buffers was released, and the older protocol buffer API was deprecated.

- https://go.dev/blog/protobuf-apiv2
- Deprecated API: https://pkg.go.dev/github.com/golang/protobuf
- New API: https://pkg.go.dev/google.golang.org/protobuf

The two API versions produce 100% wire-level compatible protobuf messages so they can be used interchangeably with no compatibility implications for client applications, SDKs or other Fabric components. This has already been proven in the [Fabric Gateway client API](https://github.com/hyperledger/fabric-gateway), which migrated seamlessly from the old to new protocol buffer API and maintained complete compatibility with Fabric.

Pre-built Go language bindings for the Fabric protobufs are already published based on both the old ([fabric-protos-go](https://github.com/hyperledger/fabric-protos-go)) and new ([fabric-protos-go-apiv2](https://github.com/hyperledger/fabric-protos-go-apiv2)). The only work required is to change Fabric code to use the new protos package.

denyeart:
Per https://github.com/hyperledger/fabric-protos/blob/main/RELEASING.md fabric-protos-go-apiv2 v0.3.0 is ready for use by fabric main branch.

bestbeforetoday:
I briefly looked at this and it seems more involved than I'd hoped, particularly since migrating fabric-gateway was very easy (see [commit](https://github.com/hyperledger/fabric-gateway/commit/fbfcc708e3702046d1a3a78e30d4754666876f04)). Other Fabric packages like [fabric-config](http://github.com/hyperledger/fabric-config) expose the older/deprecated protobuf package in their public API, which is a headache and likely a breaking change.

Since the old protobuf package got reimplemented using the newer protobuf package internally, you can translate between V1 and V2 protobuf Message structs, which might allow at least some changes to be made without breaking any public APIs. See this utility that used to be in fabric-gateway before it was migrated over entirely, which provided a few utility functions that can operate on both V1 and V2 messages by using the GeneratedMessage interface:

[...]

PR #4962, opened 2024-08-23
delete shadow copy proto-struct

pfi79:
The new fabric-protos-go-apiv2 has copy protection in the structures.
This code is adapted for the transition to the new protos-go-apiv2.


denyeart:
There is an [open issue](https://github.com/hyperledger/fabric/issues/3650) discussing potential update to protos-go-apiv2. Due to some concerns that were raised in that issue, my [latest thinking](https://github.com/hyperledger/fabric/issues/3650#issuecomment-2161393753) was to not proceed with the update. Let's continue the discussion in that issue prior to moving forward with the PRs to update to protos-go-apiv2.

pfi79:
> There is an [open issue](https://github.com/hyperledger/fabric/issues/3650) discussing potential update to protos-go-apiv2. Due to some concerns that were raised in that issue, my [latest thinking](https://github.com/hyperledger/fabric/issues/3650#issuecomment-2161393753) was to not proceed with the update. Let's continue the discussion in that issue prior to moving forward with the PRs to update to protos-go-apiv2.

Yes, it's a move towards protos-go-apiv2. But independent of it. No new protos, no new protobuf.

I specifically highlighted this code, which is independent of protos-go-apiv2. 

I would suggest that it should be considered independent.

denyeart:
> > There is an [open issue](https://github.com/hyperledger/fabric/issues/3650) discussing potential update to protos-go-apiv2. Due to some concerns that were raised in that issue, my [latest thinking](https://github.com/hyperledger/fabric/issues/3650#issuecomment-2161393753) was to not proceed with the update. Let's continue the discussion in that issue prior to moving forward with the PRs to update to protos-go-apiv2.
> 
> Yes, it's a move towards protos-go-apiv2. But independent of it. No new protos, no new protobuf.
> 
> I specifically highlighted this code, which is independent of protos-go-apiv2.
> 
> I would suggest that it should be considered independent.

Yes, it can be considered independent. But in that case the PR description should mention the rationale for the various changes. For example adding proto.Clone() in various places seems to add a dependency where none was required previously. Can you describe the benefits? And some of the other changes seem not related, ideally they would be done in a separate pull request or at least mention the rationale for the changes in the PR description.

pfi79:
> Yes, it can be considered independent. But in that case the PR description should mention the rationale for the various changes. For example adding proto.Clone() in various places seems to add a dependency where none was required previously. Can you describe the benefits? And some of the other changes seem not related, ideally they would be done in a separate pull request or at least mention the rationale for the changes in the PR description.

I think the developers of the protobuf package added shadow copy protection to the new structures for a reason. 
It means that they had good reasons for it. 
And if we can't switch to a new version of protobuf yet, we can at least remove shadow copying in them. And in those places where copying is required make it more explicit for further code maintenance.

But if you don't consider this an important reason, let's wait until I prepare a version with all the changes and present it in the corresponding issue.

bestbeforetoday:
See this comment describing why it is not safe to make shallow copies of Go protobuf messages, which explains why they are referred to in the protobuf API only as pointers (`proto.Message` interface):

https://github.com/golang/protobuf/issues/1225#issuecomment-713699536

core/chaincode/handler_test.go
```diff
@@ -2550,7 +2550,7 @@ var _ = Describe("Handler", func() {
 		})
 
 		It("sends an execute message to the chaincode with the correct proposal", func() {
-			expectedMessage := *incomingMessage
+			expectedMessage := proto.Clone(incomingMessage).(*pb.ChaincodeMessage)
```
bestbeforetoday: I think the answer here is just to reference the protobuf message as a pointer rather than dereferencing it and causing a copy-by-value. I'm not sure there is any need to clone the message

core/chaincode/handler_test.go
```diff
@@ -2559,7 +2559,7 @@ var _ = Describe("Handler", func() {
 			Eventually(fakeChatStream.SendCallCount).Should(Equal(1))
 			Consistently(fakeChatStream.SendCallCount).Should(Equal(1))
 			msg := fakeChatStream.SendArgsForCall(0)
-			Expect(msg).To(Equal(&expectedMessage))
+			Expect(msg).To(Equal(expectedMessage))
```
bestbeforetoday: The standard Gomega `Equal` matcher is not safe for use on protobuf messages at all. It simply does a deep equal comparison, and this is not valid for protobuf messages since they can be logically equal while also having some implementation fields with different values.

Some other parts of the codebase use `proto.Equal` for protobuf message comparison, and then use Gomega to _Expect_ a True return value. I think a better approach is to implement a matcher specifically for protobuf messages, for example:

```go
func ProtoEqual(expected proto.Message) types.GomegaMatcher {
	return &protoEqualMatcher{
		expected: expected,
	}
}

type protoEqualMatcher struct {
	expected proto.Message
}

func (m *protoEqualMatcher) Match(actual interface{}) (bool, error) {
	actualMessage, ok := actual.(proto.Message)
	if !ok {
		return false, fmt.Errorf("ProtoEqual expects a proto.Message, got a %T", actual)
	}

	return proto.Equal(m.expected, actualMessage), nil
}

func (m *protoEqualMatcher) FailureMessage(actual interface{}) string {
	if message, ok := actual.(proto.Message); ok {
		return fmt.Sprintf("Expected\n\t%s\nto equal\n\t%s", message, m.expected)
	}

	return fmt.Sprintf("Expected\n\t%#v\nto equal\n\t%s", actual, m.expected)
}

func (m *protoEqualMatcher) NegatedFailureMessage(actual interface{}) string {
	if message, ok := actual.(proto.Message); ok {
		return fmt.Sprintf("Expected\n\t%s\nnot to equal\n\t%s", message, m.expected)
	}

	return fmt.Sprintf("Expected\n\t%#v\nnot to equal\n\t%s", actual, m.expected)
}
```

This is then used as follows:
```go
Expect(actualMsg).To(ProtoEqual(expectedMsg))
```

bestbeforetoday (changes requested):

core/chaincode/handler_test.go
pfi79: fix

core/chaincode/handler_test.go
pfi79: fix

PR #4963, opened 2024-08-23
Added new proto files

pfi79:
Added new proto files to replace structures that were manually made to look like proto. 
These will be difficult to maintain in the future.
Аdded a test (ccproviderproto_test.go) that shows they have the same result.


pfi79:
https://github.com/hyperledger/fabric/issues/3650

denyeart (approved):

PR #4964, opened 2024-08-24
replacing old functions that are not in the new protobuf package

pfi79:
preparing to install the new protobuf package


pfi79:
https://github.com/hyperledger/fabric/issues/3650

PR #4974, opened 2024-09-10
Bump Go to 1.23.1

denyeart:
Bump Go to 1.23.1.
Also update go.mod to use three digit version to indicate toolchain version (most Go projects have switched to this approach at this point).


go.mod
```diff
@@ -1,6 +1,6 @@
 module github.com/hyperledger/fabric
 
-go 1.23
+go 1.23.1
```
pfi79, line 3: Why didn't it work to keep the 2-digit version? 1.23

denyeart: > Why didn't it work to keep the 2-digit version? 1.23

It did work with the 2-digit version. However all the Go documentation says to use a 3-digit version for the toolchain. Since we don't have a different toolchain version listed, it will use the Go directive version instead. Therefore to be aligned with the documentation it is better to switch to the 3-digit version so that nothing unexpected happens. I realize I was reluctant to make this move earlier, but most of the Go community is moving to this approach and I'm now on board.

pfi79 (approved):

denyeart:
@satota2 Note that broken link checker is expected to fail as there are still some old doc issues. It only runs when the job file is updated, and is not a required check for merge.

satota2 (approved):
@denyeart 
Thanks for letting me know. I'll go ahead and ignore the failure and proceed with the merge.

PR #4984, opened 2024-09-16
Utilize 2.5 images for java chaincode and node chaincode

denyeart:
Java chaincode image javaenv and
Node chaincode image nodeenv
will continue to be maintained at version 2.5
instead of creating versions 3.0 of them.

Therefore, instead of using the current Fabric TWO_DIGIT_VERSION for these chaincodes,
hardcode them at 2.5 in the default sample config.

### DIFF
--- a/core/chaincode/chaincodetest.yaml
+++ b/core/chaincode/chaincodetest.yaml
@@ -294,7 +294,8 @@ chaincode:
         path:
         name:
 
-    # Generic builder environment, suitable for most chaincode types
+    # Generic builder image with Go pre-installed.
+    # TWO_DIGIT_VERSION represents Fabric major.minor version.
     builder: $(DOCKER_NS)/fabric-ccenv:$(TWO_DIGIT_VERSION)
 
     # Enables/disables force pulling of the base docker images (listed below)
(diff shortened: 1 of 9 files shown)


core/chaincode/chaincodetest.yaml
```diff
@@ -303,22 +304,23 @@ chaincode:
     pull: false
 
     golang:
-        # golang will never need more than baseos
+        # Compiled Go chaincodes will be copied to the fabric-baseos runtime image.
+        # TWO_DIGIT_VERSION represents Fabric major.minor version.
```
yacovm, line 308: why not do that for Golang too? The state of Golang chaincode is the same as the others :-)

yacovm (dismissed):

core/chaincode/chaincodetest.yaml
denyeart: Go chaincode's fabric-ccenv and fabric-baseos images are built and published alongside fabric in this repository, might as well use the most recent images with latest Go.

satota2:
@denyeart I have a few minor points that caught my attention, so I'd like to leave the following comments:
- This is a minor point, but I believe it would be better to remove javaenv and nodeenv from the following two lines when they do not match TWO_DIGIT_VERSION:
  - https://github.com/hyperledger/fabric/blob/main/scripts/install-fabric.sh#L147
  - https://github.com/hyperledger/fabric/blob/main/scripts/bootstrap.sh#L45
- While it might be obvious from the default sample config, I think it would be more user-friendly to explicitly mention somewhere in the documentation that the "Node chaincode image nodeenv will continue to be maintained at version 2.5 as the latest version." What do you think?
- In fabric-samples, both the full-stack-asset-transfer-guide and test-network-k8s use javaenv and/or nodeenv. Should these also be hardcoded to version 2.5?

denyeart:
> @denyeart I have a few minor points that caught my attention, so I'd like to leave the following comments:
> 
> * This is a minor point, but I believe it would be better to remove javaenv and nodeenv from the following two lines when they do not match TWO_DIGIT_VERSION:
>   
>   * https://github.com/hyperledger/fabric/blob/main/scripts/install-fabric.sh#L147
>   * https://github.com/hyperledger/fabric/blob/main/scripts/bootstrap.sh#L45
> * While it might be obvious from the default sample config, I think it would be more user-friendly to explicitly mention somewhere in the documentation that the "Node chaincode image nodeenv will continue to be maintained at version 2.5 as the latest version." What do you think?
> * In fabric-samples, both the full-stack-asset-transfer-guide and test-network-k8s use javaenv and/or nodeenv. Should these also be hardcoded to version 2.5?

Thank you @satota2 , I've removed mention of javaenv and nodeenv from the install scripts since those scripts don't pull the javaenv and nodeenv images.

For the documentation, I've realized that we don't describe the default chaincode builder process anywhere in the docs. I've updated the chaincode lifecycle topic to at least mention the docker images and refer readers to the config files for more details so that we don't have to include version information into the docs which may get outdated.

For the samples, they still utilize Fabric v2.5, but I will soon update the samples to support both 2.5 and 3.0 and will handle the chaincode image versions accordingly.

satota2 (approved):
@denyeart Thank you for the updates, including the changes to the documentation. The changes look good to me!
I also understand the plan for updating fabric-samples. Thanks for the clarification.

PR #4987, opened 2024-09-18
Create home directory in docker images

denyeart:
The recent change to use `useradd -M` in baseos and ccenv Docker images caused building of Go chaincodes to fail with error:

```
"failed to initialize build cache at /home/chaincode/.cache/go-build: mkdir /home/chaincode: permission denied"
```

Changing it back to allow creation of the chaincode home directory with `useradd -m` solves the problem.


satota2 (approved):
@denyeart LGTM!

PR #4993, opened 2024-09-24
Fix argument format for calculatepackageid command

satota2:
#### Type of change

- Bug fix
- Documentation update

#### Description

This patch updates the argument format for `calculatepackageid` command, enclosing `packageFile` in square brankets to indicate that the specific file name should be put.


andrew-coleman (approved):

bestbeforetoday:
Square brackets typically indicate that a parameter is optional. Is this the case here? Angle brackets are typically used for required parameters, for example: `go mod <command> [arguments]`

PR #4996, opened 2024-09-28
change the for loop to a new for range - part 1

pfi79:
I suggest replacing the construction: 
`for i:= ...; i < ...; i++`
by the new construction:
`for range ...`
or 
`for i := range ... `

there are a lot of changes, but I'll break them down into a few pr

### DIFF
--- a/cmd/cryptogen/main.go
+++ b/cmd/cryptogen/main.go
@@ -344,9 +344,9 @@ func extendPeerOrg(orgSpec OrgSpec) {
 	publicKeyAlg := getPublicKeyAlg(orgSpec.Users.PublicKeyAlgorithm)
 	// TODO: add ability to specify usernames
 	users := []NodeSpec{}
-	for j := 1; j <= orgSpec.Users.Count; j++ {
+	for j := range orgSpec.Users.Count {
 		user := NodeSpec{
-			CommonName:         fmt.Sprintf("%s%d@%s", userBaseName, j, orgName),
+			CommonName:         fmt.Sprintf("%s%d@%s", userBaseName, j+1, orgName),
 			PublicKeyAlgorithm: publicKeyAlg,
 		}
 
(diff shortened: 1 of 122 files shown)


denyeart:
Although Go 1.23 expanded where you can use `range`, that doesn't necessarily mean we should change every for loop throughout the existing code.

For simple counters I think the for loop is fine and maybe even more readable by most developers. Plus, changing code unnecessarily in the main branch will make backports more difficult. And then there is also risk of an "off by 1" mistake when making bulk changes in the code.

Is there a compelling reason to make the change? Maybe we should only update occurrences where there is a compelling need, or if we are updating a section of code regardless.

pfi79:
> Although Go 1.23 expanded where you can use `range`, that doesn't necessarily mean we should change every for loop throughout the existing code.
> 
> For simple counters I think the for loop is fine and maybe even more readable by most developers. Plus, changing code unnecessarily in the main branch will make backports more difficult. And then there is also risk of an "off by 1" mistake when making bulk changes in the code.
> 
> Is there a compelling reason to make the change? Maybe we should only update occurrences where there is a compelling need, or if we are updating a section of code regardless.

There's no good reason.

There is no uniformity in the `for ; ;i++` construct. 
That's why they invented a substitute for it, in which it is harder to make a mistake. So I decided to make it monotonous.

You may not accept my suggestion and just close it.

But if you decide to move in this direction, I can close this pr and create smaller ones to make it easier for the checkers to track down the “off by 1” error.

denyeart:
> > Although Go 1.23 expanded where you can use `range`, that doesn't necessarily mean we should change every for loop throughout the existing code.
> > For simple counters I think the for loop is fine and maybe even more readable by most developers. Plus, changing code unnecessarily in the main branch will make backports more difficult. And then there is also risk of an "off by 1" mistake when making bulk changes in the code.
> > Is there a compelling reason to make the change? Maybe we should only update occurrences where there is a compelling need, or if we are updating a section of code regardless.
> 
> There's no good reason.
> 
> There is no uniformity in the `for ; ;i++` construct. That's why they invented a substitute for it, in which it is harder to make a mistake. So I decided to make it monotonous.
> 
> You may not accept my suggestion and just close it.
> 
> But if you decide to move in this direction, I can close this pr and create smaller ones to make it easier for the checkers to track down the “off by 1” error.

My opinion is to not make blanket changes in the code base for the purpose of syntax preferences. Different developers may have different preferences. If there are generally accepted bad Go practices in the code base then I agree those should be changed, but I don't think this one meets that criteria.

Let's keep it open for another day to see if there are any other thoughts from other maintainers.

yacovm:
> Let's keep it open for another day to see if there are any other thoughts from other maintainers.

Yeah I second @denyeart .  

If it ain't broke, don't fix it :-)

PR #5009, opened 2024-10-07
changing the order in which integration tests

pfi79:
changing the order in which integration tests are executed to reduce the total testing time
24 min vs 19 min


denyeart (approved):
Thanks! I've changed the branch protection rules accordingly.

PR #5014, opened 2024-10-07
change logger for unit test smartbft

pfi79:
There are unit tests in smartbft, where some logs are written via logger and some via testing.Logf.

As a result, when the test terminates with an error. Some of the sequential logs are in one place, some in another.

I decided to bring the logs to a single view through the logger. In order to deal with the floating error in the future

### DIFF
--- a/orderer/consensus/smartbft/chain_test.go
+++ b/orderer/consensus/smartbft/chain_test.go
@@ -12,6 +12,7 @@ import (
 	"testing"
 	"time"
 
+	"github.com/hyperledger/fabric-lib-go/common/flogging"
 	cb "github.com/hyperledger/fabric-protos-go-apiv2/common"
 	"github.com/hyperledger/fabric-protos-go-apiv2/msp"
 	"github.com/hyperledger/fabric/common/channelconfig"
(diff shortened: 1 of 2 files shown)


orderer/consensus/smartbft/chain_test.go
```diff
@@ -358,6 +359,9 @@ func TestAddAndRemoveNodeWithoutStop(t *testing.T) {
 	dir := t.TempDir()
 	channelId := "testchannel"
 
+	flogging.ActivateSpec("debug")
```
yacovm, line 362: why do we need this?

pfi79: The default logs seem to be Info.
In case of failure I would like to see debug logs from the smartbft library

yacovm: Why? the tests aren't flaky in CI

orderer/consensus/smartbft/util_network_test.go
```diff
@@ -64,6 +64,7 @@ type NetworkSetupInfo struct {
 	genesisBlock *cb.Block
 	tlsCA        tlsgen.CA
 	configInfo   *ConfigInfo
+	Logger       *flogging.FabricLogger
```
yacovm, line 67: as a general rule of thumb - we should not couple this logger field with a specific implementation, but with an interface.

yacovm (approved):

orderer/consensus/smartbft/chain_test.go
pfi79: https://github.com/hyperledger/fabric/blob/main/orderer/consensus/smartbft/synchronizer_bft_test.go#L43

like what's done here
and there's more than one place like this.

pfi79: Right. The TestAddAndRemoveNodeWithoutStop test is not stable. Quite often it terminates with an error. I can't catch the error and fix it yet.

PR #5019, opened 2024-10-08
bump github.com/syndtr/goleveldb to 126854af5e6d

pfi79:


common/ledger/util/leveldbhelper/leveldb_helper.go
```diff
@@ -107,14 +107,16 @@ func (dbInst *DB) Get(key []byte) ([]byte, error) {
...
+	if errors.Is(err, leveldb.ErrNotFound) {
+		return nil, nil
 	}
 	if err != nil {
 		logger.Errorf("Error retrieving leveldb key [%#v]: %s", key, err)
 		return nil, errors.Wrapf(err, "error retrieving leveldb key [%#v]", key)
 	}
+	if value == nil {
+		value = []byte{}
+	}
```
denyeart, line 123: Why is this change needed? Doesn't it change the behavior in a way that may impact callers?

pfi79: Without this change, everything stops working.

Previously the goleveldb library returned []byte{}
now it returns []byte(nil)

And in the fabric returning nil it is considered that no key is found in the database
Instead of finding and fixing all the places where this behavior is used, I simulated our old behavior.

I should probably add a comment

pfi79: https://github.com/syndtr/goleveldb/commit/75fe51978d5f82f9b62d1025e1d061cb2008c38b

here's the commit that made this change

denyeart: Thanks @pfi79 , the original author @manish-sethi will review this one.

manish-sethi: LGTM, given the change in leveldb behavior.

denyeart (approved):

PR #5023, opened 2024-10-10
bump github.com/alecthomas/kingpin/v2 to v2.4.0

pfi79:


docs/source/commands/configtxlator.md
```diff
@@ -26,8 +26,9 @@ usage: configtxlator start [<flags>]
 
 Start the configtxlator REST server
 
+
 Flags:
-  --help                Show context-sensitive help (also try --help-long and
+  --[no-]help           Show context-sensitive help (also try --help-long and
```
denyeart, line 31: What is no-help?

pfi79: I don't know, after upgrading the library. this prompt is now generated

pfi79: image
As I understand it, this is done to explicitly set some parameter to false

pfi79: https://github.com/alecthomas/kingpin/commit/32660ac033599f364e8cc2dc8c97ebcb12e839f1#diff-5150d3041d43020ac2b39e44af9b692972d5627a5777782e52a0e2094c420af9R80

modified here

pfi79:
@denyeart The author of this library is weakly supportive, which is what he wrote [about](https://github.com/alecthomas/kingpin). He himself switched to the [kong](https://github.com/alecthomas/kong) library.

Perhaps we should get rid of it in fabric. Should I switch to cobra and viper, as is done in peer, or to the same [kong](https://github.com/alecthomas/kong)?

denyeart:
> @denyeart The author of this library is weakly supportive, which is what he wrote [about](https://github.com/alecthomas/kingpin). He himself switched to the [kong](https://github.com/alecthomas/kong) library.
> 
> Perhaps we should get rid of it in fabric. Should I switch to cobra and viper, as is done in peer, or to the same [kong](https://github.com/alecthomas/kong)?

From my memory kingpin was simpler than cobra, so I'd be ok leaving it on the older kingpin (no update) to keep things simple. If you really want to invest your time in an update to kong or cobra then I'd be fine with it, up to you.

pfi79:
> > @denyeart The author of this library is weakly supportive, which is what he wrote [about](https://github.com/alecthomas/kingpin). He himself switched to the [kong](https://github.com/alecthomas/kong) library.
> > Perhaps we should get rid of it in fabric. Should I switch to cobra and viper, as is done in peer, or to the same [kong](https://github.com/alecthomas/kong)?
> 
> From my memory kingpin was simpler than cobra, so I'd be ok leaving it on the older kingpin (no update) to keep things simple. If you really want to invest your time in an update to kong or cobra then I'd be fine with it, up to you.

Okay, maybe sometime in the future.

pfi79:
> I'd be ok leaving it on the older kingpin (no update) to keep things simple

then I'm closing the pr?

denyeart:
Yes, let's close the PR. The command utilities have been very stable, and the `no-help` thing is confusing, let's keep current dependency.

PR #5024, opened 2024-10-11
fixed an error in use smartbft

pfi79:
fixed an error after deleting a node when it participated in block signing immediately before deletion

Recently the [TestAddAndRemoveNodeWithoutStop](https://github.com/hyperledger/fabric/blob/main/orderer/consensus/smartbft/chain_test.go#L358) test has been terminating with an error very often in unit tests.

It went like this:
- [prepared](https://github.com/hyperledger/fabric/blob/main/orderer/consensus/smartbft/chain_test.go#L413) a new config with node 5 removed from the cluster
- [send](https://github.com/hyperledger/fabric/blob/main/orderer/consensus/smartbft/chain_test.go#L414) it to the orderers
- The 5th node participated in the consensus to accept this block
- [deleted](https://github.com/hyperledger/fabric/blob/main/orderer/consensus/smartbft/chain_test.go#L424) the 5th node
- [sent](https://github.com/hyperledger/fabric/blob/main/orderer/consensus/smartbft/chain_test.go#L430) a simple transaction to check that everything was working.
- In the very first stage of consensus, the smartBFT library takes the previous commit with signatures and checks for correctness. It sends it to the fabric function [VerifyConsenterSig](https://github.com/hyperledger/fabric/blob/main/orderer/consensus/smartbft/verifier.go#L211). In the previous commit the signature of node 5 is present, but in the current config it is not. This causes an error [here](https://github.com/hyperledger/fabric/blob/main/orderer/consensus/smartbft/verifier.go#L216).

My suggestion for a fix: 
- keep the previous set of nodes in memory
- if we don't find a signature in the current set and the block is equal and/or less than the current block number, check in the previous set

Perhaps there is a better and more beautiful solution.

### DIFF
--- a/orderer/consensus/smartbft/assembler.go
+++ b/orderer/consensus/smartbft/assembler.go
@@ -11,11 +11,14 @@ import (
 	"sync/atomic"
 
 	"github.com/hyperledger-labs/SmartBFT/pkg/types"
+	"github.com/hyperledger/fabric-lib-go/bccsp"
 	"github.com/hyperledger/fabric-lib-go/common/flogging"
 	cb "github.com/hyperledger/fabric-protos-go-apiv2/common"
+	"github.com/hyperledger/fabric/common/channelconfig"
 	"github.com/hyperledger/fabric/orderer/common/cluster"
 	"github.com/hyperledger/fabric/protoutil"
 	"github.com/pkg/errors"
+	"google.golang.org/protobuf/proto"
 )
 
 //go:generate mockery -dir . -name Ledger -case underscore -output mocks
@@ -138,28 +141,52 @@ func lastConfigBlockFromLedger(ledger Ledger) (*cb.Block, error) {
 	return lastConfigBlock, nil
 }
 
-func PreviousConfigBlockFromLedgerOrPanic(ledger Ledger, logger Logger) *cb.Block {
-	block, err := previousConfigBlockFromLedger(ledger)
+func NextToLastConfigBlockFromLedgerOrPanic(ledger Ledger, bccsp bccsp.BCCSP, logger Logger) *cb.Block {
+	block, err := nextToLastConfigBlockFromLedger(ledger, bccsp)
 	if err != nil {
 		logger.Panicf("Failed retrieving previous config block: %v", err)
 	}
 	return block
 }
 
-func previousConfigBlockFromLedger(ledger Ledger) (*cb.Block, error) {
-	previousBlockSeq := ledger.Height() - 2
-	if ledger.Height() == 1 {
-		previousBlockSeq = 0
+func nextToLastConfigBlockFromLedger(ledger Ledger, bccsp bccsp.BCCSP) (*cb.Block, error) {
+	block, err := lastConfigBlockFromLedger(ledger)
+	if err != nil {
+		return nil, err
 	}
+	if block.Header.Number == 0 {
+		return nil, nil
+	}
+
+	previousBlockSeq := block.Header.Number - 1
 	previousBlock := ledger.Block(previousBlockSeq)
 	if previousBlock == nil {
 		return nil, errors.Errorf("unable to retrieve block [%d]", previousBlockSeq)
 	}
-	previousConfigBlock, err := cluster.LastConfigBlock(previousBlock, ledger)
+	previousLastConfigBlock, err := cluster.LastConfigBlock(previousBlock, ledger)
 	if err != nil {
 		return nil, err
 	}
-	return previousConfigBlock, nil
+
+	env := &cb.Envelope{}
+	if err = proto.Unmarshal(previousLastConfigBlock.Data.Data[0], env); err != nil {
+		return nil, errors.Wrap(err, "failed unmarshaling envelope of config block")
+	}
+	bundle, err := channelconfig.NewBundleFromEnvelope(env, bccsp)
+	if err != nil {
+		return nil, errors.Wrap(err, "failed getting a new bundle from envelope of config block")
+	}
+
+	oc, ok := bundle.OrdererConfig()
+	if !ok {
+		return nil, errors.New("no orderer config in config block")
+	}
+
+	if oc.ConsensusType() != "BFT" {
+		return nil, nil
+	}
+
+	return previousLastConfigBlock, nil
 }
 
 // LastBlockFromLedgerOrPanic returns the last block from the ledger
--- a/orderer/consensus/smartbft/chain.go
+++ b/orderer/consensus/smartbft/chain.go
@@ -141,12 +141,20 @@ func NewChain(
 
 	lastBlock := LastBlockFromLedgerOrPanic(support, c.Logger)
 	lastConfigBlock := LastConfigBlockFromLedgerOrPanic(support, c.Logger)
+	nextToLastConfigBlock := NextToLastConfigBlockFromLedgerOrPanic(support, bccsp, c.Logger)
 
+	var err error
 	rtc := RuntimeConfig{
 		logger: logger,
 		id:     selfID,
 	}
-	rtc, err := rtc.BlockCommitted(lastConfigBlock, bccsp)
+	if nextToLastConfigBlock != nil {
+		rtc, err = rtc.BlockCommitted(nextToLastConfigBlock, bccsp)
+		if err != nil {
+			return nil, errors.Wrap(err, "failed constructing RuntimeConfig")
+		}
+	}
+	rtc, err = rtc.BlockCommitted(lastConfigBlock, bccsp)
 	if err != nil {
 		return nil, errors.Wrap(err, "failed constructing RuntimeConfig")
 	}
--- a/orderer/consensus/smartbft/util.go
+++ b/orderer/consensus/smartbft/util.go
@@ -46,6 +46,7 @@ type RuntimeConfig struct {
 	LastCommittedBlockHash string
 	RemoteNodes            []cluster.RemoteNode
 	ID2Identities          NodeIdentitiesByID
+	ID2IdentitiesPrev      NodeIdentitiesByID
 	LastBlock              *cb.Block
 	LastConfigBlock        *cb.Block
 	Nodes                  []uint64
@@ -65,6 +66,7 @@ func (rtc RuntimeConfig) BlockCommitted(block *cb.Block, bccsp bccsp.BCCSP) (Run
 		LastCommittedBlockHash: hex.EncodeToString(protoutil.BlockHeaderHash(block.Header)),
 		Nodes:                  rtc.Nodes,
 		ID2Identities:          rtc.ID2Identities,
+		ID2IdentitiesPrev:      rtc.ID2IdentitiesPrev,
 		RemoteNodes:            rtc.RemoteNodes,
 		LastBlock:              block,
 		LastConfigBlock:        rtc.LastConfigBlock,
@@ -91,6 +93,7 @@ func (rtc RuntimeConfig) configBlockCommitted(block *cb.Block, bccsp bccsp.BCCSP
 		LastCommittedBlockHash: hex.EncodeToString(protoutil.BlockHeaderHash(block.Header)),
 		Nodes:                  nodeConf.nodeIDs,
 		ID2Identities:          nodeConf.id2Identities,
+		ID2IdentitiesPrev:      rtc.ID2Identities,
 		RemoteNodes:            nodeConf.remoteNodes,
 		LastBlock:              block,
 		LastConfigBlock:        block,
--- a/orderer/consensus/smartbft/verifier.go
+++ b/orderer/consensus/smartbft/verifier.go
@@ -8,6 +8,7 @@ package smartbft
 
 import (
 	"bytes"
+	"encoding/asn1"
 	"encoding/base64"
 	"encoding/hex"
 	"fmt"
@@ -209,11 +210,30 @@ func (v *Verifier) verifyRequest(rawRequest []byte, noConfigAllowed bool) (types
 
 // VerifyConsenterSig verifies consenter signature
 func (v *Verifier) VerifyConsenterSig(signature types.Signature, prop types.Proposal) ([]byte, error) {
-	id2Identity := v.RuntimeConfig.Load().(RuntimeConfig).ID2Identities
+	rtc := v.RuntimeConfig.Load().(RuntimeConfig)
 
-	identity, exists := id2Identity[signature.ID]
+	var (
+		identity []byte
+		exists   bool
+	)
+	identity, exists = rtc.ID2Identities[signature.ID]
 	if !exists {
-		return nil, errors.Errorf("node with id of %d doesn't exist", signature.ID)
+		if len(prop.Header) == 0 {
+			return nil, errors.New("proposal header cannot be nil")
+		}
+		hdr := &asn1Header{}
+		if _, err := asn1.Unmarshal(prop.Header, hdr); err != nil {
+			return nil, errors.Wrap(err, "bad header")
+		}
+		numBlockProp := hdr.Number.Uint64()
+		if numBlockProp <= rtc.LastBlock.GetHeader().GetNumber() {
+			identity, exists = rtc.ID2IdentitiesPrev[signature.ID]
+			if !exists {
+				return nil, errors.Errorf("node with id of %d doesn't exist", signature.ID)
+			}
+		} else {
+			return nil, errors.Errorf("node with id of %d doesn't exist", signature.ID)
+		}
 	}
 
 	sig := &Signature{}


yacovm:
> In the very first stage of consensus, the smartBFT library takes the previous commit with signatures and checks for correctness.

If I remember correctly, we only add the previous signatures, [when we have leader election enabled](https://github.com/hyperledger-labs/SmartBFT/blob/main/internal/bft/view.go#L977), and in Fabric we don't have leader election enabled. If the leader doesn't send the previous signatures, then the followers won't verify them. 

However in the tests, [the decisions per leader is 0](https://github.com/hyperledger/fabric/blob/main/orderer/consensus/smartbft/util_network_test.go#L762)... it is therefore odd that the previous signatures verification is the cause this behavior.

Can you perhaps add a stack trace dump in the [place](https://github.com/hyperledger-labs/SmartBFT/blob/main/internal/bft/view.go#L631) where we verify the consenter signature? 

Just to give some background - we introduced the previous commit signatures in order to know how to prune the blacklist. However, we don't use leader rotation nor the blacklist in Fabric, so before we go ahead and fix something, let's be 100% sure of it's cause. 


> My suggestion for a fix:


> - keep the previous set of nodes in memory
> - if we don't find a signature in the current set and the block is equal and/or less than the current block number, check in the previous set

Regarding your fix - I don't think we should leniently just check the previous set if we don't find a signature in the current set. Moreover, in your commit, you check the previous set if you don't find an identity, and not a signature. This function is used also to collect commit messages, so we should not carelessly change it like this. 

I think there is a more "holistic" fix:

1. When verifying a block, we inspect the Fabric header number and look at the block number. If the last block we have committed (we check the [runtime config](https://github.com/hyperledger/fabric/blob/main/orderer/consensus/smartbft/util.go#L49)) is the previous block to the block we're verifying, then we proceed as usual. 
2. Else, if it is a block we have not committed yet (I know this is not possible, but for completeness, I am writing this case) we return an error. 
3. Else, it's a block from the past, so we fetch the block with this number from the ledger, and look at its last config block. We then fetch that config block, and proceed to use it, to verify the requested block.

The above "fix" is very easy to test, as it only changes a single function and preserves the intended verification logic.

yacovm (changes requested):
> In the very first stage of consensus, the smartBFT library takes the previous commit with signatures and checks for correctness.

If I remember correctly, we only add the previous signatures, [when we have leader rotation enabled](https://github.com/hyperledger-labs/SmartBFT/blob/main/internal/bft/view.go#L977), and in Fabric we don't have leader rotation enabled. If the leader doesn't send the previous signatures, then the followers won't verify them. 

However in the tests, [the decisions per leader is 0](https://github.com/hyperledger/fabric/blob/main/orderer/consensus/smartbft/util_network_test.go#L762)... it is therefore odd that the previous signatures verification is the cause this behavior.

Can you perhaps add a stack trace dump in the [place](https://github.com/hyperledger-labs/SmartBFT/blob/main/internal/bft/view.go#L631) where we verify the consenter signature? 

Just to give some background - we introduced the previous commit signatures in order to know how to prune the blacklist. However, we don't use leader rotation nor the blacklist in Fabric, so before we go ahead and fix something, let's be 100% sure of it's cause. 


> My suggestion for a fix:


> - keep the previous set of nodes in memory
> - if we don't find a signature in the current set and the block is equal and/or less than the current block number, check in the previous set

Regarding your fix - I don't think we should leniently just check the previous set if we don't find a signature in the current set. Moreover, in your commit, you check the previous set if you don't find an identity, and not a signature. This function is used also to collect commit messages, so we should not carelessly change it like this. 

I think there is a more "holistic" fix:

1. When verifying a block, we inspect the Fabric header number and look at the block number. If the last block we have committed (we check the [runtime config](https://github.com/hyperledger/fabric/blob/main/orderer/consensus/smartbft/util.go#L49)) is the previous block to the block we're verifying, then we proceed as usual. 
2. Else, if it is a block we have not committed yet (I know this is not possible, but for completeness, I am writing this case) we return an error. 
3. Else, it's a block from the past, so we fetch the block with this number from the ledger, and look at its last config block. We then fetch that config block, and proceed to use it, to verify the requested signature.

The above "fix" is very easy to test, as it only changes a single function and preserves the intended verification logic.

pfi79:
> Can you perhaps add a stack trace dump in the [place](https://github.com/hyperledger-labs/SmartBFT/blob/main/internal/bft/view.go#L631) where we verify the consenter signature?

easily

```
goroutine 126 [running]:
github.com/hyperledger/fabric/orderer/consensus/smartbft.(*Verifier).VerifyConsenterSig(0xc000032e80, {0x5, {0xc000447340, 0x6f, 0x6f}, {0xc0009a2b80, 0x77, 0x77}}, {{0xc000c2c000, 0xa5cf, ...}, ...})
	/Users/f/go/src/github.com/hyperledger/fabric/orderer/consensus/smartbft/verifier.go:218  0xb14
github.com/hyperledger-labs/SmartBFT/internal/bft.(*View).verifyPrevCommitSignatures(0xc0000f6600, {0xc000e9a240, 0x4, 0x4}, 0x0)
	/Users/f/go/src/github.com/hyperledger/fabric/vendor/github.com/hyperledger-labs/SmartBFT/internal/bft/view.go:631  0x7c6
github.com/hyperledger-labs/SmartBFT/internal/bft.(*View).verifyProposal(0xc0000f6600, {{0xc000efe580, 0x75, 0x75}, {0xc00063e230, 0x49, 0x49}, {0xc0009567b0, 0x4, 0x4}, ...}, ...)
	/Users/f/go/src/github.com/hyperledger/fabric/vendor/github.com/hyperledger-labs/SmartBFT/internal/bft/view.go:588  0xb45
github.com/hyperledger-labs/SmartBFT/internal/bft.(*View).processProposal(0xc0000f6600)
	/Users/f/go/src/github.com/hyperledger/fabric/vendor/github.com/hyperledger-labs/SmartBFT/internal/bft/view.go:386  0x5e5
github.com/hyperledger-labs/SmartBFT/internal/bft.(*View).doPhase(0xc0000f6600)
	/Users/f/go/src/github.com/hyperledger/fabric/vendor/github.com/hyperledger-labs/SmartBFT/internal/bft/view.go:291  0x46
github.com/hyperledger-labs/SmartBFT/internal/bft.(*View).run(0xc0000f6600)
	/Users/f/go/src/github.com/hyperledger/fabric/vendor/github.com/hyperledger-labs/SmartBFT/internal/bft/view.go:277  0x215
github.com/hyperledger-labs/SmartBFT/internal/bft.(*View).Start.func1()
	/Users/f/go/src/github.com/hyperledger/fabric/vendor/github.com/hyperledger-labs/SmartBFT/internal/bft/view.go:140  0x1c
created by github.com/hyperledger-labs/SmartBFT/internal/bft.(*View).Start in goroutine 85
	/Users/f/go/src/github.com/hyperledger/fabric/vendor/github.com/hyperledger-labs/SmartBFT/internal/bft/view.go:139  0x20d

```

pfi79:
> > In the very first stage of consensus, the smartBFT library takes the previous commit with signatures and checks for correctness.
> 
> If I remember correctly, we only add the previous signatures, [when we have leader rotation enabled](https://github.com/hyperledger-labs/SmartBFT/blob/main/internal/bft/view.go#L977), and in Fabric we don't have leader rotation enabled. If the leader doesn't send the previous signatures, then the followers won't verify them.
> 
> However in the tests, [the decisions per leader is 0](https://github.com/hyperledger/fabric/blob/main/orderer/consensus/smartbft/util_network_test.go#L762)... it is therefore odd that the previous signatures verification is the cause this behavior.
> 
> Can you perhaps add a stack trace dump in the [place](https://github.com/hyperledger-labs/SmartBFT/blob/main/internal/bft/view.go#L631) where we verify the consenter signature?
> 
> Just to give some background - we introduced the previous commit signatures in order to know how to prune the blacklist. However, we don't use leader rotation nor the blacklist in Fabric, so before we go ahead and fix something, let's be 100% sure of it's cause.
> 
> > My suggestion for a fix:
> 
> > * keep the previous set of nodes in memory
> > * if we don't find a signature in the current set and the block is equal and/or less than the current block number, check in the previous set
> 
> Regarding your fix - I don't think we should leniently just check the previous set if we don't find a signature in the current set. Moreover, in your commit, you check the previous set if you don't find an identity, and not a signature. This function is used also to collect commit messages, so we should not carelessly change it like this.
> 
> I think there is a more "holistic" fix:
> 
> 1. When verifying a block, we inspect the Fabric header number and look at the block number. If the last block we have committed (we check the [runtime config](https://github.com/hyperledger/fabric/blob/main/orderer/consensus/smartbft/util.go#L49)) is the previous block to the block we're verifying, then we proceed as usual.
> 2. Else, if it is a block we have not committed yet (I know this is not possible, but for completeness, I am writing this case) we return an error.
> 3. Else, it's a block from the past, so we fetch the block with this number from the ledger, and look at its last config block. We then fetch that config block, and proceed to use it, to verify the requested signature.
> 
> The above "fix" is very easy to test, as it only changes a single function and preserves the intended verification logic.

error itself:

```
2024-10-11 15:32:51.289 MSK 181e WARN [orderer.consensus.smartbft.consensus] processProposal -> 1 received bad proposal from 1: failed verifying consenter signature of 5: node with id of 5 doesn't exist channel=testchannel
```

pfi79:
> > > In the very first stage of consensus, the smartBFT library takes the previous commit with signatures and checks for correctness.
> > 
> > 
> > If I remember correctly, we only add the previous signatures, [when we have leader rotation enabled](https://github.com/hyperledger-labs/SmartBFT/blob/main/internal/bft/view.go#L977), and in Fabric we don't have leader rotation enabled. If the leader doesn't send the previous signatures, then the followers won't verify them.
> > However in the tests, [the decisions per leader is 0](https://github.com/hyperledger/fabric/blob/main/orderer/consensus/smartbft/util_network_test.go#L762)... it is therefore odd that the previous signatures verification is the cause this behavior.
> > Can you perhaps add a stack trace dump in the [place](https://github.com/hyperledger-labs/SmartBFT/blob/main/internal/bft/view.go#L631) where we verify the consenter signature?
> > Just to give some background - we introduced the previous commit signatures in order to know how to prune the blacklist. However, we don't use leader rotation nor the blacklist in Fabric, so before we go ahead and fix something, let's be 100% sure of it's cause.
> > > My suggestion for a fix:
> > 
> > 
> > > * keep the previous set of nodes in memory
> > > * if we don't find a signature in the current set and the block is equal and/or less than the current block number, check in the previous set
> > 
> > 
> > Regarding your fix - I don't think we should leniently just check the previous set if we don't find a signature in the current set. Moreover, in your commit, you check the previous set if you don't find an identity, and not a signature. This function is used also to collect commit messages, so we should not carelessly change it like this.
> > I think there is a more "holistic" fix:
> > 
> > 1. When verifying a block, we inspect the Fabric header number and look at the block number. If the last block we have committed (we check the [runtime config](https://github.com/hyperledger/fabric/blob/main/orderer/consensus/smartbft/util.go#L49)) is the previous block to the block we're verifying, then we proceed as usual.
> > 2. Else, if it is a block we have not committed yet (I know this is not possible, but for completeness, I am writing this case) we return an error.
> > 3. Else, it's a block from the past, so we fetch the block with this number from the ledger, and look at its last config block. We then fetch that config block, and proceed to use it, to verify the requested signature.
> > 
> > The above "fix" is very easy to test, as it only changes a single function and preserves the intended verification logic.
> 
> error itself:
> 
> ```
> 2024-10-11 15:32:51.289 MSK 181e WARN [orderer.consensus.smartbft.consensus] processProposal -> 1 received bad proposal from 1: failed verifying consenter signature of 5: node with id of 5 doesn't exist channel=testchannel
> ```

Total:
- The processProposal() function is executed in the COMMITTED phase
- after PrePrepare is generated and received, verifyProposal is executed, where PrevCommitSignatures are passed to
- after that everything is passed to the verifyPrevCommitSignatures function without any conditions.
- there each signer with a signature is verified by calling VerifyConsenterSig from the fabric
- and already VerifyConsenterSig ends with an error.

pfi79:
@yacovm you can see for yourself. 
- put the stopper [here](https://github.com/hyperledger/fabric/blob/main/orderer/consensus/smartbft/verifier.go#L216)
- and run this [test](https://github.com/hyperledger/fabric/blob/main/orderer/consensus/smartbft/chain_test.go#L358) in debug

out of 10 runs, one of them will definitely work.

pfi79:
> Regarding your fix - I don't think we should leniently just check the previous set if we don't find a signature in the current set. Moreover, in your commit, you check the previous set if you don't find an identity, and not a signature. This function is used also to collect commit messages, so we should not carelessly change it like this.
> 
> I think there is a more "holistic" fix:
> 
> 1. When verifying a block, we inspect the Fabric header number and look at the block number. If the last block we have committed (we check the [runtime config](https://github.com/hyperledger/fabric/blob/main/orderer/consensus/smartbft/util.go#L49)) is the previous block to the block we're verifying, then we proceed as usual.
> 2. Else, if it is a block we have not committed yet (I know this is not possible, but for completeness, I am writing this case) we return an error.
> 3. Else, it's a block from the past, so we fetch the block with this number from the ledger, and look at its last config block. We then fetch that config block, and proceed to use it, to verify the requested signature.
> 
> The above "fix" is very easy to test, as it only changes a single function and preserves the intended verification logic.

image

I have the block number check.

Do you want me to get the config and participant set in the same place?

yacovm:
I don't think we should be sending the previous commit signatures if we don't need them. Why do that? It's just excess CPU cycles, network bandwidth and disk storage.

I think we should not [attach](https://github.com/hyperledger-labs/SmartBFT/blob/main/internal/bft/view.go#L966) them if leader rotation is disabled.

pfi79:
> I don't think we should be sending the previous commit signatures if we don't need them. Why do that? It's just excess CPU cycles, network bandwidth and disk storage.
> 
> I think we should not [attach](https://github.com/hyperledger-labs/SmartBFT/blob/main/internal/bft/view.go#L966) them if leader rotation is disabled.

What if at the time the 5th node is removed, the leader is the 5th node? 
- leader change
- the previous signatures would be
- The 5th node can participate.

and we'll be in the same situation again

yacovm:
> > I don't think we should be sending the previous commit signatures if we don't need them. Why do that? It's just excess CPU cycles, network bandwidth and disk storage.
> > I think we should not [attach](https://github.com/hyperledger-labs/SmartBFT/blob/main/internal/bft/view.go#L966) them if leader rotation is disabled.
> 
> What if at the time the 5th node is removed, the leader is the 5th node?
> 
> * leader change
> * the previous signatures would be
> * The 5th node can participate.
> 
> and we'll be in the same situation again

If we're not attaching signatures of the previous block then how can we have a problem in verifying the signatures of the previous block which we do not attach?

pfi79:
> > > I don't think we should be sending the previous commit signatures if we don't need them. Why do that? It's just excess CPU cycles, network bandwidth and disk storage.
> > > I think we should not [attach](https://github.com/hyperledger-labs/SmartBFT/blob/main/internal/bft/view.go#L966) them if leader rotation is disabled.
> > 
> > 
> > What if at the time the 5th node is removed, the leader is the 5th node?
> > 
> > * leader change
> > * the previous signatures would be
> > * The 5th node can participate.
> > 
> > and we'll be in the same situation again
> 
> If we're not attaching signatures of the previous block then how can we have a problem in verifying the signatures of the previous block which we do not attach?

during the change of leadership?

pfi79:
> > > I don't think we should be sending the previous commit signatures if we don't need them. Why do that? It's just excess CPU cycles, network bandwidth and disk storage.
> > > I think we should not [attach](https://github.com/hyperledger-labs/SmartBFT/blob/main/internal/bft/view.go#L966) them if leader rotation is disabled.
> > 
> > 
> > What if at the time the 5th node is removed, the leader is the 5th node?
> > 
> > * leader change
> > * the previous signatures would be
> > * The 5th node can participate.
> > 
> > and we'll be in the same situation again
> 
> If we're not attaching signatures of the previous block then how can we have a problem in verifying the signatures of the previous block which we do not attach?

Or is it only needed when rotation is on?
and when rotation is disabled it is not necessary, even when the leader has changed?

yacovm:
exactly, it's only used when **periodical** leader rotation is active. A regular leader failover has nothing to do with it.

pfi79:
> exactly, it's only used when **periodical** leader rotation is active. A regular leader failover has nothing to do with it.

Okay.
then I'll refine the task. 
to leave adding signatures of the previous comitee only if leader rotation is enabled and DecisionsInView==0 

right?

yacovm:
> > exactly, it's only used when **periodical** leader rotation is active. A regular leader failover has nothing to do with it.
> 
> Okay. then I'll refine the task. to leave adding signatures of the previous comitee only if leader rotation is enabled and DecisionsInView==0
> 
> right?

I don't understand how decisions in view is related to what we're talking about. Did you mean DecisionsPerLeader?

pfi79:
> I don't understand how decisions in view is related to what we're talking about. Did you mean DecisionsPerLeader?

DecisionsPerLeader > 0 - indicator that rotation is enabled

DecisionsInView==0 - indicator that the leader has just been changed
Or is the second one unnecessary?

yacovm:
> > I don't understand how decisions in view is related to what we're talking about. Did you mean DecisionsPerLeader?
> 
> DecisionsPerLeader > 0 - indicator that rotation is enabled
> 
> DecisionsInView==0 - indicator that the leader has just been changed Or is the second one unnecessary?

The former is a configuration parameter. 
The latter is a byproduct of the consensus protocol.

So we need to consider the former, not the latter.

pfi79:
> > > I don't understand how decisions in view is related to what we're talking about. Did you mean DecisionsPerLeader?
> > 
> > 
> > DecisionsPerLeader > 0 - indicator that rotation is enabled
> > DecisionsInView==0 - indicator that the leader has just been changed Or is the second one unnecessary?
> 
> The former is a configuration parameter. The latter is a byproduct of the consensus protocol.
> 
> So we need to consider the former, not the latter.

https://github.com/hyperledger-labs/SmartBFT/pull/599

take a look, please.

PR #5029, opened 2024-10-15
test "verifies private data is pulled when joining a new peer with new certs" fix

pfi79:
test "verifies private data is pulled when joining a new peer with new certs"  is very unstable. 

my tests showed that he was running out of time.
I changed it so that the test would have time to work correctly


denyeart (approved):
Thanks!

PR #5035, opened 2024-10-17
update Dockerfiles to remove the warnings

pfi79:


images/orderer/Dockerfile
```diff
@@ -48,7 +48,7 @@ RUN make orderer GO_TAGS=${GO_TAGS} FABRIC_VER=${FABRIC_VER}
 ###############################################################################
 
 ARG UBUNTU_VER
-FROM ubuntu:${UBUNTU_VER}
+FROM ubuntu:${UBUNTU_VER:-20.04}
```
denyeart: 22.04?

pfi79: fix

images/baseos/Dockerfile
```diff
@@ -19,7 +19,7 @@
 ARG GO_VER
 ARG UBUNTU_VER
 
-FROM ubuntu:${UBUNTU_VER} as base
+FROM ubuntu:${UBUNTU_VER:-22.04} AS base
```
denyeart: One of the reasons we use variables like this is so that we only have to maintain them in one place. Setting the default value everywhere defeats the purpose of that.

Is there a compelling reason to set the default?

pfi79: > Is there a compelling reason to set the default?

This is the kind of warning from docker.

https://docs.docker.com/reference/build-checks/invalid-default-arg-in-from/

```
 1 warning found (use docker --debug to expand):
 - InvalidDefaultArgInFrom: Default value for ARG ubuntu:${UBUNTU_VER} results in empty or invalid base image name (line 24)
```

But maybe that's not a good reason.

denyeart: The risk from that warning is understood and pretty much nil, so I'm thinking it's acceptable to not fix it.

pfi79: ok. then do we close pr or keep some of the changes?

denyeart (approved):
Looks good now, thanks!

Issue #5037, opened 2024-10-18
vendor directory


pfi79:
### Current Status

Currently there is a vendor folder

I have a question: why is it needed?

It is written in the documentation for [vendoring](https://go.dev/ref/mod#vendoring): 
```
Vendoring may be used to allow interoperation with older versions of Go, or to ensure that all files used for a build are stored in a single file tree.
```

After trying to update github.com/prometheus/client_golang, the peer and orderer programs in docker stopped building. However, if the vendor folder is removed, everything builds fine. 

### Goal

It seems to me that this is already redundant in the fabric project. 
And it can be deleted.

### Solution

Try deleting the vendor folder and check that no functionality has fallen off.

C0rWin:
> After trying to update github.com/prometheus/client_golang, the peer and orderer programs in docker stopped building. However, if the vendor folder is removed, everything builds fine.

Can you put here an error?

pfi79:
> > After trying to update github.com/prometheus/client_golang, the peer and orderer programs in docker stopped building. However, if the vendor folder is removed, everything builds fine.
> 
> Can you put here an error?

https://github.com/hyperledger/fabric/pull/5032

https://github.com/hyperledger/fabric/issues/4052

```
vendor/github.com/prometheus/client_golang/prometheus/registry.go:62:15: undefined: NewGoCollector
```

a command that causes an error

```
make docker
```

PR #5039, opened 2024-10-21
delete folder vendor

pfi79:
https://github.com/hyperledger/fabric/issues/5037

I think that the vendor folder is already unnecessary and can be deleted.
It is more of a hindrance.


denyeart:
Whether to include vendor directory or not was discussed early in the project. It was decided to check in the vendor folder to ensure all dependencies remain available for every version of Fabric. Basically it is the first Pro here:

https://stackoverflow.com/questions/47346607/is-it-best-practice-to-commit-the-vendor-directory

> It's the only way to get truly reproducible builds, as it guards against upstream renames, deletes and commit history overwrites.

PR #5045, opened 2024-10-24
Test network fix for Linux hosts with SELinux enabled, e.g. Fedora, RHEL or others.

twoGiants:
#### Type of change

- Bug fix
- Documentation update

#### Description

Currently the _test network_ can only be used on Linux machines with SElinux disabled. When SELinux is enabled on the host two issues appear:
1. Peer and orderer are crashing after start when using Docker or Podman.
2. Chaincode can not be deployed when using Docker.

The reason for the first issue are permission errors with Docker/Podman volumes. This can be seen in the container logs:

```bash
# peer 1&2
ERRO [main] InitCmd -> Fatal error when initializing core config : error when reading core config file: Config File "core" Not Found in "[/etc/hyperledger/peercfg]"

# orderer
PANI [orderer.common.server] loadLocalMSP -> Failed to get local msp config: could not initialize BCCSP Factories: Failed initializing BCCSP: Could not initialize BCCSP SW [Failed to initialize software key store: open /var/hyperledger/orderer/msp/keystore: permission denied]
```
To resolve the errors the volume sections of the compose files needs to be modified.

The reason for the second issue is that the peer container can not access the Docker socket. When deploying a new chaincode in a _test network_ running on Docker the following error happens:

```bash
Error: chaincode install failed with status: 500 - failed to invoke backing implementation of 'InstallChaincode': could not build chaincode: docker build failed: docker image inspection failed: Get "http://unix.sock/images/dev-peer0.org1.example.com-basic_1.0-c6a45e2d5563c883869149c3dbd941c22fbe27daa21f0552834f5a53fbb8058a-fe69b7bdc0bbe5769bbff0572aa6986343c77b61c84077999a9b65f29c5c0025/json": dial unix /host/var/run/docker.sock: connect: permission denied
Chaincode installation on peer0.org1 has failed
Deploying chaincode failed
```

To resolve this error the Docker compose file needs to be modified. The volume section needs to be updated and the security option disabled for this container.

The errors in the container logs and during chaincode deployment were added to the test_network.md 'Troubleshooting' section as an aid if someone searches the documentation for those errors.

The necessary changes are documented in a new section called 'SELinux' under the 'Tutorials/Using the Fabric test network' docs.

#### Additional details

The documented changes were discovered and fixed in a _test network_ running on a **Fedora 40** machine.

I deployed a version of the documentation with my changes for the reviewer [here](https://twogiants-fabric.readthedocs.io/en/selinux-test-network-fix/test_network.html#selinux). Scroll down for the entry in the _test network_ troubleshooting section, its the last two entries.

Two questions to the reviewer (Update 24.10): 
1. Should I create an issue for this PR? => _no_
2. Should the fix be added to the fabric samples _test network_ setup? => _can be done, but not needed now_

### DIFF
--- a/docs/source/test_network.md
+++ b/docs/source/test_network.md
@@ -24,7 +24,7 @@ current main branch, it is possible that you will encounter errors.
 
 Before you can run the test network, you need to install Fabric Samples in your
 environment. Follow the instructions on [getting_started](getting_started.html)
-to install the required software.
+to install the required software. If you are using a Linux distribution with SELinux enabled you need to add a small change to the Docker/Podman compose files described in the [SELinux section](#selinux) below.
 
 ## Bring up the test network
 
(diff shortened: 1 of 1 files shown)


docs/source/prereqs.md
```diff
@@ -115,6 +115,57 @@ Optional: Install the latest version of [Go](https://golang.org/doc/install) (on
 
 Optional: Install the latest version of [jq](https://stedolan.github.io/jq/download/) (only required for the tutorials related to channel configuration transactions).
 
+### Distros with SELinux enabled
```
denyeart: Thank you for this enhancement. I'd like to have this content in a test-network specific doc topic, or in fabric-samples repository itself. But I'd rather not clutter the main project prereq topic with information for a specific sample. Perhaps the prereq topic could link to the ultimate destination so at least people know about the potential issue at install time, while not being a distraction to users not on SELinux.

denyeart:
To answer the questions in the Description:

- No need to create an issue, it can be discussed and fixed in this PR. Typically we create an issue for things that are not yet being PRed.
- If it can be fixed without causing too much disruption to test-network setup that would be ideal. If it will make test-network setup significantly more complex, then I agree with the doc approach. However I'd keep the doc in a test network doc topic or in a fabric-samples markdown file, rather than in the main prereq doc topic.

docs/source/prereqs.md
twoGiants: Happy to help!

You're right. I have two proposals then:
1. How about [here](https://twogiants-fabric.readthedocs.io/en/selinux-test-network-fix/test_network.html#troubleshooting) right above _Troubleshooting_ in the main documentation about the test network? With a mention of this case in _Before you begin_ up top [here](https://twogiants-fabric.readthedocs.io/en/selinux-test-network-fix/test_network.html#before-you-begin).
2. Or [here](https://github.com/hyperledger/fabric-samples/tree/main/test-network#podman) in the fabric-samples repo test network `README.md` below _Podman_. And then a reference to this entry in _Before you begin_ [here](https://twogiants-fabric.readthedocs.io/en/selinux-test-network-fix/test_network.html#before-you-begin) as in proposal 1.

twoGiants:
>   - If it can be fixed without causing too much disruption to test-network setup that would be ideal. If it will make test-network setup significantly more complex, then I agree with the doc approach...

Not to complex but not worth the effort at this point. No issues open and no complains so far. The documentation will tell how to resolve it. I think that should suffice for now.

docs/source/prereqs.md
denyeart: I like your proposal 1. Add the content to the test network troubleshooting section and mention/link to it closer to the top. Thanks!

denyeart:
Note - you also need to add your email signoff at the bottom of the commit message to pass the DCO check. More details at https://github.com/hyperledger/fabric/pull/5045/checks?check_run_id=32014394611.

docs/source/prereqs.md
twoGiants: Added changes as discussed.

denyeart (approved):
Perfect now, thank you!

PR #5058, opened 2024-11-27
docs: update docs for go files

johson-ll:
#### Type of change

- Bug fix
- New feature
- Test update
- Documentation update

### DIFF
--- a/protoutil/blockutils.go
+++ b/protoutil/blockutils.go
@@ -89,24 +89,43 @@ func GetChannelIDFromBlockBytes(bytes []byte) (string, error) {
 	return GetChannelIDFromBlock(block)
 }
 
-// GetChannelIDFromBlock returns channel ID in the block
+
+// GetChannelIDFromBlock extracts the channel ID from the provided block.
+// It parses the block data to retrieve the envelope, payload, and channel header,
+// and finally retrieves the channel ID from the channel header.
+//
+// Parameters:
+// - block: A pointer to a cb.Block object containing the blockchain block data.
+//
+// Returns:
+// - (string): The channel ID extracted from the block.
+// - (error): An error if the channel ID cannot be retrieved or the block is malformed.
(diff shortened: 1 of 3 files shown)


bestbeforetoday (commented):
Many of the comments within the code simply state exactly the same thing as the line of code they are commenting. I think this actually detracts from the readability and maintainability of the code, and should should be avoided.

For example:

```go
// Unmarshal the payload from the envelope.
payload, err := UnmarshalPayload(envelope.Payload)
```

The code itself is already perfectly clear.

bestbeforetoday:
See the [contributing guide](https://hyperledger-fabric.readthedocs.io/en/latest/CONTRIBUTING.html#legal-stuff) for why the PR build has failed the DCO check.

protoutil/blockutils.go
```diff
@@ -242,18 +261,31 @@ type policy interface { // copied from common.policies to avoid circular import.
 	EvaluateSignedData(signatureSet []*SignedData) error
 }
 
+
+// BlockSignatureVerifier returns a BlockVerifierFunc that validates the signatures of a block based on the provided policy and consenter set.
+// It checks the metadata of the block to retrieve and verify the signatures against the given policy.
```
denyeart, line 266: For functions that are missing a function comment, I agree they should be added. But I would keep them to one or two sentences (like above), and avoid the full listing of parameters (like below).

denyeart (commented):
I agree with the comment from @bestbeforetoday .

Thank you for adding the leading function comments, those should indeed exist for all exported functions and it is good to add them.

For the comments within the functions, most of them here are overkill and will bloat the code. We typically only add comments within the functions if the code intent is not clear (some of the updates here help, but most of them are unnecessary).

Please update the PR to reduce the number of comments.

denyeart:
More DCO guidance can be found by clicking the link in the failed check:
https://github.com/hyperledger/fabric/pull/5058/checks?check_run_id=33591103081

PR #5078, opened 2024-12-17
Fixes #5001 

### LINKED ISSUE #5001: Panic on leveldb
### Description

Sometimes, when load is heavy on the peers, the peer crashes with the following error:

```
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x1c pc=0x12df6f8]

goroutine 9216936 [running]:
[github.com/syndtr/goleveldb/leveldb.(*dbIter).Release(0xc00c41dd40)](https://github.com/syndtr/goleveldb/leveldb.(*dbIter).Release(0xc00c41dd40))
 /vendor/[github.com/syndtr/goleveldb/leveldb/db_iter.go:352](https://github.com/syndtr/goleveldb/leveldb/db_iter.go:352) +0x118
[github.com/hyperledger/fabric/core/ledger/kvledger/txmgmt/statedb/stateleveldb.(*kvScanner).Close(0x0?)](https://github.com/hyperledger/fabric/core/ledger/kvledger/txmgmt/statedb/stateleveldb.(*kvScanner).Close(0x0?))
 /core/ledger/kvledger/txmgmt/statedb/stateleveldb/stateleveldb.go:338 +0x1c
[github.com/hyperledger/fabric/core/ledger/kvledger/txmgmt/txmgr.(*resultsItr).Close(...)](https://github.com/hyperledger/fabric/core/ledger/kvledger/txmgmt/txmgr.(*resultsItr).Close(...))
 /core/ledger/kvledger/txmgmt/txmgr/query_executor.go:495
 ```

### Steps to reproduce

It's hard to reproduce it since it happens with a lot of load, 300tps/second with chaincode that has custom logic and most likely needs optimization.

But the conditions are:
- High load on the peers
- State db (leveldb) under stress
- Chaincode as a service in Kubernetes

### PR DESCRIPTION
**Pull request by dviejokfs**
#### Type of change

#### Description

This pull request introduces several improvements in the `HandleGetStateByRange` function within the `core/chaincode/handler.go` file and the `GetStateRangeScanIteratorWithPagination` function within the `core/ledger/kvledger/txmgmt/statedb/stateleveldb/stateleveldb.go` file. The changes include:

- Adding panic recovery in the `HandleGetStateByRange` function to log errors and clean up any query contexts that may have been created.
- Validating start and end keys to ensure they are not both empty.
- Adding validation for the page size to ensure it is greater than zero.
- Cleaning up query contexts if errors occur during the range query or if the range query iterator is nil.
- Wrapping the iterator creation in a recover block in `GetStateRangeScanIteratorWithPagination` to handle potential panics and return meaningful error messages.
- Ensuring that a nil iterator is not returned from the database.

#### Additional details

These changes help improve the robustness of the code by handling potential panics and validating inputs, which can help prevent unexpected errors and improve the overall stability of the system.

#### Related issues

#5001

### DIFF
--- a/gossip/gossip/algo/pull.go
+++ b/gossip/gossip/algo/pull.go
@@ -319,11 +319,18 @@ func (engine *PullEngine) OnRes(items []string, nonce uint64) {
 }
 
 func (engine *PullEngine) newNONCE() uint64 {
-	n := uint64(0)
-	for {
-		n = util.RandomUInt64()
-		if !engine.outgoingNONCES.Exists(n) {
+	engine.lock.Lock()
+	defer engine.lock.Unlock()
+
+	maxAttempts := 100
+	for i := 0; i < maxAttempts; i++ {
+		n := util.RandomUInt64()
+		if n != 0 && !engine.outgoingNONCES.Exists(n) {
 			return n
 		}
 	}
+
+	// If we couldn't generate a unique NONCE after max attempts,
+	// use time-based fallback
+	return uint64(time.Now().UnixNano())
 }
--- a/gossip/gossip/algo/pull.go
+++ b/gossip/gossip/algo/pull.go
@@ -7,6 +7,8 @@ SPDX-License-Identifier: Apache-2.0
 package algo
 
 import (
+	"crypto/rand"
+	"encoding/binary"
 	"sync"
 	"sync/atomic"
 	"time"
@@ -322,7 +324,7 @@ func (engine *PullEngine) newNONCE() uint64 {
 	engine.lock.Lock()
 	defer engine.lock.Unlock()
 
-	maxAttempts := 100
+	maxAttempts := 1000000
 	for i := 0; i < maxAttempts; i++ {
 		n := util.RandomUInt64()
 		if n != 0 && !engine.outgoingNONCES.Exists(n) {
@@ -330,7 +332,18 @@ func (engine *PullEngine) newNONCE() uint64 {
 		}
 	}
 
-	// If we couldn't generate a unique NONCE after max attempts,
-	// use time-based fallback
-	return uint64(time.Now().UnixNano())
+	// If we still couldn't generate a unique NONCE after max attempts,
+	// keep trying with crypto/rand until we find one
+	var n uint64
+	for {
+		bytes := make([]byte, 8)
+		if _, err := rand.Read(bytes); err != nil {
+			// Fallback to time-based if crypto/rand fails
+			return uint64(time.Now().UnixNano())
+		}
+		n = binary.BigEndian.Uint64(bytes)
+		if n != 0 && !engine.outgoingNONCES.Exists(n) {
+			return n
+		}
+	}
 }
--- a/gossip/gossip/algo/pull.go
+++ b/gossip/gossip/algo/pull.go
@@ -7,8 +7,6 @@ SPDX-License-Identifier: Apache-2.0
 package algo
 
 import (
-	"crypto/rand"
-	"encoding/binary"
 	"sync"
 	"sync/atomic"
 	"time"
@@ -321,28 +319,10 @@ func (engine *PullEngine) OnRes(items []string, nonce uint64) {
 }
 
 func (engine *PullEngine) newNONCE() uint64 {
-	engine.lock.Lock()
-	defer engine.lock.Unlock()
-
-	maxAttempts := 1000000
-	for i := 0; i < maxAttempts; i++ {
-		n := util.RandomUInt64()
-		if n != 0 && !engine.outgoingNONCES.Exists(n) {
-			return n
-		}
-	}
-
-	// If we still couldn't generate a unique NONCE after max attempts,
-	// keep trying with crypto/rand until we find one
-	var n uint64
+	n := uint64(0)
 	for {
-		bytes := make([]byte, 8)
-		if _, err := rand.Read(bytes); err != nil {
-			// Fallback to time-based if crypto/rand fails
-			return uint64(time.Now().UnixNano())
-		}
-		n = binary.BigEndian.Uint64(bytes)
-		if n != 0 && !engine.outgoingNONCES.Exists(n) {
+		n = util.RandomUInt64()
+		if !engine.outgoingNONCES.Exists(n) {
 			return n
 		}
 	}
--- a/gossip/util/misc.go
+++ b/gossip/util/misc.go
@@ -18,10 +18,7 @@ import (
 	"github.com/spf13/viper"
 )
 
-var (
-	r      *rand.Rand
-	rMutex sync.Mutex
-)
+var r *rand.Rand
 
 func init() { // do we really need this?
 	var seed [32]byte
@@ -190,8 +187,6 @@ func RandomInt(n int) int {
 // If we want a rand that's non-global and specific to gossip, we can
 // establish one. Otherwise this uses the process-global locking RNG.
 func RandomUInt64() uint64 {
-	rMutex.Lock()
-	defer rMutex.Unlock()
 	return r.Uint64()
 }
 
--- a/gossip/util/misc.go
+++ b/gossip/util/misc.go
@@ -7,6 +7,7 @@ SPDX-License-Identifier: Apache-2.0
 package util
 
 import (
+	crand "crypto/rand"
 	"fmt"
 	"math/rand/v2"
 	"reflect"
--- a/gossip/util/misc.go
+++ b/gossip/util/misc.go
@@ -23,7 +23,9 @@ var (
 )
 
 func init() { // do we really need this?
-	r = rand.New(rand.NewSource(time.Now().UnixNano()))
+	var seed [32]byte
+	_, _ = crand.Read(seed[:])
+	r = rand.New(rand.NewChaCha8(seed))
 }
 
 // Equals returns whether a and b are the same
--- a/gossip/util/misc.go
+++ b/gossip/util/misc.go
@@ -7,9 +7,8 @@ SPDX-License-Identifier: Apache-2.0
 package util
 
 import (
-	crand "crypto/rand"
 	"fmt"
-	"math/rand/v2"
+	"math/rand"
 	"reflect"
 	"runtime"
 	"sync"
@@ -18,12 +17,13 @@ import (
 	"github.com/spf13/viper"
 )
 
-var r *rand.Rand
+var (
+	r      *rand.Rand
+	rMutex sync.Mutex
+)
 
 func init() { // do we really need this?
-	var seed [32]byte
-	_, _ = crand.Read(seed[:])
-	r = rand.New(rand.NewChaCha8(seed))
+	r = rand.New(rand.NewSource(time.Now().UnixNano()))
 }
 
 // Equals returns whether a and b are the same
@@ -179,7 +179,7 @@ func SetVal(key string, val interface{}) {
 // RandomInt returns, as an int, a non-negative pseudo-random integer in [0,n)
 // It panics if n <= 0
 func RandomInt(n int) int {
-	return r.IntN(n)
+	return r.Intn(n)
 }
 
 // RandomUInt64 returns a random uint64
@@ -187,6 +187,8 @@ func RandomInt(n int) int {
 // If we want a rand that's non-global and specific to gossip, we can
 // establish one. Otherwise this uses the process-global locking RNG.
 func RandomUInt64() uint64 {
+	rMutex.Lock()
+	defer rMutex.Unlock()
 	return r.Uint64()
 }
 
--- a/gossip/util/misc.go
+++ b/gossip/util/misc.go
@@ -8,7 +8,7 @@ package util
 
 import (
 	"fmt"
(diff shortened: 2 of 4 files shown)

### CODE REVIEW

File: gossip/util/misc.go [hunk 1]
```diff
@@ -18,12 +17,13 @@ import (
...
+var (
+	r      *rand.Rand
+	rMutex sync.Mutex
+)
 
 func init() { // do we really need this?
-	var seed [32]byte
-	_, _ = crand.Read(seed[:])
-	r = rand.New(rand.NewChaCha8(seed))
+	r = rand.New(rand.NewSource(time.Now().UnixNano()))
```
yacovm: why would we seed this with a deterministic and non-random seed? 
what was before was better, in my opinion.

gossip/gossip/algo/pull.go
```diff
@@ -319,11 +319,18 @@ func (engine *PullEngine) OnRes(items []string, nonce uint64) {
...
+	defer engine.lock.Unlock()
+
+	maxAttempts := 100
+	for i := 0; i < maxAttempts; i++ {
+		n := util.RandomUInt64()
+		if n != 0 && !engine.outgoingNONCES.Exists(n) {
 			return n
 		}
 	}
+
```
yacovm: We shouldn't try 100 attempts, but more attempts, like 1000,000. 

and if we still fail, then we should loop until we find a nonce that is not in use using the crypto rand.

yacovm:
I don't understand how the gossip nonce generation is related to the panic, can you explain?

gossip/util/misc.go
```diff
@@ -7,9 +7,8 @@ SPDX-License-Identifier: Apache-2.0
 package util
 
 import (
-	crand "crypto/rand"
 	"fmt"
-	"math/rand/v2"
+	"math/rand"
```
pfi79: Why did you change the library from v2 to the previous library?

gossip/util/misc.go
dviejokfs: ok, rolling back

gossip/gossip/algo/pull.go
dviejokfs: For the second part it can result in slow nonces if this happens right?

gossip/util/misc.go
dviejokfs: sorry, didn't mean to, changed back

gossip/util/misc.go
```diff
@@ -187,6 +190,8 @@ func RandomInt(n int) int {
 // If we want a rand that's non-global and specific to gossip, we can
 // establish one. Otherwise this uses the process-global locking RNG.
 func RandomUInt64() uint64 {
+	rMutex.Lock()
+	defer rMutex.Unlock()
```
pfi79: why is the lock/unlock call only here?

dviejokfs: This is the panic that happens when there's a lot of concurrency:

panic: runtime error: index out of range [-1]

goroutine 3305 [running]:
math/rand.(*rngSource).Uint64(0xc014386101?)
/usr/local/go/src/math/rand/rng.go:249 +0x7b
math/rand.(*Rand).Uint64(0xc0180cf980?)
/usr/local/go/src/math/rand/rand.go:104 +0x22
github.com/hyperledger/fabric/gossip/util.RandomUInt64(...)
/gossip/util/misc.go:187
github.com/hyperledger/fabric/gossip/gossip/algo.(*PullEngine).newNONCE(0xc006953000)
/gossip/gossip/algo/pull.go:324 +0x25
github.com/hyperledger/fabric/gossip/gossip/algo.(*PullEngine).initiatePull(0xc006953000)
/gossip/gossip/algo/pull.go:179 +0x11b
github.com/hyperledger/fabric/gossip/gossip/algo.NewPullEngineWithFilter.func1()
/gossip/gossip/algo/pull.go:126 +0x27
created by github.com/hyperledger/fabric/gossip/gossip/algo.NewPullEngineWithFilter in goroutine 1
/gossip/gossip/algo/pull.go:120 +0x2d5

I suppose the problem is because we are accessing this function a lot of times, so adding a lock ensures that there are no concurrency issues

pfi79: 1. Above r is used in 2 more places without lock. Why is lock not used there?

2. yes, this is a bug in the math/rand package. But after the release of v3.0.0 the package used https://github.com/hyperledger/fabric/commit/9b28d51052ce670bc70a244897e0712a4e0ab08b was changed, this change is not in the release yet. And your stacktrace is not using v2 code.

3. Need proof that you actually fixed the bug suddenly it was fixed by changing the package to version v2. You need a test to prove that your fixes are correct

```
/usr/local/go/src/math/rand/rng.go:249 +0x7b
math/rand.(*Rand).Uint64(0xc0180cf980?)
/usr/local/go/src/math/rand/rand.go:104 +0x22
```

pfi79: 1. Above r is used in 2 more places without lock. Why is lock not used there?

2. yes, this is a bug in the math/rand package. But after the release of v3.0.0 the package used https://github.com/hyperledger/fabric/commit/9b28d51052ce670bc70a244897e0712a4e0ab08b was changed, this change is not in the release yet. And your stacktrace is not using v2 code.

3. Need proof that you actually fixed the bug suddenly it was fixed by changing the package to version v2. You need a test to prove that your fixes are correct

```
/usr/local/go/src/math/rand/rng.go:249 +0x7b
math/rand.(*Rand).Uint64(0xc0180cf980?)
/usr/local/go/src/math/rand/rand.go:104 +0x22
```

yacovm: > This is the panic that happens when there's a lot of concurrency:
> 
> panic: runtime error: index out of range [-1]
> 
> goroutine 3305 [running]: math/rand.(*rngSource).Uint64(0xc014386101?) /usr/local/go/src/math/rand/rng.go:249 +0x7b math/rand.(*Rand).Uint64(0xc0180cf980?) /usr/local/go/src/math/rand/rand.go:104 +0x22 github.com/hyperledger/fabric/gossip/util.RandomUInt64(...) /gossip/util/misc.go:187 github.com/hyperledger/fabric/gossip/gossip/algo.(*PullEngine).newNONCE(0xc006953000) /gossip/gossip/algo/pull.go:324 +0x25 github.com/hyperledger/fabric/gossip/gossip/algo.(*PullEngine).initiatePull(0xc006953000) /gossip/gossip/algo/pull.go:179 +0x11b github.com/hyperledger/fabric/gossip/gossip/algo.NewPullEngineWithFilter.func1() /gossip/gossip/algo/pull.go:126 +0x27 created by github.com/hyperledger/fabric/gossip/gossip/algo.NewPullEngineWithFilter in goroutine 1 /gossip/gossip/algo/pull.go:120 +0x2d5
> 
> I suppose the problem is because we are accessing this function a lot of times, so adding a lock ensures that there are no concurrency issues

If there is a panic stemming from gossip then I fail to see how is the chaincode related to the problem. 

We shouldn't fix two different problems in the same PR. 

Can you move all the non gossip related code changes to a new PR?

dviejokfs: yeah, but you see the problem, right? gossip is invoking randomUint64, and it panicking,

dviejokfs: you are right, these are two different problems for two different panic errors

gossip/gossip/algo/pull.go
```diff
@@ -319,10 +321,28 @@ func (engine *PullEngine) OnRes(items []string, nonce uint64) {
 }
 
 func (engine *PullEngine) newNONCE() uint64 {
-	n := uint64(0)
+	engine.lock.Lock()
```
yacovm: we don't need to lock at all. We only create nonces from [here](https://github.com/hyperledger/fabric/blob/0501b0ddcfca9992114a3e137ed5b05156c7e13f/gossip/gossip/algo/pull.go#L179) and we already lock there. 

Just loop until `engine.outgoingNONCES.Exists(n)` is false, no need for attempts or anything.

gossip/gossip/algo/pull.go
yacovm: after looking at the code myself, I think we can just use `util.RandomUInt64` until it no longer exists in the outgoing nonces.

gossip/util/misc.go
pfi79: panic calls a different code than the one you're correcting.

dviejokfs: So @pfi79 compiling the code in main should fix the issue right about gossip right?

dviejokfs:
@yacovm @pfi79 
I have removed the changes for the gossip issue, this PR only contains the issue about the leveldb

gossip/util/misc.go
pfi79: > So @pfi79 compiling the code in main should fix the issue right about gossip right?

It's possible. After updating the package to v2 I stopped catching the panic error in tests.

pfi79:
> @yacovm @pfi79 I have removed the changes for the gossip issue, this PR only contains the issue about the leveldb

please update the pr description.

there is not enough evidence that your change fixes the error when working with leveldb

and please look at the unit test. after your changes they dropped exactly on leveldb.

dviejokfs:
> > @yacovm @pfi79 I have removed the changes for the gossip issue, this PR only contains the issue about the leveldb
> 
> please update the pr description.
> 
> there is not enough evidence that your change fixes the error when working with leveldb
> 
> and please look at the unit test. after your changes they dropped exactly on leveldb.

I just updated the description, I'm going to check the unit tests

yacovm:
Since you removed all gossip code from the PR, can't you just force push and get rid of the gossip code change commits? squash all changes together and force push so it will be more clear what's going on when this is merged, otherwise we'll merge a bunch of commits that cancel each other out.

yacovm:
And I will tag @denyeart here, since the only code change in this PR is related to things he has more context about than me and pfi79.

denyeart:
Thanks for the update @dviejokfs , although the PR has become confusing - 9 commits, a gossip title and bunch of gossip comments, but no gossip code, and 6 bullets in description related to chaincode and ledger, with no tests. It is unclear how everything relates. I suggest to reset - close this PR and open separate PRs for each individual improvement that can stand by itself. Each individual PR should have the smallest unit of code needed for the respective improvement, along with related unit test updates. The unit tests should demonstrate the problem being fixed (failing before the change, and passing after the change). Then the smaller individual PRs can each be judged on their own merit with clarity and focused discussion.

Note that to keep the commits clean we often use the commit amend approach. This way you don't have to go through the process of squashing and then fixing up the commit/PR description. See details at https://hyperledger-fabric.readthedocs.io/en/latest/github/github.html#updating-a-pull-request.

dviejokfs:
@denyeart 
Related to the unit tests, these don't change, how can we reproduce something that happens when there's a lot of concurrency? The issue is in the leveldb library:


panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x1c pc=0x12df6f8]

goroutine 9216936 [running]:
[github.com/syndtr/goleveldb/leveldb.(*dbIter).Release(0xc00c41dd40)](https://github.com/syndtr/goleveldb/leveldb.(*dbIter).Release(0xc00c41dd40))
 /vendor/[github.com/syndtr/goleveldb/leveldb/db_iter.go:352](https://github.com/syndtr/goleveldb/leveldb/db_iter.go:352) +0x118
[github.com/hyperledger/fabric/core/ledger/kvledger/txmgmt/statedb/stateleveldb.(*kvScanner).Close(0x0?)](https://github.com/hyperledger/fabric/core/ledger/kvledger/txmgmt/statedb/stateleveldb.(*kvScanner).Close(0x0?))
 /core/ledger/kvledger/txmgmt/statedb/stateleveldb/stateleveldb.go:338 +0x1c
[github.com/hyperledger/fabric/core/ledger/kvledger/txmgmt/txmgr.(*resultsItr).Close(...)](https://github.com/hyperledger/fabric/core/ledger/kvledger/txmgmt/txmgr.(*resultsItr).Close(...))
 /core/ledger/kvledger/txmgmt/txmgr/query_executor.go:495

pfi79:
> Related to the unit tests, these don't change, how can we reproduce something that happens when there's a lot of concurrency? The issue is in the leveldb library:
> 
> panic: runtime error: invalid memory address or nil pointer dereference [signal SIGSEGV: segmentation violation code=0x1 addr=0x1c pc=0x12df6f8]

leveldb has been [updated](https://github.com/hyperledger/fabric/commit/eb3a5ebb3121b605d78992dfedc8d608650c767c). is the error still there or not? i don't know.
your changes concern the previous version.

denyeart:
@dviejokfs Agreed it may not be possible to cause the error in unit tests or integration tests, but it should at least be considered whether the problem can be induced dependably so that we know when it is resolved, even if that requires a manual test. In the extreme cases we've even sometimes temporarily updated Fabric code to make a problem more likely so that we can reproduce and ensure it is fixed. Do you have ideas around this?

dviejokfs:
@denyeart @yacovm  @pfi79   I just created a new PR containing a unit test and the minimal fix #5084

Issue #5076, opened 2024-12-16
The ordering nodes panic during synchronization.


w4z5:
### Description

I have deployed a consensus cluster with 10 orderer nodes in Hyperledger Fabric version 3.0.0-beta. Among these, orderer8, orderer9, and orderer10 are randomly started and stopped. During the experiment, I repeatedly use `docker stop` and `docker start` commands to stop and start these nodes. At any given time, only one node is stopped, with a downtime of 2 minutes, and a recovery time of 5 minutes after it is started again. During the experiment, the following panic occurs during the node recovery phase.
![Image](https://github.com/user-attachments/assets/7d9f789f-fec7-46ce-a713-c03e866ffce2)
Upon reviewing the logs of orderer9, I found the following two log entries:

> 2024-11-09 09:06:46.618 UTC 164fa DEBU [orderer.consensus.smartbft.chain] getBlocksFromSyncBuffer -> Fetched and committed block [223] from cluster channel=channel1
> 2024-11-09 09:06:46.896 UTC 168b5 DEBU [orderer.consensus.smartbft.chain] getBlocksFromSyncBuffer -> Will fetch sequences [223-236] channel=channel1

Even after block 223 was committed, it was still selected as a block to synchronize. When the node attempted to submit block 223 again, a panic occurred.
Could the developers please help assess if such a panic represents a security risk in Fabric? I would appreciate any insights on this

bestbeforetoday:
Do you see better behaviour using the v3.0.0 release version instead of the 3.0.0-beta?

pfi79:
A few clarifying questions:
- What does `recovery time of 5 minutes after it is started again`, please specify.
- Is it the same on 7 knots with stop 7 and 6 or on 4 knots
- What load is applied to the fabric
- What happens to the node after a panic, does it remain inoperable or does it recover after some time and work normally

pfi79:
It would also be my wish, if possible: 
- Reduce the number of nodes to 7 or 4 and have the error repeat.
- put order logs in "debug" for fabric and smart-bft, and when the error repeats, all orderers logs in different files should be archived and attached to the discussion.

w4z5:
> A few clarifying questions:
> 
> * What does `recovery time of 5 minutes after it is started again`, please specify.
> * Is it the same on 7 knots with stop 7 and 6 or on 4 knots
> * What load is applied to the fabric
> * What happens to the node after a panic, does it remain inoperable or does it recover after some time and work normally

[...]

PR #5081, opened 2024-12-18
fix nodes panic during synchronization

pfi79:
https://github.com/hyperledger/fabric/issues/5076

**What's going on?**

During synchronisation, the synchroniser reads blocks from another node and [writes](https://github.com/hyperledger/fabric/blob/main/orderer/consensus/smartbft/synchronizer_bft.go#L261) them to itself.
Block [writes](https://github.com/hyperledger/fabric/blob/main/orderer/common/multichannel/blockwriter.go#L166) are started asynchronously, and we have no control over when this goroutine will actually be executed. 
Without waiting for the block to be written, the synchroniser proceeds to the next step.
When finished, the synchroniser exits and the code immediately detects that the node has not reached altitude and enters a new synchronisation cycle.

The `startHeight := s.Support.Height()` method is executed as the start of the range to synchronise. But it happens that the previous block has not been written yet (but will be soon) and we choose the number of the block that is about to be written as the start. Thus it is possible to write two identical blocks in a row.

**Proposal** 
Set block recording to synchronous mode


yacovm:
Thanks for looking into this.

1. What about a unit test that reproduces the problem?
2. I don't understand why we need such a complex solution, when we can just [write the blocks synchronously](https://github.com/hyperledger/fabric/blob/main/orderer/common/multichannel/blockwriter.go#L182)?

pfi79:
> 2\. I don't understand why we need such a complex solution, when we can just [write the blocks synchronously](https://github.com/hyperledger/fabric/blob/main/orderer/common/multichannel/blockwriter.go#L182)?

I was thinking about synchronous writing, but as I think the idea here was to parallele getting the next block from puller and writing the current one to state. That is a time saver.
But if other maintainers agree to convert to synchronous recording, I will.

pfi79:
> 1. What about a unit test that reproduces the problem?

I'm at work on a test. 
Was the first to throw a pr with changes to get corrected right away.

yacovm:
> > 2. I don't understand why we need such a complex solution, when we can just [write the blocks synchronously](https://github.com/hyperledger/fabric/blob/main/orderer/common/multichannel/blockwriter.go#L182)?
> 
> I was thinking about synchronous writing, but as I think the idea here was to parallele getting the next block from puller and writing the current one to state. That is a time saver. But if other maintainers agree to convert to synchronous recording, I will.

we do get the next block from the remote node though, whether you are doing a sync write or an async one.

Because when you [retrieve](https://github.com/hyperledger/fabric/blob/main/orderer/consensus/smartbft/sync_buffer.go#L46) a block from the sync buffer, a new block is [fed](https://github.com/hyperledger/fabric/blob/main/orderer/consensus/smartbft/sync_buffer.go#L39) into it.

pfi79:
> we do get the next block from the remote node though, whether you are doing a sync write or an async one.
> 
> Because when you [retrieve](https://github.com/hyperledger/fabric/blob/main/orderer/consensus/smartbft/sync_buffer.go#L46) a block from the sync buffer, a new block is [fed](https://github.com/hyperledger/fabric/blob/main/orderer/consensus/smartbft/sync_buffer.go#L39) into it.

That's right. I'll redo it.

yacovm (approved):

yacovm:
I know that you didn't provide a test, but I am perfectly fine with using the sync block writing when we synchronize.

PR #5108, opened 2025-01-17
chore: update go tool->path mapping

bitsark:
#### Type of change
- Bug fix

#### Description

gotools.mk is not update to tools/tools.go, so `make gotools` may failed

#### Related issues


denyeart:
Updated in https://github.com/hyperledger/fabric/pull/5109

PR #5110, opened 2025-01-20
Improve gossip handling of in-memory expired certs

### LINKED ISSUE #5111: Gossip doesn't replace PKI-ID of a peer that has renewed an expired certificate
### Description

When a peer gets a renewed enrollment cert before it expires, gossip on other peers recognizes the change and replaces the in-memory PKI-ID, like this:

```
INFO [gossip.comm] createConnection -> Peer 127.0.0.1:22510 changed its PKI-ID from 3257ef1cb73517a1c07b06c0c8f06903cb537199ffb11ecd23643b900f3f43d2 to 9f3e40a456e9f1a22c73e1206bd78b3b1b0ed7276ffbb37b9de00d23530f0877
INFO [gossip.discovery] purge -> Purging 3257ef1cb73517a1c07b06c0c8f06903cb537199ffb11ecd23643b900f3f43d2 from membership
```

However, if an enrollment cert expires before gossip on other peers sees the renewal, gossip is left in a bad state, like this:

```
WARN [gossip.gossip] func3 -> Unable to determine org of message tag:EMPTY  alive_msg:{membership:{endpoint:"127.0.0.1:22510"  pki_id:"\x9f>@\xa4V\xe9\xf1\xa2,s\xe1 k׋;\x1b\x0e\xd7'o\xfb\xb3{\x9d\xe0\r#S\x0f\x08w"}  timestamp:{inc_num:1737346511720016000  seq_num:29}}
WARN [gossip.gossip] disclosurePolicy -> Cannot determine organization of Endpoint: 127.0.0.1:22510, InternalEndpoint: , PKI-ID: 9f3e40a456e9f1a22c73e1206bd78b3b1b0ed7276ffbb37b9de00d23530f0877, Metadata: 
ERRO [peer.gossip.sa] OrgByPeerIdentity -> Invalid Peer Identity. It must be different from nil.
WARN [gossip.comm] createConnection -> Remote endpoint claims to be a different peer, expected 9f3e40a456e9f1a22c73e1206bd78b3b1b0ed7276ffbb37b9de00d23530f0877 but got b4944bc458640ad309569bc34eb58d66ad6a67e642f2ee89edead2e2fabd7cd8
```

The result is that gossip membership bounces between established and not established until the other peers are restarted.

### Steps to reproduce

See integration test to reproduce https://github.com/hyperledger/fabric/pull/5110

denyeart:
If a peer's cert expires while it is still in gossip memory
for another peer, membership cannot be re-established
after the cert is renewed.

The fix is to acknowledge that the expired cert's peer
is no longer in gossip identity store and accept
new connection with the new PKI-ID and identity.

Add new integration test:
-Test renew cert before expiration (passes)
-Test renew cert after expiration (previously failed)

Resolves #5111

### DIFF
--- a/gossip/comm/comm_impl.go
+++ b/gossip/comm/comm_impl.go
@@ -162,6 +162,14 @@ func (c *commImpl) createConnection(endpoint string, expectedPKIID common.PKIidT
 		return nil, errors.WithStack(err)
 	}
 
+	if len(expectedPKIID) > 0 {
+		identity, _ := c.idMapper.Get(expectedPKIID)
+		if len(identity) == 0 {
+			c.logger.Warningf("Identity of %x is no longer found in the identity store, aborting connection", expectedPKIID)
+			return nil, errors.New("identity no longer recognized")
+		}
+	}
+
 	ctx, cancel = context.WithCancel(context.Background())
 	if stream, err = cl.GossipStream(ctx); err == nil {
 		connInfo, err = c.authenticateRemotePeer(stream, true, false)
--- a/integration/gossip/gossip_test.go
+++ b/integration/gossip/gossip_test.go
@@ -7,6 +7,10 @@
 package gossip
 
 import (
+	"crypto/ecdsa"
+	"crypto/rand"
+	"crypto/x509"
+	"encoding/pem"
 	"fmt"
 	"os"
 	"path/filepath"
@@ -285,7 +289,129 @@ var _ = Describe("Gossip State Transfer and Membership", func() {
 			assertPeerMembershipUpdate(network, peer1Org1, []*nwo.Peer{peer0Org2, peer1Org2}, nwprocs, expectedMsgFromExpirationCallback)
 		})
 	})
+
+	It("updates membership for a peer with a renewed certificate", func() {
+		network.Bootstrap()
+		orderer := network.Orderer("orderer")
+		nwprocs.ordererRunner = network.OrdererRunner(orderer)
+		nwprocs.ordererProcess = ifrit.Invoke(nwprocs.ordererRunner)
+		Eventually(nwprocs.ordererProcess.Ready(), network.EventuallyTimeout).Should(BeClosed())
+
+		peer0Org1 := network.Peer("Org1", "peer0")
+		peer0Org2 := network.Peer("Org2", "peer0")
+
+		By("bringing up a peer in each organization")
+		startPeers(nwprocs, false, peer0Org1, peer0Org2)
+
+		channelparticipation.JoinOrdererAppChannel(network, "testchannel", orderer, nwprocs.ordererRunner)
+
+		By("joining peers to channel")
+		network.JoinChannel(channelName, orderer, peer0Org1, peer0Org2)
+
+		By("verifying membership of both peers")
+		bothPeers := []*nwo.Peer{peer0Org1, peer0Org2}
+		network.VerifyMembership(bothPeers, channelName)
+
+		fmt.Println("===MEMBERSHIP VERIFIED===")
+		time.Sleep(5 * time.Second)
+
+		By("stopping, renewing peer0Org2 certificate before expiration, and restarting")
+		stopPeers(nwprocs, peer0Org2)
+		renewPeerCertificate(network, peer0Org2, time.Now().Add(time.Minute))
+
+		fmt.Println("===STOPPED AND RENEWED peer0Org2, RESTARTING, CHECK FOR PKI-ID REPLACEMENT===")
+		time.Sleep(5 * time.Second)
+
+		startPeers(nwprocs, false, peer0Org2)
+
+		By("ensuring that peer0Org1 replaces peer0Org2 PKI-ID")
+		peer0Org1Runner := nwprocs.peerRunners[peer0Org1.ID()]
+		Eventually(peer0Org1Runner.Err(), network.EventuallyTimeout).Should(gbytes.Say("changed its PKI-ID from"))
+
+		fmt.Println("===PKI-ID REPLACED, WAIT FOR MEMBERSHIP===")
+		time.Sleep(5 * time.Second)
+
+		By("verifying membership after cert renewed")
+		network.VerifyMembership(bothPeers, channelName)
+
+		fmt.Println("===MEMBERSHIP VERIFIED WITH RENEWED CERT, WAIT FOR CERT TO EXPIRE AND THEN RENEW AGAIN ===")
+		time.Sleep(5 * time.Second)
+
+		By("waiting for cert to expire within a minute")
+		Eventually(peer0Org1Runner.Err(), network.EventuallyTimeout).Should(gbytes.Say("gossipping peer identity expired"))
+
+		By("stopping, renewing peer0Org2 certificate again after its expiration, restarting")
+		stopPeers(nwprocs, peer0Org2)
+		renewPeerCertificate(network, peer0Org2, time.Now().Add(time.Hour))
+
+		fmt.Println("===STOPPED AND RENEWED peer0Org2 AGAIN AFTER EXPIRATION, RESTARTING, CHECK FOR PKI-ID REPLACEMENT AGAIN===")
+		time.Sleep(5 * time.Second)
+
+		startPeers(nwprocs, false, peer0Org2)
+
+		By("ensuring that peer0Org1 replaces peer0Org2 PKI-ID after it expired")
+		Eventually(peer0Org1Runner.Err(), network.EventuallyTimeout).Should(gbytes.Say("changed its PKI-ID from"))
+	})
 })
+
+// renewPeerCertificate renews the certificate with a given expirationTime and re-writes it to the peer's signcert directory
+func renewPeerCertificate(network *nwo.Network, peer *nwo.Peer, expirationTime time.Time) {
+	peerDomain := network.Organization(peer.Organization).Domain
+
+	peerCAKeyPath := filepath.Join(network.RootDir, "crypto", "peerOrganizations", peerDomain, "ca", "priv_sk")
+	peerCAKey, err := os.ReadFile(peerCAKeyPath)
+	Expect(err).NotTo(HaveOccurred())
+
+	peerCACertPath := filepath.Join(network.RootDir, "crypto", "peerOrganizations", peerDomain, "ca", fmt.Sprintf("ca.%s-cert.pem", peerDomain))
+	peerCACert, err := os.ReadFile(peerCACertPath)
+	Expect(err).NotTo(HaveOccurred())
+
+	peerCertPath := filepath.Join(network.PeerLocalMSPDir(peer), "signcerts", fmt.Sprintf("peer0.%s-cert.pem", peerDomain))
+	peerCert, err := os.ReadFile(peerCertPath)
+	Expect(err).NotTo(HaveOccurred())
+
+	renewedCert, _ := expireCertificate(peerCert, peerCACert, peerCAKey, expirationTime)
+	err = os.WriteFile(peerCertPath, renewedCert, 0o600)
+	fmt.Println(peerCertPath)
+	Expect(err).NotTo(HaveOccurred())
+}
+
+// expireCertificate re-creates and re-signs a certificate with a new expirationTime
+func expireCertificate(certPEM, caCertPEM, caKeyPEM []byte, expirationTime time.Time) (expiredcertPEM []byte, earlyMadeCACertPEM []byte) {
+	keyAsDER, _ := pem.Decode(caKeyPEM)
+	caKeyWithoutType, err := x509.ParsePKCS8PrivateKey(keyAsDER.Bytes)
+	Expect(err).NotTo(HaveOccurred())
+	caKey := caKeyWithoutType.(*ecdsa.PrivateKey)
+
+	caCertAsDER, _ := pem.Decode(caCertPEM)
+	caCert, err := x509.ParseCertificate(caCertAsDER.Bytes)
+	Expect(err).NotTo(HaveOccurred())
+
+	certAsDER, _ := pem.Decode(certPEM)
+	cert, err := x509.ParseCertificate(certAsDER.Bytes)
+	Expect(err).NotTo(HaveOccurred())
+
+	cert.Raw = nil
+	caCert.Raw = nil
+	// The certificate was made 1 minute ago (1 hour doesn't work since cert will be before original CA cert NotBefore time)
+	cert.NotBefore = time.Now().Add((-1) * time.Minute)
+	// As well as the CA certificate
+	caCert.NotBefore = time.Now().Add((-1) * time.Minute)
+	// The certificate expires now
+	cert.NotAfter = expirationTime
+
+	// The CA signs the certificate
+	certBytes, err := x509.CreateCertificate(rand.Reader, cert, caCert, cert.PublicKey, caKey)
+	Expect(err).NotTo(HaveOccurred())
+
+	// The CA signs its own certificate
+	caCertBytes, err := x509.CreateCertificate(rand.Reader, caCert, caCert, caCert.PublicKey, caKey)
+	Expect(err).NotTo(HaveOccurred())
+
+	expiredcertPEM = pem.EncodeToMemory(&pem.Block{Bytes: certBytes, Type: "CERTIFICATE"})
+	earlyMadeCACertPEM = pem.EncodeToMemory(&pem.Block{Bytes: caCertBytes, Type: "CERTIFICATE"})
+	return
+}
 
 func runTransactions(n *nwo.Network, orderer *nwo.Orderer, peer *nwo.Peer, chaincodeName string, channelID string) {
 	for i := 0; i < 5; i++ {
--- a/integration/gossip/gossip_test.go
+++ b/integration/gossip/gossip_test.go
@@ -9,9 +9,11 @@ package gossip
 import (
 	"crypto/ecdsa"
 	"crypto/rand"
+	"crypto/sha256"
 	"crypto/x509"
 	"encoding/pem"
 	"fmt"
+	"math/big"
 	"os"
 	"path/filepath"
 	"syscall"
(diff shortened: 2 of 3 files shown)


integration/gossip/gossip_test.go
```diff
@@ -285,8 +289,130 @@ var _ = Describe("Gossip State Transfer and Membership", func() {
...
+		channelparticipation.JoinOrdererAppChannel(network, "testchannel", orderer, nwprocs.ordererRunner)
+
+		By("joining peers to channel")
+		network.JoinChannel(channelName, orderer, peer0Org1, peer0Org2)
+
+		By("verifying membership of both peers")
+		bothPeers := []*nwo.Peer{peer0Org1, peer0Org2}
+		network.VerifyMembership(bothPeers, channelName)
+
+		fmt.Println("===MEMBERSHIP VERIFIED===")
```
pfi79, line 320: Maybe here and below instead of 
`fmt.Println("===MEMBERSHIP VERIFIED===")`
use this
`By("===MEMBERSHIP VERIFIED===")`
This kind of thing in the logs is very handy to search for
`STEP: ===MEMBERSHIP VERIFIED=== @ 01/20/25 05:32:41.315`

denyeart: I expect we will remove the Println() statements and some of the sleeps once a fix is available and the test is passing, and keep the existing By() statements. The additional Println() information is useful for now if you are watching the logs during test execution.

yacovm:
Dave, FYI - the test is kind of flaky, at least on my laptop. It fails often before the expiration part. 
If you merge a fix with this test, it will make the integration tests suite flaky.

Now, to the point - Here is what is happening in the "failing" part of the test (after the certificate expires): 

1. The second peer's identity gets expired and the first peer [purges](https://github.com/hyperledger/fabric/blob/main/gossip/identity/identity.go#L134-L142) its from its identity store once its certificate expires.
2. The next time the first peer connects to the second peer, it does so because it remembers its endpoint and its PKI-ID [0]. 
3. When a connection is established to the second peer, gossip's communication layer expects to connect to the former PKI-ID. The code has a [safeguard](https://github.com/hyperledger/fabric/commit/95837c9a49130b9db3c822a2a3229f1e171bf286) that agrees to recognize the change in PKI-ID, as long as the peer (or rather, the endpoint we connect to) didn't change its organization affiliation [1]. However as mentioned before, the peer's membership layer doesn't remember the second peer's organization, only its PKI-ID. It therefore asks the identity store for the organization of the old PKI-ID, but it doesn't have it, as it has purged it from its memory once its x509 certificate has expired. The peer therefore refuses to connect to the new incarnation of the second peer. 

A simple fix would be to naively assume that if the peer's identity was purged from the identity store, to take the new identity returned from the handshake:

```
diff --git a/gossip/comm/comm_impl.go b/gossip/comm/comm_impl.go
index 6a443ae0c..57341bf1f 100644
--- a/gossip/comm/comm_impl.go
+++ b/gossip/comm/comm_impl.go
@@ -173,6 +173,11 @@ func (c *commImpl) createConnection(endpoint string, expectedPKIID common.PKIidT
                                // If the identity isn't present, it's nil - therefore OrgByPeerIdentity would
                                // return nil too and thus would be different than the actual organization
                                identity, _ := c.idMapper.Get(expectedPKIID)
+                               // Identity is no longer present in the identity store, it was probably purged due to being invalidated
+                               if len(identity) == 0 {
+                                       // Optimistically use the new identity of the peer, to merge with the logic flow of this program.
+                                       identity = connInfo.Identity
+                               }
                                oldOrg := c.sa.OrgByPeerIdentity(identity)
                                if !bytes.Equal(actualOrg, oldOrg) {
                                        c.logger.Warning("Remote endpoint claims to be a different peer, expected", expectedPKIID, "but got", pkiID)
```
This fix, however, would expose the peer to a network spoofing attack in case the identity expires and an adversary routes the communication to the endpoint to itself.

[0] It doesn't remember its organization ID, because... well, when this code was written, we didn't have organizations in Fabric (I feel so old writing this...) and we added organizations to gossip only later on. Since the PKI-ID indirectly is also bound to the organization, it's safe to only use the PKI-ID. 

[1] This was put as a safeguard against a network spoofing attack, because the peer gives gossip all TLS CAs across all organizations as a [single bundle](https://github.com/hyperledger/fabric/blob/main/internal/peer/node/start.go#L1193) and not on a per-org basis. 

Another fix which I think is more natural is to fail fast in case the identity is no longer present in the identity store, and let the second peer with the renewed identity connect to the first peer instead with its new PKI-ID and identity.

The fix is as follows:

```
diff --git a/gossip/comm/comm_impl.go b/gossip/comm/comm_impl.go
index 6a443ae0c..f06c1bb84 100644
--- a/gossip/comm/comm_impl.go
+++ b/gossip/comm/comm_impl.go
@@ -162,6 +162,15 @@ func (c *commImpl) createConnection(endpoint string, expectedPKIID common.PKIidT
                return nil, errors.WithStack(err)
        }
 
+
+       if len(expectedPKIID) > 0 {
+               identity, _ := c.idMapper.Get(expectedPKIID)
+               if len(identity) == 0 {
+                       c.logger.Warningf("Identity of %x is no longer found in the identity store, aborting connection", expectedPKIID)
+                       return nil, errors.New("identity no longer recognized")
+               }
+       }
+
        ctx, cancel = context.WithCancel(context.Background())
        if stream, err = cl.GossipStream(ctx); err == nil {
                connInfo, err = c.authenticateRemotePeer(stream, true, false)
diff --git a/integration/gossip/gossip_test.go b/integration/gossip/gossip_test.go
index 5d4c3ce09..ad5365588 100644
--- a/integration/gossip/gossip_test.go
+++ b/integration/gossip/gossip_test.go
@@ -350,7 +350,8 @@ var _ = Describe("Gossip State Transfer and Membership", func() {
                startPeers(nwprocs, false, peer0Org2)
 
                By("ensuring that peer0Org1 replaces peer0Org2 PKI-ID after it expired")
-               Eventually(peer0Org1Runner.Err(), network.EventuallyTimeout).Should(gbytes.Say("changed its PKI-ID from"))
+               time.Sleep(5 * time.Second)
+               network.VerifyMembership(bothPeers, channelName)
        })
 })
```

denyeart:
@yacovm The last fix you mention sounds reasonable.

I think the integration test is flaky because the time for gossip membership establishment varies greatly, sometimes it is almost instant and sometimes it takes a minute. Do you have any ideas on how to make the membership establishment time quicker or more consistent, either thru a config change or a code change?

yacovm:
> @yacovm The last fix you mention sounds reasonable.
> 
> I think the integration test is flaky because the time for gossip membership establishment varies greatly, sometimes it is almost instant and sometimes it takes a minute. Do you have any ideas on how to make the membership establishment time quicker or more consistent, either thru a config change or a code change?

I think it's something with the nwo, because when it flakes, the logs clearly show that both peers are known, yet it continues querying from some odd reason. Something with the matchers is probably wrong.

It's also weird that only one of the peers is queried in `DiscoverPeers`. I would expect both of them to be queried... this needs to be checked out.

integration/gossip/gossip_test.go
```diff
@@ -285,8 +289,137 @@ var _ = Describe("Gossip State Transfer and Membership", func() {
...
+		fmt.Println("===PKI-ID REPLACED===")
+		time.Sleep(5 * time.Second)
+
+		//By("verifying membership after cert renewed")
+		// Discovery check fails since discovered peer's cert is not identical to expected peer's cert (content is the same but signature has changed, perhaps due to how cert is re-created in expireCertificate)
+		// Eventually(
+		//	nwo.DiscoverPeers(network, peer0Org1, "User1", "testchannel"),
+		//	60*time.Second,
+		//	100*time.Millisecond).Should(ContainElements(network.DiscoveredPeer(network.Peer("Org2", "peer0"), "_lifecycle"))
+		// )
```
denyeart: I checked the weird Discovery check behavior... it appears that discovered peer's cert is not identical to expected peer's cert and therefore the matcher fails. The cert content matches expectations, but signature has changed, perhaps due to how cert is re-created in expireCertificate() below.

gossip/comm/comm_impl.go
```diff
@@ -162,6 +162,14 @@ func (c *commImpl) createConnection(endpoint string, expectedPKIID common.PKIidT
...
 	}
 
+	if len(expectedPKIID) > 0 {
+		identity, _ := c.idMapper.Get(expectedPKIID)
+		if len(identity) == 0 {
+			c.logger.Warningf("Identity of %x is no longer found in the identity store, aborting connection", expectedPKIID)
+			return nil, errors.New("identity no longer recognized")
+		}
+	}
+
```
denyeart, line 172: This came close to fixing the problem... membership is indeed established now, but the new warning and one of the existing warnings keeps getting repeated in the logs forever:

```
      2025-01-23 00:59:56.269 EST 00a4 WARN [gossip.gossip] disclosurePolicy -> Cannot determine organization of Endpoint: 127.0.0.1:22510, InternalEndpoint: , PKI-ID: fd8548f3b6fb8164a711b6d6dfa2a55f5ae013f512015f884e9af282ef553aa5, Metadata: 
      2025-01-23 00:59:56.273 EST 00a5 WARN [gossip.comm] createConnection -> Identity of 66643835343866336236666238313634613731316236643664666132613535663561653031336635313230313566383834653961663238326566353533616135 is no longer found in the identity store, aborting connection
      2025-01-23 00:59:56.273 EST 00a6 WARN [gossip.comm] sendToEndpoint -> Failed obtaining connection for 127.0.0.1:22510, PKIid:fd8548f3b6fb8164a711b6d6dfa2a55f5ae013f512015f884e9af282ef553aa5 reason: identity no longer recognized
      2025-01-23 00:59:57.778 EST 00a7 WARN [gossip.gossip] func3 -> Unable to determine org of message tag:EMPTY alive_msg:{membership:{endpoint:"127.0.0.1:22510" pki_id:"\xfd\x85H\xf3\xb6\xfb\x81d\xa7\x11\xb6\xd6ߢ\xa5_Z\xe0\x13\xf5\x12\x01_\x88N\x9a\xf2\x82\xefU:\xa5"} timestamp:{inc_num:1737611884416490000 seq_num:28}}
      2025-01-23 01:00:01.780 EST 00a8 WARN [gossip.gossip] func3 -> Unable to determine org of message tag:EMPTY alive_msg:{membership:{endpoint:"127.0.0.1:22510" pki_id:"\xfd\x85H\xf3\xb6\xfb\x81d\xa7\x11\xb6\xd6ߢ\xa5_Z\xe0\x13\xf5\x12\x01_\x88N\x9a\xf2\x82\xefU:\xa5"} timestamp:{inc_num:1737611884416490000 seq_num:28}}
      2025-01-23 01:00:05.781 EST 00a9 WARN [gossip.gossip] func3 -> Unable to determine org of message tag:EMPTY alive_msg:{membership:{endpoint:"127.0.0.1:22510" pki_id:"\xfd\x85H\xf3\xb6\xfb\x81d\xa7\x11\xb6\xd6ߢ\xa5_Z\xe0\x13\xf5\x12\x01_\x88N\x9a\xf2\x82\xefU:\xa5"} timestamp:{inc_num:1737611884416490000 seq_num:28}}
```

denyeart (commented):
@yacovm I made progress on the fix and the test based on your suggestions but still a few issues remain...

integration/gossip/gossip_test.go
yacovm: Could it be that the cert needs to be sanitized?

gossip/comm/comm_impl.go
yacovm: > This came close to fixing the problem... membership is indeed established now, but the new warning and one of the existing warnings keeps getting repeated in the logs forever:
> 
> ```
>       2025-01-23 00:59:56.269 EST 00a4 WARN [gossip.gossip] disclosurePolicy -> Cannot determine organization of Endpoint: 127.0.0.1:22510, InternalEndpoint: , PKI-ID: fd8548f3b6fb8164a711b6d6dfa2a55f5ae013f512015f884e9af282ef553aa5, Metadata: 
>       2025-01-23 00:59:56.273 EST 00a5 WARN [gossip.comm] createConnection -> Identity of 66643835343866336236666238313634613731316236643664666132613535663561653031336635313230313566383834653961663238326566353533616135 is no longer found in the identity store, aborting connection
>       2025-01-23 00:59:56.273 EST 00a6 WARN [gossip.comm] sendToEndpoint -> Failed obtaining connection for 127.0.0.1:22510, PKIid:fd8548f3b6fb8164a711b6d6dfa2a55f5ae013f512015f884e9af282ef553aa5 reason: identity no longer recognized
>       2025-01-23 00:59:57.778 EST 00a7 WARN [gossip.gossip] func3 -> Unable to determine org of message tag:EMPTY alive_msg:{membership:{endpoint:"127.0.0.1:22510" pki_id:"\xfd\x85H\xf3\xb6\xfb\x81d\xa7\x11\xb6\xd6ߢ\xa5_Z\xe0\x13\xf5\x12\x01_\x88N\x9a\xf2\x82\xefU:\xa5"} timestamp:{inc_num:1737611884416490000 seq_num:28}}
>       2025-01-23 01:00:01.780 EST 00a8 WARN [gossip.gossip] func3 -> Unable to determine org of message tag:EMPTY alive_msg:{membership:{endpoint:"127.0.0.1:22510" pki_id:"\xfd\x85H\xf3\xb6\xfb\x81d\xa7\x11\xb6\xd6ߢ\xa5_Z\xe0\x13\xf5\x12\x01_\x88N\x9a\xf2\x82\xefU:\xa5"} timestamp:{inc_num:1737611884416490000 seq_num:28}}
>       2025-01-23 01:00:05.781 EST 00a9 WARN [gossip.gossip] func3 -> Unable to determine org of message tag:EMPTY alive_msg:{membership:{endpoint:"127.0.0.1:22510" pki_id:"\xfd\x85H\xf3\xb6\xfb\x81d\xa7\x11\xb6\xd6ߢ\xa5_Z\xe0\x13\xf5\x12\x01_\x88N\x9a\xf2\x82\xefU:\xa5"} timestamp:{inc_num:1737611884416490000 seq_num:28}}
> ```

This happens because the peer with the renewed certificate sends the first peer a membership request ("give me all the peers you know about, alive or dead").

The first peer still has the old incarnation of the second peer in its memory. It then asks itself whether it can disclose knowledge of that peer in its membership response, but it cannot determine that because it no longer knows what is the identity corresponding to the PKI-ID of its previous incarnation.

If you run the peer for long enough (I think up to 2 hours?) it will garbage collect that peer's alive messages and it will no longer print that. 

Ideally we would purge everything we know about a peer once its certificate has expired. 
However I consider this a separate problem and it's orthogonal to this issue. 
The fact that the peer issues warnings in the log doesn't mean something is necessarily wrong.

denyeart: There exists functionality to update PKI-ID and to purge as can be seen in the regular (non-expired) cert renewal scenario, e.g.:

```
INFO [gossip.comm] createConnection -> Peer 127.0.0.1:22510 changed its PKI-ID from 3257ef1cb73517a1c07b06c0c8f06903cb537199ffb11ecd23643b900f3f43d2 to 9f3e40a456e9f1a22c73e1206bd78b3b1b0ed7276ffbb37b9de00d23530f0877
INFO [gossip.discovery] purge -> Purging 3257ef1cb73517a1c07b06c0c8f06903cb537199ffb11ecd23643b900f3f43d2 from membership
```

Why wouldn't we do the same in the case of an expired cert renewal?

yacovm: > There exists functionality to update PKI-ID and to purge as can be seen in the regular (non-expired) cert renewal scenario, e.g.:
> 
> ```
> INFO [gossip.comm] createConnection -> Peer 127.0.0.1:22510 changed its PKI-ID from 3257ef1cb73517a1c07b06c0c8f06903cb537199ffb11ecd23643b900f3f43d2 to 9f3e40a456e9f1a22c73e1206bd78b3b1b0ed7276ffbb37b9de00d23530f0877
> INFO [gossip.discovery] purge -> Purging 3257ef1cb73517a1c07b06c0c8f06903cb537199ffb11ecd23643b900f3f43d2 from membership
> ```
> 
> Why wouldn't we do the same in the case of an expired cert renewal?

Yeah I know the code :-) 

I didn't have the time to look whether it's the best option from a code change perspective. 

I'll look at it more carefully when I have time

integration/gossip/gossip_test.go
denyeart: You're right again @yacovm , I updated the test to force Low-S (sanitized) signatures for the renewed certs and now it works consistently.

yacovm:
So I think that just making the discovery module forget about these PKI-IDs is enough. No need to make changes in the communication layer if there won't be a reason for the communication layer to reach out to these PKI-IDs in the first place. 

Take a look at [this commit](https://github.com/yacovm/fabric/commit/b7bf2965b3563561ddd1e28c48cf8b9b47560f66) which consists of my code and your code squashed into one commit.

yacovm:
gossip/gossip/gossip_test.go needs to be reverted, as well as the debug/printstack statements and the integration test line you added.

The API change is a bit hacky, but I think this is the way to go for an easy and simple fix.

denyeart:
> So I think that just making the discovery module forget about these PKI-IDs is enough. No need to make changes in the communication layer if there won't be a reason for the communication layer to reach out to these PKI-IDs in the first place.
> 
> Take a look at [this commit](https://github.com/yacovm/fabric/commit/b7bf2965b3563561ddd1e28c48cf8b9b47560f66) which consists of my code and your code squashed into one commit.

Thanks @yacovm , I've tested your new change in my own reproduction environment and it works. I don't follow exactly how it works yet, but that's mostly because I'm not up to speed on gossip code in general, maybe the addition of some comments in the gossip code would help me understand it better.

BTW with the prior fix that I attempted, those "Unable to determine org" messages kept spamming the log, I waited 24 hours, so I'm definitely glad to see the new fix.

Go ahead and open a pull request. Note that your commit didn't pick up my latest gossip_test.go changes to force the Low-S renewal.

yacovm:
> > So I think that just making the discovery module forget about these PKI-IDs is enough. No need to make changes in the communication layer if there won't be a reason for the communication layer to reach out to these PKI-IDs in the first place.
> > Take a look at [this commit](https://github.com/yacovm/fabric/commit/b7bf2965b3563561ddd1e28c48cf8b9b47560f66) which consists of my code and your code squashed into one commit.
> 
> Thanks @yacovm , I've tested your new change in my own reproduction environment and it works. I don't follow exactly how it works yet, but that's mostly because I'm not up to speed on gossip code in general, maybe the addition of some comments in the gossip code would help me understand it better.
> 
> BTW with the prior fix that I attempted, those "Unable to determine org" messages kept spamming the log, I waited 24 hours, so I'm definitely glad to see the new fix.
> 
> Go ahead and open a pull request. Note that your commit didn't pick up my latest gossip_test.go changes to force the Low-S renewal.

Fine, I opened a [PR](https://github.com/hyperledger/fabric/pull/5120), but can you please take a look at the integration test you wrote? It failed locally for me and I don't want to waste time figuring out why :-)

Issue #5086, opened 2024-12-22
Peer to chaincode optimization


pfi79:
All pr, for the implementation of [0011-peer-chaincode-optimisation.md](https://github.com/hyperledger/fabric-rfcs/blob/main/text/0011-peer-chaincode-optimization.md) will be bound here

PR #5113, opened 2025-01-21
add support WriteBatch operations

pfi79:
https://github.com/hyperledger/fabric/issues/5086

### DIFF
--- a/integration/chaincode/multi/chaincode.go
+++ b/integration/chaincode/multi/chaincode.go
@@ -1,0 +1,99 @@
+/*
+Copyright IBM Corp. All Rights Reserved.
+
+SPDX-License-Identifier: Apache-2.0
+*/
+
+package multi
+
+import (
+	"fmt"
+	"strconv"
+
+	"github.com/hyperledger/fabric-chaincode-go/v2/shim"
+	pb "github.com/hyperledger/fabric-protos-go-apiv2/peer"
+)
+
+type Operations struct{}
+
+// Init initializes chaincode
+// ===========================
+func (t *Operations) Init(_ shim.ChaincodeStubInterface) *pb.Response {
+	return shim.Success(nil)
+}
+
+// Invoke - Our entry point for Invocations
+// ========================================
+func (t *Operations) Invoke(stub shim.ChaincodeStubInterface) *pb.Response {
+	function, args := stub.GetFunctionAndParameters()
+	fmt.Println("invoke is running " + function)
+
+	switch function {
+	case "invoke":
+		startWB, _ := strconv.ParseBool(args[0])
+		if startWB {
+			stub.StartWriteBatch()
+		}
+		if len(args) != 2 {
+			return shim.Error("Incorrect number of arguments. Expecting 1")
+		}
+		return t.put(stub, args[1])
+	case "get-key":
+		if len(args) != 1 {
+			return shim.Error("Incorrect number of arguments. Expecting 1")
+		}
+		return t.getKey(stub, args[0])
+	case "put-private-key":
+		startWB, _ := strconv.ParseBool(args[0])
+		if startWB {
+			stub.StartWriteBatch()
+		}
+		if len(args) != 2 {
+			return shim.Error("Incorrect number of arguments. Expecting 1")
+		}
+		return t.putPrivateKey(stub, args[1])
+	default:
+		// error
+		fmt.Println("invoke did not find func: " + function)
+		return shim.Error("Received unknown function invocation")
+	}
+}
+
+// both params should be marshalled json data and base64 encoded
+func (t *Operations) put(stub shim.ChaincodeStubInterface, numberCallsPut string) *pb.Response {
+	cntCall, _ := strconv.Atoi(numberCallsPut)
+
+	for i := range cntCall {
+		key := "key" + strconv.Itoa(i)
+		err := stub.PutState(key, []byte(key))
+		if err != nil {
+			return shim.Error(err.Error())
+		}
+	}
+	return shim.Success(nil)
+}
+
+func (t *Operations) getKey(stub shim.ChaincodeStubInterface, keyUniq string) *pb.Response {
+	key := "key" + keyUniq
+	resp, err := stub.GetState(key)
+	if err != nil {
+		return shim.Error(err.Error())
+	}
+
+	return shim.Success(resp)
+}
+
+// put to private key with a collection is not exists
+func (t *Operations) putPrivateKey(stub shim.ChaincodeStubInterface, numberCallsPut string) *pb.Response {
+	cntCall, _ := strconv.Atoi(numberCallsPut)
+
+	for i := range cntCall {
+		key := "key" + strconv.Itoa(i)
+		col := "col" + strconv.Itoa(i)
+		err := stub.PutPrivateData(col, key, []byte(key))
+		if err != nil {
+			return shim.Error(err.Error())
+		}
+	}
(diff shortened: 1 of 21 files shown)


pfi79:
In the e2e integration tests, benchmark test has been added and 2 speed tables appear with and without the use of batch.

image

image

integration/e2e/bench_write_batch_test.go
```diff
@@ -0,0 +1,143 @@
...
+				RunInvokeOnlyPutState(network, orderer, peer, "invoke", useWriteBatch, 10000)
+			}, gmeasure.SamplingConfig{N: 10})
+		})
+	},
+		Entry("without write batch", "without write batch", false),
+		Entry("with write batch", "with write batch", true),
+	)
+})
+
+func RunInvokeOnlyPutState(n *nwo.Network, orderer *nwo.Orderer, peer *nwo.Peer, fn string, startBatch bool, initial int) {
```
denyeart: Can we use a more expressive variable name than `initial`? It is not obvious what this is...

And shouldn't `startBatch` really be called `useWriteBatch` to be consistent?

integration/chaincode/multi/chaincode.go
```diff
@@ -0,0 +1,61 @@
...
+	fmt.Println("invoke is running " + function)
+
+	startWB, _ := strconv.ParseBool(args[0])
+	if startWB {
+		stub.StartWriteBatch()
+	}
+
+	switch function {
+	case "invoke":
+		return t.put(stub, args[1:])
```
denyeart: Only a single argument is passed right? e.g. `args[1]` , better to use a variable with an expressive name to make it clear what exactly is being passed here.

integration/e2e/bench_write_batch_test.go
```diff
@@ -0,0 +1,143 @@
...
+			}, gmeasure.SamplingConfig{N: 10})
+
+			experiment.SampleDuration("invoke N-10 cycle-10000", func(idx int) {
+				RunInvokeOnlyPutState(network, orderer, peer, "invoke", useWriteBatch, 10000)
+			}, gmeasure.SamplingConfig{N: 10})
+		})
+	},
+		Entry("without write batch", "without write batch", false),
+		Entry("with write batch", "with write batch", true),
+	)
```
denyeart: It would be good to read at least one of the keys in a chaincode query after writing the transaction, to ensure end-to-end behavior is consistent regardless of `useWriteBatch`.

sampleconfig/core.yaml
```diff
@@ -659,6 +659,14 @@ chaincode:
     # Format for the chaincode container logs
     format: "%{color}%{time:2006-01-02 15:04:05.000 MST} [%{module}] %{shortfunc} -> %{level:.4s} %{id:03x}%{color:reset} %{message}"
 
+    # AdditionalParams section for parameters that are passed to chaincode
+    # to specify additional operation modes in which the peer operates.
+    additionalParams:
```
denyeart: Maybe `runtimeParams` would be a better name than `additionalParams` in the config.
I realize the proto is called ChaincodeAdditionalParams, but we can give it a more expressive name in the context of the config.

pfi79: done

integration/e2e/bench_write_batch_test.go
pfi79: No. In this case it is passing a sign that the chaincode should call startBatch or not.

```
	startWB, _ := strconv.ParseBool(args[0])
	if startWB {
		stub.StartWriteBatch()
	}
```

pfi79: I'm gonna run these tests

| useWriteBatch (on peer) | StartWriteBatch (on chaincode) |
| --- | --- |
| true | false |
| false | true |
| true | true |

integration/chaincode/multi/chaincode.go
pfi79: done

integration/e2e/bench_write_batch_test.go
pfi79: done

integration/e2e/bench_write_batch_test.go
pfi79: done

C0rWin (commented):
Overall looks good to me, while I think we should take care to add tests to make sure behaviour remain consistent as @manish-sethi mentioned is his comment https://github.com/hyperledger/fabric-rfcs/pull/58#issuecomment-2458605718

pfi79:
> Overall looks good to me, while I think we should take care to add tests to make sure behaviour remain consistent as @manish-sethi mentioned is his comment [hyperledger/fabric-rfcs#58 (comment)](https://github.com/hyperledger/fabric-rfcs/pull/58#issuecomment-2458605718)

Thanks @C0rWin 
I'm willing to finish any tests.
But I'm not quite sure what those tests should be.
If they give me scenarios, I'd be happy to implement them.

C0rWin:
> > Overall looks good to me, while I think we should take care to add tests to make sure behaviour remain consistent as @manish-sethi mentioned is his comment [hyperledger/fabric-rfcs#58 (comment)](https://github.com/hyperledger/fabric-rfcs/pull/58#issuecomment-2458605718)
> 
> Thanks @C0rWin I'm willing to finish any tests. But I'm not quite sure what those tests should be. If they give me scenarios, I'd be happy to implement them.

One of the possible scenarios related to the update of the private data is where there is a check whenever a collection with a given name exists, see https://github.com/hyperledger/fabric/blob/6c2e87c0824604c7cd7a8f4f93b2c5876af76875/core/ledger/kvledger/txmgmt/txmgr/tx_simulator.go#L105... The test should, for example, check what happens when you start batch where you are trying to write into the non-existing collection, without batch it's resulted with error, need to check this behavior is consistent.

Another example would be to check whenever chaincode does write into private data and then query private data it supposes to result in an error based on this check, see here: https://github.com/hyperledger/fabric/blob/6c2e87c0824604c7cd7a8f4f93b2c5876af76875/core/ledger/kvledger/txmgmt/txmgr/tx_simulator.go#L234

pfi79:
> > > Overall looks good to me, while I think we should take care to add tests to make sure behaviour remain consistent as @manish-sethi mentioned is his comment [hyperledger/fabric-rfcs#58 (comment)](https://github.com/hyperledger/fabric-rfcs/pull/58#issuecomment-2458605718)
> > 
> > 
> > Thanks @C0rWin I'm willing to finish any tests. But I'm not quite sure what those tests should be. If they give me scenarios, I'd be happy to implement them.
> 
> One of the possible scenarios related to the update of the private data is where there is a check whenever a collection with a given name exists, see
> 
> https://github.com/hyperledger/fabric/blob/6c2e87c0824604c7cd7a8f4f93b2c5876af76875/core/ledger/kvledger/txmgmt/txmgr/tx_simulator.go#L105
> 
> ... The test should, for example, check what happens when you start batch where you are trying to write into the non-existing collection, without batch it's resulted with error, need to check this behavior is consistent.
> Another example would be to check whenever chaincode does write into private data and then query private data it supposes to result in an error based on this check, see here:
> 
> https://github.com/hyperledger/fabric/blob/6c2e87c0824604c7cd7a8f4f93b2c5876af76875/core/ledger/kvledger/txmgmt/txmgr/tx_simulator.go#L234

@C0rWin @denyeart 
Added the right tests.
Please take a look at my proposal.

denyeart:
LGTM

C0rWin (approved):

PR #5120, opened 2025-01-26
Properly handle expired identities in gossip

yacovm:
When a peer's certificate expires, gossip still retains past messages it has sent, and gossips them to other peers.

Aside from peers doing redundant work, this also impairs their connectivity to the peer with the renewed certificate.

The reason is that peers try connect to the peer of the renewed certificate but abort because they cannot find its (old) PKI-ID in the identity store, which purged its old PKI-ID once its certificate has expired.

This commit fixes this problem by making the peer forget about peers that their identities have been purged from the identity store.

### DIFF
--- a/gossip/comm/comm.go
+++ b/gossip/comm/comm.go
@@ -45,7 +45,7 @@ type Comm interface {
 	PresumedDead() <-chan common.PKIidType
 
 	// IdentitySwitch returns a read-only channel about identity change events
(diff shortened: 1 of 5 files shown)


denyeart:
Ok, I've fixed the integration test (worked around it), but wasn't able to push to this branch, I've opened yet another PR for it - https://github.com/hyperledger/fabric/pull/5121

PR #5122, opened 2025-01-27
Properly handle expired identities in gossip (release-2.5)

denyeart:
When a peer's certificate expires, gossip still retains past messages it has sent, and gossips them to other peers.

Aside from peers doing redundant work, this also impairs their connectivity to the peer with the renewed certificate.

The reason is that peers try connect to the peer of the renewed certificate but abort because they cannot find its (old) PKI-ID in the identity store, which purged its old PKI-ID once its certificate has expired.

This commit fixes this problem by making the peer forget about peers that their identities have been purged from the identity store.

### DIFF
--- a/gossip/comm/comm.go
+++ b/gossip/comm/comm.go
@@ -45,7 +45,7 @@ type Comm interface {
 	PresumedDead() <-chan common.PKIidType
 
 	// IdentitySwitch returns a read-only channel about identity change events
-	IdentitySwitch() <-chan common.PKIidType
+	IdentitySwitch() chan common.PKIidType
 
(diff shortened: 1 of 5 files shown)


yacovm (dismissed):

denyeart:
@yacovm I had to update the integration tests for different timings in release-2.5 to avoid flakes.
This should be the last approval needed for the issue.
Also made main branch consistent in https://github.com/hyperledger/fabric/pull/5123

Finally, I've also opened PRs for v2.5.11 release to deliver the fix - #5124 and #5125

yacovm (approved):

PR #5128, opened 2025-01-31
Release action - Separate runner for amd64 and arm64

denyeart:
Previously arm64 images were created on a standard amd64 runner using qemu.
qemu emulation is very slow and there have been recent issues.
Native image creation is faster and more reliable.

This change creates separate runners for amd64 and arm64.
Also updates to ubuntu-22.04 since the new arm64 runners are only available for ubuntu-22.04 and ubuntu-24.04.

A new job is then needed to combine the image digests into a multi-architecture manifest that gets pushed to Docker Hub and ghcr.

This PR is against release-2.5 branch so that release v2.5.11 can be created. Upon success the change will be cherry-picked to main branch.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -10,7 +10,7 @@ on:
 
 env:
   GO_VER: 1.23.5
-  UBUNTU_VER: 20.04
+  UBUNTU_VER: 22.04
   FABRIC_VER: ${{ github.ref_name }}
 
 permissions:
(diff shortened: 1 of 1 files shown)


denyeart:
@hyperledger/fabric-core-maintainers 
Not many of the maintainers have github action experience to review this one, but it is well tested and blocking our v2.5.11 release (qemu cross-architecture image build issue). Please take a look so that we can release!

pfi79 (approved):

Issue #5119, opened 2025-01-24
Peer panic from `fabric/gossip/util`


TaoGunner:
### Description

1-2 times a month my peers randomly crashed with the message `invalid memory address or nil pointer dereference`.
The system uses more than 50 peers and 50 channels, so exchanges through gossip are very active.

Peers, Orderers v2.5.5

The section of code where the bug appears clearly requires adding a check for the existence of a map element.

**Panic example 1**
```
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x8 pc=0x151112d]

goroutine 32595521 [running]:
github.com/hyperledger/fabric/gossip/util.(*Set).Remove(0x0, {0x191b6e0, 0xc0051e4020})
        /gossip/util/misc.go:118 +0x2d
github.com/hyperledger/fabric/gossip/util.(*PubSub).unSubscribe(0xc002b89840, 0xc0051e4020)
        /gossip/util/pubsub.go:115 +0x8d
github.com/hyperledger/fabric/gossip/util.(*PubSub).Subscribe.func1()
        /gossip/util/pubsub.go:107 +0x1b
created by time.goFunc
        /usr/local/go/src/time/sleep.go:176 +0x2d
```

**Panic example 2**
```
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x8 pc=0x151112d]

goroutine 1909191 [running]:
github.com/hyperledger/fabric/gossip/util.(*Set).Remove(0x0, {0x191b6e0, 0xc0064e7fa0})
        /gossip/util/misc.go:118 +0x2d
github.com/hyperledger/fabric/gossip/util.(*PubSub).unSubscribe(0xc005aaf980, 0xc0064e7fa0)
        /gossip/util/pubsub.go:115 +0x8d
github.com/hyperledger/fabric/gossip/util.(*PubSub).Subscribe.func1()
        /gossip/util/pubsub.go:107 +0x1b
created by time.goFunc
        /usr/local/go/src/time/sleep.go:176 +0x2d
```

### Steps to reproduce

1. Make more then 50 peers.
2. Make more then 50 channels.
3. idk what else (It happens unpredictably).

yacovm:
This looks like a race that was caused by a topic collision.

The below code should fix it. Do you want to give it a try?

```
[...]

PR #5138, opened 2025-02-08
fix peer panic from fabric/gossip/util

pfi79:
https://github.com/hyperledger/fabric/issues/5119


denyeart:
Wow, good catch, thanks for doing the backport to release-2.5 as well!

denyeart (approved):

PR #5146, opened 2025-02-12
Default to GHCR instead of DockerHub for install

### LINKED ISSUE #5145: Pull Docker images from GitHub Container Registry in preference to DockerHub
### Current Status

DockerHub has imposed more restrictive limits on image pulls, and has also raised costs for business users. This makes DockerHub a less desirable default platform from which to pull Docker images during Fabric installation. GitHub Container Registry (GHCR) remains freely available, and recent Fabric releases publish Docker images there in addition to DockerHub.

### Expected

Use GHCR to pull Docker images.

### Solution

Where Docker image versions are available from GHCR, the Fabric install script should pull from there. For versions that are not available from GHCR, the install script should fall back to DockerHub.

The install script should continue to honour the Docker registry override using environment variables with no change in this behaviour.

Yes!

bestbeforetoday:
DockerHub has imposed more restrictive image pull limits and costs for business users. GitHub Container Registry (GHCR) remains freely accessible.

This change updates the Fabric install script to use GHCR to pull Docker images, where those image versions are available. For image versions not available on GHCR, fall back to DockerHub.

Closes #5145

### DIFF
--- a/scripts/install-fabric.sh
+++ b/scripts/install-fabric.sh
@@ -24,14 +24,15 @@ _arg_comp=('' )
 _arg_fabric_version="2.5.11"
 _arg_ca_version="1.5.15"
 
-REGISTRY=${FABRIC_DOCKER_REGISTRY:-docker.io/hyperledger}
-
 OS=$(uname -s|tr '[:upper:]' '[:lower:]'|sed 's/mingw64_nt.*/windows/')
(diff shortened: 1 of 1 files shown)


denyeart (approved):
It works! Thank you. Living in the 2020s!

PR #5148, opened 2025-02-14
Run vulnerability scan on latest release tags

bestbeforetoday:
Vulnerability scans were previously run only on the latest state of currently developed branches. This provided assurance that the current branch state did not contain known vulnerabilities in dependencies, but did not provide assurance that the currently released code was free of vulnerabilities.

This change runs additional vulnerability scans on the most recent release version tag for currently developed branches. Scan failures now indicate that a new release is required to address vulnerabilities in dependencies.

### DIFF
--- a/.github/workflows/vulnerability-scan.yml
+++ b/.github/workflows/vulnerability-scan.yml
@@ -13,7 +13,7 @@ permissions:
   contents: read
 
 jobs:
-  scan:
+  latest:
     # Only run the scheduled job in hyperledger/fabric repository, not on personal forks
     if: github.event_name != 'schedule' || (github.event_name == 'schedule' && github.repository == 'hyperledger/fabric')
     runs-on: ${{ github.repository == 'hyperledger/fabric' && 'fabric-ubuntu-22.04' || 'ubuntu-22.04' }}
@@ -24,7 +24,8 @@ jobs:
           - main
           - release-2.5
     steps:
-      - uses: actions/checkout@v4
+      - name: Checkout ${{ matrix.ref }} branch
+        uses: actions/checkout@v4
         with:
           ref: ${{ matrix.ref }}
       - name: Set up Go
@@ -32,4 +33,34 @@ jobs:
         with:
           go-version: 1.24.0
       - name: Scan
-        run: make scan
+        run: go run github.com/google/osv-scanner/cmd/osv-scanner@b37c83e19af3b2555864457cbd0b08ef0e1f9d7d scan --lockfile=go.mod || (( $? > 1 && $? < 127 ))
+
+  release:
+    # Only run the scheduled job in hyperledger/fabric repository, not on personal forks
+    if: github.event_name != 'schedule' || (github.event_name == 'schedule' && github.repository == 'hyperledger/fabric')
+    runs-on: ${{ github.repository == 'hyperledger/fabric' && 'fabric-ubuntu-22.04' || 'ubuntu-22.04' }}
+    strategy:
+      fail-fast: false
+      matrix:
+        ref:
+          - main
+          - release-2.5
+    steps:
+      - name: Checkout ${{ matrix.ref }} branch
+        uses: actions/checkout@v4
+        with:
+          ref: ${{ matrix.ref }}
+          fetch-depth: 0
+      - name: Get latest release tag
+        id: latest-release
+        run: echo "tag=$(git -c versionsort.suffix=- tag --merged HEAD --sort=-version:refname | head -1)" >> "${GITHUB_OUTPUT}"
+      - name: Checkout ${{ steps.latest-release.outputs.tag }}
+        uses: actions/checkout@v4
+        with:
+          ref: ${{ steps.latest-release.outputs.tag }}
+      - name: Set up Go
+        uses: actions/setup-go@v5
+        with:
+          go-version: 1.24.0
+      - name: Scan
+        run: go run github.com/google/osv-scanner/cmd/osv-scanner@b37c83e19af3b2555864457cbd0b08ef0e1f9d7d scan --lockfile=go.mod || (( $? > 1 && $? < 127 ))
--- a/Makefile
+++ b/Makefile
@@ -106,19 +106,19 @@ help: ## List all commands with documentation
 	@awk 'BEGIN {FS = ":.*?## "}; /^[a-zA-Z_-]+:.*?## / {printf "\033[36m%-15s\033[0m %s\n", $$1, $$2}' $(MAKEFILE_LIST)
 
 .PHONY: all
-all: ## Builds all targets and runs all non-integration tests/checks 
+all: ## Builds all targets and runs all non-integration tests/checks
 	check-go-version native docker checks
 
 .PHONY: checks
-checks: ## Runs basic checks along with unit and integration tests 
+checks: ## Runs basic checks along with unit and integration tests
 	basic-checks unit-test integration-test
 
 .PHONY: basic-checks
-basic-checks: check-go-version license spelling references trailing-spaces linter check-help-docs check-metrics-doc filename-spaces check-swagger ## Performs basic checks like license, spelling, trailing spaces and linter 
+basic-checks: check-go-version license spelling references trailing-spaces linter check-help-docs check-metrics-doc filename-spaces check-swagger ## Performs basic checks like license, spelling, trailing spaces and linter
 
 
 .PHONY: desk-checks
-desk-check: ## Runs linters and verify to test changed packages 
+desk-check: ## Runs linters and verify to test changed packages
 	checks verify
 
 .PHONY: help-docs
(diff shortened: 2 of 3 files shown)


.github/workflows/vulnerability-scan.yml
```diff
@@ -24,12 +24,23 @@ jobs:
...
         with:
           ref: ${{ matrix.ref }}
+          fetch-depth: 0
+      - name: Get latest release tag
+        id: latest-release
+        run: echo "tag=$(git -c versionsort.suffix=- tag --merged HEAD --sort=-version:refname | head -1)" >> "${GITHUB_OUTPUT}"
+      - name: Checkout ${{ steps.latest-release.outputs.tag }}
+        uses: actions/checkout@v4
+        with:
+          ref: ${{ steps.latest-release.outputs.tag }}
```
denyeart, line 60: I see the value in running against the latest tags. But we also want to know if the current branches have all vulnerabilities resolved, especially heading into an upcoming release. So maybe we need to run for BOTH current branch AND latest tags?

.github/workflows/vulnerability-scan.yml
```diff
@@ -24,12 +24,23 @@ jobs:
...
       - name: Set up Go
         uses: actions/setup-go@v5
         with:
-          go-version: 1.24.0
+          # Always use the latest Go release to avoid false positives from older
+          # versions of the Go standard library
+          go-version: stable
```
denyeart: Shouldn't we run with the same Go version that the rest of the repository uses? To know if we need to update Go to address vulnerabilities.

pfi79: I'll support @denyeart . Go should be the one you build and test with.

pfi79:
I don't like that you put everything in one action. 
Try to consider changing the ‘Security vulnerability scan’ so that you can put different badges for each branch and last tags in the [Readme](https://github.com/hyperledger/fabric/blob/main/README.md).

bestbeforetoday:
> I don't like that you put everything in one action. Try to consider changing the ‘Security vulnerability scan’ so that you can put different badges for each branch and last tags in the [Readme](https://github.com/hyperledger/fabric/blob/main/README.md).

This PR doesn't introduce or change that behaviour. It might be good to capture this requirement in an issue so it isn't forgotten.

.github/workflows/vulnerability-scan.yml
bestbeforetoday: Typically I don't worry too much about dependency versions until there is either a vulnerability in dependencies for the current release or I am about to make a new release. In both cases I usually look to bump up all dependency versions. While I'm doing that I can run a scan manually.

There is little effort in also scanning the current code base too though. No harm in having it.

.github/workflows/vulnerability-scan.yml
bestbeforetoday: Since Fabric is publishing binaries rather than being consumed as a library, that is a good point. However, that makes me think that the way Fabric is handling Go versions is wrong. Using the Go version _currently_ being used to build Fabric doesn't help identify standard library issues when scanning _previously released_ versions. Probably the go.mod file should be specifying an explicit toolchain version. This then allows the correct Go version for a given release to be used when retrospectively scanning for vulnerabilities. It also removes the need to keep updating specific Go versions across all the GitHub Actions workflows. They can just use `stable` and the Go toolchain version will actually be used. What do you think?

pfi79:
> > I don't like that you put everything in one action. Try to consider changing the ‘Security vulnerability scan’ so that you can put different badges for each branch and last tags in the [Readme](https://github.com/hyperledger/fabric/blob/main/README.md).
> 
> This PR doesn't introduce or change that behaviour. It might be good to capture this requirement in an issue so it isn't forgotten.

I don't agree. It does.

Right now the badge shows the vulnerability of what's in the main on top.
```
[![Security vulnerability scan](https://github.com/hyperledger/fabric/actions/workflows/vulnerability-scan.yml/badge.svg?branch=main)](https://github.com/hyperledger/fabric/actions/workflows/vulnerability-scan.yml)
```

After your changes, it will show that one or more of the 4 builds is corrupted.

And we have 2 different badges for main and release-2.5 on passing tests

.github/workflows/vulnerability-scan.yml
denyeart: I'm not sure I'm on board yet.

From https://go.dev/doc/toolchain:

> When the go or toolchain line is newer than the bundled toolchain, the go command runs the newer toolchain instead.

This tells me if the installed (bundled) version from the workflow is later than `toolchain` version, it will use installed (bundled) version. This seems to contradict your proposal of using `stable` for the workflows and a `toolchain` version in go.mod, that is, I don't think it will actually report vulnerabilities in prior release's `toolchain` as you were hoping.

Also: 

> "A go.mod that says go 1.21.0 with no toolchain line is interpreted as if it had a toolchain go1.21.0 line."

Because of this, we had decided to switch the go.mod `go` version to the three digit version, and not set the `toolchain` version at all. Just let the `go` version dictate the `toolchain` version implicitly (fewer moving parts to manage and less to think about). A matching workflow Go version and go.mod `go` version seemed the most clear and locked-down, with fewest considerations and edge case to think about.

In summary, I do see some value in your proposal to scan with prior release's Go version, however I'm not sure it actually will do it, or if it is compelling enough to change our current approach. I may have misinterpreted something however...

bestbeforetoday: Specifying the toolchain either with an explicit toolchain directive or using the go directive is fine. Keeping the version of Go used to run the Fabric build in step with the version in the go directive also seems fine to me. Unfortunately, the version in the go.mod file has not kept in step with the version used in the GitHub Actions workflows.

For libraries, I like keeping the go/toolchain directive in the go.mod file to the lowest supported Go version (including patch level) since this imposes fewest restrictions on consumers, who are then free to use a more up-to-date  Go version to avoid standard library vulnerabilities. For a binary, it might make more sense for the go/toolchain directive to match the version of Go used to build.

bestbeforetoday:
I have updated this change to:

1. Retain the scan of the current branch state, with an additional scan of the last release versions.
2. Use OSV-Scanner instead of govulncheck. This uses govulncheck internally, but scans standard libraries at the Go toolchain version. Provided the go/toolchain directive in the go.mod file indicates the version used to build, this should give an accurate assessment of vulnerabilities present in the prebuilt binaries.

.github/workflows/vulnerability-scan.yml
```diff
@@ -24,12 +24,43 @@ jobs:
...
-      - uses: actions/checkout@v4
+      - name: Checkout ${{ matrix.ref }} branch
+        uses: actions/checkout@v4
         with:
           ref: ${{ matrix.ref }}
       - name: Set up Go
         uses: actions/setup-go@v5
         with:
-          go-version: 1.24.0
+          go-version: stable
```
pfi79: We have everywhere the version of go in go.mod matches the version of go we build images with and test them with.

Your last change may lead to vulnerability checking with a version of go that fabric doesn't support yet.

bestbeforetoday: > We have everywhere the version of go in go.mod matches the version of go we build images with and test them with.

v3.0.0 release matches:

- [.github/workflows/release.yml](https://github.com/hyperledger/fabric/blob/v3.0.0/.github/workflows/release.yml) says `1.23.1`.
- [go.mod](https://github.com/hyperledger/fabric/blob/v3.0.0/go.mod) says `1.23.1`.

v2.5.11 release does **not** match:

- [.github/workflows/release.yml](https://github.com/hyperledger/fabric/blob/v2.5.11/.github/workflows/release.yml) says `1.23.5`
- [go.mod](https://github.com/hyperledger/fabric/blob/v2.5.11/go.mod) says `1.23.1`.

Go 1.23.1 contains standard library vulnerabilities that are used by Fabric, so v3.0.0 contains vulnerabilities our current scanning is blissfully unaware of.

> Your last change may lead to vulnerability checking with a version of go that fabric doesn't support yet.

The version of Go in the workflow is used only to run the osv-scanner tool. It has no relation to the version of Go used by Fabric.

pfi79: As a developer, I'm interested in the vulnerabilities that are currently in the last commit of the main branch. This check should be done with the version of go that is being tested, not the stable version.

bestbeforetoday: To repeat what I've said before, the version of Go used for the scan is the one specified as the toolchain in the go.mod file, not the one in this workflow file. I was trying to save you some work keeping versions updated by just using the latest Go version here since it makes no difference to the scan results. It is fine by me if you prefer to keep maintaining the version in this file, so I've updated the PR as you requested,

denyeart (approved):
Looks good now, I'll merge and give it a run to verify all is good.

denyeart:
@bestbeforetoday
@pfi79 
The updated action is reporting failure now for latest releases.

This is expected for v3.0.0, I'll plan to resolve by doing a v3.0.1 release.

For v2.5.11 I think we are actually ok since the release builds actually used Go v1.23.5 (I had forgotten to update go.mod so it is still being scanned at Go v1.23.1).

Going forward we will keep the Go versions in sync, and when we notice a last release turns red we can make sure dependencies are up to date and then cut the next patch release.

Make sense?

PR #5152, opened 2025-02-23
Update go version in go.mod to match build

bestbeforetoday:
The build workflows use Go 1.23.5. The go.mod specified 1.23.1. This change updates go.mod to match the Go version used by the build.


denyeart (approved):
Thanks Mark, I do try to keep them in sync but I missed this file last time!

PR #5155, opened 2025-02-25
Use Go version from go.mod in CI workflows

bestbeforetoday:


.github/workflows/verify-build.yml
```diff
@@ -30,10 +29,7 @@ jobs:
       - uses: actions/setup-go@v5
         name: Install Go
         with:
-          go-version: ${{ env.GO_VER }}
-      - name: Check Go module version
-        run: |
-          [[ "$(go list -m -f '{{.Version}}' go)" == "${GO_VER}" ]]
```
bestbeforetoday, line 36: Check that the CI and go.mod versions match is no longer required since the CI uses the go.mod version.

.github/workflows/release.yml
```diff
@@ -94,10 +92,13 @@ jobs:
...
     steps:
-
       - name: Checkout
         uses: actions/checkout@v4
-        
+
+      - name: Set GO_VER environment variable from go.mod
+        run: |
+          awk '/^go[ /t]/ { gsub(/^go[ \t]+|[ \t]+^/, ""); print "GO_VER="$0; exit }' < go.mod >> "${GITHUB_ENV}"
+
```
bestbeforetoday, line 101: Alternative approach would be to set up go using actions/setup-go and the go.mod file, then read the `go-version` output from that action.

bestbeforetoday:
@denyeart Using the Go version from go.mod was easier than I thought since actions/setup-go provides a [go-version-file](https://github.com/actions/setup-go?tab=readme-ov-file#getting-go-version-from-the-gomod-file) input option.

This change could use some additional eyes to see if I missed anything. And perhaps a trial of the release workflow since that uses some scripting magic.

.github/workflows/release.yml
denyeart: This approach looks good.

denyeart (approved):
Great to see the go-version-file option!

This looks good, I just did a test of the release from my personal fork and GO_VER does indeed get picked up correctly.

PR #5172, opened 2025-03-08
refactor: return nil when error has already been checked

MarkDaveny:
#### Type of change

#### Description

Since err != nil has been judged before, nil is returned directly here, which is more obvious, readable and less error-prone.


denyeart (approved):
Thanks for the improvement!

PR #5186, opened 2025-03-20
Use OSV-Scanner v2 for vulnerability scan

bestbeforetoday:
Replace use of OSV-Scanner v1 with v2.

### DIFF
--- a/.github/workflows/vulnerability-scan.yml
+++ b/.github/workflows/vulnerability-scan.yml
@@ -33,7 +33,7 @@ jobs:
         with:
           go-version-file: go.mod
       - name: Scan
-        run: go run github.com/google/osv-scanner/cmd/osv-scanner@b37c83e19af3b2555864457cbd0b08ef0e1f9d7d scan --lockfile=go.mod || (( $? > 1 && $? < 127 ))
+        run: go run github.com/google/osv-scanner/v2/cmd/osv-scanner@latest scan --lockfile=go.mod || (( $? > 1 && $? < 127 ))
 
   release:
     # Only run the scheduled job in hyperledger/fabric repository, not on personal forks
@@ -63,4 +63,4 @@ jobs:
         with:
           go-version-file: go.mod
       - name: Scan
-        run: go run github.com/google/osv-scanner/cmd/osv-scanner@b37c83e19af3b2555864457cbd0b08ef0e1f9d7d scan --lockfile=go.mod || (( $? > 1 && $? < 127 ))
+        run: go run github.com/google/osv-scanner/v2/cmd/osv-scanner@latest scan --lockfile=go.mod || (( $? > 1 && $? < 127 ))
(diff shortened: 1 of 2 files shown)


.github/workflows/vulnerability-scan.yml
```diff
@@ -33,7 +33,7 @@ jobs:
         with:
           go-version-file: go.mod
       - name: Scan
-        run: go run github.com/google/osv-scanner/cmd/osv-scanner@b37c83e19af3b2555864457cbd0b08ef0e1f9d7d scan --lockfile=go.mod || (( $? > 1 && $? < 127 ))
+        run: go run github.com/google/osv-scanner/v2/cmd/osv-scanner@latest scan --lockfile=go.mod || (( $? > 1 && $? < 127 ))
```
pfi79, line 36: Why don't you want to use https://github.com/google/osv-scanner-action?

For example:
```
  osv-scanner:
    uses: "google/osv-scanner-action/.github/workflows/osv-scanner-reusable.yml@main"
```

bestbeforetoday: I think the v2 action is still in development.

pfi79: Ok.
But developers recommend.
https://github.com/google/osv-scanner/issues/1682

osv-scanner v2 appeared 4 days ago, before that it was also in development, but we used it.

bestbeforetoday: Fabric did not use OSV-Scanner v2 before this PR.

pfi79: I mean, before this, the fabric used not a tag but a commit hash. That is, a version that is in development. That is, not a release.

bestbeforetoday: A commit hash can equally point to a specific release version. You are correct that the current hash is not a release version though. It was required to get a version that included a fix for call analysis of CGO applications to avoid scan issues with the Fabric codebase. This was not available in a v1 release and the next release of OSV-Scanner containing this fix is v2.0.0.

bestbeforetoday: The objective of this PR is to update to use OSV-Scanner v2. If you disagree with using the latest release version of OSV-Scanner, we can close this PR. If you see an issue with the implementation that means it does not successfully use OSV-Scanner v2, that is helpful review feedback and I can address that.

pfi79: See https://github.com/google/osv-scanner-action/blob/main/osv-scanner-action/action.yml or https://github.com/google/osv-scanner-action/pull/64/files. The osv-scanner-action from main (yes, until the tag is released) will run osv-scanner v2.0.0
So why not use it?

bestbeforetoday: That is not something I have time for right now. A PR is not an invitation for you to assign additional work to people. If you want to use the action, you should submit a PR to do that.

denyeart (approved):
My suggestion - let's go ahead and make this simple update.

We can consider osv-scanner-action in a subsequent PR.

denyeart (approved):

PR #5188, opened 2025-03-24
NixOS support (Multi-Linux Distro Support)

AadithyanRaju:
NixOS support (Multi-Linux Distro Support)

#### Type of change

- Bug fix

#### Description

In nixos, there is no /bin/bash, so it fails. So had to change the shebang which work in any environment.
Tested it in NixOS and Nix Docker container.


denyeart (approved):
Thank you!

PR #5214, opened 2025-05-04
goroutine leak fixed

pfi79:
Because grpc clients are not closed, there is a leak of goroutines.
This change fixes this

### DIFF
--- a/common/deliverclient/blocksprovider/bft_header_receiver.go
+++ b/common/deliverclient/blocksprovider/bft_header_receiver.go
@@ -32,6 +32,7 @@ type BFTHeaderReceiver struct {
 	errorStopTime          time.Time
 	endpoint               string
 	client                 orderer.AtomicBroadcast_DeliverClient
+	clientCloserFunc       func()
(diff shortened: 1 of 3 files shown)


common/deliverclient/blocksprovider/bft_header_receiver.go
```diff
@@ -186,8 +189,7 @@ func (hr *BFTHeaderReceiver) Stop() error {
 
 	hr.logger.Infof("[%s][%s] Stopping", hr.chainID, hr.endpoint)
```
yacovm, line 190: well at least we have a stop function :-)

yacovm (approved):

pfi79:
@yacovm @C0rWin @denyeart 
Recently fixed 2 serious bugs: goroutine leak and deadlock in SmartBFT.
Please consider releasing a new version v3.1.1

tock-ibm:
@pfi79 Thanks for picking that up! At least I left TODOs in the right places ;-)

PR #5222, opened 2025-05-17
refactor: use slices.Contains to simplify code

pengqiseven:
#### Type of change

#### Description

use slices.Contains to simplify code

### DIFF
--- a/gossip/gossip/msgstore/msgs_test.go
+++ b/gossip/gossip/msgstore/msgs_test.go
@@ -9,6 +9,7 @@ package msgstore
 import (
 	crand "crypto/rand"
 	"math/rand/v2"
+	"slices"
 	"sync"
 	"sync/atomic"
 	"testing"
@@ -78,17 +79,8 @@ func TestNewMessagesInvalidates(t *testing.T) {
 }
 
 func TestMessagesGet(t *testing.T) {
-	contains := func(a []interface{}, e interface{}) bool {
-		for _, v := range a {
-			if v == e {
-				return true
-			}
-		}
-		return false
-	}
-
 	msgStore := NewMessageStore(alwaysNoAction, Noop)
-	expected := []int{}
+	expected := make([]interface{}, 0)
 	for i := 0; i < 2; i++ {
 		n := r.Int()
 		expected = append(expected, n)
@@ -96,7 +88,7 @@ func TestMessagesGet(t *testing.T) {
 	}
 
 	for _, num2Search := range expected {
-		require.True(t, contains(msgStore.Get(), num2Search), "Value %v not found in array", num2Search)
+		require.True(t, slices.Contains(msgStore.Get(), num2Search), "Value %v not found in array", num2Search)
 	}
 }
 
--- a/gossip/util/misc.go
+++ b/gossip/util/misc.go
@@ -31,16 +31,6 @@ type Equals func(a interface{}, b interface{}) bool
 
 var viperLock sync.RWMutex
 
-// Contains returns whether a given slice a contains a string s
-func Contains(s string, a []string) bool {
-	for _, e := range a {
-		if e == s {
-			return true
-		}
-	}
-	return false
-}
-
 // IndexInSlice returns the index of given object o in array, and -1 if it is not in array.
 func IndexInSlice(array interface{}, o interface{}, equals Equals) int {
 	arr := reflect.ValueOf(array)
--- a/gossip/discovery/discovery_impl.go
+++ b/gossip/discovery/discovery_impl.go
@@ -10,6 +10,7 @@ import (
 	"bytes"
 	"fmt"
 	"math"
+	"slices"
 	"strconv"
 	"strings"
 	"sync"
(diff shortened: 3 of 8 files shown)


gossip/gossip/msgstore/msgs_test.go
```diff
@@ -79,12 +80,7 @@ func TestNewMessagesInvalidates(t *testing.T) {
 
 func TestMessagesGet(t *testing.T) {
 	contains := func(a []interface{}, e interface{}) bool {
-		for _, v := range a {
-			if v == e {
-				return true
-			}
-		}
-		return false
+		return slices.Contains(a, e)
```
pfi79: it seems to me that in your case the contain function can be removed. Just use slices.Contains below

gossip/util/misc.go
```diff
@@ -33,12 +34,7 @@ var viperLock sync.RWMutex
 
 // Contains returns whether a given slice a contains a string s
 func Contains(s string, a []string) bool {
-	for _, e := range a {
-		if e == s {
-			return true
-		}
-	}
-	return false
+	return slices.Contains(a, s)
```
pfi79: Look at the places where the Contains function is used. Should this function be removed and slices.Contains be used locally?

gossip/gossip/msgstore/msgs_test.go
pengqiseven: It seems that it cannot be modified here, otherwise there will be a compilation error:

```shell
in call to slices.Contains, S (type []interface{}) does not satisfy ~[]Ecompiler[CannotInferTypeArgs](https://pkg.go.dev/golang.org/x/tools/internal/typesinternal#CannotInferTypeArgs)
```

gossip/util/misc.go
pengqiseven: Will have the error:

```shell
in call to slices.Contains, S (type string) does not satisfy ~[]Ecompiler[CannotInferTypeArgs](https://pkg.go.dev/golang.org/x/tools/internal/typesinternal#CannotInferTypeArgs)
```

gossip/gossip/msgstore/msgs_test.go
pfi79: seems to need to be converted to a different data type

pengqiseven: Yes, We need to swap the order of the parameters. I will modify it and submit it shortly, please wait a few minutes.

gossip/util/misc.go
pengqiseven: The method has already been removed, but the above modifications would result in compilation errors due to issues with interface and generic constraints. I suggest not making any changes.

gossip/gossip/msgstore/msgs_test.go
pengqiseven: > seems to need to be converted to a different data type

The method has already been removed, but  modifications  in there  would result in compilation errors due to issues with interface and generic constraints. I suggest not making any changes.

gossip/util/misc.go
pfi79: Go ahead, convert everything to the same data type.

pfi79: please sort it out, don't stop in the middle.

gossip/gossip/msgstore/msgs_test.go
pfi79: like this, but I'm sure there's a better way to do it.
```
        var n2s interface{}
	for _, num2Search := range expected {
		n2s = num2Search
		require.True(t, slices.Contains(msgStore.Get(), n2s), "Value %v not found in array", num2Search)
	}
```

pfi79:
```
func TestMessagesGet(t *testing.T) {
	msgStore := NewMessageStore(alwaysNoAction, Noop)
	expected := make([]interface{}, 0)
	for i := 0; i < 2; i++ {
		n := r.Int()
		expected = append(expected, n)
		msgStore.Add(n)
	}

	for _, num2Search := range expected {
		require.True(t, slices.Contains(msgStore.Get(), num2Search), "Value %v not found in array", num2Search)
	}
}
```

pengqiseven: Sorry to deal with it so late.

I understand what you mean, it has been modified, please review again!

pfi79 (dismissed):

pfi79:
```
gossip/gossip/msgstore/msgs_test.go
The gofumpt command 'gofumpt -l -s -w' must be run for these files
make: *** [Makefile:194: linter] Error 1
```

pengqiseven:
> ```
> gossip/gossip/msgstore/msgs_test.go
> The gofumpt command 'gofumpt -l -s -w' must be run for these files
> make: *** [Makefile:194: linter] Error 1
> ```

Oh..

Thanks for your reply and help.

I have execute `gofumpt -l -s -w gossip/gossip/msgstore/msgs_test.go`. 

Please review.

pfi79 (approved):

pengqiseven:
Unfortunately, the CI has failed. However, I checked the reason, and it seems unrelated to this change.

```shell
2025-06-03 15:09:04.068 UTC 015b WARN [gateway] Endorse -> Endorse call timed out while collecting endorsements channel=test_channel chaincode=test_chaincode txID= numEndorsers=1
--- FAIL: TestEndorse (0.41s)
    --- FAIL: TestEndorse/endorse_retry_-_org3_fail_&_1_org2_peer_fail_-_requires_2_from_org1 (0.01s)
        endorse_test.go:582: 
            	Error Trace:	/home/runner/work/fabric/fabric/internal/pkg/gateway/endorse_test.go:582
            	Error:      	Received unexpected error:
            	            	rpc error: code = Aborted desc = failed to collect enough transaction endorsements, see attached details for more info
            	Test:       	TestEndorse/endorse_retry_-_org3_fail_&_1_org2_peer_fail_-_requires_2_from_org1
```

Issue #3763, opened 2022-11-01
Support post-quantum crypto algorithms


bestbeforetoday:
**As a** business application developer
**I want** to use post-quantum crypto algorithms with Fabric
**So that** the integrity of business data is maintained as quantum computing technology advances

NIST has selected some quantum-resistant cryptography algorithms:

https://www.nist.gov/news-events/news/2022/07/nist-announces-first-four-quantum-resistant-cryptographic-algorithms

It would be good for Fabric to support use of one or more of these algorithms (or future recommendations) as an alternative (or in addition) to the existing ECDSA cryptography for securing messages and ledger data. This could also extend to the cryptography used for secure (TLS) gRPC connections, although that might be better tracked as a separate issue.

Current post-quantum crypto algorithms are available as a Golang module here:

https://github.com/open-quantum-safe/liboqs-go

SamYuan1990:
https://github.com/hyperledger/fabric/discussions/3880, maybe for post-quantum crypto algorithms we can impl in a specific repo as decouple bccsp from fabric repo?

InLindquistWeTrust:
Currently implemented Falcon and Kyber keys into Hyperledger fabric for an IoT network. While the latency is low,I am trying to push the latency even lower for efficiency. Any suggestions on this would be greatly appreciated. ![image](https://github.com/user-attachments/assets/653d26f5-761a-4ef6-a749-00bdccdd7aa1)

bestbeforetoday:
Post-quantum crypto support in the Go standard libraries looks to be tracked in golang/go#64537.

The comments for that issue mention the [cloudflare/circl](https://github.com/cloudflare/circl) Go library containing post-quantum crypto implementations.

SamYuan1990:
> Post-quantum crypto support in the Go standard libraries looks to be tracked in [golang/go#64537](https://github.com/golang/go/issues/64537).
> 
> The comments for that issue mention the [cloudflare/circl](https://github.com/cloudflare/circl) Go library containing post-quantum crypto implementations.

IMO, there lots of PQC impls libraries today, but the blocking is on x509 as IETF is on draft version.
But, maybe we should get ready for Alg level and once IEFT finalized x509 format, we can go ahead.
Maybe for now, starts with either [cloudflare/circl](https://github.com/cloudflare/circl) or [liboqs](https://github.com/open-quantum-safe/liboqs-go) and then move to go native as a switch later on.(I am not sure if go native need to wait with openssl.... or not?)

[...]

PR #5232, opened 2025-06-08
try to install open-quantum-safe in base image file

SamYuan1990:
In recent I had been asked via mail for PQC in fabric, but it seems I lost connection.
So I try to the sample and discussion here, and make the discussion in show me you code way.
related with https://github.com/hyperledger/fabric-rfcs/pull/34

but further considering that IEFT still not publish https://datatracker.ietf.org/doc/draft-ietf-jose-pqc-kem/ and https://datatracker.ietf.org/doc/draft-ietf-lamps-cms-ml-dsa/ as they are still draft at  this PR creation Jun 8th 2025(beijing time),  I suppose it's still too early for fabric or hyperledger community to make some progress but only early research/investigation just enough. So I make this PR as draft.

this PR just part of invoke the crypto libraries at the container file/container level, and then people can ref https://github.com/open-quantum-safe/liboqs-go to impls bccsp, msp and modules in fabric to archive PQC in fabric.

### DIFF
--- a/images/baseos/Dockerfile
+++ b/images/baseos/Dockerfile
@@ -24,6 +24,14 @@ FROM ubuntu:${UBUNTU_VER} AS base
 RUN apt update && apt install -y \
     tzdata
 
+RUN apt-get -y update && \
+    apt-get install -y build-essential git cmake libssl-dev golang
+
+# Install liboqs
+RUN cmake -S liboqs -B liboqs/build -DBUILD_SHARED_LIBS=ON && \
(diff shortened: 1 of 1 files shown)


denyeart:
I think it would be better to open an Issue or Discussion rather than PR at this early stage.

bestbeforetoday:
Issue [#3763: Support post-quantum crypto algorithms](https://github.com/hyperledger/fabric/issues/3763) might be a good place for discussion.

PR #5236, opened 2025-06-16
remove sampleconfig/msp/keystore/key.pem from Dockerfiles

pfi79:
Previously we added a test private key `sampleconfig/msp/keystore/key.pem` to Docker images.
In this fix, I suggest not adding the `sampleconfig/msp` folder to Docker images at all.
Please take a close look at this fix.
Even though all the tests have passed, I think that I have missed something.


C0rWin:
Copying crypto materials inside a container wasn't a good idea to begin with. Thanks for updating it @pfi79

C0rWin (approved):

PR #5238, opened 2025-06-21
I suggest the "per channel create" command to make deprecated.

pfi79:
For this functionality, you need to use the Order Service Node (OSN)

Delete it after about a year or two.

### DIFF
--- a/docs/source/commands/peerchannel.md
+++ b/docs/source/commands/peerchannel.md
@@ -31,7 +31,7 @@ Usage:
   peer channel [command]
 
 Available Commands:
-  create               Create a channel
+  create               [DEPRECATED] Create a channel
   fetch                Fetch a block
   getinfo              get blockchain information of a specified channel.
   join                 Joins the peer to a channel.
@@ -59,7 +59,7 @@ Use "peer channel [command] --help" for more information about a command.
 
 ## peer channel create
 ```
-Create a channel and write the genesis block to a file.
+[DEPRECATED] Create a channel and write the genesis block to a file. Instead of this command, use Orderer Service Node (OSN).
 
 Usage:
   peer channel create [flags]
@@ -277,7 +277,7 @@ Global Flags:
 
 ## Example Usage
 
-### peer channel create examples
+### peer channel create examples (deprecated)
 
 Here's an example that uses the `--orderer` global flag on the `peer channel
 create` command.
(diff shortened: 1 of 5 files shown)


C0rWin:
@pfi79 have you checked this with fabric-samples?

pfi79:
> @pfi79 have you checked this with fabric-samples?

In fabric-samples, only `osnadmin channel join` is used. 
The `peer channel create` is not used there.

docs/source/commands/peerchannel.md
```diff
@@ -31,7 +31,7 @@ Usage:
   peer channel [command]
 
 Available Commands:
-  create               Create a channel
+  create               deprecated: Create a channel
```
satota2, line 34:
```suggestion
  create               [DEPRECATED] Create a channel
```
I think it would be better to use the same expression for deprecated as for other commands.
https://hyperledger-fabric.readthedocs.io/en/latest/commands/configtxgen.html

satota2 (commented):
@pfi79
Thank you for submitting this PR.

This is a minor suggestion, but I think it would be better to align the wording with other deprecated commands for consistency.

Also, it might be helpful to explicitly indicate that the command is deprecated in the following documentation section as well:

https://github.com/hyperledger/fabric/blob/main/docs/wrappers/peer_channel_postscript.md#peer-channel-create-examples

For example:

`peer channel create examples (deprecated)`

internal/peer/channel/create.go
```diff
@@ -38,11 +38,12 @@ func (e InvalidCreateTx) Error() string {
...
 }
 
+// deprecated
 func createCmd(cf *ChannelCmdFactory) *cobra.Command {
 	createCmd := &cobra.Command{
 		Use:   "create",
-		Short: "Create a channel",
-		Long:  "Create a channel and write the genesis block to a file.",
+		Short: "deprecated: Create a channel",
+		Long:  "deprecated: Use Orderer Service Node (OSN). Create a channel and write the genesis block to a file.",
```
satota2:
```suggestion
		Short: "[DEPRECATED] Create a channel",
		Long:  "[DEPRECATED] Use Orderer Service Node (OSN). Create a channel and write the genesis block to a file.",
```

I think it would be better to use the same expression for deprecated as for other commands.
https://hyperledger-fabric.readthedocs.io/en/latest/commands/configtxgen.html

pfi79: @satota2 Thank you for your comments. 
I fixed them.

If you need to make such changes somewhere else, please tell me where. I do not know all the places.

satota2 (approved):
@pfi79 Thank you for the updates. LGTM!

PR #5244, opened 2025-07-07
added to OSN new handler for update config of channel

pfi79:
The `peer channel update` internally only accesses orderers. 
I think it's better to add this command to OSN.

### DIFF
--- a/integration/nwo/configblock.go
+++ b/integration/nwo/configblock.go
@@ -7,19 +7,19 @@ SPDX-License-Identifier: Apache-2.0
 package nwo
 
 import (
+	"net/http"
 	"os"
 	"path/filepath"
-	"strings"
-
-	"github.com/hyperledger/fabric/common/channelconfig"
-	"github.com/hyperledger/fabric/common/policies"
 
 	"github.com/hyperledger/fabric-protos-go-apiv2/common"
 	"github.com/hyperledger/fabric-protos-go-apiv2/msp"
 	protosorderer "github.com/hyperledger/fabric-protos-go-apiv2/orderer"
+	"github.com/hyperledger/fabric/common/channelconfig"
+	"github.com/hyperledger/fabric/common/policies"
 	"github.com/hyperledger/fabric/integration/nwo/commands"
 	"github.com/hyperledger/fabric/internal/configtxlator/update"
 	"github.com/hyperledger/fabric/protoutil"
+	. "github.com/onsi/ginkgo/v2"
 	. "github.com/onsi/gomega"
 	"github.com/onsi/gomega/gbytes"
 	"github.com/onsi/gomega/gexec"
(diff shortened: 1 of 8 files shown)


pfi79:
pr is very heavy and difficult, I will break it down into parts to make it easier and clearer.

integration/raft/config_test.go
```diff
@@ -617,11 +613,13 @@ var _ = Describe("EndToEnd reconfiguration and onboarding", func() {
 			}
 
 			By(fmt.Sprintf("Rotating cert on leader %d", leader))
+			time.Sleep(5 * time.Second)
```
C0rWin: Please use Eventual with condition why are you waiting here instead of sleep.

integration/raft/config_test.go
```diff
@@ -617,11 +613,13 @@ var _ = Describe("EndToEnd reconfiguration and onboarding", func() {
...
 			rotate(leaderIndex)
 
 			By("Rotating certificates of other orderer nodes")
 			for i := range certificateRotations {
 				if i != leaderIndex {
+					time.Sleep(5 * time.Second)
```
C0rWin: Please use Eventual

integration/raft/config_test.go
```diff
@@ -707,6 +705,7 @@ var _ = Describe("EndToEnd reconfiguration and onboarding", func() {
 				port := network.OrdererPort(o, nwo.ClusterPort)
 
 				By(fmt.Sprintf("Adding the future certificate of orderer node %d", i))
+				time.Sleep(5 * time.Second)
```
C0rWin: Same as above

integration/raft/config_test.go
```diff
@@ -733,6 +732,7 @@ var _ = Describe("EndToEnd reconfiguration and onboarding", func() {
 				assertBlockReception(expectedBlockNumPerChannel[i*2], orderers, peer, network)
 
 				By("Removing the previous certificate of the old orderer")
+				time.Sleep(5 * time.Second)
```
C0rWin: Same as above, please use Eventual

integration/raft/config_test.go
pfi79: fixed

integration/raft/config_test.go
pfi79: fixed

integration/raft/config_test.go
pfi79: fixed

integration/raft/config_test.go
pfi79: fixed

integration/nwo/configblock.go
```diff
@@ -209,47 +221,57 @@ func UpdateOrdererConfig(n *Network, orderer *Orderer, channel string, current,
...
+	Expect(err).NotTo(HaveOccurred())
 
-		return strings.Contains(string(sess.Err.Contents()), "Successfully submitted channel update")
-	}, n.EventuallyTimeout).Should(BeTrue())
+	ready := make(chan struct{})
+	go func() {
+		defer GinkgoRecover()
+		Update(n, orderer, channel, updateEnvelope)
+		close(ready)
+	}()
```
C0rWin, line 236: why do you need this here?

pfi79: If the Update execution freezes "forever", Eventually will terminate the test earlier with an error.

C0rWin (approved):

PR #5268, opened 2025-08-13
Allow network to keep working after renewing certs with the same key

dviejokfs:
#### Type of change

Improvement

#### Description

This PR allows enrollment certificates of BFT orderers to be renewed without having to do a channel configuration update, as long as the renewed certificates have the same public key.

#### Related issues

No related issues

#### Release Note

Added support for seamless enrollment certificate renewal in orderer clusters when public keys remain unchanged. This prevents connection failures and NOT_FOUND errors during certificate rotation operations.

### DIFF
--- a/integration/smartbft/smartbft_test.go
+++ b/integration/smartbft/smartbft_test.go
@@ -13,7 +13,6 @@ import (
 	"context"
 	"crypto/ecdsa"
 	"crypto/rand"
-	"crypto/sha256"
 	"crypto/x509"
 	"encoding/pem"
 	"fmt"
@@ -2908,52 +2907,72 @@ func renewOrdererTLSCertificates(network *nwo.Network, orderers ...*nwo.Orderer)
 	}
 
 	for filePath, certPEM := range serverTLSCerts {
-		renewedCert, _ := expireCertificate(certPEM, ordererTLSCACert, ordererTLSCAKey, time.Now().Add(time.Hour))
+		renewedCert := renewCertificate(certPEM, ordererTLSCACert, ordererTLSCAKey, time.Now().Add(time.Hour))
 		err = os.WriteFile(filePath, renewedCert, 0o600)
 		log.Printf("Previous cert %s and next cert %s for orderer", string(certPEM), string(renewedCert))
 		Expect(err).NotTo(HaveOccurred())
 	}
 }
 
-func expireCertificate(certPEM, caCertPEM, caKeyPEM []byte, expirationTime time.Time) (expiredcertPEM []byte, earlyMadeCACertPEM []byte) {
+// renewCertificate generates a new certificate with the same public key and subject as the original,
+// but with a new NotAfter (expiration time). Only the expiration is changed.
+func renewCertificate(certPEM, caCertPEM, caKeyPEM []byte, notAfter time.Time) (renewedCertPEM []byte) {
+	// Parse CA private key
 	keyAsDER, _ := pem.Decode(caKeyPEM)
 	caKeyWithoutType, err := x509.ParsePKCS8PrivateKey(keyAsDER.Bytes)
 	Expect(err).NotTo(HaveOccurred())
 	caKey := caKeyWithoutType.(*ecdsa.PrivateKey)
 
+	// Parse CA certificate
 	caCertAsDER, _ := pem.Decode(caCertPEM)
 	caCert, err := x509.ParseCertificate(caCertAsDER.Bytes)
 	Expect(err).NotTo(HaveOccurred())
 
+	// Parse the original certificate
 	certAsDER, _ := pem.Decode(certPEM)
 	cert, err := x509.ParseCertificate(certAsDER.Bytes)
 	Expect(err).NotTo(HaveOccurred())
 
-	cert.Raw = nil
-	caCert.Raw = nil
-	// The certificate was made 1 hour ago
-	cert.NotBefore = time.Now().Add((-1) * time.Hour)
-	// As well as the CA certificate
-	caCert.NotBefore = time.Now().Add((-1) * time.Hour)
-	// The certificate expires now
-	cert.NotAfter = expirationTime
+	// Create a new certificate template with the same fields as the original,
+	// but with a new NotAfter (expiration time)
+	newCert := &x509.Certificate{
+		SerialNumber:          cert.SerialNumber,
+		Subject:               cert.Subject,
+		NotBefore:             cert.NotBefore,
+		NotAfter:              notAfter,
+		KeyUsage:              cert.KeyUsage,
+		ExtKeyUsage:           cert.ExtKeyUsage,
+		UnknownExtKeyUsage:    cert.UnknownExtKeyUsage,
+		BasicConstraintsValid: cert.BasicConstraintsValid,
+		IsCA:                  cert.IsCA,
+		DNSNames:              cert.DNSNames,
+		EmailAddresses:        cert.EmailAddresses,
+		IPAddresses:           cert.IPAddresses,
+		URIs:                  cert.URIs,
+		SubjectKeyId:          cert.SubjectKeyId,
+		AuthorityKeyId:        cert.AuthorityKeyId,
+		SignatureAlgorithm:    cert.SignatureAlgorithm,
+		PublicKeyAlgorithm:    cert.PublicKeyAlgorithm,
+		PublicKey:             cert.PublicKey,
+		PolicyIdentifiers:     cert.PolicyIdentifiers,
+		CRLDistributionPoints: cert.CRLDistributionPoints,
+		OCSPServer:            cert.OCSPServer,
+		IssuingCertificateURL: cert.IssuingCertificateURL,
+		ExtraExtensions:       cert.ExtraExtensions,
+		Extensions:            cert.Extensions,
+	}
 
-	// The CA signs the certificate
-	certBytes, err := x509.CreateCertificate(rand.Reader, cert, caCert, cert.PublicKey, caKey)
+	// The CA signs the new certificate
+	certBytes, err := x509.CreateCertificate(rand.Reader, newCert, caCert, cert.PublicKey, caKey)
 	Expect(err).NotTo(HaveOccurred())
 
-	// The CA signs its own certificate
-	caCertBytes, err := x509.CreateCertificate(rand.Reader, caCert, caCert, caCert.PublicKey, caKey)
-	Expect(err).NotTo(HaveOccurred())
-
-	expiredcertPEM = pem.EncodeToMemory(&pem.Block{Bytes: certBytes, Type: "CERTIFICATE"})
-	earlyMadeCACertPEM = pem.EncodeToMemory(&pem.Block{Bytes: caCertBytes, Type: "CERTIFICATE"})
+	renewedCertPEM = pem.EncodeToMemory(&pem.Block{Bytes: certBytes, Type: "CERTIFICATE"})
 	return
 }
 
 // renewOrdererEnrollmentCertificates renews the signcert for each orderer with a given expirationTime
 // and re-writes it to the orderer's signcerts directory, matching the actual crypto structure.
-func renewOrdererEnrollmentCertificates(network *nwo.Network, expirationTime time.Time, orderers ...*nwo.Orderer) {
+func renewOrdererEnrollmentCertificates(network *nwo.Network, notAfter time.Time, orderers ...*nwo.Orderer) {
 	if len(orderers) == 0 {
 		return
 	}
@@ -2985,69 +3004,63 @@ func renewOrdererEnrollmentCertificates(network *nwo.Network, expirationTime tim
 		ordererSignCert, err := os.ReadFile(ordererSignCertPath)
 		Expect(err).NotTo(HaveOccurred())
 
-		renewedCert, _ := expireOrdererSignCertificate(ordererSignCert, ordererCACert, ordererCAKey, expirationTime)
+		renewedCert := renewCertificate(ordererSignCert, ordererCACert, ordererCAKey, notAfter)
 		err = os.WriteFile(ordererSignCertPath, renewedCert, 0o600)
 		Expect(err).NotTo(HaveOccurred())
 	}
 }
 
-// expireCertificate re-creates and re-signs a certificate with a new expirationTime
-func expireOrdererSignCertificate(certPEM, caCertPEM, caKeyPEM []byte, expirationTime time.Time) (expiredcertPEM []byte, earlyMadeCACertPEM []byte) {
+// renewOrdererSignCertificate generates a new certificate with the same public key and subject as the original,
+// but with a new expirationTime. The only field that is changed is NotAfter (the expiration time).
+func renewOrdererSignCertificate(certPEM, caCertPEM, caKeyPEM []byte, notAfter time.Time) (renewedCertPEM []byte) {
+	// Parse CA private key
 	keyAsDER, _ := pem.Decode(caKeyPEM)
 	caKeyWithoutType, err := x509.ParsePKCS8PrivateKey(keyAsDER.Bytes)
 	Expect(err).NotTo(HaveOccurred())
 	caKey := caKeyWithoutType.(*ecdsa.PrivateKey)
 
+	// Parse CA certificate
 	caCertAsDER, _ := pem.Decode(caCertPEM)
 	caCert, err := x509.ParseCertificate(caCertAsDER.Bytes)
 	Expect(err).NotTo(HaveOccurred())
 
+	// Parse original certificate
 	certAsDER, _ := pem.Decode(certPEM)
 	cert, err := x509.ParseCertificate(certAsDER.Bytes)
 	Expect(err).NotTo(HaveOccurred())
 
-	cert.Raw = nil
-	caCert.Raw = nil
-	// The certificate was made 1 minute ago (1 hour doesn't work since cert will be before original CA cert NotBefore time)
-	cert.NotBefore = time.Now().Add((-1) * time.Minute)
-	// As well as the CA certificate
-	caCert.NotBefore = time.Now().Add((-1) * time.Minute)
-	// The certificate expires now
-	cert.NotAfter = expirationTime
+	// Create a new certificate template with the same fields as the original,
+	// but with a new NotAfter (expiration) time. NotBefore is kept the same as the original.
+	newCert := &x509.Certificate{
+		SerialNumber:          cert.SerialNumber,
+		Subject:               cert.Subject,
+		NotBefore:             cert.NotBefore,
+		NotAfter:              notAfter,
+		KeyUsage:              cert.KeyUsage,
+		ExtKeyUsage:           cert.ExtKeyUsage,
+		UnknownExtKeyUsage:    cert.UnknownExtKeyUsage,
+		BasicConstraintsValid: cert.BasicConstraintsValid,
+		IsCA:                  cert.IsCA,
+		DNSNames:              cert.DNSNames,
+		EmailAddresses:        cert.EmailAddresses,
+		IPAddresses:           cert.IPAddresses,
+		URIs:                  cert.URIs,
+		SubjectKeyId:          cert.SubjectKeyId,
+		AuthorityKeyId:        cert.AuthorityKeyId,
+		SignatureAlgorithm:    cert.SignatureAlgorithm,
+		PublicKeyAlgorithm:    cert.PublicKeyAlgorithm,
+		PublicKey:             cert.PublicKey,
+		CRLDistributionPoints: cert.CRLDistributionPoints,
+		PolicyIdentifiers:     cert.PolicyIdentifiers,
+		ExtraExtensions:       cert.ExtraExtensions,
+		Extensions:            cert.Extensions,
+	}
 
-	// The CA creates and signs a temporary certificate
-	tempCertBytes, err := x509.CreateCertificate(rand.Reader, cert, caCert, cert.PublicKey, caKey)
+	// Sign the new certificate with the CA
+	certBytes, err := x509.CreateCertificate(rand.Reader, newCert, caCert, cert.PublicKey, caKey)
 	Expect(err).NotTo(HaveOccurred())
 
-	// Force the certificate to use Low-S signature to be compatible with the identities that Fabric uses
-
-	// Parse the certificate to extract the TBS (to-be-signed) data
-	tempParsedCert, err := x509.ParseCertificate(tempCertBytes)
-	Expect(err).NotTo(HaveOccurred())
-
-	// Hash the TBS data
-	hash := sha256.Sum256(tempParsedCert.RawTBSCertificate)
-
-	// Sign the hash using forceLowS
-	r, s, err := forceLowS(caKey, hash[:])
-	Expect(err).NotTo(HaveOccurred())
-
-	// Encode the signature (DER format)
-	signature := append(r.Bytes(), s.Bytes()...)
-
-	// Replace the signature in the certificate with the low-s signature
-	tempParsedCert.Signature = signature
-
-	// Re-encode the certificate with the low-s signature
-	certBytes, err := x509.CreateCertificate(rand.Reader, tempParsedCert, caCert, cert.PublicKey, caKey)
-	Expect(err).NotTo(HaveOccurred())
-
-	// The CA signs its own certificate
-	caCertBytes, err := x509.CreateCertificate(rand.Reader, caCert, caCert, caCert.PublicKey, caKey)
-	Expect(err).NotTo(HaveOccurred())
-
-	expiredcertPEM = pem.EncodeToMemory(&pem.Block{Bytes: certBytes, Type: "CERTIFICATE"})
-	earlyMadeCACertPEM = pem.EncodeToMemory(&pem.Block{Bytes: caCertBytes, Type: "CERTIFICATE"})
+	renewedCertPEM = pem.EncodeToMemory(&pem.Block{Bytes: certBytes, Type: "CERTIFICATE"})
 	return
 }
 
--- a/integration/smartbft/smartbft_test.go
+++ b/integration/smartbft/smartbft_test.go
@@ -18,7 +18,6 @@ import (
 	"fmt"
 	"io"
 	"log"
-	"math/big"
 	"os"
 	"path/filepath"
 	"strconv"
@@ -265,6 +264,22 @@ var _ = Describe("EndToEnd Smart BFT configuration test", func() {
 			})
 
 			assertBlockReception(map[string]int{"testchannel1": 2}, network.Orderers, peer, network)
+
+			By("Try deploying chaincode")
+			peers := network.PeersWithChannel(channel)
+			Expect(len(peers)).ToNot(Equal(0))
+
+			// deploy the chaincode
+			deployChaincode(network, channel, testDir)
+
+			assertBlockReception(map[string]int{"testchannel1": 6}, network.Orderers, peer, network)
+
+			// test the chaincodes
+			invokeQuery(network, peer, orderer, channel, 90)
+			invokeQuery(network, peer, orderer, channel, 80)
+			invokeQuery(network, peer, orderer, channel, 70)
+			invokeQuery(network, peer, orderer, channel, 60)
+
 		})
 		It("smartbft node addition and removal", func() {
 			networkConfig := nwo.MultiNodeSmartBFT()
@@ -3008,74 +3023,4 @@ func renewOrdererEnrollmentCertificates(network *nwo.Network, notAfter time.Time
 		err = os.WriteFile(ordererSignCertPath, renewedCert, 0o600)
 		Expect(err).NotTo(HaveOccurred())
 	}
-}
-
-// renewOrdererSignCertificate generates a new certificate with the same public key and subject as the original,
-// but with a new expirationTime. The only field that is changed is NotAfter (the expiration time).
-func renewOrdererSignCertificate(certPEM, caCertPEM, caKeyPEM []byte, notAfter time.Time) (renewedCertPEM []byte) {
-	// Parse CA private key
-	keyAsDER, _ := pem.Decode(caKeyPEM)
-	caKeyWithoutType, err := x509.ParsePKCS8PrivateKey(keyAsDER.Bytes)
-	Expect(err).NotTo(HaveOccurred())
-	caKey := caKeyWithoutType.(*ecdsa.PrivateKey)
-
-	// Parse CA certificate
-	caCertAsDER, _ := pem.Decode(caCertPEM)
-	caCert, err := x509.ParseCertificate(caCertAsDER.Bytes)
-	Expect(err).NotTo(HaveOccurred())
-
-	// Parse original certificate
-	certAsDER, _ := pem.Decode(certPEM)
-	cert, err := x509.ParseCertificate(certAsDER.Bytes)
-	Expect(err).NotTo(HaveOccurred())
-
-	// Create a new certificate template with the same fields as the original,
-	// but with a new NotAfter (expiration) time. NotBefore is kept the same as the original.
-	newCert := &x509.Certificate{
-		SerialNumber:          cert.SerialNumber,
-		Subject:               cert.Subject,
-		NotBefore:             cert.NotBefore,
-		NotAfter:              notAfter,
-		KeyUsage:              cert.KeyUsage,
-		ExtKeyUsage:           cert.ExtKeyUsage,
-		UnknownExtKeyUsage:    cert.UnknownExtKeyUsage,
-		BasicConstraintsValid: cert.BasicConstraintsValid,
-		IsCA:                  cert.IsCA,
-		DNSNames:              cert.DNSNames,
-		EmailAddresses:        cert.EmailAddresses,
-		IPAddresses:           cert.IPAddresses,
-		URIs:                  cert.URIs,
-		SubjectKeyId:          cert.SubjectKeyId,
-		AuthorityKeyId:        cert.AuthorityKeyId,
-		SignatureAlgorithm:    cert.SignatureAlgorithm,
-		PublicKeyAlgorithm:    cert.PublicKeyAlgorithm,
-		PublicKey:             cert.PublicKey,
-		CRLDistributionPoints: cert.CRLDistributionPoints,
-		PolicyIdentifiers:     cert.PolicyIdentifiers,
-		ExtraExtensions:       cert.ExtraExtensions,
-		Extensions:            cert.Extensions,
-	}
-
-	// Sign the new certificate with the CA
-	certBytes, err := x509.CreateCertificate(rand.Reader, newCert, caCert, cert.PublicKey, caKey)
-	Expect(err).NotTo(HaveOccurred())
-
-	renewedCertPEM = pem.EncodeToMemory(&pem.Block{Bytes: certBytes, Type: "CERTIFICATE"})
-	return
-}
-
-// forceLowS ensures the ECDSA signature's S value is low
-func forceLowS(priv *ecdsa.PrivateKey, hash []byte) (r, s *big.Int, err error) {
-	r, s, err = ecdsa.Sign(rand.Reader, priv, hash)
-	Expect(err).NotTo(HaveOccurred())
-
-	curveOrder := priv.Curve.Params().N
-	halfOrder := new(big.Int).Rsh(curveOrder, 1) // curveOrder / 2
-
-	// If s is greater than half the order, replace it with curveOrder - s
-	if s.Cmp(halfOrder) > 0 {
-		s.Sub(curveOrder, s)
-	}
-
-	return r, s, nil
 }
--- a/integration/smartbft/smartbft_test.go
+++ b/integration/smartbft/smartbft_test.go
@@ -13,11 +13,13 @@ import (
 	"context"
 	"crypto/ecdsa"
 	"crypto/rand"
+	"crypto/sha256"
 	"crypto/x509"
 	"encoding/pem"
 	"fmt"
 	"io"
 	"log"
+	"math/big"
 	"os"
 	"path/filepath"
 	"strconv"
@@ -176,8 +178,6 @@ var _ = Describe("EndToEnd Smart BFT configuration test", func() {
 			network.GenerateConfigTree()
 			network.Bootstrap()
 
-			network.EventuallyTimeout *= 1
-
 			var ordererRunners []*ginkgomon.Runner
 			for _, orderer := range network.Orderers {
 				runner := network.OrdererRunner(orderer)
@@ -193,7 +193,6 @@ var _ = Describe("EndToEnd Smart BFT configuration test", func() {
 
 			Eventually(peerProcesses.Ready(), network.EventuallyTimeout).Should(BeClosed())
 			peer := network.Peer("Org1", "peer0")
-			_ = peer
 
 			sess, err := network.ConfigTxGen(commands.OutputBlock{
 				ChannelID:   "testchannel1",
@@ -242,8 +241,11 @@ var _ = Describe("EndToEnd Smart BFT configuration test", func() {
 				Eventually(ordererProcesses[i].Wait(), network.EventuallyTimeout).Should(Receive())
 			}
 
-			By("Renewing the certificates of the orderers")
-			renewOrdererCertificates(network, network.Orderers...)
+			By("Renewing the TLS certificates of the orderers")
+			renewOrdererTLSCertificates(network, network.Orderers...)
+
+			By("Renewing the enrollment certificates of the orderers")
+			renewOrdererEnrollmentCertificates(network, time.Now().Add(time.Hour), network.Orderers...)
 
 			By("Starting the orderers again")
 			for i := range network.Orderers {
@@ -2882,7 +2884,7 @@ func createPrePrepareRequest(
 	return req, block
 }
 
-func renewOrdererCertificates(network *nwo.Network, orderers ...*nwo.Orderer) {
+func renewOrdererTLSCertificates(network *nwo.Network, orderers ...*nwo.Orderer) {
 	if len(orderers) == 0 {
 		return
 	}
@@ -2947,4 +2949,120 @@ func expireCertificate(certPEM, caCertPEM, caKeyPEM []byte, expirationTime time.
 	expiredcertPEM = pem.EncodeToMemory(&pem.Block{Bytes: certBytes, Type: "CERTIFICATE"})
 	earlyMadeCACertPEM = pem.EncodeToMemory(&pem.Block{Bytes: caCertBytes, Type: "CERTIFICATE"})
 	return
+}
+
+// renewOrdererEnrollmentCertificates renews the signcert for each orderer with a given expirationTime
+// and re-writes it to the orderer's signcerts directory, matching the actual crypto structure.
+func renewOrdererEnrollmentCertificates(network *nwo.Network, expirationTime time.Time, orderers ...*nwo.Orderer) {
+	if len(orderers) == 0 {
+		return
+	}
+
+	for _, orderer := range orderers {
+		ordererDomain := network.Organization(orderer.Organization).Domain
+		// Use the orderer name as it appears in the file system, not the nwo.Orderer.ID()
+		// The directory is .../orderers/<ordererName>.<domain>/msp/signcerts/<ordererName>.<domain>-cert.pem
+		ordererName := orderer.Name
+		ordererFQDN := fmt.Sprintf("%s.%s", ordererName, ordererDomain)
+
+		// CA key and cert for the org
+		ordererCAKeyPath := filepath.Join(
+			network.RootDir, "crypto", "ordererOrganizations", ordererDomain, "ca", "priv_sk",
+		)
+		ordererCAKey, err := os.ReadFile(ordererCAKeyPath)
+		Expect(err).NotTo(HaveOccurred())
+
+		ordererCACertPath := filepath.Join(
+			network.RootDir, "crypto", "ordererOrganizations", ordererDomain, "ca", fmt.Sprintf("ca.%s-cert.pem", ordererDomain),
+		)
+		ordererCACert, err := os.ReadFile(ordererCACertPath)
+		Expect(err).NotTo(HaveOccurred())
+
+		// Path to the orderer's signcert
+		ordererSignCertPath := filepath.Join(
+			network.RootDir, "crypto", "ordererOrganizations", ordererDomain, "orderers", ordererFQDN, "msp", "signcerts", fmt.Sprintf("%s-cert.pem", ordererFQDN),
+		)
+		ordererSignCert, err := os.ReadFile(ordererSignCertPath)
+		Expect(err).NotTo(HaveOccurred())
+
+		renewedCert, _ := expireOrdererSignCertificate(ordererSignCert, ordererCACert, ordererCAKey, expirationTime)
+		err = os.WriteFile(ordererSignCertPath, renewedCert, 0o600)
+		Expect(err).NotTo(HaveOccurred())
+	}
+}
+
+// expireCertificate re-creates and re-signs a certificate with a new expirationTime
+func expireOrdererSignCertificate(certPEM, caCertPEM, caKeyPEM []byte, expirationTime time.Time) (expiredcertPEM []byte, earlyMadeCACertPEM []byte) {
+	keyAsDER, _ := pem.Decode(caKeyPEM)
+	caKeyWithoutType, err := x509.ParsePKCS8PrivateKey(keyAsDER.Bytes)
+	Expect(err).NotTo(HaveOccurred())
+	caKey := caKeyWithoutType.(*ecdsa.PrivateKey)
+
+	caCertAsDER, _ := pem.Decode(caCertPEM)
+	caCert, err := x509.ParseCertificate(caCertAsDER.Bytes)
+	Expect(err).NotTo(HaveOccurred())
+
+	certAsDER, _ := pem.Decode(certPEM)
+	cert, err := x509.ParseCertificate(certAsDER.Bytes)
+	Expect(err).NotTo(HaveOccurred())
+
+	cert.Raw = nil
+	caCert.Raw = nil
+	// The certificate was made 1 minute ago (1 hour doesn't work since cert will be before original CA cert NotBefore time)
+	cert.NotBefore = time.Now().Add((-1) * time.Minute)
+	// As well as the CA certificate
+	caCert.NotBefore = time.Now().Add((-1) * time.Minute)
+	// The certificate expires now
+	cert.NotAfter = expirationTime
+
+	// The CA creates and signs a temporary certificate
+	tempCertBytes, err := x509.CreateCertificate(rand.Reader, cert, caCert, cert.PublicKey, caKey)
+	Expect(err).NotTo(HaveOccurred())
+
+	// Force the certificate to use Low-S signature to be compatible with the identities that Fabric uses
+
+	// Parse the certificate to extract the TBS (to-be-signed) data
+	tempParsedCert, err := x509.ParseCertificate(tempCertBytes)
+	Expect(err).NotTo(HaveOccurred())
+
+	// Hash the TBS data
+	hash := sha256.Sum256(tempParsedCert.RawTBSCertificate)
+
+	// Sign the hash using forceLowS
+	r, s, err := forceLowS(caKey, hash[:])
+	Expect(err).NotTo(HaveOccurred())
+
+	// Encode the signature (DER format)
+	signature := append(r.Bytes(), s.Bytes()...)
+
+	// Replace the signature in the certificate with the low-s signature
+	tempParsedCert.Signature = signature
+
+	// Re-encode the certificate with the low-s signature
+	certBytes, err := x509.CreateCertificate(rand.Reader, tempParsedCert, caCert, cert.PublicKey, caKey)
+	Expect(err).NotTo(HaveOccurred())
+
+	// The CA signs its own certificate
+	caCertBytes, err := x509.CreateCertificate(rand.Reader, caCert, caCert, caCert.PublicKey, caKey)
+	Expect(err).NotTo(HaveOccurred())
+
+	expiredcertPEM = pem.EncodeToMemory(&pem.Block{Bytes: certBytes, Type: "CERTIFICATE"})
+	earlyMadeCACertPEM = pem.EncodeToMemory(&pem.Block{Bytes: caCertBytes, Type: "CERTIFICATE"})
+	return
+}
+
+// forceLowS ensures the ECDSA signature's S value is low
+func forceLowS(priv *ecdsa.PrivateKey, hash []byte) (r, s *big.Int, err error) {
+	r, s, err = ecdsa.Sign(rand.Reader, priv, hash)
+	Expect(err).NotTo(HaveOccurred())
+
+	curveOrder := priv.Curve.Params().N
+	halfOrder := new(big.Int).Rsh(curveOrder, 1) // curveOrder / 2
+
+	// If s is greater than half the order, replace it with curveOrder - s
+	if s.Cmp(halfOrder) > 0 {
+		s.Sub(curveOrder, s)
+	}
+
+	return r, s, nil
 }
--- a/integration/smartbft/smartbft_test.go
+++ b/integration/smartbft/smartbft_test.go
@@ -279,7 +279,6 @@ var _ = Describe("EndToEnd Smart BFT configuration test", func() {
 			invokeQuery(network, peer, orderer, channel, 80)
 			invokeQuery(network, peer, orderer, channel, 70)
 			invokeQuery(network, peer, orderer, channel, 60)
-
 		})
 		It("smartbft node addition and removal", func() {
 			networkConfig := nwo.MultiNodeSmartBFT()
--- a/integration/smartbft/smartbft_test.go
+++ b/integration/smartbft/smartbft_test.go
@@ -11,10 +11,13 @@ import (
 	"bytes"
 	"compress/gzip"
 	"context"
+	"crypto/ecdsa"
+	"crypto/rand"
 	"crypto/x509"
 	"encoding/pem"
 	"fmt"
 	"io"
+	"log"
 	"os"
 	"path/filepath"
 	"strconv"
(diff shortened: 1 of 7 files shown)


dviejokfs:
@yacovm @pfi79 @denyeart 

Integration tests are missing, and probably unit testing.

But I want to make sure this is the right approach.

I don't like having to modify logic in the common/cluster logic, such as:

https://github.com/hyperledger/fabric/pull/5268/files#diff-634c9f3bda7675ea45dd2a5bae942ef0a1ab059e41e0949cc47fef434b299e96R188-R195

But I don't see any other solution, since if I change membership to be the public key instead of a certificate, we also need to modify the logic of common/cluster.

orderer/consensus/smartbft/consenter.go
```diff
@@ -275,6 +312,25 @@ func (c *Consenter) IsChannelMember(joinBlock *cb.Block) (bool, error) {
 	return member, nil
 }
 
+// extractPublicKeyFromCert extracts the public key from an X.509 certificate
+func extractPublicKeyFromCert(certBytes []byte) ([]byte, error) {
+	// First decode PEM if it's PEM encoded
+	block, _ := pem.Decode(certBytes)
+	var der []byte
+	if block != nil {
```
yacovm: only one of these `if` branches should be the expected branch the execution follows, and the other `if` branch should be treated as an error, wouldn't you agree?

Besides, it's incorrect to assume that if the `block` is nil then it's because the function has been passed a DER and not a PEM. 

Let's find out which one of the two branches is the right one and treat the other as an error. 

That being said, if you follow the code of `IsChannelMember` I think you'll agree with me that we always pass a DER into this function.

orderer/consensus/smartbft/consenter.go
```diff
@@ -261,12 +262,48 @@ func (c *Consenter) IsChannelMember(joinBlock *cb.Block) (bool, error) {
...
+
+	// Extract public key using the same approach as IsConsenterOfChannel
+	bl, _ := pem.Decode(santizedCert)
+	if bl == nil {
+		return false, errors.Errorf("node identity certificate %s is not a valid PEM", string(santizedCert))
+	}
+
+	myPublicKey, err := extractPublicKeyFromCert(bl.Bytes)
+	if err != nil {
+		c.Logger.Debugf("Failed to extract public key from own certificate: %v", err)
```
yacovm: I think this log entry should be at least a warning, as this is an abnormal situation that points at a channel misconfiguration.

orderer/consensus/smartbft/consenter.go
```diff
@@ -261,12 +262,48 @@ func (c *Consenter) IsChannelMember(joinBlock *cb.Block) (bool, error) {
...
+		return false, err
+	} else {
+		c.Logger.Debugf("Extracted own public key: %x", myPublicKey)
+	}
+
 	for _, consenter := range oc.Consenters() {
 		santizedCert, err := crypto.SanitizeX509Cert(consenter.Identity)
 		if err != nil {
+			c.Logger.Debugf("Failed to sanitize consenter %d identity: %v", consenter.Id, err)
```
yacovm: This should be a warning as well

orderer/consensus/smartbft/consenter.go
```diff
@@ -261,12 +262,48 @@ func (c *Consenter) IsChannelMember(joinBlock *cb.Block) (bool, error) {
...
 			return false, err
 		}
-		if bytes.Equal(c.Identity, santizedCert) {
+
+		// Extract public key using the same approach as IsConsenterOfChannel
+		bl, _ := pem.Decode(santizedCert)
+		if bl == nil {
+			c.Logger.Debugf("Consenter %d: failed to decode PEM for identity", consenter.Id)
```
yacovm: should be a warning as well.

orderer/consensus/smartbft/consenter.go
```diff
@@ -261,12 +262,48 @@ func (c *Consenter) IsChannelMember(joinBlock *cb.Block) (bool, error) {
...
+			continue
+		}
+
+		publicKey, err := extractPublicKeyFromCert(bl.Bytes)
+		if err != nil {
+			c.Logger.Debugf("Consenter %d: failed to extract public key from cert: %v", consenter.Id, err)
```
yacovm: should be a warning as well

orderer/common/cluster/clusterservice.go
```diff
@@ -44,6 +46,25 @@ type ChannelMembersConfig struct {
 	nextStreamID      uint64
 }
 
+// extractPublicKeyFromCert extracts the public key from an X.509 certificate
+func extractPublicKeyFromCert(certBytes []byte) ([]byte, error) {
+	// First decode PEM if it's PEM encoded
+	block, _ := pem.Decode(certBytes)
```
yacovm: `certBytes` should be a DER, why do we assume it's a PEM?

orderer/consensus/smartbft/consenter.go
```diff
@@ -275,6 +312,25 @@ func (c *Consenter) IsChannelMember(joinBlock *cb.Block) (bool, error) {
 	return member, nil
 }
 
+// extractPublicKeyFromCert extracts the public key from an X.509 certificate
+func extractPublicKeyFromCert(certBytes []byte) ([]byte, error) {
```
yacovm: we have this function in `common/cluster` already, so we can just export it and import it. 

Probably worth moving it to `util.go`.

orderer/common/cluster/clusterservice.go
```diff
@@ -175,6 +201,31 @@ func (s *ClusterService) VerifyAuthRequest(stream orderer.ClusterNodeService_Ste
 	return authReq, nil
 }
 
+func compareCertPublicKeys(cert1, cert2 []byte) (bool, error) {
```
yacovm: this function should probably be moved to `util.go`.

integration/smartbft/smartbft_test.go
```diff
@@ -165,7 +168,103 @@ var _ = Describe("EndToEnd Smart BFT configuration test", func() {
...
 		})
+		It("disregards certificate renewal if only the validity period changed", func() {
+			networkConfig := nwo.MultiNodeSmartBFT()
+			networkConfig.Channels = nil
+
+			network = nwo.New(networkConfig, testDir, client, StartPort(), components)
+			network.GenerateConfigTree()
+			network.Bootstrap()
+
+			network.EventuallyTimeout *= 1
```
yacovm: what are we trying to accomplish here? `1` is the identity element for integer multiplication.

integration/smartbft/smartbft_test.go
```diff
@@ -165,7 +168,103 @@ var _ = Describe("EndToEnd Smart BFT configuration test", func() {
...
+
+			var ordererRunners []*ginkgomon.Runner
+			for _, orderer := range network.Orderers {
+				runner := network.OrdererRunner(orderer)
+				runner.Command.Env = append(runner.Command.Env, "FABRIC_LOGGING_SPEC=orderer.common.cluster=debug:orderer.consensus.smartbft=debug:policies.ImplicitOrderer=debug")
```
yacovm: this is for debugging, do we really need it?

```
runner.Command.Env = append(runner.Command.Env, "FABRIC_LOGGING_SPEC=orderer.common.cluster=debug:orderer.consensus.smartbft=debug:policies.ImplicitOrderer=debug")
```

integration/smartbft/smartbft_test.go
```diff
@@ -165,7 +168,103 @@ var _ = Describe("EndToEnd Smart BFT configuration test", func() {
...
+				ordererProcesses = append(ordererProcesses, proc)
+				Eventually(proc.Ready(), network.EventuallyTimeout).Should(BeClosed())
+			}
+
+			peerRunner := network.PeerGroupRunner()
+			peerProcesses = ifrit.Invoke(peerRunner)
+
+			Eventually(peerProcesses.Ready(), network.EventuallyTimeout).Should(BeClosed())
+			peer := network.Peer("Org1", "peer0")
+			_ = peer
```
yacovm: this `_ = peer` is redundant.

integration/smartbft/smartbft_test.go
```diff
@@ -2782,3 +2881,70 @@ func createPrePrepareRequest(
...
 	return req, block
 }
+
+func renewOrdererCertificates(network *nwo.Network, orderers ...*nwo.Orderer) {
+	if len(orderers) == 0 {
+		return
+	}
+	ordererDomain := network.Organization(orderers[0].Organization).Domain
+	ordererTLSCAKeyPath := filepath.Join(network.RootDir, "crypto", "ordererOrganizations",
+		ordererDomain, "tlsca", "priv_sk")
```
yacovm, line 2905: The certificate tree we are supposed to modify is not the TLS certificate tree but the enrollment certificate tree. 

So, this test changes the TLS certificates, but BFT works with enrollment certificates.

yacovm (changes requested):
Please see comments. 

Comments on production code are mostly nits, but the biggest problem is the comment about the integration test - it changes TLS certificates instead of enrollment certificates, so it doesn't prove the feature works, unfortunately.

orderer/common/cluster/clusterservice.go
```diff
@@ -163,7 +184,12 @@ func (s *ClusterService) VerifyAuthRequest(stream orderer.ClusterNodeService_Ste
 		return nil, errors.Errorf("node %d is not member of channel %s", authReq.ToId, authReq.Channel)
 	}
 
-	if !bytes.Equal(toIdentity, s.NodeIdentity) {
+	equal, err := compareCertPublicKeys(toIdentity, s.NodeIdentity)
+	if err != nil {
+		return nil, errors.Wrap(err, "failed to compare cert public keys")
+	}
```
pfi79: It seems there are not enough unit tests.

This is about the scenario.
1. The certificate is compared with the reference and everything is OK.
2. Something is changing.
3. And again it is compared with the standard.
bytes.Equal - should fail
compareCertPublicKeys - should be successful

integration/smartbft/smartbft_test.go
dviejokfs: I just updated the test to renew both, TLS and enrollment certificates.

orderer/common/cluster/clusterservice.go
dviejokfs: Thanks @pfi79 can you check the new added tests?

integration/smartbft/smartbft_test.go
```diff
@@ -2782,3 +2883,186 @@ func createPrePrepareRequest(
...
+	certAsDER, _ := pem.Decode(certPEM)
+	cert, err := x509.ParseCertificate(certAsDER.Bytes)
+	Expect(err).NotTo(HaveOccurred())
+
+	cert.Raw = nil
+	caCert.Raw = nil
+	// The certificate was made 1 minute ago (1 hour doesn't work since cert will be before original CA cert NotBefore time)
+	cert.NotBefore = time.Now().Add((-1) * time.Minute)
+	// As well as the CA certificate
+	caCert.NotBefore = time.Now().Add((-1) * time.Minute)
```
yacovm: why do you need to change the CA cert? you only need to expire the orderer cert.

yacovm:
Actually I am not sure about something - how does this integration test even work? The default policy that is encoded by configtxgen when the channel is BFT, is a policy that [requires explicit identity comparison](https://github.com/hyperledger/fabric/blob/73c227371a7ddff41a1a2fa421bd5935946edc50/common/policies/bft.go#L25-L41).

Even if we have communication between the nodes - how can the peers verify the signature? :confused: 
And how can the orderers [verify](https://github.com/hyperledger/fabric/blob/73c227371a7ddff41a1a2fa421bd5935946edc50/orderer/consensus/smartbft/verifier.go#L214) each other's signature?

dviejokfs:
@pfi79 could you do a revision after the recent changes?

yacovm:
> @pfi79 could you do a revision after the recent changes?

Please wait a couple of days, @dviejokfs , I will review again

yacovm (changes requested):
Parking this PR, I will revisit it in a couple of days.

orderer/common/cluster/clusterservice_test.go
```diff
@@ -785,3 +933,60 @@ func TestClusterRequestAsString(t *testing.T) {
 		})
 	}
 }
+
+// generateCertWithSameKey creates a new certificate using the provided signer (private key)
```
yacovm, line 937: this function isn't used, you can remove it.

integration/smartbft/smartbft_test.go
```diff
@@ -2782,3 +2895,129 @@ func createPrePrepareRequest(
...
+	Expect(err).NotTo(HaveOccurred())
+
+	renewedCertPEM = pem.EncodeToMemory(&pem.Block{Bytes: certBytes, Type: "CERTIFICATE"})
+	return
+}
+
+// renewOrdererEnrollmentCertificates renews the signcert for each orderer with a given expirationTime
+// and re-writes it to the orderer's signcerts directory, matching the actual crypto structure.
+func renewOrdererEnrollmentCertificates(network *nwo.Network, notAfter time.Time, orderers ...*nwo.Orderer) {
+	if len(orderers) == 0 {
```
yacovm, line 2988: you don't need this `if`, a for loop over an empty slice is a no-op

integration/smartbft/smartbft_test.go
```diff
@@ -2782,3 +2895,129 @@ func createPrePrepareRequest(
...
+		IssuingCertificateURL: cert.IssuingCertificateURL,
+		ExtraExtensions:       cert.ExtraExtensions,
+		Extensions:            cert.Extensions,
+	}
+
+	// The CA signs the new certificate
+	certBytes, err := x509.CreateCertificate(rand.Reader, newCert, caCert, cert.PublicKey, caKey)
+	Expect(err).NotTo(HaveOccurred())
+
+	renewedCertPEM = pem.EncodeToMemory(&pem.Block{Bytes: certBytes, Type: "CERTIFICATE"})
```
yacovm, line 2981: you can just `return  em.EncodeToMemory(&pem.Block{Bytes: certBytes, Type: "CERTIFICATE"})`

integration/smartbft/smartbft_test.go
```diff
@@ -2782,3 +2895,129 @@ func createPrePrepareRequest(
...
+	for filePath, certPEM := range serverTLSCerts {
+		renewedCert := renewCertificate(certPEM, ordererTLSCACert, ordererTLSCAKey, time.Now().Add(time.Hour))
+		err = os.WriteFile(filePath, renewedCert, 0o600)
+		Expect(err).NotTo(HaveOccurred())
+	}
+}
+
+// renewCertificate generates a new certificate with the same public key and subject as the original,
+// but with a new NotAfter (expiration time). Only the expiration is changed.
+func renewCertificate(certPEM, caCertPEM, caKeyPEM []byte, notAfter time.Time) (renewedCertPEM []byte) {
```
yacovm, line 2931: no need for the `renewedCertPEM` just keep the `[]byte`

yacovm (approved):
a few nits but not critical, you can make a follow-up PR to take care of them.

LGTM, nice work and thanks! 

I'll hold off from merging for a week to let @pfi79 or any other maintainer to take a second look.

pfi79 (approved):

tock-ibm:
> a few nits but not critical, you can make a follow-up PR to take care of them.
> 
> LGTM, nice work and thanks!
> 
> I'll hold off from merging for a week to let @pfi79 or any other maintainer to take a second look.

Thanks @yacovm , please let me have a couple of days to review this.

integration/smartbft/smartbft_test.go
```diff
@@ -165,7 +167,118 @@ var _ = Describe("EndToEnd Smart BFT configuration test", func() {
...
+			for i := range network.Orderers {
+				ordererProcesses[i].Signal(syscall.SIGTERM)
+				Eventually(ordererProcesses[i].Wait(), network.EventuallyTimeout).Should(Receive())
+			}
+
+			By("Renewing the TLS certificates of the orderers")
+			renewOrdererTLSCertificates(network, network.Orderers...)
+
+			By("Renewing the enrollment certificates of the orderers")
+			renewOrdererEnrollmentCertificates(network, time.Now().Add(time.Hour), network.Orderers...)
```
tock-ibm, line 245: I have to admit I am confused by this. The enrollment certificates are in the config block consenter map. They can only be changed by a config TX, correct? Are you suggesting that now orderers ignore the certificate and only use a public key from it?

tock-ibm: I thought the aim of the PR is to change the binding from TLS Cert hash to TLS public key, and yet the code shows that it is the logic of enrollment certs in the cluster service that changes. Can you explain how this feature is supposed to work in practice? Is the issue about renewal of TLS certs, enrollment certs, or both?

integration/smartbft/smartbft_test.go
```diff
@@ -165,7 +167,118 @@ var _ = Describe("EndToEnd Smart BFT configuration test", func() {
...
+			for i := range network.Orderers {
+				ordererRunner := network.OrdererRunner(network.Orderers[i])
+				ordererRunners[i] = ordererRunner
+				ordererProcesses[i] = ifrit.Invoke(ordererRunner)
+				Eventually(ordererProcesses[i].Ready(), network.EventuallyTimeout).Should(BeClosed())
+			}
+			updateBatchSize(network, peer, orderer, channel, func(batchSize *ordererProtos.BatchSize) {
+				batchSize.AbsoluteMaxBytes = 1000000
+				batchSize.MaxMessageCount = 300
+			})
```
tock-ibm, line 257: Why is `updateBatchSize` relevant to this case? it is a config change, but why is it relevant?

orderer/common/cluster/clusterservice.go
```diff
@@ -163,7 +163,12 @@ func (s *ClusterService) VerifyAuthRequest(stream orderer.ClusterNodeService_Ste
 		return nil, errors.Errorf("node %d is not member of channel %s", authReq.ToId, authReq.Channel)
 	}
 
-	if !bytes.Equal(toIdentity, s.NodeIdentity) {
+	equal, err := CompareCertPublicKeys(toIdentity, s.NodeIdentity)
+	if err != nil {
+		return nil, errors.Wrap(err, "failed to compare cert public keys")
+	}
+	if !equal {
+		s.Logger.Debugf("node id mismatch for node %d, toIdentity: %s, s.NodeIdentity: %s", authReq.FromId, string(toIdentity), string(s.NodeIdentity))
```
tock-ibm, line 171: This means that the node sending the request knows the node accepting the request by an enrollment cert that is stale; i.e. only the public key is the same. Why is this legitimate? I thought the problem was with TLS certs, not enrollment certs?

orderer/common/cluster/util.go
```diff
@@ -823,3 +827,38 @@ func EncodeTimestamp(t *timestamppb.Timestamp) []byte {
...
+	cert, err := x509.ParseCertificate(der)
+	if err != nil {
+		return nil, errors.Wrap(err, "failed to parse certificate")
+	}
+
+	return x509.MarshalPKIXPublicKey(cert.PublicKey)
+}
+
+func CompareCertPublicKeys(cert1, cert2 []byte) (bool, error) {
+	// Extract public key using the same approach as IsConsenterOfChannel
```
tock-ibm, line 842: Note that IsConsenterOfChannel with BFT compares enrollment certs, it is the raft version of IsConsenterOfChannel that compares public keys.

orderer/common/cluster/util_test.go
```diff
@@ -1112,3 +1116,280 @@ func TestChainParticipant(t *testing.T) {
...
+	require.NoError(t, err)
+
+	// Verify that public keys are the same even though certificate bytes differ
+	require.True(t, bytes.Equal(pubKey1, pubKey2),
+		"Public keys should be identical even though certificate bytes differ")
+
+	// This test demonstrates the scenario described in the PR comment:
+	// - bytes.Equal(firstCert, secondCert) returns false (different certificate bytes)
+	// - But the public keys are the same, so compareCertPublicKeys would return true
+	// - This ensures that certificate renewal/reissuance with same key pair works correctly
```
tock-ibm, line 1171: Renewal means extending the notAfter right? this ignores every aspect of the cert including key usage, name, etc.

orderer/consensus/smartbft/consenter.go
```diff
@@ -261,12 +261,46 @@ func (c *Consenter) IsChannelMember(joinBlock *cb.Block) (bool, error) {
...
+		}
+
+		publicKey, err := cluster.ExtractPublicKeyFromCert(bl.Bytes)
+		if err != nil {
+			c.Logger.Warnf("Consenter %d: failed to extract public key from cert: %v", consenter.Id, err)
+			continue
+		}
+		c.Logger.Debugf("Consenter %d: extracted public key: %x", consenter.Id, publicKey)
+		if bytes.Equal(myPublicKey, publicKey) {
+			c.Logger.Debugf("Found matching public key for consenter %d", consenter.Id)
```
tock-ibm, line 303: How is this change related to what is stated in the problem statement?
This means we ignore enrollment certs as identity and only care about the public key. In the description it says the problem is the renewal of TLS certs, which, in BFT, are not part of node identity. 

The code here is a significant change to how fabric orderer BFT handles consenters identity, isn't it?

orderer/consensus/smartbft/consenter.go
```diff
@@ -297,16 +331,29 @@ func pemToDER(pemBytes []byte, id uint64, certType string, logger *flogging.Fabr
...
+
+		certAsDER, err := pemToDER(cst.Identity, uint64(cst.Id), "server", c.Logger)
 		if err != nil {
+			c.Logger.Errorf("Failed to convert node identity certificate to DER: %s", err)
 			return 0, err
 		}
-		if bytes.Equal(c.Comm.NodeIdentity, santizedCert) {
-			return cst.Id, nil
+
+		if crypto.CertificatesWithSamePublicKey(thisNodeCertAsDER, certAsDER) == nil {
```
tock-ibm, line 351: How is this change related to what is stated in the problem statement?
This means we ignore enrollment certs as identity and only care about the public key. In the description it says the problem is the renewal of TLS certs, which, in BFT, are not part of node identity.

The code here is a significant change to how fabric orderer BFT handles consenters identity, isn't it?

orderer/consensus/smartbft/consenter.go
tock-ibm: The problem statement is:

https://github.com/hyperledger/fabric/pull/5268#:~:text=Currently%2C%20when%20TLS,maintaining%20backward%20compatibility.

IE: "when TLS certs are renewed". There is nothing there about enrollment certs.

tock-ibm (dismissed):

integration/smartbft/smartbft_test.go
yacovm: > Are you suggesting that now orderers ignore the certificate and only use a public key from it?

Yeah that's the point of the PR.

> I thought the aim of the PR is to change the binding from TLS Cert hash to TLS public key, and yet the code shows that it is the logic of enrollment certs in the cluster service that changes. Can you explain how this feature is supposed to work in practice? Is the issue about renewal of TLS certs, enrollment certs, or both?

Yeah that's a good point, the PR description indeed is not what the code is about.

@dviejokfs you should change the PR description...

integration/smartbft/smartbft_test.go
yacovm: I think he's just trying to make a test transaction

orderer/common/cluster/clusterservice.go
yacovm: Disregard the TLS certificate thing, I don't know why he wrote that.

> This means that the node sending the request knows the node accepting the request by an enrollment cert that is stale; i.e. only the public key is the same. Why is this legitimate?

When you authenticate to a remote node, you send also its [identity identifier](https://github.com/hyperledger/fabric/blob/main/orderer/common/cluster/clusterservice.go#L64). 

If the identity identifier is wrong then the node receiving the connection attempt will reject the connection attempt. 
Here, we just modify the code to only look at the public key.

I'm not sure why you think this is illegitimate. This has no cryptographic significance, it's just a way to detect channel misconfiguration of a node and also make the connection message non transferable.

orderer/common/cluster/util.go
yacovm: That's what he's doing here, he's changing the implementation to fit the Raft style where we only care about the public key, not the entire cert. So that if you need to rotate your enrollment certificate, you won't need to do a channel config.

orderer/common/cluster/util_test.go
yacovm: That's true, I think ideally he would just copy the certificate itself and just change the notAfter/notBefore, but I thought the gist of the test is clear so I let it go.

orderer/consensus/smartbft/consenter.go
yacovm: > The code here is a significant change to how fabric orderer BFT handles consenters identity, isn't it?

I guess significant is relative, are you trying to say such a change requires an RFC? Because I don't think it does, there is precedent for doing this without an RFC - like we did [here](https://github.com/hyperledger/fabric/pull/1771) in Raft.

tock-ibm: I actually think we need a deeper discussion on this. I started reading the description thinking the problem is the binding of TLS Cert hash.... I expected to see changes in the delivery.go and areas around it, but found a completely different feature altogether. I think such a change deserves (at least) an issue and some attention by crypto minded people. 
@ale-linux @adecaro @C0rWin 

This is a BFT system, not Raft,  so allowing orderers to operate without valid enrollment certs defined properly in the config block seems to me like a pretty big change. I am not comfortable with such a change entering without at least an issue discussing it properly. Yes, rotating certs is somewhat disruptive and annoying, but there is a reason behind it. 

Note the current implementation completely ignores everything in the cert but the key.... and it is allowed on all nodes at once... In addition this does not extend to the follower... or the BFT synchronizer if I am not mistaken... 

In addition, how does this solve the problem of TLS cert binding? this problem seems to be unchanged.

yacovm: > I actually think we need a deeper discussion on this. I started reading the description thinking the problem is the binding of TLS Cert hash.... 

This issue is not about TLS. In BFT we authenticate using enrollment certificates. The original code uses the enrollment certificate as-is and compares it in its raw byte form. The change proposed here is to just take into account the public key and not the entire certificate, in order that if you happen to rotate the enrollment certificate and keep the public key, then you wouldn't need a config update. This is how Raft works except with regarding to TLS certificates instead of enrollment certificates. 

> I expected to see changes in the delivery.go and areas around it, but found a completely different feature altogether. 

This is not for block pulling but for the intra-orderer cluster communication for BFT. It has nothing to do with the deliver API.


> This is a BFT system, not Raft, so allowing orderers to operate without valid enrollment certs defined properly in the config block seems to me like a pretty big change. 

The original code doesn't require the certificate to be valid at all, so there is no change in that aspect. 
Having a valid certificate chain is required when verifying a leaf certificate you are given and that **you have not seen before**.
But here, the certificates (or rather, public keys) are already known to you, since they're in the channel config that you hold. If I meet with you face to face and give you a certificate and say "hey Yoav this is my certificate" and later someone sends you a message and a signature that is verified by the former certificate, how do you know it's from me? You know it's from me because we have met before and I gave you my certificate. Here we have the exact same situation. You have a certificate in the channel config and someone sends you a message and a signature verifiable by that certificate from the channel config. Why do you need the certificate to be "valid"? The validity is implicit by it being in the channel config... 

> I am not comfortable with such a change entering without at least an issue discussing it properly. Yes, rotating certs is somewhat disruptive and annoying, but there is a reason behind it.

Yes I understand the reason behind it, that's why I am also fine with this change. It's disruptive and annoying and therefore for BFT orderers to be usable, @dviejokfs decided to make this change to improve the user experience. He actually (unlike both of us) deploys Fabric in production so this requirement of having to do a channel config for every orderer node was inconvenient enough to make a code contribution.


> Note the current implementation completely ignores everything in the cert but the key.... and it is allowed on all nodes at once... In addition this does not extend to the follower... or the BFT synchronizer if I am not mistaken...

You're not mistaken but this is completely irrelevant. Why would it extend to followers or to sync?


> In addition, how does this solve the problem of TLS cert binding? this problem seems to be unchanged.

I am not sure what you mean. You'll have to be more specific

tock-ibm: >The original code doesn't require the certificate to be valid at all, so there is no change in that aspect.
Having a valid certificate chain is required when verifying a leaf certificate you are given and that you have not seen before.

This is the problem I see here. The config update process should validate new certificates we are given as identities (or at least vigilant admins should not allow such invalid certificates and avoid endorsing them). However here there is no facility to verify that the certificate given in the context is even valid. So this essentially by-passes the endorsement process of orderer admins.

That is, certificates in the config block may be expired, but everything continues to work just because an orderer renewed it in its local file system and rebooted?

I think that this goes against security best practices in which an expired cert is a red flag. I think that if an organization sees a certificate of another organization becomes expired in the channel config it is a red flag.

It is like connecting to a web service and getting a warning "the certificate is expired, do you want to continue or go back to safety?" In the code proposed here there is no option to go back to safety...

yacovm: > > The original code doesn't require the certificate to be valid at all, so there is no change in that aspect.
> > Having a valid certificate chain is required when verifying a leaf certificate you are given and that you have not seen before.
> 
> This is the problem I see here. The config update process should validate new certificates we are given as identities (or at least vigilant admins should not allow such invalid certificates and avoid endorsing them). However here there is no facility to verify that the certificate given in the context is even valid. So this essentially by-passes the endorsement process of orderer admins.


Whether the config update process should or should not do sanity checks to the input is completely irrelevant to this PR. This PR does not change anything in this aspect. All that is done here is to ignore the other fields of the certificate and looking only at the public key, when mapping the authentication message sent over the wire to the consenter. 
This doesn't bypass the endorsement process of orderer admins because the only way to get anything into the channel config anyway is via a config update, which this PR does not change.

> 
> That is, certificates in the config block may be expired, but everything continues to work just because an orderer renewed it in its local file system and rebooted?


The enrollment certificate of the BFT orderer in the channel config is used for two things:

1. Establishing an authenticated communication channel
2. Signing blocks and protocol messages

Since (1) happens over a TLS connection, the TLS certificate and TLS certification chain also play a part in establishing an authentication communication channel. If the TLS certificate expires, then you won't be able to connect.

In the code **previous to this PR**, if the enrollment certificate of the orderer expires, then as long as it's in the channel config, you can still authenticate a remote BFT orderer as long as its TLS certificate is still valid. 
Therefore, if this is a problem, then it's a problem in the existing Fabric code, and this PR doesn't introduce a new problem. **Without this PR**, if we add expiry checks to enrollment certificates (which I reiterate, we currently do not have, in the current code) then we would have to do a config update for each renewal of the validity period of a certificate which is a painful thing. **With this PR**, it would be _less painful_ to the users if we introduce validity checks to the enrollment certificates which we currently do not have in the current code.

Therefore, without this PR, if you have changed the enrollment certificate but kept the previous public key, you do not gain any security but you do gain the pain to do the channel config update. With this PR you at least don't get the pain.

> 
> I think that this goes against security best practices in which an expired cert is a red flag. I think that if an organization sees a certificate of another organization becomes expired in the channel config it is a red flag.
> 
> It is like connecting to a web service and getting a warning "the certificate is expired, do you want to continue or go back to safety?" In the code proposed here there is no option to go back to safety...

Notice that the PR only deals with certificates that they have been renewed but the public key remains the same (and therefore so does the private key). 

The entire concept of validity checks for certificates is there to prevent leaked private keys from being used after the private keys have been regenerated. 
If the private key has been leaked and the certificate of a web server changed but its public key is the same, an attacker that has the leaked private key can still impersonate the web server because it's not a problem getting the renewed certificate (it's public and is presented to you whenever you connect to the real web server). 

I'm _not against_ adding support for enrollment certificate expiration checks in the authenticated communication establishment path, but if you have a leaked enrollment key then you essentially can sign blocks on behalf of that orderer, right? So... the only way to protect against a key leaking is changing the public keys in the channel config anyway!

tock-ibm: What's missing here is the fact that the set of enrollment certificates is visible also to the peers, and they use it to validate blocks. It is also visible to the clients. The users of the system are not just admins who may find it "painful" to do a config change.

Looking at a system which, in its source of truth - the config block in the blockchain - are expired certificates, a user would lose faith and trust in such a system. I see no way of getting around the necessity to keep the enrollment certs fresh and prevent them from expiring. How fabric does that right now is beside the issue. But going in the opposite direction, I think is against the interest of the systema as a whole.

A better approach to handle the administrator pain is to automate the cert renewal process. Another approach is to streamline the config change and make it easier.

yacovm: > What's missing here is the fact that the set of enrollment certificates is visible also to the peers, and they use it to validate blocks. It is also visible to the clients. The users of the system are not just admins who may find it "painful" to do a config change.

Well, when peers validate blocks they don't take into account that the enrollment certificates of ordereres and peers may have expired. There is actually a [bypass](https://github.com/hyperledger/fabric/blob/main/msp/mspimplvalidate.go#L293-L306) in the MSP that does that.


> Looking at a system which, in its source of truth - the config block in the blockchain - are expired certificates, a user would lose faith and trust in such a system. I see no way of getting around the necessity to keep the enrollment certs fresh and prevent them from expiring. How fabric does that right now is beside the issue. But going in the opposite direction, I think is against the interest of the systema as a whole.
> A better approach to handle the administrator pain is too automate the cert renewal process. Another approach is to streamline the config change and make it easier.

The Raft orderer [lets you keep TLS certificates of consenters in the config block expired](https://github.com/hyperledger/fabric/pull/1771), as long as the certificates they hold are still valid and have public keys matching to the config block.

I made that change years ago, based on requests of **users who actually deploy Fabric**.

Now there is a request of a user who deploys Fabric in the real world (who was skilled enough to make a PR and contribute code).

It is not up to you (or me) to tell users when they should or should not lose faith and trust in the system they themselves deploy. 

I have explained already that there is **zero security benefit** in keeping the same public key and renewing the certificate's validity period. Therefore, you should (logically) either be in favor of banning certificate renewal with the same public key entirely (which I think should not be up to us but to the users who deploy Fabric) or not block this PR from being merged, because you understand that if the same public key is retained in the enrollment certificate, the rest of the fields of the enrollment certificate are meaningless from a security perspective, so we can just ignore them when identifying which node send an authentication request.

dviejokfs:
Thank you, @tock-ibm and @yacovm, for your comments and discussion.

The issue right now is that Raft and BFT has different logic on renewing the certificates of the consenters.

On the other side, in BFT, TLS certs from the channel config don't seem to be used (correct me if I'm wrong), since I have a network with expired TLS certs, and BFT keeps working.

At the end, the problem is that if I have 10 networks in BFT, I want to be able to renew the certificates without having to do 10 * x orderers config updates.

I need this for my clients who are using BFT; one of them is a member of LFDT.

yacovm:
> 
> On the other side, in BFT, TLS certs from the channel config don't seem to be used (correct me if I'm wrong), since I have a network with expired TLS certs, and BFT keeps working.
> 


That's impossible. The TLS certificate of the node needs to be valid, otherwise no one would be able to connect to it. Not orderers and not peers, not clients, etc.

dviejokfs:
> > On the other side, in BFT, TLS certs from the channel config don't seem to be used (correct me if I'm wrong), since I have a network with expired TLS certs, and BFT keeps working.
> 
> That's impossible. The TLS certificate of the node needs to be valid, otherwise no one would be able to connect to it. Not orderers and not peers, not clients, etc.

I mean TLS certs in the channel config are expired, not the certs used by the orderer. These are renewed using the previous keypair.

tock-ibm:
David, @dviejokfs 

First of all, thank you for your effort and engagement. I appreciate your time and skill trying to make Fabric better!

Indeed Raft and BFT differ in how things work. In Raft, the TLS certificate pinning in the config block is the basis of authenticating orderer nodes with each other. As @yacovm mentioned, in Raft, there were indeed some shortcuts made when renewing said certificates when the public key is the same.

In BFT, although the *_tls_cert field still appear in the consenters map, it is unused (it is ignored and may be left empty, @yacovm correct me if I am wrong), and the `identity` field carries the enrollment certificate, which is also called the signing certificate.

```
// Consenter represents a consenting node (i.e. replica).
message Consenter {
    uint32 id = 1;
    string host = 2;
    uint32 port = 3;
    string msp_id = 4;
    bytes identity = 5;                <<< used in BFT
    bytes client_tls_cert = 6;       <<< used in Raft
    bytes server_tls_cert = 7;      <<< used in Raft
}
```

In BFT the enrollment cert is used both to sign blocks, and to establish authentication between orderer nodes.

Peers, as well as third party auditors, watch the stream of blocks coming out of the ordering service. These blocks are signed by orderers. My position is that they can expect that the super majority of orderers signing the blocks to have valid certificates (at least 2f+1 of them). 

What the proposed change will do, is allow the ordering service to continue and emit signed blocks even when the signing certificates of the entire cluster - as reported in the config block - had expired. The peers and the auditors don't know about the "self renewals with same key" going on under the cover. All they see is the last config block. This is a downgrade of the fabric security model, in my opinion. It cannot be done without warning to all other network owners out there, that have fabric 3.x BFT networks deployed and expect it to work in a certain way. 

In addition, I speculate that a large part of them will want to keep the default strict model.

In light of the lively discussion here, I propose to move this line of work to an RFC, and wait for the evaluation of additional maintainers and stake holders like @C0rWin @ale-linux @bestbeforetoday @andrew-coleman @denyeart @manish-sethi etc.

Again, @dviejokfs I am highly appreciative of your contribution, but I think this issue deserve further evaluation.

pfi79:
> Peers, as well as third party auditors, watch the stream of blocks coming out of the ordering service. These blocks are signed by orderers. My position is that they can expect that the super majority of orderers signing the blocks to have valid certificates (at least 2f+1 of them).

Let's continue with your version. 3-5 years have passed and the certificate has become invalid. And I connect a new peer to the channel, which starts downloading blocks starting from 0. The peer does not know anything about the fact that the certificates have already been updated. And the certificates that signed the blocks are already expired.

What to do?

tock-ibm:
> > Peers, as well as third party auditors, watch the stream of blocks coming out of the ordering service. These blocks are signed by orderers. My position is that they can expect that the super majority of orderers signing the blocks to have valid certificates (at least 2f+1 of them).
> 
> Let's continue with your version. 3-5 years have passed and the certificate has become invalid. And I connect a new peer to the channel, which starts downloading blocks starting from 0. The peer does not know anything about the fact that the certificates have already been updated. And the certificates that signed the blocks are already expired.
> 
> What to do?

In BFT, the peer actually does know the full sequence of enrollment cert renewals, because the config block is in the block chain. The block validation policy is rebuilt dynamically on every config block fetched, so the signatures on the block are evaluated according to the last config block that had preceded it. This way an auditor, even 5 years past the event, can evaluate the veracity of the blockchain.

This code is already here in the BFT block puller and BFT synchronizer.

yacovm:
> What the proposed change will do, is allow the ordering service to continue and emit signed blocks even when the signing certificates of the entire cluster had expired. This is a downgrade of the fabric security model

That's not true. Which part of the PR makes you think that is the case? I don't think you understand the code changes made in the PR.

tock-ibm:
> > What the proposed change will do, is allow the ordering service to continue and emit signed blocks even when the signing certificates of the entire cluster had expired. This is a downgrade of the fabric security model
> 
> That's not true. Which part of the PR makes you think that is the case? I don't think you understand the code changes made in the PR.

The integration test where all certs are self renewed with same key?
If you continue to renew locally as in the test, eventually the config block will carry expired certs, that is what I mean.

Just like what @dviejokfs reported is done with Raft - certs are renewed locally with same key, but not in the config block; right?

I clarify:
 "What the proposed change will do, is allow the ordering service to continue and emit signed blocks even when the signing certificates of the entire cluster - **as reported in the config block** - had expired. "

pfi79:
> > > Peers, as well as third party auditors, watch the stream of blocks coming out of the ordering service. These blocks are signed by orderers. My position is that they can expect that the super majority of orderers signing the blocks to have valid certificates (at least 2f+1 of them).
> > 
> > 
> > Let's continue with your version. 3-5 years have passed and the certificate has become invalid. And I connect a new peer to the channel, which starts downloading blocks starting from 0. The peer does not know anything about the fact that the certificates have already been updated. And the certificates that signed the blocks are already expired.
> > What to do?
> 
> In BFT, the peer actually does know the full sequence of enrollment cert renewals, because the config block is in the block chain. The block validation policy is rebuilt dynamically on every config block fetched, so the signatures on the block are evaluated according to the last config block that had preceded it. This way an auditor, even 5 years past the event, can evaluate the veracity of the blockchain.
> 
> This code is already here in the BFT block puller and BFT synchronizer.

Yes, I know. But what does the expiration of the certificate have to do with it? The peer will be required to accept the block, even if it is signed with an expired certificate.

tock-ibm:
> > > > Peers, as well as third party auditors, watch the stream of blocks coming out of the ordering service. These blocks are signed by orderers. My position is that they can expect that the super majority of orderers signing the blocks to have valid certificates (at least 2f+1 of them).
> > > 
> > > 
> > > Let's continue with your version. 3-5 years have passed and the certificate has become invalid. And I connect a new peer to the channel, which starts downloading blocks starting from 0. The peer does not know anything about the fact that the certificates have already been updated. And the certificates that signed the blocks are already expired.
> > > What to do?
> > 
> > 
> > In BFT, the peer actually does know the full sequence of enrollment cert renewals, because the config block is in the block chain. The block validation policy is rebuilt dynamically on every config block fetched, so the signatures on the block are evaluated according to the last config block that had preceded it. This way an auditor, even 5 years past the event, can evaluate the veracity of the blockchain.
> > This code is already here in the BFT block puller and BFT synchronizer.
> 
> Yes, I know. But what does the expiration of the certificate have to do with it? The peer will be required to accept the block, even if it is signed with an expired certificate.

You have a point in the sense that the fabric blockchain does not carry an explicit true sense of time. However, in real time, a peer can tell if the current ordering service network has expired certs defined in the last config block, correct?

The peer is not **required** to check the cert expiry, But it can if it wants to. A blockchain explorer can. In my opinion the job of the channel admins is to keep the certs fresh in the config block. 

If you want to add a feature in which this is not required, maybe make it optional?

pfi79:
> > > > > Peers, as well as third party auditors, watch the stream of blocks coming out of the ordering service. These blocks are signed by orderers. My position is that they can expect that the super majority of orderers signing the blocks to have valid certificates (at least 2f+1 of them).
> > > > 
> > > > 
> > > > Let's continue with your version. 3-5 years have passed and the certificate has become invalid. And I connect a new peer to the channel, which starts downloading blocks starting from 0. The peer does not know anything about the fact that the certificates have already been updated. And the certificates that signed the blocks are already expired.
> > > > What to do?
> > > 
> > > 
> > > In BFT, the peer actually does know the full sequence of enrollment cert renewals, because the config block is in the block chain. The block validation policy is rebuilt dynamically on every config block fetched, so the signatures on the block are evaluated according to the last config block that had preceded it. This way an auditor, even 5 years past the event, can evaluate the veracity of the blockchain.
> > > This code is already here in the BFT block puller and BFT synchronizer.
> > 
> > 
> > Yes, I know. But what does the expiration of the certificate have to do with it? The peer will be required to accept the block, even if it is signed with an expired certificate.
> 
> You have a point in the sense that the fabric blockchain does not carry an explicit true sense of time. However, in real time, a peer can tell if the current ordering service network has expired certs defined in the last config block, correct?
> 
> The peer is not **required** to check the cert expiry, But it can if it wants to. A blockchain explorer can. In my opinion the job of the channel admins is to keep the certs fresh in the config block.
> 
> If you want to add a feature in which this is not required, maybe make it optional?

I'm sorry, but I disagree with you. 
The last block in the channel could have been made 5-10 years ago. 
And the peer who has now received it cannot draw the conclusions you described.

tock-ibm:
> > > > > > Peers, as well as third party auditors, watch the stream of blocks coming out of the ordering service. These blocks are signed by orderers. My position is that they can expect that the super majority of orderers signing the blocks to have valid certificates (at least 2f+1 of them).
> > > > > 
> > > > > 
> > > > > Let's continue with your version. 3-5 years have passed and the certificate has become invalid. And I connect a new peer to the channel, which starts downloading blocks starting from 0. The peer does not know anything about the fact that the certificates have already been updated. And the certificates that signed the blocks are already expired.
> > > > > What to do?
> > > > 
> > > > 
> > > > In BFT, the peer actually does know the full sequence of enrollment cert renewals, because the config block is in the block chain. The block validation policy is rebuilt dynamically on every config block fetched, so the signatures on the block are evaluated according to the last config block that had preceded it. This way an auditor, even 5 years past the event, can evaluate the veracity of the blockchain.
> > > > This code is already here in the BFT block puller and BFT synchronizer.
> > > 
> > > 
> > > Yes, I know. But what does the expiration of the certificate have to do with it? The peer will be required to accept the block, even if it is signed with an expired certificate.
> > 
> > 
> > You have a point in the sense that the fabric blockchain does not carry an explicit true sense of time. However, in real time, a peer can tell if the current ordering service network has expired certs defined in the last config block, correct?
> > The peer is not **required** to check the cert expiry, But it can if it wants to. A blockchain explorer can. In my opinion the job of the channel admins is to keep the certs fresh in the config block.
> > If you want to add a feature in which this is not required, maybe make it optional?
> 
> I'm sorry, but I disagree with you. The last block in the channel could have been made 5-10 years ago. And the peer who has now received it cannot draw the conclusions you described.

tock-ibm:
@yacovm 

I went through the existing code again to study it further. You were saying:

 "I'm not against adding support for enrollment certificate expiration checks in the authenticated communication establishment path, but if you have a leaked enrollment key then you essentially can sign blocks on behalf of that orderer, right? So... the only way to protect against a key leaking is changing the public keys in the channel config anyway!"
 
 I actually think it is a **good idea**, the TLS handshake does it on the TLS certs, so why not here?
 
 Another point you made:
 
 "I have explained already that there is zero security benefit in keeping the same public key and renewing the certificate's validity period."

You actually convinced me that keeping the same public key and renewing the certificate's validity has negative security impact, as it prolongs the vulnerability period of the private key, and is usually frowned upon by many security experts. 
 
"Therefore, you should (logically) either be in favor of banning certificate renewal with the same public key entirely (which I think should not be up to us but to the users who deploy Fabric) or ...."

I am not in favor of banning it - there is a third option - document that decision by network admins in the blockchain, and have clients decide whether they want to do business with this chain. That is - do it in a config change. This way it is out in the open and not hidden in the local file system of each ordering node.

And last, I have a question for @yacovm @dviejokfs: 
Right now, a node will not start if it has an expired enrollment cert because the local code at that node checks for validity and expiry via the CA cert. However, a rogue node may easily disable that check. Is it possible that such a node will self sign a new enrollment cert with same pub key and send it in the authentication flow? what will break at remote nodes if only the pub key is checked? (In Raft we assume that does not happen, CFT...)

Again, I am highly appreciative of your work, @dviejokfs & @yacovm, and thanks for your patience.

dviejokfs:
Thanks for the comments @yacovm and @tock-ibm 

I see valid points for both.

The problem is not for a network that's self managed to modify the certificates. I have been doing that for a long time and it works.

However to update the consenter certificates we need the Orderer Admin Policy to be met.

The problem is in the operational aspect. Imagine we have 5 orderer organizations with 1 orderer each one. The amount of work to renew the orderers it's so much.

This is a short-term solution; a long-term solution would be that organizations could change their certificates once they have consenter nodes, in the same way that a peer org can change their anchor peers without needing the Application Policy Admin to be met.

yacovm:
> > > What the proposed change will do, is allow the ordering service to continue and emit signed blocks even when the signing certificates of the entire cluster had expired. This is a downgrade of the fabric security model
> > 
> > 
> > That's not true. Which part of the PR makes you think that is the case? I don't think you understand the code changes made in the PR.
> 
> The integration test where all certs are self renewed with same key? If you continue to renew locally as in the test, eventually the config block will carry expired certs, that is what I mean.
> 
> Just like what @dviejokfs reported is done with Raft - certs are renewed locally with same key, but not in the config block; right?
> 
> I clarify: "What the proposed change will do, is allow the ordering service to continue and emit signed blocks even when the signing certificates of the entire cluster - **as reported in the config block** - had expired. "

You can still operate with expired enrollment certificates in the config block (or outside of it, actually) even now, before this change.

What this change does, is simply makes it possible to continue operating if you have rotated your enrollment certificate (but didn't change the public key).

The integration test doesn't even make the certificates expired, it just changes their validity time to make the certificate bytes change, that's all.

yacovm:
> Thanks for the comments @yacovm and @tock-ibm
> 
> I see valid points for both.
> 
> The problem is not for a network that's self managed to modify the certificates. I have been doing that for a long time and it works.
> 
> However to update the consenter certificates we need the Orderer Admin Policy to be met.
> 
> The problem is in the operational aspect. Imagine we have 5 orderer organizations with 1 orderer each one. The amount of work to renew the orderers it's so much.
> 
> This is a short-term solution; a long-term solution would be that organizations could change their certificates once they have consenter nodes, in the same way that a peer org can change their anchor peers without needing the Application Policy Admin to be met.

With anchor peers, an organization can add any number of anchor peers it wants. We cannot allow an organization to freely add orderers to the channel config as it wants, as it can do that repeatedly until it controls a third or more orderer nodes and then BFT consistency can be compromised if the organization is malicious.

The root cause of this problem is that the SmartBFT treats every node as its own party and has no notion of nodes belonging to parties. So if you have two nodes from the same organization, SmartBFT will consider them as two different votes in the protocol, even though the nodes have the same administrator.

dviejokfs:
> > Thanks for the comments @yacovm and @tock-ibm
> > I see valid points for both.
> > The problem is not for a network that's self managed to modify the certificates. I have been doing that for a long time and it works.
> > However to update the consenter certificates we need the Orderer Admin Policy to be met.
> > The problem is in the operational aspect. Imagine we have 5 orderer organizations with 1 orderer each one. The amount of work to renew the orderers it's so much.
> > This is a short-term solution; a long-term solution would be that organizations could change their certificates once they have consenter nodes, in the same way that a peer org can change their anchor peers without needing the Application Policy Admin to be met.
> 
> With anchor peers, an organization can add any number of anchor peers it wants. We cannot allow an organization to freely add orderers to the channel config as it wants, as it can do that repeatedly until it controls a third or more orderer nodes and then BFT consistency can be compromised if the organization is malicious.
> 
> The root cause of this problem is that the SmartBFT treats every node as its own party and has no notion of nodes belonging to parties. So if you have two nodes from the same organization, SmartBFT will consider them as two different votes in the protocol, even though the nodes have the same administrator.

That's why I said not to add/remove consenter (in this case admin orderer policyt must be required)

But each orderer org could modify THEIR consenter certificates so that networks with a lot of orderer organizations that are truly decentralized don't have a maintenance overhead every year.

yacovm:
> @yacovm
> 
> I went through the existing code again to study it further. You were saying:
> 
> "I'm not against adding support for enrollment certificate expiration checks in the authenticated communication establishment path, but if you have a leaked enrollment key then you essentially can sign blocks on behalf of that orderer, right? So... the only way to protect against a key leaking is changing the public keys in the channel config anyway!"
> 
> I actually think it is a **good idea**, the TLS handshake does it on the TLS certs, so why not here?

I am not sure what you mean regarding the TLS handshake. 

> 
> Another point you made:
> 
> "I have explained already that there is zero security benefit in keeping the same public key and renewing the certificate's validity period."
> 
> You actually convinced me that keeping the same public key and renewing the certificate's validity has negative security impact, as it prolongs the vulnerability period of the private key, and is usually frowned upon by many security experts.

But in the end of the day it is up to the organization that runs the node to decide the key rotation schedule of its nodes. We don't actually use any properties of the x509 standard for orderer enrollment certificates, so they should really be treated as public keys.
We encode certificates (and not public keys) into the channel config because tools that generate x509 certificates are very widespread so it's convenient for users to use x590 certificates. 

Besides, enrollment certificates can be stored in HSM or in similar software solutions that prevent the private key from existing on disk, so an organization should be able to opt out of key rotation for its enrollment certificates if it wishes to.

> 
> "Therefore, you should (logically) either be in favor of banning certificate renewal with the same public key entirely (which I think should not be up to us but to the users who deploy Fabric) or ...."
> 
> I am not in favor of banning it - there is a third option - document that decision by network admins in the blockchain, and have clients decide whether they want to do business with this chain. That is - do it in a config change. This way it is out in the open and not hidden in the local file system of each ordering node.

This proposal has nothing to do with this PR. The current state of the code even without this PR makes it that enrollment certificate expiration in the config block is not taken into account. This PR doesn't change anything in that aspect.

If you want to make a PR that makes Fabric even more complex to use, be my guest.

> 
> And last, I have a question for @yacovm @dviejokfs: Right now, a node will not start if it has an expired enrollment cert because the local code at that node checks for validity and expiry via the CA cert. However, a rogue node may easily disable that check. Is it possible that such a node will self sign a new enrollment cert with same pub key and send it in the authentication flow? 
> 

Even outside of Fabric, in every system with x509 based authentication, if a node has the private key of a certificate, then if the certificate the node has is expired, there are two mutually exclusive scenarios:

1. The real node has not changed its public key. In that case, the real node has a valid certificate. The rogue node can download the certificate of the real node and then impersonate it. 
2. The real node has changed its public key. In that case, the rogue node cannot use its valid certificate because it cannot produce a signature that is verified under that certificate. 

Having the private key but not a valid corresponding certificate should be treated as having a private key and a valid corresponding certificate, because a certificate [is not a secret ](https://en.wikipedia.org/wiki/Kerckhoffs%27s_principle).

And as me and @pfi79 explained already, certificate validity does not play a part in on-chain validation. So if you have stolen a private key, you can forge signatures of the orderer on past blocks, and certificate validity doesn't help here. 

> what will break at remote nodes if only the pub key is checked? (In Raft we assume that does not happen, CFT...)

I think you're misunderstanding what CFT means. CFT doesn't mean an attacker doesn't try to impersonate a node. CFT just means the protocol assumes that a node that has the credentials (making it part of the system), doesn't deviate from the consensus protocol. But the authentication mechanisms still need to be in place, otherwise malicious actors outside of the system can impersonate nodes from within the system. 

> > > Thanks for the comments @yacovm and @tock-ibm
> > > I see valid points for both.
> > > The problem is not for a network that's self managed to modify the certificates. I have been doing that for a long time and it works.
> > > However to update the consenter certificates we need the Orderer Admin Policy to be met.
> > > The problem is in the operational aspect. Imagine we have 5 orderer organizations with 1 orderer each one. The amount of work to renew the orderers it's so much.
> > > This is a short-term solution; a long-term solution would be that organizations could change their certificates once they have consenter nodes, in the same way that a peer org can change their anchor peers without needing the Application Policy Admin to be met.
> > 
> > 
> > With anchor peers, an organization can add any number of anchor peers it wants. We cannot allow an organization to freely add orderers to the channel config as it wants, as it can do that repeatedly until it controls a third or more orderer nodes and then BFT consistency can be compromised if the organization is malicious.
> > The root cause of this problem is that the SmartBFT treats every node as its own party and has no notion of nodes belonging to parties. So if you have two nodes from the same organization, SmartBFT will consider them as two different votes in the protocol, even though the nodes have the same administrator.
> 
> That's why I said not to add/remove consenter (in this case admin orderer policyt must be required)
> 
> But each orderer org could modify THEIR consenter certificates so that networks with a lot of orderer organizations that are truly decentralized don't have a maintenance overhead every year.

I might be wrong, but I don't think the config transaction framework lets you have a sub-tree of an organization that allows you only modify the certificates but not add new entries. At this point, the channel config framework is regarded as legacy code left to us by an ancient civilization and no one has the appetite to change it.

tock-ibm:
>But in the end of the day it is up to the organization that runs the node to decide the key rotation schedule of its nodes. We don't actually use any properties of the x509 standard for orderer enrollment certificates, so they should really be treated as public keys.
We encode certificates (and not public keys) into the channel config because tools that generate x509 certificates are very widespread so it's convenient for users to use x590 certificates.

Are you sure we don't use ANY of the features? the SmartBFT verifier or block validation policy will not verify the signature via the certs (from the config block of course), which also includes verifying the certification chain all the way to the CA? 

My question above was what happens in the auth flow here if the connecting client passes a cert that has the same pub key but is not signed by the correct CA. I have not received an answer... 

@dviejokfs this should be easy to check, just run the same integration test, but when you renew the enrollment certs, do it with another CA that is not in the config block (Only for enrollment!). I expect things to fail.

yacovm:
> > But in the end of the day it is up to the organization that runs the node to decide the key rotation schedule of its nodes. We don't actually use any properties of the x509 standard for orderer enrollment certificates, so they should really be treated as public keys.
> > We encode certificates (and not public keys) into the channel config because tools that generate x509 certificates are very widespread so it's convenient for users to use x590 certificates.
> 
> Are you sure we don't use ANY of the features? the SmartBFT verifier or block validation policy will not verify the signature via the certs (from the config block of course), which also includes verifying the certification chain all the way to the CA?
> My question above was what happens in the auth flow here if the connecting client passes a cert that has the same pub key but is not signed by the correct CA. I have not received an answer...

Both in the authentication flow, and when validating block signatures, the certification chain from the certificate to the CA does not play any part. 

The reason is that the config block is the source of truth here. So, in other words, even in the existing code before this PR, if I am a malicious admin and I can modify the code of my orderer as I see fit, and I possess a private key of an enrollment certificate that its certificate is in the channel config, no one is going to check my certificate both for protocol messages and for block signatures.

The reason is that the channel config is used as the source of truth for which certificates are valid, not the CA. The model where you verify a certificate's certification chain when you check its signature was invented to be used for leaf certificates that you do not trust and need to verify them somehow. But if you have the certificate in the config block, then you trust it because it's in the config block. If someone tells you it has a signature that is valid under the certificate from the config block there is no point in asking "oh yeah? then show me your certificate!" because the certificate is right there... in the config block. 

In the current code, if you take an orderer with a private key and a corresponding certificate that is in the channel config, and modify its certificate to be signed by a foreign CA but keep the same public key, the orderer will not start because it wouldn't find its certificate in the consenter set. However if you modify its code to start and use this certificate, then any other orderer that runs correct code will fully cooperate with this orderer and all peers will be able to validate blocks that are signed with signatures that one of them are from that orderer.

tock-ibm:
> Member

A few follow-up questions:

- What stops a local node from restarting with an expired enrollment cert? 
- What happens when a node keeps running with an expired enrollment cert?
- Before the proposed code change, a "self-renewal" with either a recognized CA or a foreign CA will exclude the node from consensus, correct?

yacovm:
> > Member
> 
> A few follow-up questions:
> 
>     * What stops a local node from restarting with an expired enrollment cert?
> 

The local MSP will [fail](https://github.com/hyperledger/fabric/commit/6f14c4db429862155d9f1bfdf2899712d988cbb0) if you initialize it with an expired certificate.

>     * What happens when a node keeps running with an expired enrollment cert?

Nothing... 

> 
>     * Before the proposed code change, a "self-renewal" with either a recognized CA or a foreign CA will exclude the node from consensus, correct?

Only if you run a node "by the book". However as I said before, an adversary can always impersonate an orderer it has its private key.

C0rWin:
@tock-ibm, I understand your security concerns, but I believe this change is essential for operational sustainability without compromising actual security. Let me explain with concrete facts from the codebase.

The fundamental reality is that **Fabric already doesn't validate certificate expiration in any security-critical path**. The only place expiration is checked is during local MSP initialization, not during:
- Block validation - `VerifyBlockSignature()` only checks signatures against public keys
- Consensus messages - SmartBFT verifier extracts public keys, ignores expiration
- Peer synchronization - Historical blocks with expired certs are always accepted

This means a network running for 2+ years with 1-year certificates **already works fine** - except when restarting orderers. The security model you're concerned about protecting doesn't actually exist in the implementation.

Raft consensus has been doing exactly this since Fabric v1.4.1. Look at `orderer/consensus/etcdraft/util.go`:
```go
// CompareCertificate returns true if the public keys of the certificates are the same
```

After 5+ years in production, there have been zero security incidents from this approach. Why should BFT orderers face operational burdens that Raft orderers don't?

As @dviejokfs explained, managing certificate renewals across multiple organizations is a coordination nightmare. For 5 orderer organizations across multiple channels, we're talking about dozens of config updates annually, each requiring multi-party approval. This isn't just inconvenient - it's a availability risk when renewals are delayed.

The current requirement forces organizations to choose between:
1. Risky coordinated updates that could fail
2. Setting very long certificate lifetimes (reducing security)
3. Operating with "expired" certificates in config blocks (what many do today)

The entire PR boils down to changing:
```go
// From: bytes.Equal(cert1.Raw, cert2.Raw)
// To: cert1.PublicKey.Equal(cert2.PublicKey)
```

That's it. It's not removing security checks - it's removing a byte comparison that provided no security value because:
- If an attacker has the private key, they can already sign messages regardless of certificate validity
- If they don't have the private key, the certificate comparison is irrelevant

You mentioned this is a "downgrade of the fabric security model," but I respectfully disagree. **We can't downgrade something that doesn't exist**. Certificate expiration enforcement in the config block is security theater - it looks like security but provides none because the actual validation paths ignore it.

This PR removes operational friction that causes real production outages while maintaining all actual cryptographic guarantees. It's not about making Fabric less secure - it's about making it less fragile by removing requirements that never provided security benefits in the first place.

C0rWin (approved):

yacovm:
I actually didn't expect this PR to have such rare support - not only it has approvals of three maintainers, but the three maintainers that did approve are pretty much the only ones active in the last year. Despite @tock-ibm tagging other maintainers several times, no one else has commented against or in favor. 

@tock-ibm - I've answered every question you had and I hope I explained how and why the Fabric orderer authentication mechanism works as it does. 

If you have further (unanswered) doubts why this PR shouldn't be merged, please do bring them forward.

tock-ibm:
@yacovm @dviejokfs @pfi79 
First, thanks for your contribution and patience. 

@C0rWin thanks for your response.

I have discussed this with the original MSP developers, and indeed your observations are true. As one put it: "Fabric identities cannot be expired, they can only be revoked". So it is perfectly fine to work with an expired signing certs in the config file. Indeed, the only expiration check is when a node starts.

So I remove my objection to this PR, merge it if you want. 

However I ask you this - a simpler approach would have been to simply disable the expiration check of signing (enrollment) certs on startup... wouldn't it? 
(maybe gated by a local config variable "allow_expired_signing_identity"?)

yacovm:
> However I ask you this - a simpler approach would have been to simply disable the expiration check of signing (enrollment) certs on startup... wouldn't it? (maybe gated by a local config variable "allow_expired_signing_identity"?)

That's a simpler check conceptually but unfortunately it would also affect any kind of MSP - also clients or peers. 
And then, when a user will use a peer or a client with an expired certificate, it will start up but will be crippled. 

> As one put it: "Fabric identities cannot be expired, they can only be revoked". 

Conceptually yes but practically not so much.
In Fabric, the peer checks by default (can be disabled) that the client's certificate has not expired. Similarly, the orderer checks that the client's certificate hasn't expired. 

The rational for this, is that client certificates aren't encoded in the channel config so they should be validated similarly to the x509 way - with validity periods, and revoking them is not scalable as it involves traversing through a linearly growing revocation list. 
Revoking an orderer though is very simple - just remove it from the channel config. 

We could of course expand the MSP config but that would involve changing the protobufs and also expanding the orderer local config which historically has caused problems to users when we changed its schema.

So going forward with this change will have the least pain to users.

PR #5280, opened 2025-08-27
added to OSN new handler for fetch a block of channel

pfi79:
part 1

### DIFF
--- a/orderer/common/multichannel/registrar.go
+++ b/orderer/common/multichannel/registrar.go
@@ -11,6 +11,7 @@ package multichannel
 
 import (
 	"path/filepath"
+	"strconv"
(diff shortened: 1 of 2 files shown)


dviejokfs:
Could I ask why this is needed? I don't see any usage of FetchBlock apart from the test @pfi79

pfi79:
> Could I ask why this is needed? I don't see any usage of FetchBlock apart from the test @pfi79 

To attach a peer or orderer to a channel, a block is required.

this in particular, but in general: there is functionality in peer cli that has nothing to do with peer. I want to move it to OSN because it concerns orderers

C0rWin (approved):

PR #5290, opened 2025-09-05
added to OSN new handler for fetch a block of channel part 2

pfi79:


orderer/common/channelparticipation/restapi.go
```diff
@@ -75,7 +80,7 @@ func NewHTTPHandler(config localconfig.ChannelParticipation, registrar ChannelMa
 		router:    mux.NewRouter(),
 	}
 
-	// swagger:operation GET /v1/participation/channels/{channelID} channels listChannel
+	// swagger:operation GET /participation/v1/channels/{channelID} channels listChannel
```
C0rWin, line 83: why did you change it?

pfi79: According to the code, everywhere orderer and osnadmin we use ["/participation/v1/"](https://github.com/hyperledger/fabric/blob/main/orderer/common/channelparticipation/restapi.go#L29). And only in the swagger file the path is reversed "/v1/participation/". This option is not working. This means that no one has used the swagger file yet.
Of the two options, I chose to leave the one that is exactly being used, respectively, and fixed the other one.

pfi79: The description of `// swagger:operation` only helps the program generate the file swagger.json doesn't affect anything else. 
The actual path pattern is defined below, and usually uses the constant `URLBaseV1`

yeasy:
Why is the description empty?
It would be helpful to add some commit message at least.

orderer/common/channelparticipation/restapi.go
yeasy: Putting the version number ahead is preferable. When you update to a newer version, it is easy to distinguish from the very beginning.

orderer/common/channelparticipation/validator.go
```diff
@@ -104,3 +105,21 @@ func ValidateUpdateConfigEnvelope(env *cb.Envelope) (channelID string, err error
 
 	return configUpdate.ChannelId, nil
 }
+
+func ValidateFetchBlockID(blockID string) error {
```
yeasy, line 110: Better add method comments:
ValidateFetchBlockID will validate that the blockID can only be newest|oldest|config|(number).

orderer/common/channelparticipation/validator_test.go
```diff
@@ -158,6 +158,66 @@ func TestValidateUpdateEnvelope(t *testing.T) {
 	}
 }
 
+func TestValidateFetchBlockID(t *testing.T) {
```
yeasy, line 161: Do we consider the hex number such as 0xff.
Otherwise, need to specify only supporting decimal format.

orderer/common/channelparticipation/restapi.go
pfi79: Alas, this is inherited. I know it's better this way, but then we'll break backward compatibility. Because so far everything has worked when the version was in second place.

orderer/common/channelparticipation/validator_test.go
pfi79: I didn't quite understand your question. I repeated the tests, I don't remember whose, but I didn't invent my own.

orderer/common/channelparticipation/validator.go
pfi79: Thanks for the comment. Added it.

orderer/common/channelparticipation/restapi.go
pfi79: Now there is a lot of code that already uses the version in the 2nd place. 
For example: https://github.com/hyperledger/fabric-admin-sdk/blob/main/internal/osnadmin/join.go#L19

orderer/common/channelparticipation/restapi.go
```diff
@@ -275,6 +309,41 @@ func (h *HTTPHandler) redirectBaseV1(resp http.ResponseWriter, req *http.Request
...
+	channelID, err := h.extractChannelID(req, resp)
+	if err != nil {
+		return
+	}
+
+	blockID, err := h.extractBlockID(req, resp)
+	if err != nil {
+		h.sendResponseJsonError(resp, http.StatusBadRequest, err)
+		return
+	}
```
C0rWin, line 328: don't you need to sanitize and check inputs? like to see whenever channelD and blockID both are correct?

C0rWin (approved):

orderer/common/channelparticipation/restapi.go
pfi79: the validation takes place inside these functions.

yacovm:
@C0rWin shall we cut a release on the weekend?

pfi79:
> shall we cut a release on the weekend?

Please wait for 2 pr. I need to finish the theme. 
Add new functionality and integration tests.

yacovm:
> > shall we cut a release on the weekend?
> 
> Please wait for 2 pr. I need to finish the theme. Add new functionality and integration tests.

Sure

PR #5310, opened 2025-10-12
Fabric release v3.1.2

yacovm:
Update release notes for v3.1.2

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -46,7 +46,7 @@
 #   - verify - runs unit tests for only the changed package tree
 
 UBUNTU_VER ?= 24.04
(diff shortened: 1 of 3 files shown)


yacovm:
@C0rWin @pfi79 please let me know if i missed something

pfi79:
When and through what procedure will we add v3 to the LTS Release?

yacovm:
> When and through what procedure will we add v3 to the LTS Release?

When we decide it has reached enough maturity. We released it roughly a year ago. If you ask me, we should wait until mid 2026 or so.

pfi79 (approved):

PR #5328, opened 2025-11-01
fix broken link checker

pfi79:
- add 403 status
- we only check en/latest


pfi79:
I think muffet is not a working tool. When checking links, he constantly finds errors 429 or 403. And this is a false positive. The recommendations of the muffet author are not valid.

I suggest disabling it.

PR #5364, opened 2025-12-18
fix(gossip): stop membership tracker ticker on channel shutdown

remo-lab:
## Bug Description:
A time.Ticker is created when initializing the membership tracker, but only the ticker’s channel is retained.
During shutdown, other resources are cleaned up, however the ticker itself is never stopped, violating Go’s ticker lifecycle requirements.

## Need for Change:
-The gossip channel creates a time.Ticker for membership tracking but does not stop it when the channel is shut down.
This leads to a resource leak where ticker goroutines continue running after the channel lifecycle has ended.

-In long-running peers or environments where gossip channels are frequently started and stopped, this can accumulate and cause unnecessary memory and CPU usage.

## Summary of Fix:
-This PR ensures proper cleanup of the membership tracker ticker by:
-Storing a reference to the created time.Ticker in the membership tracker
-Explicitly stopping the ticker during gossip channel shutdown
-Adding a nil check to safely handle cleanup
The fix is minimal and does not alter any existing behavior beyond proper resource management.

## Impact:
-Prevents goroutine and memory leaks
-Improves stability of long-running peers
-Ensures correct lifecycle management of background timers
-Particularly important when multiple gossip channels are created over time

## Verification:
-Existing gossip channel tests pass
-No linter or build errors
-Ticker is now correctly stopped during shutdown
-The issue was reproduced using a small standalone program that repeatedly starts and stops gossip channels and compares -goroutine counts before and after shutdown. A terminal screenshot demonstrating this behavior is attached.
Screenshot 2025-12-18 104221
A dedicated unit test was not added to avoid flaky goroutine-count assertions in CI environments.

### DIFF
--- a/gossip/gossip/channel/channel.go
+++ b/gossip/gossip/channel/channel.go
@@ -295,6 +295,7 @@ func NewGossipChannel(pkiID common.PKIidType, org api.OrgIdentityType, mcs api.M
 		getPeersToTrack: gc.GetPeers,
 		report:          gc.reportMembershipChanges,
 		stopChan:        make(chan struct{}, 1),
+		ticker:          ticker,
 		tickerChannel:   ticker.C,
 		metrics:         metrics,
 		chainID:         channelID,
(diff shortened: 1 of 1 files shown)


remo-lab:
Hi @denyeart @C0rWin @yacovm  ,This PR addresses a small resource cleanup issue in the gossip channel by ensuring the membership tracker ticker is properly stopped during shutdown.
Whenever you have some time, I’d really appreciate a review. Thanks!

yacovm (approved):

PR #5372, opened 2026-01-15
Fix BTF to BFT in test_network.md

sngh938:
Corrected the acronym from BTF to BFT for clarity in the documentation regarding the ordering service on SELinux.

#### Type of change

- Documentation update


sngh938:
Note that the btf version of the yml doesn't exist in test-network/compose/podman folder in the samples repo, but I just left the reference in the doc as is since it was there before, just corrected the typo.

pfi79 (approved):

sngh938:
Thanks @pfi79 for merging.

PR #5373, opened 2026-01-17
Fix gossip shutdown hang caused by blocking stop signal

remo-lab:
**This PR fixes a shutdown hang in the gossip channel state management logic.**

#### Problem

`stateInfoCache.Stop()` was using a **blocking send** on an unbuffered channel:

```go
func (cache *stateInfoCache) Stop() {
    cache.stopChan <- struct{}{}
}
```

If `Stop()` was called while the background goroutine was executing `Purge()`, the goroutine wasn’t listening on `stopChan`, causing the send to block indefinitely.
This could hang `gossipChannel.Stop()` and prevent peers from shutting down cleanly, especially under load.

#### Fix

The stop signaling was updated to use a **non-blocking, idempotent close pattern**:

1. Added a `sync.Once` guard to `stateInfoCache`
2. Replaced the blocking send with a guarded `close(stopChan)`

```go
func (cache *stateInfoCache) Stop() {
    cache.stopOnce.Do(func() {
        close(cache.stopChan)
    })
}
```

#### Why this works

* `close()` never blocks, unlike a channel send
* `sync.Once` ensures the channel is closed exactly once
* Closing the channel reliably unblocks the background goroutine
* This matches the shutdown pattern already used elsewhere in Fabric (e.g. `messageStoreImpl.Stop()`)

#### Testing

All channel-level tests pass, including shutdown-related coverage:

```bash
go test ./gossip/gossip/channel/... -v
```
Screenshot 2026-01-17 215925
Screenshot 2026-01-17 215948

**This change is minimal, localized, and only affects shutdown behavior by removing the possibility of an indefinite block.**

---

### DIFF
--- a/gossip/gossip/channel/channel.go
+++ b/gossip/gossip/channel/channel.go
@@ -1068,7 +1068,12 @@ func (cache *stateInfoCache) delete(msg *protoext.SignedGossipMessage) {
 }
 
 func (cache *stateInfoCache) Stop() {
-	cache.stopChan <- struct{}{}
+	select {
+	case <-cache.stopChan:
+		return
+	default:
+		close(cache.stopChan)
+	}
 }
 
 // GenerateMAC returns a byte slice that is derived from the peer's PKI-ID


remo-lab:
Hi @C0rWin @denyeart  @yacovm ,
This fixes a shutdown hang in the gossip channel caused by a blocking stop signal and switches it to a non-blocking, idempotent close pattern.

Whenever you have time, I’d really appreciate a review. Thanks!

gossip/gossip/channel/channel.go
```diff
@@ -1068,7 +1069,9 @@ func (cache *stateInfoCache) delete(msg *protoext.SignedGossipMessage) {
 }
 
 func (cache *stateInfoCache) Stop() {
-	cache.stopChan <- struct{}{}
+	cache.stopOnce.Do(func() {
```
yacovm: There is an easier way of fixing it, without adding the `sync.Once`. 

Just do:

```
select {
    case <- cache.stopChan:
         return
   default:
          close(cache.stopChan)
}
```

remo-lab:
Hi @yacovm ,
I’ve updated the fix based on your feedback.
Switched to a simple select-based channel close instead of the blocking send, and kept the struct unchanged (no `sync.Once`).
I’ve tested this locally and all channel tests are passing.

yacovm (approved):

PR #5375, opened 2026-01-21
Fix missing error when certificate directory is empty

DorKatzelnick:
#### Type of change

Improvement to code

#### Description

The goal of this code change is to correctly return an error when the directory contains no certificates.

#### Additional details

The function `getPemMaterialFromDir` returns a nil error and an empty list of certificates when the directory contains no certificates. As a result, calling `errors.Wrapf` or `errors.WithMessagef` with a nil error produces a nil error, and no error is propagated. The proposed fix addresses this issue by explicitly handling this case and ensuring that an appropriate error is returned.

### DIFF
--- a/msp/configbuilder.go
+++ b/msp/configbuilder.go
@@ -177,8 +177,10 @@ func GetLocalMspConfig(dir string, bccspConfig *factory.FactoryOpts, ID string) 
 	}
 
 	signcert, err := getPemMaterialFromDir(signcertDir)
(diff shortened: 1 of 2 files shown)


msp/configbuilder.go
tock-ibm, line 1: Can you add a unit test there? ideally one that fails before the fix and passes after?

DorKatzelnick: added few tests

tock-ibm (approved):

PR #5378, opened 2026-01-26
fix(gossip): prevent panic on state provider shutdown

Ady0333:
#### Type of change
- Bug fix

#### Description

This PR fixes a race condition in the gossip state provider that can cause a peer to panic during shutdown.

Gossip message handlers run in separate goroutines and may still be processing messages when Stop() is called. During shutdown, Stop() closes the state request and response channels. If an in-flight handler tries to send on these channels after they are closed, the peer crashes with panic: send on closed channel.

The fix ensures that message handlers respect the shutdown signal and exit cleanly instead of panicking.

#### Additional details

Channel sends in directMessage() are now guarded using select statements that also listen on stopCh.

This makes shutdown safe while preserving existing behavior during normal operation.
_Example:_
`select {
case s.stateRequestCh <- msg:
case <-s.stopCh:
}`

Tested locally with `go test ./gossip/state/...`.
The affected package passes cleanly. Broader gossip/service
tests fail due to missing orderer endpoints in the local environment.
image

#### Release Note
Fixes a race condition in the gossip state provider that could cause a peer to
panic during shutdown when gossip messages are in flight.

### DIFF
--- a/gossip/state/state.go
+++ b/gossip/state/state.go
@@ -389,16 +389,17 @@ func (s *GossipStateProviderImpl) directMessage(msg protoext.ReceivedMessage) {
 
 	if incoming.GetStateRequest() != nil {
 		if len(s.stateRequestCh) < s.config.StateChannelSize {
-			// Forward state request to the channel, if there are too
-			// many message of state request ignore to avoid flooding.
-			s.stateRequestCh <- msg
+			select {
+			case s.stateRequestCh <- msg:
+			case <-s.stopCh:
+			}
 		}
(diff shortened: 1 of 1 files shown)


Ady0333:
Hi @C0rWin @denyeart @yacovm,

This PR fixes a shutdown race in the gossip state provider that can cause a `send on closed channel` panic during peer shutdown.
I’ve kept the change minimal and tested the affected package locally.
Would appreciate your review when you get a chance. Thanks!

yacovm (approved):

PR #5379, opened 2026-01-27
gossip: fix TOCTOU race in discovery alive handling

Ady0333:
#### Type of change
- Bug fix

#### Description
This PR fixes a race condition in gossip discovery that could cause a peer to panic.

In handleAliveMessage(), membership state was checked under multiple read locks. If a member was purged concurrently between these checks, the code could reach a panic path even though this is a valid concurrent case.

The fix makes the membership check atomic and handles the edge case by re-learning the member instead of panicking.

#### Additional details
Tested locally with go test ./gossip/discovery/....
image

#### Release Note
Fixes a TOCTOU race in gossip discovery that could cause a peer to panic when a
member is concurrently purged while processing an alive message.

### DIFF
--- a/gossip/discovery/discovery_impl.go
+++ b/gossip/discovery/discovery_impl.go
@@ -505,6 +505,8 @@ func (d *gossipDiscoveryImpl) handleAliveMessage(m *protoext.SignedGossipMessage
 
 	d.lock.RLock()
 	_, known := d.id2Member[string(pkiID)]
+	_, isAlive := d.aliveLastTS[string(pkiID)]
+	lastDeadTS, isDead := d.deadLastTS[string(pkiID)]
 	d.lock.RUnlock()
 
 	if !known {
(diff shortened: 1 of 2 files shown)


Ady0333:
Hi @C0rWin @denyeart @yacovm,
This PR fixes a TOCTOU race in gossip discovery that could cause a peer panic under concurrent member purge.

Would appreciate your review when you have time.
 Thanks!

C0rWin:
> Hi @C0rWin @denyeart @yacovm, This PR fixes a TOCTOU race in gossip discovery that could cause a peer panic under concurrent member purge.
> 
> Would appreciate your review when you have time. Thanks!

It would be beneficial if you could add a unit test that demonstrates how the bug manifests, so it's clear you actually fixed the issue, and no regression will be introduced in the future.

gossip/discovery/discovery_impl.go
```diff
@@ -505,20 +505,18 @@ func (d *gossipDiscoveryImpl) handleAliveMessage(m *protoext.SignedGossipMessage
...
 		return
 	}
 
-	d.lock.RLock()
-	_, isAlive := d.aliveLastTS[string(pkiID)]
-	lastDeadTS, isDead := d.deadLastTS[string(pkiID)]
-	d.lock.RUnlock()
-
 	if !isAlive && !isDead {
-		d.logger.Panicf("Member %s is known but not found neither in alive nor in dead lastTS maps, isAlive=%v, isDead=%v", m.GetAliveMsg().Membership.Endpoint, isAlive, isDead)
```
C0rWin, line 521: While I think we shouldn't have paniced here, I still think it would be great to see unit test added explaining the case.

Ady0333:
Thanks @C0rWin for the suggestion. I’ve added a unit test that covers the concurrent purge case and verifies that handleAliveMessage() no longer panics.

The test passes locally with go test ./gossip/discovery/.... and 
go test ./gossip/discovery -run TestHandleAliveMessage_RelearnsMemberAfterConcurrentPurge

image
image

C0rWin:
@Ady0333, first of all,l thanks for taking the time to suggest the fix. I have two concerns:

To start with, the test creates the inconsistent state directly rather than actually triggering a concurrent race. While this is acceptable for unit testing (race conditions are hard to reproduce reliably), it doesn't prove the race can occur in production. However, the code structure before the fix clearly shows the race window exists, so the fix is justified.

Having said that, I think that your fix correctly addresses a real race condition by consolidating map reads into a single atomic operation. Therefore, I am in favour of merging it.

C0rWin (approved):

PR #5380, opened 2026-01-29
fix(ledger): prevent panic on snapshot manager shutdown

Ady0333:
#### Type of change
- Bug fix

#### Description
This PR fixes a shutdown race in the ledger snapshot manager that could cause a peer to panic.

Snapshot generation runs asynchronously in goroutines. If a peer shuts down while a snapshot is still in progress, those goroutines could attempt to send events on the events channel after it has been closed in shutdown(), resulting in a send on closed channel panic.

The fix coordinates snapshot goroutines with shutdown so that no sends occur after shutdown begins.

#### Additional details
Snapshot goroutines are now tracked and are signaled to stop during shutdown.
shutdown() waits for in-flight snapshot work to finish before closing channels.

Tested locally with `go test ./core/ledger/kvledger/...`

#### Release Note
Fixes a shutdown race in the ledger snapshot manager that could cause a peer
panic when snapshots are generated during shutdown.

### DIFF
--- a/core/ledger/kvledger/snapshot_mgmt.go
+++ b/core/ledger/kvledger/snapshot_mgmt.go
@@ -43,6 +43,8 @@ type snapshotMgr struct {
 	requestResponses          chan *requestResponse
 	stopped                   bool
 	shutdownLock              sync.Mutex
+	stopCh                    chan struct{}
+	wg                        sync.WaitGroup
 }
 
 type event struct {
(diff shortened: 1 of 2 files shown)


Ady0333:
Hi @C0rWin @denyeart @yacovm,

This PR fixes a shutdown race in the snapshot manager where a snapshot goroutine could send on a closed channel and panic.
The change makes shutdown safe even if a snapshot is still in progress.

Thanks for taking a look.

C0rWin:
Do you mind please adding unit tests to the PR?

Ady0333:
Yes sir, I will add the unit test shortly. Please allow me some time.

core/ledger/kvledger/snapshot_mgmt.go
```diff
@@ -261,6 +269,8 @@ func (m *snapshotMgr) shutdown() {
 	}
 
 	m.stopped = true
+	close(m.stopCh)
+	m.wg.Wait()
```
cendhu, line 273: As mentioned in the comment, ignoring the on-going snapshot activity was intentional. 

```
// shutdown sends a snapshotMgrShutdown event and close all the channels, which is called
// when the ledger is closed. For simplicity, this function does not consider in-progress commit
// or snapshot generation. The caller should make sure there is no in-progress commit or
// snapshot generation. Otherwise, it may cause panic because the channels have been closed.
```

Assume that the `processSnapshotMgmtEvents()` is at `l.generateSnapshot()` and the peer receives a shutdown command. If we do `m.wg.Wait()`, the peer would not shutdown immediately but wait for the snapshot to be generated fully which would take a significant time and also become useless as we are shutting the peer down. To correctly resolve this, we should pass a context to `l.generateSnapshot()` and cancel it when shutdown is called. As this change was intrusive, a simple approach was chosen.

cendhu (changes requested):
Thank you for taking the time to hunt down this bug and provide a fix. It's much appreciated. However, the current design was chosen for simplicity; a proper solution would require more intrusive changes.

core/ledger/kvledger/snapshot_mgmt.go
Ady0333: @cendhu. Thanks for the clarification. You're right that waiting for snapshot completion would delay peer shutdown. Since the intended approach would require propagating a cancellable context into generateSnapshot(), which is a more intrusive change, I'll close this PR for now.

PR #5384, opened 2026-02-03
Fix panic on send to closed channel during snapshot shutdown

Ady0333:
#### Type of change
- Bug fix

#### Description
This PR fixes a shutdown race in the ledger snapshot manager that could cause a peer panic.

Snapshot generation runs asynchronously in background goroutines. If the peer shuts down while a snapshot is still in progress, snapshotMgr.shutdown() could close internal channels before those goroutines finish, leading to a send on closed channel panic.

The fix coordinates shutdown with in-flight snapshot goroutines so shutdown is safe even when snapshot generation is ongoing.

#### Additional details
The snapshot manager now signals shutdown to background goroutines and waits for them to exit before closing channels.

Tested locally with `go test ./core/ledger/kvledger/....`
Also ran `go test ./core/ledger/kvledger -run Snapshot -v`
image

#### Release Note
Fixes a shutdown race in the ledger snapshot manager that could cause a peer
panic when snapshots are generated during shutdown.

### DIFF
--- a/core/ledger/kvledger/snapshot_mgmt.go
+++ b/core/ledger/kvledger/snapshot_mgmt.go
@@ -41,6 +41,8 @@ type snapshotMgr struct {
 	events                    chan *event
 	commitProceed             chan struct{}
 	requestResponses          chan *requestResponse
+	stopCh                    chan struct{}
+	wg                        sync.WaitGroup
 	stopped                   bool
 	shutdownLock              sync.Mutex
 }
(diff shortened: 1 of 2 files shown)


Ady0333:
Hi @CorWin @denyeart @yacovm , 
This PR fixes the shutdown panic and snapshot tests pass locally.
Please let me know if anything else is needed. Thanks!

core/ledger/kvledger/snapshot_mgmt.go
```diff
@@ -261,6 +271,8 @@ func (m *snapshotMgr) shutdown() {
 	}
 
 	m.stopped = true
+	close(m.stopCh)
+	m.wg.Wait()
```
cendhu, line 275: Refer to this comment -- https://github.com/hyperledger/fabric/pull/5380#discussion_r2903156731

cendhu (changes requested):

core/ledger/kvledger/snapshot_mgmt.go
Ady0333: Yes @cendhu , I will close this pr for now.

PR #5397, opened 2026-02-12
Fix TOCTOU race in gossip discovery handleAliveMessage

Ady0333:
#### Type of change
- Bug fix

---

#### Description
This PR fixes a TOCTOU race in gossip discovery that could crash a peer during normal operation.

`handleAliveMessage()` previously read membership state using multiple separate `RLock` / `RUnlock` sections and then treated the combined result as consistent. If a concurrent `purge()` ran between these lock gaps (for example during identity changes or message expiration), a member could appear as “known” but neither alive nor dead, triggering a `Panicf`. In another path, the same race could lead to a nil pointer dereference in `learnExistingMembers()`.

The fix consolidates all membership state reads into a single atomic `RLock` section and adds a defensive nil guard when updating existing members. This ensures inconsistent intermediate states are never observable and avoids crashing the peer.

---

#### Additional details
The change is localized to gossip discovery logic and does not affect normal behavior.  
Tested locally with `go test ./gossip/...`.

---

#### Release Note
Fixes a gossip discovery race where concurrent member purging could cause
a peer panic during normal operation.


Ady0333:
Hi @yacovm, when you get a chance, could you please review this PR?  
Thanks!

C0rWin:
@Ady0333, can you please add a unit test that manifests the issue you are solving here? Exactly like you did with https://github.com/hyperledger/fabric/pull/5379?

Ady0333:
Hello @C0rWin , extremely sorry for the late response. There was some technical issue.

I have added the unit test that manifests the issue. Please check it once and let me know if something more has to be changed.

pfi79:
@Ady0333  Will you still be working on this pr?

Atishyy27 (commented):
The nil guard in `learnExistingMembers` looks like the right fix — the purge can interleave between `handleAliveMessage`'s read-locked check and `learnExistingMembers` taking the write lock, so a nil check at the point of use is the reliable place to handle it.

The branch currently has unresolved merge-conflict markers committed in `discovery_impl.go` (`<<<<<<< fix-gossip-discovery-toctou-race` … `>>>>>>> main`), and the new test in `discovery_test.go` looks like it was spliced into `TestHandleAliveMessage_RelearnsMemberAfterConcurrentPurge` during the merge (missing closing brace) — which is why Basic Checks are failing. Could you rebase and clean those up?

One more thought: moving the `aliveLastTS` read into the first RLock block narrows the race window but doesn't fully close it (the locks are still released before the write path), so since the nil guard already handles the interleaving, it might read more clearly to leave that read where it was and let the guard be the fix.

PR #5405, opened 2026-02-14
Fix orphaned pvtdata missing entries from stale iterator during elg c…

Ady0333:
Fix orphaned eligible missing private data entries during collection eligibility conversion

#### Type of change

- Bug fix

#### Description

This PR fixes a race in private data storage that could silently create orphaned eligible missing data entries when a peer becomes newly eligible for a collection.

`processCollElgEvents()` converts ineligible missing data entries to eligible ones in batches. When the number of entries is large, it temporarily releases `purgerLock` between batches. During this unlock window, a purge cycle may delete expired missing data entries.

However, the conversion logic continued using a LevelDB iterator created before the lock was released. Because the iterator operates on a snapshot, it could still yield entries that were already deleted by the purger. Converting these stale entries recreates eligible prioritized missing data entries without corresponding expiry metadata, meaning they are never cleaned up by future purge cycles.

The fix re-creates the iterator after re-acquiring the lock so that only entries that still exist in the database are converted.

#### Additional details

- The change is minimal and localized to the eligibility conversion path
- No behavior changes outside `processCollElgEvents()`
- All existing tests pass locally

Tested with:
`go test ./core/ledger/...`

#### Release Note
Fixes an issue where private data eligibility conversion could create
orphaned missing data entries that were never cleaned up by purge cycles.


Ady0333:
@yacovm  and @pfi79 . Plese review this pr once you guys have time.
Thanks!!

pfi79:
We need a test that would show the importance of your changes.

Ady0333:
Thanks for the feedback @pfi79  sir.

I agree. I’ll add a unit test in pvtdatastorage that reproduces the eligibility-conversion + purge overlap and asserts that no orphaned eligible missing data entries are created.

Ady0333:
Hello @pfi79 ,
I am really really  sorry for this much delay. There was some technical error with my Git Account due to which i was nt able to code.

For now as requested, I have added a test which would show importance of my changes.

pfi79:
@Ady0333 Will you still be working on this pr?

Ady0333:
@pfi79 Yes, i will be working on this pr till final. Not only this pr but all the PRs which i have opened, I will work on them till end. 
Apologies for any inconvience happened by me. I have solve the merged conflicts 👍

Atishyy27 (commented):
Thanks for adding regression coverage here — this purge/eligibility interleaving had no test before.

I traced the store.go change against current main and I think it may now be redundant: #5424 (merged in March) already releases `collItr` before `purgerLock` is dropped and re-creates it after re-acquiring the lock, so the re-created iterator's snapshot already reflects purges that ran during the sleep window. Re-opening from `startKey` instead of `nextKey` also doesn't seem to change the observed key set — every key before `nextKey` was deleted by the batch flushed just before the sleep, and `purgeExpiredData` deletes the `InelgMissingData` keys too (store.go L839-848), so both start points see the same remaining keys; `startKey` just re-walks tombstones.

Related: the added `collItr.Release()` after re-acquiring the lock looks like it re-releases an iterator that was already released before the unlock (the line just above the sleep), so it can probably be dropped.

Could you check whether `TestProcessCollElgEventsIteratorRecreatedAfterPurge` fails on current main? If it passes there, it might be worth rebasing this down to a test-only PR — the regression test is genuinely valuable as a guard for the #5424 behavior. One small thing on the test: the `time.Sleep(100ms)` synchronization against the 500ms batch interval could be flaky on slow CI runners; a deterministic sync point (e.g. hooking the batch flush) would make it more robust.

PR #5413, opened 2026-03-07
Implement SeekInfoNewestHeader in delivery requester

siewcapital:
## What this PR does
- implements `DeliveryRequester.SeekInfoNewestHeader()` by creating and signing a `DELIVER_SEEK_INFO` envelope
- uses `Start=Newest` and `Stop=Newest` so the request returns exactly the newest header
- sets `ContentType=HEADER_WITH_SIG` and keeps `Behavior=BLOCK_UNTIL_READY`
- adds unit test coverage for the new method and verifies the encoded `SeekInfo` fields

## Why
`SeekInfoNewestHeader()` was still returning "not implemented yet", which prevented callers from requesting a single latest header through `DeliveryRequester`.

## Testing
- `go test ./common/deliverclient/blocksprovider -run TestDeliveryRequester -count=1`
- `go test ./common/deliverclient/blocksprovider -count=1`

### DIFF
--- a/common/deliverclient/blocksprovider/delivery_requester.go
+++ b/common/deliverclient/blocksprovider/delivery_requester.go
@@ -95,8 +95,26 @@ func seekInfoFrom(height uint64, contentType orderer.SeekInfo_SeekContentType) *
 // SeekInfoNewestHeader produces a signed SeekInfo envelope requesting the newest header (block attestation) available
 // to the orderer. Only a single header is expected in response, not a stream.
 func (dr *DeliveryRequester) SeekInfoNewestHeader() (*common.Envelope, error) {
-	// TODO
-	return nil, errors.New("not implemented yet")
+	seekInfo := &orderer.SeekInfo{
+		Start: &orderer.SeekPosition{
+			Type: &orderer.SeekPosition_Newest{Newest: &orderer.SeekNewest{}},
+		},
+		Stop: &orderer.SeekPosition{
+			Type: &orderer.SeekPosition_Newest{Newest: &orderer.SeekNewest{}},
+		},
+		Behavior:    orderer.SeekInfo_BLOCK_UNTIL_READY,
+		ContentType: orderer.SeekInfo_HEADER_WITH_SIG,
+	}
+
+	return protoutil.CreateSignedEnvelopeWithTLSBinding(
+		common.HeaderType_DELIVER_SEEK_INFO,
+		dr.channelID,
+		dr.signer,
+		seekInfo,
+		int32(0),
+		uint64(0),
+		dr.tlsCertHash,
+	)
 }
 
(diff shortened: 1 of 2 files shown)


pfi79:
Looks fine but you need to append your signed-off-by

https://www.secondstate.io/articles/dco/

pfi79:
@siewcapital  Will you still be working on this pr?

siewcapital:
Updated the branch on top of current `main` and rewrote it as a signed-off commit, so the previous conflict and DCO blockers should be cleared. The checks are rerunning now.

pfi79:
Something's wrong with DCO. In my experience, it will remain that way. Maybe we should wait for 2 days and throw another commit here. As a last resort, close this pr and open a new one.

pfi79:
> Updated the branch on top of current `main` and rewrote it as a signed-off commit, so the previous conflict and DCO blockers should be cleared. The checks are rerunning now.

please try to do something again, maybe this time the DCO will pass.

Co-Messi:
Pushed a fresh signed-off follow-up commit to retrigger the DCO/branch checks without changing the implementation. The branch is otherwise the same code-wise; this is just to give the PR a clean re-evaluation path.

pfi79:
Thank you for your contribution.

But let's get this straight, I haven't found where this function would be called. In other words, no one has ever received the "not implemented yet" error from this place.
I want to understand why you implemented it? 
Where will it be used in the future? 
And most importantly: what will it be used for?

siewcapital:
I checked the tree again and you are right: there is not a current in-tree caller for `SeekInfoNewestHeader()` today.

I added it for symmetry with `SeekInfoBlocksFrom()` / `SeekInfoHeadersFrom()` and because the BFT path already works with `HEADER_WITH_SIG` attestations, but I cannot point to a concrete current code path that needs the newest-header variant right now.

If you want to keep `DeliveryRequester` strictly demand-driven, I am fine closing this PR and only bringing it back together with a real caller.

pfi79:
> I checked the tree again and you are right: there is not a current in-tree caller for `SeekInfoNewestHeader()` today.
> 
> I added it for symmetry with `SeekInfoBlocksFrom()` / `SeekInfoHeadersFrom()` and because the BFT path already works with `HEADER_WITH_SIG` attestations, but I cannot point to a concrete current code path that needs the newest-header variant right now.
> 
> If you want to keep `DeliveryRequester` strictly demand-driven, I am fine closing this PR and only bringing it back together with a real caller.

Yes, I would like it to be closer to reality.

Co-Messi:
Closing this one — you're right that there's no in-tree caller for `SeekInfoNewestHeader()` and it doesn't make sense to merge dead code just for symmetry. I'll bring it back alongside a real caller when there's a concrete use case. Thanks for the review.

PR #5418, opened 2026-03-14
Optimize peer startup by avoiding unnecessary chaincode package reads

C0rWin:
During startup `InitializeLocalChaincodes()` reads and fully parses every installed chaincode tar.gz — two gzip decompressions per package — only to extract `PackageID`, `Label`, `Type`, and `Path`. Of these, `PackageID` and `Label` are already available from the filename (`<label>.<sha256hash>.tar.gz`), and `Type`/`Path` have no consumers anywhere in the codebase. With 500+ installed packages this takes around 5 minutes.

This change skips `Store.Load()` and `PackageParser.Parse()` during startup and uses the metadata that `ListInstalledChaincodes()` already extracts from filenames. The full package is still read lazily when actually needed — during chaincode build, runtime install, or DB artifact extraction. No changes to any other code paths.

### DIFF
--- a/core/chaincode/lifecycle/cache.go
+++ b/core/chaincode/lifecycle/cache.go
@@ -148,15 +148,17 @@ func (c *Cache) InitializeLocalChaincodes() error {
 	}
 
 	for _, ccPackage := range ccPackages {
-		ccPackageBytes, err := c.Resources.ChaincodeStore.Load(ccPackage.PackageID)
-		if err != nil {
-			return errors.WithMessagef(err, "could not load chaincode with package ID '%s'", ccPackage.PackageID)
+		// Use metadata from the package filename rather than reading and parsing
+		// the full tar.gz archive. The filename encodes the label and content hash
+		// (e.g., "mycc.abc123...tar.gz" → label="mycc", hash=abc123...).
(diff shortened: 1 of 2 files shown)


pfi79 (approved):

TaoGunner:
@C0rWin @pfi79 

Thanks for this optimization. Any chance this could be backported to release-2.5? We're still on the 2.5 LTS line (v2.5.15/2.5.16) and have 500+ installed chaincodes, startup at ~5 min is a real pain for us. The change looks safe and self-contained (no API/behavior change, lazy load only). If there's a policy reason 2.5 doesn't get such changes, I'd be glad to hear it.

pfi79:

pfi79:
> @C0rWin @pfi79
> 
> Thanks for this optimization. Any chance this could be backported to release-2.5? We're still on the 2.5 LTS line (v2.5.15/2.5.16) and have 500+ installed chaincodes, startup at ~5 min is a real pain for us. The change looks safe and self-contained (no API/behavior change, lazy load only). If there's a policy reason 2.5 doesn't get such changes, I'd be glad to hear it.

can you prepare a pr?

PR #5430, opened 2026-03-20
fix(pvtdatastorage): release iterators in deleteDataMarkedForPurge

sridhar-panigrahi:
## Summary

Spotted that `deleteDataMarkedForPurge()` never releases the LevelDB iterators it opens — both `purgeMarkerIter` and the inner `hashedIndexIter` per marker. Every other iterator in the same file uses `defer itr.Release()`, this one was just missed.

Since each open iterator holds a LevelDB snapshot that blocks compaction, and this function runs every purge cycle, the snapshots quietly pile up over time — growing memory and disk until the peer starts to degrade.

## Fix

- `defer Release()` for the outer `purgeMarkerIter`
- Explicit `Release()` for `hashedIndexIter` after each inner loop and on error paths

## Test plan

- [x] All existing `pvtdatastorage` unit tests pass
- [x] `go build ./core/ledger/...` clean

### DIFF
--- a/core/ledger/pvtdatastorage/store.go
+++ b/core/ledger/pvtdatastorage/store.go
@@ -885,6 +885,7 @@ func (s *Store) deleteDataMarkedForPurge() error {
 	if err != nil {
 		return err
 	}
+	defer purgeMarkerIter.Release()
 
 	for purgeMarkerIter.Next() {
 		if err := purgeMarkerIter.Error(); err != nil {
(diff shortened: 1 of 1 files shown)


sridhar-panigrahi:

sridhar-panigrahi:
No rush at all @pfi79! Since the fix is pretty small (just a couple of Release calls to match the rest of the file), I figured it'd be worth putting up so it's ready whenever you have time to look at it. Happy to make any changes if needed.

pfi79 (approved):

PR #5432, opened 2026-03-23
fix(orderer): cancel all streams in RemoteContext.Abort()

sridhar-panigrahi:
`RemoteContext.Abort()` calls `sync.Map.Range()` but the callback returns `false`, so it stops after canceling only the first stream. Any other active streams (e.g. a consensus stream + a submit stream to the same peer) keep running with their goroutines and gRPC resources still held.

Changed `return false` to `return true` so all streams get canceled.

Added `TestAbortCancelsAllStreams` — opens three streams, calls Abort(), verifies all three are canceled. Fails on old code, passes with the fix. Full cluster test suite passes.


sridhar-panigrahi:

pfi79 (approved):
Thank you for correcting the found error.

PR #5433, opened 2026-03-23
The Quorum (GoQuorum) project has been taken over by Consensys

pfi79:
The old link is no longer available.


satota2 (approved):
@pfi79 Thank you for the fix. LGTM.

satota2:
@pfi79 We might also need to update the v2.5 documentation as well.

pfi79:
> @pfi79 We might also need to update the v2.5 documentation as well.

You're right, I probably need to add a check for broken links for 2.5 while it's in LTS.

PR #5434, opened 2026-03-24
2.5 The Quorum (GoQuorum) project has been taken over by Consensys

pfi79:
The old link is no longer available.


satota2 (approved):
@pfi79 Thank you for your fix!

PR #5443, opened 2026-04-04
Feat: implement result reporting for payloads buffer push

FirePheonix:

#### Description

Currently, the Push method in the Gossip PayloadsBuffer returns nothing (a void). 
This results in "silent drops" where payloads (blocks) are rejectedfor example, because they are duplicates or have outdated sequence numbers,without the caller or system operator being notified. 

#### A TODO was left in the codebase suggesting that this should return a boolean to allow for better status reporting and logging.

This PR simply implements the suggested boolean return value for the Push operation to enhance the observability of the Gossip state transfer service.


FirePheonix:
@pfi79 could you please review this one too..

pfi79 (approved):

PR #5472, opened 2026-05-06
Recover from panic in HandleTransaction to prevent peer crash

Jaskirat-s7:
## Summary

When a transaction times out during a range query, the timeout path in `Execute()` closes LevelDB iterators via `CloseQueryIterators()` while the handler goroutine (spawned by `handleMessageReadyState`) is still actively calling `iter.Next()`. This causes a nil pointer dereference in the LevelDB iterator that crashes the peer process.

## Changes

- Added a deferred `recover()` in `HandleTransaction` (`core/chaincode/handler.go`) to catch panics from concurrent iterator access and return an error response instead of crashing the peer
- Added regression tests that verify the panic is recovered and an error response is sent, and that the transaction ID is properly deregistered

## Testing

All 320 existing tests pass, including the 2 new regression tests:

### DIFF
--- a/core/chaincode/handler.go
+++ b/core/chaincode/handler.go
@@ -245,12 +245,41 @@ type handleFunc func(*pb.ChaincodeMessage, *TransactionContext) (*pb.ChaincodeMe
 // returned by the delegate are sent to the chat stream. Any errors returned by the
 // delegate are packaged as chaincode error messages.
 func (h *Handler) HandleTransaction(msg *pb.ChaincodeMessage, delegate handleFunc) {
+	startTime := time.Now()
+	meterLabels := []string{
+		"type", msg.Type.String(),
+		"channel", msg.ChannelId,
+		"chaincode", h.chaincodeID,
+	}
+
+	// Recover from panics that can occur when the transaction context is cleaned up
+	// (e.g., iterators closed) due to execution timeout while this goroutine is still
+	// actively using those resources. This prevents peer crashes from nil pointer
+	// dereferences in the underlying LevelDB iterator.
+	// See https://github.com/hyperledger/fabric/issues/5048
+	defer func() {
+		if r := recover(); r != nil {
+			chaincodeLogger.Errorf("[%s] Recovered from panic handling %s: %v", shorttxid(msg.Txid), msg.Type, r)
+			resp := &pb.ChaincodeMessage{
+				Type:      pb.ChaincodeMessage_ERROR,
+				Payload:   []byte(fmt.Sprintf("%s failed: transaction ID: %s: panic during execution", msg.Type, msg.Txid)),
+				Txid:      msg.Txid,
+				ChannelId: msg.ChannelId,
+			}
+			h.ActiveTransactions.Remove(msg.ChannelId, msg.Txid)
+			h.serialSendAsync(resp)
+
+			meterLabels = append(meterLabels, "success", "false")
+			h.Metrics.ShimRequestDuration.With(meterLabels...).Observe(time.Since(startTime).Seconds())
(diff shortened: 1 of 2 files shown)


core/chaincode/handler.go
```diff
@@ -245,6 +245,25 @@ type handleFunc func(*pb.ChaincodeMessage, *TransactionContext) (*pb.ChaincodeMe
...
+		if r := recover(); r != nil {
+			chaincodeLogger.Errorf("[%s] Recovered from panic handling %s: %v", shorttxid(msg.Txid), msg.Type, r)
+			resp := &pb.ChaincodeMessage{
+				Type:      pb.ChaincodeMessage_ERROR,
+				Payload:   []byte(fmt.Sprintf("%s failed: transaction ID: %s: recovered from panic: %v", msg.Type, msg.Txid, r)),
+				Txid:      msg.Txid,
+				ChannelId: msg.ChannelId,
+			}
+			h.ActiveTransactions.Remove(msg.ChannelId, msg.Txid)
+			h.serialSendAsync(resp)
```
pfi79, line 270: Since you've decided to clean up everything, you need to add metric processing.

core/chaincode/handler.go
```diff
@@ -245,6 +245,25 @@ type handleFunc func(*pb.ChaincodeMessage, *TransactionContext) (*pb.ChaincodeMe
...
+	// (e.g., iterators closed) due to execution timeout while this goroutine is still
+	// actively using those resources. This prevents peer crashes from nil pointer
+	// dereferences in the underlying LevelDB iterator.
+	// See https://github.com/hyperledger/fabric/issues/5048
+	defer func() {
+		if r := recover(); r != nil {
+			chaincodeLogger.Errorf("[%s] Recovered from panic handling %s: %v", shorttxid(msg.Txid), msg.Type, r)
+			resp := &pb.ChaincodeMessage{
+				Type:      pb.ChaincodeMessage_ERROR,
+				Payload:   []byte(fmt.Sprintf("%s failed: transaction ID: %s: recovered from panic: %v", msg.Type, msg.Txid, r)),
```
pfi79: It seems to me that `r` should be output to the logger, but it should not be sent in the response. the `r` in the response may violate determinism.

core/chaincode/handler.go
Jaskirat-s7: i have added ShimRequestDuration and ShimRequestsCompleted recording in the recover path. Also moved startTime before the defer so it's always initialized.

core/chaincode/handler.go
Jaskirat-s7: You're right , i have fixed it, r is now only logged at ERROR level. The response payload uses a generic "panic during execution" message that will be identical across peers.

core/chaincode/handler.go
```diff
@@ -245,12 +245,41 @@ type handleFunc func(*pb.ChaincodeMessage, *TransactionContext) (*pb.ChaincodeMe
...
+				Txid:      msg.Txid,
+				ChannelId: msg.ChannelId,
+			}
+			h.ActiveTransactions.Remove(msg.ChannelId, msg.Txid)
+			h.serialSendAsync(resp)
+
+			meterLabels := []string{
+				"type", msg.Type.String(),
+				"channel", msg.ChannelId,
+				"chaincode", h.chaincodeID,
```
pfi79: can meterLabels be set immediately after startTime and before defer?

Jaskirat-s7: Done , i have moved meterLabels right after startTime, before the defer. The recover path now just appends "success", "false" to the shared slice instead of creating a new one.

pfi79 (approved):

pfi79 (approved):

pfi79:

PR #5484, opened 2026-05-26
chore: harden GHA

ryjones:
#### Type of change

- Improvement to CI security

#### Description

I ran a [zizmor](https://github.com/zizmorcore/zizmor) scan of the actions in this repo and applied some suggested remediations.

#### Related issues

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -42,11 +42,14 @@ jobs:
     runs-on: ubuntu-24.04
     steps:
       - name: Checkout Fabric Code
-        uses: actions/checkout@v5
+        uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
+        with:
+          persist-credentials: false
       - name: Install Go
-        uses: actions/setup-go@v6
+        uses: actions/setup-go@4a3601121dd01d1626a1e23e37211e3254c1c06c # v6.4.0
         with:
           go-version-file: go.mod
+          cache: false
       - name: Compile Binary and Create Tarball
         run: ./ci/scripts/create_binary_package.sh
         env:
@@ -54,7 +57,7 @@ jobs:
           RELEASE: ${{ env.FABRIC_VER }}
 
       - name: Publish Release Artifact
-        uses: actions/upload-artifact@v4
+        uses: actions/upload-artifact@043fb46d1a93c77aae656e7c1c64a875d1fc6a0a # v7.0.1
         with:
           # <name> of the artifact must not collide between platform/arch builds
           name: release-${{ matrix.target }}-${{ matrix.arch }}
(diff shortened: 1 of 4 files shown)


.github/workflows/release.yml
```diff
@@ -42,19 +42,22 @@ jobs:
     runs-on: ubuntu-24.04
     steps:
       - name: Checkout Fabric Code
-        uses: actions/checkout@v5
+        uses: actions/checkout@93cb6efe18208431cddfb8368fd83d5badbf9bfd # v5
+        with:
+          persist-credentials: false
       - name: Install Go
-        uses: actions/setup-go@v6
+        uses: actions/setup-go@4a3601121dd01d1626a1e23e37211e3254c1c06c # v6
```
bestbeforetoday: I think it would be better to include the full version number that the commit hash corresponds to in the trailing comment. If it is correct, this version comment will get updated by dependabot when it updates the action.

```suggestion
        uses: actions/setup-go@4a3601121dd01d1626a1e23e37211e3254c1c06c # v6.4.0
```

bestbeforetoday (commented):
An observation (with example) on the action version trailing comments in-line below. Same applies to all of the changes.

Once all the actions are referenced by commit hash, the **Require actions to be pinned to a full-length commit SHA** setting can be enabled for the repository to ensure they stay that way.

.github/workflows/release.yml
ryjones: I went through and pinned everything explicitly.

ryjones:
I updated the pointer to google/osv-scanner-action to actually be the 2.3.8 commit

.github/workflows/release.yml
```diff
@@ -204,11 +209,16 @@ jobs:
...
+        env:
+          MATRIX_REGISTRY: ${{ matrix.registry }}
 
       - name: Inspect image
         run: |
-          docker buildx imagetools inspect ${{ matrix.registry }}/${{ github.repository_owner }}/fabric-${{ matrix.component.name }}:${{ steps.meta.outputs.version }}
+          docker buildx imagetools inspect ${MATRIX_REGISTRY}/${{ github.repository_owner }}/fabric-${{ matrix.component.name }}:${STEPS_META_OUTPUTS_VERSION}
+        env:
+          MATRIX_REGISTRY: ${{ matrix.registry }}
+          STEPS_META_OUTPUTS_VERSION: ${{ steps.meta.outputs.version }}
```
bestbeforetoday, line 221: I am _guessing_ the change to use environment variables here is to avoid having anything sensitive appear in the command-line recorded in the build log. Instead the command-line shows environment variable names and sensitive information is supplied within those environment variables.

Nothing here is sensitive though. I don't see any obvious harm in this change but I don't see any obvious need for it either.

pfi79 (approved):

PR #5506, opened 2026-06-24
Handle TLS between Peer and CouchDB

mffrench:
Add TLS Support for Peer-to-CouchDB Connections

#### Type of change

- New feature

#### Description

This pull request adds native TLS support for connections between Hyperledger Fabric Peers and CouchDB state databases, addressing a critical security vulnerability where all peer-to-CouchDB communication was previously unencrypted.

**Problem:**
Currently, Fabric peers connect to CouchDB using unencrypted HTTP connections, exposing CouchDB authentication credentials (transmitted in plaintext via HTTP Basic Auth) and the complete blockchain world state data. While documentation recommends co-locating peers and CouchDB on the same VM, this is insufficient for modern cloud-native deployments, particularly in Kubernetes environments where network traffic can be monitored by cloud provider administrators or untrusted parties.

**Solution:**
This PR implements comprehensive TLS support with server certificate verification, configurable TLS settings, and TLS 1.3 minimum version enforcement. The implementation is backward compatible (TLS disabled by default) and includes proper error handling and security logging.

**Key Changes:**
- Extended `CouchDBConfig` structure with TLS fields (`TLSEnabled`, `TLSCACertFile`, `TLSSkipVerify`)
- Modified `createCouchInstance()` to support HTTPS scheme when TLS is enabled
- Added `loadTLSConfig()` function for TLS configuration and certificate loading
- Fixed `couchInstance.url()` method to return https:// when TLS is enabled
- Added configuration mappings and environment variable support
- Updated sample configuration with TLS settings
- Added comprehensive documentation for TLS setup

**Configuration Example:**
```yaml
ledger:
  state:
    stateDatabase: CouchDB
    couchDBConfig:
      couchDBAddress: couchdb.example.com:6984
      username: admin
      password: adminpw
      tls:
        enabled: true
        caFile: /path/to/ca-cert.pem
        skipVerify: false
```

**Security Impact:**
- Encrypts all peer-to-CouchDB communication
- Protects CouchDB credentials from network sniffing
- Prevents unauthorized access to blockchain world state data
- Enables compliance with data protection regulations (GDPR, PCI-DSS, HIPAA, SOC 2)

#### Additional details

**Testing:**
- Verified TLS connection establishment with valid certificates
- Tested CA certificate loading and validation
- Confirmed proper error handling for invalid certificates
- Validated configuration parsing and environment variable mappings
- Tested backward compatibility with TLS disabled
- Verified logging output for security warnings

**Implementation Notes:**
- Uses Go's standard `crypto/tls` package
- TLS configuration loaded once during peer startup for efficiency
- HTTP client and transport reused across all CouchDB operations (thread-safe)
- Comprehensive logging for troubleshooting
- Minimum TLS 1.3 enforcement for security best practices

**Migration Path:**
Existing deployments can migrate to TLS by:
1. Configuring CouchDB with TLS certificates
2. Updating peer configuration to enable TLS
3. Restarting the peer (no data migration required)

#### Related issues

Addresses security vulnerability: Lack of TLS support for peer-to-CouchDB connections exposes sensitive ledger data in cloud/Kubernetes deployments.

#### Release Note

Added TLS support for peer-to-CouchDB connections to encrypt communication and protect sensitive ledger data. New configuration options allow peers to connect to CouchDB over HTTPS with server certificate verification. This feature is particularly important for cloud-native and Kubernetes deployments where network traffic may be accessible to untrusted parties. TLS is disabled by default for backward compatibility. See documentation for configuration details and migration guidance.

### DIFF
--- a/core/ledger/kvledger/txmgmt/statedb/statecouchdb/couchdbutil.go
+++ b/core/ledger/kvledger/txmgmt/statedb/statecouchdb/couchdbutil.go
@@ -8,10 +8,13 @@ package statecouchdb
 
 import (
 	"bytes"
+	"crypto/tls"
+	"crypto/x509"
 	"encoding/hex"
 	"net"
 	"net/http"
 	"net/url"
+	"os"
 	"regexp"
 	"strconv"
 	"strings"
@@ -42,10 +45,19 @@ var (
 )
 
 func createCouchInstance(config *ledger.CouchDBConfig, metricsProvider metrics.Provider) (*couchInstance, error) {
+	// Determine URL scheme based on TLS configuration
+	scheme := "http"
+	if config.TLSEnabled {
+		scheme = "https"
+		couchdbLogger.Info("TLS is enabled for CouchDB connection")
+	} else {
+		couchdbLogger.Warn("Connecting to CouchDB without TLS (unencrypted connection)")
+	}
+
 	// make sure the address is valid
 	connectURL := &url.URL{
 		Host:   config.Address,
-		Scheme: "http",
+		Scheme: scheme,
 	}
 	_, err := url.Parse(connectURL.String())
 	if err != nil {
@@ -55,6 +67,8 @@ func createCouchInstance(config *ledger.CouchDBConfig, metricsProvider metrics.P
 			config.Address,
 		)
 	}
+
+	couchdbLogger.Infof("Connecting to CouchDB at: %s", connectURL.String())
 
 	// Create the http client once
 	// Clients and Transports are safe for concurrent use by multiple goroutines
@@ -74,6 +88,26 @@ func createCouchInstance(config *ledger.CouchDBConfig, metricsProvider metrics.P
 		TLSHandshakeTimeout:   10 * time.Second,
 		ExpectContinueTimeout: 1 * time.Second,
 		DisableKeepAlives:     disableKeepAlive,
+	}
+
+	couchdbLogger.Debugf("HTTP client configuration - RequestTimeout: %v, MaxIdleConns: %d, MaxIdleConnsPerHost: %d, TLSHandshakeTimeout: %v",
+		config.RequestTimeout, transport.MaxIdleConns, transport.MaxIdleConnsPerHost, transport.TLSHandshakeTimeout)
+
+	// Configure TLS if enabled
+	if config.TLSEnabled {
+		tlsConfig, err := loadTLSConfig(config)
+		if err != nil {
+			return nil, errors.WithMessage(err, "failed to load TLS configuration for CouchDB")
+		}
+		transport.TLSClientConfig = tlsConfig
+
+		// Log TLS configuration details at DEBUG level
+		couchdbLogger.Debugf("TLS configuration applied - MinVersion: TLS 1.2, InsecureSkipVerify: %v", tlsConfig.InsecureSkipVerify)
+		if tlsConfig.RootCAs != nil {
+			couchdbLogger.Debug("TLS: Custom CA certificate pool configured for server verification")
+		} else {
+			couchdbLogger.Debug("TLS: Using system CA certificate pool for server verification")
+		}
 	}
 
 	client.Transport = transport
@@ -101,6 +135,49 @@ func createCouchInstance(config *ledger.CouchDBConfig, metricsProvider metrics.P
 	}
 
 	return couchInstance, nil
+}
+
+// loadTLSConfig loads and configures TLS settings for CouchDB connection
+func loadTLSConfig(config *ledger.CouchDBConfig) (*tls.Config, error) {
+	couchdbLogger.Info("Loading TLS configuration for CouchDB connection")
+
+	tlsConfig := &tls.Config{
+		MinVersion: tls.VersionTLS13, // Enforce minimum TLS 1.3
+	}
+
+	couchdbLogger.Debug("TLS minimum version set to TLS 1.3")
+
+	// Configure server certificate verification
+	if config.TLSSkipVerify {
+		couchdbLogger.Warn("TLS certificate verification is disabled - this is insecure and should only be used for testing")
+		tlsConfig.InsecureSkipVerify = true
+		// When skipping verification, don't load CA certificate as it won't be used
+		couchdbLogger.Info("TLS configuration successfully loaded for CouchDB connection")
+		return tlsConfig, nil
+	}
+
+	couchdbLogger.Debug("TLS server certificate verification is enabled")
+
+	// Load CA certificate for server verification
+	if config.TLSCACertFile != "" {
+		couchdbLogger.Debugf("Reading CA certificate from: %s", config.TLSCACertFile)
+		caCert, err := os.ReadFile(config.TLSCACertFile)
+		if err != nil {
+			return nil, errors.WithMessagef(err, "failed to read CA certificate file: %s", config.TLSCACertFile)
+		}
+
+		caCertPool := x509.NewCertPool()
+		if !caCertPool.AppendCertsFromPEM(caCert) {
+			return nil, errors.Errorf("failed to parse CA certificate from file: %s", config.TLSCACertFile)
+		}
+		tlsConfig.RootCAs = caCertPool
+		couchdbLogger.Infof("Successfully loaded CA certificate for CouchDB server verification from: %s", config.TLSCACertFile)
+	} else {
+		couchdbLogger.Debug("No custom CA certificate specified, will use system CA certificate pool")
+	}
+
+	couchdbLogger.Info("TLS configuration successfully loaded for CouchDB connection")
+	return tlsConfig, nil
 }
 
 func checkCouchDBVersion(version string) error {
--- a/docs/source/couchdb_tls_configuration.md
+++ b/docs/source/couchdb_tls_configuration.md
@@ -1,0 +1,207 @@
+# CouchDB TLS Configuration Guide
+
+## Overview
+
+Hyperledger Fabric Peer supports TLS/SSL encrypted connections to CouchDB for secure state database communication. This guide explains how to configure TLS for CouchDB connections.
+
+## Configuration Options
+
+### YAML Configuration (core.yaml)
+
+Add TLS configuration under `ledger.state.couchDBConfig.tls`:
+
+```yaml
+ledger:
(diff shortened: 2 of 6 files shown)


pfi79:
@mffrench Please tell me, did you use AI to generate the changes?

mffrench:
@pfi79 : I used AI and reviewed/tested by myself and team (so more human than AI still) ... 

but AI still very usefull for documenting I think...

sampleconfig/core.yaml
```diff
@@ -688,42 +688,20 @@ ledger:
     # Limit on the number of records to return per query
     totalQueryLimit: 100000
     couchDBConfig:
-      # It is recommended to run CouchDB on the same server as the peer, and
```
mffrench, line 691: there by the way it seems to me we should not remove these parameters... I missed that on my review

mffrench: fixed

pfi79:
> @pfi79 : I used AI and reviewed/tested by myself and team (so more human than AI still) ...
> 
> but AI still very usefull for documenting I think...

Yes, I see that you have modified `sampleconfig/core.yaml`. Thank you. 
The fact is that if a human had done this, they would never have allowed the deletion of a part of the `sampleconfig/core.yaml` file without explanation.
I came to the conclusion that the AI made the change, but the person who created the PR did not check it.

mffrench:
> > @pfi79 : I used AI and reviewed/tested by myself and team (so more human than AI still) ...
> > but AI still very usefull for documenting I think...
> 
> Yes, I see that you have modified `sampleconfig/core.yaml`. Thank you. The fact is that if a human had done this, they would never have allowed the deletion of a part of the `sampleconfig/core.yaml` file without explanation. I came to the conclusion that the AI made the change, but the person who created the PR did not check it.

@pfi79 : well I'm not sure. It could be myself while I was reporting some changes... I didn't found that change in my initial PR in our own Fabric fork... Any other feedbacks which can help improve that PR to make it merged ?

jimthematrix:
This is a reasonable requirement and the code updates look pretty good to me. Any chance one of the existing committers can help with review/approve and merge it?

docs/source/couchdb_tls_configuration.md
pfi79, line 1: From where, from which page, will the user have to go here?
So far, this documentation page is not linked anywhere.

sampleconfig/core.yaml
```diff
@@ -724,6 +724,20 @@ ledger:
...
+        # Environment variable: CORE_LEDGER_STATE_COUCHDBCONFIG_TLS_ENABLED
+        enabled: false
+        # Path to CA certificate file for verifying the CouchDB server certificate
+        # If not provided, the system's default CA certificate pool will be used.
+        # Environment variable: CORE_LEDGER_STATE_COUCHDBCONFIG_TLS_CACERTFILE
+        caCertFile:
+        # Skip server certificate verification (default: false)
+        # WARNING: This is insecure and should only be used for testing purposes.
+        # Environment variable: CORE_LEDGER_STATE_COUCHDBCONFIG_TLS_SKIPVERIFY
+        skipVerify: false
```
pfi79, line 740: The tls setting style already exists in the project (see core.yaml and orderer.yaml). Try to do it in the same style. If this is not possible, please explain why.

core/ledger/kvledger/txmgmt/statedb/statecouchdb/couchdbutil.go
```diff
@@ -42,10 +45,19 @@ var (
 )
 
 func createCouchInstance(config *ledger.CouchDBConfig, metricsProvider metrics.Provider) (*couchInstance, error) {
+	// Determine URL scheme based on TLS configuration
+	scheme := "http"
+	if config.TLSEnabled {
+		scheme = "https"
+		couchdbLogger.Info("TLS is enabled for CouchDB connection")
```
pfi79, line 52: You have 6 places in the code where INFO level logging is used. Consider making them DEBUG

pfi79 (changes requested):

sampleconfig/core.yaml
mffrench: I'm not sure I understand the setting style you're speaking about. Here we want to set client TLS only to the couchdb and not both client & server. Is it the `caCertFile:` you would like to  change to something like: 

```
    # X.509 certificate used for TLS when making client connections.
    # If not set, peer.tls.cert.file will be used instead
    clientCert:
      file:
```

?

core/ledger/kvledger/txmgmt/statedb/statecouchdb/couchdbutil.go
mffrench: will do

docs/source/couchdb_tls_configuration.md
mffrench: I don't know... Maybe should it be accessible the same way https://github.com/hyperledger/fabric/blob/main/docs/source/couchdb_tutorial.rst is accessed by the user maybe ? Note however I acceded that doc by browsing `docs/source` directory only. I didn't know each doc in there was linked to some main doc and I'll be interested to know where is that main doc.

sampleconfig/core.yaml
mffrench: Or is it the skyVerify you would like to remove ? I would say that can be an interesting option for some setups and I would advocate to keep it.

`tls.enabled` looks already be reused from the 'tls style' ... 

Back to my previous comment about the `caCertFile`: `clientCert` example is not equivalent here. This the client certificate of the TLS client, not the certificate of the CA trusting the couchdb TLS identity. Also, we dont need Couchdb Client TLS identity because CouchDB doesn't handle mTLS. So finally I would tend to keep it that way. 

Can you be please a bit more explicit if you think there is a blocker on the current proposal ?

docs/source/couchdb_tls_configuration.md
pfi79: See. Documentation is generated from the docs folder. And by searching or clicking on links, you can get to every page. The search is OK, let's leave it. But clicking on the links is necessary.  

In other words, there must be a transition from somewhere to the page you are creating. This means that you need to add a link to your page to an existing page.

Look around, there are many examples.

sampleconfig/core.yaml
pfi79: Here are some examples. If you look closely, they are all made in the same style, both clients and servers. I would like your change to be similar to these.  
Sorry, I don't really like variety.

example of a tls section in core.yaml
```
  # TLS configuration for the operations endpoint
  tls:
    # TLS enabled
    enabled: false

    # path to PEM encoded server certificate for the operations server
    # The paths in this section may be relative to FABRIC_CFG_PATH or an absolute path.
    cert:
      file:

    # path to PEM encoded server key for the operations server
    key:
      file:

    # most operations service endpoints require client authentication when TLS
    # is enabled. clientAuthRequired requires client certificate authentication
    # at the TLS layer to access all resources.
    clientAuthRequired: false

    # paths to PEM encoded ca certificates to trust for client authentication
    clientRootCAs:
      files: []
```

example of tls section in orderer.yaml
```
  # TLS configuration for the admin endpoint
  TLS:
    # TLS enabled
    Enabled: false

    # Certificate is the location of the PEM encoded TLS certificate
    Certificate:

    # PrivateKey points to the location of the PEM-encoded key
    PrivateKey:

    # Most admin service endpoints require client authentication when TLS
    # is enabled. ClientAuthRequired requires client certificate authentication
    # at the TLS layer to access all resources.
    #
    # NOTE: When TLS is enabled, the admin endpoint requires mutual TLS. The
    # orderer will panic on startup if this value is set to false.
    ClientAuthRequired: true

    # Paths to PEM encoded ca certificates to trust for client authentication
    ClientRootCAs: []
```

tls example from osnadmin
```
	if *caFile != "" {
		osnURL = fmt.Sprintf("https://%s", *orderer)
		var err error
		caCertPool = x509.NewCertPool()
		caFilePEM, err := os.ReadFile(*caFile)
		if err != nil {
			return "", 1, fmt.Errorf("reading orderer CA certificate: %s", err)
		}
		if !caCertPool.AppendCertsFromPEM(caFilePEM) {
			return "", 1, errors.New("failed to add ca-file PEM to cert pool")
		}

		tlsClientCert, err = tls.LoadX509KeyPair(*clientCert, *clientKey)
		if err != nil {
			return "", 1, fmt.Errorf("loading client cert/key pair: %s", err)
		}
```

docs/source/couchdb_tls_configuration.md
pfi79: Alternatively, you can add changes to the documentation to an existing page. Maybe shorten the text.  
Sorry, I haven't read the gist of the documentation yet.

sampleconfig/core.yaml
mffrench: I did read these examples already but I'm personally not agree to add configuration complexity on a use case which is much more simpler than Peer and Order TLS setup which handle both server TLS setup and client mTLS  setup. HLF Peer to CouchDB connection is about configuring a client on the Peer for a simple TLS connection to a CouchDB server (and with no mTLS as CouchDB doesn't handle it).

Again, here we are configuring an HTTPs client only and we are requiring only 2 mandatory param: 

			TLSEnabled:  wether or not we activate TLS between CouchDB & Peer
			TLSCACertFile:  the CA cert file of the _CouchDB_ server to make work TLS handshake

The last one is to reduce complexity on dev environment in some scenario : 

			TLSSkipVerify:         viper.GetBool("ledger.state.couchDBConfig.tls.skipVerify"),
			
I dont see reasonable rational to add extra params & complexity to follow some examples which do not match the use case. 

While I understand you don't like variety, please understand I don't like adding useless complexity : please explain the rational to increase code change complexity in that use case.

pfi79: Fabric is a very complex system. 
In my opinion, adding a new tls assignment method, unlike the previous methods, complicates the entire system.
And I think it's ugly, but monotonous.

mffrench: Still bad rationals from my perspective.
 
The complexity you're complaining is very well located from DEV perspective (couchdb conf (3 new parameters) and statecouchdb package (~150 new code lines)). 

No reason to handle TLS configuration CouchDB is not even able to handle (mTLS) and OPS will not understand the purpose. 

Will not do.

Issue #5504, opened 2026-06-22
Send auth credentials to the Docker socket if Docker config file is mounted


waduhek:
### Current Status

Currently, if a Docker config file were to be mounted to the peer containers to either `/root/.dockercfg` or `/root/.docker/config.json` with credentials for a private repository, `fabric-peer` won't send the configured credentials to the socket. This becomes an issue when the peer tries to pull the chaincode build environments (`fabric-baseos`, etc.) from a private repository.

### Expected

If a Docker config file has been mounted to the peer, the peer should check if auth credentials are provided in the config file and send them to the socket.

### Please let us know if you plan to work on this.

Yes!

pfi79:
what is your version of the fabric? 
has this problem just appeared, or has it always been?

waduhek:
Hi @pfi79, we are currently using 2.4.7. We've also tested with 2.5 which also gave the same result. Seems like it has always been this way. Our initial debugging helped us narrow down to the `buildImage` function in [`core/container/dockercontroller/dockercontroller.go`](https://github.com/hyperledger/fabric/blob/86c1172bec37e6590c6b1ae6f847833df48d65fc/core/container/dockercontroller/dockercontroller.go#L111). The function doesn't seem to pick the credentials from the config file and provide it in the build options.

dpereowei:
Hi, I’d like to work on this.
I confirmed that the current Docker chaincode builder calls `ImageBuild` without populating `ImageBuildOptions.AuthConfigs`, so credentials from the peer’s mounted Docker configuration are not sent to the Docker daemon.
I propose loading static registry credentials from the standard Docker configuration locations, including `~/.docker/config.json` and `~/.dockercfg`, and passing them through `AuthConfigs` during chaincode image builds. I’ll add focused tests covering both formats, missing configuration, and malformed configuration while ensuring credentials are never logged.
Please let me know if there are additional compatibility requirements you would like included, such as Docker credential-helper support.

PR #5507, opened 2026-06-28
Use auth configs provided in Docker config file

waduhek:
#### Type of change

- New feature

#### Description

Detects a mounted Docker configuration file in either `$HOME/.dockercfg` or `$HOME/.docker/config.json` and parses it for authentication information. If either one of the files contains authentication information, the details are sent over to the Docker client library to be sent to the Docker socket for building the chaincode image.

#### Additional details

I ran the updated test files but I currently do have access to a private repository against which I can test the changes.

#### Related issues

See #5504.

### DIFF
--- a/core/container/dockercontroller/dockercontroller.go
+++ b/core/container/dockercontroller/dockercontroller.go
@@ -118,14 +118,12 @@ func (vm *DockerVM) buildImage(ccid string, reader io.Reader) error {
 		return err
 	}
 
-	authConfigs := vm.getAuthFromDockerConfig()
-
 	startTime := time.Now()
 	res, err := vm.Client.ImageBuild(context.Background(), reader, dcli.ImageBuildOptions{
 		Tags:        []string{id},
 		PullParent:  vm.ChaincodePull,
 		NetworkMode: vm.NetworkMode,
-		AuthConfigs: authConfigs,
+		AuthConfigs: getAuthFromDockerConfig(),
 	})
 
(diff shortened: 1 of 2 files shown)


waduhek:
@pfi79, please have a look at my PR when possible. I'd also like to request some help in testing these changes on an actual private repository. I tried spinning up a local Nexus repository but I'm having some trouble on that front since Fabric is attempting to use HTTPS which I couldn't setup yet. If I do succeed in testing these changes on a private repository, I'll report back.

waduhek:
Update on testing this change on a private repository. I tried testing using a private repository on Docker Hub and found that the same issue persists. Moving this PR to draft mode while I work on getting it to work.

Here's what I did: I pushed a re-tagged version of `fabric-baseos` to the private repository, updated the `core.yaml` to use this private image and started the test network. Then, I tried deploying the asset transfer basic chaincode which resulted in an error saying that pull access is denied for the image.

waduhek:
I've managed to figure out how authentication works with the Docker engine. Turns out, the Docker engine does not pick up the base-64 encoded "auth" string from the config file. It only recognises "username" and "password" fields for a particular registry. The PR has been updated to reflect this. Additionally, I've also removed support for the `~/.dockercfg` file since Docker CLI does not use this file and I'm trying to keep this change as close to the behaviour of Docker CLI as possible.

I've also managed to test this successfully by using the same method I've mentioned in my previous comment. Marking this PR as ready for review.

PR #5544, opened 2026-08-26
Fix OSV

ryjones:
#### Type of change

- Test update

#### Description

This updates go packages to non-vulnerable versions, which allows osv to run.


pfi79 (commented):
1. golang.org/x/crypto we also have go-swagger in our tools.  In the comments to the exclusion, you don't mention it.
 
2. Do you think we can't wait for all the libraries we use to get rid of golang.org/x/crypto ?

ryjones:
ok

PR #5558, opened 2026-09-10
Use protobuf getters to access member variables

bestbeforetoday:
Use only protobuf getter methods to access member variables on protobuf messages. This protects against panics if accessing members of a nil message, which the getter handles gracefully by returning nil or a default value for a nil receiver.


pfi79:
have you used `protogetter --fix ./...` ?

bestbeforetoday:
Effectively yes. I used golangci-lint configured with just the protogetter linter and `--fix`.

I considered adding a simple golangci-lint lint check as part of this change but your golangci-lint change is already in flight so thought it better to wait for that to go in.

pfi79 (approved):

